PR #41, opened 2019-04-11
Fix minor freeones scraper bug

bnkai:
Stash
i found out that stash by default listens to all available interfaces , even the ipv6 one that can be a bit problematic.
I added a few extra options for the config file
**address**
**port**  
**httpsport**
that are optional and as non invasive as i could(created a new viper instance just for the server config which only reads the conf)
By default without having any options it behaves as before


StashAppDev:
Thanks for the contribution.  I actually went a bit different route with [this commit](https://github.com/stashapp/stash/commit/29a40c5bdd476587843f9946c0cc4807e00bee2c) which makes most of this work unnecessary at this point.  I'd like to get the freeones fix in though if you want to simplify the change to just that.

bnkai:
It's all good stash , i had the codde for my custom build anyway.

PR #44, opened 2019-04-19
Dockerfile for production.

Leopere:
Since GitHub thrashed my PR message body I'll re-write it in post submission.

This is a working Dockerfile and a few other untested objects.  A final version requires that Stash has environment variable configuration alternatives to core config options so that it can be stateless and low maintenance.


Leopere:
Things like these should be added ideally; however at this point, it could be using it won't do anything statelessly until it has environment variable configuration options.

https://github.com/bnkai/stash/commit/faf1c4b82fa71728c005488ae541637d6ead2178
https://github.com/bnkai/stash/commit/3b8913b563e72f6db8bf8f32e81a649c0301e453

Leopere:
Assumptions that are made here are that eventually, someone will add an automatic Docker container build at https://hub.docker.com/r/stashapp/stash:1804_x86_64

bnkai:
@Leopere  In your compose file `STASH_STASH=/stash/` should be changed to `STASH_STASH=/data/` or otherwise change the volume directive for data.

Leopere:
This runs now.

Issue #48, opened 2019-04-28
ffmpeg error when generating sprites and previews


rudi123github:
When I generate previews and spites, I get this kind of error on some vids:

ERRO[3587] ffmpeg error when running command </usr/bin/ffmpeg -v quiet -ss 1180 -t 0.75 -i /home/user/Desktop/sort/Goga-Huxley ++.wmv -y -c:v libx264 -pix_fmt yuv420p -profile:v high -level 4.2 -preset veryslow -crf 21 -threads 4 -vf scale=640:-2 -c:a aac -b:a 128k -strict -2 /home/ziggy/stash/generated/tmp/preview010.mp4>

I ran the command manually from bash, and it does work if I add the option "-max_muxing_queue_size 400" to ffmpeg.

rudi123github:
Ok, I figured out that the change needs to happen in pkg/ffmpeg/encoder_scene_preview_chunk.go. I added the same parameter I found in another module as extra argument to function ScenePreviewVideoChunk:

"-max_muxing_queue_size", "1024", // https://trac.ffmpeg.org/ticket/6375

I rebuilt stash with the docker container and now everything runs fine locally. I cannot push the update to github, so I ask one of the devs to make this change please.

Leopere:
@rudi123github did you still get this issue or can it be closed?

rudi123github:
I fixed this in my local build but the change should go into the development branch.

Leopere:
If you file a PR it can get reviewed / worked on.  As far as I'm aware this app is far from 1.0 status but whatever works.

rudi123github:
Okay, I just figured out how to do that. The PR is there now.

ExceptionalError:
I am still getting errors when trying to generate preview images.
Since my filename contains spaces, I am not sure if it should be quoted.

Can anyone confirm it works with spaces?
Also the ````-max_muxing_queue_size 1024```` is missing.
````
msg="ffmpeg error when running command </usr/bin/ffmpeg -v quiet -ss 1562 -y -i /data/MyVideo [x264].mp4 -vframes 1 -q:v 1 -vf scale=160:-1 -f image2 /generated/tmp/thumbnail071.jpg>"
````

rudi123github:
Hello, this is a different function. Can you please run this from a terminal:

[...]

PR #64, opened 2019-06-21
Update encoder_scene_preview_chunk.go

rudi123github:
Apply ffmpeg max_muxing_queue_size option fix to avoid problems with some videos


Leopere:
fixes #48

Leopere:
@rudi123github does this break anything?

bnkai:
it shouldn't break anything, the ffmpeg settings changed only affect the sprite generation, i would wait though for stash to merge it himself.
The  value though (1024) needs more extended testing just to be sure.

rudi123github:
We use the same option already in function "SceneMarkerVideo". So I think it is safe and does not break anything.

Leopere:
Accepted thanks!

Issue #62, opened 2019-06-02
[Request] Add support to play files in external video player?


Cutters1982:
Hi there

Really liking the app so far, but the internal video player and having to transcode multiple files is a real pain, and it will take up a lot of disk space.

Was wondering if you plan to allow support to play the video files externally via VLC/MPCHC which would eliminate the need in having to transcode the whole video? Or is that defeating the objective of what you want to do with this app?

I was hoping that maybe the app could act as a hub/media server, but the files play externally, but you would still have the little animated video previews etc)

Cheers

rudi123github:
Similar to issue #52

Leopere:
I think the idea is ultimately to attempt to leave the original file alone and do live transcode if needed to the browser.

If you consider this and the previous comment as a sufficient answer please close the ticket.

rexii2300:
Yes I would love this feature. A large number of my files are mp4/flv encoded in mpeg4 instead of h264 and non of them are able to be played. All I see is black with the seekbar moving.

Leopere:
Its a good feature if it gets submitted as a PR however it wouldn't make
sense for us to have a streaming app that doesn't stream.  We're just
currently waiting on @stash to get time to look into live transcoding which
seems to be the best solution.

rexii2300:
Live transcoding would be fantastic!

How is everyone handling the problem right now? Is it best just to encode them all to h264 in mp4 container? sounds like a huge task with a big library?

Leopere:
The other project that I would look at is https://github.com/JellyFin/JellyFin that is not really ideal for this kind of content but its a very mature streaming app.

rexii2300:
jellyfin looks very polished for a steaming app. Most people seem to be recommending Plex or Emby.

The application of Stash can potentially be really wide. I was looking for a Lightroom equivalent for videos. Surprisingly nothing is out there. Most recommended Bridge but it is still clunky and metadata handling is horrible. Some more 'pro' level apps such as Kyno writes to a XML sidecar file and is pretty Spartan in terms of features. A non-adult version of Stash could be revolutionary for managing video content.

I'm a big fan of transcoding and streaming. Its magical when JW player works. Transcoding might be an issue for people that runs this in docker inside VMs. I see the two biggest strength of Stash as cataloging and streaming. Having the simple option and flexibility of play it via an external player will give people choices and only bring more people to Stash.
________________________________

Leopere:
[...]

Issue #52, opened 2019-05-01
Add stream link to scene web page


rudi123github:
Feature reqest: Add stream link to scene page. Some media formats cannot be played by the JW player. However, the stream URLs play perfectly in a media player such as VLC and mpv. If the scene page has a link to the stream, the video can easily be opened in a media player. At the moment, it is too complicated to get the stream URL (e.g. http://127.0.0.1:8080/scene/49/stream.mp4) out of stash to view a vid that is not supported by JW player.

rudi123github:
I created a pull request for this.

WithoutPants:
This looks like it can be closed. @Leopere

Leopere:
Fixed by #65

Leopere:
Closed by #65 and resolves #62

PR #65, opened 2019-07-02
Add anchor to scene path, and display stream anchor

rudi123github:
This allows to play videos which are not supported by JWplayer. Stash will display links to the local file and the stream in the "File Info" panel. This allows to open the stream or local file directly in the browser (if supported) or the app that is registered for the respective file type (e.g. VLC).

This addresses issues #52 and #62 (which are basically duplicates)

### DIFF
--- a/ui/v2/src/components/scenes/SceneDetails/SceneFileInfoPanel.tsx
+++ b/ui/v2/src/components/scenes/SceneDetails/SceneFileInfoPanel.tsx
@@ -23,7 +23,16 @@ export const SceneFileInfoPanel: FunctionComponent<ISceneFileInfoPanelProps> = (
     return (
       <tr>
         <td>Path</td>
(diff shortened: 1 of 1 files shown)


Leopere:
I think that the plan is to have the server side use FFMPEG to do live transcoding for that kind of stream at some point but this might be nice in general so long as it's got a CLI flag like --verbose was just PR'd recently.  I know that this thing isn't designed to host content on the internet but being able to switch on and off additional features would be nice.

bnkai:
Imho the specific feature could be a an extra tick option in the interface configuration (like Display titles and tags) at least for the short term untill live transcoding is possible.CLI options are convenient but mainly for the backend.

bnkai:
The local file link is only valid if stash is running on your local machine obviously but the the stream link works fine :-) and is indeed very useful for wmv , hevc etc media files.

PR #79, opened 2019-07-25
Add basic live transcoding for non-streamable files

WithoutPants:
Adds very basic live transcoding to allow playing of files that are not directly streamable.

Kills the ffmpeg process when it detects that the connection is closed.

Does not currently support seeking or even resuming a running stream (it restarts from the start of the video) because of the lack of byte range support. 

Currently encodes to webm with a fixed crf. 

It's a very basic start, but it's better than nothing. Assuming no one else looks into this in the short term, I'll look into seeking next, but based on my previous experience, it's a non-trivial problem. It would probably also be worth caching transcoded files so that we can retrieve and play them, rather than re-transcode them each time.

### DIFF
--- a/pkg/api/routes_scene.go
+++ b/pkg/api/routes_scene.go
@@ -1,12 +1,14 @@
 package api
 
 import (
+	"io"
 	"context"
 	"github.com/go-chi/chi"
 	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/manager"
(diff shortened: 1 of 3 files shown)


StashAppDev:
I know I wont be able to look into this in the short term.  I think seeking is a pretty important thing, so it would be good to look into that.  I'll go ahead and merge this though since I agree it's better than nothing.

Leopere:
Elaboration on why Squash & Rebase makes sense. https://archive.fo/K3F4H

PR #84, opened 2019-07-30
Add seeking for live transcoding via video.js

WithoutPants:
Uses video.js for non-streamable files. It overrides the duration to use the duration from the scene info, and overrides the currentTime function to support streaming. Back-end stream.mp4 endpoint was updated to accept a start time, which modifies the ffmpeg call accordingly.

I've chosen to use video.js because jwplayer doesn't appear to allow the required overrides for duration and seeking. jwplayer is still used where the file can be streamed directly.

### DIFF
--- a/pkg/api/routes_scene.go
+++ b/pkg/api/routes_scene.go
@@ -1,17 +1,18 @@
 package api
 
 import (
-	"io"
 	"context"
+	"io"
+	"net/http"
+	"strconv"
+	"strings"
+
 	"github.com/go-chi/chi"
+	"github.com/stashapp/stash/pkg/ffmpeg"
 	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/manager"
 	"github.com/stashapp/stash/pkg/models"
 	"github.com/stashapp/stash/pkg/utils"
-	"github.com/stashapp/stash/pkg/ffmpeg"
-	"net/http"
-	"strconv"
-	"strings"
 )
 
 type sceneRoutes struct{}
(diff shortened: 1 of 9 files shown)


bnkai:
I tried that out and seeking seems to work well
might i suggest adding `"-row-mt", "1",`to the encoder_transcode fmpeg options for  vp9 ?
It enables multithreading which in my case was needed to avoid stuttering

WithoutPants:
I've just encountered a bug that causes an error when navigating to a streamable scene after viewing a transcoded scene. Will look into it further. I'll add the multi-threading change as well.

bnkai:
I think i needed to refresh the page for the player to transition from video.js to jwplayer for the streamable. I'll have a look later

WithoutPants:
Would probably prefer to set scaling as a configuration option rather than hard-coding it, since performance is likely to be conditional on the server machine.

bnkai:
I noticed on some avi,wmv videos that video.js is used seeking is fine 
In a few hevc mkvs i tried though the player used (looks like jw don't know though with the hack you added ?) seems to be missing the duration and so seeking doesn't work

bnkai:
example file 
https://www.h265files.com/ tears of steel
While avi files or wmv files i tried work ok for the above file and some more mkvs using hevc the seeking doesn't work. 
The media player shown for the avi,wmv files looks like the one from video.js but when i play the hevc file the one used is jw player and the file is transcoded without the seeking abilty.

WithoutPants:
> example file
> https://www.h265files.com/ tears of steel
> While avi files or wmv files i tried work ok for the above file and some more mkvs using hevc the seeking doesn't work.
> The media player shown for the avi,wmv files looks like the one from video.js but when i play the hevc file the one used is jw player and the file is transcoded without the seeking abilty.

Changed `IsStreamable` to return based on video codec instead of MIME type. This addresses the issue based on my testing.

Leopere:
At this point it sounds like its up to @bnkai to give a thumbs up and I can merge.

bnkai:
Everything seems in order.No problems with with hevc anymore

Issue #83, opened 2019-07-30
Setting general configuration sometimes wipes cache directory config


WithoutPants:
This is a (frustratingly) intermittent issue. It doesn't happen every time, and I've found restarting the server sometimes works to reproduce the issue.

To reproduce:
* navigate to Settings -> Configuration
* change Generated Path to a different value and click Save
* refresh the page

The server redirects to the setup page. With some logging, I've determined that the cache config setting is being wiped out (`viper.IsSet(cache)` returns false). What's worse is that sometimes the corrupted config is written to the config.yml file.

This looks like a [race condition bug in viper](https://github.com/spf13/viper/issues/174) due to the use of `WatchConfig`. I'd suggest removing the `WatchConfig` call until the bug is resolved in viper, since it makes modifying the configuration in the UI dangerous.

WithoutPants:
I have little faith that `WatchConfig` will be fixed upstream, and I don't think live config reloading is that essential to stash. Closing.

PR #85, opened 2019-07-30
Basic password security #3

WithoutPants:
Addresses #3 

Uses basic authentication. Added fields to the configuration settings tab for username and password. If both values are non-empty, then authentication is performed. Stores the username and password hash in the config file.

Includes a change to SettingsConfigurationPanel.tsx to use `value` instead of `defaultValue` since the latter was causing the settings to become unsynced (I cleared a field and the internal model wouldn't update its value).

Note that due to #83 it's dangerous to change the settings in the configuration settings tab. I fixed this by removing the watch on the config file. I can add that change to this branch if required.


Leopere:
Thanks for using bcrypt

Leopere:
What does the auth protect specifically?  Is it for the configuration or the entire app?

bnkai:
From what i understand it protects the app when it  starts? If we authenticate once it keeps running unprotected? I couldn't find any session cookies in the code ?

WithoutPants:
It protects the entire app. Basic authentication works by setting the credentials once in the browser. While the browser is open, those credentials are kept. If the required credentials change, then the user must relogin with the new credentials. Otherwise, the user is kept logged in until they close the browser.

bnkai:
Cool. I missed the call to basic auth....
I'll give it a try tonight.

bnkai:
Just tried it and it works good like you said.

WithoutPants:
Resolved conflicts. Ready to retest.

bnkai:
I will test later allong with the other PRs

bnkai:
Update
Retested!
Everything works fine ....

WithoutPants:
I'm not sure where to document this:

Be aware that having user credentials prevents the node debugger version of the UI (on port 3000) from accessing the system due to CORS issues. Removing the username and password fixes the issue.

Leopere:
I think that at the very least @WithoutPants documenting things in the README.md is helpful.

PR #91, opened 2019-08-02
Update doublestar to fix #38

### LINKED ISSUE #38: Content locations using Windows UNC paths are inaccessible
Just documenting this issue per our discussion on Discord.

> I spent an hour or so looking into UNC paths, kind of hit a wall.  The first problem was the fact that the library I'm using for globbing (doublestar) doesn't seem to support UNC paths.  I tried removing that and writing the logic myself, but for some reason I'm unable to access any of the folders within the share.  I can see the share itself and its contents, but none of the folders inside the share.  I get an error saying The system cannot find the path specified. even though it's there and I can access it.

Content locations using Windows UNC paths, //servername/folder, are inaccessible to the app.

Windows users will have to map network locations to a drive either using the [file explorer](https://support.microsoft.com/en-ca/help/4026635/windows-map-a-network-drive) or [command line method](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/gg651155(v%3Dws.11)) until the issue is resolved .

WithoutPants:
Resolves #38 

Updates doublestar to v1.1.5, which fixes UNC paths on Windows.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/stashapp/stash
 require (
 	github.com/99designs/gqlgen v0.9.0
 	github.com/PuerkitoBio/goquery v1.5.0
-	github.com/bmatcuk/doublestar v1.1.1
+	github.com/bmatcuk/doublestar v1.1.5
 	github.com/disintegration/imaging v1.6.0
 	github.com/fsnotify/fsnotify v1.4.7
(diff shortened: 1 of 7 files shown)


StashAppDev:
Dope.  Made a PR for the library and everything, that's awesome.  Thanks a lot.

Issue #80, opened 2019-07-28
Apply tags and metadata to search results


koala640:
to quickly bulk tag you should be able to apply tags to search results:

1 - search for something
2 - apply tag or metadata like studio, performer (select one more more tags/performers/etc. or create new ones)
3 - when clicking ok it applies the tags to all content in search results (if already exists then ignore)

this is a good workaround to more complex bulk tagging

GernBlanston12:
I would add, just for giggles, that this sort of "add x tag to everything I see on the screen/search results" functionality would likely work well when implementing playlists, if that is ever a thing. If you've got a file list to display, and I imagine you do, it's just syntax to create/append an .asx file or a playlist.  I think the bulk tag function would be easier to implement than filename-based auto-tagging, and be the next best thing, but I do think filename-based is also hugely important.

Leopere:
This is mostly resolved with #92

PR #92, opened 2019-08-02
Bulk update scenes

WithoutPants:
Adds checkboxes to the scene grid view. Clicking one or more checkboxes shows a bulk-update row which allows setting Rating, Studio, Performers and Tags. Server-side API is added.

Need to iron out a few bugs (mostly due to my inexperience with React):
- ~bulk update row sometimes doesn't hide when everything is de-selected~ [fixed]
- ~fields are supposed to update based on what is selected - ie if all selected scenes have the same tags, then those tags should be populated in the bulk update field. It works for the first selected scene, but not the subsequent ones~ [fixed]

Would also like ~select all/none buttons~ [added], and ~add shift-click shortcut to select multiple scenes~ [added].

### DIFF
--- a/graphql/documents/mutations/scene.graphql
+++ b/graphql/documents/mutations/scene.graphql
@@ -35,8 +35,8 @@ mutation BulkSceneUpdate(
   $rating: Int,
   $studio_id: ID,
   $gallery_id: ID,
-  $performer_ids: [ID],
-  $tag_ids: [ID]) {
+  $performer_ids: [ID!],
+  $tag_ids: [ID!]) {
 
   bulkSceneUpdate(input: {
                         ids: $ids,
--- a/graphql/documents/mutations/scene.graphql
+++ b/graphql/documents/mutations/scene.graphql
@@ -26,6 +26,34 @@ mutation SceneUpdate(
   }
 }
 
+mutation BulkSceneUpdate(
+  $ids: [ID!] = [],
+  $title: String,
+  $details: String,
+  $url: String,
+  $date: String,
+  $rating: Int,
+  $studio_id: ID,
+  $gallery_id: ID,
+  $performer_ids: [ID!] = [],
+  $tag_ids: [ID!] = []) {
+
+  bulkSceneUpdate(input: {
+                        ids: $ids,
+                        title: $title,
+                        details: $details,
+                        url: $url,
+                        date: $date,
+                        rating: $rating,
+                        studio_id: $studio_id,
+                        gallery_id: $gallery_id,
+                        performer_ids: $performer_ids,
+                        tag_ids: $tag_ids
+                      }) {
+      ...SceneData
+  }
+}
+
 mutation SceneDestroy($id: ID!, $delete_file: Boolean, $delete_generated : Boolean) {
   sceneDestroy(input: {id: $id, delete_file: $delete_file, delete_generated: $delete_generated})
 }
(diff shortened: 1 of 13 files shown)


WithoutPants:
Addresses #80

WithoutPants:
Currently breaks the Create Tag from multiselect behaviour.

bnkai:
From a quick test adding tags,performers and rating to scenes using the checkboxes seems to work fine.
Doing the same for studio doesn't seem to make any changes even though the update scenes message is shown

bnkai:
@WithoutPants can you merge with master before i retest?

bnkai:
@WithoutPants 
Studios -> None can't be selected (so we cant unset studios)
Rating -> _ can be selected but doesn't apply correctly

Also when you multiselect and choose a studio and then apply it you can't choose the (No Selection) in the studio tab again unless you deselect and refresh the page

WithoutPants:
@bnkai 
Should be addressed with the latest commit.

bnkai:
I see the following bugs.

1. Select 6 scenes for example , apply a rating or a studio
    Deselect the 3 of the 6 for example and apply a different/rating studio
   Changes are applied to the 6 scenes not the 3 selected

2. Applying the empty rating (unset) only works once. For it to work again you need to refresh the page
example select a number of scenes.
apply rating 1-> ok
apply raiting null -> ok
apply raing 2 -> ok
apply null nothing happens
apply rating 3 ->ok

3. Something similar to 2 happens with studios but i can't reproduce it 100% or at least i can't find the exact sequence that it fails

bnkai:
With the latest commit everything seems to work as it should ..
If there is a bug left it remains to be found :-)
Ok for merge i think.

Leopere:
This just requires conflicts to be resolved and merging should be good to go.

WithoutPants:
The UI for this change is still flaky on my end. I'm going to need some time to rework it.

WithoutPants:
I burned the UI changes (mostly) to the ground and re-implemented. Cosmetically nothing has changed, but I'm hoping that it is more stable. Please retest when you get the chance.

bnkai:
With the latest changes there is no way for to select a few scenes with different performers and same studio and set the studio name leaving the performers alone

```

Example use case 

- filter scenes by lets say dontbreakme

- select some scenes (some with perfomers some without)

- from the drop down select dontbreakme from the studios , empty on the performers
```
**result**

```
scenes get the studio correctly but the performers are gone
same thing happens with rating or tag instead off studio
```

Selecting a bunch of scenes to set the performer without messing with the studio (no selection) works ok

WithoutPants:
The performers/tags bug was a latent bug that existed in the older version as well. Should now be fixed.

bnkai:
Tests also ok for me now

WithoutPants:
After trying to solve this problem many different times, I'm confident that this is at last put to bed. The issue was that graphql unmarshals array fields into a non-nil, empty slice, so there was no way to determine if a field was left unset, or was set specifically to an empty list. I've implemented a server-side check to see if a field is left unset and the field is ignored appropriately.

This addresses the case where selected scenes have differing performers and tags and a different field was set - leaving performers and tags unset. The expected behaviour is that performers and tags are left unchanged, and only the set fields are modified.

bnkai:
Ok i also think we are done here .I tested with evey combination i could think and it checks out ok.

PR #97, opened 2019-08-12
Add delete tag button #96

### LINKED ISSUE #96: Deleting tags?
I accidentally created a tag I didn't want, but it seems as if there's no way to delete a tag. It's possible I'm missing something, of course. Any help would be greatly appreciated!

WithoutPants:
Resolves #96 

Adds a delete tag button to each tag row (with trash icon).

Changed the `TagQueryBuilder.Destroy` function to also delete references to the tag from the scenes and markers tables, since otherwise you would need to remove all of the tag references manually before deleting the tag. Can't do anything with the `primary_tag_id `on scene markers, since it cannot be unset. Scene markers referencing the tag would need to be modified manually prior to deleting the tag.

### DIFF
--- a/pkg/models/querybuilder_tag.go
+++ b/pkg/models/querybuilder_tag.go
@@ -52,6 +52,19 @@ func (qb *TagQueryBuilder) Update(updatedTag Tag, tx *sqlx.Tx) (*Tag, error) {
 }
 
 func (qb *TagQueryBuilder) Destroy(id string, tx *sqlx.Tx) error {
+	// delete tag from scenes and markers first
+	_, err := tx.Exec("DELETE FROM scenes_tags WHERE tag_id = ?", id)
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Exec("DELETE FROM scene_markers_tags WHERE tag_id = ?", id)
+	if err != nil {
+		return err
+	}
+
+	// does not unset primary_tag_id in scene_markers because it is not nullable
+
 	return executeDeleteQuery("tags", id, tx)
 }
 
--- a/pkg/models/querybuilder_tag.go
+++ b/pkg/models/querybuilder_tag.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"errors"
 	"database/sql"
 	"github.com/jmoiron/sqlx"
 	"github.com/stashapp/stash/pkg/database"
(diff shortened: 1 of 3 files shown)


pkg/models/querybuilder_tag.go
```diff
@@ -52,6 +52,19 @@ func (qb *TagQueryBuilder) Update(updatedTag Tag, tx *sqlx.Tx) (*Tag, error) {
...
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Exec("DELETE FROM scene_markers_tags WHERE tag_id = ?", id)
+	if err != nil {
+		return err
+	}
+
+	// does not unset primary_tag_id in scene_markers because it is not nullable
```
StashAppDev: It's been a while, but I think it's not nullable for a reason... I'm curious if you've tried creating a scene marker with a certain primary tag, and then delete that tag and try to view that scene... does the scene still load?  I get the feeling it might not since the marker list needs a primary tag to display properly.

If it breaks the page we will likely need to do one of two things:

1) Throw an error stating that any scene markers associated with that tag need to be updated
2) Rename the tag to something like "Deleted Tag".  This would act as a placeholder until the user updates any remaining scene markers.  Then they'd be able to delete it.

WithoutPants: To clarify, that comment simply indicates that it won't try to unset it. As a result, if you try to delete a tag where that tag is the primary tag for one or more scene markers, then the entire operation fails with "Error: GraphQL error: FOREIGN KEY constraint failed", and the tag is not deleted. In order for the tag deletion operation to be successful, then they must remove the tag from the applicable markers.

bnkai: Wouldn't it be simpler to just disable the delete button for tags that are primary for a marker?
For example if
select scene_id from scene_markers where primary_tag_id 
is not null or you can use count and >0 also
then set the state of the delete button to disabled
you can even add the count of "primary markers" with maybe a different colour as a tab after the scenes and markers

WithoutPants: From a usability perspective, I think its better and simpler to have the operation fail with a (useful) error message than to have the delete button disabled for an unknown (to the user) reason. The delete operation only fails when its in use as a primary tag, for other uses it works seamlessly.

WithoutPants: I will resolve this by adding a useful error message if there are scene markers with the primary tag set to the tag being removed.

WithoutPants:
Ready for testing.

PR #101, opened 2019-08-15
Show studios with no scenes #69

### LINKED ISSUE #69: Studios only visible after adding atleast 1 scene to it
After adding a new studio it won't be visible under Studios. It's only visible when editing a scene and assigning a studio to it.

After you have assigned the studio to at least 1 scene it does become visible. Is there a filter I am missing?

WithoutPants:
Fixes #69 
Changed join to left join to include studios with no scenes.

### DIFF
--- a/pkg/models/querybuilder_studio.go
+++ b/pkg/models/querybuilder_studio.go
@@ -86,7 +86,7 @@ func (qb *StudioQueryBuilder) Query(findFilter *FindFilterType) ([]*Studio, int)
 	var args []interface{}
 	body := selectDistinctIDs("studios")
 	body += `
-		join scenes on studios.id = scenes.studio_id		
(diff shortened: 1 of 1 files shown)


pkg/models/querybuilder_studio.go
```diff
@@ -86,7 +86,7 @@ func (qb *StudioQueryBuilder) Query(findFilter *FindFilterType) ([]*Studio, int)
 	var args []interface{}
 	body := selectDistinctIDs("studios")
 	body += `
-		join scenes on studios.id = scenes.studio_id		
+		left join scenes on studios.id = scenes.studio_id
```
StashAppDev: My SQL skills are 💩 .  Thanks for this.

Issue #63, opened 2019-06-11
[Request] Remove deleted scenes from database


ghost:
It would be nice to have a feature which removes deleted scenes (and generated content) from the database, when a scene is deleted. I currently have a few dangling scenes which have an entry in stash but can't be played because they are deleted from the filesystem. I tried to look into it myself but I am not familiar with GQL and GO and don't know where to start.

ueaslsef:
Assuming you are using linux:

Find the checksum of the scene in the ui. Copy it.
Then, open up stash's db with `sqlite3`. If not installed, install it.
```bash
sqlite3 ~/.stash/stash-go.sqlite
```
Then 
```sql
DELETE FROM scenes WHERE checksum LIKE '<CHECKSUM>';
```

dotbanana:
I second this. It would be good if the content stashapp shows is a representation of the content of a specific folder. The "Scan"-Script could do an integrity check basically if all files are still present and if not delete it from database.

bnkai:
I'm pretty sure that the clean button in the tasks section is reserved mainly for that  (when it gets implemented though )

If you wan to delete manually from DB , it's mentioned before  deleting a scene you should delete the relevant entries from tables **scene_markers** and **performers_scenes** because of foreign key constraints.
something like this for example  .... (lets say scene id is 50)

```
DELETE FROM performers_scenes WHERE scene_id = 50;
DELETE FROM scene_markers WHERE scene_id = 50;
DELETE FROM scenes WHERE id = 50 ;

```

dotbanana:
I've written a little bashscript to automate this things with your commands. This works great so far but i stumbled across a strange oddity and i don't know if i am doing something horribly wrong or there is some shennanigans going on:  

These commands only work for me up to scene 352. With a higher id the commands fail. I checked with a sqlite-viewer the stash-go.sqlite and really the listing also there goes only to 352. But my recent videofile showing in stash has id 742. Where is the info saved for these files? I looked across the stash-go.sqlite but couldn't find it. But somehow all the scenes above 352 still show up in the stash-UI and on the startpage it says Scenes 742.
In table sqlite_sequence scenes has value 352.  

My bash-script:  

`#!/bin/bash
# Script needs to be called with number of scene you want to remove
pathtofile=$( sqlite3 ~/.stash/stash-go.sqlite "SELECT path FROM scenes WHERE id = $1;" )
[...]

PR #102, opened 2019-08-15
Add Delete Scene button

### LINKED ISSUE #76: [Request] Delete-Buttons in Overview and Detail-Page
This is kind of connected to: https://github.com/stashapp/stash/issues/63  
  
It would be really handy in my opinion to have little buttons by the side of the videos in the overview and especially in the detail-page where you watch a video so that you can delete a certain scene directly from stashapp-database _and_ also from filesystem.
Maybe with an option in the settings-menu if data should be deleted in filesystem or not.  

With implementing this one could manage the collection through stashapp and without using a cron to search for deleted scenes and then delete them one by one from database manually.   
  
I use stashapp remotely on a machine which collects scenes for me automatically and i use it to check the downloaded content out and for this/my usecase also delete-buttons in the scene-overview would be nice and way quicker with sorting the collection.

WithoutPants:
Resolves #76, #8 (with the #100 PR) and should help with #63 

Adds a delete button to the edit scene panel. Clicking this button shows a dialog to confirm the delete operation, while also giving the option to delete the file from the filesystem (defaulting to false), and to delete generated files (defaulting to true).

When deleting the files from the filesystem, if any of the file deletions fail, then an error is logged, but does not prevent the operation from completing.


echo6ix:
On the Windows build:

Sometimes deleting a scene that had just been viewed deletes everything but the video file. This also happens when trying to manually delete a video from the file system that was just viewed in Stash, with the error, "File in use. This action cannot be completed because the file is open in stash.exe. Close the file and try again"

Leopere:
I am still receiving these conflicts.

## Conflicting files
pkg/api/resolver_mutation_scene.go
pkg/models/generated_exec.go
pkg/models/generated_models.go

WithoutPants:
The delete function now closes running streams/encodes before attempting to delete the file. This should resolve the issue identified by @echo6ix 

Leaving this in WIP until #146 is merged, as I will clean up some repeated code before it will be ready to merge.

bnkai:
I'll test after that then

WithoutPants:
Unified the common code between this PR and #146. As a result, the delete scene and clean functionality may need a quick verification test. Both appear to be testing ok on my end.

bnkai:
Everything checks out ok for me too.

Leopere:
Verified by @WithoutPants and @bnkai merging

PR #103, opened 2019-08-15
Set title, details and date from scene file metadata

WithoutPants:
During scan, sets the title of a scene from the title field in the files metadata, or defaults to the file's base name (instead of being an empty string currently). Sets the date to the creation date in the metadata. Sets the Details field from the Comment field in the file's metadata.

Within Windows, I was also able to set the rating and tags fields in an mp4 file, but wasn't able to extract this information out again using ffprobe.

### DIFF
--- a/pkg/ffmpeg/ffprobe.go
+++ b/pkg/ffmpeg/ffprobe.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


egon6969:
Hello,

i hope this is the right place for my request/comment.
It appears, that you use the "creation_time" tag from ffprobe.
This tag is not present in most of my files and therefore the function to automatically set the date is not working.
Would it be possible to use the file creation date (by using https://github.com/djherbis/times for example) as a backup, if ffprobe returns nothing?

Leopere:
@egon6969, unfortunately, this is a closed and merged Pull Request so it's not the correct place to request a feature.  If you wish to complain about this one feel free to open a new issue. https://github.com/stashapp/stash/issues/new

Issue #58, opened 2019-05-28
[Request] Add and display studio without image


echo6ix:
Please consider adding the following:

1. Ability to create new studio without an image. Studios that do not have an image should just appear as a text link.
2. A toggle override in the settings to display the text studio name everywhere instead of the studio image

What's the use-case for this other than preferring text over a logo? A lot of studios don't have a logo or image representation, such as very small clips 4 sale "studios".

StashAppDev:
Images are required due to how I handle the JSON export.  I needed some type of unique identifier for the studio json filenames which could be mapped back to the studio name string.  The identifier I use is the MD5 hash of the image.

For example:

* Studio A has an image which hashes to 30bcf62db486d0248f5566833074fed3.
* The metadata exports to studios/30bcf62db486d0248f5566833074fed3.json
* in `mappings.json` under the `studios` map I map "Studio A" to `30bcf62db486d0248f5566833074fed3`.  Allowing me to map the correct studio file to the given studio name

This probably isn't ideal, but I haven't had time to circle back and rethink the JSON schema and how this might work.

echo6ix:
Thanks for taking your time to explain how it works.

I don't understand why the identifier cannot be from the MD5 hash of a string value, such as the Studio Name, but I'm sure there's a good reason.

Leopere:
Closed by #109

PR #109, opened 2019-08-21
Generate performer checksum from name instead of image. Use default performer image

### LINKED ISSUE #67: Saving performers with the same image
When saving performers with the same image you get a unique constraint error. I presume because you build a checksum using only the image.

Seeing as I do not have an image for every performer I'm adding, I reuse a generic "female shape"-image for ~95% of the performers. At least until I care to spend time on actually adding an image for them.

To work around this I removed the unique key from the database and now it works fine. Though having to add an image to even be able to save it is kind of a hassle. Maybe use just the performer's name for the checksum? It's something that should be unique anyway.

WithoutPants:
Resolves #67 

Changes the checksum generation for performers to be based on name instead of image. Name should be unique for performers anyway, so it makes sense to generate the checksum on the performer name. This also means an image can be reused (for generic images).

This change also adds a default generic performer image, for when no image is provided. The current image is really just a placeholder, and I'd like a better one. Ideally, the replacement would be similarly compact. 

icons8-female-user-100


WithoutPants:
Added the same change for studios (I was going to make a separate PR, but it required the same utils method). 

Added a placeholder studio image as well.

Resolves #58 

Video-03

echo6ix:
> Added the same change for studios (I was going to make a separate PR, but it required the same utils method).
> 
> Added a placeholder studio image as well.
> 
> Resolves #58
> 
> Video-03

As discussed on Discord, it might be a good idea to create an option in `Settings > Interface Configuration > Scenes` to toggle on/off the display of the studio as a text (use the provided studio name) or as an image.

I'm not sure if providing a Studio Name is mandatory when creating a studio. If not it might be a good idea to make it mandatory to avoid outputing a null string if they turn off studio images on the scene pages.

bnkai:
@WithoutPants 
Afaik the checksum from the image was also used for the import function (for both performer/studio)
Have a look at the task_import.go file (use of ProcessBase64Image function from utils/image.go ) So there may be side effects to the point of messing up the import.
I'm not sure if it matters for the export.
Better wait for Stash to tell you in detail though

Leopere:
Is this PR a WIP?   If it is please add WIP into the PR name so its not possible to merge until you remove it.

WithoutPants:
> Is this PR a WIP? If it is please add WIP into the PR name so its not possible to merge until you remove it.

It wasn't until @bnkai 's find. Fixed up the issues found.

WithoutPants:
> Another matter is also the recalculation of the checksum for performers / studios already in the DB so that the calculation method is consistant for all. An export / import sequence probably is enough , provided that import,export is patched as mentioned above.

It shouldn't really matter if it is inconsistent for existing data. You could indeed correct it with an export and re-import, but otherwise the app is going to function the same regardless, and the checksum is only exposed to the user during the export/import anyway.

iKeats:
Yo folks. I started making a NoID performer image for the default .... I then liked the idea and made 40 of them ahah. Attached a zip with all of them. My idea, and it would be awesome if doable, is to have the system pick one at random for each new entry. This would alleviate the repetitiveness that a single picture would create. Let me know :)

Preview: 
NoName04


[NoName.zip](https://github.com/stashapp/stash/files/3535319/NoName.zip)

Leopere:
Nice work!

WithoutPants:
Added a packr box with the images @iKeats kindly provided. Changed so that a random image is selected from the images if none is provided.

WithoutPants:
Merged from master. Ready for retesting.

bnkai:
Given the lack of feedback i tested this out again. It looks ok

Leopere:
@iKeats you should make these silhouette designs for all of the tags good grief they look fantastic.

PR #110, opened 2019-08-21
Display hash and build time at startup and in about page

### LINKED ISSUE #107: Display version/build number in console and UI
It might be a good idea to show the Stash version/build in the following places:

1. Upon loading Stash the console can spit out the version/build number, eg `build="xxxxxxxxxx" time="2019-08-20T15:03:16-04:00" level=info msg="stash is running on HTTP at http://0.0.0.0:9999/"`
2. In the UI at _http://localhost:9999/settings?tab=about_

WithoutPants:
Resolves #107 

Generates version info based on the git hash and build time. This information is printed on startup and is also shown in the About page.

I also moved the packr2 call to the ui target since it needs to be rebuilt when the UI changes anyway. Running `packr2 build` shows a bunch of deprecation warnings, so I changed it to just `packr2`.


WithoutPants:
Updated to show short git hash (to be consistent with github's hashes), and merged from master. Ready for retesting.

bnkai:
@WithoutPants  while i can build it manually and it works ok , i'm not sure that the travis auto builds 
will actualy have the changes.Can you have a look at .travis.yml just to be safe?

WithoutPants:
The travis builds run `cross_compile.sh` which has now been updated and verified. However, I suspect that `.goreleaser.yml` also needs to be updated. I'll need to look into what needs to be done there. It will be difficult to verify without pushing the branch though.

bnkai:
@WithoutPants is the `.goreleaser.yml` used anywhere now? The `cross_compile.sh script` , that travis also uses, gets the goreleaser from the custom docker image Stash built ... or so i think at least.
I would say push the PR and see what happens , it's not that easy to check the travis build.

bnkai:
Update :
I built manually using the makefile and also i imitated as close as possible the travis.yaml  sequence with the cross compile script and both seem to work fine for me. So it's ok from my testing also.

WithoutPants:
> @WithoutPants is the `.goreleaser.yml` used anywhere now? The `cross_compile.sh` script , that travis also uses, gets the goreleaser from the custom docker image Stash built ... or so i think at least.
> I would say push the PR and see what happens , it's not that easy to check the travis build.

It looks like the travis.yml handles pushing the release, so `.goreleaser.yml` looks superfluous. I think this is good to go.

Leopere:
Merged under instruction from @WithoutPants to do so.

WithoutPants:
I can verify that the Windows release executable includes the build hash information.

Issue #39, opened 2019-04-02
Related / Similar scenes based on scene title, tags, studio, performers


StashAppDev:
This could show up as a tab on the scene pages or as a module for the dashboard

ueaslsef:
Doing some stuff with the db, here are some queries I am trying:

### 10 related scenes based upon tags, sorted after count of matching tags:
```sql
SELECT Scenes.id, Scenes.title, matching_tags
FROM (
  SELECT scene_id, tag_id, count(scene_id) as matching_tags
  FROM scenes_tags
  WHERE tag_id IN (SELECT tag_id FROM scenes_tags WHERE scene_id=375)
  GROUP BY scene_id
  ORDER BY matching_tags DESC
  LIMIT 10
) Tags
LEFT JOIN Scenes ON Scenes.id=Tags.scene_id;
```

Will experiment further.

compound-dumbo:
For a more sophisticated solution one could compute a similarity matrix based on tags and performers.
The main advantage of this that I see is that you're not limited to suggestions of the same performers or tags, but can take similarities between tags into account.
But I don't know if that kind of computational complexity is warranted

PR #116, opened 2019-08-27
[WIP] Adds SQL statements to query similar scenes

ueaslsef:
Adds sql statements proposed in #39.

Using them, you can query for related scenes based upon common tags or similarity of actors in this scene.

I'm not familiar with the data flow in this application, but maybe @WithoutPants is able to do his magic and implement this into a scene tab.


Leopere:
@ueaslsef where are we at with this?  It's been open for too long if this isn't dead or can be updated for the latest schema, please re-open it with the WIP tag removed when it's ready.

Issue #61, opened 2019-06-01
freeones scrape error docker


egon94:
Hello,

I am not able to add new performers by scraping because I get the following errors:

`time="2019-06-01T22:36:13+02:00" level=fatal msg="Get https://www.freeones.com/suggestions.php?q=&t=1: x509: certificate signed by unknown authority"`

in docker log and

`Error: GraphQL error: Get https://www.freeones.com/search/?t=1&q=test&view=thumbs: x509: certificate signed by unknown authority`

in the webinterface.

I deployed the container with the provided compose file.
Please inform me if you need additional information.

pmisterpp:
Docker image is missing certificate authority certificates.
I have extended the stashapp/stash docker image and installed ca-certificate package to make it work.
Here's an example of my Dockerfile:

```
FROM stashapp/stash
RUN apt update && apt install ca-certificates -y && update-ca-certificates
```

egon94:
Yeah that works. Thanks.
You should probably fix it and submit a pull-request.

ExceptionalError:
Hi, I just pulled the docker and I am still seeing this error:
````
Error: GraphQL error: Get https://www.freeones.com/search/?t=1&q=Prename%20Surname&view=thumbs: x509: certificate signed by unknown authority
````
Any ideas what went wrong?

Leopere:
Looks like we need to update the container to have The TLS certificate
store. I didn’t include it in my original builds because Stash was simpler
back then. If someone has time to look into including this before I do it
looks like the issue.

PR #129, opened 2019-10-07
Install ca-certificates in Docker container

joyov:
I believe this was accidentally added to the wrong build stage in https://github.com/stashapp/stash/commit/30a2a2cf0035ac2113afbbe69db8bfc81abee069. The package is not needed during the build process, but for the final container. Without the `ca-certificates` package, Stash will crash and return the following error message when the scrape feature is used:

```
FATA[0024] Get https://www.freeones.com/suggestions.php?q=&t=1: x509: certificate signed by unknown authority
```

This fixes https://github.com/stashapp/stash/issues/61 (this time for real).


Leopere:
This is a good addition.

PR #132, opened 2019-10-09
Remove unused and generated code

friendlycrab:
These are some cleanups intended to make collaboration easier. Specifically, by removing the vendor directory and generated files from the git repository and instead generating them when needed, merge conflicts are reduced. I also removed UI V1 because it appeared unused and at least for me it was confusing to still find it in the repository. If this PR is accepted, the old UI code will be preserved through git's history. I changed the command `make gqlgen` to `make generate` to better reflect that it includes both gqlgen and packr.


Leopere:
This no longer builds after this PR

friendlycrab:
@Leopere Fixed Travis build via 7cf41b5a44b9352436c5ca477eaba731f48f3856.

Leopere:
Thanks for fixing that @friendlycrab I'll label it hacktoberfest for you and whenever @StashAppDev gets around to reviewing/merging.  I don't really want to merge Go Code I do the docker stuff and minor merges.

WithoutPants:
I'd prefer not to remove the vendor directory. It allows us to make local changes to the vendor libraries in the event that there is an unpatched bug that isn't fixed in one of the libraries (see #38 and #83). Conflicts typically don't arise in the vendor code.

If we're changing the way to generate files, then the README.md should be updated with the new method as well.

Leopere:
Might not be worth doing this this early.

WithoutPants:
I think the rest of the PR is worth doing. V1 UI isn't useful to continue to have in the source tree, and removing the generated files would help avoid a whole heap of merge conflicts. I think this should be reopened, but separate out the vendor stuff.

Leopere:
If @friendlycrab updates with what @WithoutPants mentioned I can merge this.

friendlycrab:
Understood. I dropped f4ab98a37ed78aa92147dce5ff5827a25959fd9b and force-pushed to revert the `vendor` changes.

> the README.md should be updated with the new method as well

I already updated `README.md` to reflect the change from `make gqlgen` to `make generate`. Added the `make generate` step to the building instructions via aa19b87dcd5473f0f5e212f7248d98dcd926903e. Anything else I missed?

Slightly off-topic, but would it make sense to add `cd ui/v2 && yarn build` to `Makefile`'s `build` rule? This would be consistent with `make generate` which also includes UI's generation and it would keep building the software to just two commands, `make generate` and `make build`.

Leopere:
https://github.com/stashapp/stash/pull/132#issuecomment-541599456 seems like a good Issue topic.  Please definitely make an issue from this if there isn't one already.  

@friendlycrab could you resolve the Conflicting Files in this PR? Once this is finished I can merge as soon as we validate that @bnkai and @WithoutPants have not seen any regressions from this change.

bnkai:
I haven't checked the code for the grahpql or packr2 generation so i think @WithoutPants is better qualified to deal with this.

WithoutPants:
> I already updated `README.md` to reflect the change from `make gqlgen` to `make generate`. Added the `make generate` step to the building instructions via [aa19b87](https://github.com/stashapp/stash/commit/aa19b87dcd5473f0f5e212f7248d98dcd926903e). Anything else I missed?

There is a `ui` make target that can be used instead of `cd ui/v2`, `yarn build`. Can you add this to the Commands section and replace step 2 in Making a Release?

> Slightly off-topic, but would it make sense to add `cd ui/v2 && yarn build` to `Makefile`'s `build` rule? This would be consistent with `make generate` which also includes UI's generation and it would keep building the software to just two commands, `make generate` and `make build`.

During development, rebuilds typically fall into one of three scenarios:
- updated GQL files (and possibly server code/UI code) -> `make generate ui build`
- updated UI (and possibly server code) -> `make ui build`
- updated server code -> `make build`

You don't want to have to rebuild the UI if you changed only the server code. Adding a default target to build everything (`generate ui build`) isn't a terrible idea for a freshly downloaded working tree. Though I think you still need to run `yarn install` even before you can run `yarn build` so it may not necessarily be that much more efficient.

> @friendlycrab could you resolve the Conflicting Files in this PR? Once this is finished I can merge as soon as we validate that @bnkai and @WithoutPants have not seen any regressions from this change.

I'll feedback when I've tested this.

WithoutPants:
This looks good to me.

friendlycrab:
Rebased on top of latest `master` to resolve merge-conflicts and added `make ui` instructions via 080d80117b2958025c63945378f447d10997fae1.

WithoutPants:
Looks good.

WithoutPants:
@friendlycrab please remove the WIP tag if it is good to go.

Leopere:
I was the one I believe tho added the WIP tag.  It should be ready I imagine.  Have we tested the build or is it necessary?

WithoutPants:
> I was the one I believe tho added the WIP tag. It should be ready I imagine. Have we tested the build or is it necessary?

Yep. That's what I meant by "Looks good", but I'll be more explicit next time :)

Leopere:
Heh sorry just don't want to have to untangle this if I merge and later we regret it.

PR #136, opened 2019-10-12
Make title from file metadata optional

WithoutPants:
Amendment to #103 

The issue with the original PR is that most file releases that have the Title metadata set are garbage. As a result, I have a bunch of files with garbage titles, where the filename would have been a far more useful title. 

This change adds an option checkbox to the scan button to optionally ignore the metadata in the file and default the title to the base filename.

### DIFF
--- a/graphql/documents/queries/settings/metadata.graphql
+++ b/graphql/documents/queries/settings/metadata.graphql
@@ -6,8 +6,8 @@ query MetadataExport {
   metadataExport
 }
 
-query MetadataScan {
-  metadataScan
+query MetadataScan($input: ScanMetadataInput!) {
+  metadataScan(input: $input)
 }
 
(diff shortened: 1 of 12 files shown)


bnkai:
Just tested and looks good.

Leopere:
Thanks for the PR and the review @WithoutPants & @bnkai I'll assume that since @WithoutPants recently volunteered in Discord to handle the windows binary review and @bnkai certified functionality of the Linux build we were okay to merge.

echo6ix:
Re: http://localhost:9999/settings?tab=tasks

- Consider rephrasing the string `Set name from metadata (if present)` to `Import metadata from files if present`. I think the purpose of the function is a bit clearer like that.
- Have it unchecked by default since it seems it's likely just going to contain torrent scene info?

PR #137, opened 2019-10-12
Allow unsetting of rating, studio, gallery

### LINKED ISSUE #127: Unable to remove studio from scene
I accidentally associatated a studio with a particular studio. It is not possible to select None as an option for studio if you have already selected a studio. Need to add a None option.

WithoutPants:
Fixes #127 

Allows unsetting of rating, studio and gallery in a scene.

Also fixes unsetting the sceneID value on the scene's old gallery when the scene's gallery is changed.

### DIFF
--- a/pkg/api/resolver_mutation_scene.go
+++ b/pkg/api/resolver_mutation_scene.go
@@ -14,36 +14,36 @@ func (r *mutationResolver) SceneUpdate(ctx context.Context, input models.SceneUp
 	// Populate scene from the input
 	sceneID, _ := strconv.Atoi(input.ID)
 	updatedTime := time.Now()
-	updatedScene := models.Scene{
+	updatedScene := models.ScenePartial{
 		ID:        sceneID,
-		UpdatedAt: models.SQLiteTimestamp{Timestamp: updatedTime},
+		UpdatedAt: &models.SQLiteTimestamp{Timestamp: updatedTime},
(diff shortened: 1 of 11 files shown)


bnkai:
@WithoutPants even setting the rating with this build messes the scene's details
Everything in the file details tab turns to null File Size Duration Dimensions ....

bnkai:
update : even unsetting a gallery has the same issues

Leopere:
Can we get an update on this PR it would be awesome to have this feature fixed in the main build.

bnkai:
Patience @Leopere  i had a quick look but i can't figure out where it goes wrong.
Hopefully Withoutpants will figure it out.

Leopere:
If you can't tell I work from home.  When I get a second I check my emails. I didn't realize the timestamp when I'd posted so rather than deleting it I just let it fly.

WithoutPants:
I had time to develop the original change, but little time to investigate fixing it. Anyway, the issue @bnkai raised should now be fixed. However, this is still WIP since rating keeps resetting to the old value in the UI when the scene is saved, so I still need to fix that.

WithoutPants:
Ready for testing/review.

bnkai:
All tested ok and working

Leopere:
Thanks for everyone's hard work on this.

Issue #122, opened 2019-09-12
[Bug?] Changing filter on performers page doesn't reset page parameter


echo6ix:
Suppose you're browsing the list of performers, and you're on the 3rd page, `p=3`, using the filter `scenes_count`: http://localhost:9999/performers?disp=0&p=3&q=&sortby=scenes_count&sortdir=asc

Now suppose you change the filter to `updated_at`, http://localhost:9999/performers?disp=0&p=3&q=&sortby=updated_at&sortdir=asc

The page parameter, `p`, doesn't reset to `1`, it remains `3`. I'm not sure if this is intended behavior but it's confusing.

I had always assumed changing the filter was akin to starting a new query and only noticed it when I changed the filter to scene_count and saw the list wasn't starting with performers who had 0 scenes.

WithoutPants:
This can now be closed.

PR #140, opened 2019-10-15
Reset to first page when changing sort by

WithoutPants:
Addresses #122 

Very simple fix to reset the query to the first page when changing the sort by field. It does seem not intuitive to stay on the same page number when changing the sort by field. I wasn't sure about resetting the page when changing the sort direction (it currently does not), so I left it as is.


bnkai:
Works ok. Nothing more to say :-)

PR #141, opened 2019-10-15
Add scene list view

WithoutPants:
Adds a table list view for scenes instead of the current TODO.

Currently hard-coded to show the following columns:
- title
- rating
- duration
- tags (with links to scene query view)
- performers (with links)
- studio (with link)

Possible improvements:
- make displayed columns customisable
- sort by when clicking header column

Feedback welcomed.

### DIFF
--- a/ui/v2/src/components/scenes/SceneList.tsx
+++ b/ui/v2/src/components/scenes/SceneList.tsx
@@ -8,6 +8,7 @@ import { ListFilterModel } from "../../models/list-filter/filter";
 import { DisplayMode, FilterMode } from "../../models/list-filter/types";
 import { WallPanel } from "../Wall/WallPanel";
 import { SceneCard } from "./SceneCard";
+import { SceneListTable } from "./SceneListTable";
 
 interface ISceneListProps extends IBaseProps {}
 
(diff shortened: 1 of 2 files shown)


bnkai:
Plain and working ok.
I thinks it's ok as a start point to scratch that of the landing page.
Wouldn't mind a thumbnail column but that's just a personal preference.
Can you perhaps do it for the performers also?

Leopere:
I feel like anything is better than the TODO page for now, I don't even use the app this much but the fact that its a big gaping todo list item in the middle of the UI bothers me immensely. Since it's reviewed as functional on windows and linux thanks to @WithoutPants and @bnkai I'll merge this and someone can redo the thing later.

echo6ix:
Your list view could also be perfect for quick metadata editing of scenes. I'm taking inspiration from something like [MP3tag](https://www.mp3tag.de/images/sht_main.png), which could look something like this on Stash.

list view

This was a quick 5-minute concept thrown together in PS so I realize the UI is meh looking, but you get the [idea](https://www.mp3tag.de/images/sht_main.png).

Probably a good idea to display _a lot_ more than than 40 scenes by default on a list view though.

Leopere:
I really wish that we could do something here like Music Brainz Picard with perceptual fingerprinting analysis and a big central API / DB.

Leopere:
Looks like theres a lot to learn from how the [Picard](https://picard.musicbrainz.org/) tagger works.  https://picard.musicbrainz.org/plugins/

PR #142, opened 2019-10-15
Add performer list view.

WithoutPants:
Adds a performer list view similar to #141 

Currently uses the following hard-coded list of columns:
- name
- aliases
- favourite
- scene count
- birthdate
- height

Added thumbnails to both performer list and scene list.
Removed the list view from the TODO list on the front page.


bnkai:
Works fine. Thanks fo the thumbs

PR #143, opened 2019-10-15
Freeones fixes

bnkai:
Fix a couple of bugs with the scraper

- when there was a date of birth missing or some other random stat missing from the bio page it freaked out
example: search for `Nathaly Heaven` 
the freeones bio is missing the birth date

- when there were more than one aliases for a performer and you were searching with it
example: `Allexis Blow`

### DIFF
--- a/pkg/ffmpeg/encoder_screenshot.go
+++ b/pkg/ffmpeg/encoder_screenshot.go
@@ -21,7 +21,7 @@ func (e *Encoder) Screenshot(probeResult VideoFile, options ScreenshotOptions) {
 		"-v", options.Verbosity,
 		"-ss", fmt.Sprintf("%v", options.Time),
 		"-y",
-		"-i", `"` + probeResult.Path + `"`,
+		"-i", probeResult.Path, // TODO: Wrap in quotes?
 		"-vframes", "1",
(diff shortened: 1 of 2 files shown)


Leopere:
As this is the obverse to the latest PR's I'll need @WithoutPants to confirm functionality of the Windows builds here and then I can merge this if nobody else has any objections.

Leopere:
I still have no idea how the scraper works because I have yet to see something tag itself yet.

bnkai:
@Leopere the freeones scraper is used only when creating new performers to fetch the profile details. As for auto tagging i think it will take some more time...

Leopere:
There really doesn't seem like a lot of information gathered there, I've noticed that it doesn't scrape it's own imagery or is this because the preview(production) docker container includes a broken scraper.

WithoutPants:
Just verifying the original bug.

Scraping "Nathaly Heaven" already seems to work ok.

Scraping "Allexis Blow" results in the following error in the pre-fixed code:
```
Error: GraphQL error: Internal system error. Error <runtime error: index out of range>
```

WithoutPants:
Tests ok.

Issue #144, opened 2019-10-16
Scanning should regenerate screenshots if missing


WithoutPants:
If a scene's screenshot wasn't generated or was subsequently removed, then it is not possible to regenerate the screenshot without removing the scene and rescanning, or renaming the file and rescanning. The scan function should check for the presence of the screenshot, even if the scene is already found.

ExceptionalError:
Shouldn't the screenshot generation only be done when generating and not while scanning?

bnkai:
It is part of the scaning process anyway because we need these images anyway.
The generate part has parts that some people may not want ,need or whatever thats why they can be opted out.

Leopere:
Closed by #145

PR #145, opened 2019-10-16
Scan screens

bnkai:
addres issue #144

### DIFF
--- a/pkg/manager/task_scan.go
+++ b/pkg/manager/task_scan.go
@@ -163,31 +163,25 @@ func (t *ScanTask) makeScreenshots(probeResult *ffmpeg.VideoFile, checksum strin
 		return
 	}
 
-	if probeResult != nil {
-		// makescreenshots is called for a new file
-		t.makeScreenshot(*probeResult, thumbPath, 5, 320)
-		t.makeScreenshot(*probeResult, normalPath, 2, probeResult.Width)
-	} else {
-		// makescreenshots is called for an already existing file
-		if !thumbExists || !normalExists {
-			checkvideoFile, err := ffmpeg.NewVideoFile(instance.FFProbePath, t.FilePath)
+	if probeResult == nil {
+		var err  error 
+		probeResult, err = ffmpeg.NewVideoFile(instance.FFProbePath, t.FilePath)
 
-			if err != nil {
-				logger.Error(err.Error())
-				return
-			}
-			if !thumbExists {
-				logger.Infof("Recreating thumbnail for %s", t.FilePath)
-				t.makeScreenshot(*checkvideoFile, thumbPath, 5, 320)
-			}
-
-			if !normalExists {
-				logger.Infof("Recreating screenshot for %s", t.FilePath)
-				t.makeScreenshot(*checkvideoFile, normalPath, 2, checkvideoFile.Width)
-			}
+		if err != nil {
+			logger.Error(err.Error())
+			return
 		}
 	}
 
+	if !thumbExists {
+		logger.Debugf("Creating thumbnail for %s", t.FilePath)
+		t.makeScreenshot(*probeResult, thumbPath, 5, 320)
+	}
+
+	if !normalExists {
+		logger.Debugf("Creating screenshot for %s", t.FilePath)
+		t.makeScreenshot(*probeResult, normalPath, 2, probeResult.Width)
+	}
 }
 
 func (t *ScanTask) makeScreenshot(probeResult ffmpeg.VideoFile, outputPath string, quality int, width int) {
--- a/pkg/manager/task_scan.go
+++ b/pkg/manager/task_scan.go
@@ -81,7 +81,31 @@ func (t *ScanTask) scanScene() {
 	qb := models.NewSceneQueryBuilder()
 	scene, _ := qb.FindByPath(t.FilePath)
 	if scene != nil {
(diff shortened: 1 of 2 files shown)


pkg/manager/task_scan.go
```diff
@@ -81,7 +81,31 @@ func (t *ScanTask) scanScene() {
...
+			if !thumbExists {
+				logger.Infof("Recreating thumbnail for %s", t.FilePath)
+				t.makeScreenshot(*checkvideoFile, thumbPath, 5, 320)
+			}
+
+			if !normalExists {
+				logger.Infof("Recreating screeenshot for %s", t.FilePath)
+				t.makeScreenshot(*checkvideoFile, normalPath, 2, checkvideoFile.Width)
+			}
+		}
```
WithoutPants: Can we move this into a separate function? `scanScene `is already largish.

pkg/manager/task_scan.go
```diff
@@ -185,3 +186,33 @@ func (t *ScanTask) calculateChecksum() (string, error) {
...
+			logger.Infof("Recreating thumbnail for %s", t.FilePath)
+			t.makeScreenshot(*checkvideoFile, thumbPath, 5, 320)
+		}
+
+		if !normalExists {
+			logger.Infof("Recreating screenshot for %s", t.FilePath)
+			t.makeScreenshot(*checkvideoFile, normalPath, 2, checkvideoFile.Width)
+		}
+	}
+	return
```
WithoutPants: This is now superfluous.

pkg/manager/task_scan.go
```diff
@@ -81,7 +81,8 @@ func (t *ScanTask) scanScene() {
 	qb := models.NewSceneQueryBuilder()
 	scene, _ := qb.FindByPath(t.FilePath)
 	if scene != nil {
-		// We already have this item in the database, keep going
+		// We already have this item in the database, check for thumbnails,screenshots
+		t.checkScreenshots(scene.Checksum)
```
WithoutPants: There is already a `makeScreenshots` function that can be reused here. It checks if the screenshot files exist and only rebuild if they are not present.

I would do the following:

```
qb := models.NewSceneQueryBuilder()
scene, _ := qb.FindByPath(t.FilePath)

videoFile, err := ffmpeg.NewVideoFile(instance.FFProbePath, t.FilePath)
if err != nil {
	logger.Error(err.Error())
	return
}

t.makeScreenshots(*videoFile, scene.Checksum)

if scene != nil {
    // We already have this item in the database, keep going
    return
}
```

Then the new function isn't needed at all. Sorry for not picking this up earlier.

bnkai: https://github.com/stashapp/stash/commit/a67a45ce910f9dfd475e77ab58eb8fe3ac109ae6

it used to be like that but invoking ffprobe for every single file slowed down the scan process too much.

WithoutPants: In that case, I would change `makeScreenshots` to accept a pointer to a `VideoFile`. If this is nil, then create the `VideoFile`. Then line 85 will be `t.makeScreenshots(nil, checksum)`, and the existing call to `makeScreenshots` will be changed to pass the pointer. Basically want to avoid repeated code if possible.

bnkai: i wanted to avoid editing the function but i guess i'll just do that ... sounds good

bnkai: Should be ok now 

@WithoutPants have a look

pkg/manager/task_scan.go
```diff
@@ -150,19 +151,43 @@ func (t *ScanTask) scanScene() {
...
 
-	t.makeScreenshot(probeResult, thumbPath, 5, 320)
-	t.makeScreenshot(probeResult, normalPath, 2, probeResult.Width)
+	if probeResult != nil {
+		// makescreenshots is called for a new file
+		t.makeScreenshot(*probeResult, thumbPath, 5, 320)
+		t.makeScreenshot(*probeResult, normalPath, 2, probeResult.Width)
+	} else {
+		// makescreenshots is called for an already existing file
+		if !thumbExists || !normalExists {
```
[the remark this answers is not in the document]
bnkai: Regarding probeResult if it's nil i don't necessary need to create it
The whole point in the ifs is to avoid as much as possible creating it even if  it means a little more code , because to create  it means running ffprobe for the media file 
When the file is new we can't avoid it because we need the probe anyway but when it's in the database me must avoid that as much as possible
Also any call to makeScreenshot must be spared because it calls ffmpeg to create an image from the media file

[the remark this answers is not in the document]
WithoutPants: I agree. That's why the `probeResult` creation would be after the check for `thumbExists` and `normalExists`. It would only be created if `probeResult` is nil and a screenshot needs to be created.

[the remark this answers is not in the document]
bnkai: lol now fixing 
realized what you meant

bnkai:
it should be ok now ......
i changed the prints to Debug so  i can change them back when we have a proper log file

WithoutPants:
Code looks good.

Tested by deleting some screenshots and rescanning. There's no logging for when recreating the screenshots, but not a big deal. Also tested for regressions when scanning new files.

bnkai:
I added the printing only when we are REgenerating the screenshot,thumbnail.

PR #146, opened 2019-10-16
Add functionality to the Clean Button

endqwerty:
Add the ability to clean up db entries and generated content for files that are no longer found in the filesystem

### DIFF
--- a/pkg/manager/task_clean.go
+++ b/pkg/manager/task_clean.go
@@ -5,7 +5,9 @@ import (
 	"github.com/stashapp/stash/pkg/database"
 	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/models"
+	"github.com/stashapp/stash/pkg/utils"
 	"os"
+	"path/filepath"
 	"strconv"
 	"sync"
 )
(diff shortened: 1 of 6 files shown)


endqwerty:
use from #102 
`SceneDestroy90` with a `input.DeleteFile = false`
`deleteGeneratedSceneFiles()`

pkg/manager/task_clean.go
```diff
@@ -0,0 +1,42 @@
...
+		t.deleteScene(strconv.Itoa(t.Scene.ID))
+	}
+}
+
+func (t *CleanTask) deleteScene(id string) {
+	ctx := context.TODO()
+	qb := models.NewSceneQueryBuilder()
+	tx := database.DB.MustBeginTx(ctx, nil)
+	defer tx.Commit()
+	qb.Destroy(strconv.Itoa(t.Scene.ID), tx)
```
endqwerty: use from #102
SceneDestroy() with a input.DeleteFile = false
deleteGeneratedSceneFiles()

WithoutPants:
There are a number of unrelated files that have been updated and added in this branch: `go.mod`, `go.sum`, `metadata.graphql`, and a vscode settings file. The merge set should be a minimum set of files to perform your change, unless there is good reason to change the other files. Similarly, there are a number of changes to the style of the existing files. Preference would be to maintain existing style - again, unless there's a reason to change.

Please use #102 (`resolver_mutation_scene.go` and `querybuilder_joins.go`) to perform the scene destruction, as that code removes everything that needs to be removed. This function should also remove the generated files, in my opinion.

I would also add a confirmation dialog for the clean button because it is a destructive action.

endqwerty:
I'll clean it up. I added in the vscode stuff since the intellij stuff was already there. Figured i'd match it, I can just local ignore it though.

As for this PR needed fixes. Its going to depend on #102 getting fixed. I'll rebase onto #102, but until #102 is merged, this won't be mergeable

WithoutPants:
> As for this PR needed fixes. Its going to depend on #102 getting fixed. I'll rebase onto #102, but until #102 is merged, this won't be mergeable

I meant that you should just lift the code out of #102. I should have been more explicit. As it is, we'll revisit once #102 is merged.

endqwerty:
Ohhhh. Yeah, I can do that. I can implement just the gql functions, I don't really want to mess with the UI for scene deletion itself yet.

WithoutPants:
To be clear, I mean copying `SceneDestroy` and `deleteGeneratedSceneFiles` from `resolver_mutation_scene.go` in #102, and the required functions from `querybuilder_joins.go`. Rebasing is generally a bad idea in pull requests as it makes it harder to follow to history.

endqwerty:
I'm copying them. I get it. As for rebasing, does it matter? its my branch. I'm not rebasing your master. The point is to remove the history as its messy and useless at this point.

WithoutPants:
It rewrites the history on your (public) branch, making all the discussion and reviews hard to follow for anyone not already involved. It's important to preserve that info and shows the evolution of the branch/feature. I'm not involved in a heap of github projects, so I might not be the best authority. Maybe @Leopere can weigh in?

endqwerty:
I'm not familiar enough with gql mutations to know if the server itself should be accessing them. Quick online check implied that its was primarily for frontend implementation.

I implemented the equivalent for cleaning up associated data/files in the database if the scene is not found to exist in the filesystem. Let me know if I should be using the mutations instead.

Future work (not to be done here unless you really want it):

* clean up files that are found in stash cache folders, but have no related database entry
* clean up database entries but leave files alone for if a library is removed (should just be a quick check to see if the path for the library is in the path for the scene and if not, remove the db/generated files regardless)

WithoutPants:
Code looks good. Only remaining item I can see is the confirmation dialog to confirm the operation.

endqwerty:
Done

bnkai:
Can i test this now or are we waiting for #102  ?

endqwerty:
@bnkai test now. We've resolved its dependency on #102

bnkai:
@endqwerty things seem to work ok 

i  had to change
logger.Debugf("Deleting missing file: %s", t.Scene.Path)
to
logger.Infof("Deleting missing file: %s", t.Scene.Path) 
because i thought nothing was happening :-)
when indeed scenes with missing files where deleted.

I would also change  "Deleting missing file" to "Deleting scene with missing file" just to ne more clear

but thats it .

 Since @WithoutPants  approved i would also say ready to go

PR #151, opened 2019-10-21
Add logs to Logs page

WithoutPants:
Shows the past 30 log entries and streams any current logging. Shows newest entries first.

image

### DIFF
--- a/graphql/documents/data/log.graphql
+++ b/graphql/documents/data/log.graphql
@@ -1,0 +1,5 @@
+fragment LogEntryData on LogEntry {
+  time
+  level
+  message
+}
--- a/graphql/documents/queries/misc.graphql
+++ b/graphql/documents/queries/misc.graphql
@@ -54,6 +54,11 @@ query Stats {
   }
 }
 
+query Logs {
+  logs {
+    ...LogEntryData
+  }
+}
 query Version {
   version {
     hash,
--- a/graphql/documents/subscriptions.graphql
+++ b/graphql/documents/subscriptions.graphql
@@ -1,3 +1,9 @@
 subscription MetadataUpdate {
   metadataUpdate
 }
+
+subscription LoggingSubscribe {
+  loggingSubscribe {
+    ...LogEntryData
+  }
+}
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -37,6 +37,8 @@ type Query {
   """Organize scene markers by tag for a given scene ID"""
   sceneMarkerTags(scene_id: ID!): [SceneMarkerTag!]!
 
+  logs: [LogEntry!]!
+
   # Scrapers
 
   """Scrape a performer using Freeones"""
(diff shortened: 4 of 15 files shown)


bnkai:
I tested and it looks ok but i think you need to raise the level from debug to info
When scanning or cleaning the log gets full from screenshot alreadey present or file found respectively.You get this message for every single file.
Also due to the above if you press the  scan or clean task and switch to log before the task finishes the log seems stuck.The above is tested with 10k scebes in the database , idk if that happens when the database is small 
You have to switch from logs to tasks for example and back to log to get the final output updated

WithoutPants:
Added filter drop down that defaults to Info. I've not yet been able to reproduce the logging getting stuck so far.

bnkai:
10k+ scenes
set logging to debug
press scan from another tab
and you are stuck at this
<a href='https://postimg.cc/YLkkmnKF' target='_blank'>log-debug</a>
firefox was at 100% for some minutes, i couldn't wait for it to finish and ram usage climbed to 10G+
There are just too many things that get printed

set logging to info , as is default now and everything is ok

I know it's just an extreme case with many files and i don't see it as a problem for the merge now that we have the option to set the level

WithoutPants:
I've been testing this with a larger database, and there are definite performance issues when there are a larger amount of log lines. I had issues even with 300+ lines. I'll add a truncation function to ensure the displayed log doesn't get too big. I will also consider dropping the nice colour and formatting in favour of just plain text, as the sheer number of html elements appears to make the browser choke.

bnkai:
Even with this update firefox gets high usage for 2 minutes after the console outputs` Finished Scan` to catch up. This time it's not that unresponsive but is still needs 2 minutes of 100% usage and about 7G RAM( less than before now).If i just close the tab and open a new one i get the end of the log file instantly.I would say the js code is trying to update and catch up what it already got from the backend?
The same with cleaning.

Given the verbosity of  the debug outputs as you said and their mostly dev aspect , i would just ommit the debug level for the log page in the production build and be done with it.

WithoutPants:
The latest commits should be a cleaner and more robust implementation. Please retest and let me know. I do have some concerns around the server still potentially sending a lot of data that will be largely thrown out, but I'm thinking that it should be good enough.

Leopere:
Is there some form of JIT loading we can use for lengthy logs?

WithoutPants:
I'm not quite sure exactly what you mean. It uses graphql subscriptions via websocket to transmit the logs as they are generated to the client, which then displays them. The server buffers and throttles the logs so that batches aren't sent more than once per second. The UI truncates the displayed entries to 200 entries currently. My concern comes from the server queuing up a batch of 10000 log entries for example, then the UI throwing most of that away to just display the last 200 entries.

Leopere:
oh perfect

bnkai:
Ok now i think we are done :+1: 
Almost instanteneous results during the "torture" test with the 10k scenes.

Leopere:
@bnkai if you can resolve the conflicts I'll merge.

PR #154, opened 2019-10-23
Add logging options

WithoutPants:
Adds logging options to the config file and the Configuration UI:
- file to output logging to - leave blank to output to stderr
- flag to output logging to stderr even if also logging to file
- log level: minimum level to log
- flag to output http access to stdout

All of these options require a restart before changes are applied, with the exception of log level, which is applied immediately.

### DIFF
--- a/graphql/documents/data/config.graphql
+++ b/graphql/documents/data/config.graphql
@@ -4,6 +4,10 @@ fragment ConfigGeneralData on ConfigGeneralResult {
   generatedPath
   username
   password
+  logFile
+  logOut
+  logLevel
+  logAccess
 }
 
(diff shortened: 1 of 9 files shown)


bnkai:
Adding a log file to the configuration works fine. Log level selection also works.
As soon as you try to tick or untick ```Log to terminal``` and ```Log http acces``` you can't even save 
It shows this message in red ```Error: Network error: Response not successful: Received status code 422``` 
Stopping stash and manually editing the `logaccess` and `logout` options in the config file directly works as it should and shows the respective status as ticked or unticked in the UI.

WithoutPants:
Rebased and fixed conflicts.

PR #161, opened 2019-10-25
Use wss if origin uses https

### LINKED ISSUE #160: Mixed content errors when ssl/https enabled
## Issue
The new websocket connections creates issues for Stash deployments hosted behind a SSL layer. I'm using a letencrypt cert on my host and reverse proxying to the Stash docker container. This means Stash doesn't recognize itself as needing to use secure connections.

For reference: `REACT_APP_HTTPS` is not set.

## What it shows
Chrome, for example, will load a grey page and require manually allowing insecure connections for the page. 

The following errors are seeing in the console log.

Warning `Mixed Content: The page at '<URL>' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://<my_stash_url>/graphql'. This endpoint should be available via WSS. Insecure access is deprecated.`

Warning `Connecting to a non-secure WebSocket server from a secure origin is deprecated.`

Error `WebSocket connection to 'ws://<my_stash_url>/graphql' failed: Error during WebSocket handshake: Unexpected response code: 422`

WithoutPants:
Fixes #160 (hopefully)

Sets (always) the protocol for the websocket connection to wss if the origin connection protocol is https.  I don't have https set up to test this.

### DIFF
--- a/ui/v2/src/core/StashService.ts
+++ b/ui/v2/src/core/StashService.ts
@@ -23,8 +23,11 @@ export class StashService {
 
       if (process.env.REACT_APP_HTTPS === "true") {
         platformUrl.protocol = "https:";
-        wsPlatformUrl.protocol = "wss:";
       }
+    }
+
+    if (platformUrl.protocol === "https:") {
+      wsPlatformUrl.protocol = "wss:";
     }
     const url = platformUrl.toString().slice(0, -1) + "/graphql";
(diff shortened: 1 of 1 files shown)


Leopere:
This may be superfluous if you're using Traefik as your front end rev proxy however it might make sense to pursue regardless.

endqwerty:
👍

Leopere:
Since this fixes a recently created issue I'm willing to push this out since we still don't technically have a Production branch.

Issue #152, opened 2019-10-21
[EPIC] Better title creation from filename


timcantona:
It seems simple to do, but I was thinking the title should be a little smarter than just using the full filename.

Often there will be . or _ - etc instead of spaces, as well as extra data at the end 1080p, FULL, etc.

It would be nice if the title was a cleaned up version of the filename, for example:

**Filename**: claire-castel-at-the-service-of-theses-gentlemen-for-a-hard-dp-6699-1080p_full_mp4.mp4
**Title**: Claire Castel at the service of theses gentlemen for a hard dp 6699

and maybe even when a date is contained, that gets applied to the date field? 

**Filename**: dorcelclub.18.12.26.claire.castel.hot.night.in.club.xtrem.mp4
**Title**: Dorcelclub claire castel hot night in club xtrem
and date set to the 18.12.26 from the filename

Perhaps when #6 is implemented, you could be take more factors into account, for example in the last one you would recognise Dorcelclub as the studio, so the title would just become "Claire castel hot night in club xtrem"

Leopere:
This issue is notoriously a complex issue that is in most streaming self-hosted apps, and likely won't be fixed any time soon but is hugely useful.  One potential option is developing a scraper that compiles `.NFO` standard files and collects and stores any public domain stock imagery for scenes/preformers/etc...

Leopere:
There has been discussion in the Discord surrounding how to handle a lot of this but I think that eventually aiming for something in relation to how MusicBrainz and Picard handle tagging and file renaming and all of that fancy stuff will help make big leaps in the right direction.

I think a hugely useful potential direction will be to do things with Perceptual Fingerprinting via FFMPEG but that requires a whole new issue for discussion.

GernBlanston12:
If you implemented this alongside/under the auto-tagger for Performers and Studios, whatever is left of the file name after you drop Performers and Studios and separator charactors, would likely be the Scene name, right?

Leopere:
> Parsing text out of a file name via a list of known tags in the DB can't be that hard, can it?
It is a pretty tricky task to get right; there are a few companies that were established to help understand things like this; for example, Google was one.  

That said, we don't need to do such a complicated job as Google with something that is typically within the scope of just doing file identification and metadata curation. 

Adding things like facial recognition and p-hashing are just additional options for zeroing in on the data often times people who have a "stash" in the first place want something that can manage this similar to why MusicBrainz had been created powered by hobbyist collectors and curators.

[...]

PR #164, opened 2019-10-28
Scene filename to metadata parser

WithoutPants:
Adds a new scene filename parser page that can be accessed from the Tasks settings page.

Allows the input of a filename pattern. Fields are denoted with `{fieldName}`. Wildcards are denoted with `{}`. Everything outside of fields and wildcards are used literally in the search pattern.

Currently supports setting the title and date on scenes.

Currently supports the following fields:

- `title` - text captured within is set as the title of the scene
- `ext` - matches the end of the filename. It is not captured. Does not include the last `.` character, as the pattern is more readable as `.{ext}` (in my opinion)
- `d` - matches delimiter characters `-_.`. Not captured
- `i` - matches any ignored word. Ignored words are entered as space-delimited words in the parser input. Not captured. Use this to match release artifacts like `DVDRip` or release groups.
- `date` - matches `yyyy-mm-dd` and sets the date of the scene

It also supports the following partial date fields. The date will only be set on the scene if a date string can be built using the partial date components:

- `yyyy` - four digit year
- `yy` - two digit year. Assumes the first two digits are `20`
- `mm` - two digit month
- `dd` - two digit date

Also supports the following full date fields, using the same partial date rules as above:
- `yyyymmdd`
- `yymmdd`
- `ddmmyyyy`
- `ddmmyy`
- `mmddyyyy`
- `mmddyy`

A drop-down menu is available on the pattern to insert fields, and provides explanations for non-obvious fields.

Title generation also has the following options:

- whitespace characters: these characters are replaced with whitespace (defaults to `._`, to handle filenames like `three.word.title.avi`
- capitalize words: capitalises the first letter of each word

The results are presented in a table showing the existing and generated values of the title and date fields, along with a checkbox to determine whether or not the field will be set on each scene. The title and date can also be edited manually.

Finally, the Apply button updates all of the scenes based on the set fields.

Future features for this will be to add mapping options for tags, performers and studios, and saving and loading of parser options (which will be particularly important once the former feature is added). Importing and exporting of these options would also be beneficial as well.

### DIFF
--- a/graphql/documents/mutations/scene.graphql
+++ b/graphql/documents/mutations/scene.graphql
@@ -54,6 +54,12 @@ mutation BulkSceneUpdate(
   }
 }
 
+mutation ScenesUpdate($input : [SceneUpdateInput!]!) {
+  scenesUpdate(input: $input) {
+    ...SceneData
+  }
+}
+
 mutation SceneDestroy($id: ID!, $delete_file: Boolean, $delete_generated : Boolean) {
   sceneDestroy(input: {id: $id, delete_file: $delete_file, delete_generated: $delete_generated})
 }
--- a/graphql/documents/queries/scene.graphql
+++ b/graphql/documents/queries/scene.graphql
@@ -7,8 +7,8 @@ query FindScenes($filter: FindFilterType, $scene_filter: SceneFilterType, $scene
   }
 }
 
-query FindScenesByFilename($filter: FindFilterType) {
-  findScenesByFilename(filter: $filter) {
+query FindScenesByPathRegex($filter: FindFilterType) {
+  findScenesByPathRegex(filter: $filter) {
     count
     scenes {
       ...SlimSceneData
--- a/graphql/documents/queries/scene.graphql
+++ b/graphql/documents/queries/scene.graphql
@@ -7,6 +7,15 @@ query FindScenes($filter: FindFilterType, $scene_filter: SceneFilterType, $scene
   }
 }
 
+query FindScenesByFilename($filter: FindFilterType) {
+  findScenesByFilename(filter: $filter) {
+    count
+    scenes {
+      ...SlimSceneData
+    }
+  }
+}
+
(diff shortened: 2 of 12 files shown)


WithoutPants:
Should address #152 sufficiently.

bnkai:
From a little testing it seems to work fine.

I only noticed  that the date field is not verified / tested to be valid? Is it by design?
For a example on a few files i have the yymmdd and on others yyddmm not to mention some asian mmddyy. 
Is it up to the user to just untick the date when it doesn't seem valid ?  (for example 12-26-05 or 26-12-05 is obviously not valid ).If by chance you don't is says updated scenes ok althougth it just nulls the date while the title is renamed as it should. 

Regarding the usage i would add a few examples or a so to say recipes just to make it easier....

For example on scene releases from KTR or TRASHBIN and so on
pattern `{}.{yy}.{mm}.{dd}.{title}.{i}.{}.{ext}` with ignore `XXX`
seems to work well for now 

Tags,Studios and Performers  would be alot trickier i imagine since you have to check for existance in the DB

Great job @WithoutPants !

WithoutPants:
Some minor usability improvements. Added some sanity checking to date. Made it so that fields are only checked if they are different from the original value. Makes it easier to pick and choose what scenes to modify without having to worry about all of the unchanged ones.

bnkai:
@WithoutPants the changes work great .
I found a small bug though
If you hit find and select a few files from the list these files get their  **title** / **date** updated correctly but the unselected ones get an updated **updated_at** field  as a bonus :-)

WithoutPants:
@bnkai Should now be fixed.

bnkai:
Works fine now.

WithoutPants:
> Regarding the usage i would add a few examples or recipes maybe just to make it easier to understand ....

I've added a drop-down "recipe" field, with a few of the common scene formats I found. It will form the basis of a future save/load config, but will hopefully be a convenient feature for now.

Also made matching case insensitive.

Issue #163, opened 2019-10-26
Clean should remove scenes that are no longer covered by the library


WithoutPants:
Remove a directory from the stashes library and save the config, then click the Clean button in Tasks. No scenes are removed. 

The clean button should ensure that all scenes are covered by the currently configured stash directories, and remove entries for those that are not covered, in the same way as if the file were removed from the filesystem.

PR #169, opened 2019-10-29
Check if scenes are in the  library path when cleaning

bnkai:
It should clear #163 
It needs testing for the windows build though because of path separators.
Testing library paths in different drives would be also be ideal.


WithoutPants:
Tested against relative paths to cwd, different drive and network paths. All worked as expected and only removed expected scenes.

PR #174, opened 2019-11-03
Performers, Tags and Studio from scene filename

WithoutPants:
Marking as WIP until I have time to properly test.

This involved rewriting the existing filename parser from performing the parsing in the front-end, to performing it on the server. As such, the original functionality for #164 will need retesting.

Adds `{performer}`, `{tag}`, and `{studio}` fields to the filename parser. These fields work by extracting the field text, removing any delimiter characters (`.-_`) and performing a query for the resulting name. ~~It queries only for the first 10 results currently. It then attempts to case-insensitive match the name against all of the results, and sets if it is an exact match. Otherwise, it sets it to the first result.~~ For now, this matches against exact name only.

I've improved the UI to hide fields that aren't being parsed for, with the option to show these fields.

I'll try to give it a more rigorous test in the coming days, but I'd appreciate testing and feedback in the meantime.

### DIFF
--- a/graphql/documents/queries/scene.graphql
+++ b/graphql/documents/queries/scene.graphql
@@ -39,9 +39,15 @@ query ParseSceneFilenames($filter: FindFilterType!, $config: SceneParserInput!) 
       scene {
         ...SlimSceneData
       }
-      parserResult {
-        ...SlimSceneData
-      }
+      title
+      details
+      url
+      date
+      rating
+      studio_id
+      gallery_id
+      performer_ids
+      tag_ids
     }
   }
 }
--- a/graphql/documents/queries/scene.graphql
+++ b/graphql/documents/queries/scene.graphql
@@ -31,3 +31,17 @@ query FindScene($id: ID!, $checksum: String) {
     }
   }
 }
+
+query ParseSceneFilenames($filter: FindFilterType!, $config: SceneParserInput!) {
+  parseSceneFilenames(filter: $filter, config: $config) {
+    count
+    results {
+      scene {
+        ...SlimSceneData
+      }
+      parserResult {
+        ...SlimSceneData
+      }
+    }
+  }
+}
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -7,6 +7,8 @@ type Query {
 
   findScenesByPathRegex(filter: FindFilterType): FindScenesResultType!
 
+  parseSceneFilenames(filter: FindFilterType, config: SceneParserInput!): SceneParserResultType!
+
   """A function which queries SceneMarker objects"""
   findSceneMarkers(scene_marker_filter: SceneMarkerFilterType filter: FindFilterType): FindSceneMarkersResultType!
 
--- a/graphql/schema/types/scene.graphql
+++ b/graphql/schema/types/scene.graphql
@@ -85,8 +85,16 @@ input SceneParserInput {
 }
 
 type SceneParserResult {
-  scene: Scene!,
-  parserResult: Scene!
+  scene: Scene!
+  title: String
+  details: String
+  url: String
+  date: String
+  rating: Int
+  studio_id: ID
+  gallery_id: ID
+  performer_ids: [ID!]
+  tag_ids: [ID!]
 }
 
 type SceneParserResultType {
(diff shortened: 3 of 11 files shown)


bnkai:
With the latest commit i'm getting error as soon as i press find (default or even from recipes)
`Error: GraphQL error: Internal system error. Error <runtime error: invalid memory address or nil pointer dereference>`
ihave the error log here [error-default](https://pastebin.com/my9dN2xS)

WithoutPants:
My bad. Pushed a bad build. Recent commit should fix it.

bnkai:
Everything seems ok now.
The only thing i couldn't test extensively was the tag matching due to the lack of those from my DB filenames.
I thing it's ok for now since we might need a lot more feedback for the tweaking anyway.

Also as an improvement could we add an option to keep the matched fields in the title?
I for example would like to keep the studio or even the performers in the title even if they are matched.
If there was an option like lets say {+studio} that matches the studio but also adds that to the title it would be perfect

Leopere:
I'm not certain from @bnkai 's last reply as to whether or not this is still technically a WIP PR or not.

bnkai:
@Leopere sorry didn't make it clear imho it should merge so that we can get more feedback for fine tuning.It works and doesn't break anything.

Leopere:
@bnkai no worries I wasn't sure if it was WIP due to something undocumented so I'd rather be safe than sorry.

Ch00nassid:
I was asked to outline how i file parse in my VB app:

Example:
pc.19.09.14.kenzie.madison.make.you.cum.mp4

Create a "Leven Compare" string of the basic file name: 
PC 19 09 14 KENZIE MADISON MAKE YOU CUM
Leven compare?   https://en.wikipedia.org/wiki/Levenshtein_distance


=-=-=- Date finder
I use regex, which im sure is universal.

 Dim pattern As String = "\d{2}[\.]\d{2}[\.]\d{4}"
 Dim r As Regex = New Regex(pattern, RegexOptions.IgnoreCase)
 Dim m As Match = r.Match(dateinfilename)


=-=-=-Studio Gather
Try to find a studio
Gather all the studio name and match it against the basic leven string above

PC = princessCum

This is identified on the folder above but the filename has an alias short version.

My parser did not find a match based on filename alone.

=-=-=-=-=-Performer Gather

Gather all the Performers name and match it against the basic string above
Do the same for all known performer aliases
	Anything match? Store it in an array to compare more closeley (performers can have names that are close)


For each match
	(make sure a string can still match without spaces)  example of performer with false positive matches: Lana S

	another example, my script found "Kenzie Madison" from the file name.  I would then try to do a wildcard search on filename to confirm
find string: *Kenzie*Madison*  withing string: 	PC 19 09 14 KENZIE MADISON MAKE YOU CUM
	

	
	
=-=-=-=-=-=-LevenshteinDistance  Method
Complete a "BEST MATCH" string comparison (if required):
 
This will give you a number of changes required to make the strings match.  The lower the number the closer the match
If there are competing performers, usually the performer with the lowest number of changes to make the string match, is usually the best bet...

LevenshteinDistance(string1,string2)

```
			Public Function LevenshteinDistance(ByVal s As String,
													ByVal t As String) As Integer
					Dim n As Integer = s.Length
					Dim m As Integer = t.Length
					Dim d(n + 1, m + 1) As Integer

					If n = 0 Then
						Return m
					End If

					If m = 0 Then
						Return n
					End If

					Dim i As Integer
					Dim j As Integer

					For i = 0 To n
						d(i, 0) = i
					Next

					For j = 0 To m
						d(0, j) = j
					Next

					For i = 1 To n
						For j = 1 To m

							Dim cost As Integer
							If t(j - 1) = s(i - 1) Then
								cost = 0
							Else
								cost = 1
							End If

							d(i, j) = Math.Min(Math.Min(d(i - 1, j) + 1, d(i, j - 1) + 1),
											   d(i - 1, j - 1) + cost)
						Next
					Next

					Return d(n, m)
				End Function
```
				
I do my best to match exact performers / studios to others that might be in the file as well.	

I usually start my comparisons with an insane amount of changes (5000) and if a match is found, it sets the bar.  
If another performer is found, it needs to see if it can beat out the best match
I then start comparing string to see if it is the same performer or alias name


=-=-=-==-AFTER Math
After i have my "matches" I try to eliminate them from the basic string i started with:
And if we parsed well enough, we have something that might look like the title.

In my case i was left with:
PC   MAKE YOU CUM

I found:
KENZIE Madison
(If i had a studio alias, i *may have* been able to find PC = Princess Cum)
SEPT 14, 2019

PR #175, opened 2019-11-04
Use custom.css in same path as config file

WithoutPants:
Fix for #112 

Existing code was attempting to replicate the config file resolution code. It did not resolve `$HOME` correctly. As a result, if there were no existing `custom.css` file in the working directory, then saving custom css would not work.

I have simplified this to just use the same directory as the config file.


bnkai:
Works ok now.

PR #176, opened 2019-11-04
Fix ffmpeg error output

WithoutPants:
Setting the verbosity level to `quiet` means that even errors are suppressed. In addition, when an error does occur, the error message is written to stderr, not stdout. Changed the verbosity of the encoder calls, and changed to print stderr message in the event of an ffmpeg error. I tested this with some random files that ffmpeg doesn't like (but ffprobe is ok with). May be difficult to test unless you have some similar files.


bnkai:
The build and tests ok for me , i have no way of forcing an error as you said though.
Does using loglevel fatal instead of error work for your sample?

WithoutPants:
Loglevel error was the least verbose option I could use and still get the error output.

Issue #31, opened 2019-03-21
Native HEVC/x265 support


seducingbits:
x265 is the future. smaller file size by same quality. Or better quality by same file size :)

Right now I noticed that Stash supports x265 by transcoding the files to x264. For example, one of my files is ~3 GB at x265. The x264 transcode came out to be arounf ~ 9 GB, and it looked worse :/

StashAppDev:
I agree that x265 is great, the problem is [browser support](https://caniuse.com/#feat=hevc).  Both Chrome and Firefox have said that they will not support HEVC in their browsers, which means those types of files will always require transcoding in order to play properly in Stash.  I suppose you could compile a custom version of Chrome, but the average user is not going to do that.

The future of video on the web is really with the [AV1 codec](https://caniuse.com/#search=av1).  It's still pretty early, but all major browsers will support it.  The problem with this though is the encoding process is very resource intensive since there doesn't exist a hardware encoder for it yet.  I do want to replace x264 with AV1 for better file sizes and quality, but I think it's still too early.  I suppose I could add the option so the user can decide.

This still doesn't really solve the issue with HEVC though, but I don't think there is anything I can do.  Anyone have any ideas?

seducingbits:
Bummer! Did not knew that. But AV1 sounds great, haven't seen that in the wild yet, probably for the missing hardware encoding you mentioned.
What about live transcoding of x265? This works great for me with Plex.

bnkai:
I have also a lot of x265 and own encodes but browser support is a problem.

If you want to  try webm and vp9 is ok as long as you stick to 8bit encoding and firefox , chrome
and even android browsers work fine with stash and webm.
The compression level is almost the same with x265 but a little slower for the same quality.
I just batch encode my downloads wth ffmpeg and then just let stash handle the compressed webm files.

StashAppDev:
I already use webm / vp9 for some of the generated files, so it shouldn't be a stretch to use that for the transcodes.  I'll work on providing some configuration options so that you can choose the encoding parameters rather than how it's hard coded right now.

As for live transcoding, that's something I'd need to build support for.  Doable, but no ETA.

ExceptionalError:
Live transcoding is possible with [ffmpeg](https://trac.ffmpeg.org/wiki/StreamingGuide), too.
I am not sure how to consume the stream on the webpage though ...

WithoutPants:
Is this resolved now with the live transcoding? @Leopere @bnkai

[...]

Issue #10, opened 2019-02-11
User facing FFMPEG configuration


StashAppDev:
https://github.com/stashapp/StashServer/issues/41

WithoutPants:
stash currently generates various images and videos with the following settings:
**Screenshots:**
- time location: (0.2 * duration)
- quality: 2
- width: video width
- output format: jpg
- exposed via `/screenshot`

**Thumbnails:**
- time location: (0.2 * duration)
- quality: 5
- width: 320
- output format: jpg

**Note:** thumbnails are currently not used by anything

**Preview videos:**
- video chunks: 12
- width: 640
- chunk duration: 0.75s
- output format: mp4 (x264/aac)
- exposed via `/preview`

**Preview video webp:**
- width: 640
- fps: 12
- output format: webp
- exposed via `/webp`

**Scene marker videos:**
- width: 640
- duration: 20s
- output format: mp4 (x264/aac)
- exposed via `/stream`

**Scene marker images:**
- width: 640
- duration 5s
- output format: webp
- exposed via `/preview`

**Transcode videos:**
- width: original width
- output format: mp4 (x264/aac)

**Streaming transcodes**
- width: original width
- output format: webm

WithoutPants:
With #178 sorting out the transcode sizes, do we think there's anything else needed for this issue, or can it be closed?

Leopere:
This was requested to be addressed on November 17th 2019 I'll close this pending any further commentary.

PR #178, opened 2019-11-04
Add configurable maximum transcode sizes

WithoutPants:
Adds options to the configuration page to allow setting a maximum size for generated transcoded files, and for streaming transcoded files.

Default value is "Original" which uses the existing resolution of the source video. Other values are "240p", "480p", "720p", "1080p", "4K". The transcoding code uses this as the maximum size to down-scale to. If a scene has a lower resolution than the max, then that resolution is used. This scaling is applied to the height for normal landscape videos, and applied to the width for portrait videos.

Should help address the performance issues identified in #31. Related to #10.

### DIFF
--- a/graphql/documents/data/config.graphql
+++ b/graphql/documents/data/config.graphql
@@ -2,6 +2,8 @@ fragment ConfigGeneralData on ConfigGeneralResult {
   stashes
   databasePath
   generatedPath
+  maxTranscodeSize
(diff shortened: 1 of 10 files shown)


bnkai:
Testes for transcode and livestreamsizes , everything seems ok.
Now even 1080p live transcoded files can be seen without freezes.

PR #181, opened 2019-11-05
Stop tasks and show task progress

### LINKED ISSUE #177: Ability to stop scan/generation
As discussed in the Discord-channel, we should have the ability to stop scans/generations/re-encodes via the webgui.

WithoutPants:
Fixes #177 

Adds current task status in the Tasks setting page. Shows the current job, and where possible the progress.

Also adds a stop button. Clicking the stop button instructs the system to stop the running task. The limitation of this is that it will stop processing at the next scene/file. This means that unfortunately, it will not currently kill a hanging ffmpeg process. Does not currently apply to import/export.


Leopere:
If this is "good enough" we should create an issue focused on making it ideal however I will merge this since it looks and sounds ready.

PR #187, opened 2019-11-08
Update index.scss

### LINKED ISSUE #186: [bug] z-index property conflict with bp3-navbar
The new `.grid-item label.card-select` class shares the same z-index property value with the `.bp3-navbar` class. When the scene cards scroll beneath the navbar, the new scene tick boxes overlap the navbar content as a result.

Annotation 2019-11-07 185840

I think the simple fix would be to change the z-index value for the following class

```
.bp3-navbar {
     z-index: 99;
}
```

It worked for me in ChromeDevTools. Please confirm.

echo6ix:
Fixing a z-index property conflict with the bp3-navbar class identified in fixes #186

### DIFF
--- a/ui/v2/src/index.scss
+++ b/ui/v2/src/index.scss
@@ -91,7 +91,7 @@ code {
   position: absolute;
   padding-left: 15px;
   margin-top: -12px;
-  z-index: 10;
+  z-index: 9;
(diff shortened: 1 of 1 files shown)


bnkai:
Looks ok. Tested ok also.

PR #188, opened 2019-11-08
Calculate and print job totals for scan and generate tasks

bnkai:
Calculates the number of new files we need to scan for the scan task and the total number of sprites,previews,transcodes,markers we need to generate.

### DIFF
--- a/pkg/manager/manager_tasks.go
+++ b/pkg/manager/manager_tasks.go
@@ -70,7 +70,7 @@ func (s *singleton) Scan(nameFromMetadata bool) {
 		}
 
 		total := len(results)
-		logger.Infof("Starting scan of %d files.%d New files found", total, s.neededScan(results))
+		logger.Infof("Starting scan of %d files. %d New files found", total, s.neededScan(results))
 
 		var wg sync.WaitGroup
 		s.Status.Progress = 0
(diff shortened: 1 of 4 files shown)


pkg/manager/manager_tasks.go
```diff
@@ -70,7 +70,7 @@ func (s *singleton) Scan(nameFromMetadata bool) {
 		}
 
 		total := len(results)
-		logger.Infof("Starting scan of %d files", total)
+		logger.Infof("Starting scan of %d files.%d New files found", total, s.neededScan(results))
```
WithoutPants: Add a space between `.` and `%d`.

pkg/manager/manager_tasks.go
```diff
@@ -155,6 +155,8 @@ func (s *singleton) Generate(sprites bool, previews bool, markers bool, transcod
 			logger.Info("Stopping due to user request")
 			return
 		}
+		totalsNeeded := s.neededGenerate(scenes, sprites, previews, markers, transcodes)
+		logger.Infof("Generate Starting.Generating %d sprites %d previews %d markers %d transcodes", totalsNeeded.sprites, totalsNeeded.previews, totalsNeeded.markers, totalsNeeded.transcodes)
```
WithoutPants: Suggest improving levity and formatting: 
```
logger.Infof("Generating %d sprites, %d previews, %d markers, %d transcodes", totalsNeeded.sprites, totalsNeeded.previews, totalsNeeded.markers, totalsNeeded.transcodes)
```

WithoutPants:
Seems to test ok. I'm wondering if this will add unnecessary overhead to scans/generates on slow links. It's been reported that scanning/generating is already slow on google drive for example, due to the amount of walking the directory that the scanner does to get the total workload. I'm concerned that this will slow it even further for such users.

bnkai:
@WithoutPants the time for the exra directory scan is negligible for lans.It's almost the same time from pressing scan to seeing the message Starting scan from the "plain" scan version.It's only penalty is the doublestar glob time the rest is db and local fs. 
I don't have access to gdrive to test but i think the extra time needed won't be even close to the scan/generate of a single file from there and if you are using a slow link anyway it makes a lot more sense to know exactcly how many files you need to generate.
I'll ask around on discord to get an estimate on the that.

bnkai:
Ok got feedback from Charlie on discord , even with a big collection on gdrive the needed time for doublestar is indeed negligible.

WithoutPants:
I'm satisfied that this is good to merge. @Leopere

Leopere:
Written by @bnkai and verified by @WithoutPants

PR #190, opened 2019-11-10
[WIP] README: Added additional notes to dev install

echo6ix:
* Go download url/instructions
* `yarn install`
* `mingw32-make` for Windows


bnkai:
I would also add that yarn requires a fairly new nodejs as a warning (if it spits out an error it is usually from there) but otherwise it seems ok to me.

WithoutPants:
I think this is good enough as is to merge @Leopere

Issue #194, opened 2019-11-12
Regression in marker creation / modification


StashAppDev:
Two issues:

1) When creating a new scene marker the scene will no longer update to display the newly created marker.  A hard refresh will show it.
2) After creating a new scene marker it is no longer possible to modify the tags associated to that marker.

I get the feeling the issue lies somewhere in the `FilterMultiSelect.tsx` file.  It might be related to this pull request (https://github.com/stashapp/stash/pull/92), but I have not confirmed.

StashAppDev:
@WithoutPants I've confirmed the causes of the issues

1) Creating / deleteing scene markers no longer updates the page due to this change: https://github.com/stashapp/stash/pull/167/files#diff-ed7b93b175ef2f33dfecc3d28b8b3644L149

2) Not being able to modify the tags associated to a marker is due to something in https://github.com/stashapp/stash/pull/92.  Reverting this diff fixes the issue.

Issue #166, opened 2019-10-28
Creating and deleting performers and studios does not refresh the applicable view


WithoutPants:
Create a new performer or studio, then return to the performer/studio page. The new item is not added to the list. It requires a page refresh to show the new item. Likewise, the item is still shown in the list after returning to the page after deleting.

Appears to be something to do with graphql's cache.

WithoutPants:
There is likewise an issue with the select boxes after creating a new tag, studio or performer. 

Create one of these, then navigate to edit a scene to add the new tag/studio/performer. The new entry will not be present.

There are a number of ways to address this, with varying complexity:

- change all requests to `network-only`. This means that nothing is cached, and all requests will go to the server. The data is essentially guaranteed to be valid and current, but at the cost of increased queries to the server.
- change mutation calls to update the cache after mutation. Based on my reading, this is error-prone and complicated. You would need to iterate through the cache for _any_ query impacted, and adjust it. For some queries, this will be simple. But for others it could be quite complex. For example, a mutation to remove a tag from a scene would need to go through any queries that have been called that filter on that tag and remove the scene from it, plus adjust the count. In addition, it would also need to adjust possible scene_marker queries that filter by scene tag. This will be complex to implement and is likely to be error prone.
- For any call that mutates the data, call `ApolloClient.resetStore`. This wipes the client cache. This means that any subsequent queries will need to hit the server again, but it also means that any data that it gets is now valid and correct. The downside to this is that unrelated queries now need to be refetched. This seems like a decent expense to take in order to get current data.

Ideally, we would also invalid the cache when certain backend tasks are completed, such as scanning, cleaning or importing.

StashAppDev:
Reopening due to https://github.com/stashapp/stash/issues/195#issuecomment-553115201

thundergreen:
When Trying to create a new Actor i get this: 

```
Error: GraphQL error: Internal system error. Error <runtime error: invalid memory address or nil pointer dereference>
```

WithoutPants:
Please raise a new issue with the details. The stack trace should be output to the console when this happens, so please include that in there as well.

PR #196, opened 2019-11-12
Refetch tags on scene update #195

### LINKED ISSUE #195: [bug?] Tags disappear in edit panel when saving to new scene
Recently on the latest builds, tags that are added to new scenes are disappearing from the tag input box in the edit panel when clicking the save button.

To recreate scenario: Add new scene -> Scene edit panel -> Add tags - > Click the save button

The `updated scene` notification banner appears, then the tags disappear. A page refresh is required to see them in the edit panel.

insta

WithoutPants:
Fixes #195. Minor regression on #166 

Refetches `AllTagsForFilter` so that tags drop-down isn't wiped. Causes scene count on tags page to have possibly stale data.


WithoutPants:
Added fix for #194 .

StashAppDev:
Just tested it.  This is pretty close, but It doesn't completely fix #194.  Create a new marker and then try to edit that marker to add more tags.  Nothing happens.

WithoutPants:
That issue was related to #92 like you said. The `initialIds` meaning changed to be value. Updated to show current value of scene marker tags instead of initial value.

StashAppDev:
Awesome work!  That fixed it.

Issue #34, opened 2019-03-24
Add a task for auto assigning performers


iwiwdt:
It would be very useful if it were possible to scan the files and if there is a performer full name in the file path, like in the file name or some directory further up, then the performer is very likely in the file.

StashAppDev:
I actually had something like this in the old version which I never did commit.  Mine was a bit different though in that it would iterate through each performer and look for the performer name or aliases in the details, title, or path.  If it found a match then it would prompt the user if they'd like to attach that performer to that scene.

I do think this is a good idea, I just need to think through the implementation details.

WithoutPants:
@Leopere can this be closed?

Leopere:
Looks like it can be.

PR #204, opened 2019-11-15
Add scene auto-tagging from filename

### LINKED ISSUE #6: Tagging based on file name
https://github.com/stashapp/StashServer/issues/33

WithoutPants:
Adds an "Auto tag" section to the tasks settings page. Here you can tick whether to include performers, studios and tags. Clicking the button starts the auto-tagging job.

image

The auto-tagger loops through all of the performers/studios/tags in the database, and searches for scenes that match the following the name - using the following regex to handle word separators: `(?: |\.|-|_)?`. For each scene it finds, it adds the performer/studio/tag to the scene.

Also adds buttons to the performer/studio and tags pages which runs the auto-tag process for a specific performer/studio/tag.

image

image

Should resolve #6 and #34

### DIFF
--- a/graphql/documents/queries/settings/metadata.graphql
+++ b/graphql/documents/queries/settings/metadata.graphql
@@ -14,6 +14,10 @@ query MetadataGenerate($input: GenerateMetadataInput!) {
   metadataGenerate(input: $input)
 }
 
+query MetadataAutoTag($input: AutoTagMetadataInput!) {
+  metadataAutoTag(input: $input)
+}
+
 query MetadataClean {
   metadataClean
 }
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -75,6 +75,8 @@ type Query {
   metadataScan(input: ScanMetadataInput!): String!
   """Start generating content. Returns the job ID"""
   metadataGenerate(input: GenerateMetadataInput!): String!
+  """Start auto-tagging. Returns the job ID"""
+  metadataAutoTag(input: AutoTagMetadataInput!): String!
   """Clean metadata. Returns the job ID"""
   metadataClean: String!
 
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -9,6 +9,15 @@ input ScanMetadataInput {
   nameFromMetadata: Boolean!
 }
 
+input AutoTagMetadataInput {
+  """IDs of performers to tag files with, or "*" for all"""
+  performers: [String!]
+  """IDs of studios to tag files with, or "*" for all"""
+  studios: [String!]
+  """IDs of tags to tag files with, or "*" for all"""
+  tags: [String!]
+}
+
 type MetadataUpdateStatus {
   progress: Float!
 	status: String!
--- a/pkg/api/resolver_query_metadata.go
+++ b/pkg/api/resolver_query_metadata.go
@@ -27,6 +27,11 @@ func (r *queryResolver) MetadataGenerate(ctx context.Context, input models.Gener
 	return "todo", nil
 }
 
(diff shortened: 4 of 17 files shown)


ExceptionalError:
Could you add a seperator in front and after the name, otherwise there will be false positives.
Also rather use this as separator: ```sep= "[. _-]+"```

```
regex = '(?:^|sep)' + name.replace(' ', sep) + '(?:sep|$')
```

Ch00nassid:
Omg i love this feature!!!!

Leopere:
@WithoutPants the TravisCI build failed due to TravisCI being Travis again.

Leopere:
Yep nevermind this time it wasn't.

WithoutPants:
> Could you add a seperator in front and after the name, otherwise there will be false positives.
> Also rather use this as separator: `sep= "[. _-]+"`
> 
> ```
> regex = '(?:^|sep)' + name.replace(' ', sep) + '(?:sep|$')
> ```

Definitely see the benefit of checking before and after name, but you don't think the separator should be optional within the name? ie `xxx.FirstNameLastName.mp4`

ExceptionalError:
> Definitely see the benefit of checking before and after name, but you don't think the separator should be optional within the name?

Ah, yes. So then
```
'(?:^|[. _-]+)' + name.replace(' ', [. _-]*) + '(?:[. _-]+|$')
```

WithoutPants:
Changed to only log when performer/studio/tag added. Improved regex to prevent false positives.

cloudstrife11:
Is this already out for using?

Ch00nassid:
Will this find stars and studios i might have missed?  OR is this just looking for tags?

WithoutPants:
> Will this find stars and studios i might have missed? OR is this just looking for tags?

It only operates on performers, studios and tags that are in the system already but not applied to scenes.

ExceptionalError:
Since this is a manual action it would be nice if it would autotag Performers and Scenes that have already some entires.

WithoutPants:
> Since this is a manual action it would be nice if it would autotag Performers and Scenes that have already some entires.

It already does this. To clarify the above comment, it doesn't detect if there are performers/studios/tags you haven't added to your system. I may have misunderstood @Ch00nassid's original question.

Ch00nassid:
No, you understood fine.  Im worried about my current scenes having missed a current performer or studio i already have in my collections.  I'd like the option to review the findings and make final approvals but this sounds like an automated "seek and destroy" or "seek and input" ;)

bnkai:
From a quick test the performers name regex seems to work better than all (at least in my DB).
The studios won't match if the filename begins with Studio eg Studio.date.Performers....
Auto tagging in the tags list doesn't update the counter
The using regex info msg you print when iterating/autotagging performers is useful also for the tags,studios to at least know that something is happening

For all of them and especially the tags that i could test better you get maximum 25 matches.

i have for example a specifc tag in 200+ filenames and it only matches 25.
i added a new performer with over of 30 scenes, only 25 matched.
autotagging again doesn't do anything to match the remainig untagged performers,scenes,etc....

Leopere:
I added this as I think it's important to this kind of stuff.  https://github.com/stashapp/stash/issues/221

bnkai:
on more
The regex used doesn't match studios,tags or performers if they are the first thing in the filename.
eg 

`WildOnCam.19.04.12.Sydney.Cole.And.Victoria.June.Lesbian.XXX.720p.MP4-KTR.mp4`

studio WildOnCam not autotagged

if you create a tag called WildOnCam that won't get autotagged also

if the filename is 
`Sydney.Cole.And.Victoria.June.Lesbian.XXX.720p.MP4-KTR.mp4`
performer Sydney Cole won't get autotagged also while Victoria june will be

WithoutPants:
Issue was due to the regex pattern not considering path separators in the scene paths. Added a test for this particular scenario. Should be ready to retest @bnkai

bnkai:
I can confirm everything now seems ok

PR #205, opened 2019-11-15
Fix panic on import with existing scene

WithoutPants:
To reproduce, export then re-import into the same database. The first scene fails to import due to the unique constraint on the checksum. However, this was causing a panic because the `scene` value was nil and being dereferenced for the log text.

### DIFF
--- a/pkg/manager/task_import.go
+++ b/pkg/manager/task_import.go
@@ -443,7 +443,7 @@ func (t *ImportTask) ImportScenes(ctx context.Context) {
 		scene, err := qb.Create(newScene, tx)
 		if err != nil {
 			_ = tx.Rollback()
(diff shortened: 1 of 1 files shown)


bnkai:
@WithoutPants @Leopere  indeed the fix is needed and ok for merge

Not related to the FIX code
@WithoutPants for the import / export stuff i think we need  to clarify #206 though first.Depending on that i can't understand how you can export and then import and get duplicate checksum.The DB is supposed to be cleared?.Unless the list is custom made and has scenes with the same checksum?

WithoutPants:
> Not related to the FIX code
> @WithoutPants for the import / export stuff i think we need to clarify #206 though first.Depending on that i can't understand how you can export and then import and get duplicate checksum.The DB is supposed to be cleared?.Unless the json is custom made and has scenes with the same checksum?

I'll address this in #206

PR #214, opened 2019-11-16
[WIP] Removed vendor folder and bumped some dependency versions

StashAppDev:


WithoutPants:
Are we moving to a module proxy at the same time?

StashAppDev:
It's supported by default in Go 1.13.  As long as you're on the latest version it should "just work".  Might need to set this environment variable if it doesn't `GOPROXY=https://proxy.golang.org,direct`

WithoutPants:
Understood, but we'd want to also enable it in the CI builds.

WithoutPants:
One other consideration is what happens in docker CI builds with this? With the existing method, docker has all of the dependencies already in vendor directory. When `cross-compile.sh` is run, does the docker container then have to pull down the dependencies that travis already got in the pre-deploy stage?

Leopere:
Marked as WIP until the conversation is resolved.

Leopere:
For now, I'm going to close this, but we can re-open and revisit in the future if we wish.

Issue #206, opened 2019-11-15
Import improvements


WithoutPants:
There's a few issues with the import functionality as it is currently implemented.

- import bombs out of importing objects as soon as it encounters an error, rolling back the entire transaction. In my opinion, objects should be imported in their own transaction, so the failure of importing one scene does not prevent the rest of the scenes from importing.
- importing tags does not check for existing tags, so duplicate tags are created if they already exist
- in my opinion, the importer should handle existing data, overwriting it with the imported data

Before I look at implementing these changes, I wanted to confirm if this is consistent with the general view of expected import behaviour.

bnkai:
@WithoutPants afaik the importer was only meant for migrations and database schema upgrades.
The first thing it does is database.Reset()  which removes the sqlite db file.
That's why it doesn't check for anything and probably why it packs everything in one transaction.
@StashAppDev  is needed to verify though

Imho there should also be a separate non destructive import so that users could share perfomers,studios,tags for example.(i imagine scenes would be tricky with the path)

StashAppDev:
To echo what @bnkai said... the import should be dropping the database right at the start.  It shouldn’t be creating duplicates and failing the entire import on error was intended.  I wanted to ensure that no data was skipped from being imported so I was pedantic about errors.  If we want to import without nuking current data I feel that should either be a new task or a flag for the existing task.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Leopere:
Ensure that there are sufficient UX notifications to let the end-user know that they are about to dump any preexisting data.

WithoutPants:
On my Windows machine, the database is definitely not deleted prior to import, so looks like there's a bug there.

I'll look at doing the update import as a separate mode.

bnkai:
`func Reset(databasePath string) {
	_ = DB.Close()
	_ = os.Remove(databasePath)
	Initialize(databasePath)
}`
On linux at least i verified that the sqlite db file gets deleted and created again.

WithoutPants:
On Windows, I added logging and better error reporting if the database `Reset` fails and got the following output:
```
time="2019-11-16T19:58:28+11:00" level=error msg="Error resetting database: Error removing database: remove stash-go.sqlite: The process cannot access the file because it is being used by another process."
```

I tried adding a `sleep` between the `Close` and the `os.Remove` call, but even with a 5 second sleep, I get the same issue. So, on Windows at least it's not working as expected.

[...]

PR #215, opened 2019-11-17
Fix Windows database.Reset

WithoutPants:
Added `error` return value to `database.Reset()`. Added `close` call to `runMigrations`. Adding this allowed the database file to be closed properly and the import to run successfully in Windows.


bnkai:
The error was not present in linux but i can confirm it doesn't interfere with the import / export code ...

WithoutPants:
Related to #206

PR #217, opened 2019-11-17
Refresh config when paths change

### LINKED ISSUE #207: [BUG] Immediatelly after setup screenshots  and generate tasks can't be created.
After the initial setup through the web setup page i think the config info isn't propagated to the manager.As a result the first run of scan or generate without a restart uses null as the generated / metadata dir and so screenshots and metadata get a plain "screenshots" or "vtt" directory (instead of the metadata+screenshots one) that depending on chance could work.
As soon as you stop the app and restart it , it reads properly the config file and all is well
I think this could also be related to #83 #114 as the watchconfig on the manager side would notice the update on the config file.
To reproduce you can just rename the .stash directory and do a clean setup with different metadata dir , just make sure you don't run stash from within the metada directory so you can see the ffmpeg errors.
I tried a bit running the manager.Initilalize() after the config.Save in api/server.go but it didn't work and i'm not sure if thats the right way anyway.
Force restarting the app after the setup is complete might be a little brute force ?

WithoutPants:
Fixes #207 

Calls `RefreshConfig` to reset the paths object when the paths are changed at init or subsequently in the config page.

### DIFF
--- a/pkg/api/resolver_mutation_configure.go
+++ b/pkg/api/resolver_mutation_configure.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 
 	"github.com/stashapp/stash/pkg/logger"
+	"github.com/stashapp/stash/pkg/manager"
 	"github.com/stashapp/stash/pkg/manager/config"
 	"github.com/stashapp/stash/pkg/models"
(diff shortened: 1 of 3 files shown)


bnkai:
Tried this on linux and works fine.

PR #218, opened 2019-11-17
Use justified layout for wall

WithoutPants:
The current wall layout has issues when portrait and landscape scenes are mixed in on the same row. The scene thumbnails end up being truncated. This change uses flickr's justified-layout to lay the scene thumbnails in a justified grid. In my opinion it results in a much cleaner layout. It's a bit hard to show the comparative results with SFW images, so you might just need to see for yourself.

### DIFF
--- a/ui/v2/package.json
+++ b/ui/v2/package.json
@@ -19,6 +19,7 @@
     "bulma": "0.7.5",
     "formik": "1.5.7",
     "graphql": "14.3.1",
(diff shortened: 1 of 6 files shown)


bnkai:
Couldn't find a lot of samples but there ...
original
original
flickr
flickr
the snapshot dimensions are a little different but you hopefully you can spot the difference

works fine for me and looks better imho

PR #222, opened 2019-11-19
Add test and integration tests to the Makefile

WithoutPants:
Currently, travis runs `go test` which only tests the top-level package. Added `test` and `it` targets to the Makefile. `test` performs unit tests on the entire package tree. `it` performs integration tests (denoted in the test files with `// +build integration`).

I've changed the travis file to run `make test` on all builds, and `make it` on master builds, though this could possibly be changed to include the develop branch as well.

The change to `generator.go` was the result of running `make test` finding an error. This was latent due to travis running `go test` only.

Finally, I've added a `gitattributes` file to prevent the `go.mod` and `go.sum` files from showing as modified on Windows after a build.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+go.mod text eol=lf
+go.sum text eol=lf
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,8 @@ before_install:
 script:
 #- make lint
 #- make vet
-- go test -mod=vendor
+- make test
+- if [ "$TRAVIS_BRANCH" = "master" ]; then make it; fi
 before_deploy:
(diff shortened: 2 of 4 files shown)


WithoutPants:
I'm wondering if we should just run the integration tests for all branches. Makes sense when evaluating pull requests. @StashAppDev @Leopere

Leopere:
Yeah, I would say integration tests for everything is nice if it doesn't cost us a colossal workload.

WithoutPants:
Changed so that integration tests are always run.

Leopere:
Ready?

WithoutPants:
`go test` unfortunately does not allow excluding files without build tags. This means that `make it` also runs the non-integration tests. I have clarified this in the Makefile, and removed `make test` from the travis configuration since it is redudant when `make it` is invoked.

Provided the travis build is successful, this should be good to merge.

PR #225, opened 2019-11-26
Config should be read from cwd before user profile

### LINKED ISSUE #219: Config should be read from cwd before user profile
Current behaviour is to look for the config file in the `$HOME/.stash` directory, and only if it is not found there, check the current working directory. In my opinion this should be reversed. The current working directory should always be checked first, falling back to the user directory if not found. 

I think also that the system should print the config file that it ends up reading from. It can be confusing to find that the system is running from a specific config and not being able to check. It could perhaps be printed in the configuration UI as well.

Finally, add the ability to specify the config file location from the command line. This will make running and testing multiple deployments easier.

I realise that this probably won't affect 90% of users, but this will improve things for people that test changes in a separate deployment.

StashAppDev:
fixes #219

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -25,6 +25,11 @@ type singleton struct {
 var instance *singleton
 var once sync.Once
 
+type flagStruct struct {
+	configFilePath string
+}
+var flags = flagStruct{}
+
 func GetInstance() *singleton {
 	Initialize()
 	return instance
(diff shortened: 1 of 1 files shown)


bnkai:
Tested in linux , it  seems ok

- checking first in cwd _ok_
- checking in HOME/.stash if the above doesn't exitst _ok_
- using filename argument  from `-c` flag  _ok_

PR #230, opened 2019-11-28
Loop and autostart flags. Save interface options

### LINKED ISSUE #147: [Request] Ability to loop video
Loving this app, but I'm having one minor issue. I have plenty of short clips that i would like to be able to loop because having to click the restart after it finishes is an inconvenience. Other than that great work guys!

WithoutPants:
Changes interface options to be saved in the configuration file, to make them persistent.

Closes #147

Added interface option "Maximum Loop Duration". This field, defaulting to 0, determines the maximum duration in seconds for a scene, under which the scene player will loop the video. For example, if this field is set to 10 seconds, then scenes under 10 seconds will be looped.

Added interface option "Auto-start video", defaulting to false. When enabled, videos will be auto-started when navigating to the scene player.

### DIFF
--- a/graphql/documents/data/config.graphql
+++ b/graphql/documents/data/config.graphql
@@ -13,6 +13,10 @@ fragment ConfigGeneralData on ConfigGeneralResult {
 }
 
 fragment ConfigInterfaceData on ConfigInterfaceResult {
+  soundOnPreview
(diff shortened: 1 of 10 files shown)


bnkai:
Everything seems to work ok

PR #231, opened 2019-11-28
[BUG] Fix previews generation bug

bnkai:
During the preview generation task on some files (very small percentage) the process seemed to take too much time to complete.
Upon investigation it  seems that the trimmed files to be concatenated had weird durations / timings and so the resulting mp4 file generated had also weird fps (0.001) making the generate task stuck to a single scene for hours.
Simply changing the order of the ffmpeg  `-t` parameter after the input `-i` fixes all weird occurences and goes past the problematic files as it should.
I couldn't find anything in common for the files affected (original,reencodes,mp4,webm ) , i could only verify through discord that the specific change is indeed generating correct previews with the problematic media files.


StashAppDev:
Hmm, interesting.  According to [this](https://trac.ffmpeg.org/wiki/Seeking) it seems like only the location of the `ss` flag would affect the processing, I can't find any info on the `t` flag.  No big deal though, I'll go ahead and merge.

Issue #220, opened 2019-11-18
[Discussion] UX/UI Design Guidelines


echo6ix:
Opening this issue to start a discussion on any possible UX UI design guidelines to achieve a consistent, clean, and simple design throughout Stash. Full disclosure, I'm not a designer and I don't really know how to write design guidelines.

Some (hopefully) relevant notes I'd add to start this off:

- Keep Stash looking simple and remember it needs to scale to mobile viewports
- There are similar apps to Stash, but their UIs are overwhelming with their disorganized and inconsistent implementation of features. Stash can offer a rich feature-set and present it in a well organized and consistent manner.
- Don't recreate the wheel. Please use established sites like [vPorn](http://vporn.com), [xHamster](http://xhamster.com), [SpankBang](http://spankbang.com), or [PornHub](http://pornhub.com) for _inspiration_ on an optimal UI. These sites make excellent use of screen real estate while providing necessary scene info, performer info, gallery info, etc. 
  - Eg. Use of overlaying (z-index) HTML DIVs on scene thumbnails and performer thumbnails, where resolution, scene length, scene count, and performer fav'ing on mouseOver are all placed in overlays instead of stacking rows of text above or beneath the thumbnails.
- Design consistency between each of Stash's main sections (Scenes, Performers, Markers, Tags, etc).
- The filter container is Stash's version of the global search box. It should be viewable in each section of Stash from scenes to tags, and it's filters and parameters should adapt to the section it is located in.
- Increase simplicity through reduction of redundancy/reduction of duplicate design. For example the current filter container can be optimized as follows:
  - The `grid`, `list`, and `wall view` buttons that have icons do not require text labels. Text labels should be in tooltips.
  - The `filter` button, `select none`, and `select all` buttons can be minimized by using their corresponding icons (currently with Blueprint) and mouseover tooltip text label.
  - Those few modifications alone would allow the filter container to fit in a smaller viewport and just give Stash a more simplistic look without compromising any features.

Note: A compilation of UI UX related suggestions and concepts can be found on issue #108

Leopere:
Please add this stuff when its solidified as a proper spec to #211 under an independent UI section.

Issue #108, opened 2019-08-20
[EPIC] [RFC] Front-end user interface discussion


echo6ix:
Compilation of UI UX related suggestions

- [x] [Clean, simple, and functional design](#clean-simple-and-functional-design)
- [ ] [Tags multi column list](#tags-multi-column-list)
- [x] [Tags thumbnail view](#tags-thumbnail-view)
- [ ] [Tags alphabetical side panel list](https://github.com/stashapp/stash/issues/108#issuecomment-524760553)
- [x] [Scene cards thumbnail overlays](#scene-cards-thumbnail-overlays)
- [x] Scene cards custom thumbnail (let user upload thumbnail or select timestamp from scene)
- [x] [Performer page UI paradigm](https://github.com/stashapp/stash/issues/124) from #124 
- [ ] [Performer page side panel alphabetical performer list](https://github.com/stashapp/stash/issues/118) from #118 
- [x] [Performer page performer tags and UI implementation](https://github.com/stashapp/stash/issues/120) from #120 
- [x] [Channels UI paradigm](https://github.com/stashapp/stash/issues/115#issuecomment-525997244) (aka parent and child studios, also including DVD collections as channels) from #115
- [x] [Image gallery UI paradigm](https://github.com/stashapp/stash/issues/123) from #123 
- [x] [Adaptive interface](https://github.com/stashapp/stash/issues/87) from #87 and #192 
- [x] [Infinite scrolling](https://github.com/stashapp/stash/issues/47) from #47 and #192 

# Clean, simple, and functional design
TLDR Lets not recreate the wheel for a clean, simple, and functional UI UX. We can take inspiration from sites like xHamster, PornHub, vPorn, etc. 

# Tags multi column list
Using a single column list is a poor use of space on any viewport for the default tag list view. A mutli-column flexbox list is probably the best use of space ala [this xHamster tag page layout](https://xhamster.com/categories).

![stash1](https://user-images.githubusercontent.com/37937507/63389448-bbc1a500-c379-11e9-9409-e7719ed87be1.png)

# Tags thumbnail view
Re: #5 Thumbnails view could be inspired by Pornhubs concept [found here](https://www.pornhub.com/categories). Where the list of tags would be on a left side bar and the thumbnails populate the majority of the viewport.

[...]

PR #232, opened 2019-11-28
Scene ui improvements

WithoutPants:
This is a collection of UI improvements for the scene list page. The improvements come from #220 and #108 
Each improvement is done as a separate commit so that we can pick and choose which to include.

### Scene card overlays
I lifted the code from @echo6ix 's fork and tidied it up. It removes the filesize, duration and resolution from below the thumbnail, and overlays the resolution and duration to the bottom right of the thumbnail. See below:

image

### Handling portrait videos
I've been trying to get a consistent height shown for the scene videos when there is a mix of portrait and landscape videos. The CSS has proved a bit of a nightmare. Here is the result so far:

image

The intent is to line up the thumbnails so that the title appears in the same place vertically. This appears to be working correctly, though I notice a bit of a performance hit, so I'm going to need to revisit it. Note also that the presence of the details field moves the divider down.

### Filter controls

I removed the text where possible. I'm not sold on how it looks - I think it needs further refinement.

image

The filter and sort buttons I think are improved. The view buttons are ok. The select all/none buttons I think don't really look that good. I'm wondering if there is a better way to present those buttons. I don't know why the tooltip is so damn large.

# Scene performer thumbnails

Probably my favourite of the new things:

image

Fairly self explanatory. 

Suggestions and feedback welcomed.

### DIFF
--- a/ui/v2/src/index.scss
+++ b/ui/v2/src/index.scss
@@ -85,6 +85,8 @@ code {
   width: calc(100% + 40px);
   margin: -20px 0 0 -20px;
   position: relative;
+  height: 240px;
+  max-height: 80vh;
 }
 
 .grid-item label.card-select {
@@ -95,12 +97,24 @@ code {
   opacity: 0.5;
 }
 
+.video-container {
+  width: 100%;
+  height: 100%;
+}
+
 video.preview {
   // height: 225px; // slows down the page
   width: 100%;
   // width: calc(100% + 40px);
   // margin: -20px 0 0 -20px;
   object-fit: cover;
+  margin: 0 auto;
+  display: block;
+}
+
+video.preview.portrait {
+  height: 100%;
+  width: auto;
 }
 
 .filter-item, .operation-item {
--- a/ui/v2/src/index.scss
+++ b/ui/v2/src/index.scss
@@ -85,8 +85,11 @@ code {
   width: calc(100% + 40px);
   margin: -20px 0 0 -20px;
   position: relative;
+  max-height: 240px;
+}
+
+.previewable.portrait {
   height: 240px;
-  max-height: 80vh;
 }
 
 .grid-item label.card-select {
--- a/ui/v2/src/index.scss
+++ b/ui/v2/src/index.scss
@@ -161,6 +161,23 @@ video.preview {
   text-align: center;
 }
 
+.scene-specs-overlay {
+  display: block;
+  position: absolute;
+  bottom: 1em;
+  right: .7em;
+  font-weight: 400;
+  color: #f5f8fa;
+  letter-spacing: -.03em;
+  text-shadow: 0 0 3px #000;
+}
+
+.overlay-resolution {
+  font-weight: 900;
+  text-transform: uppercase;
+  margin-right:.3em;
+}
+
 #jwplayer-container {
   margin: 10px auto;
   width: 75%;
--- a/ui/v2/src/index.scss
+++ b/ui/v2/src/index.scss
@@ -254,6 +254,23 @@ span.block {
   background-repeat: no-repeat !important;
 }
 
+.performer-tag-container {
+  max-width: 100%;
+  height: 100%;
+  display: inline-block;
+  text-align: center;
+  margin: 5px;
+}
+
+.performer-tag.image {
+  max-height: 150px;
+  max-width: 100%;
+  background-size: cover !important;
+  background-position: center !important;
+  background-repeat: no-repeat !important;
+  margin: 0 auto;
+}
+
 .studio.image {
   height: 100px;
   background-size: contain !important;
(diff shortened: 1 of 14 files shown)


WithoutPants:
### Studio overlay

Added studio logo to the upper right corner of the scene card. The image is top-right aligned. The empty space in the right most thumbnail is due to empty space in the studio image.

image

WithoutPants:
### Fade out scene overlays on hover

I can't show a screenshot of this as I don't have the required tools, but hovering over the scene card should fade out the overlays, and fade back in nicely when moving off the scene card.

Leopere:
This is exciting work!

StashAppDev:
Just checked out this branch.  The portrait display commit definitely needs work.  The fixed height of 240px is problematic.  The overlay shows up under the video rather than on top of it for a majority of my library.

I'm not sure why but it looks like the studio overlay isn't working either.  The opacity is always 1%, basically invisible.

WithoutPants:
Latest commit should fix the overlay positioning. I want to do some further work to try to align the position of the titles, as different video sizes make the titles in different vertical positions. I'm not able to reproduce the opacity issue so far.

echo6ix:
@WithoutPants, this looks good. I'm looking forward to trying it in an alpha_dev version.

I hit a wall on my end with my "compact grid view" mode, which took the scene card improvements a step further and displayed the string Studio name (using .bp3-text-overflow-ellipsis) on the thumbnail where you have the studio image. I personally think it's just a cleaner look since a lot of the studio images are meh, if they exist at all.

Would you consider adding the above as a [switch](https://blueprintjs.com/docs/#core/components/switch) option in settings. Where the switch is located at `Settings > Interface Config > Scene Cards > Display studio using logo {on/off toggle switch}`

ui/v2/src/index.scss
```diff
@@ -38,7 +38,7 @@ code {
   flex-flow: row wrap;
   justify-content: center;
   margin: $pt-grid-size $pt-grid-size 0 0;
-  padding: 0 calc(10%);
```
StashAppDev: Can we make this and line 71 an interface option?

StashAppDev:
Here's a diff which fixes the studio opacity issue, removes what seems to be unnecessary CSS, and fixes an issue I'm seeing where the performer images aren't showing:

```
diff --git a/ui/v2/src/index.scss b/ui/v2/src/index.scss
index b1ece78..b23d153 100755
--- a/ui/v2/src/index.scss
+++ b/ui/v2/src/index.scss
@@ -199,7 +199,7 @@ video.preview.portrait {
   font-weight: 400;
   width: 40%;
   height: 20%;
-  opacity: 75%;
+  opacity: .75;
   z-index: 9;
 }
 
@@ -294,19 +294,15 @@ span.block {
 }
 
 .performer-tag-container {
-  max-width: 100%;
-  height: 100%;
-  display: inline-block;
-  text-align: center;
   margin: 5px;
 }
 
 .performer-tag.image {
-  max-height: 150px;
-  max-width: 100%;
-  background-size: cover !important;
-  background-position: center !important;
-  background-repeat: no-repeat !important;
+  height: 150px;
+  width: 100%;
+  background-size: cover;
+  background-position: center;
+  background-repeat: no-repeat;
   margin: 0 auto;
 }
```

StashAppDev:
group - Copy

Another issue.  The button selection state is inverted.  It looks like it's due to the minimal class.

ui/v2/src/index.scss
WithoutPants: For line 71, I've added a zoom slider to adjust the sizes of the scene cards to one of four different sizes. These sizes can be overridden with the custom css if wanted.

For line 41, I think we should just make a reasonable default and allow it to be overridden with custom css.

WithoutPants:
Thanks @StashAppDev . I've fixed the opacity and performer issues, and I've removed the `minimal` class from the view buttons. I've also fixed the overlay so that it shows the studio when the logo is hovered. I've also added a zoom slider to allow selecting the size of the scene cards.

Things left to do:
- [ ] add option to show studio by text. In future, we may want this to be a studio-specific option, but that will require a schema change.
- [ ] possibly condense view buttons further by changing to drop-down

WithoutPants:
Updated filter bar:

image


Moved the select all/none buttons into a More button. Seemed a better fit in a separate more button, and I think it ends up looking cleaner.

I investigated changing the view buttons into a drop-down list, but the result doesn't save any space and adds an extra click to change views. The intent is a little more clear on the drop-down list, but I think ultimately having three buttons is better. Below is a comparison of the two options:

image

WithoutPants:
Also added option to display studios as text instead of logos, per @echo6ix 's request. The result of setting this option is shown below. The text highlights blue on mouseover.

image

WithoutPants:
For now I'm satisfied with these improvements, but I realise that UI design is highly opinionated. I hope I've addressed all of the concerns raised.

bnkai:
I've tested a little all of the above and they seem ok now.I especially like the slider for the screenshot size and the performer thumbnails.
A little thing i noticed is the overlay on the portrait mode pics. 
overlay

PR #236, opened 2019-11-29
Add scene metadata scraping functionality

WithoutPants:
Add similar functionality to the performer scraper. 

This PR changes the schema of the scraper json configuration files. `type` - previously a single string value - becomes `types`, an array of strings. This way a single scraping configuration can apply to performers and scenes.

Also adds `get_scene_url` and `get_scene`. The former works similar to the `get_performer_url`, accepting a url and outputting the scene details. The latter accepts a full scene object and returns a scene object - meaning that the script can use whatever it needs from the scene object to get the data.

On the UI side, this change adds a scrape button to the URL field when a scrapable URL is detected. It also adds a "Scrape with..." menu button. This menu will be populated with scene scrapers for which `get_scene` is configured.

When a scene is successfully scraped, it populates the unpopulated scene fields with the scraped data. Stash will also attempt to match studio, performers and tags by name against those already in the system, and if present sets those in the resulting data. A future feature I'd like to add is to show unmatched performer, studio, tag data and allow the user to create the missing objects as needed.

### DIFF
--- a/graphql/documents/data/scrapers.graphql
+++ b/graphql/documents/data/scrapers.graphql
@@ -1,0 +1,75 @@
+fragment ScrapedPerformerData on ScrapedPerformer {
+  name
+  url
+  birthdate
+  ethnicity
+  country
+  eye_color
+  height
+  measurements
+  fake_tits
+  career_length
+  tattoos
+  piercings
+  aliases
+}
+
+fragment ScrapedScenePerformerData on ScrapedScenePerformer {
+  id
+  name
+  url
+  twitter
+  instagram
+  birthdate
+  ethnicity
+  country
+  eye_color
+  height
+  measurements
+  fake_tits
+  career_length
+  tattoos
+  piercings
+  aliases
+}
+
+fragment ScrapedSceneStudioData on ScrapedSceneStudio {
+  id
+  name
+  url
+}
+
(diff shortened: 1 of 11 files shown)


bnkai:
It seems ok and freeones seems to have no sideeffects ...
Your previous IAFD poc scraper needs to be changed also.
Is it ready or WIP till we have some scrapers?

WithoutPants:
Marking WIP while I ponder the configuration schema.

WithoutPants:
Changed to accommodate the new yaml scraper configuration format. An example Pornhub scraper configuration is shown below:

```
name: Pornhub
performerByURL:
  - action: script
    url: 
      - pornhub.com
    script:
      - python
      - phScrape.py
      - scrapePerformerURL
sceneByURL:
  - action: script
    url: 
      - pornhub.com/view_video.php
    script:
      - python
      - phScrape.py
      - scrapeSceneURL
sceneByFragment:
  action: script
  script:
    - python
    - phScrape.py
    - scrapeSceneURL
```

I have also included an example pornhub scraper script. This should be considered a proof of concept and not a production-ready script. I only tested it to ensure the fields were populated and performers and tags included.

[pornhub.zip](https://github.com/stashapp/stash/files/3958163/pornhub.zip)

Example scrapers are not within the scope of this PR.

PR #237, opened 2019-11-29
Fix marker page error

WithoutPants:
Fix error when going directly to marker wall page.


bnkai:
This one seems to fix it for me .

PR #241, opened 2019-12-03
Fix video.js current time

WithoutPants:
The `getPosition` method on the transcoding video player is currently broken as it doesn't add the start time to the current time. This meant that if a position was seeked to, then the scrubber wouldn't show the correct position, and the marker create dialog would get the wrong position value.

Fixed so that `seek` and `getPosition` are initialised after `currentTime` is set.


bnkai:
Everything seems ok . Seeking working ok . Markers get the correct time when created.

PR #242, opened 2019-12-03
Marker time input

WithoutPants:
Changes the marker seconds input to be a time input. Also adds a button to the input to select the current time of the playing scene.

image


bnkai:
Created marker ok.
Checked time button to get current time also ok.Marker from this also ok.
All markers created look fine too ...

Leopere:
100th pull request

PR #243, opened 2019-12-04
[WIP] Responsive UI changes

WithoutPants:
This is an extension of the #232 branch. I've looked at doing a larger overhaul to try to make the stash UI more small-screen-friendly. 

Specific changes here are to convert the filter bar into a navbar with left and right aligned elements, add a small-device specific menu of the top navbar items, change first/last/previous/next pagination buttons to use icons instead, and use a smaller number of pages on smaller devices.

This may possibly replace #232 if we're happy with the additions, or I can attempt to split everything up.

### DIFF
--- a/ui/v2/src/App.tsx
+++ b/ui/v2/src/App.tsx
@@ -16,11 +16,7 @@ import { Sidebar } from "./components/Sidebar";
 interface IProps {}
 
 export const App: FunctionComponent<IProps> = (props: IProps) => {
-  const [menuOpen, setMenuOpen] = useState<boolean>(getInitialMenuState());
-
-  function getInitialMenuState() {
-    return window.innerWidth > 768;
-  }
+  const [menuOpen, setMenuOpen] = useState<boolean>(false);
 
   function getSidebarClosedClass() {
     if (!menuOpen) {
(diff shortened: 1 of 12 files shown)


WithoutPants:
Screenshots. This is what the scene filter and navbar looks like on a larger device:

image

Note that the select all/none, zoom slider and view button have been moved to the right, with the rest moved to the left - rather than all in the centre. On larger screens, the left contents are to the left of the first scene card. I don't know whether this bothers people:

image

This is what it looks like on a phone:

image

The overflow of the filter controls is not ideal, and nor is the truncating of the page size, but not a dealbreaker and is something I can look at later. Notice that the navbar page links have been replaced with a menu button. Here is what the menu looks like when opened:

image

This is what the pagination looks like on a small device:

image

Notice the chevrons instead of First/Last/Previous/Next. These are on the larger version as well, only the number of pages displayed changes.

There are plenty of pages that still don't work well on a small device. These include:
- list views, including the tags page
- performer page seems to lose the settings button
- settings page

Leopere:
It looks like #232 progressed faster than this and got merged.  A responsive UI would be an incredible value add, in my opinion.

WithoutPants:
I'm going to close this for now, in preference for breaking it up into smaller, separate PRs.

PR #246, opened 2019-12-05
Allow pasting image into performer/studio

WithoutPants:
Small quality of life change. Set a studio or performer image by copying it, then pressing Ctrl+V in the edit performer/studio page, and it will be populated with the image from the clipboard.


bnkai:
Small change but a lof of difference. Definitely makes the manual perfmormer , studio additions a lot easier / quicker. Works ok for me on linux.

Leopere:
The only thing I wonder about is how obvious the ctrl+v function is in the ui then ux would improve greatly if it was obvious somewhere.

PR #251, opened 2019-12-06
Use inputgroup instead of editabletext in performer/studio edit pages

### LINKED ISSUE #247: Performer & Studio editing fields lose focus when switching apps or tabs
Hi all

This has already been discussed on the Discord, but I wanted to make an issue for it.

When editing/adding a performer/studio and filling in the details, if you tab out and tab back in it loses focus, so you have to left-click on the field again, which is a bit of a nuisance particularly when you're adding a performer and filling out date of birth etc.

It doesn't seem to lose focus on the main scene editing section.

Would appreciate if someone could implement this fix asap because it slows down adding performers a fair bit

Cheers!

EDIT: Mentioned studio editing as well (as per WithoutPants post)

WithoutPants:
Changes the performer and studio edit pages to use input group instead of editableText. Editable text has an issue where clicking on a different window then returning loses focus on the field. I also noticed that it resized the fields when clicking onto or away from the fields. This change shows a bare value when not in edit mode, and the input group when in edit mode. This means that the placeholder values won't be visible in read mode when there is no value for a given field.

I have left the name fields as editable text, since these are heading values and don't look as nice as input group fields. I can look at styling them larger if needed, but for now I think it should be ok.

This PR also includes a fix that hides the scrape URL button when in read mode.

Fixes #247

### DIFF
--- a/ui/v2/src/components/Studios/StudioDetails/Studio.tsx
+++ b/ui/v2/src/components/Studios/StudioDetails/Studio.tsx
@@ -183,7 +183,7 @@ export const Studio: FunctionComponent<IProps> = (props: IProps) => {
 
           <HTMLTable style={{width: "100%"}}>
             <tbody>
-              {TableUtils.renderEditableTextTableRow({title: "URL", value: url, isEditing, onChange: setUrl})}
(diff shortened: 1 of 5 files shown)


bnkai:
Works great.Imho the name fiels is fine as you left it

PR #252, opened 2019-12-06
Beautify scene list table

echo6ix:
Added modifier props to HTMLTable component.


bnkai:
Nice.Looks ok to me.

WithoutPants:
I added a commit to apply this change to the performers list as well.

PR #253, opened 2019-12-07
Add exclude file from scan feature

bnkai:
Added a basic regexp pattern matching file exclusion.
You can add the patterns in the config file or directly from the UI

For the config file 
```
exclude: 
- "sample\\.mp4$"
- "/\\.[[:word:]]+/"
- "c:\\\\stash\\\\videos\\\\exclude"
- "^/stash/videos/exclude/"
- "exclude/"
```
the first excludes all files ending in `sample.mp4`
the second hidden directories `/.directory/`
the third should work for windows based directories (the \ needs double escaping)
 and so on

If added through the UI you don't need to escape the `\` character 

exl

### DIFF
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -30,7 +30,7 @@ input ConfigGeneralInput {
   logLevel: String!
   """Whether to log http access"""
   logAccess: Boolean!
-  """Array of file regexp to exlude from scan"""
+  """Array of file regexp to exclude from Scan"""
   excludes: [String!]
 }
 
@@ -57,7 +57,7 @@ type ConfigGeneralResult {
   logLevel: String!
   """Whether to log http access"""
   logAccess: Boolean!
-  """Array of file regexp to exlude from scan"""
+  """Array of file regexp to exclude from Scan"""
   excludes: [String!]!
 }
 
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -30,6 +30,8 @@ input ConfigGeneralInput {
   logLevel: String!
   """Whether to log http access"""
   logAccess: Boolean!
+  """Array of file regexp to exlude from scan"""
+  excludes: [String!]
 }
 
 type ConfigGeneralResult {
@@ -55,6 +57,8 @@ type ConfigGeneralResult {
   logLevel: String!
   """Whether to log http access"""
   logAccess: Boolean!
+  """Array of file regexp to exlude from scan"""
+  excludes: [String!]!
 }
 
 input ConfigInterfaceInput {
--- a/pkg/manager/manager_tasks.go
+++ b/pkg/manager/manager_tasks.go
@@ -2,7 +2,9 @@ package manager
 
 import (
 	"path/filepath"
+	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -78,6 +80,7 @@ func (s *singleton) Scan(nameFromMetadata bool) {
 			return
 		}
 
+		results, _ = excludeFiles(results, config.GetExcludes())
 		total := len(results)
 		logger.Infof("Starting scan of %d files. %d New files found", total, s.neededScan(results))
 
@@ -491,4 +494,47 @@ func (s *singleton) neededGenerate(scenes []*models.Scene, sprites, previews, ma
 		}
 	}
 	return &totals
+}
+
+func excludeFiles(files []string, patterns []string) ([]string, int) {
+	if patterns == nil {
+		logger.Infof("No excludes in config.")
+		return files, 0
+	} else {
+		var results []string
+		var exclCount int
+		var fileRegexps []*regexp.Regexp
+
+		for _, pattern := range patterns {
+			reg, err := regexp.Compile(strings.ToLower(pattern))
+			if err != nil {
+				logger.Errorf("Aborting Exclude:%v", err)
+				return files, 0
+			}
+			fileRegexps = append(fileRegexps, reg)
+		}
+
+		if len(fileRegexps) == 0 {
+			return files, 0
+		}
+
+		for i := 0; i < len(files); i++ {
+			match := false
+			for _, regPattern := range fileRegexps {
+				//if pattern matches remove file from list
+				if regPattern.Match([]byte(strings.ToLower(files[i]))) {
+					logger.Infof("File %s excluded from scan ", files[i])
+					match = true
+					exclCount++
+					break
+				}
+
+			}
+			if !match {
+				results = append(results, files[i])
+			}
+		}
+		logger.Infof("Excluded %d file(s) from scan ", exclCount)
+		return results, exclCount
+	}
 }
--- a/pkg/manager/manager_tasks.go
+++ b/pkg/manager/manager_tasks.go
@@ -508,10 +508,10 @@ func excludeFiles(files []string, patterns []string) ([]string, int) {
 		for _, pattern := range patterns {
 			reg, err := regexp.Compile(strings.ToLower(pattern))
 			if err != nil {
-				logger.Errorf("Aborting Exclude:%v", err)
-				return files, 0
+				logger.Errorf("Exclude :%v", err)
+			} else {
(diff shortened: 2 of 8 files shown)


Leopere:
will the exclude file follow gitignore styling or some kind of YAML configuration in some way?

bnkai:
It is in the config file as en extra option `exclude` so it's either json or yml depending on what your config file is currently.

pkg/manager/manager_tasks.go
```diff
@@ -492,3 +495,46 @@ func (s *singleton) neededGenerate(scenes []*models.Scene, sprites, previews, ma
...
+		var results []string
+		var exclCount int
+		var fileRegexps []*regexp.Regexp
+
+		for _, pattern := range patterns {
+			reg, err := regexp.Compile(strings.ToLower(pattern))
+			if err != nil {
+				logger.Errorf("Aborting Exclude:%v", err)
+				return files, 0
+			}
```
WithoutPants: As a user, if I made a mistake with one of the regex patterns, I would still expect the scan to exclude using the patterns that were valid. I think not applying any exclude patterns due to a single invalid pattern is unintuitive behaviour.

graphql/schema/types/config.graphql
```diff
@@ -55,6 +57,8 @@ type ConfigGeneralResult {
   logLevel: String!
   """Whether to log http access"""
   logAccess: Boolean!
+  """Array of file regexp to exlude from scan"""
```
WithoutPants: Typo

pkg/manager/manager_tasks_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestExcludeFiles(t *testing.T) {
+
+	filenames := []string{
+		"/stash/videos/filename.mp4",
```
WithoutPants: For completeness, can we add windows network paths as well? `\\\\network-share\\path\\etc`

ui/v2/src/components/Settings/SettingsConfigurationPanel.tsx
```diff
@@ -65,13 +68,36 @@ export const SettingsConfigurationPanel: FunctionComponent<IProps> = (props: IPr
...
+      setExcludes(conf.general.excludes);
     }
   }, [data]);
 
   function onStashesChanged(directories: string[]) {
     setStashes(directories);
   }
 
+  function excludeRegexChanged(idx: number,value: string) {
+	const newExcludes = excludes.map((regex,i)=> {
```
WithoutPants: We use spaces for tabs with a tab-size of 2 for the UI files. Tabs appear throughout the new code. Also for consistent formatting, add a space after `,` characters.

pkg/manager/manager_tasks.go
bnkai: Yeah , it was my first step after the must compile change ...
I'll print an error warning and let it continue

pkg/manager/manager_tasks_test.go
bnkai: Added a few.
A lot of `\` escapes needed ....

ui/v2/src/components/Settings/SettingsConfigurationPanel.tsx
bnkai: Noted.
Should be ok now.

Leopere:
One last QA check before we merge this.

pkg/manager/manager_tasks_test.go
WithoutPants: Windows network paths are with double backslash `\\network-share\etc`.
```
\\\\\\\\network\\\\videos\\\\filename  windows network.mp4
```
translates to 
```
\\\\network\\videos\\filename  windows network.mp4
```
It should be similar to the `c:\\` ones - ie:
```
\\\\network\\videos\\filename  windows network.mp4
```
which would translate to 
```
\\network\videos\filename  windows network.mp4
```
which is the correct form.

ui/v2/src/components/Settings/SettingsConfigurationPanel.tsx
```diff
@@ -65,13 +68,36 @@ export const SettingsConfigurationPanel: FunctionComponent<IProps> = (props: IPr
...
   function onStashesChanged(directories: string[]) {
     setStashes(directories);
   }
 
+  function excludeRegexChanged(idx: number, value: string) {
+    const newExcludes = excludes.map((regex, i)=> {
+      const ret = ( idx !== i ) ? regex : value ;
+      return ret
+      })
+  setExcludes(newExcludes);
```
WithoutPants: Indenting still broken here.

ui/v2/src/components/Settings/SettingsConfigurationPanel.tsx
```diff
@@ -65,13 +68,36 @@ export const SettingsConfigurationPanel: FunctionComponent<IProps> = (props: IPr
...
+  }
+
+  function excludeRemoveRegex(idx: number){
+  const newExcludes = excludes.filter((regex, i) => i!== idx );
+
+  setExcludes(newExcludes);
```
WithoutPants: Indenting still broken here.

ui/v2/src/components/Settings/SettingsConfigurationPanel.tsx
```diff
@@ -65,13 +68,36 @@ export const SettingsConfigurationPanel: FunctionComponent<IProps> = (props: IPr
...
+  }
+
+  function excludeAddRegex(){
+  const demo = "sample\\.mp4$"
+  const newExcludes = excludes.concat(demo);
+
+  setExcludes(newExcludes);
+  }
```
WithoutPants: Indenting still broken here, and please add space between `)` and `{`.

ui/v2/src/components/Settings/SettingsConfigurationPanel.tsx
```diff
@@ -148,6 +174,26 @@ export const SettingsConfigurationPanel: FunctionComponent<IProps> = (props: IPr
...
+
+       { (excludes) ? excludes.map((regexp, i) => {
+                                            return(
+                                              <InputGroup
+                                              value={regexp}
+                                              onChange={(e: any) => excludeRegexChanged(i, e.target.value)}
+                                              rightElement={<Button icon="minus" intent="danger" onClick={(e: any) => excludeRemoveRegex(i)} />}
+                                              />
+                                              );
+                                            })   : null
```
WithoutPants: This should be indented by 2 spaces from line 184. Remove extra two spaces between `)` and `: null`.

ui/v2/src/components/Settings/SettingsConfigurationPanel.tsx
```diff
@@ -148,6 +174,26 @@ export const SettingsConfigurationPanel: FunctionComponent<IProps> = (props: IPr
...
+       }
+
+          <Button icon="plus" onClick={(e: any) => excludeAddRegex()} />
```
WithoutPants: I think adding `minimal={true}` to the add and remove buttons looks a bit nicer. What do you think?

WithoutPants:
From testing, I think the exclusions should also apply when performing a clean. Currently, it does not apply these.

bnkai:
> From testing, I think the exclusions should also apply when performing a clean. Currently, it does not apply these.

it's supposed to be only for the Scan, i'll create a separate PR with an option for the Clean

pkg/manager/manager_tasks_test.go
bnkai: i mistook your `\\\\network-share\\path\\etc` as the non escaped path instead of already escaped , now fixed

ui/v2/src/components/Settings/SettingsConfigurationPanel.tsx
bnkai: ... you are probably right
normal
normal
minimal
minimal
minimal2

i'll set it to minimal for both

bnkai:
I think i changed everything needed , not 100% sure on the tsx formatting i did it manually.

PR #255, opened 2019-12-09
Add Play random button to scenes and scene markers page

### LINKED ISSUE #249: Add "Play Random" functionality with filter controls.
In the scenes section a play random feature would be a great addition.  It could also be applicable in other areas such as stars and markers?  

Play random would also take into account your current filters, so if you wanted to play random scenes that were only of a rating of 5 then it would only take those into account.

WithoutPants:
Adds a Play Random item to the More button on scenes and scene markers. Chooses a random item using the current filter and navigates to the scene/marker and auto-plays it. 

Fixes #249

### DIFF
--- a/ui/v2/src/components/list/ListFilter.tsx
+++ b/ui/v2/src/components/list/ListFilter.tsx
@@ -1,8 +1,6 @@
 import {
-  AnchorButton,
   Button,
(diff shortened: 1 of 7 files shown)


bnkai:
Everything checks out ok for me.

PR #257, opened 2019-12-11
Make top level navbar responsive

WithoutPants:
Changes the top-level navbar to be a button when the navbar won't fit on the screen. On a larger screen, navbar is unchanged. On smaller screens, navbar is replaced as follows:

image

Menu only opens when the button is pressed, and overlays the content pane.


bnkai:
Tried resizing , tried from smartphone , all seems ok.

PR #259, opened 2019-12-12
Make scene metadata from file metadata optional

WithoutPants:
This change extends #136 to make all of the file-based metadata extraction optional, since it is frequently incorrect. It makes all of the behaviour introduced in #103 optional, with the exception of the title defaulting to the filename. The checkbox in the settings for this has been renamed to "Set name, date, details from metadata (if present)" and is now unchecked by default.


bnkai:
Works ok for me.
I think if we exclude a few people that tag their media files the majority of files have comments and titles that are either irrelevant or not well defined anyway.

Issue #35, opened 2019-03-24
Custom scene cover


StashAppDev:
As discussed in discord:

> maybe 20% by default, but having the option in the UI to upload an image as well. what ive been doing is pulling the "cover art" image from  the torrent tracker site, and using that. useually is a image that came from a paysite gallery for the scene thats been perfectly croped and had sites logo added or somthing of the kind. looks better imho than just scrolling though random screenshots ( plus they tend to all look the same about 20% in on the same catagory of scene)

It would be nice to use a custom cover for a scene (like the official site image).  Right now Stash generates screenshot files at 20% into the video.  For this change this would act as a default, but we would also expand the scene model.  There would need to be a new BLOB cover image property in the database.  Stash should use this image if it exists, otherwise fall back to the screenshot.

echo6ix:
I look forward to this feature. Also consider generating a custom scene cover/thumbnail from the a user defined time during the scene. Similar to how a marker is currently generated.

rexii2300:
yep this would be fantastic!

WithoutPants:
> I look forward to this feature. Also consider generating a custom scene cover/thumbnail from the a user defined time during the scene. Similar to how a marker is currently generated.

This is covered by #9 

> It would be nice to use a custom cover for a scene (like the official site image). Right now Stash generates screenshot files at 20% into the video. For this change this would act as a default, but we would also expand the scene model. There would need to be a new BLOB cover image property in the database. Stash should use this image if it exists, otherwise fall back to the screenshot.

Is there any reason we can't just replace the existing screenshot file with the submitted one? I'm not sure why we need to store a separate image in the database unless we're using the thumbnail image for something else.

ueaslsef:
@Leopere this seems done, should we close it?

bnkai:
should be closed when #376 is merged

bnkai:
closing since #376 was merged

PR #262, opened 2019-12-13
Allow uploading of custom scene covers

WithoutPants:
Adds a "Cover Image" section to the Edit Scene page. This is an expandable section that is collapsed by default, as follows:

image

Expanding this shows the current cover image, with a file selector to select a new cover image:

image

Like the other image selection pages, Ctrl-V with an image in the clipboard will paste the image directly.

Note that saving the scene after updating the cover image does not update the cover image in the video player until the page is reloaded.

Should address #35


bnkai:
Looks ok.
Tested the addition through file browse, paste.
Works ok in the scenes grid,list and the player.
It replaces the original screenshot, thumbnail so no problem with the clean task also.

PR #263, opened 2019-12-13
Fix vtt for chapter display in scene players

WithoutPants:
Fixes the VTT generation so that markers are shown correctly in the seek bar in jwplayer, and in the chapters drop-down in videojs. Note that videojs only shows the chapters drop-down if there are more than one marker in the scene.


Leopere:
Looks like the build process is just getting stuck on the upload to transfer.sh so if someone can build this locally and verify the changes I can merge.

PR #268, opened 2019-12-16
jwplayer transcode seek support. Remove video.js

WithoutPants:
This change means that all video files are shown only in jwplayer, instead of showing video.js for non-streamable files. This change includes a local change to the jwplayer code, which is maintained in the `hooks` branch of my personal fork of the jwplayer project, available here: https://github.com/WithoutPants/jwplayer/tree/hooks

This version of jwplayer allows hooking into the `getDuration`, `seek` and `getPosition` methods to override them. The stash code was changed to make use of these hooks in the same way that it hooked into the same methods in video.js.

Since video.js is no longer used or required, it was removed.


bnkai:
Tests fine for me.

PR #269, opened 2019-12-16
Add stash scraper type

WithoutPants:
Adds stash as a scraper type. This allows one to scrape performer and scene metadata from another remote stash instance.

Here is an example yaml config for a stash scraper:
```
name: stash
performerByName:
  action: stash
performerByFragment:
  action: stash
sceneByFragment:
  action: stash
stashServer:
  url: http://username:password@stashaddress:9999
```

Allows searching performer by name and scene by fragment. `sceneByFragment` is implemented by searching by scene checksum.


bnkai:
Performers seem to work ok.

For the scenes i found the following from a quick test
- details and date seem ok
- title,rating,url are not updated even though the scene was found and had field data
- tag,studio,performers are getting the scraper's id ok but if the client has different name already for the same id  we end up with  wrong tag,studio,performer
- if you scrap a scene and a result is found then on save you get `Error: GraphQL error: FOREIGN KEY constraint failed` even if nothing seems changed in the UI. 
  EDIT this seems to be related to the above ,it happens if the scraper has tags , or studios or performers or anything other with an ID that doesn't exist in the client

WithoutPants:
Note that the current implementation does not overwrite fields if they are already populated.

bnkai:
> Note that the current implementation does not overwrite fields if they are already populated.

that might explain the title .... but the url and rating were definetely blank on the test i did..
also the title by default gets the filename from the scan , so it will always be populated ...

WithoutPants:
URL should now be populating correctly. Rating was left off intentionally as I didn't think it should be included by default. In future, I think we should show the difference between the current and scraped values and allow the user to set accordingly.

The IDs of performers, studio and tags should never have been returned by the scrape, as they will seldom match between instances, and the local stash sets the IDs explicitly after matching them. This has now been fixed.

bnkai:
URL is ok now
Performers,studio and tag seem ok too.

A small bug left...
I get this error
 `Error: GraphQL error: Internal system error. Error <runtime error: invalid memory address or nil pointer dereference>` 
when the scene is not found and this is the [dump](https://pastebin.com/xBWhNWLN).

Leopere:
Can we have the system programmatically generate a clean demo config so that the end-users always have the most accurate example in their installs?

bnkai:
Everyhing seems fine now ...

@Leopere the config for this PR is an extra file like this ~/.stash/scrapers/stash.yml , it's doesn't change anything for the main config file
If you want a clean config file or maybe better a config file sample we can just run stash , tick/fill out  all options and hit save i guess...
In due time i think making a config section in the wiki might  be more convenient

PR #271, opened 2019-12-17
Add cache breaker for screenshot using modtime

### LINKED ISSUE #270: [Bug Report] Session "caching" issue with cover photos with scenes
**Describe the bug**
ok, i choose the picture, it shows up in the edit screen
i hit "save" and it reverts
i hit refresh
still same "old" picture
(or cover)

**To Reproduce**
Open multiple sessions of chrome
1. Go to 'any scene'
2. Go to edit tab
3. Click on "> cover image"
4. Browse to the desired picture
5. Image should appear
6.  Click save under the edit tab
<picture will revert to the old picture>
REFRESH the page
the picture will remain as the "old" cover image

Other browser instances will have the updated image.  Once you close all instances of chrome and try again, I'm sure this problem would be resolved.

**Expected behavior**
I believe that the image should update regardless of the number of instances of the browser you have open.

 - OS: Windows
 - Browser [Chrome]
 - Version: 78.0.3904.108 (Official Build) (64-bit)

**Additional context**
briefly discussed on the general chat within discord

WithoutPants:
Fixes #270 

Adds a cache-breaker query parameter to the generated screenshot url based on the modification time of the scene. When the scene is updated (and the cover image possibly changed) then the query parameter value will change, ensuring that the latest copy of the cover image is downloaded.

### DIFF
--- a/pkg/api/resolver_model_scene.go
+++ b/pkg/api/resolver_model_scene.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+
 	"github.com/stashapp/stash/pkg/api/urlbuilders"
 	"github.com/stashapp/stash/pkg/manager"
 	"github.com/stashapp/stash/pkg/models"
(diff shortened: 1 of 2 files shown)


bnkai:
It works ok now even on chrome without refreshing.

PR #274, opened 2019-12-18
Added exclude patterns support for Clean Task

bnkai:
Extend the code from #253 so the files matched from those patterns are also deleted from the DB during the Clean Task


pkg/manager/task_clean.go
```diff
@@ -73,3 +77,36 @@ func (t *CleanTask) pathInStash() bool {
 	logger.Debugf("File %s is out from stash path", t.Scene.Path)
 	return false
 }
+
+func includeFile(file string, patterns []string) bool {
```
WithoutPants: Suggest moving this method, and `excludeFiles` from manager_tasks.go into a new `exclude.go` (or something like that) file so that the exclude list code is in the one place. It will also make it easier to refactor the common code in the two methods into helper methods.

pkg/manager/exclude_files.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil
+	} else {
+		return fileRegexps
+	}
+
+}
+
+func matchFileSimple(file string, regExps []*regexp.Regexp) bool {
+	for _, regPattern := range regExps {
+		if regPattern.Match([]byte(strings.ToLower(file))) {
```
WithoutPants: Use `MatchString` instead.

pkg/manager/exclude_files.go
```diff
@@ -0,0 +1,89 @@
...
+
+	} else {
+		fileRegexps := generateRegexps(patterns)
+
+		if len(fileRegexps) == 0 {
+			return false
+		}
+
+		for _, regPattern := range fileRegexps {
+			if regPattern.Match([]byte(strings.ToLower(file))) {
```
WithoutPants: Use `MatchString` instead. That way you don't need to cast to `[]byte`.

WithoutPants:
Tests ok. With those minor changes sorted this should be good to go.

Issue #124, opened 2019-09-13
Performer profile page UI enhancements


echo6ix:
Following the theme of my other [front-end UI concepts](https://github.com/stashapp/stash/issues/108#issuecomment-524760553) #108 , I think each performers page, `http://localhost:9999/performers/n`, could act as their profile, summarizing their content on Stash.

This concept utilizes #120 (performer tags feature enhancement request), #118 (side panel performer list feature enhancement request), and #123 (broadening the scope of galleries)

#### 1. The default profile page showing the performer's bio.
The attached image is self-explanatory. Some notes are:

![64833549-6a898780-d5ac-11e9-86ae-3abcf1e8d1a4](https://user-images.githubusercontent.com/37937507/91378703-185a6a00-e7ef-11ea-98a7-2ebd4d2127c5.png)

- The performer's "profile" image would be different from their main image that's currently displayed throughout Stash. Because the profile image would have smaller dimensions a user can focus on a cropped image of just the performers face.
- The performer can be favorited and unfavorited directly from the header at any time by clicking the heart icon (icon via [FontAwesome](https://fontawesome.com/icons/heart)).
- Birthdate is displayed in a user-friendly verbose format (`Month, D, YYYY`)

#### 2. Profile page showing the performer's images from galleries per #123 (galleries gridview and other enhancements)

![64834062-58a8e400-d5ae-11e9-9fd7-fa9c7800f298](https://user-images.githubusercontent.com/37937507/91378788-4cce2600-e7ef-11ea-840d-e6f5b141142e.png)

- Images displayed as cards, as scenes are. I wanted to add rounded corners, but I suck.
- Clicking an image shows it in "full-screen" mode like Lightbox, where the user can navigate to the prior or previous image, or can escape out of full-screen mode.

![64834187-dbca3a00-d5ae-11e9-8bd6-4d3bd9b19b96](https://user-images.githubusercontent.com/37937507/91378920-869f2c80-e7ef-11ea-82c6-465c69795fbb.png)

#### 3. Profile page showing the edit panel

![64834206-edabdd00-d5ae-11e9-931b-18e04cd6ea04](https://user-images.githubusercontent.com/37937507/91379001-aa627280-e7ef-11ea-8319-a9528a3a5501.png)

Yet again, it's self-explanatory in the image. The only real difference is I have showed the option
to upload a profile image.

##### Profile image
[...]

PR #280, opened 2019-12-20
Add scenes tab to performer page

WithoutPants:
This change adds a set of tabs to the performer page: Details, Edit, Scenes and Operations. 

Details and Edit tabs should look like the old page, with the fields only editable in the Edit tab. The Delete and Scrape buttons have been moved into this tab.

image

image

The Scenes tab shows a scenes list filtered on the performer. 

image

The Auto Tag button has been moved to Operations.

image

The only thing I'm really unhappy with for this change is that when the right panel scrolls, then the image scrolls up with it and there is a bunch of empty space on the left. I wanted to make the right panel independently scrollable, but I haven't been able to get the CSS to work correctly for it. It would also have to be compatible with smaller screens as well. To my mind, the

### DIFF
--- a/ui/v2/src/components/performers/PerformerDetails/Performer.tsx
+++ b/ui/v2/src/components/performers/PerformerDetails/Performer.tsx
@@ -16,6 +16,7 @@ import { PerformerDetailsPanel } from "./PerformerDetailsPanel";
 import { PerformerOperationsPanel } from "./PerformerOperationsPanel";
 import { PerformerScenesPanel } from "./PerformerScenesPanel";
 import { TextUtils } from "../../../utils/text";
+import Lightbox from "react-images";
 
 interface IPerformerProps extends IBaseProps {}
 
@@ -25,6 +26,7 @@ export const Performer: FunctionComponent<IPerformerProps> = (props: IPerformerP
   // Performer state
   const [performer, setPerformer] = useState<Partial<GQL.PerformerDataFragment>>({});
   const [imagePreview, setImagePreview] = useState<string | undefined>(undefined);
+  const [lightboxIsOpen, setLightboxIsOpen] = useState<boolean>(false);
 
   // Network state
   const [isLoading, setIsLoading] = useState(false);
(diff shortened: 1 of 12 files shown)


WithoutPants:
Related to #124 and #108

bnkai:
Everything seems to function ok.
A cosmetic suggestion , when you switch to the Scenes or Operations tabs it seems a little weird with the performer name missing. Maybe add the performer name there also or keep the performer name above the tabs element?
For the scrolling i get what you mean , i played a little with fixed positioning but it has its problems also...

WithoutPants:
Marked as WIP while I do another iteration on this to make it more like the mock ups in #124

WithoutPants:
Changed the layout. Favourite button added to heading, along with the URL link as an icon. Instagram and Twitter not added in as blueprint doesn't have the icons for these. Age and aliases now shown in the header. Tabs moved below the image. Image can be clicked to show a full-screen view.

image

URL and social links now displayed as actual links:

image

Edit screen has normal name and alias fields now, and favourite has been removed in favour of the button at the top:

image

I'm still not super-sold on all the empty space to the right of the image, but we could probably fill it with other details, like number of scenes, performer tags (if/when implemented).

The new performer page still looks like the old version.

bnkai:
It looks ok to me the only thing **if possible** is what i mentioned in discord `is there a way to split the Tabs so that the the titles go to the right of the performer image while the panel contents go underneath? `. Otherwise works/tests ok for me.

Cutters1982:
Looks brilliant! Thank you

Please can you confirm that if you select a field, tab out, tab back in, does it lose focus or keep it?

:)

WithoutPants:
Confirmed that tabbing out and back does not lose focus on fields.

Blueprint implements tab contents within the tabs object. In order to have the tab headers next to the performer image, the tabs would need to be in a different element than the tab contents. I could probably hack a way around it, but I don't think it's worth it right now. We can re-evaluate with the UI refresh.

Cutters1982:
thanks for confirming mate :)

PR #285, opened 2019-12-23
Xpath scraping from URL

WithoutPants:
Adds the ability to scrape performer and scene details by URL within stash itself, using an XPath scraping configuration.

XPath scraping requires the xpath scraping configuration to be included in the scraper yaml config file. XPath scrapers are configured under `xPathScrapers`. Each key inside this object corresponds to the name of the xpath scraper config.

The xpath behaviour is configured with `action: scrapeXPath`. Currently, this action is only valid for `performerByURL` and `sceneByURL` scrape types. When `action` is set to `scrapeXPath`, the `scraper` field must also be configured. The value of the `scraper` field must match the name of one of the xpath scrapers configured in `xPathScrapers`.

XPath scraper configurations have up to three possible keys within: `common`, `performer`, and `scene`. 

The `common` object stores common xpath fragments that are referenced in other places. I have used a convention of `$name` to denote these, but no particular naming convention is enforced.

The `performer` and `scene` objects contain the fields corresponding to the scraped performer and scene objects. Note that these fields are case sensitive. In addition to this, the scene configuration has `Tags`, `Performers` and `Studio` objects, which themselves have fields.

I think this is best demonstrated with an example, so below is an xpath scraper configuration for pornhub, that scrapes performers (pornstars and community members) and scenes:

```
name: Pornhub
performerByURL:
  - action: scrapeXPath
    url: 
      - pornhub.com
    scraper: performerScraper
sceneByURL:
  - action: scrapeXPath
    url: 
      - pornhub.com/view_video.php
    scraper: sceneScraper

xPathScrapers:
  performerScraper:
    common:
      $infoPiece: //div[@class="infoPiece"]/span
    performer:
      Name: //h1[@itemprop="name"]
      Birthdate: //span[@itemprop="birthDate"]
      Measurements: $infoPiece[text() = 'Measurements:']/../span[@class="smallInfo"]
      Height: $infoPiece[text() = 'Height:']/../span[@class="smallInfo"]
      Ethnicity: $infoPiece[text() = 'Ethnicity:']/../span[@class="smallInfo"]
      FakeTits: $infoPiece[text() = 'Fake Boobs:']/../span[@class="smallInfo"]
      Piercings: $infoPiece[text() = 'Piercings:']/../span[@class="smallInfo"]
      Tattoos: $infoPiece[text() = 'Tattoos:']/../span[@class="smallInfo"]
      CareerLength: $infoPiece[text() = 'Career Start and End:']/../span[@class="smallInfo"]
  sceneScraper:
    common:
      $performer: //div[@class="pornstarsWrapper"]/a[@data-mxptype="Pornstar"]
      $studio: //div[@data-type="channel"]/a
    scene:
      Title: //div[@id="main-container"]/@data-video-title
      Tags: 
        Name: //div[@class="categoriesWrapper"]//a[not(@class="add-btn-small ")]
      Performers:
        Name: $performer/@data-mxptext
        URL: $performer/@href
      Studio:
        Name: $studio
        URL: $studio/@href
```

Let me know if there is anything that needs to be clarified. 

Future direction for this will be to include post-processing of fields, such as using regex to manipulate the field values.

Going forward, it will be useful to maintain a repository of these scraper configurations to share between users.


Leopere:
I fully agree we can potentially determine a structure in which will keep it organized such that version compatibility is evident and clear to anyone.

bnkai:
It seems to work ok to me (data gets fetched, perfomers, studios added to scene if they exist with same name) , i think it will be easier to evolve this as more scrapers get added.

The only think i have to report is the birthdate from the  performer part of the scraper. It gets added as a string correctly into the input group  but when you hit save it's gone. An extra format `Jan 2, 2006` needs to be added to `/pkg/utils/date.go` to parse the pornhub date correctly. 
This will need to be done for any new scraper whose  date format is not yet supported (unless you want to somehow add it  to the scraper configuration?)

I assume you kept the former scrapers behaviour the same , if a field is not null it won't be changed? Also in the scenes scraper if a scene already has a performer (not the same with the one found from the scraper) the scraped one won't be added.


The url field (which i already saved/entered to use the scraper function) gets null as soon as you use the scrape button , is this expected?

WithoutPants:
Performer URL is fixed so that it only overwrites if the result URL is set.

I'd like to leave date parsing to a future iteration on this feature. My intention is to configure it in the scraper config. For now, it will have to be a manual post-scrape step for the user to do.

I have left the performer and tags fields as is - if they are set, then they are not added to. I think a future iteration should show the before and after fields (something like the filename parser) and allow the user to adjust accordingly.

bnkai:
Tests ok , good to go.

PR #288, opened 2019-12-25
Add link to gthub releases to about page

ueaslsef:
image
I would prefer to actually check the github page for newer releases and possibly even a button to update, but I don't know where and how to propperly implement this. If anyone want's to give me a hint, feel free :)

### DIFF
--- a/ui/v2/src/components/Settings/SettingsAboutPanel.tsx
+++ b/ui/v2/src/components/Settings/SettingsAboutPanel.tsx
@@ -52,6 +52,9 @@ export const SettingsAboutPanel: FunctionComponent<IProps> = (props: IProps) => 
       {!data || loading ? <Spinner size={Spinner.SIZE_LARGE} /> : undefined}
       {!!error ? <span>error.message</span> : undefined}
       {renderVersion()}
+      <p>
+        New updates might be available on <a href="https://github.com/stashapp/stash/releases">GitHub</a>.
+      </p>
     </>
   );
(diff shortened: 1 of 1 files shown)


Leopere:
There are examples for checking for newer builds via the github api this is definitely a thing that could be figured out.

bnkai:
This was certainly in my TODO list , if no one else has any plans for it.
I was thinking of checking first once in the console when stash starts and maybe a button to check for updates somewhere in the UI ?
As it is now there are only two  actual "versions"  v0.0.0-alpha and v0.0.0-alpha_dev (we haven't decided on numbering/versioning yet ) , the first for the "stable" and the second for the "dev".The only thing that changes is the build hash.I was thinking of getting the hash from the github tag api(there only the two tagged versions there).If the build hash in the github api for each of these versions is different  then there is a newer version....
[tag api](https://api.github.com/repos/stashapp/stash/tags)

bnkai:
@ueaslsef is your version a self compiled one?
The "version" tag gets added through travis , thats how it looks when you get it from the releases page


version

bnkai:
@ueaslsef i made a PR with the things mentioned above , have a look if it covers you. The update stuff is a bit more tricky though.

ueaslsef:
@bnkai Great, this obsoletes my sloppy text.

Leopere:
Thank you all the same for getting the motion started on this @ueaslsef

PR #289, opened 2019-12-25
Add autorefresh of task status to tasks page

ueaslsef:


Leopere:
It could be a reasonable value add, but I feel there has to be a better way than just to refresh the page on an interval breaking the user's interaction flow.

bnkai:
It's not exactly refresh page @Leopere  , it more of a pooling of the tasks status...
@ueaslsef was there something wrong , not optimal with the current implementation ?

WithoutPants:
This should not be necessary. The websocket subscription should be automatically updating when tasks start and finish.

Leopere:
> This should not be necessary. The websocket subscription should be automatically updating when tasks start and finish.

Perhaps the interface is failing in just indicating that there is progress being made.  Perhaps we should tweak the interface in some way to indicate that the app is doing something and actually progressing in some way.

ueaslsef:
The ui never updated by itself, so this was the quick and dirty fix. No, it does not reload the page, just poll the endpoint every 2.5 seconds until it is idle again.

WithoutPants:
> The ui never updated by itself, so this was the quick and dirty fix. No, it does not reload the page, just poll the endpoint every 2.5 seconds until it is idle again.

In that case, an issue should be raised detailing the bug. There should be no need to poll the endpoint when we already have a websocket connection and graphql subscription that handles task updates.

bnkai:
I don't seem to have the issue , the tasks seem to update the status bar ok.
Maybe close and open an issue as @WithoutPants mentioned to reproduce the "bug" if possible ?

Leopere:
Is this redundant now? 

Closing unless is not redundant then feel free to re-open.

ueaslsef:
Yes it is, sorry for beeing absent for so long.

PR #295, opened 2019-12-28
Fix usage of Box.Bytes causing depreciation message

FleetingOrchard:
Save the trees, use less logspace.

This is basically directly mocked off of how packr currently implements .Bytes (by passing the args to .Find and then ignoring the second part of the 2-tuple).

I'm also sneaking in a `release` target into the make file that effectively mimics what the project readme states should be done to make a release build. Don't tell anyone. It'll be our secret.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ ifeq ($(OS),Windows_NT)
   SET := set
 endif
 
+release: generate ui build
+
 build:
 	$(eval DATE := $(shell go run scripts/getDate.go))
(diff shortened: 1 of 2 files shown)


bnkai:
Simple and looks/works ok for me.

Leopere:
I'm terribly curious about screenshots for the before and after for those of us who don't speak Go

FleetingOrchard:
There isn't any sort of difference. Bytes is implemented, under the hood, as literally just a call to Find with the same args (and additionally throwing the depreciation message).

PR #296, opened 2019-12-29
Add check version functionality

bnkai:
Adds initial support for new version checking. Mentioned also in #288 
As it is now it uses the github api to get the shorthash commits from the repo tags.
It matches the current version ( `v0.0.0-alpha` or 'v0.0.0-alpha_dev' ) to the respective repo tags and gets the related commit hash from there.If it's different than the local that means that a new release build is available. 
It first prints a message in the console when the app is started and you can see it also when you open the about tab from the settings.An extra button is also available for manual checking.

for dev testing the build , since version is added through TRAVIS you need to add it to the ldflags in the Makefile
`-X 'github.com/stashapp/stash/pkg/api.version=v0.0.0-alpha_dev' `
or
`-X 'github.com/stashapp/stash/pkg/api.version=v0.0.0-alpha'`

console

ui

### DIFF
--- a/pkg/api/check_version.go
+++ b/pkg/api/check_version.go
@@ -1,0 +1,179 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/stashapp/stash/pkg/logger"
+	"io/ioutil"
+	"net/http"
+	"runtime"
+	"time"
+)
+
+//we use the github REST V3 API as no login is required
+const apiURL string = "https://api.github.com/repos/stashapp/stash/tags"
+const apiReleases string = "https://api.github.com/repos/stashapp/stash/releases"
+const apiAcceptHeader string = "application/vnd.github.v3+json"
+
+var stashReleases = func() map[string]string {
+	return map[string]string{
+		"windows/amd64": "stash-win.exe",
+		"linux/amd64":   "stash-linux",
+		"darwin/amd64":  "stash-osx",
+		"linux/arm":     "stash-pi",
+	}
+}
+
+type githubTagResponse struct {
+	Name        string
+	Zipball_url string
+	Tarball_url string
+	Commit      struct {
+		Sha string
+		Url string
+	}
+	Node_id string
+}
+
+type githubReleasesResponse struct {
+	Url              string
+	Assets_url       string
+	Upload_url       string
+	Html_url         string
+	Id               int64
+	Node_id          string
+	Tag_name         string
+	Target_commitish string
+	Name             string
+	Draft            bool
+	Author           githubAuthor
+	Prerelease       bool
+	Created_at       string
+	Published_at     string
+	Assets           []githubAsset
+	Tarball_url      string
+	Zipball_url      string
+	Body             string
+}
+
+type githubAuthor struct {
+	Login               string
+	Id                  int64
+	Node_id             string
+	Avatar_url          string
+	Gravatar_id         string
+	Url                 string
+	Html_url            string
+	Followers_url       string
+	Following_url       string
+	Gists_url           string
+	Starred_url         string
+	Subscriptions_url   string
+	Organizations_url   string
+	Repos_url           string
+	Events_url          string
+	Received_events_url string
+	Type                string
+	Site_admin          bool
+}
+
+type githubAsset struct {
+	Url                  string
+	Id                   int64
+	Node_id              string
+	Name                 string
+	Label                string
+	Uploader             githubAuthor
+	Content_type         string
+	State                string
+	Size                 int64
+	Download_count       int64
+	Created_at           string
+	Updated_at           string
+	Browser_download_url string
+}
+
+//gets latest version (git commit hash) from github API
+//the repo's tags are used to find the latest version
+//of the "master" or "develop" branch
+func GetLatestVersion(shortHash bool) (latestVersion string, latestRelease string, err error) {
+
+	platform := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
+	wantedRelease := stashReleases()[platform]
+
+	branch, _, _ := GetVersion()
+	if branch == "" {
+		return "", "", fmt.Errorf("Stash doesn't have a version. Version check not supported.")
+	}
+
+	client := &http.Client{
+		Timeout: 3 * time.Second,
+	}
+
+	req, _ := http.NewRequest("GET", apiReleases, nil)
+
+	req.Header.Add("Accept", apiAcceptHeader) // gh api recommendation , send header with api version
+	response, err := client.Do(req)
+
+	input := make([]githubReleasesResponse, 0)
+
+	if err != nil {
+		return "", "", fmt.Errorf("Github API request failed: %s", err)
+	} else {
+
+		defer response.Body.Close()
+
+		data, err := ioutil.ReadAll(response.Body)
+		if err != nil {
+			return "", "", fmt.Errorf("Github API read response failed: %s", err)
+		} else {
+			err = json.Unmarshal(data, &input)
+			if err != nil {
+				return "", "", fmt.Errorf("Unmarshalling Github API response failed: %s", err)
+			} else {
+
+				for _, ghApi := range input {
+					if ghApi.Tag_name == branch {
+
+						if shortHash {
+							latestVersion = ghApi.Target_commitish[0:7] //shorthash is first 7 digits of git commit hash
+						} else {
+							latestVersion = ghApi.Target_commitish
+						}
+						if wantedRelease != "" {
+							for _, asset := range ghApi.Assets {
+								if asset.Name == wantedRelease {
+									latestRelease = asset.Browser_download_url
+									break
+								}
+
+							}
+						}
+						break
+					}
+				}
+
+			}
+		}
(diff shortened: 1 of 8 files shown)


Leopere:
Eventually, I really would like to update automagically by default.  Potentially allow people to disable the update mechanism via some command line launch argument or something not in the UI.

WithoutPants:
Reviewed but not yet tested.

pkg/api/check_version.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if err != nil {
+		return "", fmt.Errorf("Github API request failed: %s", err)
+	} else {
+
+		defer response.Body.Close()
+
+		data, err := ioutil.ReadAll(response.Body)
+		if err != nil {
+			return "", fmt.Errorf("Github API read respone failed: %s", err)
```
WithoutPants: Typo - respon**s**e

pkg/api/check_version.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+//gets latest version (git commit hash) from github API
+//the repo's tags are used to find the latest version
+//of the "master" or "develop" branch
+func GetLatestVersion(shortHash bool) (string, error) {
+
+	branch, _, _ := GetVersion()
+	if branch == "" {
+		return "", fmt.Errorf("Stash doesn't have a version.Version check not supported.")
```
WithoutPants: Add space between sentences.

Makefile
```diff
@@ -6,7 +6,7 @@ endif
 build:
 	$(eval DATE := $(shell go run scripts/getDate.go))
 	$(eval GITHASH := $(shell git rev-parse --short HEAD))
-	$(SET) CGO_ENABLED=1 $(SEPARATOR) go build -mod=vendor -v -ldflags "-X 'github.com/stashapp/stash/pkg/api.buildstamp=$(DATE)' -X 'github.com/stashapp/stash/pkg/api.githash=$(GITHASH)'"
+	$(SET) CGO_ENABLED=1 $(SEPARATOR) go build -mod=vendor -v -ldflags "-X 'github.com/stashapp/stash/pkg/api.buildstamp=$(DATE)' -X 'github.com/stashapp/stash/pkg/api.githash=$(GITHASH)' "
```
WithoutPants: Why is this change necessary?

WithoutPants:
Since we are getting the URL anyway, is there any reason we can't or shouldn't include a direct link to the latest download in the about page?

pkg/api/check_version.go
bnkai: fiixe

pkg/api/check_version.go
bnkai: fixed

Makefile
bnkai: Sorry  , missed that. It was leftover from manual editing the Make file to add the version and so a space was left there.

bnkai:
> Since we are getting the URL anyway, is there any reason we can't or shouldn't include a direct link to the latest download in the about page?

The API endpoint i am quering is this one [tags](https://api.github.com/repos/stashapp/stash/tags).
It only has the versions and the URL of the zipfile , tarball **source code** no releases.
There is also the [releases](https://api.github.com/repos/stashapp/stash/releases) one that can be queried ,  it is more complex to parse though.
I'll have a look at it , switching to WIP till i settle this.

ueaslsef:
The download path should be rather straight forward: `https://github.com/stashapp/stash/releases/download/$version/stash-$platform`. Maybe we could cheat and skip the releases-api parsing all together and just use this until we found a better way? It could even be a static string inserted by the ci on build time.

bnkai:
@ueaslsef  since we are doing it we might as well go all the way  ..
The releases API has extra stuff also like times downloaded or date released. Do we want any of that ?

bnkai:
squashed and rebased , also fixed a merge conflict due to latest commits.
this the latest version with the proper release added as download link.
releases

pkg/api/check_version.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return latestVersion, latestRelease, nil
+
+}
+
+func printLatestVersion() {
+	_, githash, _ = GetVersion()
+	latest, _, err := GetLatestVersion(true)
+	if err != nil {
+		logger.Errorf("Couldn't find latest version :%s\n", err)
```
WithoutPants: Change to `version: %s`. Looks weird in the log currently. The `\n` is unnecessary and pollutes the output.

ui/v2/src/components/Settings/SettingsAboutPanel.tsx
```diff
@@ -26,32 +28,72 @@ export const SettingsAboutPanel: FunctionComponent<IProps> = (props: IProps) =>
...
           <tr>
-            <td>Build hash:</td>
-            <td>{data.version.hash}</td>
+            < td>Latest Version Build Hash: </td>
+            < td>{maybeRenderLatestVersion()} </td>
           </tr>
           <tr>
-            <td>Build time:</td>
-            <td>{data.version.build_time}</td>
+            <td><Button minimal={true} onClick={() => refetch()} text="Check for new version!" /></td>
```
WithoutPants: Having `minimal={true}` makes the button look weird in this context when not moused over. I'd get rid of it. I'm not sure that the exclamation mark is necessary, but that's just a subjective aesthetic opinion.

pkg/api/check_version.go
```diff
@@ -0,0 +1,179 @@
...
+	} else {
+		if githash == latest {
+			logger.Infof("Version: (%s) is already the latest released.\n", latest)
```
WithoutPants: Remove `\n`

pkg/api/check_version.go
```diff
@@ -0,0 +1,179 @@
...
+		} else {
+			logger.Infof("New version: (%s) available.\n", latest)
```
WithoutPants: Remove `\n`.

ui/v2/src/components/Settings/SettingsAboutPanel.tsx
```diff
@@ -26,32 +28,72 @@ export const SettingsAboutPanel: FunctionComponent<IProps> = (props: IProps) =>
...
     return (
-      <>
       <HTMLTable>
         <tbody>
-          {maybeRenderTag()}
           <tr>
-            <td>Build hash:</td>
-            <td>{data.version.hash}</td>
+            < td>Latest Version Build Hash: </td>
+            < td>{maybeRenderLatestVersion()} </td>
```
WithoutPants: Spaces before the `td` tag need to be fixed.

ui/v2/src/components/Settings/SettingsAboutPanel.tsx
```diff
@@ -26,32 +28,72 @@ export const SettingsAboutPanel: FunctionComponent<IProps> = (props: IProps) =>
...
+        </>
+      );
+    }
+
+    return (
+      <>{dt.latestversion.shorthash}</>
+    );
+  }
+
+  function RenderLatestVersion() {
```
WithoutPants: First letter should be lowercase.

ui/v2/src/components/Settings/SettingsAboutPanel.tsx
```diff
@@ -26,32 +28,72 @@ export const SettingsAboutPanel: FunctionComponent<IProps> = (props: IProps) =>
...
   return (
     <>
       <H4>About</H4>
       {!data || loading ? <Spinner size={Spinner.SIZE_LARGE} /> : undefined}
-      {!!error ? <span>error.message</span> : undefined}
+      {!!error ? <span>{error.message}</span> : undefined}
+      {!!er ? <span>{er.message}</span> : undefined}
       {renderVersion()}
+      {!dt || ld || networkStatus == 4 ? <Spinner size={Spinner.SIZE_SMALL} /> : undefined}
+      {RenderLatestVersion()}
```
WithoutPants: I think these variables (`dt`, `ld`, `er`) need to be better named.

I would make it so that if the spinner is showing (ie the query is in progress) then `renderLatestVersion` is not called.

pkg/api/check_version.go
bnkai: i think i added it for the console, removed anyway

ui/v2/src/components/Settings/SettingsAboutPanel.tsx
bnkai: i tried the combo minimal and ! and plain button without the !
i'll change back to plain button

ui/v2/src/components/Settings/SettingsAboutPanel.tsx
bnkai: should be covered with latest commit ....

pkg/api/check_version.go
WithoutPants: I need to proofread my feedback. This was supposed to be two separate items: the colon position and the `\n` character. Currently, the log looks like this:
```
time="2020-01-05T09:54:31+11:00" level=error msg="Couldn't find latest version :Stash doesn't have a version. Version check not supported."
```

Should be `version: %s` rather than `version :%s`.

bnkai: missed that, fixed.

PR #277, opened 2019-12-18
Re-License to AGPL

Leopere:
fixes #276 requires all previous contributors to sign off on the issue before we can merge this thusly replacing the License.  All previous commits and versions can technically be forked under the MIT license if people disagree with the final decision.

Issue #276, opened 2019-12-18
[RFC] Re-License to AGPL or Other for the sake of keeping this open source long term.


Leopere:
Reply with a license and version we want and then contributors up to date will need to thumbs up to agree with a specific license all contributors must agree.

##  Long Form

If you wish to suggest and pitch your own license alternative you're welcomed to pitch one.  My suggestion is to drive Stash to be a forever FOSS project and to have a fully open source API where sustainability is possible and the governance model doesn't matter.  

If all tools are required to be open sourced and licensed under AGPL we can ensure that the API will forever be publicly available for the world despite any financial renumeration schema we adopt.

## Process Steps

- [ ] Collect thumbs ups on whatever licenses are proposed.
- [ ] Screenshot and upload into the git issue logs of all of the Thumbs Up's.
- [ ] Screenshot and upload into the git issue logs all of the contributors to date.
- [ ] Merge PR #277 

## Reference Reading

https://choosealicense.com/licenses/agpl-3.0/

FleetingOrchard:
I have to admit, I'm not sure the thumbs up would be a viable way to show that a given contributor has relicenced their contributions to stash under the new, desired licence. It would probably be easier to simply ask each contributor for a short statement to the effect of relicencing their work.

But it's a good idea, and AGPL is a good pick.

Leopere:
Realistically we just need some record of an agreement.   If anything we can commit the screenshots to the repository for one commit/PR or something then remove the file in subsequent builds just so that it's in the history.

StashAppDev:
@friendlycrab @endqwerty @ExceptionalError @echo6ix @rudi123github @joyov @InfiniteTF 

Tagging every contributor who I haven't seen comment yet.  Do you folks mind commenting or adding a +1 with your approval / disapproval?

endqwerty:
:+1:
Any changes now or in the future to the license or otherwise for any content which I have contributed to this project are approved by me.

ExceptionalError:
I approve the change to AGPL.

WithoutPants:
I approve the change to AGPL.

WithoutPants:
Assuming that we're not retrospectively changing the license of existing code and the license change applies only once the PR is merged, I think we should set a deadline for objections and just go ahead otherwise.

WithoutPants:
I also don't think screenshotting is necessary.

Leopere:
[...]

PR #297, opened 2019-12-29
Add LICENSE until we get all signers on the switch

Leopere:
For now, all subsequent commits should be against the appropriate license.

Related pre-requisites for switching from a SUBLICENSE file to just LICENSING this code to AGPL fully.
#276 #277


bnkai:
I think it is better to add this ASAP . Is there antyhing else needed to change to make every commit from now on with this license or to select this license when you make a PR ?

WithoutPants:
Why the urgency for this? Unless we get an influx of PRs from new contributors (which we don't have to approve until we change the license), I don't see why we can't wait for the sign off and change the license entirely.

FleetingOrchard:
Not to be a jerk, but WP is on the money here. The license is only ever going to be relevant if IP drama goes down -- as in, at that specific moment in time, and not sooner.

In the meantime, if we want to continue to bring in new contributors (assuming somehow the switch ends up taking longer than expected), maintainers can just make getting an RP in conditional on that contributor "signing off" on AGPL.

Ultimately I'd suggest not worrying too much about this. There's enough being done here already that's not technically by the book but will probably fly on count of commonly understood meaning and intent, having things be a bit in limbo for a while is going to be fine.

Leopere:
No real urgency so much as an option because we have a number of people contributing if new contributors contribute and we're planning on the switch we would thus need them to agree to the license change along with their PR before we could accept it however with a SUBLICENSE it might be less stress involved.

As I am still somewhat iffy on this myself I put it up for review from the other major contributors to Stash.

Issue #291, opened 2019-12-25
[Feature] Splooge button


echo6ix:
Provide a splooge button ala Pornganizer for scenes as an additional way to rate scenes, and let the end-user filter scenes based on splooge count eg. `splooge < 1` will ideally provide fresh scenes

**Describe the solution you'd like**
Each scene would have a splooge counter. When the scene's splooge button is pressed the counter will increment by 1.

**Additional context**
The basic UI of the splooge button can be `{splooge icon button} {visual counter} {bp3-icon-caret-down button}` in a `ButtonGroup`  where,

- **{splooge icon button}** is a splooge icon and button. The `tint` icon from [Blueprint](https://blueprintjs.com/docs/#icons) is the only suitable icon they have imo. Pressing the button increments the counter by 1.
- **{visual counter}** displays the number of splooges to the right of the `tint` icon on the splooge button.
- **{bp3-icon-caret-down button}** indicates a small dropdown menu (I think Blueprint calls this a popover wrapper?) that would let the user:
 -- `Decrement` the splooge counter by one for any accidental button presses
 -- `Reset` the splooge counter to 0

**Mockup concepts**

Scene cards

![Annotation 2019-12-25 205057](https://user-images.githubusercontent.com/37937507/71453267-5d0d9d00-2758-11ea-8c95-518a4ad1dec1.png)

Leopere:
I get a laugh out of the concept, but I'd imagine just a heart or a star would suffice and be a bit more professional.  I'm indifferent if people follow my lead for "professionalism" here but just wanted to pitch my thoughts.

WithoutPants:
I'd prefer not to call it "splooge" as well, not just for aesthetics, but also because it makes it a carbon copy of the feature in PGV. I agree with a heart or star. As a text/db field, times hearted might be sufficient.

dotbanana:
I like the idea very much and i personally see no reason to not use "Splooge" or some other 'ordinary expression', because it exactly describes what this feature is for. A star or a heart are used in contemporary social media semiotics for other purposes -> For favorites and for likes.
And imo there is no need to be sensitive about the wording - this script is in the end for porn-organization and not for winning some coding contest or using it for the purpose of getting a job (or so do i assume/hope ^^).

Leopere:
Sure, and you have a perspective on the matter, but I would return with something along the lines of just that, not everyone splooges per-say, and some may prefer something a bit more generic.

echo6ix:
[...]

PR #298, opened 2019-12-30
[WIP/RFC] Add rudimentary "watch count" scene attribute

FleetingOrchard:
This is still pretty rudimentary (and mainly intended, at the moment, to spur a bit of discussion as a response to/more functional alternative to #291), but hey, it works. And doesn't seem to trash data on migrations. Success!

Currently it's just an inert number that doesn't auto-increment, or provide a "+1" button (mostly because I can't figure out a good place to put such a thing in the current UI), so don't actually merge this in.

### DIFF
--- a/graphql/documents/data/scene-slim.graphql
+++ b/graphql/documents/data/scene-slim.graphql
@@ -6,6 +6,7 @@ fragment SlimSceneData on Scene {
   url
   date
   rating
+  watch_count
   path
 
   file {
--- a/graphql/documents/data/scene.graphql
+++ b/graphql/documents/data/scene.graphql
@@ -6,6 +6,7 @@ fragment SceneData on Scene {
   url
   date
   rating
+  watch_count
   path
 
   file {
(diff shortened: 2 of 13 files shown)


Leopere:
Even the +1 button would be useful however the one problem is what conditions trigger allowing it's potentially repeated use.

WithoutPants:
A "watch count" field should be something that is automatically incremented when user watches a scene - which is a possible future feature. If we're to add functionality based on #291 then I'd prefer it not be named "watch count".

FleetingOrchard:
I appreciate that, I'm suggesting this as an alternate to 291 -- as to my eye, a watch count more accurately fulfills the actual use case (establishing if something is novel or familiar to the user, and to what degree).

Auto increment just isn't something I've hooked up yet -- I unfortunately didn't have time to do so earlier, and will be adding to this PR as time goes on. I figured it would be worth getting some eyeballs on the idea in the meantime, though, given I'm suggesting here implementing a request by adding something that is, semantically, not that request.

Leopere:
If this tracking stuff is done it should be toggled in settings.

WithoutPants:
#334 has been merged which addresses #291. This could be a good basis for a watch counter functionality, but given the original intention, I'll close this for now.

PR #299, opened 2019-12-30
Show localhost in console output instead of 0.0.0.0

InfiniteTF:
0.0.0.0 is not a valid address, and can be confusing to novice users. In most cases localhost is more correct.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -200,6 +200,12 @@ func Start() {
 		}
 	})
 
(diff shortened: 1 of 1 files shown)


Leopere:
Just closing the loop between discord and here the eventual function here will be to have it say http://localhost:{port}/ but this is the IP that we're listening on.

StashAppDev:
Closing since this was merged in with #300

Issue #227, opened 2019-11-27
DVD Section?


Cutters1982:
Hi there

Apologies if this has already been covered under another issue, I had a bit of a search and couldnt find it discussed.

I wanted to request a DVD section be added to Stash to be alongside the Scenes section, but with obvious differences where you can upload the front & back cover images of the DVD, and add the individual scenes under the DVD entry with subsequent performer tagging and scene infos. So you would left-click on the DVD in the list, and it would show all of the scenes of that DVD as if it were the Scenes section

The app is coming along really nicely, very excited to see what will come in the next UI release!

Cheers

binarygeek119:
i would also like to add with cover uploads disc image upload this could tie into the metadata where it could upload missing images for the ui part i like to see  in dvd broswer part just the front image when hovering over dvd it inlarges and then after X secs fades to back cover

in the dvd movie part (milfs 3) the front and back cover show side by side with disc in middle
with the disc spinning
![poc3](https://user-images.githubusercontent.com/47513721/69692035-76a6c080-1096-11ea-9a35-9fbbb262b5e4.gif)
![poc2](https://user-images.githubusercontent.com/47513721/69692037-773f5700-1096-11ea-821f-78c9b67f1e34.jpg)
![poc1](https://user-images.githubusercontent.com/47513721/69692038-773f5700-1096-11ea-87e2-1f675b097b01.jpg)

Leopere:
If we do a DVD section could we aim for something that shows how complete the data is?  Like includes the main video content or also includes alternative scene data (dvd extras) and whatnot.

WithoutPants:
#338 addresses this and is now merged, so closing. If there are outstanding features, please raise a separate issue.

PR #316, opened 2020-01-12
[WIP] DVD Section #227

caustico:
Hi,
I have been modifying the code to add a DVDs section to group the DVDs's scenes.

The changes included are many and this changes modify the server and the client (the model of Database, the export / import models, ....

I'm going to add images to show the new funtionality.

Before performing any test, I recommend to you make a backup of your database and jsons files.

Regards

Movies- Scenes
Movies-List
Movies-Movie-Detail
Movies-Scenes-Edit
movies-scenes-movies
Scenes-Movie-Filter


Cutters1982:
This is much needed, thank you so much! :-)

Makefile
```diff
@@ -1,3 +1,4 @@
+
```
WithoutPants: Unnecessary change.

go.mod
```diff
@@ -12,19 +12,21 @@ require (
...
+	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
 	golang.org/x/image v0.0.0-20190118043309-183bebdce1b2 // indirect
-	golang.org/x/net v0.0.0-20190522155817-f3200d17e092
+	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
 	gopkg.in/yaml.v2 v2.2.2
 )
 
 replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
+
+go 1.13
```
WithoutPants: Pretty sure we target 1.10 from memory? We should avoid changing this unless necessary.

graphql/schema/types/dvd.graphql
```diff
@@ -0,0 +1,51 @@
+type Dvd {
```
WithoutPants: Convention in go and elsewhere is for classnames that are acronyms should be all in capital - ie this should `DVD`. As such the references to this should changed as well - `allDVD` for example. Where it is in lowercase, it can stay as such - `dvdCreate` is fine for example. This should be changed in the graphql schema and in the golang code.

graphql/schema/types/dvd.graphql
```diff
@@ -0,0 +1,51 @@
...
+  id: ID!
+  checksum: String!
+  name: String!
+  aliases: String
+  durationdvd: String
```
WithoutPants: Is there a reason this is not named `duration`?

graphql/documents/mutations/dvd.graphql
```diff
@@ -0,0 +1,36 @@
+mutation DvdCreate(
+  $name: String!,
+  $aliases: String,
+  $durationdvd: String,
+  $year: String,
+  $director: String,
+  $synopsis: String,
+  $url: String,
+  $frontimage: String,
+  $backimage: String) {
```
WithoutPants: These should be `$front_image` and `$back_image`. There are other places where these fields exist that should also be fixed. I won't include all of them in the review.

pkg/api/resolver_model_dvd.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/stashapp/stash/pkg/api/urlbuilders"
+	"github.com/stashapp/stash/pkg/models"
+)
+
+func (r *dvdResolver) Name(ctx context.Context, obj *models.Dvd) (string, error) {
+	if obj.Name.Valid {
+		return obj.Name.String, nil
+	}
+	panic("null name") // TODO make name required
+}
```
WithoutPants: Name should be required. This resolver won't be required if it is.

graphql/schema/types/dvd.graphql
```diff
@@ -0,0 +1,51 @@
+type Dvd {
+  id: ID!
+  checksum: String!
+  name: String!
+  aliases: String
```
WithoutPants: What is the purpose of a DVD alias?

pkg/database/migrations/3_dvd.up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE `dvds` (
+  `id` integer not null primary key autoincrement,
+  `name` varchar(255),
```
WithoutPants: Make this `not null`, and the golang model can be changed from `NullString` to `string`, and the resolver for name is no longer necessary.

pkg/models/querybuilder_tag.go
```diff
@@ -33,7 +33,23 @@ func (qb *TagQueryBuilder) Create(newTag Tag, tx *sqlx.Tx) (*Tag, error) {
...
+		return nil, err
+	}
+
+
 	return &newTag, nil
+
+
+
+
+
```
WithoutPants: Unnecessary whitespace.

ui/v2/src/components/Dvds/DvdList.tsx
```diff
@@ -0,0 +1,36 @@
...
+      return (
+        <div className="grid">
+          {result.data.findDvds.dvds.map((dvd) => (<DvdCard key={dvd.id} dvd={dvd} />))}
+        </div>
+      );
+    } else if (filter.displayMode === DisplayMode.List) {
+      return <h1>TODO</h1>;
+    } else if (filter.displayMode === DisplayMode.Wall) {
+      return <h1>TODO</h1>;
+    }
```
WithoutPants: I'm thinking a wall view is probably not necessary for DVDs, at least in this iteration.

InfiniteTF:
First of all, great stuff, this will make a lot of people happy.

I do have some comments on the schema though. A lot of scenes are re-released many times on compilations, etc., which means they can be attached to multiple dvds. I think we should consider having a scene_dvd table to allow scenes to be on arbitrary amount of dvds.

I also think we should consider having an index property to denote which number the scene is so we can sort the scenes properly without putting the scene number in the title.

Lastly, and this is maybe a bit of bikeshedding, but I wonder if we should use a more neutral term than DVD, something like Films or Movies for instance. DVD excludes a significant amount of releases both before and after the rise/fall of dvds.

ui/v2/src/components/Stats.tsx
```diff
@@ -33,6 +33,13 @@ export const Stats: FunctionComponent = () => {
             <p className="title">{data.stats.studio_count}</p>
           </div>
         </div>
+        <div className="level-item has-text-centered">
+          <div>
+            <p className="heading">Dvds</p>
```
WithoutPants: I'd make this `DVDs`.

ui/v2/src/components/select/FilterMultiSelect.tsx
```diff
@@ -106,12 +107,19 @@ export const FilterMultiSelect: React.FunctionComponent<IProps> = (props: IProps
         translateData = () => { let perfData = data as GQL.AllPerformersForFilterQuery; setItems(!!perfData && !!perfData.allPerformers ? perfData.allPerformers : []); };
         break;
       }
-      case "studios": {
+     /*  case "studios": {
```
WithoutPants: What is the purpose of commenting this out? If it's superfluous it should probably just be removed altogether.

bnkai:
Most of what i see looks good, i would like to test after the changes @InfiniteTF @WithoutPants  proposed.
I think also DVDs is a little too specific? Probably movies or collections (changing the name only on the UI would suffice imho) would be more appropriate.
As @InfiniteTF mentioned scenes tend to be reused for different  DVD , movie releases so making an extra table dvd_scenes might be better. You can have a look at the scene_tags or performer_scenes tables.A couple of fk, the id and maybe an extra index of the scene (as @InfiniteTF mentioned) should be enough.
Also is there a reason for the Year instead of Date  in the dvd  fields? I don't mind leaving it like that if it's for simplicity , given that scenes already have dates.

Makefile
caustico: I will revert the change

go.mod
caustico: I will revert the change

Cutters1982:
> Most of what i see looks good, i would like to test after the changes @InfiniteTF @WithoutPants proposed.
> I think also DVDs is a little too specific? Probably movies or collections (changing the name only on the UI would suffice imho) would be more appropriate.
> As @InfiniteTF mentioned scenes tend to be reused for different DVD , movie releases so making an extra table dvd_scenes might be better. You can have a look at the scene_tags or performer_scenes tables.A couple of fk, the id and maybe an extra index of the scene (as @InfiniteTF mentioned) should be enough.
> Also is there a reason for the Year instead of Date in the dvd fields? I don't mind leaving it like that if it's for simplicity , given that scenes already have dates.

Yeah, you're right, "Movies" is fine with me, instead of it saying Dvds / DVDs

graphql/schema/types/dvd.graphql
caustico: I used "durationdvd" because I had a problem with the "duration" field in the [scenes] table. There is a "left join" in a query what caused me this problem.

graphql/schema/types/dvd.graphql
caustico: I'm going to try  to adapated to the convention,

graphql/documents/mutations/dvd.graphql
caustico: I'm going to try to resolve this too.

pkg/models/querybuilder_tag.go
caustico: I will delete the whitespace.

pkg/database/migrations/3_dvd.up.sql
caustico: I will change it.

ui/v2/src/components/Stats.tsx
caustico: I will change it.

ui/v2/src/components/select/FilterMultiSelect.tsx
caustico: I will revise it.

caustico:
Modifications and stuff for the next iteration:

-Revert unnecessary changes and try to make the requested changes in the code.

-Change the name of Dvds to Movies (for now in the user interface) and I will probably move it between scenes and bookmarks instead of on the right side.

-Create a "scenes_dvd" table to allow assign scenes to different DVDs | Movies ... I think that for this change it is necessary to create a new Panel in Scene Detail.

-"Scene Index" to indicate the scene number in the movie.

-I choosed the "Year" field instead of the "Date" field because the "Year" is known in most movies and it is harder to find the "Date" release. Anyway, I can change the "Year" field to the "Date" field. No problem. I used the "Aliases" field because some movies have differents titles in differents languages.


-Due to changes in the database. I'm going to add one or more field too ... Maybe I will add a "Rating" field to rate the movie.

I think I don't forget anything.

graphql/schema/types/dvd.graphql
WithoutPants: That can be resolved by adding the table qualifier to the query - `scenes.duration`, `dvd.duration`, for example.

WithoutPants:
I've marked this as WIP while you make the changes. Rename it back when it's ready to review.

Leopere:
This feels hugely out of scope, but if the community wants it, then that sounds like an excellent feature added.  I think that this is already an issue relatively well handled by Jellyfin, but healthy competition is essential.

WithoutPants:
Why was this closed?

StashAppDev:
@WithoutPants https://github.com/stashapp/stash/pull/338

PR #317, opened 2020-01-14
Fix react warnings

WithoutPants:
This removes most of the warnings generated by eslint. There are 11 warnings remaining that will require further refactoring.

- update dependencies and set the go version to the same as that used in the travis configuration
- remove unused imports
- remove unused variables
- add alt text to images
- replace `==` with `===`
- replace `throw` clause with bare string to `Error`
- replace `a` elements displayed as buttons with buttons
- add missing `useEffect` dependencies where feasible


bnkai:
Looks ok to me.

PR #319, opened 2020-01-14
Refresh yarn.lock

bnkai:
Recreated yarn.lock
Should also take care of https://cdn.discordapp.com/attachments/559169859351740418/663058576075653131/unknown.png

### DIFF
--- a/ui/v2/yarn.lock
+++ b/ui/v2/yarn.lock
@@ -9,12 +9,12 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5", "@babel/code-frame@^7.8.0":
-  version "7.8.0"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.8.0.tgz#8c98d4ac29d6f80f28127b1bc50970a72086c5ac"
-  integrity sha512-AN2IR/wCUYsM+PdErq6Bp3RFTXl8W0p9Nmymm7zkpsCmh+r/YYcckaCGpU8Q/mEKmST19kkGRaG42A/jxOWwBA==
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5", "@babel/code-frame@^7.8.3":
+  version "7.8.3"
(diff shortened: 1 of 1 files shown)


WithoutPants:
I deleted and regenerated this file and ended up with some different results:
- `babel` packages resolved as `7.8.3` instead of `7.8.0`.
- `@types/istanbul-lib-report@*` resolved to `3.0.0` instead of `1.1.1`.
- `@types/node@*` resolved to `13.1.8` instead of `13.1.6`.
- and plenty more

How did you regenerate?

bnkai:
deleted the cache and the lock file and rerun yarn if i remember correctly.

WithoutPants:
Yes, mine removed handlebars as well. Can you try regenerating again? I find it odd that I'm getting a different version for `babel` than you have here.

bnkai:
I just recreated, the difference was because of new versions of packages.
I have the `yarn why` of the ones you mentioned here
[before](https://pastebin.com/fpydmE0F)
[latest commit](https://pastebin.com/c9DFhBrg)

WithoutPants:
I think this is good to go ahead, but I'm pretty confused by this file.

Deleting the yarn.lock file and running `yarn install` resulted in a couple of differences, but not as many as before.

However, if I restored to this branch version of yarn.lock, then ran `yarn install`, I ended up with a bunch of dependencies removed: `deep-extend`, `detect-libc`, `fs-minipass`, `ignore-walk`, `minipass`, `minizlib`, `needle`, `node-pre-gyp`, `nopt`, and a bunch more.

I haven't tried removing `node_modules` and regenerating.

[This issue](https://github.com/yarnpkg/yarn/issues/7770) might provide some answers.

I verified this change by running `yarn install --frozen-lockfile`, then performing a full build and doing a quick test of the app. It all looks ok.

I'm going to raise a new PR to use `--frozen-lockfile` in the travisci file, and in the README instructions, so that we end up with deterministic dependencies across builds.

WithoutPants:
Just as an addendum to this. I ran `yarn install --frozen-lockfile` on the pre-merged code, then ran a build and ended up with a bunch of missing dependencies. I had to remove the yarn.lock file, then run `yarn install`, before I could build successfully. After the merge, it all works as expected.

PR #324, opened 2020-01-17
Add timeout to neededGenerate function

bnkai:
Added a timeout to neededGenerate, the function that calculates how many screenshots,previews,sprites and transcodes are needed prior to starting the actual generation.
A default of 90 seconds was added that can tweaked further if needed.

### DIFF
--- a/pkg/manager/manager_tasks.go
+++ b/pkg/manager/manager_tasks.go
@@ -165,8 +165,11 @@ func (s *singleton) Generate(sprites bool, previews bool, markers bool, transcod
 			return
 		}
 		totalsNeeded := s.neededGenerate(scenes, sprites, previews, markers, transcodes)
-		logger.Infof("Generating %d sprites %d previews %d markers %d transcodes", totalsNeeded.sprites, totalsNeeded.previews, totalsNeeded.markers, totalsNeeded.transcodes)
-
+		if totalsNeeded == nil {
+			logger.Infof("Generating content")
+		} else {
+			logger.Infof("Generating %d sprites %d previews %d markers %d transcodes", totalsNeeded.sprites, totalsNeeded.previews, totalsNeeded.markers, totalsNeeded.transcodes)
+		}
 		for i, scene := range scenes {
 			s.Status.setProgress(i, total)
 			if s.Status.stopping {
@@ -208,6 +211,7 @@ func (s *singleton) Generate(sprites bool, previews bool, markers bool, transcod
 
 			wg.Wait()
 		}
+		logger.Infof("Generate finished")
 	}()
 }
 
@@ -461,6 +465,12 @@ type totalsGenerate struct {
 func (s *singleton) neededGenerate(scenes []*models.Scene, sprites, previews, markers, transcodes bool) *totalsGenerate {
 
 	var totals totalsGenerate
+	chTimeout := make(chan struct{})
+	go func() {
+		time.Sleep(90 * time.Second)
+		chTimeout <- struct{}{}
+	}()
+
 	for _, scene := range scenes {
 		if scene != nil {
 			if sprites {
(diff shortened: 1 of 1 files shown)


pkg/manager/manager_tasks.go
```diff
@@ -461,6 +465,12 @@ type totalsGenerate struct {
 func (s *singleton) neededGenerate(scenes []*models.Scene, sprites, previews, markers, transcodes bool) *totalsGenerate {
 
 	var totals totalsGenerate
+	chTimeout := make(chan struct{})
+	go func() {
+		time.Sleep(90 * time.Second)
```
WithoutPants: Make `90` into a const.

pkg/manager/manager_tasks.go
```diff
@@ -461,6 +465,12 @@ type totalsGenerate struct {
 func (s *singleton) neededGenerate(scenes []*models.Scene, sprites, previews, markers, transcodes bool) *totalsGenerate {
 
 	var totals totalsGenerate
+	chTimeout := make(chan struct{})
```
WithoutPants: Can you add a comment explaining why this is added?

pkg/manager/manager_tasks.go
```diff
@@ -165,8 +165,11 @@ func (s *singleton) Generate(sprites bool, previews bool, markers bool, transcod
 			return
 		}
 		totalsNeeded := s.neededGenerate(scenes, sprites, previews, markers, transcodes)
-		logger.Infof("Generating %d sprites %d previews %d markers %d transcodes", totalsNeeded.sprites, totalsNeeded.previews, totalsNeeded.markers, totalsNeeded.transcodes)
-
+		if totalsNeeded == nil {
```
WithoutPants: I think we should add an info log here just saying that it is determining job length or something, because currently there's no logging for 90 seconds, which may lead the user to thinking something is wrong.

pkg/manager/manager_tasks.go
```diff
@@ -166,6 +166,7 @@ func (s *singleton) Generate(sprites bool, previews bool, markers bool, transcod
 		}
 		totalsNeeded := s.neededGenerate(scenes, sprites, previews, markers, transcodes)
 		if totalsNeeded == nil {
+			logger.Infof("Counting content timeout!")
```
WithoutPants: Is this supposed to be here?

pkg/manager/manager_tasks.go
```diff
@@ -465,12 +466,18 @@ type totalsGenerate struct {
...
 	chTimeout := make(chan struct{})
+
+	//run the timeout function in a separate thread
 	go func() {
-		time.Sleep(90 * time.Second)
+		time.Sleep(timeoutSecs)
 		chTimeout <- struct{}{}
 	}()
 
+	logger.Infof("Counting content to generate.Please wait ...")
```
WithoutPants: The `Please wait ...` is probably not necessary.

pkg/manager/manager_tasks.go
bnkai: Yes. If the timeout occurs i thought we should be a little more verbose than just printing the generating content message. It probably isn't necessary  but thats where it should be if we decide to keep it.

pkg/manager/manager_tasks.go
bnkai: I just wanted to make sure the user knows something is going on that might take some time(nothing will be printed before counting is finished or timeout occurs) , maybe just add the ... like this "Counting content to generate ..." ?

WithoutPants: Yep, `Counting content to generate...` sounds good to me.

pkg/manager/manager_tasks.go
WithoutPants: Yeah, I get it now. I'd probably change the phrasing. `Taking too long to count content. Skipping...`, or something like that.

bnkai: changed.

pkg/manager/manager_tasks.go
bnkai: done

Issue #99, opened 2019-08-14
[Feature] Allow user to set default scene page parameters in settings


echo6ix:
Whenever I go to my scene page I always have to set my _sortby_ to _created_at_, _sortdir_ to _desc_, and other sorting, filter, and display parameters exactly how I want them.

Consider having a section in the settings page that lets the user define the default parameters for the scene page.

WithoutPants:
A better alternative might be to save these settings in a cookie, so that settings are persisted between visits to the scenes page.

echo6ix:
I should have included the reasoning behind my suggestion: given the nature of the content being viewed, a lot of people may be like myself and only view Stash on an incognito/private browser mode rendering cookies inconsequential.

Maybe implementing both would be a good idea?

joyov:
Another option would be to save the last used settings in the server's database. I do prefer a solution that doesn't require cookies too.

WithoutPants:
#326 fixes the issue of queries persisting between page navigations. I'm thinking of tying this issue in with a broader Saved Queries feature, with the option of selecting a given query as the "default" query, which is what will be displayed if there is no saved query.

PR #326, opened 2020-01-22
Maintain filter parameters in session

WithoutPants:
Changes so that the last query parameters are stored in local browser storage. This means that if you navigate to another page, then click back onto Scenes, for example, it will show the results of the last query you ran, including the page you were on. This is overridden if you go to a URL with the search parameters included.

For incognito browsers, this will only last as long as the browser instance is running. For non-incognito browsers, this should be persisted between sessions.

This forms a sort-of workaround for #99

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,8 @@ require (
 	github.com/golang-migrate/migrate/v4 v4.3.1
 	github.com/gorilla/websocket v1.4.0
 	github.com/h2non/filetype v1.0.8
+	// this is required for generate
+	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/jmoiron/sqlx v1.2.0
 	github.com/mattn/go-sqlite3 v1.10.0
 	github.com/rs/cors v1.6.0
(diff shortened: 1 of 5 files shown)


bnkai:
A minor issue i encountered upon testing.
When migrating to this version to test i get this (on firefox)
error
Closing the browser i  and reopening doesn't make a difference (despite reasking the password to access stash). Opening in a new private window or a different browser that i don't usually use stash (opera, chrome) everything works ok.
I had to clear data to make it work again in my firefox
clear.
It seems that some data is left on the browser with the old settings that interferes with  new session settings.

WithoutPants:
I haven't been able to reproduce this issue, but I think the last commit should have resolved it.

bnkai:
I'm having trouble reproducing this also now ( mainly because i cleared my main firefox's problematic history) .
The change look ok though and the latest commit works fine so i think it is ok.

WithoutPants:
I've managed to reproduce the error or a variant of it by building the stable version, running the UI against that after clearing the forage database, then updating to this branch and reloading. I'm investigating now.

WithoutPants:
False alarm. I was running against the pre-fix commit, which reproduced the error. Once I updated to the latest commit, the problem was resolved, so the fix is verified. Ready to merge.

Issue #302, opened 2020-01-01
[Enhancement] Capability of running Scene Filename Parser on certain folders


Cutters1982:
Hi there

Happy New Years guys.

I would like to suggest an enhancement to the Scene Filename Parser where you can select what folders in your Archive that you wish to run it on, rather than it scanning everything every time, otherwise you keep having to untick stuff that you dont want it to rename.

This would be beneficial when adding a folder into your archive of stuff from a studio, you can then go into the filename parser and just tick that studio and carry out the title renaming

Regards

Leopere:
My problem is there are various sources for various pieces of content and they all have their own structures by default so I can see something like this being reasonable at least if you could select and add regex patterns for the various subdirs.

bnkai:
Should be covered by #327.
Feel free to reopen otherwise.

PR #327, opened 2020-01-22
Allow path separator in scene filename parser pattern

WithoutPants:
Intended to address #302 

Currently, the scene filename parser matches the regex against the basename of the scene filename only. This PR changes so that if a path separator is included in the pattern (`/` or `\\` - due to regex escaping), then the pattern is matched against the full path of the scene filename, as stored in the database.

This allows patterns such as `{}/{studio}/{title}.{ext}`, for example, but also allows the scope of the search to be applied to specific folders: `{}/specific_folder/{title}.{ext}`

### DIFF
--- a/pkg/manager/filename_parser.go
+++ b/pkg/manager/filename_parser.go
@@ -335,9 +335,15 @@ func (h *sceneHolder) postParse() {
 }
 
 func (m parseMapper) parse(scene *models.Scene) *sceneHolder {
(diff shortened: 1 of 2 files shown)


Cutters1982:
This is great, thank you!

bnkai:
Looks ok, tests ok. Really useful if you have your collection in organised directories by studio or performer also.

PR #328, opened 2020-01-22
Accept random seed from UI for random sorting

### LINKED ISSUE #321: [Bug Report] Seed for Random sort only set on server launch
The seed for the Random sort is only set once, on server launch. 

**To Reproduce**
Use the Random scene sort mode. It will always return scenes in the same order for the life of the server process.

**Expected behavior**
The sort order is randomized every time the scene list is generated.

**Additional context**
From querybuilder_sql.go, it looks like randomSortFloat is set to a single random value outside of any function:

image

If it were set to a new random value inside of getSort(), perhaps it would work as expected.

WithoutPants:
Fixes #321 

Sets a random seed when sort by random is selected. Subsequent queries reuse the same seed so the order is maintained for a query, until a different sort by is chosen. When random is chosen once again, a new seed is generated and a new random order is created.

### DIFF
--- a/pkg/models/querybuilder_sql.go
+++ b/pkg/models/querybuilder_sql.go
@@ -93,6 +93,8 @@ func getSort(sort string, direction string, tableName string) string {
 		direction = "ASC"
 	}
 
+	const randomSeedPrefix = "random_"
+
 	if strings.Contains(sort, "_count") {
 		var relationTableName = strings.Split(sort, "_")[0] // TODO: pluralize?
 		colName := getColumn(relationTableName, "id")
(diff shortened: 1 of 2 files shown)


InfiniteTF:
Would it be difficult to write it to the query string so the seed doesn't get lost if the user accidentally refreshes?

WithoutPants:
It's included in the url that is sent to the server, so refreshing should maintain the same seed.

bnkai:
The seed is indeed in the url `http://localhost:9998/scenes?disp=0&p=1&sortby=random_98392816&sortdir=desc`. Refreshing the page or going to next page keeps the same url and thus also order.
Changing the sort selection and then choosing random creates a new one.
Works as intended for me.

PR #329, opened 2020-01-22
Use --frozen-lockfile in travis calls to yarn install

WithoutPants:
Updates the travis configuration to use `--frozen-lockfile` in the `yarn install` call, to ensure that node dependencies are deterministic across builds. I've also updated the README file to include the option in the `yarn install` instructions, and to update the http port to the correct one.


bnkai:
Good find, i never really noticed the issue. I thought that this was the default behaviour when a lock file was present but apparently the --frozen-lockfile is needed for that as you mentioned in the previous PR.

PR #334, opened 2020-01-25
Add O-counter

WithoutPants:
Adds an "o-counter". This is represented as a 💦 looking icon. When the counter is > 0, then it is included in the scene card:

image

On the scene details page, I have implemented a button for the counter. My default, it shows the icon and counter value. When clicked, this button increments the counter. If the button is hovered for 1 second, and the counter value > 0, then a popover menu is shown with decrement and reset items:

image

I need advice as to where to put this button in the scene details page. Below is a screenshot of possible placements:

image

To me the best here is probably on the tabs line. We could also right-align it:

image

Another option is overlaying the button on the scene player when the player is moused-over:

image

The button is not shown in the fullscreen view.

Let me know where you'd like it placed.

TODO:
- add filter for o-counter (<, >, !=, =)
- add button functionality to the scene card (so that you can click the o-counter button from the scene query page)
- add an operation on the scene query page to reset o-counter for selected scenes

### DIFF
--- a/graphql/documents/data/scene-slim.graphql
+++ b/graphql/documents/data/scene-slim.graphql
@@ -6,6 +6,7 @@ fragment SlimSceneData on Scene {
   url
   date
   rating
+  o_counter
   path
 
   file {
--- a/graphql/documents/data/scene.graphql
+++ b/graphql/documents/data/scene.graphql
@@ -6,6 +6,7 @@ fragment SceneData on Scene {
   url
   date
   rating
+  o_counter
   path
 
   file {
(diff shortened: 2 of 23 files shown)


bnkai:
Imho the tabs right align one has a cleaner look.
right align

WithoutPants:
I've added a filter and I think I'll leave it there for now. Performing other functions on the query page requires more structural changes, so I'll leave it for a future iteration.

WithoutPants:
Resolves #291

bnkai:
Works ok.
Just a note for the  filter. IS NULL or NOT NULL aren't usefull for the O-Counter since it's NOT NULL always  (0 by default)

PR #335, opened 2020-01-26
Fix to avoid creating generated folder inside data folder.

thundergreen:
This Fix is to avoid that ```generated``` folder will be created inside the movie /video files forder which leads to see the generated clips of the movies in SCENES.

### DIFF
--- a/docker/production/docker-compose.yml
+++ b/docker/production/docker-compose.yml
@@ -10,8 +10,8 @@ services:
     logging:
       driver: "json-file"
       options:
-        max-file: "10"
-        max-size: "200k"
+        max-file: "100"
+        max-size: "2000k"
     environment:
       - STASH_STASH=/data/
       - STASH_GENERATED=/generated/
@@ -20,12 +20,14 @@ services:
     volumes:
       - /etc/localtime:/etc/localtime:ro
       ## Keep configs here.
-      - ./config:/root/.stash
+      - /your/path/to/Stash/config:/root/.stash
       ## Point this at your collection.
-      - ./data:/data
+      - /your/path/to/your/files:/data
       ## This is where pre-generated transcodes live.
-      - ./transcodes:/transcodes
+      - /your/path/to/Stash/transcodes:/transcodes
       ## This is where your stash's metadata lives
-      - ./metadata:/metadata
+      - /your/path/to/Stash/metadata:/metadata
       ## Any other cache content.
-      - ./cache:/cache
+      - /your/path/to/Stash/cache:/cache
+      ## Generated
+      - /your/path/to/Stash/generated:/generated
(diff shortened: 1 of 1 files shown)


docker/production/docker-compose.yml
```diff
@@ -10,8 +10,8 @@ services:
     logging:
       driver: "json-file"
       options:
-        max-file: "10"
-        max-size: "200k"
+        max-file: "100"
+        max-size: "2000k"
```
bnkai: is there a reason why we need to increase the number of the log files? (they will get rotated anyway)
max-size: "2m" is more readable
also 
compress: "true" 
is an option

docker/production/docker-compose.yml
```diff
@@ -20,12 +20,14 @@ services:
     volumes:
       - /etc/localtime:/etc/localtime:ro
       ## Keep configs here.
-      - ./config:/root/.stash
+      - /your/path/to/Stash/config:/root/.stash
```
bnkai: maybe  
/path/to/your/stash/config  
instead? same for the rest

docker/production/docker-compose.yml
```diff
@@ -20,12 +20,14 @@ services:
...
       ## This is where pre-generated transcodes live.
-      - ./transcodes:/transcodes
+      - /your/path/to/Stash/transcodes:/transcodes
       ## This is where your stash's metadata lives
-      - ./metadata:/metadata
+      - /your/path/to/Stash/metadata:/metadata
       ## Any other cache content.
-      - ./cache:/cache
+      - /your/path/to/Stash/cache:/cache
+      ## Generated
```
bnkai: maybe 
## Where to store generated content (screenshots,previews,...)

bnkai:
I think we need to change also the docker image so that a default /generated/ path is created that now is missing , just to be safer.
The subdirs in the generated path seem to be created afterwards with the dev docker image not with the stable though.

docker/production/docker-compose.yml
Leopere: I’m fine with this I just shipped something small as to not fill up people’s machines with logs.

docker/production/docker-compose.yml
Leopere: I would prefer the relative path personally. This way we don’t inadvertently start filling people’s host file system with junk.

docker/production/docker-compose.yml
Leopere: Same thing here as my previous

Leopere:
I’m not sure if some of the suggested edits are really a good way to go for resolving these issues unfortunately I don’t like assuming that we can expect that the end user is going to tweak the docker compose we ship before they turn it on for the first time and I don’t want to start littering in peoples file systems.

bnkai:
As recommended by @Leopere edited back the volumes to be relative.
I also removed the transcodes volume as it wasn't used anyway. (We didn't set any enc variable for it and it is created  ok by default in the `generated` folder anyway)
Added an extra example comment to make it more clear hot to change the paths.

PR #336, opened 2020-01-28
Change thumbnail default size and resize algorithm

InfiniteTF:
Change thumbnail generation algorithm to 'Box' rather than NearestNeighbor, and change the default size from 100px to 200px.

Also added a size parameter to url query to allow for variable quality in galleries.

Box is about 25% slower than Nearest Neighbor by my measurements, but the quality is well worth it. A future enhancement might be to make the generation algo a user setting, since some users might prefer Lanczos or similar that offers sharper results.

Also, please excuse the format diff noise.


pkg/api/routes_gallery.go
```diff
@@ -27,9 +27,16 @@ func (rs galleryRoutes) File(w http.ResponseWriter, r *http.Request) {
 	thumb := r.URL.Query().Get("thumb")
 	w.Header().Add("Cache-Control", "max-age=604800000") // 1 Week
 	if thumb == "true" {
-		_, _ = w.Write(gallery.GetThumbnail(fileIndex))
-	} else {
+		_, _ = w.Write(gallery.GetThumbnail(fileIndex, 200))
+	} else if thumb == "" {
 		_, _ = w.Write(gallery.GetImage(fileIndex))
+	} else {
+		height, err := strconv.ParseInt(thumb, 0, 64)
```
bnkai: isn't this `width` actually instead of `height` ?

pkg/api/routes_gallery.go
```diff
@@ -27,9 +27,16 @@ func (rs galleryRoutes) File(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			http.Error(w, http.StatusText(400), 400)
+			return
+		}
+		_, _ = w.Write(gallery.GetThumbnail(fileIndex, int(height)))
```
bnkai: `width` instead of `height` ?

bnkai:
Tests ok , two possible semantic typos.

pkg/api/routes_gallery.go
InfiniteTF: Yup, forgot to update this file. Fixed.

PR #337, opened 2020-01-28
Add cache headers to static files and etags to performer images

InfiniteTF:
Adds one week cache headers to all files served from /static/*/*. All of the css/js assets are hashed by webpack, so there's no risk of a stale cache.

Also adds etags to performer images based on the md5 of the image. Since the tag is the hash it should also never go stale.

Both taken together shaves the load time of my `/performers`page from 1500 to ~500ms when the cache is hot.

### DIFF
--- a/pkg/api/routes_performer.go
+++ b/pkg/api/routes_performer.go
@@ -2,10 +2,13 @@ package api
 
 import (
 	"context"
+	"crypto/md5"
+	"fmt"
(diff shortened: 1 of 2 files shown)


bnkai:
It seems to work ok.
The difference in load time / payload is more evident  when the perfomer images are pngs / bigger in size.

PR #340, opened 2020-02-01
Fix paging not working

WithoutPants:
Quick fix of bug #326 that stopped paging from working. 

There is an outstanding bug in that the page number is supposed to also be maintained and persisted, but this is currently not working. I'll get to that later.


WithoutPants:
Changed fix - prevent rendering of pagination and content while loading. This will prevent the pagination control from screwing up the page numbers while in loading state. Should address both the paging issue and the page number persistence issue.

Issue #339, opened 2020-01-30
[RFC] v0.1 release


WithoutPants:
We've got some pretty large changes in the pipeline, and I'd like to do a formal v0.1 release before we merge those in.

### Pending PRs to include

- [x] #326
- [x] #333
- [x] #335 
- [x] #336
- [x] #337 
- [x] #350 

### Other required tasks

- [x] Update travis configuration to _not_ build the master branch
- [x] Merge develop into master
- [x] Manually tag v0.1 and create release
- [x] Merge master back into develop

InfiniteTF:
Good plan. #336 and #337 are good to merge unless there are any complaints.

bnkai:
I would also add _update check version functionality_ to the required tasks.
Currently it relies on the hardcoded version and the fact that we have single tag for stable/dev builds.
It could also be useful if we can decide on a naming scheme for the version.

333,336,337 are ok by me also
326 i need to make a quick test for the latest commit
335 it need a little feedback . but is necessary to include

WithoutPants:
> It could also be useful if we can decide on a naming scheme for the version.

Based on semver naming, is there anything wrong with v0.1?

WithoutPants:
Propose the following changes before we release:

- change the current `v0.0.0-alpha_dev` tag to `latest` instead. This tag represents the latest version on develop and should be documented as such.
- travis will tag the current commit with `latest` when on the develop branch. It will not automatically tag commits in the master branch. This should mean that the master branch will only be released when already tagged.
- use `git describe --tags --exclude latest` to generate the `STASH_VERSION` environment variable - this will not be used to tag the release, see above. git describe will show the current tag if the commit has one, or the latest tag (excluding `latest`) with a suffix showing the number of commits since the last tag, and the commit hash. For example, running `git describe --tags --exclude *_dev`, gives the result `v0.0.0-alpha-252-g066295f`. This gives us a decent version string to deal with for the `latest` builds. The `STASH_VERSION` generation can be moved out of travis and into the makefile instead.
- `latest` can be have `prerelease` set to true. The master builds should have it set to false.
- In order for `git describe` to work correctly, the tagged release commit must exist on both master and develop branches. I think this will mean we need to fast-forward merge to master, and hotfixes to releases will need to be commit-merged to the develop branch from master. Someone with better git experience might need to help me out here.
[...]

PR #345, opened 2020-02-04
Prepare for 0.1.0 release

WithoutPants:
Version is now generated in the Makefile and `cross-compile.sh` script, and uses `git describe --tags --exclude latest_develop` to generate a version string. From the `git describe` help page:

> If the tag points to the commit, then only the tag is shown. Otherwise, it suffixes the tag name with the number of additional commits on top of the tagged object and the abbreviated object name of the most recent commit.

The latest develop release is now tagged with `latest_develop`. It will be named for the version string generated above.

Builds on the `master` branch will now only be released if the commit is tagged with a tag that is not `latest_develop`. Builds on this branch will be done as a `draft` allowing maintainers to confirm the release before making it publicly available. It also allows changelogs to be filled in as needed.

Changed the check version code so that it determines whether it is a development build by checking the version string. If the version string contains the generated suffixes, then it must be a development build. For development builds, it now calls the github API `/repos/stashapp/stash/releases/tags/latest_develop` endpoint, which returns the release for the `latest_develop` tag. For release builds, it now calls the github API `/repos/stashapp/stash/releases/latest` endpoint, which returns the most recent full release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ deploy:
     skip_cleanup: true
     overwrite: true
     name: "${STASH_VERSION}: Latest development build"
-    body: This is always the latest committed version on the develop branch. Use as your own risk!
+    body: ${RELEASE_DATE}\n This is always the latest committed version on the develop branch. Use as your own risk!
     prerelease: true
     on:
       repo: stashapp/stash
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,15 +18,12 @@ script:
 #- make vet
 - make it
 after_success:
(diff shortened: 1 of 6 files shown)


WithoutPants:
Related to #339

WithoutPants:
I've also added a two minute limit to the transfer.sh upload call so that slow upload speeds don't kill the build.

bnkai:
Most changes here depend on Travis and the new tag you created. 
From what i checked they look ok but we can't really test unless travis provides a binary and the new tag gets created...

PR #350, opened 2020-02-05
Fix URL update not reflected in list hook

### LINKED ISSUE #348: [Bug Report] Performer/Tag links in scenes page do not work
**Describe the bug**
Clicking the link anchor for a tag or performer should load a query with that tag/performer. It currently does not. This looks like a regression due to #326 

**To Reproduce**
Go to the scenes page.
Mouse over the tags or performers for any scene, then click on the name anchor link. Nothing happens. 

**Expected behavior**
The page should be updated with a filter on the selected performer/tag.

WithoutPants:
Fixes #348 

Refactored ListHook again so that it initialises from interface forage once before the rest of the component is even rendered. When the forage is loaded, it initialises the filter from the saved value if and only if the search parameters on the URL are not present. Once this has been done, any updates to the current URL are applied to the filter.

Tests performed:

- navigating to other screens then back to the scenes page maintains the same query filter
- clicking on the tag/performer link correctly updates the filter and results
- navigating to other pages in the results works as expected

### DIFF
--- a/ui/v2/src/hooks/ListHook.tsx
+++ b/ui/v2/src/hooks/ListHook.tsx
@@ -95,6 +95,15 @@ export interface IListHookOptions {
   renderSelectedOptions?: (result: QueryHookResult<any, any>, selectedIds: Set<string>) => JSX.Element | undefined;
 }
 
+function updateFromQueryString(queryStr: string, setFilter: (value: React.SetStateAction<ListFilterModel>) => void) {
+  const queryParams = queryString.parse(queryStr);
+  setFilter((f) => {
+    const newFilter = _.cloneDeep(f);
+    newFilter.configureFromQueryParameters(queryParams);
+    return newFilter;
+  });
+}
+
 export class ListHook {
   public static useList(options: IListHookOptions): IListHookData {
     const [filter, setFilter] = useState<ListFilterModel>(new ListFilterModel(options.filterMode));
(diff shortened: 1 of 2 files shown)


bnkai:
Above work ok for me also.
Setting different filters for scenes/performers/studios , closing browser reopening works fine also

**A minor glitch**
When clicking on the performer link (thumbnail) the filter is updated but not with the current sorting order but with the default instead
_Before cliking the thumbnail_
`http://localhost:9998/scenes?disp=0&p=1&perPage=40&sortby=created_at&sortdir=desc`
_After (edited and cleaned)_
`http://localhost:9998/scenes?c={ type : performers , value :[{ id : 66 , label : Performer Name }], modifier : INCLUDES_ALL }&disp=0&p=1&perPage=40&sortby=date&sortdir=asc`
I would expect current `sortby` and `sortdir`  to be kept (`created_at` and `desc`)

WithoutPants:
I was going to leave that issue for a later date, but I came up with a reasonable solution that I hope doesn't cause any other regressions. The solution was to change the links to exclude items per page, sort by, sort direction and display mode. Then when the list hook is initialised, if the URL is set and those items are not present in the URL, then they are initialised from the local forage instead. 

The only thing I have tested with this latest commit is the paging, since I don't have sufficient data on the machine I'm on.

bnkai:
Seems to work ok to me.

PR #355, opened 2020-02-06
Make performer image contain

### LINKED ISSUE #349: [Bug Report] Performer images are cut off in the performers page
**Describe the bug**
Performer images are not resized according to the bounding box of the performer card. The image is centred meaning that the top of the image is often cut off. 

**To Reproduce**

I've used this image, which I'm aggressively cropped to demonstrate the issue (2.34:6).

image

Set this as the image for a performer, then view the performers page. The performer image is shown like this:

image

**Workaround**

This custom css fragment crops from the bottom of the image, rather than the top:

```
a.performer.image {
 background-position: top !important;
}
```

**Expected behavior**

The image should be resized to fit within the boundaries of the performer card, with no cropping.

WithoutPants:
Fixes #349 

Simple CSS fix to changed performer image to be contain instead of cover.

Without fix:

image

With fix:

image

### DIFF
--- a/ui/v2/src/index.scss
+++ b/ui/v2/src/index.scss
@@ -359,7 +359,7 @@ span.block {
 .performer.image {
   height: 50vh;
   min-height: 400px;
-  background-size: cover !important;
+  background-size: contain !important;
   background-position: center !important;
   background-repeat: no-repeat !important;
 }


WithoutPants:
Reverted due to incorrect merge.

PR #357, opened 2020-02-09
Add new v2.5 UI

InfiniteTF:
New UI based on a refactor of the existing v2 UI.

**The user facing changes are:**
* Replacing blueprintjs with react-bootstrap. Not a very sexy framework, but it's well maintained, and has a fully responsive layout. Additionally, the select dropdowns have been replaced with react-select.
* Responsive mobile layout. This is still somewhat WIP, particularly the editing pages and settings pages do not work great, but browsing performers/scenes/studios should work well.
* Added react-intl to serve as an i18n framework. For the time being only main menu and the stats page has been translated into (bad) German, to serve as an example. The language can be changed in the interface config. To add new text, wrap it in `<FormattedMessage />` and run `yarn extract` to populate the message files in `src/locale`. To support this all text should be wrapped in `<FormattedMessage>` or `formatMessage`, all dates wraped inn `<FormattedDate>`, and all numbers in `<FormattedNumber>`. For more info see the [react-intl documentation](https://github.com/formatjs/react-intl/blob/master/docs/Components.md). New languages can be added by editing the `yarn extract` command and the dropdown in the config menu, and then editing the resulting [language-code].json in the locale folder.
* Tweaked performer and studio page layouts.

**There have also been a large amount of behind the scenes changes:**
* Fixed all linting errors. This also involved refactoring many of the components to remove or clean up `useEffect` hooks which are prone to create confusing and brittle code.
* Added additional linting rules based on airbnb-typescript, as well as stylelint rules, to ensure that the style is more unified and we avoid verbose or potentially buggy code.
* Disallowed `any` in the linter, and rewrote or exempted all uses of it.
* Ran the entire project through [prettier](https://prettier.io/), which formats all whitespace according to a strict set of rules. This can be rerun on demand with `yarn format`.
* Added 'src/' as an import root, and switched all imports to use `src/component|hooks|utils/Name/File`. This makes it easier to see where imports and coming from, and means they won't break when copy-pasted, or the file is moved.
* Moved all component-specific styles to the individual components so that it's easier to see where styles belong, and what they should affect. Also, went through all styles and did a general cleanup.
* Upgraded typescript to 3.7. This supports [optional chaining and nullish coalescing](https://blog.logrocket.com/optional-chaining-and-nullish-coalescing-in-javascript/) which is extremely handy when dealing with graphql data.
* Ugpraded apollo-client and all related libraries to the latest version, as well as graphql-codegen. The latter involved some changes to the calling signature of the queries/mutations, as well as the type names, but nothing major.
* Upgraded react-router and switched to using useHistory, useParams, and useLocation hooks rather than passing router props around.
* Upgraded all other libraries to the latest version as well, except for react-images which is inexplicably very over-complicated in the latest version.
* Renamed all component folders to be uppercase, and moved some of them down since they are more or less standalone, specifically ScenePlayer and SceneFilenameParser.
* Probably way more stuff that I can't remember right now.

**Some things to note:**
* The code is up to date with the master develop branch. Some of the more recent commits had to be rewritten to work, but should otherwise be fine.
* Outside of the `ui/v2.5` folder the only change is adding the language config setting to graphql. This shouldn't affect the v2 branch at all, and they can in fact be run side by side without problem as long as v2.5 is manually started in dev mode. To switch release version the packr2 box has to be changed from v2 to v2.5, and stash rebuilt.
* All icons are now from the font-awesome project. To add an icon, simply use the shared `<Icon>` component and specify an icon from [fontawesome](https://fontawesome.com/icons?d=gallery) with the icon prop.
* There are bugs. A few I know of, but there are probably many more I don't.


bnkai:
Looks awsome. It may not be "sexy" but i think it's more user friendly especially the mobile version.

A few things i encountered while "playing" with it.

- the status popup is not readable (it's light grey while the font is white and you can't make out anything)

- the ethnicity selector in the performers details panel is active even when we are in view mode

- in the scene tab, the tag/performer "thumbnails" have a `&sortby=date` to their link that changes the current sort order

- the filters/sort order/.. nav bar in the scenes section of the performer page doesn't work properly

- when you are in `mobile` mode the `stash` home at the nav bar is kinda of invisible because it misses an icon. You can click on it and see the stats though.

- the german de.json is missing the `markers` part ( i know it is just an example :-) )
Also in the options part of the language selector i think languages should be presented in their native name. German should be Deutsch for example.

InfiniteTF:
@bnkai Thanks for the feedback. I've rewritten the list filter hooks so it behaves the same as v2, and added a hamburger menu in mobile mode. Also fixed the other issues you mentioned.

ui/v2.5/src/components/Scenes/SceneCard.tsx
```diff
@@ -0,0 +1,248 @@
...
+          loop
+          className={cx("scene-card-video", { portrait: isPortrait() })}
+          poster={props.scene.paths.screenshot || ""}
+          ref={videoHoverHook.videoEl}
+        >
+          {previewPath ? <source src={previewPath} /> : ""}
+        </video>
+      </Link>
+      <div className="card-section">
+        <h5 className="text-truncate">
```
WithoutPants: The old version didn't truncate the title text. Is there a reason why we're truncating it now?

pkg/api/resolver_query_configuration.go
```diff
@@ -66,5 +68,6 @@ func makeConfigInterfaceResult() *models.ConfigInterfaceResult {
 		ShowStudioAsText:    &showStudioAsText,
 		CSS:                 &css,
 		CSSEnabled:          &cssEnabled,
+        Language:                  &language,
```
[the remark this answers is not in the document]
InfiniteTF: Not sure why `go fmt` doesn't auto correct this, but fixed now at least.

ui/v2.5/src/components/Scenes/SceneCard.tsx
InfiniteTF: v2 truncates in the same way: https://github.com/stashapp/stash/blob/develop/ui/v2/src/components/scenes/SceneCard.tsx#L240

WithoutPants: The v2 is missing `white-space: nowrap`, which prevents the truncation (and rendering the other two styles superfluous). So v2.5 successfully truncates, while v2 does not. In my opinion, not truncating is better.

InfiniteTF: Ah, I think I see what's happening. Filenames without whitespace are being truncated in both versions, but titles with whitespace are only being truncated in 2.5. I was just testing with scene titles. I'll remove the whitespace no-wrap.

InfiniteTF:
> Just a couple of minor changes.
> 
> ....
> 
> I think if we can get the bugs ironed out, we should get this merged and sort out the move from v2 in a separate PR, once the two versions are fully aligned.

All the issues you pointed out should be resolved now.

bnkai:
Looks great.
A couple of things i found.

- the loading animation is a bit unpredictable. On a new database i tried  with only scenes and no performers/studios/markers/galleries when you switch from the scenes to the studios or performers/markers the loading animation keeps going without stopping. 

- on the mobile view as soon as you press on the galleries menu , the menu shows as empty. You have to press above or below the galleries so that it gets visible again.

PR #360, opened 2020-02-11
Fix database locked on image cover too big bug (#359)

### LINKED ISSUE #359: [Bug Report] Cover images dont get updated when they are too big
**Describe the bug**
When updating a cover it seems that if the filesize is big enough something goes wrong
The error reported is this one `Error: GraphQL error: database is locked`
The response for example `{"errors":[{"message":"database is locked","path":["sceneUpdate"]}],"data":{"sceneUpdate":null}}`

**To Reproduce**
1. Go to a random scene
2. Click on edit
3. Either copy/paste a big image or use the sample below with browse image for the cover
4. Try to save
5. The only error reported is from graphql: `Error: GraphQL error: database is locked`. No error is reported in the console

**Expected behavior**
The cover should be updated to new image

**Sample**
should be 6mb+
4k

**Desktop:**
 - OS: Linux
 - Browser Chrome, Firefox 
 - Version 79.0.3945.88, 69.0.2

Firefox seems to cope a little better , needs a little bigger filesize to choke.

**Additional Comments**
Using the same image as a performer / studio image works ok.

bnkai:
Fixes #359
It seems to me that when  `scene, err := qb.Update(updatedScene, tx)` is called
it returns while  a lock is still active.
With big (filesize)  cover images the lock is still active till we reach to the call to `scene, err := qb.Find(sceneID)` and thus we get an error of database locked.
`scene` isn't actually needed  to be "found" anyway because we already have it from the call to `qb.Update` above.

### DIFF
--- a/pkg/api/resolver_mutation_scene.go
+++ b/pkg/api/resolver_mutation_scene.go
@@ -163,10 +163,6 @@ func (r *mutationResolver) sceneUpdate(input models.SceneUpdateInput, tx *sqlx.T
 
 	// only update the cover image if provided and everything else was successful
 	if coverImageData != nil {
-		scene, err := qb.Find(sceneID)
-		if err != nil {
-			return nil, err
-		}
 
 		err = manager.SetSceneScreenshot(scene.Checksum, coverImageData)
 		if err != nil {


WithoutPants:
I believe this occurred because `Find` doesn't accept a transaction object and creates a new one. Since it is within a mutation operation, the transaction should be passed in, and @bnkai verified that this also resolves the original bug. In future, I think we should add a `tx` parameter to all `Find` functions, setting it to nil where it is not needed (ie outside of transactions).

PR #364, opened 2020-02-12
Dockerfiles for production, develop and build

### LINKED ISSUE #353: [Feature] Better Production containers for a stable release.
* [x] Multi Stage Docker builds
* [ ] Every version of the Docker containers should be compile from source so people can take any commit from Version 0.1.0 forward and have all of the tools required to produce binaries for testing locally. 
* [ ] Documentation for Development
* [ ] Security updates to [user privileges](https://github.com/stashapp/stash/blob/develop/docker/production/x86_64/Dockerfile#L20-L23) of the production container.  This may benefit from a permissions assertion binary that asserts permissions for specific folders for the container on container run.  [Docker run lines for a perms assertion app](https://github.com/lbryio/lbry-docker/blob/fced92808905e6c23110d01c63df38d07290d5e2/lbrycrd/Dockerfile-linux-x86_64-production#L4-L16) [c binary for asserting perms in a static directory path](https://github.com/lbryio/lbry-docker/blob/fced92808905e6c23110d01c63df38d07290d5e2/lbrycrd/stuff/fix-permissions.c) this is how it's [invoked](https://github.com/lbryio/lbry-docker/blob/fced92808905e6c23110d01c63df38d07290d5e2/lbrycrd/stuff/start.sh#L52-L53)
* [ ] Production and Dev containers need reworked totally.

I would like to have some time to write this and test a multi-stage build Dockerfile for production and dev that basically both do the same thing.  Perhaps pulls the compiler image from the upstream but have the option to pull a local one.

Open to suggestions. 

**Additional context**
I would like to keep transparency as best as possible everywhere so logs being where people expect them and keep the technical debt down to avoid maintenance hell.

Blocks #339

WithoutPants:
Resolves #353 

Changed production dockerfile to get the latest release (which won't work until we perform an actual release).

Added new develop dockerfile which is the same as the production one, but grabs the `latest_develop` tagged release.

Added `--http1.1` because I was getting `curl: (16) Error in the HTTP2 framing layer` when downloading ffmpeg in the docker build.

Added new build dockerfile. This is a two-stage dockerfile that builds the minimum compiler image to build with (since the compiler image is overkill and very large), builds stash, then in the second stage does the same thing as the other dockerfiles do.

### DIFF
--- a/docker/build/x86_64/Dockerfile
+++ b/docker/build/x86_64/Dockerfile
@@ -46,9 +46,7 @@ RUN make build
 
 FROM ubuntu:19.10 as app
 
-RUN apt-get update && \
-    apt-get -y install ca-certificates && \
-    adduser stash --gecos GECOS --shell /bin/bash --disabled-password --home /home/stash
+RUN apt-get update && apt-get -y install ca-certificates
 COPY --from=compiler /stash/stash /ffmpeg/ffmpeg /ffmpeg/ffprobe /usr/bin/
 
 EXPOSE 9999
--- a/docker/build/x86_64/Dockerfile
+++ b/docker/build/x86_64/Dockerfile
@@ -2,7 +2,7 @@
 # ie from top=level stash:
 # docker build -t stash/build -f docker/build/x86_64/Dockerfile .
 
-FROM golang:1.11.5 as compiler
+FROM golang:1.11.13 as compiler
 
 RUN apt-get update && apt-get install -y apt-transport-https
 RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
@@ -44,12 +44,12 @@ RUN make generate
 RUN make ui 
 RUN make build
 
-FROM ubuntu:18.04 as app
+FROM ubuntu:19.10 as app
 
 RUN apt-get update && \
     apt-get -y install ca-certificates && \
     adduser stash --gecos GECOS --shell /bin/bash --disabled-password --home /home/stash
-COPY --from=compiler /stash /ffmpeg/ffmpeg /ffmpeg/ffprobe /usr/bin/
+COPY --from=compiler /stash/stash /ffmpeg/ffmpeg /ffmpeg/ffprobe /usr/bin/
 
 EXPOSE 9999
 CMD ["stash"]
--- a/docker/build/x86_64/Dockerfile
+++ b/docker/build/x86_64/Dockerfile
@@ -1,0 +1,57 @@
+# this dockerfile must be built from the top-level stash directory
+# ie from top=level stash:
+# docker build -t stash/build -f docker/build/x86_64/Dockerfile .
+
+FROM golang:1.11.5 as compiler
+
+RUN apt-get update && apt-get install -y apt-transport-https
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+
+RUN apt-get update && \
+    apt-get install -y nodejs yarn xz-utils --no-install-recommends || exit 1; \
+	rm -rf /var/lib/apt/lists/*;
+	
+ENV PACKR2_VERSION=2.0.2
+ENV PACKR2_SHA=f95ff4c96d7a28813220df030ad91700b8464fe292ab3e1dc9582305c2a338d2
+ENV PACKR2_DOWNLOAD_FILE=packr_${PACKR2_VERSION}_linux_amd64.tar.gz
+ENV PACKR2_DOWNLOAD_URL=https://github.com/gobuffalo/packr/releases/download/v${PACKR2_VERSION}/${PACKR2_DOWNLOAD_FILE}
+
+WORKDIR /
+RUN wget ${PACKR2_DOWNLOAD_URL}; \
+	echo "$PACKR2_SHA $PACKR2_DOWNLOAD_FILE" | sha256sum -c - || exit 1; \
+	tar -xzf $PACKR2_DOWNLOAD_FILE -C /usr/bin/ packr2; \
+	rm $PACKR2_DOWNLOAD_FILE;
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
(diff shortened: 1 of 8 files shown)


WithoutPants:
WIP while I fix the following issue with the build dockerfile:

```
error: unknown option `exclude'
usage: git describe [<options>] [<commit-ish>...]
   or: git describe [<options>] --dirty

    --contains            find the tag that comes after the commit
    --debug               debug search strategy on stderr
    --all                 use any ref
    --tags                use any tag, even unannotated
    --long                always use long format
    --first-parent        only follow first parent
    --abbrev[=<n>]        use <n> digits to display SHA-1s
    --exact-match         only output exact matches
    --candidates <n>      consider <n> most recent tags (default: 10)
    --match <pattern>     only consider tags matching <pattern>
    --always              show abbreviated commit object as fallback
    --dirty[=<mark>]      append <mark> on dirty working tree (default: "-dirty")
```

WithoutPants:
Updated golang image to the latest fixed the issue with git.

bnkai:
They look ok to me.
We should probably update the documentation also.
For the new `build` docker file the use case would be that a user clones the repo, makes the changes he wants or checkouts a specific tag/commit and then builds the customized docker image , right ?

docker/build/x86_64/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+RUN make generate 
+RUN make ui 
+RUN make build
+
+FROM ubuntu:18.04 as app
+
+RUN apt-get update && \
+    apt-get -y install ca-certificates && \
+    adduser stash --gecos GECOS --shell /bin/bash --disabled-password --home /home/stash
+COPY --from=compiler /stash /ffmpeg/ffmpeg /ffmpeg/ffprobe /usr/bin/
```
bnkai: should be
`COPY --from=compiler /stash/stash /ffmpeg/ffmpeg /ffmpeg/ffprobe /usr/bin/`

docker/build/x86_64/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+RUN yarn --cwd ui/v2 install --frozen-lockfile
+
+COPY . /stash/
+ENV GO111MODULE=on
+
+RUN make generate 
+RUN make ui 
+RUN make build
+
+FROM ubuntu:18.04 as app
```
bnkai: with the change to go 1.11.13 as the compiler (from 1.11.5) we need glibc at least 2.28
changing to `FROM ubuntu:19.10 as app` covers that

docker/build/x86_64/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+RUN make ui 
+RUN make build
+
+FROM ubuntu:19.10 as app
+
+RUN apt-get update && \
+    apt-get -y install ca-certificates && \
+    adduser stash --gecos GECOS --shell /bin/bash --disabled-password --home /home/stash
+COPY --from=compiler /stash/stash /ffmpeg/ffmpeg /ffmpeg/ffprobe /usr/bin/
+
```
Leopere: There should be a `USER` directive here declaring that the app itself should execute as the unprivileged user otherwise we just create the user for no reason.

The only issue this creates is file permissions corrections so if we can live with support requests for that its good to add the USER directive.

docker/develop/x86_64/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    tar xf /ffmpeg.tar.xz && \
+    rm ffmpeg.tar.xz && \
+    mv /ffmpeg*/ /ffmpeg/
+
+FROM ubuntu:18.04 as app
+RUN apt-get update && \
+    apt-get -y install ca-certificates && \
+    adduser stash --gecos GECOS --shell /bin/bash --disabled-password --home /home/stash
+COPY --from=prep /stash /ffmpeg/ffmpeg /ffmpeg/ffprobe /usr/bin/
+EXPOSE 9999
```
Leopere: USER directive required to implement the additional user thats added.  Same issues will persist from previous comment.

docker/build/x86_64/Dockerfile
WithoutPants: This is copied from the production dockerfile, so is an existing issue. During testing, adding the `USER` directive resulted in a lot of permissions issues that I was unable to workaround. 

Having done some reading on this, this looks like a non-trivial issue that I'd prefer to push out to another issue and PR. 

[This stackoverflow answer](https://stackoverflow.com/a/39398511) suggests one solution using `ENTRYPOINT` to run the container as root, set the file permissions, then run stash as the stash user. Changing this around will also impact the `docker-compose` file.

Alternatively, I can remove the `adduser` from all of the docker files, but I'd rather keep it in there and fix the container user issue separately.

Leopere: I added it long before it was looking like anyone was really going to keep up on maintaining this and haven't really touched it since.  I'd suggest just remove it, for now, we can re-add it later.  

The only other thing we could do is have a tiny fix permissions go binary that would execute as you said a file permissions change on a static set of paths such that nobody can hijack or escalate anything. 

This could be run on container start with a `start.sh` or the `entrypoint.sh` really either is fine.   The biggest concern that I have is just anything that is done by the restricted user should be just that restricted.  

```c
#include <unistd.h>
int main() {
  // This program needs to run with setuid == root
  // This needs to be in a compiled language because you cannot setuid bash scripts
  setuid(0);
  execle("/bin/bash", "bash", "-c",
         "/bin/chown -R stash:stash /data && /bin/chmod -R 755 /data/",
         (char*) NULL, (char*) NULL);
}
```
This above little c code was used on another container I had worked on before to let the user summon the `chmod -R 755` via root level permissions except as the unpriv'd user.  I think that this method would be potentially ideal.  However it would likely be smarter if it handled a number of directories common to Docker but would be written in Go and be stored in another repository.

Leopere: How it was implemented looked like https://github.com/lbryio/lbry-docker/blob/fced92808905e6c23110d01c63df38d07290d5e2/lbrycrd/Dockerfile-linux-x86_64-production#L16

WithoutPants:
Removed `adduser` from the docker files.

bnkai:
The "adduser" was a placeholder anyway as it was now.
Also as stated the use case for the "build" docker is such that it must not be tracked/build in dockerhub. A user can either create an image from either the current source code or his custom code or specific commit depending on the status of the code in the directory he is using.

Issue #368, opened 2020-02-13
[Bug Report] BaseURL is set to reverse proxy IP, not web client's


twotobinary:
**Describe the bug**
I'm using stash on my Kubernetes cluster with Istio mTLS enabled and [Pomerium](https://www.pomerium.io/) as my SSO provider which acts as a reverse proxy so this results in following access:
1. `https://<external stash domain>`
2. Istio routes to pomerium service within kubernetes cluster -> `pomerium-proxy.namespace.svc.cluster.local:443`
3. Pomerium routes to stash service -> `stash.namespace.svc.cluster.local:9999`
4. Pomerium sets baseURL to `stash.namespace.svc.cluster.local:9999` when it should be my external stash domain.

**To Reproduce**
Access stash from a reverse proxy and the host will be improperly set at [BaseURLMiddleware](https://github.com/stashapp/stash/blob/cd9c6e998ef7b51295f48963222d36c60ca52040/pkg/api/server.go#L299).

It would be nice if we can override this value with some configuration value such as `external_host: https://external.host.com`

**Expected behavior**
I expect to see my base URL set to the address my client is accessing stash from.

As a workaround, I can still stream all files as long as I manually set the domain to the external domain.

WithoutPants:
Closed via #369

PR #369, opened 2020-02-13
fix(middleware): set baseURL to external host if provided

twotobinary:
Added a fix for issue with reverse proxy: https://github.com/stashapp/stash/issues/368

To test this, I added `external_host: https://my.domain.com` to the config yml and created a docker image. Paths no longer contain the internal address that is inaccessible by the client.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -296,6 +296,11 @@ func BaseURLMiddleware(next http.Handler) http.Handler {
 		}
 		baseURL := scheme + "://" + r.Host
 
+		externalHost := config.GetExternalHost()
+		if externalHost != "" {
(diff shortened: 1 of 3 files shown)


WithoutPants:
Code looks fine.

bnkai:
Looks ok to me also.
@twotobinary  did you test also the ENV part ? Instead of changing the config adding the STASH_EXTERNAL_HOST env to the compose file for your docker test? (should be ok though)

twotobinary:
@bnkai I just finished testing by adding env variable: `STASH_EXTERNAL_HOST` (and removed from config) and it works as expected. I also did a quick test without the variable to make sure original behavior was restored.

PR #333, opened 2020-01-24
Add Xpath post processing and performer name query

WithoutPants:
Note: by necessity, this PR supercedes and extends #332. I can close the other PR, or just rebase this one when it is merged (ideally the latter).

This change adds some post-processing functionality to the xpath scraping configuration.

In an xpath scraper, a field value may now be either a string xpath selector value, or a sub-object. 

If it is a sub-object, it must contain the `selector` field, which has the xpath selector value. Within the sub-object, further fields are available to perform post-processing:

`concat`: if an xpath matches multiple elements, and `concat` is present, then all of the elements will be concatenated together
`replace`: contains an array of sub-objects. Each sub-object must have a `regex` and `with` field. The `regex` field is the regex pattern to replace, and `with` is the string to replace it with. `$` is used to reference capture groups - `$1` is the first capture group, `$2` the second and so on. Replacements are performed in order of the array.
`parseDate`: if present, the value is the date format using go's reference date (2006-01-02). For example, if an example date was `14-Mar-2003`, then the date format would be `02-Jan-2006`. See the [time.Parse documentation](https://golang.org/pkg/time/#Parse) for details. When present, the scraper will convert the input string into a date, then convert it to the string format used by stash (`YYYY-MM-DD`).

Post-processing is done in order of the fields above - `concat`, then `regex`, then `parseDate`.

Below are two example scrapers that will hopefully illustrate these concepts.

This Boobpedia scraper illustrates the `concat` and `parseDate` operations:
```
name: Boobpedia
performerByURL:
  - action: scrapeXPath
    url: 
      - boobpedia.com/boobs/
    scraper: performerScraper

xPathScrapers:
  performerScraper:
    performer:
      Name: //h1
      URL: //table//tr/td//b/a[text()='Official website']/@href
      Twitter: //table//tr/td/b/a[text()='Twitter']/@href
      Instagram: //table//tr/td/b/a[text()='Instagram']/@href
      Birthdate:
        selector: //table//tr/td//b[text()='Born:']/../following-sibling::td/a
        # two elements - concatenate together
        concat: " "
        parseDate: January 2 2006
      Ethnicity: //table//tr/td/b[text()='Ethnicity:']/../following-sibling::td/a
      Country: //table//tr/td/b[text()='Nationality:']/../following-sibling::td/a
      EyeColor: //table//tr/td/b[text()='Eye color:']/../following-sibling::td/a
      Height: //table//tr/td/b[text()='Height:']/../following-sibling::td
[...]

PR #370, opened 2020-02-13
Add image scraping support

### LINKED ISSUE #344: [Feature] Add image support for scrapers
The performer scrapers should support scraping an image to populate the performer image. 

The scene scraper should support scraping and setting a cover image.

WithoutPants:
Resolves #344 

Adds the ability to scrape performer images and scene cover images.

This change also introduces the `subScraper` xpath post-processing option. If `subScraper` appears in an attribute xpath configuration, then the sub-scraper will be executed after all other post-processes are complete. It then takes the value and performs an http request, using the value as the URL. Within the `subScraper` config is a nested scraping configuration. This allows you to traverse to other webpages to get the attribute value you are after.

For example, from the Boobpedia scraper config in #333 :

```
...
performerScraper:
  performer:
    # ..snip..
    Image:
      selector: //table[@class="infobox"]//tr[2]//a/@href
      # URL is a partial url, add the first part
      replace:
        - regex: ^
          with: http://www.boobpedia.com
      subScraper:
        selector: //div[@class="fullImageLink"]/a/@href
        replace:
          - regex: ^
            with: http://www.boobpedia.com
```

This fragment gets the URL from the xpath `//table[@class="infobox"]//tr[2]//a/@href`, adds the `http://www.boobpedia.com` prefix with the `replace` post-process. Then the sub-scraper post-process is run. It requests the document from the resulting URL, then gets the URL from `//div[@class="fullImageLink"]/a/@href` of the resulting page, followed by the replace post-process.

The `Image` value is expected to be a URL itself, which the system will subsequent request and encode.

Also adds image scraping to the stash scraper.

### DIFF
--- a/pkg/scraper/image.go
+++ b/pkg/scraper/image.go
@@ -4,10 +4,15 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/stashapp/stash/pkg/models"
 	"github.com/stashapp/stash/pkg/utils"
 )
+
+// Timeout to get the image. Includes transfer time. May want to make this
+// configurable at some point.
+const imageGetTimeout = time.Second * 30
 
 func setPerformerImage(p *models.ScrapedPerformer) error {
 	if p == nil || p.Image == nil || !strings.HasPrefix(*p.Image, "http") {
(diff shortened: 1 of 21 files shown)


pkg/scraper/image.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	s.Image = img
+
+	return nil
+}
+
+func getImage(url string) (*string, error) {
+	// assume is a URL for now
+	resp, err := http.Get(url)
```
bnkai: should we use a new http client with a default appropriate timeout instead of a plain get ?

bnkai:
Tests ok with me.
Images are fetched ok and saved with the boobpedia demo scraper.

WithoutPants:
Rebased and ported UI changes to 2.5. @bnkai can you please review on v2 and v2.5 of the UI?

MrX292:
https://github.com/stashapp/CommunityScrapers/pull/2 some scrapers for it

PR #371, opened 2020-02-14
Add gender support for performer

hiddenpants255:
I have added "gender" field in "performers" table and update the UI to allow gender selection.


hiddenpants255:
I was branched from wrong commit. Reversing b6d722a, which is for different PR.

bnkai:
From a quick look at the code it seems to me that the import/export functionality (task_import.go task_export.go ) isn't updated to reflect to the addition of the `gender` attribute.

WithoutPants:
I'd prefer to have the gender be an enum of possible values (like in https://github.com/stashapp/stash-box/blob/stash_prototype/graphql/schema/types/performer.graphql) rather than a free-text field.

yoshnopa:
Can I do anything to help on this one?
The import and export are easy enough and I can add them, but I don't quite understand what about the enum should be done. There are no enums in the current stash code in that regard (for example, compare the code for ethnicity, which should work in a similar if not identical fashion). Instead, the code you show is from another project. Could you maybe explain @WithoutPants

WithoutPants:
The stashbox code I linked is adapted from the same codebase as stash, so shouldn't be too difficult to port over. You'll need to add the gender enum and change the gender types of the Performer type and the inputs, as well as the filter. The stashbox code should also give an indication as to how the translation works between enum and string.

WithoutPants:
Changed the gender field to an enum of: Male, Female, Transgender Male, Transgender Female, Intersex.

Added import/export changes.
Bumped the migration version number. May need to do it again depending on when it is merged.
Ported UI changes to 2.5 and added filtering.

lots0logs:
bump

:innocent:

PR #376, opened 2020-02-20
Generate cover image

### LINKED ISSUE #9: Generated image / video regeneration
https://github.com/stashapp/StashServer/issues/40

WithoutPants:
Resolves #9 

Note: this change also moves the scan/generate/import/export operations from `query` to `mutation`, since these are mutating operations. This will impact external scripts using these operations.

Adds a new tab to the scene page - Operations. Within this tab, there are two buttons. One generates the cover image for the scene based on the current position of the scene player, the other generates the cover image using the default settings.

Note that a page refresh may be required to see the updated images after generation, due to caching.

### DIFF
--- a/graphql/documents/mutations/metadata.graphql
+++ b/graphql/documents/mutations/metadata.graphql
@@ -1,0 +1,27 @@
+mutation MetadataImport {
+  metadataImport
+}
+
+mutation MetadataExport {
+  metadataExport
+}
+
+mutation MetadataScan($input: ScanMetadataInput!) {
+  metadataScan(input: $input)
+}
+
+mutation MetadataGenerate($input: GenerateMetadataInput!) {
+  metadataGenerate(input: $input)
+}
+
+mutation MetadataAutoTag($input: AutoTagMetadataInput!) {
+  metadataAutoTag(input: $input)
+}
+
+mutation MetadataClean {
+  metadataClean
+}
+
+mutation StopJob {
+  stopJob
+}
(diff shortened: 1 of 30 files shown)


bnkai:
Is this WIP?

First by changing the operations to mutation i get this `Invariant Violation: Invariant Violation: 10 (see https://github.com/apollographql/invariant-packages)` when i try to run any changed operation from the settings panel ( scan , autotag , generate)

Also you are using the generator from the previews task that was supposed to generate the previews and the webp for the wall view  not the screenshots. This leads to this error when trying to "create a thumbnail" 
`ERRO[0084] ffmpeg error when running command </usr/bin/ffmpeg -v error -ss 272.40798 -y -i /edited/path/to/filename.webm -vframes 1 -q:v 1 -vf scale=640:-1 -f image2 /edited/mystash/meta/tmp/b33cb3c05a5ba5b017573a43fdb1e74e.webp>: [image2 @ 0x55c5e91be700] Could not get frame filename number 2 from pattern '/edited/mystash/meta/tmp/b33cb3c05a5ba5b017573a43fdb1e74e.webp'. Use '-frames:v 1' for a single image, or '-update' option, or use a pattern such as %03d within the filename.`



 which is because you are using image generate options (-f image2) that are not compatible with webp (g.ImageFilename). 
The paths also are not the correct ones for the screenshot  and the thumbnail generation  is missing also...


The screenshots code should work like task_scan.go->makeScreenshots 
take the outputpath of the normal image, thumb and put it to the options output path before calling the encoder. This changes the output file to jpg instead of the webp now and also works directly at the screenshots directory  (we don't need / use  an intermediate tmp dir for the screenshot,thumbnail generation)

WithoutPants:
Yeah, you're right. The earlier version did that, but forgot to port that in during the rework.

WithoutPants:
Added setting image in the database and ported to 2.5 UI. Ready to review.

PR #380, opened 2020-02-24
Merge master back to develop

WithoutPants:
Merge master back to develop so that develop branch gets the v0.1 tag.


bnkai:
should be merged after 381

PR #382, opened 2020-02-24
Fix version check for v0.1.0+

bnkai:
Fix `panic: runtime error: slice bounds out of range`
Adding the tag to the master release changed the target_commitish part of the version in the releases endpoint to "master" from the sha value it had before
The only place to look for the sha value for v.0.1.0 ( "master ") is now the tags endpoint of github api and the commit.sha part specific
So if we don't have the sha from the releases endpoint we fallback to a call to the tags endpoint using the `version` as the tag name

### DIFF
--- a/pkg/api/check_version.go
+++ b/pkg/api/check_version.go
@@ -14,6 +14,7 @@ import (
 
 //we use the github REST V3 API as no login is required
 const apiReleases string = "https://api.github.com/repos/stashapp/stash/releases"
+const apiTags string = "https://api.github.com/repos/stashapp/stash/tags"
 const apiAcceptHeader string = "application/vnd.github.v3+json"
 const developmentTag string = "latest_develop"
 
@@ -84,6 +85,17 @@ type githubAsset struct {
 	Browser_download_url string
 }
 
+type githubTagResponse struct {
+	Name        string
+	Zipball_url string
+	Tarball_url string
+	Commit      struct {
+		Sha string
+		Url string
+	}
+	Node_id string
+}
+
 // GetLatestVersion gets latest version (git commit hash) from github API
 // If running a build from the "master" branch, then the latest full release
 // is used, otherwise it uses the release that is tagged with "latest_develop"
@@ -144,7 +156,9 @@ func GetLatestVersion(shortHash bool) (latestVersion string, latestRelease strin
 	}
 
 	if release.Prerelease == usePreRelease {
-		if shortHash {
+		if len(release.Target_commitish) != 40 { // sha is not present in the commit
+			latestVersion = getShaFromTags(shortHash, version)
+		} else if shortHash {
 			latestVersion = release.Target_commitish[0:7] //shorthash is first 7 digits of git commit hash
 		} else {
 			latestVersion = release.Target_commitish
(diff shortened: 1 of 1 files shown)


WithoutPants:
Documenting an alternative. 

Change the `GetLatestVersion` function to return a struct containing hash, tag, and release url. If the `Target_commitish` is `master`, then the hash is left blank. Tag is populated with `Tag_name` from the response. In `printLatestVersion`, if the hash is empty, then compare on the tag/version instead. Likewise in `LatestVersion`, if hash is empty, then return the tag. In the UI, change `Latest Version Build Hash` to be `Latest Version`. This will show either the hash or the newest release tag. This eliminates the need for a second API request.

See below code:

(check_version.go)
```golang
type githubRelease struct {
	Hash    string
	Tag     string
	Release string
}

// GetLatestVersion gets latest version (git commit hash) from github API
// If running a build from the "master" branch, then the latest full release
// is used, otherwise it uses the release that is tagged with "latest_develop"
// which is the latest pre-release build.
func GetLatestVersion(shortHash bool) (ret *githubRelease, err error) {

	platform := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
	wantedRelease := stashReleases()[platform]

	version, _, _ := GetVersion()
	if version == "" {
		return nil, fmt.Errorf("Stash doesn't have a version. Version check not supported.")
	}

	// if the version is suffixed with -x-xxxx, then we are running a development build
	usePreRelease := false
	re := regexp.MustCompile(`-\d+-g\w+$`)
	if re.MatchString(version) {
		usePreRelease = true
	}

	client := &http.Client{
		Timeout: 3 * time.Second,
	}

	url := apiReleases
	if !usePreRelease {
		// just get the latest full release
		url += "/latest"
	} else {
		// get the release tagged with the development tag
		url += "/tags/" + developmentTag
	}
	req, _ := http.NewRequest("GET", url, nil)

	req.Header.Add("Accept", apiAcceptHeader) // gh api recommendation , send header with api version
	response, err := client.Do(req)

	release := githubReleasesResponse{}

	if err != nil {
		return nil, fmt.Errorf("Github API request failed: %s", err)
	}

	if response.StatusCode != http.StatusOK {
		return nil, fmt.Errorf("Github API request failed: %s", response.Status)
	}

	defer response.Body.Close()

	data, err := ioutil.ReadAll(response.Body)
	if err != nil {
		return nil, fmt.Errorf("Github API read response failed: %s", err)
	}
	err = json.Unmarshal(data, &release)
	if err != nil {
		return nil, fmt.Errorf("Unmarshalling Github API response failed: %s", err)
	}

	ret = &githubRelease{}

	if release.Prerelease == usePreRelease {
		ret.Tag = release.Tag_name

		if release.Target_commitish != "master" {
			if shortHash {
				ret.Hash = release.Target_commitish[0:7] //shorthash is first 7 digits of git commit hash
			} else {
				ret.Hash = release.Target_commitish
			}
		}

		if wantedRelease != "" {
			for _, asset := range release.Assets {
				if asset.Name == wantedRelease {
					ret.Release = asset.Browser_download_url
					break
				}
			}
		}
	}

	if ret.Hash == "" && ret.Tag == "" {
		return nil, fmt.Errorf("No version found for \"%s\"", version)
	}
	return ret, nil
}

func printLatestVersion() {
	version, githash, _ = GetVersion()
	release, err := GetLatestVersion(true)
	if err != nil {
		logger.Errorf("Couldn't find latest version: %s", err)
	} else {
		compare := githash
		latest := release.Hash
		if latest == "" {
			latest = release.Tag
			compare = version
		}

		if compare == latest {
			logger.Infof("Version: (%s) is already the latest released.", latest)
		} else {
			logger.Infof("New version: (%s) available.", latest)
		}
	}
}
```

(resolver.go)
```golang
func (r *queryResolver) Latestversion(ctx context.Context) (*models.ShortVersion, error) {
	release, err := GetLatestVersion(true)
	if err != nil {
		logger.Errorf("Error while retrieving latest hash: %s", err)
	}

	ver := release.Hash
	if ver == "" {
		ver = release.Tag
	}

	return &models.ShortVersion{
		Shorthash: ver,
		URL:       release.Release,
	}, err
}
```

The major issue I see with this is that the graphql schema explicitly calls it `shorthash`, and there is no version field. We could possibly look at implementing some variation of this in the future.

bnkai:
The issue with the alternate code is that you expect to get "master" in the target_commitish which i'm not sure that it will be next time we tag the version. The whole issue was unpredictable anyway. Thus If target_commitish!=master it doesn't mean 100% that it will be a hash so the extra check for the length is needed to make sure we don't get again the slice out of bounds.
i chose the extra call to the api for simplicity and as a tradeoff between lesser code to change and an extra network call for the master version check

PR #384, opened 2020-02-25
Add detection of container/video_codec/audio_codec compatibility for live file streaming or transcoding

### LINKED ISSUE #354: [Bug Report] This video file cannot be played. (Error Code: 224003)
**Describe the bug**
When trying to play certain videos in the interface, I get an error:

```
This video file cannot be played.
(Error Code: 224003)
```

**To Reproduce**
1. Open up Stash
2. Click on a video to open it
2. Click on the "Play" arrow to start playback

**Expected behavior**
Video should play

**Screenshots**

 - OS: MacOS 10.15 (19A582a)
 - Browser: Chrome
 - Version: 81.0.4013.0 (Official Build) canary (64-bit)

According to the FIleinfo tab, the codec type is h264. (for my reference - e.g. checksum 8c15c34eb54244659c75e213887a317b).

Here is the output from ffprobe:

```
$ ffprobe test.mp4
ffprobe version 4.1.4-1~deb10u1 Copyright (c) 2007-2019 the FFmpeg developers
  built with gcc 8 (Debian 8.3.0-6)
  configuration: --prefix=/usr --extra-version='1~deb10u1' --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --arch=amd64 --enable-gpl --disable-stripping --enable-avresample --disable-filter=resample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libaom --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libcodec2 --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libjack --enable-libmp3lame --enable-libmysofa --enable-libopenjpeg --enable-libopenmpt --enable-libopus --enable-libpulse --enable-librsvg --enable-librubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxml2 --enable-libxvid --enable-libzmq --enable-libzvbi --enable-lv2 --enable-omx --enable-openal --enable-opengl --enable-sdl2 --enable-libdc1394 --enable-libdrm --enable-libiec61883 --enable-chromaprint --enable-frei0r --enable-libx264 --enable-shared
  libavutil      56. 22.100 / 56. 22.100
  libavcodec     58. 35.100 / 58. 35.100
  libavformat    58. 20.100 / 58. 20.100
  libavdevice    58.  5.100 / 58.  5.100
  libavfilter     7. 40.101 /  7. 40.101
  libavresample   4.  0.  0 /  4.  0.  0
  libswscale      5.  3.100 /  5.  3.100
  libswresample   3.  3.100 /  3.  3.100
  libpostproc    55.  3.100 / 55.  3.100
[mpegts @ 0x55ef7f6f4000] PES packet size mismatch
    Last message repeated 1 times
Input #0, mpegts, from 'test.mp4':
  Duration: 02:36:51.19, start: 0.999000, bitrate: 7451 kb/s
  Program 1
    Stream #0:0[0x810]: Video: h264 (High) ([27][0][0][0] / 0x001B), yuv420p(tv, bt709, top first), 1920x1080 [SAR 1:1 DAR 16:9], 29.97 fps, 59.94 tbr, 90k tbn, 59.94 tbc
    Stream #0:1[0x814](eng): Audio: aac (LC) ([15][0][0][0] / 0x000F), 48000 Hz, stereo, fltp, 196 kb/s
```

bnkai:
Fixes #354 
As it is now stash only selects which files are "streamable" by using the video_codec information from ffprobe stored in DB. In some files while the codec is indeed supported by the browser/jwplayer the container isn't and thus instead of live transcoding the files we get the error message from #354 

- the container isn't stored in the database so we either need to store it or simply call ffprobe before we begin the streaming of a media file. For now i chose to go with the second option and it doesn't seem to add any measurable player startup delay

- a combination of valid codec/container pairs is used in addition to the codec validation. mp4 and webm seem to be the "safe" choises for containers so i whitelisted those 

- ffprobe treats webm and matroska as the same which is problematic for firefox and probably a lot more browsers (chrome doesn't have a problem). I had to use a "magic_file" reader to determine the actual container (reads 4kb of the file, no measurable delay here either) for this case .

As it is now all the error cases from #354 get live transcoded ok.

TODO / TBD

- Browser Detection ( **EDIT** not to be implemented ,  the user agent header is way too abused to make a consistent identification of the web browser )
<del>We could detect which browser is used and adjust the valid combinations accordingly ( user agent header is available in routes_scene ,i left a TODO there ). For example chrome (and probably most chromium browsers ) suppports matroska as a container for h264 while firefox doesn't. Prior to this PR mkv and h264 played directly in chrome while after this fix (as it is now) the files will get live transcoded</del>
A force matroska option was implemented instead 

- Transcode generation ( **EDIT** completed)
<del>I haven't touched the file transcode code yet.If we adjust it over there then there should be a new profile that copys the video stream and doesn't reencode as the codec is supported and only the container isn't. Something like `ffmpeg -i input h264_media_file.avi -c:v copy  .... output.mp4` </del>
Transcoded files (mp4) are generated correctly when the codec/container is not supported. For h264 files the video stream is directly copied avoiding time and quality loss.

- Stream copy profile  ( **EDIT** not to be implemented , the main benefit would be h264 encoded files but the resulting mp4 file will not be fully seekable )
Same as above for files that the container is not supported but the codec is. Instead of live transcoding using the webm live stream profile is it possible to stream only `copying` the video stream? Most use  cases would be h264 in mpegts,flv,avi that the video could be copied to mp4 ( no reduction in  quality compared to transcoding and almost no cpu usage ) . 
**EDIT** _streaming is possible by copying the video stream but for h264 (mkv,flv,ts,avi) the resulting mp4 file won't be fully seekable (the duration keeps increasing till the file is fully loaded to the player). <del>Is this ok or should we transcode this cases also to vp9.webm ?_ </del>
**EDIT** After some more testing the fragmented mp4 (for the h264 mkv,flv,ts,avi ) that is not fully seekable is not acceptable imo.
Despite transcoding and loosing a bit in quality the file seeking/seek to sprite of the webm encoded file makes a better user experience.

- Add container as an extra field to the scenes table. ( **EDIT** completed)
Is a cleaner solution
<del>For new users it should be fairly easy to implement, it's a bit more tricky for users with existing DBs and more complex to implement than the current solution.</del>
A single scan is needed to populate the format field (container) for already existing files. 

- Audio codec (**EDIT** completed)
Is it worth the extra complexity / code to check for valid audio codec ? I didn't see any issues with files playing the video part and not the audio.


bnkai:
Added container detection to transcode  task.
For h264 files that are not supported (ts,flv,avi,mkv ...) the video part is copied instead of being reencoded.

pkg/api/routes_scene.go
```diff
@@ -42,13 +42,29 @@ func (rs sceneRoutes) Routes() chi.Router {
 // region Handlers
 
 func (rs sceneRoutes) Stream(w http.ResponseWriter, r *http.Request) {
+
+	//TODO get user agent and use to fine tune media compatibility
+	//ua := r.Header.Get("User-Agent")
+	//logger.Infof("[stream] user agent is: %s \n", ua)
```
WithoutPants: If we're not intended on using browser detection, then we can remove this.

pkg/api/routes_scene.go
```diff
@@ -42,13 +42,29 @@ func (rs sceneRoutes) Routes() chi.Router {
...
+	videoFile, err := ffmpeg.NewVideoFile(manager.GetInstance().FFProbePath, scene.Path)
+	if err != nil {
+		logger.Errorf("[stream] error reading video file: %s", err.Error())
+		return
+	}
+	container := ffmpeg.FfprobeToContainer[videoFile.Container]
+	if container == ffmpeg.Matroska {
+		container = ffmpeg.MagicContainer(scene.Path) // use magic file instead of ffprobe for matroska,webm
+	}
+
```
WithoutPants: Need to consider the performance impacts of this and compare against storing the container in the database instead.

pkg/api/routes_scene.go
bnkai: it was left there for internal testing. will be removed once i remove the WIP

pkg/api/routes_scene.go
bnkai: The container needs to be stored to the database eventually because of the generate transcodes task penalty .

If we want to merge this to the stable though as a bug fix release ( eg 0.1.2) we can leave the db implementation for 0.2.0 and a seperate PR. If we don't care about pushing to stable but only for dev / 0.2.0 then i should implement the db part in this PR.

The generate transcodes part has a noticable penalty and thus i  can revert it till the db implementation is ready.

bnkai:
Container info added to DB.
Before testing make sure to backup the database file as db version is increased to 4.
After migrating to this version a scan is needed to populate the missing container to all media files.
This scan will take some time as it ffprobes all files to get the info. 
Once the first scan is finished all the next ones are as fast as before.

WithoutPants:
Can you please rebase against `develop` and update the migration to be 5? The schema version has been bumped on develop.

WithoutPants:
I'm seeing regression in behaviour for some combinations. It appears that previous combinations (tested with mp4/flv combination) that did not work for me now appear to work, but they do not allow seeking. That may be a limitation somewhere for this scenario. However, I noticed that on the current version in Chrome, the mp4/matroska combination works correctly with seeking, but in this PR, the seeking is disabled.

bnkai:
Yes as i stated at the beginning i wanted the browser detection mainly for matroska and Chromium based browsers. What do you think about adding an extra conf option to "force" mkv as supported? Some files are non seekable anyway though and i don't think there is a way to fix that for live streaming ( only fully transcoding deals with that)

WithoutPants:
> What do you think about adding an extra conf option to "force" mkv as supported?

Yep, I'd like this.

The seeking stuff is because `IsStreamable` (in `manager/utils.go`) hasn't been updated with your new logic.

bnkai:
Pushed a fix for `manager/utils.go`.
I switched to WIP till i implement the `force mkv` option.

bnkai:
Is ` ui/v2.5/src/core/generated-graphql.tsx` supposed to be tracked by git ?
I didn't notice that it was generated it got added to the commit....

WithoutPants:
> Is ` ui/v2.5/src/core/generated-graphql.tsx` supposed to be tracked by git ?

No. I just haven't yet bothered to add it to the gitignore file. Doesn't matter.

WithoutPants:
When I set the force mkv to true, the mkv files are not transcoded as expected, but the seeking doesn't work - regardless of where you seek to, it starts from the start of the file.

bnkai:
Does it print anything when you play those mkvs?
It shouldn't print anything(transcoding file ....) if they are served directly. Can you try the same files with the dev/master version?

WithoutPants:
The same files stream directly and correctly with the dev version. There is no log entry indicating that the file is being transcoded, so I'm confident that it is not. When it does transcode (when the checkbox is cleared), the seeking works correctly.

bnkai:
Can you try reloading the page with  player / restarting stash after you save the configuration with the mkv as forced? The only thing i can think of is that you get a cached response for the `isStreamable` the player is requesting. I added a couple of prints that i'll remove later to check this out.

WithoutPants:
You are correct. Refreshing the page fixes the issue. We may want to consider resetting the cache when setting those flags.

I also noticed that when I forced the mkv support, I didn't get sound on my particular video. The audio coded was ac3. I haven't yet checked if other ac3 videos work. Note that this happens with the current version as well, so not something introduced by your change. Unchecking the force mkv flag transcodes the file so the audio is present.

One other possibility, could we determine codec/container support on the client-side browser and send that down as a parameter to the stream request? That might side-step needing to set the flags at all.

bnkai:
> Audio codec
Is it worth the extra complexity / code to check for valid audio codec ? I didn't see any issues with files playing the video part and not the audio.
mkv supports a lot of audio codecs and ac3 is common in movies, browsers only support aac in mkv i think (not sure about opus or vorbis). If we need it then i have to extend the combo check to the audio codec also and transcode as needed.
I wouldn't want to add more to the client side. Using these flags is supposed to be a one time only depending on the browser one uses , a page refresh after the save is not that big of a deal for that imho. 
Have you tried the problematic avi btw?
also https://github.com/henrypp/chromium/releases has windows builds of chromium that should support hevc

bnkai:
Added checks for audio codec / player compatibility.
As a special case if `MKV` is forced as supported then when the media file is mkv and only its audio codec is not supported eg `ac3` we copy the video stream as is, transcode the audio as opus and stream the media file as mkv(mime video/x-matroska) instead of webm. This results in less cpu usage and better video quality ( no video transcoding) for the specific case.

WithoutPants:
A possible future iteration to investigate (not necessary for this PR): https://www.nomensa.com/blog/2011/detecting-browser-compatibility-html5-video-and-audio

Determine from the client-side if the video is directly streamable, if not then do request to `stream.mp4?transcode=1`

Forgot to mention what is actually working. This fixed streaming the files which were previously not working for me.

I'll do another review pass and test with the latest build as soon as I'm able.

bnkai:
I think those solutions are too much trouble for little gain. They move the selection code from the server to the  client. The extra code in the server side is used also to make "better" decisions for the transcode part ( copy or transcode a stream when needed instead of the whole video ) . 
If and when needed we should probably checkout ways to invalidate the isStreamable query or make it non cachable.

pkg/api/routes_scene.go
```diff
@@ -69,16 +89,49 @@ func (rs sceneRoutes) Stream(w http.ResponseWriter, r *http.Request) {
...
+	if audioCodec == ffmpeg.MissingUnsupported {
+		//ffmpeg fails if it trys to transcode a non supported audio codec
+		stream, process, err = encoder.StreamTranscodeVideo(*videoFile, startTime, config.GetMaxStreamingTranscodeSize())
+	} else {
+		copyVideo := false        // try to be smart if the video to be transcoded is in a Matroska container
+		if config.GetForceMKV() { // If MKV is forced as supported and video codec is also supported then only transcode audio
+			if ffmpeg.Container(container) == ffmpeg.Matroska {
+				if config.GetForceMKV() {
+					switch videoCodec {
+					case "h264", "vp9", "vp8":
```
WithoutPants, line 105: Add constants to `ffmpeg` module and refer to them here.

pkg/ffmpeg/ffprobe.go
```diff
@@ -10,11 +10,89 @@ import (
...
+}
+
+var FfprobeToContainer = map[string]Container{
+	"mov,mp4,m4a,3gp,3g2,mj2": Mp4, // browsers support all of them so we don't  care
+	"asf":                     Wmv,
+	"avi":                     Avi,
+	"flv":                     Flv,
+	"mpegts":                  Mpegts,
+	"matroska,webm":           Matroska,
+}
```
WithoutPants, line 74: We're reusing string literals here, so we should create and/or refer to constants for them.

pkg/ffmpeg/ffprobe.go
```diff
@@ -10,11 +10,89 @@ import (
...
+	if container == "" { // if format is not in our Container list leave it as ffprobes reported format_name
+		container = Container(format)
+	}
+	return container
+}
+
 func IsValidCodec(codecName string) bool {
+	forceHEVC := config.GetForceHEVC()
+	if forceHEVC {
+		if codecName == "hevc" {
```
WithoutPants, line 91: Replace string literal with constant.

pkg/ffmpeg/ffprobe.go
```diff
@@ -23,6 +101,78 @@ func IsValidCodec(codecName string) bool {
...
+	}
+	return false
+}
+
+//extend stream validation check to take into account container
+func IsValidCombo(codecName string, format Container) bool {
+	forceMKV := config.GetForceMKV()
+	forceHEVC := config.GetForceHEVC()
+	switch codecName {
+	case "h264":
```
WithoutPants, line 148: Replace string literals with constants.

pkg/ffmpeg/media_detection.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// detect file format from magic file number
+// https://github.com/lex-r/filetype/blob/73c10ad714e3b8ecf5cd1564c882ed6d440d5c2d/matchers/video.go
+
+func mkv(buf []byte) bool {
+	return len(buf) > 3 &&
+		buf[0] == 0x1A && buf[1] == 0x45 &&
+		buf[2] == 0xDF && buf[3] == 0xA3 &&
+		containsMatroskaSignature(buf, []byte{'m', 'a', 't', 'r', 'o', 's', 'k', 'a'})
```
WithoutPants, line 16: It should be possible to construct the byte slice from a string literal rather than an array of bytes.

pkg/ffmpeg/media_detection.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func webm(buf []byte) bool {
+	return len(buf) > 3 &&
+		buf[0] == 0x1A && buf[1] == 0x45 &&
+		buf[2] == 0xDF && buf[3] == 0xA3 &&
+		containsMatroskaSignature(buf, []byte{'w', 'e', 'b', 'm'})
```
WithoutPants, line 23: As above.

pkg/manager/task_transcode.go
```diff
@@ -41,24 +61,52 @@ func (t *GenerateTranscodeTask) Start(wg *sync.WaitGroup) {
 		MaxTranscodeSize: transcodeSize,
 	}
 	encoder := ffmpeg.NewEncoder(instance.FFMPEGPath)
-	encoder.Transcode(*videoFile, options)
+
+	if videoCodec == "h264" { // for non supported h264 files stream copy the video part
```
WithoutPants, line 65: Replace string literal with `ffmpeg` constant.

pkg/manager/utils.go
```diff
@@ -12,12 +12,30 @@ func IsStreamable(scene *models.Scene) (bool, error) {
...
 
 	videoCodec := scene.VideoCodec.String
-	if ffmpeg.IsValidCodec(videoCodec) {
+	audioCodec := ffmpeg.MissingUnsupported
+	if scene.AudioCodec.Valid {
+		audioCodec = ffmpeg.AudioCodec(scene.AudioCodec.String)
+	}
+
+	if ffmpeg.IsValidCodec(videoCodec) && ffmpeg.IsValidCombo(videoCodec, container) && ffmpeg.IsValidAudioForContainer(audioCodec, container) {
+		fmt.Printf("File is streamable %s, %s, %s\n", videoCodec, audioCodec, container)
```
WithoutPants, line 34: Is this a debug output? Should this be changed to a `logger` call?

pkg/manager/utils.go
```diff
@@ -12,12 +12,30 @@ func IsStreamable(scene *models.Scene) (bool, error) {
...
 		return true, nil
 	} else {
 		hasTranscode, _ := HasTranscode(scene)
+		fmt.Printf("File is not streamable , transcode is needed  %s, %s, %s\n", videoCodec, audioCodec, container)
```
WithoutPants, line 38: See above.

pkg/ffmpeg/media_detection.go
bnkai: The specific code was pullled as is from the project mentioned. Although it should work if we change it as you say, i think it's in this format to emphasize that  it is being used for binary reading a file and to make sure the exact byte count is correct. 
This probably explains it better https://medium.com/@xoen/go-strings-vs-byte-slices-5beb4c411bda.
If it's for cosmetic reasons i would like to keep it as is.

WithoutPants: Given that you've documented the origin, I'm cool with this.

pkg/api/routes_scene.go
```diff
@@ -69,16 +89,49 @@ func (rs sceneRoutes) Stream(w http.ResponseWriter, r *http.Request) {
...
+						copyVideo = true
+					case "hevc":
+						if config.GetForceHEVC() {
+							copyVideo = true
+						}
```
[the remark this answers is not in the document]
bnkai: you were right one is redudant.

bnkai:
The reload is only once though when you change the settings anyway. I didn't have to change / look back at  the mkv settings since i started the PR for example.
I think all the "cache" issues we have in stash can be better addressed when apollo client gets to the next version https://blog.apollographql.com/previewing-the-apollo-client-3-cache-565fadd6a01e
A lot more fine grained control over cache.

PR #388, opened 2020-03-01
Fix performer height filter error

InfiniteTF:
The performer height filter is broken due to a column with the same name in the scenes table that is being joined in.


WithoutPants:
Replaced all of the criteria with qualified names to avoid existing and future latent bugs.

PR #389, opened 2020-03-02
Update browser cache when images are updated

InfiniteTF:
Fetching with `{ cache: reload }` forces the browser to update the local cache with a new copy. This prevents the issue where the image remains the old one after it has been updated.

I wanted to do the same for scenes, but jwplayer ignores the fact that the cache has been updated, so I left the existing method alone.


WithoutPants:
Would doing the same thing for scenes improve the scenes page?

PR #390, opened 2020-03-05
Fix various issues with v2.5 UI

WithoutPants:
- Fixed the breakpoint for the navbar so that it collapses before items go off-screen
- Fixed the breakpoint of the settings page so that it collapses before the left menu overlaps the content
- Fixed the height discrepancy of the filter components
- Changed the styling around to make it as close as possible to v2
- Fixed the scene card orientation and zoom sizes
- Fixed custom css not working
- Changed default settings tab to Tasks
- Fixed the progress bar quickly appearing when entering tasks page, made the bar animated and made the percentage text an integer
- Fixed not being able to unset scene studio
- Fixed scene gallery field not being populated in edit screen

### DIFF
--- a/ui/v2.5/src/components/MainNavbar.tsx
+++ b/ui/v2.5/src/components/MainNavbar.tsx
@@ -89,7 +89,7 @@ export const MainNavbar: React.FC = () => {
       <Navbar.Collapse className="order-3 order-md-1">
         <Nav className="mr-md-auto">
           {menuItems.map(i => (
-            <Nav.Link eventKey={i.href} key={i.href}>
+            <Nav.Link eventKey={i.href} as="div" key={i.href}>
               <LinkContainer
                 activeClassName="active"
                 exact
--- a/ui/v2.5/src/components/MainNavbar.tsx
+++ b/ui/v2.5/src/components/MainNavbar.tsx
@@ -73,23 +73,23 @@ export const MainNavbar: React.FC = () => {
       variant="dark"
       bg="dark"
       className="top-nav"
-      expand="sm"
+      expand="md"
     >
-      <Navbar.Brand as="div" className="order-1 order-sm-0">
+      <Navbar.Brand as="div" className="order-1 order-md-0">
         <Link to="/">
-          <Button className="minimal brand-link d-none d-sm-inline-block">
+          <Button className="minimal brand-link d-none d-md-inline-block">
             Stash
           </Button>
-          <Button className="minimal brand-icon d-inline d-sm-none">
+          <Button className="minimal brand-icon d-inline d-md-none">
             <img src="favicon.ico" alt="" />
           </Button>
         </Link>
       </Navbar.Brand>
       <Navbar.Toggle className="order-0" />
-      <Navbar.Collapse className="order-3 order-sm-1">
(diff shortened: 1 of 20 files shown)


WithoutPants:
One outstanding issue: Edit tab disappears when viewport width drops below 576px.

ui/v2.5/src/components/MainNavbar.tsx
```diff
@@ -73,23 +73,23 @@ export const MainNavbar: React.FC = () => {
...
             <img src="favicon.ico" alt="" />
           </Button>
         </Link>
       </Navbar.Brand>
       <Navbar.Toggle className="order-0" />
-      <Navbar.Collapse className="order-3 order-sm-1">
+      <Navbar.Collapse className="order-3 order-md-1">
         <Nav className="mr-md-auto">
           {menuItems.map(i => (
-            <Nav.Link eventKey={i.href} as="div" key={i.href}>
```
InfiniteTF: The `as="div"` is unfortunately kind of necessary since otherwise it renders as an anchor. The issue is that the button below is also an anchor, and nesting anchors causes a console error.

ui/v2.5/src/index.scss
```diff
@@ -23,16 +23,40 @@ body {
...
+}
+
+.text-input, .text-input:focus {
+  background-color: $textfield-bg;
+}
+
+.input-control, .input-control:focus {
+  background-color: $secondary;
+}
+
```
InfiniteTF: Would it be possible to apply these styles to all input elements rather than classes? 
`input[type=text], textarea, select {
  ...
}`

ui/v2.5/src/index.scss
```diff
@@ -56,40 +80,84 @@ code {
...
 
 .scene-card-video {
   height: auto;
   width: 100%;
+}
 
-  .zoom-0 {
-    height: 11.25rem;
+/* this is a bit of a hack, because we can't supply direct class names
+   to the react-select controls */
```
InfiniteTF: Yeah, react-select is quite tedious to style. You can supply styles to it, but it has to be set on each individual part of the select and leads to a ton of boilerplate. This seems like a better approach tbh.

ui/v2.5/src/index.scss
```diff
@@ -56,40 +80,84 @@ code {
 
 @media (min-width: 576px) {
   .zoom-0 {
-    width: 15rem;
+    width: 240px;
+
+    & .scene-card-video {
```
InfiniteTF: The ampersand is implied when two styles are nested, so it's not strictly speaking necessary.

ui/v2.5/src/components/MainNavbar.tsx
WithoutPants: I don't even remember why I changed it in the first place. Reverted.

ui/v2.5/src/index.scss
WithoutPants: Addressed this and optimised a little after doing some reading up on sass.

ui/v2.5/src/index.scss
InfiniteTF: The default type of a Form.Control is text, but yeah it can be hard to override a lot of the bootstrap styles due to precedence rules. No big deal either way.

PR #395, opened 2020-03-12
Fix error when viewing scenes related to objects with illegal characters in name

### LINKED ISSUE #394: [Bug Report] Syntax error when movies have a # symbol in the name
**Describe the bug**
When a movie has a # in the name, clicking on "Scenes" then causes a syntax error. 
`Something went wrong.
Details
SyntaxError: Unexpected end of JSON input`

**To Reproduce**
1. Create a movie with a # in the name
2. Add a scene to this movie
3. Click on "Scenes" for this movie

**Expected behavior**
Behavior should mirror movies without a # symbol

 - Windows 10
 - Opera 67.0.3575.53

WithoutPants:
Fixes #394 

Encodes `ILabeledId` objects so that illegal characters (`#;,/?:@&=+$`) don't break the query URL parsing.

Tested by creating Movies, Studios, Performers and tags with illegal characters in the names, then clicking on the query link on the relevant page. Change has been ported to UI v2.5 except for Movies, since that is not yet in 2.5.

Also includes linting and format changes for 2.5.

### DIFF
--- a/ui/v2.5/src/models/list-filter/types.ts
+++ b/ui/v2.5/src/models/list-filter/types.ts
@@ -22,6 +22,12 @@ export interface ILabeledValue {
   value: string;
 }
 
+export function encodeILabeledId(o: ILabeledId) {
+  let ret = Object.assign({}, o);
+  ret.label = encodeURIComponent(o.label);
+  return ret;
+}
+
 export interface IOptionType {
   id: string;
   name?: string;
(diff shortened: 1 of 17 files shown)


ui/v2.5/src/models/list-filter/types.ts
```diff
@@ -22,6 +22,12 @@ export interface ILabeledValue {
   value: string;
 }
 
+export function encodeILabeledId(o: ILabeledId) {
+  const ret = { ...o };
+  ret.label = encodeURIComponent(o.label);
+  return ret;
+}
```
InfiniteTF: You can just do `return { ...o, label: encodeURIComponent(o.label) }`. Properties are set in the order they are passed in, so even if `o` has a label, it will be overwrittenby encodeURIComponent.

WithoutPants: Done.

PR #412, opened 2020-03-20
Add modes for performer/tag for bulk scene editing

### LINKED ISSUE #250: [Enhancement] Insert tag & overwrite tag button group for bulk scene editing
The current implementation of bulk scene editing overwrites existing tags. As discussed on Discord, enhance this feature to allow insertion of new tags into existing tags and to visually notify users what mode they are in: `insert tag` or `overwrite tag` mode.

In the `operation-container` to right of the tags input box add a `button-group` with two icons, an insert tags icon and an overwrite tags icon which can be represented with Blueprint's `insert` icon and the `highlight` icon for overwrite. I suppose icon tooltips would be useful. 

I suspect that the `insert` icon should be the default active button since it is less destructive.

WithoutPants:
Resolves #250 

Adds a button to the performer and tag controls in the bulk scene update UI. This button shows the current mode of the operation - Add, Remove and Set. The set mode overwrites the existing ids with the ids supplied. Add and Remove are iterative operations using existing data.

Added to version 2 and 2.5 UI.

### DIFF
--- a/graphql/documents/mutations/scene.graphql
+++ b/graphql/documents/mutations/scene.graphql
@@ -39,8 +39,8 @@ mutation BulkSceneUpdate(
   $rating: Int,
   $studio_id: ID,
   $gallery_id: ID,
-  $performer_ids: [ID!],
-  $tag_ids: [ID!]) {
+  $performer_ids: BulkUpdateIds,
+  $tag_ids: BulkUpdateIds) {
 
   bulkSceneUpdate(input: {
                         ids: $ids,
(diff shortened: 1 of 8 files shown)


bnkai:
Looks ok , tests ok.
When you select a few scenes and they have the same tags or performers  they used to appear in the controls list. Now to make them appear you have to cycle to the edit button. Is this the expected behaviour?

WithoutPants:
Given that add/remove operations are iterative, I made the design decision to only show the common tags/performers when in Set mode. I think it would be unhelpful to fill in these values in the other modes, but I'm not convinced that my implementation is the most intuitive either, so I am open to suggestion.

bnkai:
Just wasn't sure what to expect, it works for me. Might be better to merge it as is and get feedback  from more users.

PR #415, opened 2020-03-21
Backup database if a migration is needed

bnkai:
When a migration is needed the database gets backed up.
The backup is stored in the same folder as the original db with a filename like this `stash-go.sqlite.4.20200321_221931` 
**stash-go.sqlite** was the filename of stash's DB
**4** is the db version of the original DB
**20200321_221931** is the date_time of the backup

I had to use a newer version of mattn's sqlite library (1.13) that includes the `VACUUM INTO` support hence the changes to the `go.mod` file and the `vendor` directory (go mod vendor).


pkg/database/database.go
```diff
@@ -55,6 +57,59 @@ func Reset(databasePath string) error {
...
+	}
+	defer db.Close()
+
+	var dbVersion uint
+	err = db.Get(&dbVersion, "SELECT version FROM schema_migrations")
+	if err != nil {
+		return err
+	}
+
+	if dbVersion > appSchemaVersion {
```
WithoutPants: `runMigrations` already gets the database versions and checks the difference. Can we not call this stuff from there? We can ensure that backup doesn't run on new databases by checking for schema version `0`.

bnkai: i made it independent so that it might be called from elsewhere for future purely backup purposes
eg task -> backup database

bnkai: Also in runMigrations as it is now you don't have access to the database to run a query.
migrate.New is used to directly access the db and it only returns a migrate type.
If we wanted to do it there we would need to add the sql.Open there , check for errors , run the migrate.NewWithDatabaseInstance and so on ....

WithoutPants:
Added a PR for your branch.

PR #416, opened 2020-03-21
docker-compose.yml url fix

bnkai:


bnkai:
this is the docker-compose.yml of the dev branch.
now that master is back it works but should it point out to the master?
shouldn't it point to its own branch?

PR #417, opened 2020-03-22
added .grid class and updated various components to use it

Anon247:
Done to support a much more flexible grid layout. Also removes the need to change potentially 100's of classNames in the DOM at the same time.

### DIFF
--- a/ui/v2.5/src/components/Performers/styles.scss
+++ b/ui/v2.5/src/components/Performers/styles.scss
@@ -7,19 +7,24 @@
 }
 
 #performer-page {
-  flex-direction: row;
-  margin: 10px auto;
-  overflow: hidden;
+  margin: 10px auto 0;
+  max-width: 1600px;
 
-  .image-container .performer {
-    max-height: 960px;
-    max-width: 100%;
+  &.grid{
+    grid-template-columns: auto 1fr;
+    grid-template-rows: auto 1fr;
+  }
+
+  .performer-image {
+    grid-row: 1 / 3;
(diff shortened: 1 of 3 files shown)


Anon247:
Still seem to be struggling with line seperators... I'll hold off on that commit for now

InfiniteTF:
I don't think the scene card change is worthwhile. When rapidly changing the zoom levels the performance is definitely worse with grid than the existing flexbox layout.

The performer page layout is an improvement though, both in terms of performance and simplicity. However, I think there should be a bit more whitespace, and maybe a smaller max-height on the performer image since it takes over the entire page

Anon247:
Fair enough, I'll update the PR to remove the grid scene page stuff

Anon247:
@InfiniteTF Added those changes you requested and rebased onto latest commit so no need to worry about conflicts, please review.

ui/v2.5/src/components/Performers/styles.scss
```diff
@@ -7,19 +7,24 @@
...
+  &.grid{
+    grid-template-columns: auto 1fr;
+    grid-template-rows: auto 1fr;
+  }
+
+  .performer-image {
+    grid-row: 1 / 3;
+    cursor: pointer;
+    max-width: 50vw;
+    max-height: 70vh;
```
WithoutPants, line 22: Why 70vh? Ends up with a lot of wasted vertical space underneath.

WithoutPants:
Closing this for inactivity for now. Feel free to reopen if needed.

PR #418, opened 2020-03-22
Add SVG studio image support, and studio image caching

### LINKED ISSUE #408: [Feature] SVG support for studio images
I've got some studio logos that are in SVG format, yet they do not display properly once saved. If these could be displayed the same way as PNGs and JPGs that would be ideal. Right now I'm having to convert the SVG files to PNGs.

InfiniteTF:
Fixes #408 

Many studios use SVG logos natively, and it's preferable to keep them that way since they're small and scale perfectly. Also added ETAG caching, same as with th performer images.

### DIFF
--- a/pkg/api/routes_studio.go
+++ b/pkg/api/routes_studio.go
@@ -35,7 +35,7 @@ func (rs studioRoutes) Image(w http.ResponseWriter, r *http.Request) {
 	}
 
 	contentType := http.DetectContentType(studio.Image)
-	if contentType == "text/xml; charset=utf-8" {
+	if contentType == "text/xml; charset=utf-8" || contentType == "text/plain; charset=utf-8" {
 		contentType = "image/svg+xml"
 	}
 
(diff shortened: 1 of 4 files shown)


Anon247:
#408 will be closed when this is merged

pkg/api/routes_studio.go
```diff
@@ -23,6 +26,21 @@ func (rs studioRoutes) Routes() chi.Router {
...
+	etag := fmt.Sprintf("%x", md5.Sum(studio.Image))
+	if match := r.Header.Get("If-None-Match"); match != "" {
+		if strings.Contains(match, etag) {
+			w.WriteHeader(http.StatusNotModified)
+			return
+		}
+	}
+
+	contentType := http.DetectContentType(studio.Image)
+	if contentType == "text/xml; charset=utf-8" {
```
bnkai: for some svgs i tested  `contentType == "text/xml; charset=utf-8" || contentType == "text/plain; charset=utf-8"` is needed

InfiniteTF: Updated

WithoutPants:
For consistency, I think this should either wait for the 2.5 UI cutover, or be ported to v2 UI.

PR #419, opened 2020-03-22
UI fixes

InfiniteTF:
* Change list filter query event handler from onChange to onInput to fix an issue where it wouldn't trigger when deleting the entire query.
* Change new movie view to full width with 1200px max-width so it doesn't start scaling down until the screen size goes below 1200px.
* Set spellcheck=false on folder path input to prevent the browser from "correcting" the path.
* Add margin and set cursor to pointer for image input button.
* Linting/format

Note that index/styles scss seems to have wonky line endings. I had to force the file to Windows style to prevent a full file diff, but there's still some noise.


InfiniteTF:
Also changed some settings input handlers from onChange to onInput. Same issue, deleting the auth information and saving didn't actually wipe the auth info.

PR #427, opened 2020-03-27
Add library size to main stats page

### LINKED ISSUE #426: [Feature] Total Library File Size
I think it would be useful to be able to see the total file size of your collection on the landing page where the totals numbers for scene, movies, performers, tags, etc are shown. I'd love the ability to quickly reference that info.

bnkai:
Add the library size ( sum of the sizes of all scenes ) to the stats page.

closes #426


pkg/models/querybuilder_scene.go
```diff
@@ -191,6 +192,14 @@ func (qb *SceneQueryBuilder) Count() (int, error) {
 	return runCountQuery(buildCountQuery("SELECT scenes.id FROM scenes"), nil)
 }
 
+func (qb *SceneQueryBuilder) SizeCount() (string, error) {
+	sum, err := runSumQuery("SELECT SUM(size) as sum FROM scenes", nil)
+	if err != nil {
+		return "O", err
```
InfiniteTF: Out of curiosity, what's the purpose of returning O? Can you not return an empty string?

bnkai: 0 seemed better than an empty string since we are talking about size. But doesn't actually matter , do you think `""` would look better?

InfiniteTF: I don't really have a preference since it's an error anyway, I was just curious since it returns the letter O. 0 makes more sense.

bnkai: oh it was supposed to be "0" (zero) ! Just noticed that now , typo...

PR #429, opened 2020-03-31
[Feature] Add image count to gallery list

ueaslsef:
As always just a small addition from me. Might be worthwile to take with us to the new ui.
image
_Images taken from unsplash._


bnkai:
looks ok, do you mind applying the same to the v2.5 file ?

PR #432, opened 2020-04-01
Add scene rating to scene filename parser

### LINKED ISSUE #406: [Feature] Parse rating in filename
A field in the scraper to parse filenames including a field for {rating}
For example-
9_[clips4sale]studio title.mp4 where 9 is the rating

WithoutPants:
Fixes #406 

Adds `{rating}` as a parsable field to the scene filename parser. Added to v2 and v2.5 of the UI.

Also fixes up display issues in the v2.5 version of the scene filename parser page.


bnkai:
Works ok.
Just a question about intended usage. 
Is using the filename parser to batch remove some ratings considered normal / acceptable ? I now i get `Error: GraphQL error: strconv.Atoi: parsing "": invalid syntax` when trying to unset some ratings using the filename parser (clear the rating field and hit apply).

WithoutPants:
Fixed v2.5 so that it is possible to unset rating. Fixed parser results refreshing twice after applying changes. Fixed rating selection not being correctly disabled when checkbox was unchecked.

Also includes a change to ignore line endings in the scss files so that the formatter works properly on Windows.

bnkai:
It seems to work ok now. Can confirm that double loading is gone and setting, unsetting are fine also.

PR #433, opened 2020-04-01
Cut over to v2.5 UI

InfiniteTF:
Also changed port back to 3000 and updated the README from create-react-app.


bnkai:
Seems ok to me.
I also tested the resulting linux binary from transfer.sh just to be sure.

Issue #436, opened 2020-04-02
[Feature] Add "Missing Gender" filter to performer section


VonTittyslappen:
I believe that it would be very beneficial for there to be a filter where you can view all performers that are missing a gender within the performers section

Cheers

bnkai:
the "Is Missing" filter would be useful to double check for missing gender, url , birthdate, ethnicity ....

Cutters1982:
Yep, definitely! :)

Anon247:
I'll work on this one

PR #438, opened 2020-04-02
Added various missing filters to performer page

Anon247:
Filters have been added to the performer page. I think that the 'Is Missing' criterion will need to be cleaned up and abstracted soon if we're going to be adding the 'Is Missing' criterion to other pages; I wasn't sure what would be the best way of doing it so I just duplicated it for now. Also, only applied changes to V2.5 UI as the switchover is soon.

Supported missing parameters:
- url
- twitter
- instagram
- ethnicity
- country
- eye_color
- height
- measurements
- fake_tits
- career_length
- tattoos
- piercings
- aliases
- gender
- scenes

### DIFF
--- a/pkg/models/querybuilder_performer.go
+++ b/pkg/models/querybuilder_performer.go
@@ -158,6 +158,45 @@ func (qb *PerformerQueryBuilder) Query(performerFilter *PerformerFilterType, fin
 		query.addArg(gender.Value.String())
 	}
 
+	if isMissingFilter := performerFilter.IsMissing; isMissingFilter != nil && *isMissingFilter != "" {
+		switch *isMissingFilter {
+		case "image":
+			query.addWhere("performers.image IS NULL")
+		case "url":
+			query.addWhere("performers.url IS NULL")
+		case "twitter":
+			query.addWhere("performers.twitter IS NULL")
+		case "instagram":
+			query.addWhere("performers.instagram IS NULL")
+		case "ethnicity":
+			query.addWhere("performers.ethnicity IS NULL")
+		case "country":
+			query.addWhere("performers.country IS NULL")
+		case "eye_color":
+			query.addWhere("performers.eye_color IS NULL")
+		case "height":
+			query.addWhere("performers.height IS NULL")
+		case "measurements":
+			query.addWhere("performers.measurements IS NULL")
+		case "fake_tits":
+			query.addWhere("performers.fake_tits IS NULL")
+		case "career_length":
+			query.addWhere("performers.career_length IS NULL")
+		case "tattoos":
+			query.addWhere("performers.tattoos IS NULL")
+		case "piercings":
+			query.addWhere("performers.piercings IS NULL")
+		case "aliases":
+			query.addWhere("performers.aliases IS NULL")
+		case "gender":
+			query.addWhere("performers.gender IS NULL")
+		case "scenes":
+			query.addWhere("scenes_join.scene_id IS NULL")
+		default:
+			query.addWhere("performers." + *isMissingFilter + " IS NULL")
+		}
+	}
+
 	handleStringCriterion(tableName+".ethnicity", performerFilter.Ethnicity, &query)
 	handleStringCriterion(tableName+".country", performerFilter.Country, &query)
 	handleStringCriterion(tableName+".eye_color", performerFilter.EyeColor, &query)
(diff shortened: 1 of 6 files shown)


bnkai:
closes #436

pkg/models/querybuilder_performer.go
```diff
@@ -158,6 +158,45 @@ func (qb *PerformerQueryBuilder) Query(performerFilter *PerformerFilterType, fin
 		query.addArg(gender.Value.String())
 	}
 
+	if isMissingFilter := performerFilter.IsMissing; isMissingFilter != nil && *isMissingFilter != "" {
+		switch *isMissingFilter {
+		case "image":
+			query.addWhere("performers.image IS NULL")
```
WithoutPants, line 164: Image is never null because it is required to be not null in the schema.

pkg/models/querybuilder_performer.go
```diff
@@ -158,6 +158,45 @@ func (qb *PerformerQueryBuilder) Query(performerFilter *PerformerFilterType, fin
...
+		case "piercings":
+			query.addWhere("performers.piercings IS NULL")
+		case "aliases":
+			query.addWhere("performers.aliases IS NULL")
+		case "gender":
+			query.addWhere("performers.gender IS NULL")
+		case "scenes":
+			query.addWhere("scenes_join.scene_id IS NULL")
+		default:
+			query.addWhere("performers." + *isMissingFilter + " IS NULL")
```
WithoutPants, line 196: Everything other than `"scenes"` seems to be the same as the default case. How about we get rid of all of the other cases above?

Anon247:
@WithoutPants any thoughts on duplicating the IsMissingCriterion?

pkg/models/querybuilder_performer.go
Anon247: Is there an easy way of identifying the default images directly in the database?

Anon247: Removed for now.

bnkai:
They seem to work ok.

ui/v2.5/src/models/list-filter/criteria/is-missing.ts
```diff
@@ -1,8 +1,8 @@
 import { CriterionModifier } from "src/core/generated-graphql";
 import { Criterion, CriterionType, ICriterionOption } from "./criterion";
 
-export class IsMissingCriterion extends Criterion {
-  public type: CriterionType = "isMissing";
+export class SceneIsMissingCriterion extends Criterion {
+  public type: CriterionType = "sceneIsMissing";
```
WithoutPants, line 5: In terms of abstracting, I did the following:

Keep `IsMissingCriterion` as an abstract class as follows:
```
export abstract class IsMissingCriterion extends Criterion {
  public parameterName: string = "is_missing";
  public modifierOptions = [];
  public modifier = CriterionModifier.Equals;
  public value: string = "";
}
```

Have the two concrete classes extend from `IsMissingCriterion` and remove the `parameterName`, `modifierOptions`, `modifier`, and `value` fields. 

I also changed the `is_missing` assignments in `filter.ts` to cast to `IsMissingCriterion` instead of the concrete classes.

I can push these changes myself if you prefer.

Functionally not much different, but I guess it saves some LOC.

Cutters1982:
Thanks guys, whilst this is being added, would functionality to bulk remove performers (ie. ones that have 0 scenes) be possible? Do you want me to raise that as a feature "issue" instead?

ui/v2.5/src/models/list-filter/criteria/is-missing.ts
Anon247: Yeah, good suggestions, I'm not going to be able to push these changes till tomorrow, so if you've got the time, go ahead 👍

Anon247:
Yes, this would be a separate PR so raise another issue for it 👍

VonTittyslappen:
Unfortunately something has broken, if I remove a date of birth and save, the DOB disappears, but when I refresh the DOB comes back, this used to work without issue, I think this commit has borked this somehow

WithoutPants:
> Unfortunately something has broken, if I remove a date of birth and save, the DOB disappears, but when I refresh the screen, the DOB comes back, this used to work without issue, I think this commit has borked this somehow

Tested on the version prior to this one and also got the same result, so this PR is not the cause. Please raise a bug for this. Not sure which commit may have caused this regression.

WithoutPants:
> Unfortunately something has broken, if I remove a date of birth and save, the DOB disappears, but when I refresh the screen, the DOB comes back, this used to work without issue, I think this commit has borked this somehow

Tested on the version prior to this one and also got the same result, so this PR is not the cause. Please raise a bug for this. Not sure which commit may have caused this regression.

PR #440, opened 2020-04-04
Replace basic auth with cookie authentication

### LINKED ISSUE #361: [Bug Report] Mobile UI doesn’t load when using authentication (IOS)
As of c99ba68 (and in prior builds), when using a password, the UI doesn't load on iPhone--either via Safari or Chrome.   No authentication input pops up, and no error is shown.  Page just loads as a blank white screen.  Problem persists when requesting "desktop site." Removing authentication allows UI to load.

Tried to inline auth info (using http://username:password@StashIP:Port), no change.

Using HTTP (not HTTPS).  Need to test whether problem persists w/ HTTPs.

iPhone 7, IOS 13.3.1, latest safari/chrome.

WithoutPants:
Replaces the existing basic authentication with cookie-based authentication.

If credentials are configured and no session is active, then the UI will redirect to a login page, which consists of only the following:

image

(key is a Chrome addition)

Entering invalid credentials gives an error as follows:

image

The login operation should redirect to the page that the user was attempting to access after logging in.

Cookies are JWT based and persist through server restarts.

Added a new configuration option: Maximum Session Age. This is the maximum age - in seconds - of the session cookie before it expires. The cookie is renewed during every request, so this option represents the maximum idle time before the session is expired.

Added a logout button next to the Settings button. This button is only shown when credentials are required and the user is logged in.

image

Clicking on this button expires the session cookie and returns the user to the login page.

Resolves #361.


pkg/api/server.go
```diff
@@ -38,37 +39,66 @@ var uiBox *packr.Box
 //var legacyUiBox *packr.Box
 var setupUIBox *packr.Box
 
+func allowUnauthenticated(r *http.Request) bool {
+	return r.URL.Path == "/login" || r.URL.Path == "/css" || strings.HasPrefix(r.URL.Path, "/setup")
```
bnkai, line 43: Adding /setup here poses a security risk
Anyone can access and mess with the setup even if a user/password is set

WithoutPants: Moved login files to a separate packr box and amended the line. No longer possible to access `/setup` without logging in.

PR #445, opened 2020-04-07
Added gqlgen command before React start & build

Anon247:
Keep forgetting to run gqlgen before building and starting react, so added a script to run it automatically.


WithoutPants:
We don't need to regenerate the graphql code everytime we run/build. It's as it is because graphql generation should only occur when the graphql is changed. It's a vote to close for me.

Issue #442, opened 2020-04-04
[Bug Report] Drop Down Arrow appears on Details page when it shouldnt do


VonTittyslappen:
**Describe the bug**
When viewing a performers details the drop down arrow appears on the gender box, which makes it look like it is clickable to change the gender, when it isn't

**To Reproduce**
1. Go to Performers
2. Click on one of the performers
3. In the details section, drop down arrow is visible

**Expected behavior**
Drop down arrow should only be visible in the edit tab where you can actually adjust it

**Screenshots**
n/a (though can be provided if requested)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 80.0.3987.149

PR #447, opened 2020-04-08
Performer and Movie UI fixes and improvements

WithoutPants:
- Sanitises URLs so that a valid URL is created in most cases: 
  - if string starts with a protocol (`http://` or `https://`) then URL is left unchanged
  - otherwise, prepend `https://`
  - for Twitter/Instagram URLs, prepend with `https://twitter.com/` if not present.
- Make gender dropdown consistent in the read-only performer view
- Show URLs as links in Performer and Movie read-only views
- Use duration input for Movie length
- Refactored `TableUtils` to use new `EditableTextUtils`


WithoutPants:
Fixes #442

WithoutPants:
Added fix for #450 

Fixes #450

bnkai:
The change for the movie duration makes it unable to be unset/cleared.
The `scene->file info->downloaded from` url could also be adjusted? 
The rest test out fine.

bnkai:
can you squeeze this  change also ?
```
diff --git a/ui/v2.5/src/utils/text.ts b/ui/v2.5/src/utils/text.ts
index e8e6c1e..b897c25 100644
--- a/ui/v2.5/src/utils/text.ts
+++ b/ui/v2.5/src/utils/text.ts
@@ -20,7 +20,7 @@ const fileSize = (bytes: number = 0, precision: number = 2) => {
     unit++;
   }
 
-  return `${bytes.toFixed(+precision)} ${Units[unit]}`;
+  return `${count.toFixed(+precision)} ${Units[unit]}`;
 };
 
 const secondsToTimestamp = (seconds: number) => {
```
it fixes the file  size in the scene file info tab

WithoutPants:
* Movie duration field fixed
* Scene URL display fixed
* Fixed `fileSize`

PR #448, opened 2020-04-08
Include gender in performer scraper results

WithoutPants:
Add Gender to performer scraper results, and set the gender in the UI after scraping.


WithoutPants:
Changed the UI code so that it matches gender based on the enum values (case insensitive) first, then the dropdown strings (case insensitive). 

This should allow for the following case insensitive strings:
* `female`
* `male`
* `intersex`
* `transgender male` and `transgender_male`
* `transgender female` and `transgender_female`

Once this is merged, the scrape configuration wiki page will need to be updated.

PR #454, opened 2020-04-09
Stash-box tagger integration

InfiniteTF:
This PR adds a scene tagger component which queries the stash-box instance at stashdb.org and allows for updating scenes with relevant metadata, and create studios/performers as needed. The tagger can query with full text search based on saved metadata or sanitised path/filename, or query by MD5. On save it submits the scenes hash/duration back to stashdb.org for future queries.

The performer creation allows the user to select the desired image from the images associated with the performer, and sets it on creation. The scene update also has an option to update the cover image from stashdb.org. The performer data is as far as possible fetched from freeones, minus tattoos/piercings since they don't really fit the stash-box model. Genders are for the most part correct, and male performers are by default not tagged, but can be tagged if desired.

Please note that this is still a WIP, and while it works well enough to where I've started tagging my collection, it has a few quirks that need ironing out. The scene selection is also somewhat limited since it's based on scraped data, and only has about 200k scenes atm.

Also note that there's a migration to add the stash_id to studio/performer/scenes, so backup your database as it will not work with the master/develop branches.


bnkai:
Something seems wrong with the pagination or the filters for the results.
Say your results are 3 pages and you  finish tagging the first page or part of the first, if you click to go to page 2 the pages displayed are 1,2. 3 is missing and if you click 1 you get 3 pages again but with page 3 being empty. The problem is that the missing results are not the ones you tagged already and so you have to refresh page 1 to get rid of the already tagged and get a correct page count

bnkai:
I had to export some data i wanted , so i added import/export functionality for the StashID field.
I have the changes as a git [diff](https://gist.github.com/bnkai/cda8d2b982e02f5fd8e431ad4b23352d) /  [patch](https://gist.github.com/bnkai/8a80fd1935bade6ba70cbf914179cc76) .

bnkai:
Just tested and noticed that when the scene image fetched from stashdb is webp the scene update fails
eg nsfw `https://cdn.stashdb.org/36/e4/36e49533-6671-4723-9b9b-73ca167f6618`
[params](https://pastebin.com/7x4Kf0QW)
[response](https://pastebin.com/AMR8RC1s)

InfiniteTF:
> Just tested and noticed that when the scene image fetched from stashdb is webp the scene update fails
> eg nsfw `https://cdn.stashdb.org/36/e4/36e49533-6671-4723-9b9b-73ca167f6618`
> [params](https://pastebin.com/7x4Kf0QW)
> [response](https://pastebin.com/AMR8RC1s)

Yeah that's a stash bug, we've not added webp support for decoding scene covers which we definitely should:
https://github.com/stashapp/stash/blob/develop/pkg/manager/scene_screenshot.go#L41

That said, stashdb wasn't supposed to serve a 1080p webp, I'd refactored some stuff and forgot to reenable image resizing. Any image over 720p will be reencoded to a 720p jpeg. It's been fixed now.

bnkai:
A minor bug found from a brief testing. Merge tags doesn't seem to work, it always overwrites.

WithoutPants:
Can you explain the design design behind adding `stash_id` to the database and graphql schemas?

InfiniteTF:
@WithoutPants The stash_id is there to ensure that the user only ever has to match an entity once. Once `Performer A` is tagged the stash_id `123-abc`, any future scenes tagged, regardless of performer aliases used, will automatically match to the correct local performer. It also allows for updating data in the future if entities are merged/updated since the stash_id is permanent.

WithoutPants:
I think we may need a more robust solution for this particular problem. For example, what happens if the user chooses to change their stash-box endpoint after having done some tagging with another endpoint already?

InfiniteTF:
That's a fair point. I'd only considered it from the standpoint of mirrored endpoints that would share the dataset and thus stash_ids, but it's not hard to imagine a separate stash-box instance with say only cam content that doesn't share ids but has performer overlap.

I think what I'll do is create a `stash_box` table with `endpoint, apikey` columns, and create {scene/performer/studio}_stash_id join tables with `entity_id`, `stash_id` and `stash_box_id` so entities can have multiple stash_ids. Should future proof things without complicating the implementation too much.

WithoutPants:
That seems like a reasonable approach. 

We should design any integration with stash-box under the principle that anyone can roll their own server of stash-box, and users may switch between them. Servers may go down, or users may choose to switch to a different server. We want to thus decouple the two systems as much as possible. 

Incidentally, there is a reference to stashdb in an apollo configuration file that I don't think should be in there, but I don't quite understand the purpose.

InfiniteTF:
Turned out to be a lot more work than I thought, but the implementation should be pretty solid now. I dropped the stash-box instance config from sqlite since a foreign key link to the stash_ids would've been messy if accidentally deleted. Instead I've just stuck the instance config in the regular config yaml, and made the endpoint explicit per stash_id so they can still be used for updating or whatever at a later point.

I've also added the stash_ids to the studio/performer/scene details, so it's easy to navigate out to the source. The only thing missing at this point is the ability to remove stash_ids, but that's not a blocker as far as I'm concerned.

Regarding `apollo.config.js`, it should be pointed at a stash-box instance for graphql type generation. In stash this is handled by gqlgen directly from the graphql definitions, but since the stash-box definitions aren't included in this project, we instead write graphql queries manually, and use `apollo generate` to talk to a live instance to generate the types. Not anything to worry about unless you're changing any stash-box graphql queries.

InfiniteTF:
Ok I lied, not being able to remove stashIDs was quite annoying, but fixed now.

ui/v2.5/src/components/Tagger/client.ts
```diff
@@ -0,0 +1,38 @@
...
+      if (forward) return forward(operation);
+      return null;
+    });
+
+    const link = middlewareLink.concat(httpLink);
+    const cache = new InMemoryCache();
+
+    setClient(
+      new ApolloClient({
+        name: "stashdb",
```
WithoutPants, line 29: I think this should be `stash-box`?

WithoutPants:
I'm not sure I'm on board the idea of having the logic for grabbing data from stash-box in the client. It really belongs in the server. Having it in the client means that we have no means manipulating the scraping process on the server side. It also delegates setting of the stash box ids to the client, which imo is the wrong approach to take.

WithoutPants:
Given that moving this into the server is going to take some significant effort, I think it'll be ok with going with the current approach, but I think that moving it into the server should be a priority for following iterations.

InfiniteTF:
I'm not familiar enough with the scraping system so say how easily this would fit into that. I'm guessing it would be a pretty big job, but it certainly makes sense to align all the different data fetching systems so they can share ui.

In any case, as you say it's probably better handled in a future iteration. There will almost certainly be more changes necessary once more people start using it, and that will likely inform the design of a more permanent graphql interface.

WithoutPants:
Looks like it doesn't handle Windows paths very well:

image

I can't work out what the blacklist is supposed to do. I have `XXX` in the blacklist but it still appears in the query? I think filename extensions should be stripped out of the generated query string altogether.

The top-level nav link should be moved somewhere into the scenes page.

It currently does not handle errors when saving the scene. Button just shows the spinner and no error is shown.

Selecting `Dir` mode caused the UI to crash.

I couldn't work out how the "Search Fingerprints" button was supposed to work.

Performer create dialog styling should match the other dialogs.

Would be nice if the performer dialog image selection would show a loading image while loading the selected image.

I'm going to spend some time seeing if I can prototype some server-side interfaces and some UI alternatives.

graphql/schema/types/stash-box.graphql
```diff
@@ -0,0 +1,32 @@
...
+  performer_id: ID!
+  endpoint: String!
+  stash_id: String!
+}
+
+type StudioStashID {
+  studio_id: ID!
+  endpoint: String!
+  stash_id: String!
+}
```
WithoutPants, line 27: I think these should all be a single type `StashID` and remove the id. The instances of these are all within the scope of their containing objects, so they are redundant anyway.

InfiniteTF: Good point, fixed.

InfiniteTF:
> Looks like it doesn't handle Windows paths very well:
> 

Completely slipped my mind. It should work now.


> I can't work out what the blacklist is supposed to do. I have `XXX` in the blacklist but it still appears in the query? I think filename extensions should be stripped out of the generated query string altogether.

The blacklist are regular expressions which are used to strip the query of stopwords that don't belong to the scene metadata. The XXX is a bit tricky since it actually has a leading and trailing space to avoid manglig studio names like xxxpawn, etc., so it only matches solo XXX fragments.


> It currently does not handle errors when saving the scene. Button just shows the spinner and no error is shown.

It should output errors. If you see it again, please let me know if there are any graphql errors, etc. I ran into two scenes just now that took over a minute each to update, so you may be running into a similar scenario. Haven't had a chance to debug, but I'll take a look tomorrow.

> Selecting `Dir` mode caused the UI to crash.

Fixed as part of the path parsing change.

> I couldn't work out how the "Search Fingerprints" button was supposed to work.

It fires off a fingerprint search for each scene to the instance, and shows the metadata if the fingerprint was found. Doesn't do much currently since my instance doesn't have many fingerprints, and certainly not oshash which it defaults to. I'm planning to change it to use a batch endpoint in stash-box btw, so you can fire off all 40 hashes and get any scene matches back as part of one query.

> Performer create dialog styling should match the other dialogs.

Done.

> Would be nice if the performer dialog image selection would show a loading image while loading the selected image.

Yup, done.


> The top-level nav link should be moved somewhere into the scenes page.

I've moved it into a scene panel as you suggested. Search is a big improvement, although I need to shuffle the config stuff and header around since it's a bit messy right now.

WithoutPants:
> I've moved it into a scene panel as you suggested. Search is a big improvement, although I need to shuffle the config stuff and header around since it's a bit messy right now.

This is a massive improvement imo.

> It should output errors. If you see it again, please let me know if there are any graphql errors, etc. I ran into two scenes just now that took over a minute each to update, so you may be running into a similar scenario. Haven't had a chance to debug, but I'll take a look tomorrow.

The server error I'm getting is related to an access denied error setting the thumbnail, which is a local issue and not something related to your change. However, the error does cause the button to remain in the spinner state. 

This is the error:
```
{"errors":[{"message":"open <snip>.thumb.jpg: Access is denied.","path":["sceneUpdate"]}],"data":{"sceneUpdate":null}}
```

FWIW, the console log is:
```
ApolloError.ts:57 Uncaught (in promise) Error: GraphQL error: open <snip>.thumb.jpg: Access is denied.
    at new t (ApolloError.ts:57)
    at Object.next (QueryManager.ts:225)
    at g (Observable.js:135)
    at w (Observable.js:179)
    at e.value (Observable.js:235)
    at observables.ts:12
    at Set.forEach (<anonymous>)
    at Object.next (observables.ts:12)
    at g (Observable.js:135)
    at w (Observable.js:179)
```

> I'm planning to change it to use a batch endpoint in stash-box btw, so you can fire off all 40 hashes and get any scene matches back as part of one query.

Let's look at getting that into #751.

I'd like to get #751 into a sufficient state to support the (or at least most of the) functionality in this PR and merge that first, before reworking the UI to use the stash server instead.

InfiniteTF:
Tweaked the config section so it's a bit cleaner, and moved the help text to the manual. It needs to be expanded, but I'll get to that later.

I've posted a PR for the stash-box batch fingerprint endpoint. Should be good to go. https://github.com/stashapp/stash-box/pull/23

InfiniteTF:
@WithoutPants I've rebased and squashed the commits so it's more manageable. I've also removed the remaining graphql stuff so it only talks to the stash backend.

Haven't got the batch query working yet since there's no ids or fingerprints in the scraped scenes to tie it back to the scenes, but it seems to work other than that.

PR #458, opened 2020-04-10
Add Studio to movie and fix movie schema

### LINKED ISSUE #456: [Bug Report] Movie schema issues
From #398:

The scene -> movie relationship table has a unique index on movie id and scene index, meaning that you cannot assign the same movie id/scene index combination to more than one scene. I think this unique constraint should be between movie id and scene id, not the scene index.

Movie name should be not null.

Duration and rating should be numeric.

WithoutPants:
Fixes #456 

Adds a new schema migration to correct the issues with the movie schema identified in #456.

Changes `duration` to an `Integer` instead of `varchar` and converts existing values using a custom function.
Changes `rating` to `tinyint` from `varchar(1)`.
Changes `scene_index` in `movies_scenes` from `varchar(2)` to `tinyint` and removed the unique index.
Makes the `name` field not null.

The graphql schema has been updated accordingly.

This PR is waiting on #447 to be merged so that the duration input change can be accomodated for.


bnkai:
Since you are making a copy migration do you mind reordering the rows so that blobs end up in the end?
( front/back cover image). As stated in #460 it is better perfomance wise to keep blobs at the end of the table.

WithoutPants:
> Since you are making a copy migration do you mind reordering the rows so that blobs end up in the end?

Done. I've also added studio to movies. I've left producer off for now as I think this can be covered by studio.

bnkai:
A couple of bugs
If you set the rating from a movie eg 2 , you can't unset it. ( rating gets sent as null from the ui)

Something doesn't work with the duration in the ui. You can't edit it. It shows as hh:mm:ss but it only accepts input 00:00 that you cant increase/decrease and converts to numbers that don't make sense.

WithoutPants:
Also looks like Studio can't be unset.

bnkai:
Everything looks ok now.

Issue #452, opened 2020-04-08
[Bug Report] Performer list slow loading on scene edit page, causes DB to be locked when making fast edits (large database!)


VonTittyslappen:
**Describe the bug**
Since the new UI interface with the drop down list of Performers, when going to the edit page of a scene, the performer list is slow at loading (the 3 flickering dots appear for loading), this means that if you are making a fast edit on a scene page and try and save it the database is locked, as its taking a few seconds to load the performer list (sometimes takes up to 10 seconds. Whenever I go to an edit page on a scene I can hear my NAS scanning really hard to pull a list of the performers. 

**Expected behavior**
Performer list should be more instantaneous. Although I have 1900+ performers added, there needs to be a better method of this, I keep facing database locked messages because the performer list is taking too long to appear.

Happy to share my database for investigative purposes :)

**Desktop (please complete the following information):**
 - OS: Windows 10 64-Bit (Latest non-beta update)
 - Browser Chrome 80.0.3987.163 (Official Build) (64-bit)
 - Stash version v0.1.1-54-gdc37a30 (Build hash: | dc37a30)

PR #460, opened 2020-04-10
Add slim endpoints for entities to speed up filters

InfiniteTF:
Should resolve issue #452 

In my library with 123 performers and 1MB-ish performer images it speeds up AllPerformersForFilters from 0.4s to 0.04s.  For 549 studios it speeds up from 0.08s to 0.02s.

I considered changing the dropdowns to fetch on demand, but considering that scenes can have arbitrary amount of performers attached, it would likely make page load significantly slower in many instances.

### DIFF
--- a/pkg/database/migrations/7_performer_optimization.up.sql
+++ b/pkg/database/migrations/7_performer_optimization.up.sql
@@ -1,8 +1,7 @@
-PRAGMA foreign_keys = OFF;
 DROP INDEX `performers_checksum_unique`;
 DROP INDEX `index_performers_on_name`;
 DROP INDEX `index_performers_on_checksum`;
-ALTER TABLE performers RENAME TO temp_old_performers;
+ALTER TABLE `performers` RENAME TO `temp_old_performers`;
 CREATE TABLE `performers` (
   `id` integer not null primary key autoincrement,
   `checksum` varchar(255) not null,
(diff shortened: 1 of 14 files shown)


WithoutPants:
Testing against a database with 1,968 performers, I'm not seeing any difference before and after this change. I'm seeing between 7.5 and 12 seconds for `allPerformersSlim` to come back.

InfiniteTF:
@WithoutPants Try now. I've added a migration that moves the `image` column to the end of the table. Apparently selecting columns past blobs incurs a large penalty in sqlite. On my instance it reduces the allPerformers query further down to 0.01s, though it seems my dataset isn't large enough to really stress the database at this point. I'm guessing at a certain size it spills to disk, causing the crazy query times.

bnkai:
I was getting consisntent speedup in the edit scene menu as stated above also. @WithoutPants is your db maybe in a spinning disk? ( my tests were in sata ssd and nvme ssd)

bnkai:
With latest commit : 
Database located on a nvme ssd
Edit Scene Tab
Time for AllPerformers
500+ performers  
before 800ms after 15-20 ms
1k+ performers
before 1150ms  after 25-30 ms

WithoutPants:
I'm getting consistent times of around 50ms for `allPerformersSlim` now. Looks good.

pkg/database/migrations/7_performer_optimization.up.sql
```diff
@@ -0,0 +1,102 @@
...
+  `piercings` varchar(255),
+  `aliases` varchar(255),
+  `favorite` boolean not null default '0',
+  `created_at` datetime not null,
+  `updated_at` datetime not null,
+  `image` blob not null
+);
+CREATE UNIQUE INDEX `performers_checksum_unique` on `performers` (`checksum`);
+CREATE INDEX `index_performers_on_name` on `performers` (`name`);
+CREATE INDEX `index_performers_on_checksum` on `performers` (`checksum`);
```
WithoutPants, line 31: There's a redundant index on checksum here (it was in the original schema). I think we can safely remove the non-unique index.

bnkai: I just checked also. _index_performers_on_checksum_ is used now for queries but when dropped _performers_checksum_unique_ gets used instead. So safe to remove.

InfiniteTF: Good point, I've removed `index_performers_on_checksum`.

PR #462, opened 2020-04-12
Parent studio support

Anon247:
Added initial draft migration schema. Input appreciated.


bnkai:
Am i missing something for the migration ?
Isn't something like 
```
ALTER TABLE studios ADD COLUMN parent_id INTEGER DEFAULT NULL CHECK ( id IS NOT parent_id ) REFERENCES studios(id) ;
CREATE INDEX index_studios_on_parent_id on studios (parent_id);
```

not enough ?

bnkai:
Also when adding functions to the schema you need to implement them, even as empty functions at the beginnung. You can have a look at the _models/generated_xxxx.go_ files after the _make generate_ stage. From a first look you need to implement `FindParentStudio(ctx context.Context, parentID string) (*Studio, error)` and `FindChildrenStudios(ctx context.Context, id string) (*FindStudiosResultType, error)` in _api/resolver_query_find_studio.go_ and `ParentID(ctx context.Context, obj *Studio) (*string, error)` in _resolver_model_studio.go_
You'll also need to adjust _models/model_studio.go_

Anon247:
> Am i missing something for the migration ?
> Isn't something like
> 
> ```
> ALTER TABLE studios ADD COLUMN parent_id INTEGER DEFAULT NULL CHECK ( id IS NOT parent_id ) REFERENCES studios(id) ;
> CREATE INDEX index_studios_on_parent_id on studios (parent_id);
> ```
> 
> not enough ?

You're correct. I was aware of issues with altering tables, specifically with adding foreign keys in SQLite, but the syntax you provided works.

pkg/models/model_studio.go
```diff
@@ -10,6 +10,7 @@ type Studio struct {
 	Checksum  string          `db:"checksum" json:"checksum"`
 	Name      sql.NullString  `db:"name" json:"name"`
 	URL       sql.NullString  `db:"url" json:"url"`
+	ParentId  int             `db:"parent_id" json:"parent_id"`
```
bnkai, line 13: Wouldnt  `sql.NullInt64` be more relevant as a type for ParentID ? In  cases with fks (eg scene,gallery) we use `sql.NullInt64` afaik

bnkai: Edit from the  integration test in travis log it seems that the type chage is needed anyway
`panic: Could not populate database: sql: Scan error on column index 7, name "parent_id": converting driver.Value type <nil> ("<nil>") to a int: invalid syntax`
Have a look at the scene code to see how sql.NullInt64 is handled

WithoutPants:
Thanks for the contribution @Anon247 . This feature has been continued in #595

Issue #473, opened 2020-04-16
[Feature] Gallery Grid View & Cover Picture


f1delio:
Currently, the gallery only shows a list view and automatically selected cover image.

**Grid View** would show more galleries at the same time making it easier to choose and navigate.

**Cover Image**
Instead of implementing something complex, I have a proposal for this simple solution in pseudo-code:

IF zip file.contains (image called "cover.jpg") THEN
use it as a cover image for gallery.
ELSE
do the thing it is doing now.
END IF

This way users could easily make a cover image for the gallery by just renaming one image.

WithoutPants:
Implemented via #622

PR #477, opened 2020-04-18
Move image with cover.jpg in name to first place in Galleries

bnkai:
Change file order in zip files to put image with _cover.jpg_ included in name at first position.
(image ending in cover.jpg  for example)

Related to #473

### DIFF
--- a/pkg/models/model_gallery.go
+++ b/pkg/models/model_gallery.go
@@ -88,7 +88,7 @@ func (g *Gallery) readZipFile(index int) ([]byte, error) {
 func (g *Gallery) listZipContents() ([]*zip.File, *zip.ReadCloser, error) {
 	readCloser, err := zip.OpenReader(g.Path)
 	if err != nil {
-		logger.Warn("failed to read zip file %s", g.Path)
+		logger.Warnf("failed to read zip file %s", g.Path)
 		return nil, nil, err
 	}
 
--- a/pkg/models/model_gallery.go
+++ b/pkg/models/model_gallery.go
@@ -113,7 +113,10 @@ func (g *Gallery) listZipContents() ([]*zip.File, *zip.ReadCloser, error) {
 	})
 	cover := Contains(filteredFiles, "cover.jpg") // first image with cover.jpg in the name
 	if cover >= 0 {                               // will be moved to the start
-		filteredFiles = Reorder(filteredFiles, cover)
+		reorderedFiles := Reorder(filteredFiles, cover)
+		if reorderedFiles != nil {
+			return reorderedFiles, readCloser, nil
+		}
 	}
 
 	return filteredFiles, readCloser, nil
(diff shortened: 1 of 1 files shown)


pkg/models/model_gallery.go
```diff
@@ -111,6 +111,42 @@ func (g *Gallery) listZipContents() ([]*zip.File, *zip.ReadCloser, error) {
...
+		reorderedFiles := Reorder(filteredFiles, cover)
+		if reorderedFiles != nil {
+			return reorderedFiles, readCloser, nil
+		}
+	}
 
 	return filteredFiles, readCloser, nil
 }
+
+func Contains(a []*zip.File, x string) int { // return index of first occurenece of string x in name of zip contents
```
WithoutPants, line 125: Does this need to be public (capitalised)?

Also, documentation for the function belongs on the line above.

pkg/models/model_gallery.go
```diff
@@ -111,6 +111,42 @@ func (g *Gallery) listZipContents() ([]*zip.File, *zip.ReadCloser, error) {
...
+	for i, n := range a { // -1 otherwise
+		if strings.Contains(n.Name, x) {
+			logger.Debugf("Cover found in zip file %s.Position:%d", n.Name, i)
+			return i
+		}
+	}
+	return -1
+}
+
+func Reorder(a []*zip.File, toFirst int) []*zip.File { // reorder slice so that element with position toFirst gets at the start
```
WithoutPants, line 135: As above. I think there may be a more efficient way of doing this via slice ranges instead of iterating through the entire slice.

pkg/models/model_gallery.go
bnkai: wasn't sure if it will useful in the future. fixing

pkg/models/model_gallery.go
bnkai: The order needs to be kept after the removal of the element so a copy is needed. I doubt that the difference will be that much. Nevertheless  fixing that also.

pkg/models/model_gallery.go
```diff
@@ -112,5 +112,42 @@ func (g *Gallery) listZipContents() ([]*zip.File, *zip.ReadCloser, error) {
...
+	}
+
 	return filteredFiles, readCloser, nil
 }
+
+// return index of first occurenece of string x in name of zip contents, -1 otherwise
+func contains(a []*zip.File, x string) int {
+	for i, n := range a {
+		if strings.Contains(n.Name, x) {
+			logger.Debugf("Cover found in zip file %s.Position:%d", n.Name, i)
```
WithoutPants, line 130: This gets output a lot. Do we need it?

bnkai: I thought it was only printed when log level is set to debug. Not needed now though so removing.

PR #478, opened 2020-04-18
Query optimizations

InfiniteTF:
This PR optimizes performer and scene queries by doing a few things:
* Selecting only columns that are actually used by graphql. The primary purpose of this is excluding performer images and scene covers from the queries since they are never used. 
* Fetching all entity data at once instead of first querying for ids, and then querying each individual id for the data. Despite the sqlite manual's claims, doing a single query is significantly faster, at least in my tests.
* Removing pointless joins. Many places entity tables are joined to do counts and existence checks, despite the join table alone being sufficient. E.g: Joining performers with scenes through performer_scenes and counting scenes.id, rather than just counting performers_scenes.scene_id. 

The performance gains for my test cases have been:
120 Performers: ~560ms -> 30ms.
60 Scenes with performers and custom covers: 220ms -> 70ms.

Note that this is with the migration to move the performer image to the last column. Otherwise the difference for performers is likely less significant.

The scene query is harder to optimize due to a larger number of joins for each entity, but it's still a nice gain.

The diff is a bit hard to follow, and I didn't test all the filters extensively, so it's possible I missed something, but generally speaking there should be no changes other than the performance improvements.

### DIFF
--- a/pkg/models/querybuilder_scene.go
+++ b/pkg/models/querybuilder_scene.go
@@ -10,33 +10,67 @@ import (
 	"github.com/stashapp/stash/pkg/database"
 )
 
-const scenesForPerformerQuery = `
-SELECT scenes.* FROM scenes
+var sceneMetadataColumns = []string {
+    "id",
+    "checksum",
+    "title",
+    "details",
+    "url",
+    "date",
+    "rating",
+    "size",
+    "duration",
+    "video_codec",
+    "audio_codec",
+    "format",
+    "width",
+    "height",
+    "framerate",
+    "bitrate",
+    "studio_id",
+    "created_at",
+    "updated_at",
+}
+
+func selectSceneMetadata() string {
+    var str strings.Builder
+
+    for _, colName := range sceneMetadataColumns {
+        str.WriteString("scenes." + colName + ", ")
+    }
+
+    returnVal := str.String()
+    return "SELECT " + returnVal[:len(returnVal)-2] + " FROM scenes "
+}
+
+
+
+var scenesForPerformerQuery = selectSceneMetadata() + `
 LEFT JOIN performers_scenes as performers_join on performers_join.scene_id = scenes.id
-LEFT JOIN performers on performers_join.performer_id = performers.id
-WHERE performers.id = ?
+WHERE performers_join.performer_id = ?
 GROUP BY scenes.id
 `
 
-const scenesForStudioQuery = `
-SELECT scenes.* FROM scenes
+var countScenesForPerformerQuery = `
+SELECT performer_id FROM performers_scenes as performers_join
+WHERE performer_id = ?
+GROUP BY scene_id
+`
+
+var scenesForStudioQuery = selectSceneMetadata() + `
 JOIN studios ON studios.id = scenes.studio_id
 WHERE studios.id = ?
 GROUP BY scenes.id
 `
-const scenesForMovieQuery = `
-SELECT scenes.* FROM scenes
+var scenesForMovieQuery = selectSceneMetadata() + `
 LEFT JOIN movies_scenes as movies_join on movies_join.scene_id = scenes.id
-LEFT JOIN movies on movies_join.movie_id = movies.id
-WHERE movies.id = ?
+WHERE movies_join.movie_id = ?
 GROUP BY scenes.id
 `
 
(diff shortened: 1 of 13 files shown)


WithoutPants:
Looks like its failing the unit tests. Can you take a look at the results of `make it`?

pkg/models/querybuilder_scene.go
```diff
@@ -10,33 +10,67 @@ import (
...
+	"audio_codec",
+	"format",
+	"width",
+	"height",
+	"framerate",
+	"bitrate",
+	"studio_id",
+	"created_at",
+	"updated_at",
+	"o_counter",
```
bnkai, line 34: "cover"
is missing

InfiniteTF: I've reverted the change to `qb.All` which is what the export task uses, so it should get the cover again.

bnkai: confirmed , export is fine now.

WithoutPants:
I really don't like the idea of changing the queries to get a static list of column names, but it'll have to do for now until we get rid of the blobs in the schema. I don't like having an extra place to remember to maintain columns for objects.

WithoutPants:
Just as an addendum to last comment, I think having an arbitrary set of columns in the `Find` functions are a bad idea. It is non-obvious to the calling code just what is being returned. 

My feeling is that the column changes should be dropped for this PR, so that we can get the performance gain of getting rid of the unnecessary joins, and perform the column stuff in a separate PR. The design of the query builder PR needs to be iterated on further imo. 

If we absolutely must include it in this PR, my thinking would be that the existing find interfaces should be left as they originally were (getting all columns) and adding a new interface that accepts a list of columns. That way, the calling code has control over what columns it queries for, and makes it more explicit that the graphql code will be querying for specific columns.

InfiniteTF:
I agree that it's not a great solution. Ideally the resolvers would figure out the required columns based on the query and only fetch those, but that would probably make for a really hairy resolver.

Unfortunately once I revert the column changes, the performance gains largely evaporate. I'll try to come up with a cleaner solution, but as you say the query builders should probably be refactored altogether, and I don't think I'm the right person to do that.

WithoutPants:
Assuming we remove the blobs from the tables, is there going to be an appreciative performance difference between doing `select *` vs `select <specific columns>`?

If there is, then it makes sense to refactor the query functions to accept columns. Otherwise, I would suggest changing the unnecessary joins and leaving the column selection as is with the view that the blob stuff should be done as priority for the next release.

I'm going to submit a PR to your branch with an alternative approach, but I am hesitant to merge it into 0.2 presuming we release reasonably soon, since there's a decent risk of regression.

WithoutPants:
I've had a quick go at it. You can take a look at the branch here: https://github.com/WithoutPants/stash/tree/select-columns

I still don't really like the idea in general. When we make a call that gets a `Scene` or `Performer` object, we should really be fully populating the object, because otherwise other functions have no idea how much an object has been populated. I can see this being a massive maintenance headache.

InfiniteTF:
Column changes have been reverted, as have the bulk fetching since it proved to be slower than fetching everything one by one without the column change.

Also removed some groupBys in performer findByScene since it's really slow and doesn't make much sense anyway. If someone for some reason attaches the same performer multiple times  to a scene then we should just return that.

Finally, refactored scene query to use the querybuilder because why not.

WithoutPants:
Changes look ok and tested ok during my quick testing. #513 is now merged, so if you can merge in the develop branch then we can ensure that at least the new unit tests pass.

pkg/models/querybuilder_scene.go
```diff
@@ -10,7 +10,9 @@ import (
 	"github.com/stashapp/stash/pkg/database"
 )
 
-var scenesForPerformerQuery = selectAll("scenes") + `
+const TABLE_NAME = "scenes"
```
WithoutPants, line 13: Golang constants are visible module-wide, so `TABLE_NAME` is not a good constant name for a scene-specific value. Suggest changing this to `sceneTable` or similar.

InfiniteTF: Duh, I keep forgetting that go is basically C, scope and all. Fixed.

PR #482, opened 2020-04-19
Don't error in IsStreamable resolver

WithoutPants:
`IsStreamable` resolver emits an error if it cannot read the source file. Makes the entire graphql request fail if the source file can't be read. This means that the UI fails when loading up a scene where the file is missing. Also means I can't browse scenes from someone else's database ;)


bnkai:
Ideally there shouldn't be stale/missing files , clean should remove those from DB ( if run )  :-)
But we get `false`  from  manager.IsStreamable if something errors out so error shouldn't be needed. 
An edge case would be read error / hdd error (file is present but can't be read) but that would be mitigated to the transcode process anyway so i think it's safe to ignore

WithoutPants:
From an API point of view, the absence of the file shouldn't prevent getting information (including is streamable) about the scene. In future, perhaps a `missing` graphql field with a suitable resolver would be worth adding.

PR #489, opened 2020-04-22
Clean missing galleries

WithoutPants:
Changed the clean operation so that it also checks and removes galleries that are no longer present. Changed the clean dialog message accordingly.

### DIFF
--- a/pkg/manager/manager_tasks.go
+++ b/pkg/manager/manager_tasks.go
@@ -519,8 +519,8 @@ func (s *singleton) Clean() {
 			wg.Wait()
 		}
 
-		for _, gallery := range galleries {
-			s.Status.incrementProgress()
+		for i, gallery := range galleries {
+			s.Status.setProgress(len(scenes)+i, total)
 			if s.Status.stopping {
 				logger.Info("Stopping due to user request")
 				return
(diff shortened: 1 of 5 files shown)


pkg/manager/manager_tasks.go
```diff
@@ -506,7 +514,26 @@ func (s *singleton) Clean() {
...
 			wg.Add(1)
 
-			task := CleanTask{Scene: *scene}
+			task := CleanTask{Scene: scene}
+			go task.Start(&wg)
+			wg.Wait()
+		}
+
+		for _, gallery := range galleries {
+			s.Status.incrementProgress()
```
bnkai, line 523: s.Status.incrementProgress() 
shouldnt this be something like s.Status.setProgress(len(scenes)+j, total) ?
where j the for index for gallery
If you want to use increment shouldn't the Task total be set first and increment also used for scenes?

WithoutPants: I can do that. It was just done to not use `len`. The calls to `setProgress` set the total.

bnkai: i didn't notice that the status total was updated that way.
you can leave it as is then

bnkai:
Not related to your changes
Can you replace in `pkg/manager/exclude_files.go` function `matchFile`
`logger.Infof("No exclude patterns in config.")`
with
`return false` ?
I noticed that it pollutes the log if no exclude pattern is set in the config.

Everything tests out ok

PR #490, opened 2020-04-22
Add index/total count to end of pagination buttons

FleetingOrchard:
Very simple, just chucks a fake button on the end of the pagination buttons. Looked a bit awkward by itself on the no-pagination case, so left it out there, but hopefully most folks know how to count up to 40.

### DIFF
--- a/ui/v2.5/src/components/List/Pagination.tsx
+++ b/ui/v2.5/src/components/List/Pagination.tsx
@@ -121,5 +121,5 @@ export const PaginationIndex: React.FC<IPaginationIndexProps> = ({
   const lastItemCount:number = Math.min(firstItemCount+(itemsPerPage-1), totalItems);
   const indexText:string = `${intl.formatNumber(firstItemCount)}-${intl.formatNumber(lastItemCount)} of ${intl.formatNumber(totalItems)}`;
 
-  return <span className="filter-container paginationIndex" onClick={onClick}>{indexText}</span>
+  return <span className="filter-container text-muted paginationIndex" onClick={onClick}>{indexText}</span>
 };
--- a/ui/v2.5/src/components/List/Pagination.tsx
+++ b/ui/v2.5/src/components/List/Pagination.tsx
@@ -9,20 +9,20 @@ interface IPaginationProps {
   onChangePage: (page: number) => void;
 }
 
+interface IPaginationIndexProps {
+  itemsPerPage: number;
+  currentPage: number;
+  totalItems: number;
+  onClick?: () => void;
+}
+
 export const Pagination: React.FC<IPaginationProps> = ({
   itemsPerPage,
   currentPage,
   totalItems,
   onChangePage,
 }) => {
-  const intl = useIntl();
-
   const totalPages = Math.ceil(totalItems / itemsPerPage);
-
-  // Build the pagination index string
-  const firstItemCount:number = Math.min((currentPage-1)*itemsPerPage+1, totalItems);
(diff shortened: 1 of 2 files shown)


ui/v2.5/src/components/List/Pagination.tsx
```diff
@@ -16,6 +16,11 @@ export const Pagination: React.FC<IPaginationProps> = ({
 }) => {
   const totalPages = Math.ceil(totalItems / itemsPerPage);
 
+  // Build the n-m/total string
+  const firstItemCount:number = Math.min((currentPage-1)*itemsPerPage+1, totalItems);
+  const lastItemCount:number = Math.min(firstItemCount+(itemsPerPage-1), totalItems);
+  const indexText:string = `${firstItemCount}-${lastItemCount}/${totalItems}`;
```
InfiniteTF, line 22: You don't actually need to specify the type since the type is implied from the initialization.

InfiniteTF, line 22: `1-40/360` is a bit hard to parse for me. I would prefer `1-40 of 360`.

ui/v2.5/src/components/List/Pagination.tsx
```diff
@@ -96,6 +101,12 @@ export const Pagination: React.FC<IPaginationProps> = ({
         <span className="d-none d-sm-inline">Last</span>
         <span className="d-inline d-sm-none">&#x300b;</span>
       </Button>
+      <Button
+        variant="secondary"
+        disabled={true}
+      >
+        {indexText}
+      </Button>
```
InfiniteTF, line 109: Not entirely sure how I feel about a disabled button. I honestly think it looks fine with just a `<div>`, it strongly implies that it's not part of the actual navigation.

ui/v2.5/src/components/List/Pagination.tsx
FleetingOrchard: Good call, actually. That's been implemented in 15befa1.

FleetingOrchard: I know, my C days just give me anxiety regarding inferred types.

ui/v2.5/src/components/List/Pagination.tsx
FleetingOrchard: Eh, I disagree, but more fundamentally -- I think an index is part of the navigation of a paginated structure. Yes, it's not a functional part, but it conveys key information about navigation to the user.

WithoutPants:
I am inclined to agree with @InfiniteTF on the placement of the index/total count. I think we should probably be looking at other sites and apps for guidance on this one. 

PH places it at the top of the page, just before the results. This is probably the ideal in my opinion:

image

Google just shows the page and the total, meh:

image

Radarr just gives you the total:

image

As it is, it just looks a little tacked-on:

image

It also looks broken in mobile view:

image

My preference would be a small section above the results, like the PH example.

FleetingOrchard:
My only concern there is that I feel we don't really have any UI elements that are analogous to a header (in the sense of results-for-query-foobar). Putting it between the filter and the content might look a bit weird as a result.

Maybe putting it between the content and pagination in ListHook? I'd be inclined to then just break it out into its own component to facilitate that, though. On the other hand, I was also thinkin' we could use it as a way to pop up a modal so users could skip to a particular range, and I don't know if it's a great idea to influence pagination _outside_ pagination.

Looking at the mobile view, and how it's implemented in Pagination.tsx -- I'm inclined to suggest we shorten the index string (dropping " of " in favor of "/") for mobile views, but I have to confess, I have no idea how that's achieved with these classNames on those first/last buttons.

PR #491, opened 2020-04-22
Add random male performer image

bnkai:
Extends the random image to performers if image is missing  functionality for male performers.
If no gender is given or gender is not male it defaults to female image.
Images used are below.

noname_male_01
noname_male_02


pkg/api/resolver_mutation_performer.go
```diff
@@ -19,7 +19,11 @@ func (r *mutationResolver) PerformerCreate(ctx context.Context, input models.Per
 	var err error
 
 	if input.Image == nil {
-		imageData, err = getRandomPerformerImage()
+		if input.Gender != nil {
+			imageData, err = getRandomPerformerImage(input.Gender.String())
+		} else {
+			imageData, err = getRandomPerformerImage("")
+		}
```
WithoutPants, line 26: Can we change this to be:
```
gender := ""
if input.Gender != nil {
	gender = input.Gender.String()
}
imageData, err = getRandomPerformerImage(gender)
```

Less verbose and less prone to maintenance errors.

PR #493, opened 2020-04-23
Refactor build

### LINKED ISSUE #492: [Bug Report] Version not included in latest builds
**Describe the bug**
The version is not being embedded correctly in the latest builds. This results in the following log message:

```
Error while retrieving latest hash: Stash doesn't have a version. Version check not supported.
```

The travis log includes the following:

```
 sh ./scripts/cross-compile.sh
fatal: No names found, cannot describe anything.
```

This is occurring due to travis cloning with a depth of 50. We'll need to disable the shallow clone to fix this.

WithoutPants:
Fixes #492 

- Adds `depth: false` to the travis configuration so that tags are correctly gathered
- Perform linting and checking of go and UI files. Build will fail if format is required. I've left `lint` off for now since that requires getting an extra dependency.
- Added new make targets: `ui-pre`, `packr`, `fmt` and `fmt-check`
- Changed `cross-compile.sh` to run packr2 once, then go build for each platform. Should result in a quicker and less verbose build.

Note that this PR will require the files to have linting and formatting issues fixed, and will fail until this is done. I've submitted it now to test the travis build.


bnkai:
I've gone through the script changes and they look ok.
I guess we need to push  and see.

The only thing i needed  to do is install revive for the lint , that i didn't have ( saw the comment on travis.yml)

bnkai:
Another thing that probalby needs to change is the first line of the travis.yml
`if: tag != develop_latest # dont build for the latest_develop tagged version`
i may have made a copy/paste error when adding that
i am pretty sure
`if: tag != latest_develop  # dont build for the latest_develop tagged version`
is needed
Can you make the change to test the travis check ?

PR #494, opened 2020-04-23
Fix yarn extract error due to "library size"

bnkai:


WithoutPants:
What is this fixing?

bnkai:
When making the library size PR i forgot that @InfiniteTF had added the internationalisation feature. Yarn extract is used to extract formatted messages that we then can then translate. It expects that those messages are static and so fails for the file size part of the library in stats page.

PR #496, opened 2020-04-24
Add a cache for gallery thumbnails

bnkai:
Uses the cache dir to store  gallery thumbnails. 
A new cache subfolder **gthumbs** is created with a tree structure and one folder (named from the checksum of the gallery) for every gallery accessed. 
Everytime a thumbnail  request is sent to the gallery router it first checks the cache and then creates it if needed. If a thumbnail is created it gets stored for  further use.
For users that want the thumbnails created ahead of time there is an extra generate task added. By default i left it unticked as it isn't required for the cache to function. ( took the liberty of  unchecking transcodes by default also)


bnkai:
The cache size check / purge is done at startup before the cache starts to read/write files as it was the easiest and safest to implement.
The thumb cache size option was added to the UI also and if its set to > 0 it "touches" the thumbnails to make the cache purge LRU. By default it is set to 0 ( no limit ) as with that we don't need to "touch" the  thumbnails.
The main cache path was missing from the UI config options so i added that too.
The root path used for the  thumbnails cache will be  cache+"gthumbs".

When scanning new galleries, if a zip file is detected with compression level not set to 0  a warning will be displayed.

pkg/utils/file.go
```diff
@@ -66,7 +76,12 @@ func EnsureDir(path string) error {
 	return err
 }
 
-// RemoveDir removes the given file path along with all of its contents
+// EnsudirAll will create a directory at the given path along with any necessary parents if they don't already exist
```
WithoutPants, line 79: Typo
```suggestion
// EnsureDirAll will create a directory at the given path along with any necessary parents if they don't already exist
```

bnkai:
Updated  the fist post , should be ready for review.

graphql/schema/types/metadata.graphql
```diff
@@ -3,6 +3,7 @@ input GenerateMetadataInput {
   previews: Boolean!
   markers: Boolean!
   transcodes: Boolean!
+  thumbnails: Boolean!
```
WithoutPants, line 6: This should either be renamed or clarified in a comment that this refers to gallery thumbnails, since this could refer to any number of thumbnail type.

pkg/manager/manager_tasks.go
```diff
@@ -191,10 +192,20 @@ func (s *singleton) Generate(sprites bool, previews bool, markers bool, transcod
...
+
 		delta := utils.Btoi(sprites) + utils.Btoi(previews) + utils.Btoi(markers) + utils.Btoi(transcodes)
 		var wg sync.WaitGroup
 		s.Status.Progress = 0
-		total := len(scenes)
+		lenScenes := len(scenes)
+		total := lenScenes
+		if thumbnails {
+			total += len(galleries)
+		}
```
WithoutPants, line 208: The `qg.All()` call should be moved in here so that the query is only run if we're generating thumbnails.

pkg/manager/paths/paths_gallery.go
```diff
@@ -15,6 +20,31 @@ func (gp *galleryPaths) GetExtractedPath(checksum string) string {
...
+func getIntraDir(checksum string) string {
+	if thumbDirDepth == 0 {
+		return "" // filepath.Join ignores empty elements
+	}
+	intraDir := checksum[0:thumbDirLength]
+	for i := 1; i < thumbDirDepth; i++ {
+		intraDir = filepath.Join(intraDir, checksum[thumbDirLength*i:thumbDirLength*(i+1)])
+	}
+	return intraDir
+}
```
WithoutPants, line 37: Can this be moved into a general utility function in the utils module with a little bit more detail in the comment?
Line 34 could use a comment explaining what it is doing.

pkg/utils/file.go
```diff
@@ -133,3 +148,132 @@ func GetHomeDirectory() string {
 	}
 	return currentUser.HomeDir
 }
+
+func PrintDuDetails(files []DuDetails) {
+	for i, file := range files {
+		fmt.Printf("%d: %s size: %d mod date: %s\n", i, file.path, file.size, file.mtime.Format("2006-01-02T15:04:05"))
+	}
+}
```
WithoutPants, line 156: Is this used anywhere? Debug function? Should probably be removed or converted to log or return a string.

pkg/utils/file.go
```diff
@@ -133,3 +148,132 @@ func GetHomeDirectory() string {
...
+// Sort DuDetails slice by mod time, older first
+func SortDuDetailsByMtime(files []DuDetails) {
+	sort.Slice(files, func(i, j int) bool {
+		return files[i].mtime.After(files[j].mtime)
+	})
+}
+
+// DuDir returns the size of the directory (only actual files are counted)
+// The filesInfo slice is populated during the recursion
+func DuDir(path string, info os.FileInfo, filesInfo *[]DuDetails) int64 {
```
WithoutPants, line 167: Are these functions used any more? Is there any value in keeping them?

pkg/utils/file.go
bnkai: removed as all of du functions.
not needed anymore

pkg/manager/paths/paths_gallery.go
bnkai: done

graphql/schema/types/metadata.graphql
bnkai: added comment

pkg/manager/manager_tasks.go
bnkai: done

PR #497, opened 2020-04-24
Add "reshuffle button" when sortby is random

FleetingOrchard:
Button, when pressed, triggers the regen of a new seed for the list filter, effectively reshuffling the list.


WithoutPants:
Just ran `yarn run format` to fix up the formatting. Looks good.

PR #500, opened 2020-04-24
Add parser support for 3-letter month

WithoutPants:
Adds an `{mmm}` parser field to the scene filename parser. Matches against three letter month such as `Jan` (case-insensitive).

### DIFF
--- a/pkg/manager/filename_parser.go
+++ b/pkg/manager/filename_parser.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/stashapp/stash/pkg/models"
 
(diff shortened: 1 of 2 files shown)


bnkai:
Looks ok, works ok also.
A small observation
If month is supported like this shouldn't the day also be supported in single digit?
Apr.5.2020 for example is more common than April.05.2020

WithoutPants:
We currently have `{d}` as delimiter. This can easily be changed to another character.

We really want either `{dd}` to accept one or two date characters, or have the new `{d}` field accept one or two date characters. A single-only date field would be of pretty limited use. I am leaning towards using a new `{d}` single/double date field currently. That said, given that the design on this isn't straightforward, I think we can push this to a separate PR rather than creeping the scope of this one.

PR #501, opened 2020-04-25
Make image extension check in zip files case insensitive

bnkai:
Also add webp as a supported extension for the galleries


pkg/models/model_gallery.go
```diff
@@ -123,10 +125,10 @@ func (g *Gallery) listZipContents() ([]*zip.File, *zip.ReadCloser, error) {
 	return filteredFiles, readCloser, nil
 }
 
-// return index of first occurenece of string x in name of zip contents, -1 otherwise
+// return index of first occurenece of string x (low case) in name of zip contents, -1 otherwise
 func contains(a []*zip.File, x string) int {
 	for i, n := range a {
-		if strings.Contains(n.Name, x) {
+		if strings.Contains(strings.ToLower(n.Name), x) {
```
WithoutPants, line 131:
```suggestion
		if strings.Contains(strings.ToLower(n.Name), strings.ToLower(x)) {
```

pkg/models/model_gallery.go
```diff
@@ -123,10 +125,10 @@ func (g *Gallery) listZipContents() ([]*zip.File, *zip.ReadCloser, error) {
 	return filteredFiles, readCloser, nil
 }
 
-// return index of first occurenece of string x in name of zip contents, -1 otherwise
+// return index of first occurenece of string x (low case) in name of zip contents, -1 otherwise
```
WithoutPants, line 128:
```suggestion
// return index of first occurrence of string x (case-insensitive) in name of zip contents, -1 otherwise
```

PR #508, opened 2020-04-29
Add flags for performer countries

InfiniteTF:
Adds SVG flags to the performer cards and the performer page.

Uses some fuzzy matching to parse common "misspellings" like USA, United Kingdom, Russia, etc. For v0.3 I'd like to migrate the country field to a fixed ISO field so we don't need to do this kind of parsing, but I'll leave it alone for now and see how it works out.

image

image


bnkai:
From a quick test it looks good.
Another entry for the fuzzy matching
Czechia : CZ , Czechia should be detected though
Also on performers were multiple countries are present ( comma or space seperated) none is displayed as a flag.

InfiniteTF:
> Another entry for the fuzzy matching
> Czechia : CZ , Czechia should be detected though
Seems like the library is a bit out of date on that point. There's a PR to fix it but the maintainer isn't super active. Anyway, added it.

> Also on performers were multiple countries are present ( comma or space seperated) none is displayed as a flag.
What would be the expected behavior? Should the first match or the last?

bnkai:
Ideally it should match all but since we didn't enforce a specific delimiter to split the countries eg comma either option mentioned is fine by me.

PR #509, opened 2020-04-29
freeones scraper tweaks

bnkai:
Remove all regexp.MustCompile  and logger.Fatal calls to make the scraper more predictable/consistent.
If a field is not valid skip it ( leave as nil) and continue to the next instead of exiting.
Add a default <del>30</del> **45** sec timeout to http requests ( no images are fetched so it should be safe, might need some more tweaking)

### DIFF
--- a/pkg/scraper/freeones.go
+++ b/pkg/scraper/freeones.go
@@ -12,7 +12,7 @@ import (
 	"github.com/stashapp/stash/pkg/models"
 )
 
-const freeonesTimeout = 30 * time.Second
+const freeonesTimeout = 45 * time.Second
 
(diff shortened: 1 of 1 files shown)


WithoutPants:
`MustCompile` is used where the regex is static and known to be valid. There is no need to change any of the `MustCompile` calls, since they are all valid expressions.

bnkai:
I just didn't like the fact that MustCompile panics if something is wrong. In the specific cases that can only happen during development ( changing the regex ) or during a pkg version upgrade if the package is bugged.
Should i change them back ?

WithoutPants:
IMO yes. I don't think changing these are necessary.

PR #510, opened 2020-04-30
Selectable wall preview type

InfiniteTF:
Adds an option to select either video, animation or image previews for walls. Also adds a setting for preview generation preset. 

I've refactored the wall components to largely only use css properties, this has the advantage of being more exact and deterministic. Also fixed a few issues like the gaps between wall items on larger screens, and the flashing when the page loads.

IMO the experience is a lot better since video playback is completely smooth, plus it saves a bunch of time and space from not having to generate webps.

### DIFF
--- a/ui/v2.5/src/components/Wall/WallPanel.tsx
+++ b/ui/v2.5/src/components/Wall/WallPanel.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React from "react";
 import * as GQL from "src/core/generated-graphql";
 import { WallItem } from "./WallItem";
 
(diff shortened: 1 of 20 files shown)


bnkai:
Using firefox (linux) with ryzen 5 1600 (6c/12t) 
The cpu usage goes from 15% to 30% ( 20 per page)  when switching from webp to mp4
When i change to  60 per page i have 43 % usage for webp and all seems ok but for mp4 firefox doesn't play correctly all the videos.
Chrome behaves ok even with 60 per page and with lower cpu usage ( 30% webp 45% mp4)

whitty-username:
+1 on this pull request. This will make the experience so much better for Safari & iOS users like myself. The wall view was one of the few things keeping me using Firefox. Would love to see this merged.

WithoutPants:
This is behaving a bit weird for me. Query for a single scene and you get a very small preview:

image

Layout is a bit odd in general if you don't get enough scenes to fill the horizontal space.

When If you mouse over a preview when `Enable sound` is checked, then the audio continues to be played after mousing away from the preview. You end up with a cacophony of sound (I tested this with family movies; testing it with porn would've been interesting). 

The quality dropdown needs a label on it in the `Generated Content` section:

image

I realise it has a description underneath, but it's a bit jarring having the checkbox and the `Slow`.

InfiniteTF:
@WithoutPants PR has been updated, although github is struggling right now and doesn't seem to have caught up. Have fixed the issue with <5 scenes, fixed sound playback, and zoom directions. Also cleared up the settings a bit.

WithoutPants:
There's a regression in the display of portrait previews in mobile display. The existing behaviour has a fixed height for all previews. The PR version doesn't limit the height, so portrait previews appear much larger than the other previews.

InfiniteTF:
@WithoutPants Ah right, fixed.

WithoutPants:
One more regression: the scene titles don't show up any more, regardless of if the option is enabled.

InfiniteTF:
@WithoutPants Oof, tried to render the function instead of calling it. Fixed.

ui/v2.5/src/components/Wall/WallPanel.tsx
```diff
@@ -10,106 +10,49 @@ interface IWallPanelProps {
...
+  if (index < 5) return "transform-origin-top";
+  if (count > 9 && count % 5 === 0 && index + 1 === count)
+    return "transform-origin-bottom-right";
+  if (count > 9 && count % 5 === 0 && index + 5 === count)
+    return "transform-origin-bottom-left";
+  if (count - (count % 5 || 5) <= index + 1) return "transform-origin-bottom";
+  if (index % 5 === 4) return "transform-origin-right";
+  if (index % 5 === 0) return "transform-origin-left";
+  return "transform-origin-center";
+};
```
WithoutPants, line 25: This needs documentation.

InfiniteTF:
@WithoutPants Good catch, I've fixed the bugs and added a generation message.

PR #513, opened 2020-05-01
Querybuilder integration tests

WithoutPants:
Adds a set of integration tests for the query builders. These aren't exhaustive (coverage is only about 4% of the package), but are intended to exercise #478.

Includes a small change to the scene query builder to remove a switch case that was impossible to hit.


pkg/models/querybuilder_scene.go
```diff
@@ -293,8 +293,6 @@ func (qb *SceneQueryBuilder) Query(sceneFilter *SceneFilterType, findFilter *Fin
 				whereClauses = append(whereClauses, "(scenes.height >= 1080 AND scenes.height < 2160)")
 			case "FOUR_K":
 				whereClauses = append(whereClauses, "scenes.height >= 2160")
-			default:
-				whereClauses = append(whereClauses, "scenes.height < 240")
```
bnkai, line 297: As it was if a non existing resolution string  was entered it would return the scenes with height < 240 which was indeed not correct
With the change  you did it will return all resolution scenes which is correct.

Imo though either  a new ultralow resolution needs to be created 0 < height  <240 or 240p adjusted to be 0 < height < 480 ( like FOUR_K)
so that we cover all the height range from scenes with the filter. ( sub 240p are practically non existant but just for cmpleteness)
the ui part in the scenecard  component which uses textutils for resolution makes the same checks. if a scene has < 240 height  it won't have a resolution label.

WithoutPants: The code was actually unreachable. `IsValid` returns false if the value isn't one of the valid enum values (which should really be referenced here). I agree regarding the lowest range. I think we should just get rid of the minimum value for `LOW` here and in the UI.

PR #516, opened 2020-05-03
Upgrade javascript libraries

InfiniteTF:
Due to stricter hook linting rules I've had to refactor `StashService` and `VideoHoverHook`.

Also fixed Navbar messages so `yarn extract` works correctly.


WithoutPants:
I noticed that you can paste an image into the performer page even when not editing. This is an existing bug, so not sure if you want to solve it in this PR or not.

InfiniteTF:
@WithoutPants Good point, I've amended the usePasteImage hook to resolve that. Also fixed the performer image setting so it cancels correctly and busts cache when the image is updated. I'd forgotten to backport that last one from v2.

PR #522, opened 2020-05-07
[TBD] Add x264 (MKV) streaming profiles

bnkai:
Adds some  x264 based streaming profiles ( ultrafast, medium, slow) using MKV as a container (can't be done in mp4) .
VP9 is left as the default option because mkv is not supported in all browsers ( firefox)

x264 ultrafast makes it possible in my case to live stream 1080p hevc, wmv  files that were stuttering with vp9.

Also added VP8 ( WEBM ) as it is "lighter" than VP9


whitty-username:
Sorry, I must be missing something, this article suggests JW Player supports MP4 as a container? HLS might also be an option? 

(For context I only have Mac's and naturally Safari is my browser of choice, AFAIK Safari won't play ball with MKV)

https://support.jwplayer.com/articles/supported-video-and-audio-formats-reference

bnkai:
mp4 as a container is supported but mp4 doesn't support live streaming(that's what i meant not supported) . 
HLS might be possible but it needs more research. I'll have a look at that when i have some more time.

bnkai:
Drafting this till we sort out #609

bnkai:
Closing this as #609 makes it obsolete

Issue #467, opened 2020-04-14
[Bug Report] Unable to select path by mouse in directory selector


michaelxxl2:
In the config section, I can not specify a path outside my user directory.
This causes problems if the stash is on an external/2nd Harddrive

bnkai:
If you mean the library stashes just click the add directory , enter /  (asuming linux) in the input field and you can select from there.

michaelxxl2:
I know. Putting a path there by hand does not seem to work and selecting does also not work after having the selected a directory, not allowing selecting a deeper sub-directory.
I currently run stash as root (just to test it out). This may be the problem.
The suggested directory is /data/
Data hast to be in /root/data/
At present I can not select a directory /media/user/drivename/data/prn/

bnkai:
![selectdirectory](https://user-images.githubusercontent.com/48220860/79311865-bd426d80-7f06-11ea-8abf-46c29a0012d6.png)
which version are you using ? the above( click add and then press `/` in the input field and you can select )  is possible in the dev version.

michaelxxl2:
Latest version. Downloaded it a few days ago.
Yes, I can select ONE directory but not dive deeper. 
Also putting the path there by hand seems not work.

Actually I only managed to run stash as root. Is this supposed to be?

bnkai:
I see what you mean.
As a workaround you can type a path at a time and then add
![2020-04-19_160654](https://user-images.githubusercontent.com/48220860/79688612-058bc380-8258-11ea-84e5-be8fc25f1d61.png)
eg for above
type /usr
type /local after that
.....
click add
By hand it seems fine , selecting with the mouse doesn't work

WithoutPants:
Fixed via #527

PR #523, opened 2020-05-08
Fix folder mouse selection

WithoutPants:
Fixes #467

PR #527, opened 2020-05-08
Overhaul look and feel of folder select

InfiniteTF:
Expanded the graphql payload so we know the starting directory, and whether there's a parent directory we can navigate to. 

Also spruced up the look a bit.

image


WithoutPants:
Supersedes #523

WithoutPants:
Fixes #467

PR #528, opened 2020-05-09
Fix edge case preview generation, tweak preview generation preset

bnkai:
On some cases the mp4 previews generated have timing "issues" making the webps generated from them huge while also taking a lot of time to generate.
You can identify files like that if you look at the screenshots directory and sort by desc size. The mp4 with the same name as the webp that is affected will show abnormal fps eg 0.01 instead of 25.
For already existing problematic files the webp AND the mp4 preview must be deleted from the screenshots folder and regenerated.
I also changed the x264 ffmpeg preset ( for the preview generation) from very slow to slow as the huge difference in encoding time isn't justified by the marginal ( if existant ) difference in visual quality.

### DIFF
--- a/pkg/ffmpeg/encoder_scene_preview_chunk.go
+++ b/pkg/ffmpeg/encoder_scene_preview_chunk.go
@@ -16,8 +16,8 @@ type ScenePreviewChunkOptions struct {
 func (e *Encoder) ScenePreviewVideoChunk(probeResult VideoFile, options ScenePreviewChunkOptions) {
 	args := []string{
 		"-v", "error",
-		"-ss", strconv.Itoa(options.Time),
 		"-i", probeResult.Path,
+		"-ss", strconv.Itoa(options.Time),
 		"-t", "0.75",
 		"-max_muxing_queue_size", "1024", // https://trac.ffmpeg.org/ticket/6375
 		"-y",
@@ -25,7 +25,7 @@ func (e *Encoder) ScenePreviewVideoChunk(probeResult VideoFile, options ScenePre
 		"-pix_fmt", "yuv420p",
 		"-profile:v", "high",
 		"-level", "4.2",
-		"-preset", "veryslow",
+		"-preset", "slow",
 		"-crf", "21",
 		"-threads", "4",
 		"-vf", fmt.Sprintf("scale=%v:-2", options.Width),


JoeSmithStarkers:
I think this PR should be reverted and reworked, as changing the position of the seek parameter (-ss) relative to the input parameter (-i) has a significant performance impact.

When the seek parameter is before the input parameter for ffmpeg a fast seek occurs.  A fast seek is a jump to the estimated position of the requested time code in the input file (the algorithm is more complex than this, eg a corse fast seek + fine output seek). However when the seek parameter is after the input parameter, then an output seek occurs, which requires ffmpeg to decode all the frames before the requested timecode.

Since this is the preview generation code, it will decode half the video file 12 times. 12 * 1/x where x = 0 .. 12, rather than decoding 12 * (0.75 + error_in_seek) seconds.

see -ss option in reference below:
http://ffmpeg.org/ffmpeg.html#Main-options
or
https://trac.ffmpeg.org/wiki/Seeking

JoeSmithStarkers:
On further thought, I suspect the problem with the webps can be solved via using an ffmpeg setpts filter. I am guessing the presentation timestamps for some reason are being preserved when extracting the preview segments. 

I've found a webp in my stash that is dramatically bigger than the rest, so i will test if a setpts filter fixes the problem.

bnkai:
As mentioned already our priority is making "correct" preview files.
Making the -ss an input filter results in files (in some) that after concat are not valid e.g. fps 0.01
There is no way to detect beforehand ( at least i couldn't find one) which files exhibit that problem.
If a preview file is like that it might seem to play ok when in scene view but the webp generated from it will be huge ( 1GB+ instead of a 1MB)

bnkai:
I think i tried the setps , not sure though.
Make sure to delete the mp4 preview prior to testing

bnkai:
As a sidenote i think adding the setpts filter also introduces a delay but i can't remember from when i tried that.

JoeSmithStarkers:
In my 28,402 generated webp files, i only have two that are bigger than the rest (9.3M & 24M). The 24MB webp file seems invalid, both ffprobe and vlc won't probe or play it. When I review the source preview mp4 file it's 4:13 seconds in length. And it has large segments of frozen picture with ongoing normal audio. This also could be a negative time stamp issue, or an audio framing issue (maybe fixed with segment_time_metadata and select=concatdec_select filter). I will investigate further. If you find your problematic source file, i would also like to review what is causing it's issue.

I don't believe making preview generation 114 times slower is the right solution to such a rare corner case.

bnkai:
The file generation includes the encoding and concatenation time it's not pure seek. I would like also to avoid that provided that a proper fix is available or at least a way to identify the problematic files and act accordingly. If you find a sample to test with, it would be appreciated as i can't seem to find my "problematic" one. Any feedback here or in the discord channel would be very helpful.

bnkai:
The change will be reverted so that we can least get some sample "problematic files" and then decide how to proceed. In some cases it may be possible that getting a fairly new ffmpeg version (4.2.2) makes the original problem less apparent.

WithoutPants:
I have reverted the seek parameter position change, leaving the preset as `slow`.

I believe that the problem this change was fixing was actually addressed in [this commit](https://github.com/stashapp/stash/commit/3f511e48e83977c541d1792532e48ce85cbe6ef7#diff-7aa487d878e91518361068b040b8d5b9), and that it was wrongly thought to be continuing to be present due testing by removing the webp file only and not the mp4 preview file. I have since tested by reverting the commit mentioned and testing with some files that generated large webp files before and reproduced the issue. Reverting the latest change and regenerating resulted in the correct webp files being generated.

I believe the original bug is squashed. Going forward, for those that still encounter this issue, we should instruct to delete the generated mp4 and webp files before regenerating with the latest build. If we still get reports of corrupt webp files after this, then we can investigate further.

Thanks for the report @JoeSmithStarkers

PR #531, opened 2020-05-10
Changelog

### LINKED ISSUE #504: [Bug Report] Excessive reads from Stashes when applying tags
**Describe the bug**
When adding to tags to videos that lacks the container information, reads are made to the Stashes.
With a large number of selected scenes, activity spikes on the drive.

**To Reproduce**
1. Have an older database without the format table in scenes filled out.
2. Click on Scenes
3. Select all scenes on the page
4. Add a tag to those scenes.
5. Disk activity should now increase.

**Expected behavior**
Expected behavior would be that dealing with tags should only affect R/W where the database is stored I guess.

 - OS: Windows 10
 - Firefox
 - Version 75.0

**Additional context**
Rescan the library and the issue disappears.

InfiniteTF:
Adds a markdown based changelog to the start page. I've used [gitmojis](https://gitmoji.carloscuesta.me/) to denote change "type".  Changes for v0.1.0 are copied from release page, changes for 0.2.0 are pulled from the commit log. I've omitted fixes for features that are implemented in v0.2.0.

Suggestions for changes very welcome!

Also tweaked the localForage hook slightly so it's a bit easier to init new storage types, and changed the default anchor color since the dark blue is very hard to read.

Resolves #504.

image

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v020.tsx
+++ b/ui/v2.5/src/components/Changelog/versions/v020.tsx
@@ -2,62 +2,64 @@ import React from "react";
 import ReactMarkdown from "react-markdown";
 
 const markup = `
-💥 **Note: After upgrading performance will be degraded until a full [scan](/settings?tab=tasks) has been completed.**
+#### 💥 **Note: After upgrading performance will be degraded until a full [scan](/settings?tab=tasks) has been completed.**
+
 &nbsp;
+### ✨ New Features
+*  Movies are now supported.
+*  Responsive layout for mobile phones.
+*  Add support for image scraping.
+*  Allow user to regenerate scene cover based on timestamp.
+*  Autoassociate galleries to scenes when scanning.
+*  Configurable scraper user agent string.
+*  Backup database if a migration is needed.
+*  Add modes for performer/tag for bulk scene editing.
+*  Add gender support for performer.
+*  Add SVG studio image support, and studio image caching.
+*  Enable sorting for galleries.
+*  Add scene rating to scene filename parser.
+*  Replace basic auth with cookie authentication.
+*  Add detection of container/video_codec/audio_codec compatibility for live file streaming or transcoding.
+*  Move image with cover.jpg in name to first place in Galleries.
+*  Add "reshuffle button" when sortby is random.
+*  Implement clean for missing galleries.
+*  Add parser support for 3-letter month.
+*  Add is-missing tags filter.
 
-#### Major Changes:
-* ✨ [Movies](/movies) are now supported.
-* 💄 Responsive layout for mobile phones.
-* ⚡️ Performance improvements and improved video support.
-* 📝 Support for localized text, dates and numbers.
+### 🎨 Improvements
+*  Performance improvements and improved video support.
+*  Support for localized text, dates and numbers.
+*  Replace Blueprint with react-bootstrap.
+*  Add image count to gallery list.
+*  Update prettier to v2.0.1 and enable for SCSS.
+*  Add library size to main stats page.
+*  Add slim endpoints for entities to speed up filters.
+*  Export performance optimization.
+*  Add random male performer image.
+*  Added various missing filters to performer page.
+*  Add index/total count to end of pagination buttons.
+*  Refactor build.
+*  Add flags for performer countries.
+*  Querybuilder integration tests.
+*  Overhaul look and feel of folder select.
+*  Add changelog to start page.
 
-#### Full list of changes:
+### 🐛 Bug fixes
+*  Update performer image in UI when it's replaced.
+*  Fix performer height filter.
+*  Fix error when viewing scenes related to objects with illegal characters in name.
+*  Fix to allow scene to be removed when attached to a movie.
+*  Make ethnicity freetext and fix freeones ethnicity panic.
+*  Fix to filter on movies from performer filter to movie filter.
+*  Delete marker preview on marker change or delete.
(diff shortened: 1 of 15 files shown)


InfiniteTF:
Should resolve #504.

WithoutPants:
This is a great idea and is looking very good. Here's my feedback:
* the expand/collapse chevron icons should have `pointer: cursor` and expand/collapse like clicking on the version headings. The first thing I did was try to click those.
* clicking on the version heading leaves the underline style up - I assume it's an `:active` css thing
* the changes should be consistent between versions. At the moment, 0.2 has "Major Changes" and "Full list of changes", while 0.1 is divided into "New Features", "Improvements" and "Bug Fixes". In my opinion we should stick with the 0.1 style, but either way it should be consistent.
* 0.1.1 just has "Fix version checking" in bold, so it needs to be brought in line with whatever format we decide on.

Eventually, I think this should be moved into the about/version page, but then you lose the note about needing to rescan, so we'd need another solution for that issue.

ui/v2.5/src/components/Changelog/versions/v020.tsx
```diff
@@ -0,0 +1,65 @@
...
+*  Implement clean for missing galleries.
+*  Add parser support for 3-letter month.
+*  Add is-missing tags filter.
+
+### 🎨 Improvements
+*  Performance improvements and improved video support.
+*  Support for localized text, dates and numbers.
+*  Replace Blueprint with react-bootstrap.
+*  Add image count to gallery list.
+*  Update prettier to v2.0.1 and enable for SCSS.
```
WithoutPants, line 34: This isn't a user-visible change, so probably not necessary.

ui/v2.5/src/components/Changelog/versions/v020.tsx
```diff
@@ -0,0 +1,65 @@
...
+*  Add library size to main stats page.
+*  Add slim endpoints for entities to speed up filters.
+*  Export performance optimization.
+*  Add random male performer image.
+*  Added various missing filters to performer page.
+*  Add index/total count to end of pagination buttons.
+*  Refactor build.
```
WithoutPants, line 41: As above, not necessary.

ui/v2.5/src/components/Changelog/versions/v020.tsx
```diff
@@ -0,0 +1,65 @@
...
+*  Add flags for performer countries.
+*  Querybuilder integration tests.
+*  Overhaul look and feel of folder select.
+*  Add changelog to start page.
+
+### 🐛 Bug fixes
+*  Update performer image in UI when it's replaced.
+*  Fix performer height filter.
+*  Fix error when viewing scenes related to objects with illegal characters in name.
+*  Fix to allow scene to be removed when attached to a movie.
```
WithoutPants, line 51: Can probably remove this. It is related to the new Movie feature and is not a bug fix between 0.1.1 and 0.2. When updating the changelog in future PRs on the pre-release builds, it makes sense to include such bug fixes, they should be removed for the release version.

ui/v2.5/src/components/Changelog/versions/v020.tsx
```diff
@@ -0,0 +1,65 @@
...
+*  Make ethnicity freetext and fix freeones ethnicity panic.
+*  Fix to filter on movies from performer filter to movie filter.
```
WithoutPants, line 53: As above, related to Movie feature.

ui/v2.5/src/components/Changelog/versions/v020.tsx
```diff
@@ -0,0 +1,65 @@
...
+*  Delete marker preview on marker change or delete.
+*  Prefer modified performer image over scraped one.
+*  Don't redirect login to migrate page.
```
WithoutPants, line 56: Fix for bug introduced in 0.2. Can probably remove.

ui/v2.5/src/components/Changelog/versions/v020.tsx
```diff
@@ -0,0 +1,65 @@
...
+*  Performer and Movie UI fixes and improvements.
```
WithoutPants, line 57: This is a little non-descript.

ui/v2.5/src/components/Changelog/versions/v020.tsx
```diff
@@ -0,0 +1,65 @@
...
+*  Include gender in performer scraper results.
+*  Include scene o-counter in import/export.
+*  Make image extension check in zip files case insensitive.
+*  Freeones scraper tweaks.
```
WithoutPants, line 61: Suggest this should be `Update built-in Freeones scraper for new API`.

bnkai:
> For future changes (ie not this PR), suggest the following:
> 
>     * new entries to the changelog should be added to the top, so that the user can see the most recent changes first
> 
>     * iterative changes and fixes for regressions introduced within a development build can be included in the changelog for the development build, but should be removed when finalising the changelog for the release.
> 
> 
> Now that I think about it, perhaps for the next development build (ie once 0.2 is released), we should put changes in chronological order mixed in (ie like your original version) rather than separating into features/improvements/bug fixes. Then during release, we can sort them into their applicable categories.
> 
> Paging @bnkai for input too.

I think this indeed makes more sense. For dev just put everything in and then squash/edit as needed for the release version. For anyone needing more details there is always the github commits page.

InfiniteTF:
Updated to resolve your comments @WithoutPants.

PR #534, opened 2020-05-12
Make auto tagger matching more flexible

### LINKED ISSUE #483: [Bug Report] Autotagger doesnt find matches between square brackets
**Describe the bug**
The autotagger does not find matches -- such as performer name -- when the matching string in the filename is between square brackets, `[ ]`.

**To Reproduce**
1. Create a performer name, `Jane Doe`.
2. Create an mp4 for the Stash video library, `Scene.2020-04-19.[Jane Doe].mp4`
3. Scan library for new file.
4. Click autotag.

**Expected behavior**
The new scene should be tagged with the performer name `Jane Doe`. 

 - OS: Ubuntu Server 18.04
 - Browser Chrome
 - Version 81

**Additional context**
Didn't test Autotagger for _studio_ or _tag_ string matches in square brackets.

WithoutPants:
Fixes #483 

Changes the auto tagger regex to be a bit more flexible when matching. The boundary pattern is changed to be any non-alphanumeric character. The intra-name separators are unchanged, so it won't match something like `firstname[lastname`, but it should match something like `{firstname.lastname}`.

### DIFF
--- a/pkg/manager/task_autotag.go
+++ b/pkg/manager/task_autotag.go
@@ -24,12 +24,10 @@ func (t *AutoTagPerformerTask) Start(wg *sync.WaitGroup) {
 func getQueryRegex(name string) string {
 	const separatorChars = `.\-_ `
 	// handle path separators
-	const endSeparatorChars = separatorChars + `\\/`
 	const separator = `[` + separatorChars + `]`
-	const endSeparator = `[` + endSeparatorChars + `]`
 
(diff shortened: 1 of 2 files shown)


bnkai:
Tried with tags, studios and it seems to work better now (didn't have any performer cases in my db ) . Words enclosed in `[ ] {}` for example seem to wok fine now.

PR #535, opened 2020-05-12
Fix 0 undefined library size on empty db

bnkai:


WithoutPants:
I'm not super thrilled with the way this is handled in general. I don't think sqlx should error out in this case. I can't see a way around it, so this fix will have to do.

WithoutPants:
This looks related: https://github.com/jmoiron/sqlx/issues/613

Issue #540, opened 2020-05-14
[Feature] Improve initial setup experience


WithoutPants:
Currently, starting a system from scratch isn't super flexible. The config and database locations are assumed, the location folders that you can set must be created manually by the user prior, and must be entered manually.

Propose the following improvements:
* if no config file has been created yet, give an option (via checkbox) to create the config in the current working directory
* allow setting the database location during initial setup, with the initial value set to the default database location
* create the generated/cache/metadata folders based on the entered location if they are not already present

I think we should also consider rolling the setup and migration pages into the react webapp, so that we can use the folder selection control to enter the folder locations.

bnkai:
As long as it doesn't affect the env ( used in the docker images )  behaviour i think that all of the above are a major improvement to the user experience during setup.

PR #546, opened 2020-05-14
Make wizard form inputs required

InfiniteTF:
Ideally we'd have a query to validate the paths, but it's probably not worth the effort.


bnkai:
Don't the paths get checked by the backend anyway?
Besides if #540 moves forward there is no need to waste time/effort for this one.

InfiniteTF:
The paths get checked, but it's better to warn about the fields being required early.

All the suggestions in #540 are good though, this is just a bandaid.

PR #555, opened 2020-05-18
Add debug logging for xpath scraping

WithoutPants:
Adds debug logging for xpath scraping. This is to assist the creation of community scrapers.

Outputs the value of each xpath field in the form of `[index][field] = value`. Also outputs regex replacements and sub-scraping operations.


bnkai:
Looks good.
A potential improvement if it's easy.
I get a list of [Name] attributes 0 to 19 that are performer and then tag names ( if the scraper i used had studios i'm guessing that also). An extra print of "performer"  or "tag" in between to distinguish would be nice.

WithoutPants:
Easiest way was to add logging before processing each part of the scene members.

PR #556, opened 2020-05-18
Change scrape matching (studio, movies, tag, performers) to case insensitive

bnkai:
Changes the FindByName/s queries to support case insensitive matches.
All calls from the scrapers code where changed to be case insensitive.
All other left to case sensitive.
I don't know if the filename parser should also be changed so i left that as it was.

A few sideeffects can occur from the matching , eg if a user has multiple tags to cover all cases upper, lower, mixed then the first name matched (LIMIT 1) will get the tag. Merging all tags into one is a complex procedure and i think should be left out to the user, if he chose to do this.
For studios, performers, movies i don't think there should be multiple "similar" named cases anyway.


WithoutPants:
Retest filename parser change after #563 is merged.

bnkai:
I will merge with master and retest once #563 is merged (it should work though).

bnkai:
Seems to work fine also for the filename parser after the merge with develop

WithoutPants:
Can you please add unit tests for the changed functions? Or let me know and I can add them.

WithoutPants:
Tests ok.

bnkai:
I think I saw the TODO in the test and got lazy :-).
I'll do that and trouble you if something goes wrong.

PR #574, opened 2020-05-23
Localize dates and numbers

### LINKED ISSUE #565: [Feature] Format all numbers and dates
Numbers, dates, weight and length measurements should be wrapped by react-intl so they can be formatted to the users locale. Unit conversion is currently not supported natively, so will need to be implemented. If en-us convert to imperial, otherwise use metric. All inputs should remain metric. 

165cm -> 5" 4'
45kg -> 99lbs
2018-05-02 -> May 5th, 2018, etc.
81787583 -> 81,787,583, 81.787.583, etc.

InfiniteTF:
Localizes dates and numbers when not editing. Also changed height rendering to imperial if locale is set to US.

Resolves #565

### DIFF
--- a/ui/v2.5/.eslintrc.json
+++ b/ui/v2.5/.eslintrc.json
@@ -47,6 +47,9 @@
       "react/prop-types": "off",
       "react/destructuring-assignment": "off",
       "react/jsx-props-no-spreading": "off",
+      "react/style-prop-object": ["error", {
+        "allow": ["FormattedNumber"]
+      }],
       "spaced-comment": ["error", "always", {
           "markers": ["/"]
       }],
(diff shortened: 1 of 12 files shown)


WithoutPants:
The height change is a little contentious:
* imo input and display should be consistent. Having the US users input in metric and display in imperial seems like a usability issue
* for those users that have not yet changed there locale away from US to UK (like me), the height is now inexplicably formatted as imperial, with no obvious reason why

IMO, we should defer the height change to a later PR with the following refinements:
* input should be in the same format as display
* a changelog entry indicating that for users that have not switched locales, the height field will be changed from previous, and how to resolve the issue
* the language field in the settings should document the impact on the display - including the height change

InfiniteTF:
I removed the feet/inches conversion for now. I considered adding a metric/imperial option, but it seemed overkill when we don't have weight anyway. I have no idea if Americans prefer feet/inches, I assume so, but we can defer it until a later point when it can be implemented in a more thorough fashion.

Issue #525, opened 2020-05-08
[Feature] Cosmetic Enhancements to Scene Page


tarzanboy76:
**Is your feature request related to a problem? Please describe.**

- Most studios put their logo in the bottom right hand corner.  When looking at the scene page, this is where the current layout places the scene duration and resolution, which can sometimes clash with the screenshot that is typically used as the cover image for the scene.

- The check-box used for selecting multiple scenes can sometimes blend into the background if the cover image for a scene is a similar pale/grey colour to the checkbox.

**Describe the solution you'd like**

Re-organise the layout of the scenes page.  Some possible options;

1. Place the studio logo in the bottom right hand section -- that way, if the scene already has a watermarked logo (most likely already in the same place), it will simply sit over the top of it.

2. Move the duration and resolution information to the top right or left hand side.

3. Move the checkbox to outside the scene (maybe next to the title??)

**Describe alternatives you've considered**

None specifically... but making the duration and resolution white on semi-transparent black might also help (see https://www.porntrex.com/) for examples.  Similarly, if the checkbox is going to remain in the upper left hand corner, it might be worthwhile looking at other designs/colours.

WithoutPants:
> Place the studio logo in the bottom right hand section -- that way, if the scene already has a watermarked logo (most likely already in the same place), it will simply sit over the top of it.
> Move the duration and resolution information to the top right or left hand side.

This can _almost_ be done already with css overrides, but the studio overlay element needs to be moved into the scene image element. 

For reference, the css override I used (once I moved the studio overlay element) is:
```
.scene-studio-overlay {
  bottom: 1rem;
  right: 0.7rem;
  height: inherit;
  top: inherit;
}

.scene-specs-overlay {
  right: 0.7rem;
  top: 0.4rem;
  bottom: inherit;
}
```

Phasetime:
Seems fixable with new UI API #4256, can be closed.

PR #580, opened 2020-05-26
Fix date rendering for missing date values

WithoutPants:
Fixes issue where today's date would be rendered for performer birthdate and movie date when these values weren't set, and epoch date would be rendered for scene date.

Also sneaks in a change to move the scene studio overlay element to within the scene preview element, to allow the studio overlay position to be customised per #525.


bnkai:
Works ok for me.
For the sneak part should the css part ( default ) be changed to the one you mention in #525 ?
Now that i look at my scenes the vast majority have the logo at the bottom right corner as mentioned in 525 so at least for me it looks better that way.

WithoutPants:
This change is sufficient to allow a css change to do that, and I'd like to spend more time working out the ideal positioning of things. I'm not sold on having the duration and resolution in the top right.

Issue #572, opened 2020-05-22
Indexer is unable to see files with the character "【" in them


victorhooi:
**Describe the bug**
The Stash scan seems to ignore folders with the character "【" in them.

Is this related to Unicode maybe, or how we scan?

**To Reproduce**
1. Add a video/scene with "【" in the filename.
2. Run a scan operation from the Stash Web UI,
3. Attempt to search for the video by name.

**Expected behavior**
The Stash scan operation should pickup all subdirectories in the specified download directory.

**Screenshots**
If applicable, add screenshots to help explain your problem please ensure that your screenshots are SFW or at least appropriately censored.

**Stash Version: (from Settings -> About):**
```
Version:	v0.1.1-148-gd74a303
Build hash:	d74a303
Build time:	2020-05-14 00:55:16
```

**Desktop (please complete the following information):**
 - OS: Debian Buster
 - Browser: Chrome
 - Version: 85.0.4151.0

victorhooi:
It appears to be this character

https://unicodelookup.com/#%E3%80%90/1

which is a fairly common punctuation character in Asian languages:

https://en.wikipedia.org/wiki/Japanese_punctuation
https://en.wikipedia.org/wiki/Chinese_punctuation

Also - to clarify, that character appears in both the parent directory of the video, and also in the video filename itself.

victorhooi:
Upon further testing, it only happens if the bracket character is in the video filename as well.

An example of a full non-working filename is this:

```
【ses23.com】BOMN-122.1080p/【ses23.com】BOMN-122.1080p.mkv
```

To workaround the issue - you can move the file (e.g. introduce a space in front of the bracket) - however, this will break things that depend on the original filename (e.g. seeding):
```
cp -R '【ses23.com】BOMN-122.1080p/' ' 【ses23.com】BOMN-122.1080p/'
cd ' 【ses23.com】BOMN-122.1080p'/
mv '【ses23.com】BOMN-122.1080p.mkv' ' 【ses23.com】BOMN-122.1080p.mkv'
```

victorhooi:
As a temporary workaround - you can make a symlink to the video file, to either introduce a space before "【", or rename the file otherwise:

```
pwd
/home/name/qbittorrent/downloads/【ses23.com】example.1080p
ln -s 【ses23.com】example.1080p.mkv " 【ses23.com】example.1080p.mkv"
```

bnkai mentioned that the third-party globbing library we use is this one:

[...]

PR #586, opened 2020-05-31
Upgrade bmatcuk/doublestar to 1.3.1

bnkai:
Fixes issue #572 
Did an update on the library (1.3.1 is the latest version fixing the issue), a `go mod tidy` and a `go mod vendor`
From my tests it seems to work ok and in my case adds an extra 32 mp4 files in my DB that where not detected/scanned because the filename started with a unicode character.


go.mod
```diff
@@ -2,17 +2,15 @@ module github.com/stashapp/stash
...
-	github.com/bmatcuk/doublestar v1.1.5
+	github.com/bmatcuk/doublestar v1.3.1
 	github.com/disintegration/imaging v1.6.0
 	github.com/go-chi/chi v4.0.2+incompatible
 	github.com/gobuffalo/packr/v2 v2.0.2
 	github.com/golang-migrate/migrate/v4 v4.3.1
 	github.com/gorilla/sessions v1.2.0
 	github.com/gorilla/websocket v1.4.0
 	github.com/h2non/filetype v1.0.8
-	github.com/inconshreveable/mousetrap v1.0.0 // indirect
```
WithoutPants, line 15: Can we leave this in for now. I think we'll want to change the packr dependency to get rid of this issue altogether, but this is required for windows builds currently, and also exists in the vendor directory.

bnkai: Yeah i keep forgetting that, it gets removed by the `go mod tidy`.
The cross-compile script seems to have made a windows binary though. Is it only needed when compiling from windows ?

WithoutPants: It seems a be a wrinkle in go. The tidy command doesn't take into account cross-platform dependencies, but the vendor command must do.

PR #589, opened 2020-06-01
Add polyfill for number unit formatting

InfiniteTF:
Units are not yet supported in Firefox and Safari, but it will be part of the spec so there's no reason not  to use it.


bnkai:
The size is displayed ok now in firefox

PR #592, opened 2020-06-03
Reload scrapers button

WithoutPants:
Adds an entry in the scrapers button to refresh loaded scrapers. This is for when scrapers are modified, added or removed. Changes to the scrapers will not come into effect until the scrapers are reloaded or the system is restarted.

### DIFF
--- a/graphql/documents/mutations/scrapers.graphql
+++ b/graphql/documents/mutations/scrapers.graphql
@@ -1,0 +1,3 @@
+mutation ReloadScrapers {
+  reloadScrapers
+}
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -53,6 +53,7 @@ type Query {
   """List available scrapers"""
   listPerformerScrapers: [Scraper!]!
   listSceneScrapers: [Scraper!]!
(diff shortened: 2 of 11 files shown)


bnkai:
I'm having trouble rendering the scrape menu in the scene edit menu. 
It doesn't show anything except the save and delete button (the scrape button appears works/ok).
I think `QueryableScrapers` is empty for some reason.

WithoutPants:
The scene scraper menu is populated by scrapers that have the `sceneByFragment` section. See #236 and #269 for examples.

Changed so that the scene scraper menu now always displays for the reload option even if there are no queryable scrapers.

bnkai:
The xpath samples I think didn't have the sceneByFragment so it didn't make it in most ( if not all) of the xpath scene community scrapers. What should be the "action :" part for an xpath scraper? The scraper name?

WithoutPants:
`sceneByFragment` doesn't currently support xpath scrapers.

WithoutPants:
The action field will need to be `script` or `stash`.

bnkai:
Works ok for me. Both scene and performer section.

Issue #585, opened 2020-05-31
[Feature] the current Raspberry build is only 32-bit, please provide a build for 64-bit images


GETandSELECT:
The [Ubuntu image is 64-bit](https://ubuntu.com/download/raspberry-pi). Thank you very much.

```
$ file stash
stash: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.3, for GNU/Linux 3.2.0, Go BuildID=i1FSHpCNTZ_KpdKf6WzO/wPsfSQrcKjopjAi_8qDM/G2i86PG6zMzbxCddoyW4/vt3BX7ksv3NmGcp-ypN5, BuildID[sha1]=4379e8240db765dfb52d70cba2af4a0bf90b526c, not stripped
$ uname -a
Linux ubuntu 5.4.0-1011-raspi #11-Ubuntu SMP Fri May 8 07:43:33 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux
```

bnkai:
We use cross  compiling to make the binaries so the ubuntu image is not needed anyway.
`GOOS=linux GOARCH=arm GOARM=5 CC=arm-linux-gnueabi-gcc` something like this is used for the raspi build. 
I guess ` GOOS=linux GOARCH=arm64 CC=aarch64-linux-gnu-gcc` could be used for a 64bit binary
Doesn't the 32bit run in your case ?

GETandSELECT:
> The official Raspbian OS has not yet been updated for a 64-bit device. 32-bit software can run on a 64-bit system (but not vice versa). I use Ubuntu which is 64-bit.

Strange, but the 32-bit bin doesn't run on my system.

`arm64` is correct.

bnkai:
@GETandSELECT i made a test build but have no way to try that
can you try https://easyupload.io/k1q1j3 ?
md5
`ea3c7d6dba468b6eb42cdc308c0dcf81  stash-arm64`

GETandSELECT:
it's working fine. Thank you very much

bobbynails:
> @GETandSELECT i made a test build but have no way to try that
> can you try https://easyupload.io/k1q1j3 ?
> md5
> `ea3c7d6dba468b6eb42cdc308c0dcf81 stash-arm64`

could you reupload this?

bnkai:
Do you need latest dev or master ?
I'll compile a new build over the weekend.

[...]

Issue #352, opened 2020-02-05
[Bug Report] Docker Compiler container won't be up to date before travis builds things. [Edge Case]


Leopere:
**Describe the bug**
This container https://github.com/stashapp/stash/tree/develop/docker/compiler if it ever has to be updated in order for a build to complete we're going to have to manually re-issue a Travis build signal for this commit.  Ideally, the compiler container will rarely need updating however it's not impossible.  This could be remedied by replacing https://github.com/stashapp/stash/blob/f87117b0d6997e1c887f00a00654133c0dbbe8cf/.travis.yml#L23 and its contents of `- docker pull stashapp/compiler:develop` to be a `- docker build` it will increase Travis build time but it will make sure that we never hit this edge case.

**To Reproduce**
1. Push a build that needs an updated Compiler before DockerHub has updated the Compiler container.

**Expected behavior**
Every build is run on the latest code including compiler containers.

WithoutPants:
I'm not convinced that covering this corner case is worth the extra build time from building the docker container every single build.

Leopere:
For the most part, it's just something that should be noted.

PR #602, opened 2020-06-07
Cross compile for armv7/armv8, raise raspi variant to armv6

hardcore-coder:
Hi,

I extended the stash cross compilation to include newer ARM platforms (current raspis and NAS'). I used the [naming convention of the official docker images](https://github.com/docker-library/official-images#architectures-other-than-amd64)) but kept the existing binary file names for backwards compatibility.

In addition, I added Dockerfiles for ARM. They are practically all the same but I decided not to put too much effort in it before I got some feedback. What we probably want is taking advantage of the [docker multi arch support](https://docs.docker.com/docker-for-mac/multi-arch/).

Disclaimer: I tested only the arm64v8 version and it's working as expected. The arm32v7 variant compiles but my newer raspis are not available for testing atm.

I don't know why the raspi version was compiled with GOARM=5, as far as I know (and could google) the first raspi supports v6, so I increased this. Did you have a specific reason to choose v5 over v6? If yes, I will revert this change.

I have not updated the `upload-pull-request.sh` script because the raspi version was already missing there.

I guess this pull request will hit this issue: #352 

Fixes: #585


hardcore-coder:
one more thing.
I stumbled upon the user being `stashappdev` in the [compiler makefile](https://github.com/stashapp/stash/blob/develop/docker/compiler/Makefile#L1) and `stashapp` in the cross compile script. It took me a moment to realize why the cross compilation did not pick up my changes to the compiler container during local development.

.travis.yml
```diff
@@ -44,6 +44,8 @@ deploy:
     - dist/stash-osx
     - dist/stash-win.exe
     - dist/stash-linux
+    - stash-linux-arm64v8
```
bnkai, line 47: - dist/stash-linux-arm64v8

.travis.yml
```diff
@@ -44,6 +44,8 @@ deploy:
...
+    - stash-linux-arm32v7
```
bnkai, line 48: `- dist/stash-linux-arm32v7`

.travis.yml
```diff
@@ -61,6 +63,8 @@ deploy:
     - dist/stash-osx
     - dist/stash-win.exe
     - dist/stash-linux
+    - stash-linux-arm64v8
```
bnkai, line 66: `- dist/stash-linux-arm64v8`

.travis.yml
```diff
@@ -61,6 +63,8 @@ deploy:
...
+    - stash-linux-arm32v7
```
bnkai, line 67: `- dist/stash-linux-arm32v7`

bnkai:
I don't remember why the version was 5 but i'm guessing for broader compatibility (there are also other arm platforms excpet raspi)? If we take the raspi as a baseline i think armv6 is acceptable.
Also any docker related PR as you mentioned won't complete as the docker image needs to be pushed beforehand.
The `upload-pull-request.sh` is mainly used to help try/debug PRs so leaving as it is isok.

docker/production/x86_64/Dockerfile
```diff
@@ -1,21 +1,25 @@
...
+RUN curl -L -o /stash $(curl -s https://api.github.com/repos/stashapp/stash/releases/latest | awk '/browser_download_url/ && /$STASH_BINARY/' | sed -e 's/.*: "\(.*\)"/\1/') && \
     chmod +x /stash && \
-    curl --http1.1 -o /ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-amd64-static.tar.xz && \
+    curl --http1.1 -o /ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-${FFMPEG_ARCH}-static.tar.xz && \
     tar xf /ffmpeg.tar.xz && \
     rm ffmpeg.tar.xz && \
     mv /ffmpeg*/ /ffmpeg/
 
-FROM ubuntu:18.04 as app
-RUN apt-get update && apt-get -y install ca-certificates
```
bnkai, line 18: any reason to change that?
copying the certs as you do later should work but isn't it better to have the package installed?

.travis.yml
bnkai, line 67: - dist/stash-linux-arm32v7

docker/production/x86_64/Dockerfile
bnkai: Fair enough. Seems ok then.

bnkai:
> Would it make sense to move the changes in the compiler Dockerfile to a separate PR? When that is merged travis should be able to test this PR, no?

I think that would be better. It will be easier to troubleshoot also if we have the compiler docker image available.

WithoutPants:
I've opened #737 to just include the cross-compilation changes. Thanks for the contribution. I'll close this and you can open a new one for the dockerfile changes.

PR #609, opened 2020-06-11
Transcode stream refactor

### LINKED ISSUE #600: [Bug Report] Cannot transcode stream videos on Safari
**Describe the bug**
Video player gives me this error for all mkv files (tried with two different browser clients and devices). Strangely enough, tested a few of these files and were playable just yesterday.

I have both options in settings turned on. (Force Matroska as supported and Force HEVC as supported). Even tried alternating between them with restarts. 

Log file does not provide any useful information.

```
20-06-06 16:04:11  Error [stream] error serving transcoded video file: write tcp 172.27.0.3:9999->10.0.1.25:42794: write: broken pipe
2020-06-06 16:04:11  Info [stream] client closed the connection. Killing stream process.
2020-06-06 16:04:11  Info [stream] transcoding video file to video/webm
2020-06-06 16:04:03  Info [stream] client closed the connection. Killing stream process.
2020-06-06 16:04:03  Info [stream] transcoding video file to video/webm
2020-06-06 16:04:03  Info [stream] client closed the connection. Killing stream process.
2020-06-06 16:04:02  Info [stream] transcoding video file to video/webm
2020-06-06 16:04:02  Error [stream] error serving transcoded video file: read |0: file already closed
2020-06-06 16:04:02  Info [stream] client closed the connection. Killing stream process.
2020-06-06 16:04:01  Info [stream] transcoding video file to video/webm
2020-06-06 16:04:01  Info [stream] client closed the connection. Killing stream process.
2020-06-06 16:04:00  Info [stream] transcoding video file to video/webm
2020-06-06 16:04:00  Debug File is not streamable , transcode is needed  h264, opus, matroska
```

Related: #354 

**Expected behavior**
For files to be transcoded, streamed and played on client.

Version: | v0.2.0
Build hash: | bd5ee9b
Build time: | 2020-06-06 05:34:33

 - OS: Mac OS Mojave (10.14)
 - Browser: Safari
 - Version: 12

 - Device: iPhone 8
 - OS: iOS 13.4
 - Browser: Safari
 - Version: Whatever ships with this iOS version. Don't know where to get info.

**Additional context**
Filename:
```
file:///data/Other2/#58 ESP8266 Sensor runs 17 days on a coin cell_transmits data to sparkfun.com and ubidots.com-20160505-IYuYTfO6iOs.mkv
```

WithoutPants:
This is a significant refactor of the transcode streaming code.

The `forceMKV` and `forceHEVC` flags have been removed, and `is_streamable` has been removed from the scene graphql type. Instead, there is a new `IsSceneStreamable` graphql query that accepts a list of supported video codecs. Similarly, the `stream.mp4` route has been changed to accept a list of codecs as well. The supported video codecs are obtained using a custom Modernizr build that only tests for video support. There is an additional test for mkv support, but it is hard-coded to check if the client browser is Chromium-based.

On the backend, the transcode code is changed so that it chooses the codec to use based on the supported codecs sent to the request. It is currently hard-coded to prefer VP9, VP8, HEVC then H264. This should probably be further refined to accept the supported codecs in order of preference. 

~~This doesn't fix the Safari bug referenced in #600 but is a decent first step.~~ This obsoletes a lot of the changes in #522 which will need to be reworked if we go ahead with this.

Fixes #600


WithoutPants:
Incidentally, this should eliminate the performance impact of not having the scene containers in the database.

pkg/ffmpeg/stream.go
```diff
@@ -0,0 +1,199 @@
...
+		"-vf", "scale="+scale,
+	)
+
+	if len(s.codec.extraArgs) > 0 {
+		args = append(args, s.codec.extraArgs...)
+	}
+
+	args = append(args,
+		// this is needed for 5-channel ac3 files
+		"-af", "channelmap=channel_layout=5.1",
```
bnkai, line 110: As mentioned in discord 
`"-af", "channelmap=channel_layout=5.1",`
makes ffmpeg error out in my case in linux
the usage of the option is wrong and even then i dont think it's recommended for non 5.1 files
i would opt for `"-ac", "2",` instead to downmix to stereo

bnkai:
Mp4 as a container is not suited for live streaming.
Try forcing h264, mp4 as the codec , container and try to play a scene that needs transcoding
Try a long scene (not a small one that can be converted at once). When you seek through it and then let it play it plays for a little while and then stops. Even playing without seeking has the same issue.
h264 and hevc is possible but only when mkv is the container (#522 makes use of that)

bnkai:
> Incidentally, this should eliminate the performance impact of not having the scene containers in the database.

Yeah i would imagine everyone at some point would have finished a new scan to update the  containers but the `is_streamable`needed to go anyway.

WithoutPants:
MP4 appears to work correctly on Chrome, but not on Firefox. I'm not sure if it's worth spending the time and effort fixing it when Firefox supports webm anyway.

The latest commit adds support for HLS for Safari. Initial testing looks good, but I need to do a more extension test run. HLS is supported in Edge but playback didn't work for me, so I've made it so that it won't use HLS unless other options aren't available.

The HLS support was assisted with knowledge from a fork of [Golang HLS Streamer](https://github.com/shimberger/gohls). The fork was a lot more basic than the original, so there's possibly some improvements to be done with further study of that project. For example, [this file](https://github.com/shimberger/gohls/blob/master/internal/hls/segment.go) looks like it might have some refinements that might improve it.

pkg/ffmpeg/stream.go
```diff
@@ -0,0 +1,273 @@
...
+	)
+
+	if len(o.Codec.extraArgs) > 0 {
+		args = append(args, o.Codec.extraArgs...)
+	}
+
+	args = append(args,
+		// this is needed for 5-channel ac3 files
+		"-ac", "2",
+		"-b:v", "0",
```
bnkai, line 221: `"-b:v", "0",` this should be removed from there and added as an extraArg in the VP8 and VP9 profile.
It is needed there of a quirk in libvpx 
It shouldn't be used in other settings because it might interfere with the crf setting

pkg/ffmpeg/stream.go
```diff
@@ -0,0 +1,273 @@
...
+
+var CodecVP9 = Codec{
+	Codec:    "libvpx-vp9",
+	format:   "webm",
+	MimeType: MimeWebm,
+	extraArgs: []string{
+		"-deadline", "realtime",
+		"-cpu-used", "5",
+		"-row-mt", "1",
+		"-crf", "30",
```
bnkai, line 81: `"-b:v", "0",
should be added here instead of the getStreamArgs() function

pkg/ffmpeg/stream.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+var CodecVP8 = Codec{
+	Codec:    "libvpx",
+	format:   "webm",
+	MimeType: MimeWebm,
+	extraArgs: []string{
+		"-deadline", "realtime",
+		"-cpu-used", "5",
+		"-crf", "30",
```
bnkai, line 92: `"-b:v", "0",
should be added here instead of the getStreamArgs() function

pkg/ffmpeg/stream.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+var CodecHEVC = Codec{
+	Codec:    "libx265",
+	format:   "mp4",
+	MimeType: MimeMp4,
+	extraArgs: []string{
+		"-movflags", "frag_keyframe",
+		"-preset", "veryfast",
+		"-crf", "30",
```
bnkai, line 103: crf 30 is probably too much , closer to 25 might be better

pkg/ffmpeg/stream.go
```diff
@@ -0,0 +1,273 @@
...
+
+var CodecHLS = Codec{
+	Codec:    "libx264",
+	format:   "mpegts",
+	MimeType: MimeMpegts,
+	extraArgs: []string{
+		"-acodec", "aac",
+		"-pix_fmt", "yuv420p",
+		"-preset", "veryfast",
+		"-crf", "30",
```
bnkai, line 57: crf 30 for libx264 might be too much , 25 could be better

pkg/ffmpeg/stream.go
```diff
@@ -0,0 +1,273 @@
...
+
+var CodecH264 = Codec{
+	Codec:    "libx264",
+	format:   "mp4",
+	MimeType: MimeMp4,
+	extraArgs: []string{
+		"-movflags", "frag_keyframe",
+		"-pix_fmt", "yuv420p",
+		"-preset", "veryfast",
+		"-crf", "30",
```
bnkai, line 69: crf 30 for libx264 might be too much , 25 could be better

pkg/ffmpeg/stream.go
bnkai, line 92: also crf 30 might be a bit too much
vp8 tends to use crfs closer to x264 then vp9
22-25 might be a better choice

bnkai:
Some quick feedback from testing on an Android Smartphone.

Microsoft Edge ( got the hint from discord ) is the only browser that i found which supports hevc natively. This can br direct streamed (no transcode)  when Force Hevc was checked in the dev build. In this version HEVC is not detected as supported.

Chrome and Edge ( again in android ) show `Choosing transcode codec from: h264,vp8,vp9,mkv,hls` when transcoding which implies that hls is supported in Android ? I tried forcing hls as a first choice when transcoding but it doesn't seem to play , it returns the infamous file not supported message from JWplayer. I don't know if the detection of hls as supported or if the implementation is  lacking. HLS support is ok imo even if it only plays in safari.

InfiniteTF:
@bnkai Android does support HLS, see https://support.jwplayer.com/articles/adaptive-streaming-reference#hls-support

However according to jwplayer it only supports an older spec, so some tweaks might be required. I can see if I can get it working.

bnkai:
@InfiniteTF i found this https://alexzambelli.com/blog/2016/05/04/understanding-hls-versions-and-client-compatibility/ regarding hls specs. I think for the job we need it even the oldest version is enough.
Wouldn't it be a lot easier to just limit hls to safari only ? Imo webm for the rest or even a mp4/mkv  with h264  extra choice for chrome based browsers should be enough. The only problem with webm/vp9 is the higher cpu usage requirements ( problem for older, low specs machines )  that can not be addressed even with the realtime setting.

WithoutPants:
What is the consequence of leaving the HLS compatibility code as is if Android Chrome only uses it if forced? I'd rather not add Safari-detection code if we can avoid it, since it's apparently very difficult to do and is brittle. I'm thinking we should consider a move away from jwplayer if we can maintain the same features.

bnkai:
With the code as it is it doesn't really matter since you have to alter the code yourself to reach hls in other browsers. The safari detection was mainly for cosmetic reasons and if it was easy to implement.
The only issue left is the Microsoft Edge in android and hevc. It might be easier to leave an advanced option in the UI for manual codec overrides.

WithoutPants:
The modernizr build I had was quite old and didn't include the H265 test. However, even with the latest version, H265 support is not detected correctly, and I was unable to come up with a test that worked, so I have restored the `forceHEVC` option. I'm still a bit unhappy with the way that video streaming works currently, so I'll probably try to tackle this again before the 0.3 release.

bnkai:
As you mention even with your latest commit hevc is not detected for me either. Tested with microsoft edge in android and a custom chromium build. Checking the restored hevc option at least makes it possible to directly stream using those browsers instead of transcoding.
Mkv support gets properly detected for firefox vs chrome

victorhooi:
Is there any way of integrating a UI toggle to force live-transcoding to this re-factor?

https://github.com/stashapp/stash/issues/654

rg9400:
Not sure if this is the proper place to comment on it, and it seems like the PR is not really focused on this issue per se. Still, I thought I would just comment on my usecase. I really like Stash, but I am often remote from my server which has limited upload. Services like Emby/Plex/Jellyfin offer not just transcoding to match codec support, but also to select streaming quality/bitrate. This way, I could select a default rate while remote to prevent buffering/hogging all available upload, while also setting a default while local with the option to change the quality on the fly during playback. Not sure if that's on the radar or not, but seems like a useful feature.

bnkai:
@rg9400 the closer we have availabe to what you need is reducing the `Maximum streaming transcode size`. That would work for files that are live-transcoded but not for the direct streamable ones. If a force live-transcoding is implemented (as @victorhooi requests ) then that would probably be closer to what you want.

InfiniteTF:
@rg9400 I've been thinking about implementing something like that. Essentially force transcode and a resolution setting. Will see if I have time in the not too distant future, once this PR is merged.

rg9400:
> @rg9400 the closer we have availabe to what you need is reducing the `Maximum streaming transcode size`. That would work for files that are live-transcoded but not for the direct streamable ones. If a force live-transcoding is implemented (as @victorhooi requests ) then that would probably be closer to what you want.

Do you happen to know the bitrates associated with each of the options in the maximum size dropdown? For example, 720p can be anywhere from 1-4 mbps, and 1080p can be anywhere from 6-24mbps.

I definitely will use this option as a workaround for now, but I do think it can be enhanced down the line. Sometimes people are streaming from Plex, and I may need to drop my quality down even further than the default, and if I am local, maybe I want to play it at original quality. I think a default, library wide option is not really necessary, though would be convenient. But having the option during playback similar to how most streaming websites display it gives more flexibility. 

image

bnkai:
As it is now we use `crf` for encoding profiles. That means we don't target a specific bitrate but rather a specific quality. This gives better single pass video quality but the bitrate varies a lot depending on the scene ( noise, movement, complexity). For the vp9 webms we use `crf 30` i think which should be low enough for remote playing ( for 720p max at least).

victorhooi:
For the video player - would moving from JW Player to video.js help with any of this?

https://github.com/stashapp/stash/issues/659

WithoutPants:
I've done a major refactor of this PR. The `supported` parameters are gone. Instead, I have exposed new endpoints in the server:

```
/stream
/stream.mkv
/stream.webm
/stream.m3u8
/stream.ts
/stream.mp4
```

The first endpoint is a direct stream of the source file. The mkv endpoint is a video copy stream and the remaining transcode to the applicable type. `m3u8` gives the HLS manifest, with `ts` giving the individual streams. 

I have replaced `isSceneStreamable` with `sceneStreams`. This returns applicable stream endpoints for the scene. Direct stream endpoints are filtered out of the returned values if the audio codec is invalid for the container, so the client doesn't attempt to direct stream. The mkv stream endpoint is only included if the scene file is an mkv file.

On the client end, I have changed it so that jwplayer receives a list of source urls, allowing it to choose the correct compatible stream. In addition, if the stream fails due to an incompatible stream, then it will fail over to the next source. I am hoping that this will make the `force` options redundant and I have removed them once again.

WithoutPants:
From Discord, hearing that hevc/aac MKV files are not streaming correctly on iPad safari.

bnkai:
The same thing happens to me in chrome/linux. 
For the same file in firefox i get the infamous 224003 error so i guess something is wrong with the codec detection or the error retrieval from jwplayer.
 In chrome mkv is supported so it plays the audio and because of that it doesn't throw an error even though the video codec isn't supported.
Any mkv hevc files with aac audio should behave the same

WithoutPants:
The failover code was broken which is why you saw the 224003 error. For the no video issue, I've added some code that detects if the video metadata width and height are zero, and if so fails over to the next source.

pkg/api/routes_scene.go
```diff
@@ -42,110 +47,126 @@ func (rs sceneRoutes) Routes() chi.Router {
...
+		w.Write([]byte("not an mkv file"))
 		return
 	}
 
-	// needs to be transcoded
+	rs.streamTranscode(w, r, ffmpeg.CodecMKVAudio)
+}
+
+func (rs sceneRoutes) StreamWebM(w http.ResponseWriter, r *http.Request) {
+	rs.streamTranscode(w, r, ffmpeg.CodecVP8)
```
bnkai, line 92: should be `rs.streamTranscode(w, r, ffmpeg.CodecVP9)`

pkg/ffmpeg/stream.go
```diff
@@ -0,0 +1,244 @@
...
+	extraArgs: []string{
+		"-deadline", "realtime",
+		"-cpu-used", "5",
+		"-row-mt", "1",
+		"-crf", "30",
+		"-b:v", "0",
+	},
+}
+
+var CodecVP8 = Codec{
```
bnkai, line 90: given that only one webm endpoint is available VP9 should be the preferred codec
thus VP8 settings can be removed altogether.
If we choose to keep them they should be tuned a little bit 
From some tests to get acceptable quality i had to alter the `crf` to `12`, `-b:v` to `3M` ( 0 doesn't seem to work like it did for VP9) and add an extra `"-pix_fmt", "yuv420p",`

bnkai:
The error retrieval seems to  work now but VP8 is used instead of VP9 so it should be changed to VP9.
Hevc detection works now for most of the cases i tried except  Edge on android were instead of direct streaming hevc it converts to webm.

victorhooi:
I've been testing this on Linux x64 - these builds seem to remove the thumbnail functionality you normally get when hovering over the scrub bar.

For example, on YouTube:

image

pkg/ffmpeg/stream.go
WithoutPants: I've left the VP8 codec in there for now, with your revised settings.

WithoutPants:
Corrected the webm codec and fixed the thumbnail sprite regression.

I've removed `stream.mp4` from the `sceneStreams` return list, since having the second `mp4` stream appeared to be confusing jwplayer on iPad, causing it to try that stream over the HLS one.

bnkai:
Everything looks ok except the minor issue with microsoft edge and hevc in android mentioned above.

WithoutPants:
> The error retrieval seems to work now but VP8 is used instead of VP9 so it should be changed to VP9.
> Hevc detection works now for most of the cases i tried except Edge on android were instead of direct streaming hevc it converts to webm (custom hevc enabled linux chromium build direct streams hevc files without a problem).

Tested the current `develop` version on Edge Android. Confirmed that when I set `forceHEVC` and tried to stream an hevc file, I got the 224003 error. Also tried streaming directly (not using jwplayer) and got the same issue. I don't think Android Edge actually has reliable support for hevc files.

bnkai:
The hevc in android support should  be a combination of browser/android version/hardware.
I have a xiaomi mi note 3 and also a user with a oneplus reported in the channel that it was working for him. In theory the other browsers should support that but only Edge was direct streaming it.
I am not sure if something can be done about it and how many people use it.

WithoutPants:
I think I might have to chalk this up as a limitation at the moment:

The issue was introduced by the no video detection added in [3b816fe](https://github.com/stashapp/stash/pull/609/commits/3b816fecd85aab4b6acf41189da1b96a8702d1f0). If this code is removed, then it streams directly in Android Edge. 

The no video detection works by hooking into the `meta` jwplayer event, and causes a failover if the video's width and height are 0. It appears that on Android Edge, the width/height are 0, but the video is still able to be played. 

One alternative would be to detect Android Edge and disable the no video detection when on that client, but I just don't think it's worth the hassle.

bnkai:
I think we can leave that as a limitation especially if #679 can be used instead in android.

PR #616, opened 2020-06-17
Refactor xpath scraper code. Add fixed and map

### LINKED ISSUE #457: [Feature] Add fixed and mapped xpath scraper fields
There is currently no simple way to set a scraped field to a fixed value. Current workarounds are to set the xpath to an element that is always in the response and perform a regex replacement to the fixed value.

Likewise for the gender field, we'd want the ability to map scraped values to other values.

For a fixed value, config should look something like:
```
Gender:
  fixed: Female
```

For mapped values, something like the following:
```
Gender:
  selector: //div[class="example element"]
  map:
    F: Female
    M: Male
```

The above would take the value from the given selector, then translate it by setting values of `F` to `Female` and `M` to `Male`. Values that cannot be mapped are left unset.

It is probably possible to do the mapping using some sort of painful regex replacement.

WithoutPants:
Refactored the xpath code so that it uses structs instead of generic maps.

Changed post-processing format so that actions can be performed in any order. Maintains backward compatibility with existing post-processing actions: `parseDate`, `replace`, `subScraper`; but should be considered deprecated. New post-processing actions should only be implemented in the `postProcess` field.

Example of the old vs new structure:
```yml
# old
Birthdate: 
  selector: //span[@itemprop="birthDate"]
  parseDate: Jan 2, 2006

# new
Birthdate: 
  selector: //span[@itemprop="birthDate"]
  postProcess:
    - parseDate: Jan 2, 2006
    - # add other post-process actions as needed
```

Adds a `Fixed` field. This field replaces `Selector` in a field and resolves to the string value. For example:
```yml
Gender:
  Fixed: Female
```

Adds a `map` post-processing action. This is only allowed in the new `postProcess` field. It maps values to other values. Example:
```yaml
Gender:
  selector: //div[class="example element"]
  postProcess:
    - map:
      F: Female
      M: Male
```

The above config changes `F` to `Female`, `M` to `Male`. Eliminates the need for multiple regex actions.

Resolves #457


WithoutPants:
Going to mark as draft until #579 is merged.

WithoutPants:
Merged and adjusted for #579 changes.

bnkai:
I finished doing some regression tests using this build and the community scrapers (tested ~150 urls).
One issue i found is evident with this scraper `https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/Xartxxx.yml`
The `Details` get an extra space at the start for the specific site with this build (compared to dev). I think the postprocess after the concat is missing or something , i went through the code but didn't notice anything suspicious, maybe something when converting to the new format?
sample urls that show the issue
```
https://www.xart.xxx/anny-aurora-and-niki-surprise-sex-for-three/
https://www.xart.xxx/izzy-hot-russian-fucking/
https://www.xart.xxx/mary-kalisy-summertime-sex/
https://www.xart.xxx/rebecca-and-viktoria-cum-in-for-an-orgy/
https://www.xart.xxx/susie-sexy-movies-cum-inside/
https://www.xart.xxx/tiffany-and-amaris-sex-and-fashion-a-threeway-project/
https://www.xart.xxx/veronica-when-you-least-expect-it/
https://www.xart.xxx/zazie-hot-office-sex/
```

bnkai:
The example for the fixed shoud be
```yaml
Gender:
  fixed: Female
```

WithoutPants:
The commits aren't showing in the feed properly, but I did another significant refactor. I've converted it to use the same cache-style system that the plugin system PR uses. I've moved a lot of generic code into `mapped.go` which I plan to reuse for an upcoming json scraping change.

I've rethought the whitespace removal code. It now removes leading and trailing whitespace, and currently will only remove newlines when it first parses the node text. I'll need another round of regression tests, and I'll also consider disabling the newline removal when a specific flag is set on the attribute, but I'll need an example of when it is needed.

bnkai:
Just finished the regression tests. I haven't gone through the code yet.

Most differences i found were an extra leading or trailing whitespace on some scrapers.
**Urls tested**
scenes - https://pastebin.com/yq26r9RX
performers - https://pastebin.com/3YLSdCdG

scrapers that had differences in results
**SCENES**
_21naturals_-> https://pastebin.com/EcuTDu4Z 
The only difference is for a single URL `https://www.21naturals.com/en/video/Under-The-Coat/144299`
```diff
103c103
<     Details: Kinky Lana Roy is excited to meet up with her hot boyfriend Raul Costa. Lana surprises him by wearing lingerie under her new coat. She wraps her hands around his throbbing cock and deep throats it. Lana then presents her round ass for Raul to fuck. She takes every inch of his cock in her tight asshole and moans in pleasure before getting a hot blast of cum all over her pussy and ass.
---
>     Details: 'Kinky Lana Roy is excited to meet up with her hot boyfriend Raul Costa. Lana surprises him by wearing lingerie under her new coat. She wraps her hands around his throbbing cock and deep throats it. Lana then presents her round ass for Raul to fuck. She takes every inch of his cock in her tight asshole and moans in pleasure before getting a hot blast of cum all over her pussy and ass. '
```
Notice the trailing whitespace in the details field for the refactor build.


_cherrypimps_ -> https://pastebin.com/NgKABVrS
Again a single trailing whitespace for the refactor build in this url `https://cherrypimps.com/trailers/16588-kaneziereeves-vinasky.html`
```diff
22c22
<     Title: Wild Girls Kenzie Reeves And Vina Sky
---
>     Title: 'Wild Girls Kenzie Reeves And Vina Sky '
```

_mypervmom_ -> https://pastebin.com/upbRP0Mb
All scene details for tested scenes get an extra leading whitespace with the refactor build 
diff file -> https://pastebin.com/GQafkRmX

**PERFORMERS**
_freeones_
all performers tested get an extra leading whitespace in their birthdate field ( failing the parsedate as a result )
diff file ->  https://pastebin.com/kGFNgwnR

_babepedia_ -> https://pastebin.com/pqbfxb4v
all performers tested get an extra leading or trailing whitespace either in aliases or height
diff file -> https://pastebin.com/v2HTDdsu

WithoutPants:
Ideally, I'd leave it as is and change the scrapers - the replace regex should include the space to remove. However, I obviously don't want to break existing behaviour, so I've committed a change to trim the whitespace after replace operations. It shouldn't cause any problems down the track, but it's the sort of magic behaviour I'd like to avoid.

bnkai:
> Ideally, I'd leave it as is and change the scrapers - the replace regex should include the space to remove. However, I obviously don't want to break existing behaviour, so I've committed a change to trim the whitespace after replace operations. It shouldn't cause any problems down the track, but it's the sort of magic behaviour I'd like to avoid.

If it wasn't for the babepedia and freeones scrapers i would opt for removing the unpredicted "magic" behaviour even if it means changing/fixing the rest of the scene scrapers. Can we revisit this issue after 0.3 ?

bnkai:
With the latest commit it seems to be ok ( backwards compatible) with the existing scrapers, at least with the  tested URLs

PR #618, opened 2020-06-18
Move image blobs into separate tables

### LINKED ISSUE #425: [RFC] Changing image storage from DB blob to file system
## Scope
The goal of this would increase the performance of the server and database by separating the image data from the database itself and instead provide file paths to the image files in the database. 

There are definitely debates to be had over whether:
- a file system solution would be faster for the size of images
- the trade-offs (losing ACID on files) are worth it
- it's a good decision to change something that already works for increased performance, potentially only noticed by a small number of users

From my reading, it seems if the files being stored are over 1MB, it would be more efficient to use file system storage.

 [This article from SQLite](https://www.sqlite.org/intern-v-extern-blob.html) suggests that files over 100KB in size would be faster read from the file system rather than the database.

## Reference Reading
https://www.sqlite.org/fasterthanfs.html
https://www.sqlite.org/intern-v-extern-blob.html
https://dzone.com/articles/which-is-better-saving-files-in-database-or-in-fil
https://dba.stackexchange.com/questions/120091/why-do-large-images-stored-in-database-hurt-performance
https://www.microsoft.com/en-us/research/publication/to-blob-or-not-to-blob-large-object-storage-in-a-database-or-a-filesystem

WithoutPants:
Moves all images into separate tables:
* scene `cover` into `scenes_cover`
* performer `image` into `performers_image`
* studio `image` into `studios_image`
* movie `front_image`/`back_image` into `movies_images`

The performance gain is potentially quite large. On a database with ~1000 scenes, all with covers, the following graphql query went from 6.39 seconds to 477ms!

graphql query in question (tested using graphql playground):
```
{
  findScenes(filter: {
    page: 1,
    per_page:500,
    sort: "date",
    direction:ASC
  }) {
    count
    scenes {
      id
  checksum
  title
  details
  url
  date
  rating
  o_counter
  path

  file {
    size
    duration
    video_codec
    audio_codec
    width
    height
    framerate
    bitrate
  }

  paths {
    screenshot
    preview
    stream
    webp
    vtt
    chapters_vtt
  }

  scene_markers {
    id
    title
    seconds
  }

  gallery {
    id
    path
    title
  }

  studio {
    id
    name
    image_path
  }

  movies {
    movie {
      id
      name
      front_image_path
    }
    scene_index
  }

  tags {
    id
    name
  }

  performers {
    id
    name
    favorite
    image_path
  }
    }
  }
}
```

This required a schema migration that essentially rebuilds the entire schema, since removing columns requires rebuilding both the table and referencing tables. As such, I have performed the following verification steps to ensure there is no loss of data:
* ran `.schema` on the before and after databases and compared the results to ensure no fields or indexes were lost
* ran an export on the before and after versions and ensured that they were identical
* ran an import of the exported after data, then re-exported and ensured that the new export was identical to the previous one

The only part that hasn't been exercised by this is movies, since the database I was using didn't have them. I will test that in the meantime.

Other changes:
* added `utils.ServeImage` which performs the etag generation added previous and applies to all database images
* scene cover will fallback to use the database cover image if the generated image is not found
* perform a vacuum on the database after all migrations are completed
* add logging to the migration process
* change `Box.Bytes` to `Box.Find` to eliminate the deprecation message

Other things to consider:
* performer, studio and movie images are no longer mandatory in the schema. This means we can look at not generating these images at creation, but serve the correct "random" image based on the checksum of the object if the database image is null. This will also allow querying for performers/studios with no image
* the movie images table should be changed so that one of the images is mandatory, not just `front_image`

Resolves #425


pkg/database/migrations/10_image_tables.up.sql
```diff
@@ -0,0 +1,514 @@
...
+
+CREATE UNIQUE INDEX `index_movie_images_on_movie_id` on `movies_images` (`movie_id`);
+
+INSERT INTO `movies_images` 
+  (
+    `movie_id`,
+    `front_image`,
+    'back_image'
+  )
+  SELECT `id`, `front_image`, 'back_image' from `_movies_old` where `front_image` is not null;
```
bnkai, line 501: `'back_image'` here adds string back_image instead of blob to db
should be backtick like the rest 
it took me  a while to figure out why the back_images where missing after the migration :-)

WithoutPants:
Thanks for the heads up. I've since tested against some movies and fixed an unrelated issue that movie studios were not being exported or imported.

PR #620, opened 2020-06-18
Add custom served folders

### LINKED ISSUE #455: [Feature] Support configurable static folder
Allowing the user to configure a "static" folder would allow css themes to reference files outside of the stash binary and will also allow UI extensions or replacements. A simple example would be to add a background image to the custom css with the URL from the static folder. A more complex example might be if someone coded a replacement UI and placed the files within the static folder.

Add a (non-exposed) configurable static folder (or folders). This folder will be accessible from the front-end under the URL `/static`. May want to consider whether this folder should be accessible with or without logging in or if it should be an option.

WithoutPants:
Resolves #455 

Allows configuration of mapped URLs to file system folders. These are served from the `/custom` url prefix. For example, in config.yml:
```yml
custom_served_folders:
  /: D:\stash\static
  /foo: D:\bar
```
With the above configuration, a request for `/custom/foo/bar.png` would return `D:\bar\bar.png`. The `/` entry matches anything that is not otherwise mapped by the other entries. For example, `/custom/baz/xyz.png` would return `D:\stash\static\baz\xyz.png`.

This configuration is not exposed in the UI to minimise possible malicious use.

Applications for this include using static images in custom css, like the Plex theme. For example, I used the following config:
```yml
custom_served_folders:
  /: <stash folder>\custom
```
I placed `background.png` and `noise.png` in the `custom` folder, then in the custom css I put:
```css
body {
	background-image: url("./custom/background.png");
...
}

#root {
	background: rgba(0, 0, 0, 0) url("./custom/noise.png") repeat scroll 0% 0%;
...
}
```

Other applications might be to drop UI extensions into this folder, allowing people to extend the UI further. I tried putting the `ui/v2.5/build` directory into my configuration, but it's built to use specific URL paths, so didn't quite work correctly.


bnkai:
Tried using the css and accessing the folders directly from browser. Works ok.
Another trick you can  do is link the error.log ( i used soft links) to the custom folder and so have access to your whole log from inside stash.

PR #622, opened 2020-06-19
Gallery list improvement

WithoutPants:
First pass of gallery improvements.

Adds a grid view for galleries, which is now the default. Gallery card shows the thumbnail, path and the number of scenes. Shows a scene icon in the card if there is an associated scene. Mousing over the icon gives the title and clicking it navigates to the scene. Grid view is zoomable like the scenes grid.

Adds is missing filter for galleries missing scenes.

Changes the scanning and cleaning behaviour so that zip files with no valid images are not included.


bnkai:
Can you change the icon for the scenes in the galleries grid layout ? I think its a little misleading.  Instead of clicking on the title i ended clicking on the play button a couple of times even though the scene title was shown correctly

WithoutPants:
Changed so that the scene icon now navigates to the scene on click.

WithoutPants:
Resolves #473

PR #629, opened 2020-06-21
Integrate build version in UI

WithoutPants:
Integrates the build version information into the react UI app so that it no longer needs to communicate with the server to get the version. Also changes the changelog so that the current development version uses the stash version (eg `v0.2.1-11-ge50f1d01`) and build date.

image

Also adds a convenience target `ui-start` to run the UI locally, using the build version information. If the build information isn't set when running `yarn start` directly, then it defaults to `v0.3.0 (Development Version)`.


bnkai:
It seems to work ok.

PR #630, opened 2020-06-22
Support deleting multiple scenes

### LINKED ISSUE #239: Add "Delete" button to multi-scene selection page.
I am suggesting that a "Delete" button be added to the UI when multiple scenes are selected. This allows you to delete multiple scenes/files without having to click through each individual file.

WithoutPants:
This change has a few cosmetic changes. To begin with, the list filter (desktop and mobile views):

image
image

The query fields were groups together, and in order of filter, sort by, page size, since these seemed like the logical order for them. The filter button has been made part of the text query field.

The view controls have been moved to the right, while the operations button sits between the two groups. When one or more scenes are selected, new buttons appear next to the operations button:

image
image

These two buttons replace the new row that used to appear to facilitate editing multiple scenes, which has been removed. To edit the scenes, the user now clicks the Edit (pencil) button and a new dialog appears:

image

I have also changed the styling of all modal dialogs to match the dark styling of the rest of the UI.

This dialog operates the same way as the old fields did.

Clicking on the delete button brings up the delete confirmation dialog, which has been extracted into its own class:

image

The scene page has been changed to use this dialog as well.

A happy side-effect to this change is that after editing or deleting scenes from the scenes page, the query is now refetched, so scenes that no longer match the query are no longer shown.

Resolves #239

### DIFF
--- a/graphql/documents/mutations/scene.graphql
+++ b/graphql/documents/mutations/scene.graphql
@@ -80,6 +80,10 @@ mutation SceneDestroy($id: ID!, $delete_file: Boolean, $delete_generated : Boole
   sceneDestroy(input: {id: $id, delete_file: $delete_file, delete_generated: $delete_generated})
 }
 
+mutation ScenesDestroy($ids: [ID!]!, $delete_file: Boolean, $delete_generated : Boolean) {
+  scenesDestroy(input: {ids: $ids, delete_file: $delete_file, delete_generated: $delete_generated})
+}
+
 mutation SceneGenerateScreenshot($id: ID!, $at: Float) {
   sceneGenerateScreenshot(id: $id, at: $at)
 }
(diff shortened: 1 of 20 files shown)


bnkai:
Something is wrong with the population of the tags / performers when you select scenes.
Tick a single scene and hit the edit button. Tags and Performers are not populated in the edit menu.
Couldn't test for more than one scenes since it isn't loaded even for one.
All other functionality i tested seems to work fine and it makes multi editing a lot easier.

WithoutPants:
That was existing behaviour from #412 - you need to set the mode to Edit and the common tags will be populated.

bnkai:
just tried that it works.
shouldn't the studio,ratings behave the same then ? it was autopopulated and that made me expect the same for the rest.
 I understand that it follows the behaviour from before but the new button says edit. Shouldn't edit be the default action when you popup the menu?

WithoutPants:
Good point. I'll change the default modes for the multi-set controls.

PR #632, opened 2020-06-24
[RFC] Fix query toolbar, performer/studio/movie page styling improvements

WithoutPants:
Changed the query bar to use ButtonToolbar with ButtonGroups, rather than messing around with columns. Before and after images follow:
(on larger screens):

image
image
Note that the text field now maintains the same size regardless of screen width.

(on smaller screens)
image
image

These break appropriately when used as a sub-component, like in the performers page.

The Performer page has been adjusted so that the performer image is centred in its column and has a maximum height that is less than the viewport height. It still needs a little further refinement, but imo looks a bit better at all sizes.

The studio page no longer makes the logo fill the width. Again, this page needs another refinement pass, but I think it looks better.

The movie page has also been adjusted to look better in smaller viewports.

### DIFF
--- a/ui/v2.5/src/components/List/ListFilter.tsx
+++ b/ui/v2.5/src/components/List/ListFilter.tsx
@@ -12,8 +12,7 @@ import {
   SafeAnchorProps,
   InputGroup,
   FormControl,
-  Col,
-  Row,
+  ButtonToolbar,
 } from "react-bootstrap";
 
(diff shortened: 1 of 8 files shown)


WithoutPants:
A good way for seeing before/after is to build before checking out the branch and run as normal, then checkout the branch and use `yarn start` to load the local UI.

PR #636, opened 2020-06-26
Strip debug symbols from release builds

WithoutPants:
Adds a `build-release` make target, and changes the `release` target to use it. The `build-release` target adds `-s -w` ldflags when compiling, which strips out the debug symbols. This results in an executable that went from 60MB to 35MB for the Windows build.

This also includes a change to the cross-compile script to use make instead of building directly. The `-tags extended` arguments were taken out since I couldn't figure out why they were there in the first place.

~~Edit: build is broken due to an old git version in the compiler docker image. Will either update the docker image or revert the other changes for now.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,6 @@ ui-validate:
 # rebuilding the UI
 .PHONY: packr
 packr:
-	env
 	packr2
 
 # runs all of the tests and checks required for a PR to be accepted
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ endif
 release: generate ui build-release
 
 pre-build:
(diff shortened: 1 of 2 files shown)


bnkai:
For  the `-tags extended` i searched through the source code for `// +build extended` directives but nothing like that was present. I think it was either some extra code that removed in the meantime or maybe it was required by some library we no longer use.

WithoutPants:
Build is now fixed. Just need to test the generated binaries.

peolic:
@WithoutPants 
Windows seems to work fine for me. Is there anything special I should check?

bnkai:
Linux build compiles and looks ok also.

WithoutPants:
Thanks guys. Good enough for me.

PR #637, opened 2020-06-27
Add keyboard shortcuts

### LINKED ISSUE #81: [Feature] keyboard shortcuts
there should be keyboard shortcuts for quickly:

- [ ] rating a scene
- [ ] tagging a scene (with autocomplete)
- [ ] marking a scene
- [ ] adding performers to scene (with autocomplete)
- [ ] adding scene to studio (with autocomplete)
- [ ] ...

WithoutPants:
Resolves #81 (with caveats, see below).

Keyboard shortcuts are documented in the [manual page](https://github.com/WithoutPants/stash/blob/hotkeys/ui/v2.5/src/docs/en/KeyboardShortcuts.md).

**Note:** this PR does not include shortcuts for selecting fields (set scene performer, tags, etc). I will open a new issue covering this deficiency, as I was unable to work out how to do it.

Additionally, this includes a new dialog when pasting an image into the edit Movie page, which allows the user to select the image to set:

image


bnkai:
Looks ok.
In the manual page you have the "f" as show filter dialog when in the code is marked as TODO and doesn't work
The "c" create marker doesn't seem to work either ( or i am not using it right :-) )

Is it easy to highlight or underline the chars that are used for shortcuts when in the edit menu or nav panels ?
Something like firefox
firefox_under

WithoutPants:
Filter shortcut has been fixed, and the marker create shortcut has been corrected in the manual (its `n` in line with other "new" commands).

I don't think underlining shortcuts is a good approach, since it will lead the user to believe that the shortcuts are `Alt+<key>` since that is what they have historically meant. I'm averse to highlighting them in other ways since I feel it will detract from the appeal of the interface. If there's a way to do it that doesn't look gross _and_ doesn't mislead the user then I could consider it, but I'm not thinking of any good options at the moment.

bnkai:
Looks ok now.

PR #641, opened 2020-06-29
Add grid view, image to tag

### LINKED ISSUE #5: [Enhancement] Tag thumbnails
https://github.com/stashapp/StashServer/issues/32

WithoutPants:
Resolves #5

Adds schema migration to add tag image.

Adds grid view to tags page. Tags without images have a default image displayed (not stored), as shown below:

image

Clicking on the scene or marker icon in the card navigates to the applicable query view.

Leaves the old (ugly) view in under the list view.

Adds an individual tag page, allowing editing and viewing the scenes and markers with the tag:

image

Tags are now exported in their own folder.

Also adds a check to ensure tag name is unique at creation/modification. I did this rather than a schema migration in case people have duplicate tag names in their database.

~~Marking as draft while I add the following features:~~
* ~~add is missing image filter~~
* ~~add filter by scene/marker count~~
* ~~add sort by scene/marker count~~


bnkai:
Tested for a couple of days, it seems ok.

echo6ix:
This looks really good!

Not sure if this is the right form but here are two suggestions:

1. Add a button to delete an existing image.
2. When uploading an image, add a crop feature while the image is being uploaded, [such as this](https://vuejsexamples.com/tag/crop/), at a set aspect ratio, such as 1:1 or 4:3, so all images are the same aspect ratio. This will make the tag images look so much better if they all share the same aspect ratio.

WithoutPants:
@echo6ix can you raise this in a new issue please? :)

PR #644, opened 2020-06-30
Scrape dialog

### LINKED ISSUE #551: [Feature] Allow Scraper To Overwrite Data
Currently when using scene scrapers if there is already data in a field (like the title or description) it is not overwritten with scraped data.

This means, ever if the existing title and description is just the mess pulled from the file like `Lubed.18.04.14.Kimmy.Granger.Every.Last.Drop.XXX.1080p.MP4-KTR` for example, it does not get updated.

I'd like the scraper to overwrite the existing data. If we are worried about some people preferring the opposite, maybe this will need a config toggle. Personally I'd always want it to overwrite as long as the scraper didn't return a blank result.

Currently my work around is to manually delete the data from those fields before scraping the scene data, but this adds extra tedium to the process.

**Additional context**
N/A

### LINKED ISSUE #646: [Bug Report] New Performer UI Buttons Spacing Isn't Consistent
**Describe the bug**
The spacing between the Scrape with button and browse for image buttons doesn't match the space between the save and scrape with button.

**To Reproduce**
1. Visit Performers Section
2. Click on 'New'

**Expected behavior**
Spacing would be consistent

**Screenshots**
unknown

v0.2.1-22-gd516947

 - OS: macOS 10.15.5
 - Browser Chrome
 - Version Version 83.0.4103.116 (Official Build) (64-bit)

**Additional context**
Brought this up in discord, figured I'd make it issue official so it doesn't get lost.

WithoutPants:
Adds a dialog when scraping scenes and performers that shows the existing and scraped values and allows choosing between them.

Example scene dialog:

image

Unlike the above image, in general any field that doesn't have scraped values is not shown in the dialog.

Example performer dialog:

image

This dialog is not shown in the new performers screen - details are piped directly to the fields.

A future iteration on this is to present new tags/performers/studios and allow adding them from this dialog.

Resolves #551 
Fixes #646


peolic:
@WithoutPants 
Looks awesome!
I believe I have a bug report:
When you click to apply (I tried Performer),
all the fields that you don't select from the scrape data, even ones that don't show up, get emptied (except maybe the image).
If you click Save, it then saves and displays correct data.

bnkai:
Can confirm the above bug for the performer edit page.
For the scenes it doesnt seem to detect performers/tags even though they are scraped correctly, their field is empty  and they exist in the DB ( i couldn't test movies because of lack of samples).

WithoutPants:
Fixes #646

WithoutPants:
Thanks for testing this.

Made a few changes:
* fixed population of performer fields after applying
* scraping dialog now shows existing as the current values, rather than the original values. This means that if you edit the fields before scraping, the dialog will now show those new values, rather than the original values.
* performers/tags are now displayed correctly
* scene URL scrape button is now inline with the URL field, to be consistent with the performer edit page

Ready for retesting.

peolic:
Retested, looks good!

peolic:
Noticed one more thing:
Not sure if you want to do this as part of this PR or a separate one,
but the tags show up even if they are the same. Maybe only when more than one tag is matched.
Performers seem to work.

image

WithoutPants:
Should be fixed. Wasn't detecting ids in different order.

peolic:
@WithoutPants Yes, that fixed it.

PR #650, opened 2020-07-04
Scan for files with ALLCAPS extensions

bnkai:
_requested from the discord channel_
Enables stash to look for files with lowercase and uppercase supported extensions during the scan task. This is mostly needed for case sensitive filesystems (e.g. linux). 
It doesn't search for mixed case extensions since doing that would require either customizing doublestar glob, using another search function or generating a lot of extension combinations. If needed/wanted though this can be done also.

### DIFF
--- a/pkg/manager/manager_tasks.go
+++ b/pkg/manager/manager_tasks.go
@@ -18,18 +18,12 @@ var extensionsToScan = []string{"zip", "m4v", "mp4", "mov", "wmv", "avi", "mpg",
 var extensionsGallery = []string{"zip"}
 
 func constructGlob() string { // create a sequence for glob doublestar from our extensions
-	extLen := len(extensionsToScan)
-	glb := "{"
-	for i := 0; i < extLen-1; i++ { // append extensions and commas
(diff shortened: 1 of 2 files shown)


pkg/manager/manager_tasks.go
```diff
@@ -20,19 +21,20 @@ func constructGlob() string { // create a sequence for glob doublestar from our
...
+		glb += strings.ToLower(extensionsToScan[i]) + "," + strings.ToUpper(extensionsToScan[i]) + ","
 	}
-	if extLen >= 1 { // append last extension without comma
-		glb += extensionsToScan[extLen-1]
+	if extLen >= 1 { // append last extension without a comma at the end
+		glb += strings.ToLower(extensionsToScan[extLen-1]) + "," + strings.ToUpper(extensionsToScan[extLen-1])
 	}
+
 	glb += "}"
 	return glb
```
WithoutPants, line 32: Suggest the following alternative:

```
	// add capitalised extensions
	extList := extensionsToScan
	for _, ext := range extensionsToScan {
		extList = append(extList, strings.ToUpper(ext))
	}
	return "{" + strings.Join(extList, ",") + "}"
```

PR #657, opened 2020-07-08
Fix tag query performance problems

WithoutPants:
On systems with a large number of scene markers, querying for tags takes a really long time.

I discovered the cause to be related to the join between tags and scene markers on the `primary_tag_id` column. I haven't been able to optimise this query to make the query time reasonable, so for now I am disabling querying/sorting on the tag scene marker count.

I will leave this commented out until a more permanent solution can be found. 

Also fixes some issues with the tracing code.

### DIFF
--- a/pkg/manager/config/config.go
+++ b/pkg/manager/config/config.go
@@ -344,7 +344,7 @@ func GetLogLevel() string {
 	const defaultValue = "Info"
 
 	value := viper.GetString(LogLevel)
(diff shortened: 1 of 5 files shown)


bnkai:
I don't have a lot of scene markers to test the issue.
Loading the tags page as a list  i get 150ms for the findtags query for the dev build and 100ms  with your changes.
Looking at the query i can't also think of something more optimal.

PR #661, opened 2020-07-09
Use referer on xpath getImage, apply printHTML to subscraper also

bnkai:
In some cases like the below scraper mentioned in discord, images are protected by referer and so we get a `403 access denied` instead of the actual image.
Adding a referer based on the host of the image's url seems to get us the picture.
```yaml
name: AmateurAllure
sceneByURL:
  - action: scrapeXPath
    url:
      - amateurallure.com/tour/scenes/
      - amateurallure.com/tour/search
    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    common:
      $sceneinfo: //div[@class="scene-info"]
      $title: //span[@class='title_bar_hilite']
    scene:
      Title: $title
      Date:
        selector: //div[@class='cell update_date']
        parseDate: 01/02/2006
      Details: //span[@class='update_description']
      Tags:
        Name: //span[@class='update_tags']//a/text()
      Performers:
        Name: //span[@class='update_models']//a
      Image:
        selector: $title
        replace:
          - regex: \s
            with: "+"
          - regex: ^
            with: "https://www.amateurallure.com/tour/search.php?st=advanced&qall=&qany=&qex="
        subScraper:
            selector: //img/@srcset
            replace:
              - regex: .+(/tour/content/contentthumbs/\d+/\d+/[^/]+\.jpg) 1920w
                with: https://www.amateurallure.com$1
debug:
  printHTML: true     

# Last Updated July 09, 2020
```
_scraper is not yet fully tested, only used as a reference for the specific issue_
(test url:  `https://www.amateurallure.com/tour/scenes/elsa-dream-jean-blonde-teen-princess-sucks-fucks-swallows-sperm-pov_vids.html`)

I extended also the usage of 
```yaml
debug:
  printHTML: true     
```
to work for the subScraper

### DIFF
--- a/pkg/scraper/xpath.go
+++ b/pkg/scraper/xpath.go
@@ -25,6 +25,8 @@ import (
 // configurable at some point.
 const scrapeGetTimeout = time.Second * 30
 
+var debugMode = false
+
 type commonXPathConfig map[string]string
 
 func (c commonXPathConfig) applyCommon(src string) string {
@@ -259,7 +261,7 @@ func (c xpathScraperAttrConfig) applySubScraper(value string) string {
 	}
 
 	logger.Debugf("Sub-scraping for: %s", value)
-	doc, err := loadURL(value, nil)
+	doc, err := loadURL(value)
 
 	if err != nil {
 		logger.Warnf("Error getting URL '%s' for sub-scraper: %s", value, err.Error())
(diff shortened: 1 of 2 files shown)


pkg/scraper/xpath.go
```diff
@@ -25,6 +25,8 @@ import (
 // configurable at some point.
 const scrapeGetTimeout = time.Second * 30
 
+var debugMode = false
```
WithoutPants, line 28: Really want to avoid global variables if possible. I'll refactor this in a later change.

bnkai: Yeah i try to avoid as much as possible  also, but in the specific case it was a lot easier/quicker doing that.

PR #663, opened 2020-07-10
Allow adding performer & studio from scenes page

### LINKED ISSUE #547: [Feature] Add performers to Stash database from scene edit page.
Not a 'problem' so much, just an opportunity to remove friction from the process of adding metadata to new scenes. When adding new scenes, if the performer is not already in the database I have to change task to go add the performers then return to the scene page. That or I have to examine the files individually to list the performers involved and enter them in before proceeding to populate the scene metadata.

I'd like to see a way to add performers to my database from the scene edit page itself. These would likely be placeholder performer profiles with just the name added, to be filled out or scraped at a later date.

If using a scraper to fetch metadata, I might be able to add performers to the database straight from this process, though recognise this might not be an ideal solution and just being able to type the names in from the scene metadata page would suffice.

### LINKED ISSUE #548: [Feature] Add Performer & Studio directly from scenes page
When going through and adding data to scenes, I often find myself needing to add a performer and/or studio that have not been added before. Currently the only way to add them is via the Performer or Studio page, which means you need to leave the scene page temporarily while you add those, then come back to the scenes page, refresh the page, and then add the data. This is very very annoying cumbersome user experience.

Like the way adding tags work. When you type in the name of a Performer or Studio that doesn't exist, a new 'Create' option for the Performer or Studio should appear. Understandably, all of the extra data for Performer or Studio cannot be added or edited from this point, but simply, a new Performer or Studio with the given name should be created, and their respective details can be edited and added later from their own respective pages .

Another thing that might be more simple, but would not really count as a solution is to make the 'Create' option simply open up a new tab that takes you directly to the Create Performer or Create Studio page. This is not a solution, but at least it does help speed things along just a little

friendlycrab:
Adds the possibility to create performers and studios directly from the scene panel. I moved some duplicate code to a new `FilterSelectComponent`. To make re-fetching work, I had to replace some `update: () => invalidateQueries` in `StashService.ts` by `refetchQueries` and make the query names upper-case. Hope this doesn't break anything? `MovieSelect` had `type="studios"` which I fixed and it had a manual "None" entry which I removed because it's inconsistent with other Select components and doesn't really do anything. I introduced a new FilterSelectComponent to reduce duplicate logic.

image

Closes #548, closes #547.


bnkai:
One regression i found with this version is that when you are scraping a scene the performer/studio/tag fields are not shown as populated when you select the fields from the scraper popup and hit apply. Clicking on the scrape button next to the scene url again shows them as it should.

friendlycrab:
@bnkai Thanks for testing. Fixed by restoring the `invalidateQueries` calls I removed in c85f315b91bf781fcf120eb4abea1442ab9cb90a and making `invalidateQueries` query name matching case-insensitive via 5fb3a401fd70a16eef1012c1ed03a0247b76bbcf. I don't understand why this is needed though, supplying `refetchQueries` to Apollo should suffice in theory.

bnkai:
Even with the latest commit the issue is still present.

Go to a scene that has no performers, studio, tags.
Go to edit.
Fill a url  so that we can scrape the scene.
Hit the scrape icon (left of url).
In the popup that appears tick the studio,performers,tags from the scraped part and hit apply.
The relevant fields in the edit performer tab don't get updated.
Clicking again the scrape icon populates these fields.

friendlycrab:
Indeed, apparently I was fixing a different bug related to newly created performers. Thanks for your detailed instructions. I made all select components' "selected ids" property controlled via 6dd875e2acac2c1885498543fb95e1125f48acb1 as was the case for all except `TagSelect` before my changes. Can you confirm this fixes the issue?

bnkai:
Confirmed. Latest commit seems to fix the regression.

PR #667, opened 2020-07-14
Add oshash support

### LINKED ISSUE #351: [RFC] Add an extra "fast" hash/checksum for the scenes. Make MD5 scene calculation optional.
Currently the checksum we use for metadata filenames, duplicate file detection is MD5. 
While it is well established / tested it has a main disadvantage like all checksums, we need to read the whole file to calculate it. 
I personally don't have any problem waiting for the scan and like the ability to verify the integrity of my collection that MD5 gives , but it was mentioned on discord that a lot people ( especially with remote hosted collections , gdrive etc...) want/need something faster.

##  Proposal

I propose we add the ability to generate an extra  hash that is easy to calculate, doesn't read the whole file and hopefully doesn't have any collisions. 
My main recommendation and the hash i tested is oshash.
This can be added as an extra to the scenes table and calculated by default since it's generation time is negligible. Users who want to could opt to use this as their main  instead of MD5.

## Points to Consider

- MD5 is now required as the `checksum` in scenes. If we add a `hash` we could make the `hash` required and MD5 checksum optional (as mentioned the time needed for the proposed oshash generation is negligible ) or even make the existence of 'hash' OR 'checksum' required.

- Checksum and hash should cooperate nicely. 
A user must be able to opt in / out without the need to rename files , all existing metadata should work without changes. 
One potential way to solve this is checking first for existence using the checksum (if it exists) first and then with the hash. Thankfully MD5 has a fixed length of 32 chars while oshash 16 so they don't clash if we use them for the filename. The drawback is that there will be mixed filenames (some 16 chars, some 32) for those that choose to opt out of MD5 and already have some metadata.

- Oshash is already used by opensubtitles and seems to be collision free for media files ( tested for my 12k scenes and also reported so from  discord)

- Oshash needs a minimum file size of 64kb to be calculated. Anything less will not be.
I don't think i have come across a video that small , even sample that i use in my media collections but i suppose we could fallback to MD5 if we want to on rare cases.

- Oshash is mainly tested against media files. It reads 64k from the end and start of the file.I don't know if it will work well with the galleries. Further investigation will be needed if we choose to implement that ( priority should be given to scenes given the limited use/size of a gallery compared to a scene )

## Benchmarks

Oshash reads in total 128kb from a file so it's very fast.
Tests were done in linux using md5sum and the go util i have in reference
The media files were located localy in a spinning hdd. 
The main bottleneck for md5 was the disk read speed. 
Oshash was run before MD5 to make sure caching wasn't favoring oshash.
I used a folder containing 127 files (51Gb in total)

```
time oshash *.*

real	0m5.476s
user	0m0.171s
sys	0m0.081s

time md5sum *.*

real	6m57.624s
user	2m27.387s
sys	0m22.549s
```

To give a perspective of the speedup:
During scanning of a new file we calculate the checksum, generate a screenshot, a thumbnail and interact with the database.
The time needed for an average scene (size ~1gb)

```
checksum (md5sum) 5.8s (depends on disk read, file size, file fragmentation)
ffmpeg screenshot 0.2s
ffmpeg thumbnail 0.2s

hash (oshash) 0.06s 

```

Time to oshash my whole stashed collection  (12k scenes ~ 3.8TB) using a batch script to get the paths from stash database  ~ 7 minutes

## Reference
[oshash](https://trac.opensubtitles.org/projects/opensubtitles/wiki/HashSourceCodes) The official page from opensubtitles
[oshash.go](https://gist.github.com/bnkai/a09e75f163932bb91996a75d4dddc9f3) A md5sum like cmd line implementation of oshash i made from the above

WithoutPants:
Adds an `oshash` column to the database and makes `checksum` nullable. Added check to the scenes table to ensure at least one of the two are not null.

Introduces two new settings `calculateMD5` and `useMD5`. 

`calculateMD5` indicates whether or not to calculate the MD5 checksum during scanning. oshash will always be calculated for scenes where it is not already present. Similarly, if `calculateMD5` is true and the checksum is empty, then the MD5 will be calculated. 

`useMD5` indicates whether to use the MD5 checksum or the oshash for generated file naming and import/export naming.

For existing systems, these settings are both defaulted to true after migration if the scene table is not empty. 

New systems have `calculateMD5` and `useMD5` set to false.

The system prevents changing `useMD5` to false if there are any missing oshash values. Similarly, changing `useMD5` to true will be prevented if there are any scenes without checksums. Finally, `calculateMD5` is prevented from being set to false if `useMD5` is true.

image

Adds `hash` to the file info panel in the UI. Missing hashes are not displayed.

image

Adds a manual migration task that renames existing generated files (not including exported metadata files) to the current naming format.

image

The oshash algorithm has been verified manually using the sample files from the opensubtitles website.

Resolves #351


pkg/utils/oshash.go
```diff
@@ -0,0 +1,82 @@
...
+	var sum uint64
+	for _, v := range ints {
+		sum += v
+	}
+
+	// add the filesize
+	sum += uint64(fileSize)
+
+	// output as hex
+	return fmt.Sprintf("%16x", sum), nil
```
bnkai, line 81: oshash should be 0 padded to 16 digits so
```go
return fmt.Sprintf("%016x", sum), nil
```

WithoutPants:
I think it's probably worth extending this to include galleries as well.

bnkai:
My only concern for galleries would be how `unique` the hash is if we use zip files or files that can be <64k ( small zip file with a single image, or plain image in the future). I think it was mainly tested with video files.

WithoutPants:
Yeah, you're right. Might be better leaving it off.

pkg/manager/task_migrate_hash.go
```diff
@@ -0,0 +1,84 @@
...
+	newPath = scenePaths.GetStreamPreviewImagePath(newHash)
+	t.migrate(oldPath, newPath)
+
+	oldPath = scenePaths.GetTranscodePath(oldHash)
+	newPath = scenePaths.GetTranscodePath(newHash)
+	t.migrate(oldPath, newPath)
+
+	oldPath = scenePaths.GetSpriteVttFilePath(oldHash)
+	newPath = scenePaths.GetSpriteVttFilePath(newHash)
+	t.migrate(oldPath, newPath)
```
bnkai, line 66: the actual sprite image is missing

```go
	oldPath = scenePaths.GetSpriteImageFilePath(oldHash)
	newPath = scenePaths.GetSpriteImageFilePath(newHash)
	t.migrate(oldPath, newPath)
```

pkg/manager/task_migrate_hash.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (t *MigrateHashTask) migrate(oldName, newName string) {
+	oldExists, _ := utils.FileExists(oldName)
+	newExists, _ := utils.FileExists(newName)
+
+	if oldExists {
+		if newExists {
```
bnkai, line 74: Is there a reason not to overwrite existing files ?
If the user changes the naming hash/checksum and forgets to migrate the names then during the next scan/content generation he will get files with both naming schemes. Doing a migrate task afterwards will only print the warning and leave files as they are.

pkg/manager/manager_tasks.go
```diff
@@ -626,6 +639,54 @@ func (s *singleton) Clean() {
...
+	qb := models.NewSceneQueryBuilder()
+
+	go func() {
+		defer s.returnToIdleState()
+
+		logger.Infof("Migrating generated files for current naming hash")
+
+		scenes, err := qb.All()
+		if err != nil {
+			logger.Errorf("failed to fetch list of scenes for cleaning")
```
bnkai, line 658: even if the task is using the clean state isn't it more helpful to print something like "migrating" or "file renaming" or something like that?

pkg/manager/manager_tasks.go
```diff
@@ -626,6 +639,54 @@ func (s *singleton) Clean() {
...
+
+		for i, scene := range scenes {
+			s.Status.setProgress(i, total)
+			if s.Status.stopping {
+				logger.Info("Stopping due to user request")
+				return
+			}
+
+			if scene == nil {
+				logger.Errorf("nil scene, skipping Clean")
```
bnkai, line 675: same as above for `Clean` 
as a side note `Clean` doesn't have a status message so nothing is printed in the status when doing the name migration

bnkai:
It seems to work ok both with an empty/existing db.
Initial oshash scan for my exitsing ~4TB, 11K scenes (stored on local HDDs) took ~6 minutes.
The name migration took < 5s ( nvme ssd)

pkg/manager/task_migrate_hash.go
WithoutPants: Good point. Fixed.

WithoutPants:
Merged in from develop branch. There was quite a bit of conflicting changes, so may need a retest.

Fixed up the logging and status stuff identified. Also fixed a nasty panic when running clean if a file is not accessible.

pkg/manager/task_scan.go
```diff
@@ -182,15 +223,33 @@ func (t *ScanTask) scanScene() {
...
+			return
+		}
+	}
+
+	sceneHash := oshash
+	if t.useMD5 {
+		sceneHash = checksum
+		scene, _ = qb.FindByChecksum(sceneHash)
+	} else {
+		// TODO scene = findbyoshash
```
bnkai, line 248: is a 
```golang
scene = qb.FindByOSHash(sceneHash)
```
here enough to remove the TODO?

pkg/manager/task_migrate_hash.go
```diff
@@ -0,0 +1,86 @@
...
+
+	oldPath = scenePaths.GetSpriteImageFilePath(oldHash)
+	newPath = scenePaths.GetSpriteImageFilePath(newHash)
+	t.migrate(oldPath, newPath)
+}
+
+func (t *MigrateHashTask) migrate(oldName, newName string) {
+	oldExists, err := utils.FileExists(oldName)
+	if err != nil {
+		logger.Errorf("Error checking existence of %s: %s", oldName, err.Error())
```
bnkai, line 76: should we switch from error level to debug or warning here?
It pollutes the log with `Error checking existence of ...` for marker,webp and transcoded files ( for the majority of my scenes i don't have those generated). The specific files not existing shouldn't be an error

bnkai:
Retested seems ok.

pkg/manager/task_migrate_hash.go
WithoutPants: Should've only been logging due to other error. Fixed.

bnkai:
I finished retesting. It wasnt an exhaustive test though, more of a check for regressions due to refactoring test. Looks ok.

For users with existing databases i think a step by step migration guide could be useful.
Eg.

1. Scan library ( updates the oshash for existing entries)
2. Untick UseMD5
3. Choose oshash as the naming scheme
4. Migrate

Issue #634, opened 2020-06-25
[Feature] Skip intro of scenes of specific sources


ConstantMoo:
Hi and thanks for this great project I discovered a few days ago. Here goes my first issue for it! :shipit: 

**Is your feature request related to a problem? Please describe.**

When browsing in wall mode, scenes from sources such as Pornhub have 2-3 sec intros that don't need to be in the preview imho. For example, as of now, the first frame of my wall looks like this:

![dfdfgf](https://user-images.githubusercontent.com/67415772/85719383-a9c63680-b6ef-11ea-8cde-04b84ee54466.png)

Depending on the content, some studios and performers also have they own intro.

**Describe the solution you'd like**

1. Not sure how the project is working and I never coded in Go, but my assumption would be to add an `Intro Length (sec)` field to `Studio`, `Movie` and `Performer` classes. Default value would be 0 which means there's no intro.

2. Then, when a scene is linked to a studio/movie/performer, the user has the ability on the edit scene page to select either:
  - Studio intro length
  - Movie intro length
  - Performer intro length
  - A custom value in case the scene is specific

3. Upon scene preview generation, the generation function would look for a intro length field value and if it's different than 0, the first frame would skip by the amount of the field value.

- It would also be possible to set a preference order for which default intro length value to use when a scene is linked to more than one entity.

- There's also the case when a scene is from a Studio that doesn't have an intro unless it's hosted on Pornhub. For these cases, we could add a `Video Provider` modifiable list somewhere in the admin panel with their name and the Intro Length value. This would be an additional option in the edit scene page.

**Describe alternatives you've considered**

I don't see alternatives beside transcoding my library of files containing such intros, which would take a while.

**Additional context**
N/A

rg9400:
+1. I think the solution you described is a bit complex and manual. If we just blankety offset the first frame by a few seconds, is there any harm? Even scenes without intros wouldn't be impacted too much I think

Shadesbird:
https://github.com/stashapp/stash/pull/673 added ability to skip first X of every scene for preview generation.

@glxzander created https://github.com/stashapp/stash/issues/6209 after the feature was introduced, so there seems to be a need for scene specific options?

I am with @rg9400 here: I would have suggested to set the default to a few seconds and close this issue... But that is only because I could not come up with a reason why it would be so valuable to have the preview start exactly at second 0 of the main content.

[...]

PR #668, opened 2020-07-16
Skip Intro for video previews

rg9400:
Entire credit goes to Fonzie#5266 on discord, but he suggested the above code as a way to skip the beginning portion of video previews as that is often disclaimers or title cards for the scene. Since this blanketly applies the skip, it would impact scenes without these intros, but I think the skip is small enough that it should not make a difference for those scenes. 

I think it is worth analyzing if 2% is the best way to skip the intro. We could offset by a static amount of time (30 seconds) to deal with shorter videos or bump up the 2%. Not sure on this.

Code is what Fonzie suggested, but it seems to make sense to me unless it has impact on other areas as well. I just moved the twoPercent variable out of the for loop since I didn't think it needed to be calculated each time.

I think this is a crude solution to #634 , but it should solve most people's usecases. To meet the requirements for that Feature Request would require more work, and it may create additional complexity with limited benefit

### DIFF
--- a/pkg/manager/generator_preview.go
+++ b/pkg/manager/generator_preview.go
@@ -93,8 +93,9 @@ func (g *PreviewGenerator) generateVideo(encoder *ffmpeg.Encoder) error {
 	}
 
 	stepSize := int(g.Info.VideoFile.Duration / float64(g.Info.ChunkCount))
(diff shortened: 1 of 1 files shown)


WithoutPants:
Refer to the alternative PR #673. If that is successful, I will close this.

PR #673, opened 2020-07-17
Allow customisation of preview generation

WithoutPants:
Alternative to #668 

Provides an alternative solution to #634 

Exposes a number of new settings for preview video generation:
* number of segments (default 12)
* duration per segment (default 0.75s)
* exclusion start (expressed in seconds or a percentage of total duration (ie `2%`), default 0)
* exclusion end (as above)

image


bnkai:
If we take into account #672 isn't it more useful though if at least the start/end time exclusions are in the generate dialog/ task page as options and not in the config?
For most use cases i would imagine that changing the default start/end shouldn't be applied to all newly scanned scenes but only to those that are grouped together by the user ( eg for a specific studio. Different studios usually have different length intros/endings ).

WithoutPants:
Good idea. I'm thinking that we keep the existing settings in there which will be used as defaults, and allow the generate dialog to override them. Thoughts?

WithoutPants:
Added Preview Options to the Generate dialog:

image
image

These values are filled in from the configuration settings so that the user can customise them appropriately.

bnkai:
It seems to work fine for me.
The override option in the generate dialog looks great.

The only thing that troubles me a bit is the cache usage . I tried force reloading in firefox/chrome to get the new overwritten preview instead of the old one with no success (preview file was generated/overwritten correctly). I had to use the dev tools in chrome and set the no-cache option to  get the new preview to play.

WithoutPants:
I've changed the preview response to include `no-cache` in the header, but I think that this is a client-side issue and there's little we can do. Closing the tab and reopening does seem to fix the issue.

bnkai:
My main concern is people reporting that it doesn't work because they can't see the newly created preview. The no-cache should help though for the newly generated files. Imho it good enough as it is now.

PR #679, opened 2020-07-19
Add button to scene page to open scene in external player

InfiniteTF:
Uses video intent on android so it should pop up app selector for video. Uses M3U in desktop/IOS. Unfortunately the only crosscompatible way on desktop is using audio/x-mpegurl which indicates audio file, but I'm assuming not that many people use M3Us anymore, so reassigning to VLC or whatever won't bother people.

I haven't tested IOS but it should hopefully work. It uses the HLS M3U mimetype which I'm led to believe will be interpreted correctly.

Note that Chrome will always download the m3u even if you select always open. There's no way around this. Firefox instead downloads the m3u to your tmp dir so it's not an issue.


WithoutPants:
I think we're going to be fielding a lot of support requests about this feature not working correctly on Chrome. Actual support seems flaky at best on desktop PCs. If we are going to include this, then I think some alternatives need to be explored:
* consider making enabling the button a configurable option, with some documentation to indicate the limitations
* consider only showing the button for certain platforms/browsers. It seemed to work ok on my Android phone, but it was very clunky on Windows Chrome and Edge, due to the limitations you described.
* consider moving it into File Info instead of prominently in the scene tabs.

WithoutPants:
Gave it a quick test on iPad Safari. The button opened a new tab in Safari playing the video, much like the direct stream link.

bnkai:
Works ok for me in firefox/linux (vlc as the player) and android.
Chrome / opera and probably the rest of the chromium based ones are a fail in linux as mentioned.

w1ldg00se:
Can confirm this works on android. Some suggestions for improvement:
- you could use the package `react-device-detect` and just check for `isAndroid` and `isIOS` instead of manually checking the useragent.

- in the android intent add the title of the video via `...;type=video/mp4;S.title=${encodeURI(title)};end`, this way the video player displays the correct title (currently it only displays 'Stream' as title). Works on 'MX Player' and 'VLC for Android' and probably on most other video players that display the title.

- the feature of the new button was explicitly made for android and ios devices, which have touchscreens. But you have to touch the video from the scenes overview to get to the scene detail page, scroll down and touch this really tiny button to open the video externally. Not the best touch experience. Maybe place a big (semi-)transparent button over the upper left edge of the video player or so? Or just replace the whole video player with a preview image and make that a link to the external player (with a setting to turn that on/off in the configuration page). And it would be nice if this feature would also be available in the scenes overview page so one doesn't always have to go into the scene detail page.

Btw, i currently auto build the develop branch with a patch which gives me that external player feature, so no need to hurry :)

milindpatel63:
Do you have the pre built binary file for linux with this button added to it...?
I am trying to build one but have some errors..

go generate -mod=vendor
# runtime
/usr/local/go/src/runtime/mgcsweepbuf.go:87:80: memstats.gc_sys undefined (type mstats has no field or method gc_sys)
/usr/local/go/src/runtime/mgcsweepbuf.go:106:102: memstats.gc_sys undefined (type mstats has no field or method gc_sys)
main.go:1: running "go": exit status 2
make: *** [Makefile:63: generate] Error 1

PR #682, opened 2020-07-20
Fix performer card name truncation

InfiniteTF:
Shuffled performer-card styles around so the width is defined on the card rather than the image, that way truncations works again. Also renamed `card-image` to `performer-card-image` to clarify that it only applies to those cards.


WithoutPants:
Breaks the thumbnail for me. With the current build, the thumbnail image is cropped to be top-centre:

image

In this branch, that positioning is lost:

image

InfiniteTF:
@WithoutPants Good catch, should be fixed now.

PR #698, opened 2020-07-31
Fix markers not autoplaying

### LINKED ISSUE #687: [Bug Report] Broken Marker Seeking
After clicking on a marker preview, the video player fails to seek. (The progress bar shows the seek but video actual starts at the start).

Example of the player url:
http://host/scenes/53309?t=1558

** Additional Context ** 
Happens with all markers
mp4 (h264+aac) video
no ffmpeg tasks running while playing.

**To Reproduce**
1. Goto Markers
2. Click a marker

**Expected behavior**
The player will seek to the marker and begin playing.

** Actual Behaviour **
The player doesn't seek to the marker and begins playing from the start. 

Following on: Clicking on the marker causes the player to seek, but the whole progress bar is messed up. The progress bar indicates being at the end, the time progress display, say bigger number (current timecode) over a small number (duration), eg 56:11 / 27:46. 

My branch is up to this: (I don't believe my modification would be affecting this issue).
commit c327a98ecbd6783c924bf052eb090b9fa7fc372d (upstream/develop)
Date:   Fri Jul 24 10:45:48 2020 +1000

Stash on linux amd64 server.
Browser on OSX 10.15.4 Chrome (Version 84.0.4147.89)
Also checked on Firefox.

WithoutPants:
Combined `firstFrame` event behaviour into single declaration. Added auto-start behaviour when navigating to a timestamp. 

Fixes #687 

The original problem was frustratingly intermittent, but I can't seem to reproduce it with this change, so I hope it has resolved it.

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
+++ b/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
@@ -95,9 +95,6 @@ export class ScenePlayerImpl extends React.Component<
 
   private onReady() {
     this.player = JWUtils.getPlayer();
-    if (this.props.timestamp > 0) {
-      this.player.seek(this.props.timestamp);
-    }
 
     this.player.on("error", (err: any) => {
       if (err && err.code === 224003) {
@@ -113,6 +110,12 @@ export class ScenePlayerImpl extends React.Component<
       ) {
         // treat this as a decoding error and try the next source
         this.handleError();
+      }
+    });
+
+    this.player.on("playlist", () => {
+      if (this.props.timestamp > 0) {
+        this.player.seek(this.props.timestamp);
       }
     });
   }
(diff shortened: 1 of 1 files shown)


bnkai:
I didn't notice the issue with firefox/chrome in the previous PR. Testing this also looks ok with those browsers.

I tried testing edge from my android phone ( no linux binary ) and markers will not autostart ( chrome works ok). Touching a marker gets you to the scene page were if you hit play it plays the marker, it just doesn't autostart. Marker images are also not diplayed in edge and the default android browser while they display fine with chrome.

WithoutPants:
I retested against Android Edge and Samsung default browser and I'm not getting any of the issues you described. Markers appear to autoplay correctly and the marker images also displayed fine on both. I'll merge this for now and see if there's any more reported issues.

Issue #587, opened 2020-06-01
[Feature] Convert Feet/Inches to CM when scraping


halorrr:
**Is your feature request related to a problem? Please describe.**
Currently when creating a performer scraper, if the site doesn't include the height in cm ([example site](https://vrcosplayx.com/cosplaygirl/liyasilver/)), the height data becomes useless since stash expects the height in cm.

**Describe the solution you'd like**
Similar to how we work with multiple date formats, it would be useful to have a heightParse post-processing option.

**Describe alternatives you've considered**
Doesn't seem any way aside from manual to handle this currently.

**Additional context**
A quick Google search turned up this which may be helpful:
https://stackoverflow.com/questions/44091338/convert-feet-and-inches-to-centimeters-in-go

bnkai:
For the date we use time.Parse that is already available in go.
Making an equivalent would be too complicated imho.
Its way easier to just use a separator eg like concat does to separate the feet from the inches. 
`5'3` or `5ft3` ...
In some cases inches are missing so that should be taken into account.

halorrr:
@bnkai but I thought the db expects the height to be in cm? So concating doesn't really do much here?

bnkai:
I meant **like** concat, instead of parsing the 5'3 or 5ft3 or whatever just use the `'` or the `ft` as separator and get the feet,inches that we then can convert to cm. It will have more usage limitations but it will be easier to implement.

woodgen:
I propose a solution for this in #711

PR #711, opened 2020-08-07
scraper/mapped: Add feetToCm post process.

woodgen:
This patch adds a toMetric post process that converts imperial feet and
inches to centimeters.

This implements feature request https://github.com/stashapp/stash/issues/587

For testing try my Jules Jordan performer scraper here:
https://github.com/woodgen/CommunityScrapers/tree/ea%2Bjj

### DIFF
--- a/pkg/scraper/mapped.go
+++ b/pkg/scraper/mapped.go
@@ -396,6 +396,9 @@ func (a mappedPostProcessAction) ToPostProcessAction() (postProcessAction, error
 		ret = &action
 	}
 	if a.FeetToCm {
+		if found != "" {
+			return nil, fmt.Errorf("post-process actions must have a single field, found %s and %s", found, "feetToCm")
+		}
 		found = "feetToCm"
 		action := postProcessFeetToCm(a.FeetToCm)
 		ret = &action
--- a/pkg/scraper/mapped.go
+++ b/pkg/scraper/mapped.go
@@ -3,8 +3,10 @@ package scraper
 import (
 	"errors"
 	"fmt"
+	"math"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/scraper/mapped.go
```diff
@@ -276,11 +278,33 @@ func (p *postProcessMap) Apply(value string, q mappedQuery) string {
...
+	reg := regexp.MustCompile("[0-9]+")
+	filtered := reg.FindAllString(value, -1)
+
+	var parsed []float64
+	for _, v := range filtered {
+		k, _ := strconv.ParseFloat(v, 64)
+		parsed = append(parsed, k)
+	}
+
+	var centimeters = parsed[0]*foot_in_cm + parsed[1]*inch_in_cm
```
WithoutPants, line 296: This is going to panic if the input only has a single number or doesn't have numbers at all.

```suggestion
	var feet float64
	var inches float64
	if len(filtered) > 0 {
		feet, _ = strconv.ParseFloat(filtered[0], 64)
	}
	if len(filtered) > 1 {
		inches, _ = strconv.ParseFloat(filtered[1], 64)
	}

	centimeters := feet*foot_in_cm + inches*inch_in_cm
```

pkg/scraper/mapped.go
```diff
@@ -276,11 +278,33 @@ func (p *postProcessMap) Apply(value string, q mappedQuery) string {
...
+	// Return rounded integer string
+	return strconv.Itoa(int(math.Round(centimeters)))
+}
+
 type mappedPostProcessAction struct {
 	ParseDate  string                   `yaml:"parseDate"`
 	Replace    mappedRegexConfigs       `yaml:"replace"`
 	SubScraper *mappedScraperAttrConfig `yaml:"subScraper"`
 	Map        map[string]string        `yaml:"map"`
+	ToMetric   bool                     `yaml:"toMetric"`
```
WithoutPants, line 307: I think `toMetric` is too generic for this purpose. `feetToCm` would probably be more suitable.

woodgen: This was my first idea as well, but then I noticed it's feet and inches to cm :D Anyway, I will take your suggestion as it's more intuitive.

pkg/scraper/mapped.go
woodgen: Thanks for the patch, this is indeed more safe.

pkg/scraper/mapped.go
```diff
@@ -368,6 +395,11 @@ func (a mappedPostProcessAction) ToPostProcessAction() (postProcessAction, error
 		action := postProcessMap(a.Map)
 		ret = &action
 	}
+	if a.FeetToCm {
+		found = "toMetric"
```
bnkai, line 399: `found = "toMetric" ` after the refactoring shouldn't it be `feetToCm` ?

woodgen: Oops, missed that one.

PR #716, opened 2020-08-09
Don't play video after failover

### LINKED ISSUE #714: [Bug Report] Some videos are auto-playing when the option is unticked
**Describe the bug**
On the scene player, some videos are auto-playing when you view them, this is when the "Auto-start video" is unticked. This only occurs with certain codecs. In my case, this is 'mpeg4' codec. This may actually apply to any codec that Stash needs to transcode but that will need other users input, as I dont have any other files to test (I dont download HEVC for example)

Version: v0.2.1-50-g5992ff8
Build hash: 5992ff8
Build time: 2020-08-06 01:28:23

 - OS: Synology NAS (uses linux)
 - Browser: Chrome
 - Version: 84.0.4147.105 (Official Build) (64-bit)

**Additional context**
It may well be a red herring that it's any file that needs transcoding is auto-playing, but that will require input here from other people affected (at least one other person has reported this on Discord so far)

Thanks

WithoutPants:
Fixes #714 

If a video stream failed to play and the UI used the failover to load the next stream, then the newly loaded stream would be played automatically. Removed call to `play` after loading the next stream.

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
+++ b/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
@@ -154,7 +154,6 @@ export class ScenePlayerImpl extends React.Component<
       // eslint-disable-next-line no-console
       console.log("Trying next source in playlist");
       this.player.load(this.playlist);
-      this.player.play();
     }
   }
 


bnkai:
Just tested against some hevc videos.
Looks ok

PR #718, opened 2020-08-09
Fix sprite vtt panic

WithoutPants:
If the sprite vtt was generated using oshash and the scene did not have an MD5, then the system would panic when trying to get the sprite image. Now determines whether MD5 or oshash is passed, using the length of the string, then calls the applicable function. Now also gives a 404 if the scene could not be found.


bnkai:
Lol missed that during testing. Works ok and no need to edit the vtts.

PR #721, opened 2020-08-11
Make scene movie number freely editable

### LINKED ISSUE #683: [Feature] make "Scene number" field writeable.
As of now when you assign a Clip to a Movie, the Scene number counter "only" goes to 10.
But there are cases when that is not enough, so it would be great if we could just write a number in there.

WithoutPants:
Resolves #683 

My personal view is that scene # should be text rather than number, but that would require a schema change, and a bit more substantial changes, so going with a quicker fix for now.

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v030.tsx
+++ b/ui/v2.5/src/components/Changelog/versions/v030.tsx
@@ -19,6 +19,7 @@ const markup = `
 *  Add support for parent/child studios.
 
 ### 🎨 Improvements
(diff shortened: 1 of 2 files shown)


bnkai:
Seems ok. Should we limit the number to only positives though ?

WithoutPants:
Don't think it's worth the effort putting the extra validation in. A future iteration will be to make a text-based, so I'm not keen on adding a validation for negative numbers.

PR #722, opened 2020-08-11
Clear image

### LINKED ISSUE #660: [Feature] Tag grid view enhancements (delete button, image cropper)
Re: #641

1. Add a button to delete an existing image. There's currently no way to remove a tag image.
2. When uploading an image, add a crop feature while the image is being uploaded, [such as this](https://vuejsexamples.com/tag/crop/), at a set aspect ratio, such as 1:1 or 4:3, so all images are the same aspect ratio. This will make the tag images look so much better if they all share the same aspect ratio.

(Side note: the image cropper idea should probably also be extended to actors image and scene image uploads).

WithoutPants:
Partially resolves #660 

Adds a clear image button to the edit tags, performers, studios and movies (front and back) pages.

image

Performer image now generates a random image using the name as a seed, and uses that image if there is none set. All other types use their default image if none is set.

Movie front image is a little bit of a hack. If the back image is set but the front image is not, then the front image will be set in the database from the default image, since the front image cannot currently be nulled, and I didn't want to add a schema change to fix it. This will mean unexpected behaviour when querying for missing front image if the back image is set.

Adds is missing image for tags, studios, performers and movies. Note that existing users will find that is missing will not find results for their existing data which had the default images populated (they aren't null).

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -89,11 +89,15 @@ input SceneFilterType {
 input MovieFilterType {
   """Filter to only include movies with this studio"""
   studios: MultiCriterionInput
+  """Filter to only include movies missing this property"""
+  is_missing: String
 }
 
 input StudioFilterType {
   """Filter to only include studios with this parent studio"""
   parents: MultiCriterionInput
+  """Filter to only include studios missing this property"""
+  is_missing: String
 }
 
 input GalleryFilterType {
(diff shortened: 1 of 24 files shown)


bnkai:
For movies that only have a front cover clicking on the **clear front image** and then saving gives you
`ErrorGraphQL error: NOT NULL constraint failed: movies_images.front_image`.
Clicking on **clear back image** afterwards and then hitting save lets you clear the front image

WithoutPants:
Movie image clearing should be fixed now.

PR #741, opened 2020-08-16
Show and allow creation of unknown performers/tags/studios/movies from scraper dialog

WithoutPants:
Shows performers/tags/studios/movies that are unknown to the stash database under the applicable entry in the scraper dialog, with a button to create and add the object to the new value.

image

I'm not sold on the presentation, so let me know if there's a better way to present this information.

Clicking on the + button removes the entry from the list and adds it to the applicable field.

Also changes the scraping operations so that they do not cache the results.


bnkai:
Testing seems ok.
Not sure on how it would look better but as a first note the tags section should probably be last in order as some sites use a lot of tags (30+). A suggestion from the discord channel for the tags section was for something like a checkbox to show/hide tags that are not present in the DB (switch between dev behaviour and the new from this PR).

WithoutPants:
I've changed the missing tags to use clickable badges. I've also added a collapse element that hides the missing tags if there are >= 10 tags.

image

image

PR #745, opened 2020-08-17
SafeMove

JoeSmithStarkers:
- Replace os.Rename with util.SafeMove to allow cross device moving.
This issue occurs when you mount parts of your generated folder from other devices. For example say temp is on a local drive but your transcode directory is on network share.

- Performance improvement in SpriteGeneration by using image.DecodeConfig instead of loading the whole image.

- Fixed annoyingly noisy transcoding progress log messages.
Output one log message per 1 % change in progress.

- Fixed minor minor issue with directories that are named "blah.mp4" being detected as files to be scanned.
Check each file to determine if it is a directory.

### DIFF
--- a/pkg/ffmpeg/encoder.go
+++ b/pkg/ffmpeg/encoder.go
@@ -100,6 +100,7 @@ func (e *Encoder) run(probeResult VideoFile, args []string) (string, error) {
 	}
 
 	buf := make([]byte, 80)
+	lastProgress := 0.0
 	var errBuilder strings.Builder
 	for {
 		n, err := stderr.Read(buf)
@@ -108,7 +109,11 @@ func (e *Encoder) run(probeResult VideoFile, args []string) (string, error) {
 			time := GetTimeFromRegex(data)
 			if time > 0 && probeResult.Duration > 0 {
 				progress := time / probeResult.Duration
-				logger.Infof("Progress %.2f", progress)
+
+				if progress > lastProgress+0.01 {
+					logger.Infof("Progress %.2f", progress)
+					lastProgress = progress
+				}
 			}
 
 			errBuilder.WriteString(data)
--- a/pkg/manager/generator_preview.go
+++ b/pkg/manager/generator_preview.go
@@ -139,7 +139,7 @@ func (g *PreviewGenerator) generateImage(encoder *ffmpeg.Encoder) error {
 	if err := encoder.ScenePreviewVideoToImage(g.Info.VideoFile, 640, videoPreviewPath, tmpOutputPath); err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


pkg/manager/manager_tasks.go
```diff
@@ -99,7 +99,13 @@ func (s *singleton) Scan(useFileMetadata bool) {
 		for _, path := range config.GetStashPaths() {
 			globPath := filepath.Join(path, "**/*."+constructGlob())
 			globResults, _ := doublestar.Glob(globPath)
-			results = append(results, globResults...)
+
+			for _, globCheck := range globResults {
```
bnkai, line 103: Imo checking for directories shouldn't happen here. It adds a lot of os.stat calls and thus delay for users with network mounted dirs (or even worse gdrive mounts ).
I would check for dirs/filter the results  in the task_scan.go file, when scaning the scene/gallery. 
If it is done there after the db findbypath 
-  if the path is in the DB  it would filter out all the files that are already in the db since directories shouldnt be there anyway due to checksum calculation error
- if the path isnt it would checksum it anyway so the os.stat doesn't `cost` us

JoeSmithStarkers: Thanks @bnkai for the excellent idea of doing it further down in the scanning process. Always felt this change added a huge amount of overhead for a very minor issue, caused by an edge case. I'll rework it.

JoeSmithStarkers: Reworked.

bnkai:
Just for completeness i would update the code in pkg/manager/task_scan.go 
from
```golang
func (t *ScanTask) associateGallery(wg *sync.WaitGroup) {
	qb := models.NewGalleryQueryBuilder()
	gallery, _ := qb.FindByPath(t.FilePath)
	if gallery == nil {
		// shouldn't happen , associate is run after scan is finished
		logger.Errorf("associate: gallery %s not found in DB", t.FilePath)
		wg.Done()
		return
	}
```
to something like that
```golang
func (t *ScanTask) associateGallery(wg *sync.WaitGroup) {
	qb := models.NewGalleryQueryBuilder()
	gallery, _ := qb.FindByPath(t.FilePath)
	if gallery == nil {
                // associate is run after scan is finished
		// should only happen if gallery is a directory or an io error occurs during checksumming
		logger.Warnf("associate: gallery %s not found in DB", t.FilePath)
		wg.Done()
		return
	}
```
If the gallery ends up being a directory a warning is enough imho.

bnkai:
Looks and tests ok.

WithoutPants:
The `AssociateGallery` warning also occurs for zip files that have no images within. That should probably be addressed at a later stage.

PR #748, opened 2020-08-18
Config for stash-box instances

InfiniteTF:


WithoutPants:
This PR was requested by me as an extraction from #454, so that I could work on a server-side stash-box interface.

The configuration interface looks like this currently:

image

The places that I anticipate the stash-box integration to be used are as follows:
* the tagger in #454 
* the "Scrape Scene" dropdown button on the edit scene page
* the "Scrape Performer" dropdown button on the new/edit performer page

Given that we are allowing the configuration of multiple concurrent stash-box instances, we need to consider how to handle them in the UI. I see three main options, two of which will require a change in this PR:
* under "Scrape With" in scene and performer page, use the endpoint URL to the stash-box. This is ugly imo.
* add a "name" field to the stash-box configuration entry. This field would be optional for a single entry, but required if multiple endpoints are configured. This name will be used for the "Scrape With" dropdowns, and will use "stash-box" if none is entered.
* ensure that only one stash-box instance can be used at a time. A single stash-box instance could be configured as "active" at a time. Only one "stash-box" entry will be shown in the "Scrape With" dropdowns. I don't think this would be a very usable alternative.

My preference is for the second alternative.

I think the edit/save button should be removed in favour of always allowing the text fields to be edited. As it currently is, I think it infers to the user that clicking the save button would save the stash-box configuration, which isn't really the case.

InfiniteTF:
Implemented your name suggestion. For once instance you can omit the name, but more than one and it will only save with name.

ui/v2.5/src/components/Settings/SettingsConfigurationPanel.tsx
```diff
@@ -82,6 +86,21 @@ export const SettingsConfigurationPanel: React.FC = () => {
...
+            api_key: b?.api_key,
+            endpoint: b?.endpoint,
+          } as GQL.StashBoxInput)
+      )
+      .filter(
+        (b) =>
+          b.api_key &&
+          b.endpoint &&
+          (stashBoxes.length < 2 || (b?.name?.length ?? 0) > 0)
+      ),
```
WithoutPants, line 103: This validation should be done in the backend instead of or in addition to here. The server should not rely on the client to sanitise its inputs. If the configuration is invalid, the user should be notified as such.

With the current implementation, if a stash-box configuration is invalid it is removed altogether, rather than indicating to the user that it is invalid. Further, this also removed the first (valid) entry when the second entry was added and no names were entered.

WithoutPants (changes requested):

InfiniteTF:
Yeah, the behavior was pretty screwy. I've added serverside validation so it will reject bad stash-box inputs. This resolves the frontend issues as well since it only updates state when all instances are valid.

PR #751, opened 2020-08-19
Stash box client interface

WithoutPants:
This branch will likely be rebased to build on the #748 changes once they are approved. For now, this branch includes the config changes from that PR.

This change adds a graphql client to the stash server and exposes some interfaces to scrape from stash-box, using the existing `ScrapedScene` graphql types. It also adds stash-box entries - for each stash-box configured - to the "Scrape With..." edit scene button. It also adds UI code to add a query button for the stash-box query functionality, but this is currently commented out until a query dialog can be created, which depends on #454.

The stash-box client queries are in `graphql/stash-box` and can be generated with `make generate-stash-box-client`. It uses `gqlgenc` to generate the client go files. However, the fragment types in the generated `client.go` file do not currently include the `json` and `graphql` tags, so these currently have to be added manually.

Currently, the code grabs the first scene image, unlike #454 which as far as I can tell gets the largest image in landscape orientation. It does not currently get performer images. I think these should be returned as a list of URLs, allowing the user to choose an image as needed. This can be added at a later point, and will require a change to the graphql schema for `ScrapedScenePerformer`.

As it is, the query functionality can't be tested in the UI. Instead, the server-side can be tested in the graphql playground. An example query is as follows:
```
query {
  queryStashBoxScene(input: {
    stash_box_index: 0
    q: "Blowjob Fridays Jessie Rogers"
  }) {
    title
    details
    url
    date
    studio {
      id
      name
      url
    }
    tags {
      name
    }
    performers {
      name
      gender
      url
      twitter
      birthdate
      ethnicity
      country
      eye_color
      height
      measurements
      career_length
      tattoos
      piercings
      fake_tits
    }
  }
}
```


InfiniteTF:
@WithoutPants I've pushed a branch of my tagger pr which supports this backend. It should work if you apply it on top of yours. https://github.com/InfiniteTF/stash/tree/tagger-with-backend

I just used a selector to transform the data into roughly the existing shape, so it didn't require too much modification.

The missing pieces are scene id/duration/fingerprints, and performer images. The scene image solution  is fine, I can modify stash-box to do the sorting, like 99% of scenes only have one image anyway. For performer images I think it would be best to pass through the array of urls directly, fetching them all is going to be way too slow. The id/width/height can be omitted.

In terms of endpoints, a queryStashBoxByFingerprints which takes an array of scene ids, and queries stash-box with the relevant fingerprints, is kinda necessary. ditto submitStashBoxFingerprint. An endpoint that validates the credentials would also be useful, but that can wait.

Let me know if you want me to pitch in.

WithoutPants:
The issue I'm having with the batch query is that the singular query tries to search by the scenes MD5 checksum (if present) first, then tries the oshash if no results were found. I'm not really sure what approach to take for a batch query. I could just iterate the scenes and query the same way, but the stash-box API would still be hit once for each scene. Alternatively, I could just make it use the currently configured hash algorithm, but that seems like it would potentially miss results.

Do you have any thoughts on how you'd expect this to work?

InfiniteTF:
> The issue I'm having with the batch query is that the singular query tries to search by the scenes MD5 checksum (if present) first, then tries the oshash if no results were found. I'm not really sure what approach to take for a batch query. I could just iterate the scenes and query the same way, but the stash-box API would still be hit once for each scene. Alternatively, I could just make it use the currently configured hash algorithm, but that seems like it would potentially miss results.
> 
> Do you have any thoughts on how you'd expect this to work?

The idea  was to use the `findScenesByFingerprints` query which was added a little while back https://github.com/stashapp/stash-box/pull/23

It accepts an arbitrary amount of hashes and returns any scenes that match one of them, so we can grab the md5 and oshash of every requested scene and fire them to stash-box in one query. It only returns unique scenes so sending both md5 and oshash isn't an issue. Afterwards we just have to match the results to the local scenes based on one of the hashes.

WithoutPants:
Batch query change is done. Should be ready for review etc.

InfiniteTF:
The batch fingerprint query seems to work great from my limited testing. I haven't wired it up to the ui yet since it's getting increasingly difficult to merge the PRs, but I'll try to do that tonight.

As soon as stash_ids are added to the entities, and duration/fingerprints to scenes, I think we're good to merge this.

PR #754, opened 2020-08-20
Update to golang 1.13

WithoutPants:
Required for #751


bnkai:
Just a note that we need to test the resulting linux binaries if they run with Ubuntu 18 due to the libc change( production, develop docker images) or we need to upgrade to u Ubuntu 19 or 20.

bnkai:
In the compiler/build Dockerfile
should we also upgrade the node version? `RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -`

From the available versions https://github.com/nodesource/distributions 
 i tried 
`RUN curl -sL https://deb.nodesource.com/setup_lts.x | bash -` and it was ok

bnkai:
builts for all targets using the updated compiler image

https://gofile.io/d/mP84Do
```
0c9b7d89c7407c63036e2f8b5aac6e46  stash-linux
6c4d1a459f2ea45156bd55d03092ea37  stash-linux-arm32v7
afc9d1910723c29f5a85b0ea157118fc  stash-linux-arm64v8
2b08832bce5ed215f53fe19bd6f79570  stash-osx
887bf6507b731cd56e64ede001b9388e  stash-pi
aee3f4403ef548603cdc1160da9b4bfb  stash-win.exe

```
the osx binary should be tested 
the linux one is ok
 arm/pi ones  should be ok also ( if anyone can try them thats also good though)

WithoutPants:
We're going to need to be careful about the impact on the compiler and any hotfixes we introduce in the meantime. We need the master branch to continue to build using the existing compiler image. We'll probably need to change the master branch cross-compile script to get a specific version of the compiler image.

bnkai:
I also think that we need more tagged versions of the compiler
We could append  the golang version to the tag for the `stable` compiler images eg `stashapp/stash:compiler-1.11.5` and another `stashapp/stash:compiler-1.13.15` when the PR is merged. I guess this would need a "manual" push of the said compiler image?
The cross compile script should do the the pull of the image it needs instead of travis also.

Any thoughts of making a tagged version of the stable stash images ? It could be useful to have them available in the docker hub.

WithoutPants:
Opened #794 to address the compiler issue. I've stuck with numeric tags for the moment, since versions can change over time irrespective of the golang version.

I'll change the cross compile script to use the next version (4), and manually push the new version to docker hub.

PR #756, opened 2020-08-20
Fix infinite load issue

### LINKED ISSUE #747: [Bug Report] Hangs on Performer Page - Scene Tab with ONLY Random Sort Option
**Describe the bug**
When viewing a performer's scene page with random sorting, Stash hangs and CPU usage spikes. 

**To Reproduce**
1. Go to performer page, scene tab
2. Change to random sorting

**Expected behavior**
Random sort of scenes for performer. 

Version:	v0.2.1-66-gecf7451
Build hash:	ecf7451
Build time:	2020-08-17 02:14:05

Win 10, Chrome

**Additional context**
Debug logs only show a lot of entries similar to this: 
2020-08-17 19:45:56
Debug   
SQL: SELECT DISTINCT scenes.id FROM scenes 
        left join scene_markers on scene_markers.scene_id = scenes.id
        left join performers_scenes as performers_join on performers_join.scene_id = scenes.id
        left join movies_scenes as movies_join on movies_join.scene_id = scenes.id
        left join studios as studio on studio.id = scenes.studio_id
        left join galleries as gallery on gallery.scene_id = scenes.id
        left join scenes_tags as tags_join on tags_join.scene_id = scenes.id
     LEFT JOIN performers ON performers_join.performer_id = performers.id WHERE performers.id IN (?) GROUP BY scenes.id  HAVING count(distinct performers.id) IS 1 ORDER BY (substr(scenes.id * 0.9666082262992859, length(scenes.id) + 2)) DESC LIMIT 40 OFFSET 0 , args: [2374]

InfiniteTF:
Selecting random sort in a scene subcomponent causes the app to lock up indefinitely due to the random seed not being saved to the query string, which regenerates it infinitely.

Resolved by removing the subcomponent logic and saving all listfilter state to the query string. Added some tab routing logic to make the query string a bit cleaner when not on the scene tab.

Closes #747.

### DIFF
--- a/ui/v2.5/src/components/List/ListFilter.tsx
+++ b/ui/v2.5/src/components/List/ListFilter.tsx
@@ -29,7 +29,6 @@ interface IListFilterOperation {
 }
 
 interface IListFilterProps {
-  subComponent?: boolean;
   onFilterUpdate: (newFilter: ListFilterModel) => void;
   zoomIndex?: number;
   onChangeZoom?: (zoomIndex: number) => void;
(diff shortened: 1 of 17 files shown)


bnkai:
The random performer scenes sort seems to work ok with this fix.

PR #760, opened 2020-08-23
Fix various console errors and graphql loading issues

InfiniteTF:
* Adds element keys in a few places, as required by react. Specifically the manual, scene-card and rating stars components.
* Resolves graphql errors in tag/performer/studio components caused by graphql scene requests being fired before the entity is loaded, causing a bad request. Fixed by checking for presence before rendering scene panel.
* Changes a few onInput to onChange in the settings panel, and defaults values to 0 rather than undefined which results in a NaN value.
* Limit the width of the tag/performer scene card popovers to a sane maximum, otherwise it expands to the entire width of the screen which looks weird and is hard to use.
* Remove console logging of video pause errors. Not sure why that was happening at all since it's not really noteworthy.
* Moved the scene-card studio link outside of the thumbnail link since nesting anchors is illegal. Position is the same.
* Refactor the ListHook by moving all content rendering logic into its own component. Looks scary but the only real change is making sure the list content doesn't start rendering until localForage is initialized. Prior to the change panel content was frequently loaded twice, once with a blank filter, and then immediately afterwards with the saved filter. This wasted resources and caused the loading indicator to show at the same time as the panel content. Also removed the `getSelectedData` hook prop since `getData` already does the same thing.

All in all, besides the loading changes, very little should have changed.

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneCard.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneCard.tsx
@@ -289,26 +289,28 @@ export const SceneCard: React.FC<ISceneCardProps> = (
         }}
       />
 
-      {maybeRenderSceneStudioOverlay()}
-      <Link
-        to={`/scenes/${props.scene.id}`}
-        className="scene-card-link"
-        onClick={handleSceneClick}
-        onDragStart={handleDrag}
-        onDragOver={handleDragOver}
-        draggable={props.selecting}
-      >
-        {maybeRenderRatingBanner()}
-        {maybeRenderSceneSpecsOverlay()}
-        <video
-          loop
-          className={cx("scene-card-video", { portrait: isPortrait() })}
-          poster={props.scene.paths.screenshot || ""}
-          ref={hoverHandler.videoEl}
+      <div className="video-section">
+        <Link
+          to={`/scenes/${props.scene.id}`}
+          className="scene-card-link"
+          onClick={handleSceneClick}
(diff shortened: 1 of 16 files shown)


ui/v2.5/src/components/Scenes/SceneCard.tsx
```diff
@@ -294,7 +299,6 @@ export const SceneCard: React.FC<ISceneCardProps> = (
         draggable={props.selecting}
       >
         {maybeRenderRatingBanner()}
-        {maybeRenderSceneStudioOverlay()}
```
WithoutPants, line 297: Moving the studio overlay will make it impossible to overlay it in the bottom-right corner of the thumbnail (see the css snippet in the wiki). I think there is an outstanding request to swap the resolution/duration and studio positions. 

I think the better option for now would be to make either the scene thumbnail or the studio overlay into something other than a link.

I've also heard requests to move the studio link to the card rather than over the thumbnail, but I think that can be done as a normal element rather than an overlay, and we'd have some sort of option to switch showing whatever element is required.

InfiniteTF: Resolved it by wrapping the video section in a block so it behaves the same way it used to. I've confirmed that the wiki snippet works.

InfiniteTF:
Also fixed a background issue on the modals which was ruining the borders and corners.

WithoutPants (approved):

PR #770, opened 2020-08-29
Selective export

WithoutPants:
Part of #206 

Adds server support for exporting a subset of scenes, tags, performers, studios, and movies. Adds menu options in the scenes page to export selected or all scenes, with an option to include/exclude dependent tags/performers/studios/movies.

The export operation creates a zip file with the `mappings.json` file and all of the required json files using the existing directory structure. The graphql mutation returns a downloadable link that the UI then downloads. The download link and file are generated to be used once only, after which the zip file is removed from the server. 

Include a beginning of an architectural redesign: created a package for scene, movie, tag, performer and studio. These package are intended to house the business logic for the applicable business objects. I have also abstracted the query builder interfaces into `ReaderWriter` interfaces so that these can be mocked and the dependent packages unit tested.

Although there is server support for exporting other object types, I have deferred UI support for this for later. The other list controls need to be adjusted to support object selection.

This functionality is of limited utility until the import side is done, which will be in a separate PR.


bnkai:
The `use once functionality` doesn't seem to work when using a download manager extension in ff (https://addons.mozilla.org/en-US/firefox/addon/multithreaded-download-manager/).The file is deleted before the download starts. I wouldn't expect it too , just mentioning because i had it enabled when doing a full export. Disabling that and using firefox or using chrome everything seems fine.

An alternative approach (would need a lot more UI changes) would be to add  a manage exports page that lists all stored  exports ( zip files in meta/downloads/ folder) and lets the user download/delete the file manually.

WithoutPants:
Changed so that download hangs around for 30 seconds after the first request, so it should allow for multiple downloads in a short period of time.

bnkai:
With the latest commit it works with the download manager enabled.

PR #774, opened 2020-08-30
Add cbz to supported extensions as gallery

### LINKED ISSUE #768: [Feature] Add support for .cbz files
The gallery feature of the program supports zip files. .cbz files are zip files that are read by programs to act like comics, manga, etc. This is really handy for hentai storage. 

The program already deals with zip files really easy so it should just be expanding what file extensions the server scans for. It should be able to deal with a cbz file the exact same as a zip file.

You could rename all cbz files to zip but then it becomes a lot harder to view the files outside of the webapp.

Also worth nothing that the cb[x] naming for extensions applies to different archives as well. If you were to add support for tar (cbt), 7zip (cb7), or rar (cbr) then it might be worth adding the comic book version of those files as well.

bnkai:
should close #768

### DIFF
--- a/pkg/manager/manager_tasks.go
+++ b/pkg/manager/manager_tasks.go
@@ -14,8 +14,8 @@ import (
 	"github.com/stashapp/stash/pkg/utils"
 )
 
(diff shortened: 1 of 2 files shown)


bnkai:
Updated the changelog mainly to force travis but it still fails for `nvm install 12`

WithoutPants (approved):

PR #778, opened 2020-08-31
Fix age filtering regression

### LINKED ISSUE #777: [Bug Report] Age/birth year filtering causes SQL error
**Describe the bug**
Doing a performer query for age = x produces an SQL error.

eg: birth year = 2000 && age > 18 gives: `GraphQL error: Internal system error. Error <not enough args to execute query: want 4 got 2>`

These errors also replace the filter control, so that it is impossible to select a different query without altering the query URL.

Birth year is NULL also does not appear to work correctly.

**Expected behavior**
Age equality filtering should return the filtered results. Any graphql filter error should show in the results pane, and not hide the filter control.

WithoutPants:
Fixes #777 

Fixes regression introduced in #478 

Adds unit test to exercise the age criterion.

Fixes UI so that loading/error status does not find the filter control.

### DIFF
--- a/pkg/models/querybuilder_performer_test.go
+++ b/pkg/models/querybuilder_performer_test.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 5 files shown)


InfiniteTF (approved):
Works great.

Issue #734, opened 2020-08-13
[Feature] Mouse hover delay


ciaron88:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when quickly moving my mouse over the screen and all the videos instantly start showing the previews. Most of the time I am going to a video in the middle/end and this is the only one I want to see the preview and I want all the other videos to stay on the thumbnails.

**Describe the solution you'd like**
Have a customizable delay on how long you have to hover over a video before it starts showing the preview

peolic:
Another solution is restoring the scene cover image when the mouse isn't hovering over the video (currently disabled in code).

https://github.com/stashapp/stash/blob/e3ea3ea85e5f551ee36668d441c925e2c848b4c4/ui/v2.5/src/components/Scenes/SceneCard.tsx#L23-L25

https://github.com/stashapp/stash/blob/e3ea3ea85e5f551ee36668d441c925e2c848b4c4/ui/v2.5/src/hooks/VideoHover.ts#L45-L49

HereIam80:
According to #576 there is no easy fix?! I would love having a delay or the thumbnail restored.

WithoutPants:
I think this should be implemented alongside #576 .

PR #787, opened 2020-09-05
Scene card preview refactor

### LINKED ISSUE #576: [Feature] Show thumbnail instead of paused preview when moving mouse away from scene
Having animated scene previews is a really cool feature, but can be frustrating when trying to browse to figure out what to watch. 

It gets frustrating when every time you hover your mouse anywhere in the scene window it starts the preview then when removed the preview stops but doesn't revert to showing the assigned thumbnail requiring you to reload the page in order to show it again.

This becomes an issue when a lot of scenes have black fade in's or other intros and your entire scene page is turned to a sea of black in just a few short swipes across the screen when you are trying to organise stuff or just browse what to watch.

When moving mouse off of the scene preview, revert to showing the assigned thumbnail while still retaining the part of the video you stopped the preview on.
This way your page will always look organised and clean but you still retain the ability to resume the playback of the preview

Can't really figure out what benefits the current system has over the suggested one, scene previews are so short and go by fast so its not like you are gonna reorganise the look of your entire scene page temporarily by picking your favourite part of the preview and avoiding moving the mouse to anywhere other then the edges of the screen.
But if the current system is a version some find perfect for them, maybe include a button in settings to switch between showing preview only when hovering and showing preview constantly.

InfiniteTF:
Changes scene cards to use separate image and video elements, and switch between them with css so that the image is restored whenever the card is no longer hovered.

I've also added a 200ms delay before the video preview is shown. I haven't made it user configurable since it would add a lot of boilerplate for little gain.

Resolves #576 and #734


ui/v2.5/src/hooks/VideoHover.ts
```diff
@@ -1,86 +0,0 @@
...
-  }, [videoEl]);
-
-  useEffect(() => {
-    const videoTag = videoEl.current;
-    if (!videoTag) {
-      return;
-    }
-    videoTag.volume = soundEnabled ? 0.05 : 0;
-  }, [soundEnabled]);
-
```
WithoutPants, line 77: This needs to be integrated into the replacement. Currently hovering over thumbnails outputs sound regardless of the `soundEnabled` flag.

WithoutPants (changes requested):

ui/v2.5/src/hooks/VideoHover.ts
InfiniteTF: Completely forgot about that. Fixed now.

WithoutPants (approved):

PR #790, opened 2020-09-08
Added streaming quality options

JoeSmithStarkers:
Added selectable streaming quality profiles from the ScenePlayer.

This is work in progress, this is the minimal set of changes required.

I'm looking for comments or advice.

Further issues/questions are:
1. Check what happens when a stash wide quality profile is selected.
2. Fix bug that occurs when switching back to direct streaming after being in a quality profile (the progress bar goes haywire, any advice would be appreciated)


InfiniteTF:
To fix the seek issue you need to reset the `start` parameter on the video element if it's a direct stream:

```
if (ScenePlayerImpl.isDirectStream(_videoTag.src) && _videoTag.dataset.start) {
  _videoTag.dataset.start = "0";
}

Seems to work really well btw, great stuff! I'd completely forgotten that jwplayer supports quality switching natively.

**Comment by bnkai** [comment]
It looks good.
The use of the `mp4` endpoint though doesnt work well with firefox. If you try to seek over a video the video will start to play but after some time it will stop. Only a chunk of it is played. Chromium based browsers support some extra stuff (mkv support also) that firefox doesnt.Thats why we used the `webm` endpoint as a safer choice in the first place before the last stream refactoring.
Can you add the `mp4` to your existing labels before the resolution and add the extra 5 `webm` ones pointing to the appropriate endpoint ?

I'm not sure about safari but i think it might needed the `hls` ones so maybe more labels/streams for that too ?

**Comment by JoeSmithStarkers** [comment]
Hi @bnkai 

Fun has been had but finally I’ve found the issue with firefox and mp4 streaming.

-movflags frag_keyframe+empty_moov

Without empty_moov Firefox has the start/stop playback issue as described by @bnkai, with empty_moov it works as expected. I've also check chrome and no regression. I plan to do some more extensive testing.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/Media_Source_Extensions_API/Transcoding_assets_for_MSE

The reference also suggest using “default_base_moof” to make it work with chrome, but I don’t see any regression with Chrome, so I’m not planing to add it.

I’ve also found this issue where Firefox is always defaulting to the last stream type that was loaded, eg I see the request for a different quality stream instead of the default direct stream option, but the player seems to think it’s playing the default direct stream.

On the subject of providing webm/mp4 quality options: JWPlayer selects the first of the provided mime types that the current browser supports and then only provides different quality streams for that mime type, thus it would have to be an external setting to allow selecting between webm/mp4.

@InfiniteTF I'm not sure where I should reset the start parameter, do i need to create a new hook for it? eg when the stream changes?

**Comment by InfiniteTF** [comment]
@JoeSmithStarkers Put it at the start of the `seekHook`. There's an event for `onLevelChanged` which would be appropriate, but you need access to the data attribute of the video element, and you only get that in the hooks.

Strictly speaking I don't think there's a need to use the data attribute, a `useState` would accomplish the same thing, but the ScenePlayer isn't a functional component so I'm not sure that is possible without a rewrite.

**Comment by JoeSmithStarkers** [comment]
@InfiniteTF thank-you, i've put in the fix for the issue with switching back to the direct stream.

**Comment by bnkai** [comment]
With the latest commit if you play a scene and choose one of the transcode streams all seems ok. When you open a new scene though after that and try to play while the direct stream is shown as selected it still uses the previous transcode stream resolution setting and transcodes the scene instead.

**Comment by JoeSmithStarkers** [comment]
Update
1. I've fixed that jwplayer was attempting to remember the quality of the stream.
2. I've added webm format quality transcodes with the same mime type as mp4, so that a user can select between the two transcoding options.

File: ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx [hunk 1]
```diff
@@ -211,28 +206,31 @@ export class ScenePlayerImpl extends React.Component<
...
         return false;
       }
 
-      // remove the start parameter
-      let { src } = _videoTag;
+      if (ScenePlayerImpl.isDirectStream(_videoTag.src)) {
+        if (_videoTag.dataset.start) {
+          /* eslint-disable no-param-reassign */
+          _videoTag.dataset.start = "0";
+          /* eslint-enable no-param-reassign */
```
InfiniteTF, line 217: You  can use `// eslint-disable-next-line no-param-reassign` rather than disabling and enabling.

JoeSmithStarkers: Fixed. Sorry i'm pretty new to modern frontend dev. My work world is full of hacky jquery.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -272,6 +270,9 @@ export class ScenePlayerImpl extends React.Component<
   }
 
   public render() {
+    // Default back to Direct Streaming
+    localStorage.removeItem("jwplayer.qualityLabel");
+
```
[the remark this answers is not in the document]
JoeSmithStarkers: Resolved.

JoeSmithStarkers:
Made two minor adjustments.
1. Move the non direct default streams to after the transcoded streams. Fixes issue when transcoding is required, so the direct stream isn't provided by the backend. This in turn means jwplayer defaults to the webm stream with the webm mime type, which prevent selection of other transcoded qualities.
2. Reorder the resolutions: biggest to smallest. seems to be the standard.

WithoutPants:
This is breaking the failover code for me. Tested on an hevc video. Loads briefly then stops. Clicking the play button just flickers the player.

JoeSmithStarkers:
> This is breaking the failover code for me. Tested on an hevc video. Loads briefly then stops. Clicking the play button just flickers the player.

I believe the fallover code was broken by #716 which fixes #714.

In this PR I've added `this.player.play();` back to the fallback code, as i believe it is needed.

I can't reproduce the original issue #714.

WithoutPants:
> > This is breaking the failover code for me. Tested on an hevc video. Loads briefly then stops. Clicking the play button just flickers the player.
> 
> I believe the fallover code was broken by #716 which fixes #714.
> 
> In this PR I've added `this.player.play();` back to the fallback code, as i believe it is needed.
> 
> I can't reproduce the original issue #714.

This does regress to cause #714 for me. Tested against `jellyfish-5-mbps-hd-hevc.audio.null.mp4` which @bnkai made for me from memory.

bnkai:
The jellyfish sample was used to test a case where jwplayer was returning an error for videos without audio part i think.
@WithoutPants was it the specific file you mentioned or the original `jellyfish-5-mbps-hd-hevc.mkv` that shows the regression ?
I have them uploaded here https://easyupload.io/m/7c6mzy

JoeSmithStarkers:
Thank you for the sample file: @bnkai 

Ether of the null audio files cause the regression for me. It seems those null audio files create a "meta" event from jwplayer, that has zero width/height reported. Thus running the below code:

image

image

The odd thing is a mkv file with valid sound doesn't create a media event on load, but a mp4 does (just with a non zero height/width).

@WithoutPants do you remember why that handler was added?

WithoutPants:
See [here](https://github.com/stashapp/stash/pull/609#issuecomment-656781683) - if the audio codec is supported but the video codec is not, then it would play the audio without the video. The handler is there to check if the video is not present and handle it the same way as an unsupported file.

JoeSmithStarkers:
> See [here](https://github.com/stashapp/stash/pull/609#issuecomment-656781683) - if the audio codec is supported but the video codec is not, then it would play the audio without the video. The handler is there to check if the video is not present and handle it the same way as an unsupported file.

To fix this issue, i can see two ways to do it.
1. Shadow the playing state of the jwplayer.
2. Handle the meta event differently to the specific video format error. (I've opted for this one because i think it's cleaner and less problematic).

WithoutPants:
I haven't tested yet, but can you put some code comments describing why you've added that change. Better to have it in the code than here. I probably should have documented the meta event change that way as well.

WithoutPants:
Testing with the same hevc files: it's now transcoding to mp4 by default instead of webm. For hevc files it seems quite slow to transcode in comparison to webm. I think it should default to webm (in Chrome at least) as per previous behaviour.

JoeSmithStarkers:
> I haven't tested yet, but can you put some code comments describing why you've added that change. Better to have it in the code than here. I probably should have documented the meta event change that way as well.

Comments added.

>Testing with the same hevc files: it's now transcoding to mp4 by default instead of webm. For hevc files it seems quite slow to transcode in comparison to webm. I think it should default to webm (in Chrome at least) as per previous behaviour.

I've swapped the order of the webm/mp4 streams, so it defaults to webm.

I also fixed an issue with Safari OSX/iOS that prevented it from playing hls/m3u8 streaming.

I've tested chrome/firefox/opera/safari osx/safari ios.

The only issue that i've found is, for a mkv with null audio, chrome requires the play button to be pressed twice. Without resorting to browser detection: I can't determine if it's the m3u8 case which breaks safari, or the null audio case which requires two play presses. So I opted to support safari over the minor inconvenience with chrome.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -98,18 +96,25 @@ export class ScenePlayerImpl extends React.Component<
...
         !metadata.height
       ) {
-        // treat this as a decoding error and try the next source
-        this.handleError();
+        // Occurs during preload when videos with supported audio/unsupported video are preloaded.
+        // Treat this as a decoding error and try the next source without playing.
+        // However on Safari we get an media event when m3u8 is loaded which needs to be ignored.
+        const currentFile = this.player.getPlaylistItem().file;
+        if (currentFile != null && !currentFile.includes("m3u8")) {
+          this.handleError(false);
```
WithoutPants, line 116:
```suggestion
          const state = this.player.getState();
          const play = state === "buffering" || state === "playing";
          this.handleError(play);
```

This fixes the issue of needing to click the play button multiple times.

WithoutPants (changes requested):
I didn't realise that the ios safari side still had some latent bugs. Still not super smooth with my testing, but it seems improved over the current behaviour.

Otherwise, testing looks pretty good. Just have a single change for that multi-click mkv issue.

PR #793, opened 2020-09-08
Update gallery view layout and switch libraries

InfiniteTF:
The gallery library we've used is old and doesn't really work right since it needs width/height of images to lay them out properly. Without dimensions it just looks bad. I've swapped the gallery for flexbin which is just a small scss snippet which does proper image block layout on the fly, which is presumably how it worked at some point.

I also switched out the lightbox library since it was old and deprecated at this point. fslightbox is functionally more or less the same as the old one, but a bit easier to use and with better mobile support.


WithoutPants:
Tests ok and looks good. Can you resolve the conflicts?

InfiniteTF:
> Tests ok and looks good. Can you resolve the conflicts?

Done

PR #808, opened 2020-09-11
Update makefile for static build

bnkai:
This link https://www.arp242.net/static-go.html explains better why the linux stash build till now was not "static"
With these changes the linux build is static
Now ( this PR )
```
ldd ~/stash-linux 
	not a dynamic executable
```
Before
```
ldd ~/stash-linux
	linux-vdso.so.1 (0x00007fffc6be4000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f906bee7000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f906bec5000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f906bd00000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f906bf00000)
```
Hopefully this doesn't effect the windows and osx buid. (travis log should verify that)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,12 +41,15 @@ endif
 
 build: pre-build
 	$(eval LDFLAGS := $(LDFLAGS) -X 'github.com/stashapp/stash/pkg/api.version=$(STASH_VERSION)' -X 'github.com/stashapp/stash/pkg/api.buildstamp=$(BUILD_DATE)' -X 'github.com/stashapp/stash/pkg/api.githash=$(GITHASH)')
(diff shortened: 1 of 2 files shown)


bnkai:
Seems it is messing with the osx build , need to update

bnkai:
Left the windows , osx build as it was and used the extra static target for the linux builds ( arm + amd64 )

WithoutPants:
We're going to want to update the build dockerfile as well.

WithoutPants (approved):

PR #814, opened 2020-09-17
Added ffmpeg/ffprobe permission correction code for linux/osx

JoeSmithStarkers:
Permission correcting code for downloaded ffmpeg/ffprobe.


bnkai:
While we are doing this should we also look for alternatives to zeranoes static builds ?
According to <https://ffmpeg.zeranoe.com/builds/> by tomorrow the download links won't work.

WithoutPants (approved):
Tested on Linux using some local code modifications.

PR #820, opened 2020-09-23
Parallel scanning/generation, and combined scanning/preview/sprite

### LINKED ISSUE #323: [Feature] Improve Scanning Button to include other functions (Auto Tag, Generate)
When clicking "Scan" allow it to do multiple actions at once if needed.  Make the other options toggle-able so that if a user does not want to generate sprites or kick off the auto-tag feature they don't have to.    

Alternatively another solution might be when clicking multiple options (Scan, Auto-Tag, Generate) queue them up so that eventually each will finish.

### LINKED ISSUE #818: [Feature] Allowing Parallelism for Large Task Queue
If the user's library contains a large number of videos or files to process, tasks such as content generation (i.e., image previews, markers, transcodes, etc.) can take abysmal amount of time to finish. For example, assuming there are several thousands of videos, and the user has requested image previews generation. ffmpeg is spawned synchronously, one at a time, to generate image previews. This can take *ages* to finish. 

Allowing users to specify maximum number of processes for parallelism, depending on the task.

**Additional context**
Potential related issues - #323

JoeSmithStarkers:
Resolves #323
Resolves #818

This PR implemented parallel scanning/generation tasks, and also configurable preview/sprite generation during scan.

The rational behind both these features is for users who use cloud drives to store files. 
1. For parallel scanning/generation, there is a large amount of latency caused by waiting for files to download.
2. For combined scanning/preview/sprite generation, reduces the number of the times the file needs to be fetched from the remote because it's cached temporally.)  

(A minor benefit maybe seen for users with local data, when running 2 generation tasks, reducing the effect of multiple invocations of ffmpeg, startup latency)

At the moment it's hardcoded to one task for both scanning/generation, and also to disable combined scanning/preview/sprite generation. (To enable both, check manager_tasks.go for "CHANGEME" comments)

There a few things left to finish up:
1. Add configuration/ui for number of tasks for both scan/generate. (need help)
2. Add ui options to enable preview/sprite generation during scan. (need help)
3. Check a GenerateMarkersTask for thread safety.
4. Check for bugs when prematurely ending scans/generation


pkg/manager/manager_tasks.go
```diff
@@ -112,7 +114,15 @@ func (s *singleton) Scan(useFileMetadata bool) {
...
 		logger.Infof("Starting scan of %d files. %d New files found", total, s.neededScan(results))
 
-		var wg sync.WaitGroup
+		start := time.Now()
+		parallelTasks := config.GetParallelTasks()
+		if parallelTasks <= 0 {
+			parallelTasks = runtime.NumCPU()
+		}
+
+		logger.Info("Scan started with %d parallel tasks", parallelTasks)
```
bnkai, line 123: `logger.Infof("Scan started with %d parallel tasks", parallelTasks)`

pkg/manager/manager_tasks.go
```diff
@@ -265,8 +279,13 @@ func (s *singleton) Generate(input models.GenerateMetadataInput) {
...
 		}
 
-		delta := utils.Btoi(input.Sprites) + utils.Btoi(input.Previews) + utils.Btoi(input.Markers) + utils.Btoi(input.Transcodes)
-		var wg sync.WaitGroup
+		parallelTasks := config.GetParallelTasks()
+		if parallelTasks <= 0 {
+			parallelTasks = runtime.NumCPU()
+		}
+
+		logger.Info("Generate started with %d parallel tasks", parallelTasks)
```
bnkai, line 287: `logger.Infof("Generate started with %d parallel tasks", parallelTasks)`

pkg/manager/manager_tasks.go
```diff
@@ -112,7 +114,15 @@ func (s *singleton) Scan(useFileMetadata bool) {
 		total := len(results)
 		logger.Infof("Starting scan of %d files. %d New files found", total, s.neededScan(results))
 
-		var wg sync.WaitGroup
+		start := time.Now()
+		parallelTasks := config.GetParallelTasks()
+		if parallelTasks <= 0 {
+			parallelTasks = runtime.NumCPU()
```
bnkai, line 120: since ffmpeg uses multiple threads by default when encoding video ( we limit it to 4 threads for the marker,chunk creation) is it wise to use that many tasks ? 
also the default configuration is `1` shouldn't we use that also here ?

pkg/manager/manager_tasks.go
```diff
@@ -265,8 +279,13 @@ func (s *singleton) Generate(input models.GenerateMetadataInput) {
 			return
 		}
 
-		delta := utils.Btoi(input.Sprites) + utils.Btoi(input.Previews) + utils.Btoi(input.Markers) + utils.Btoi(input.Transcodes)
-		var wg sync.WaitGroup
+		parallelTasks := config.GetParallelTasks()
+		if parallelTasks <= 0 {
+			parallelTasks = runtime.NumCPU()
```
bnkai, line 284: same suggestion as above , maybe use the default instead ?

pkg/manager/manager_tasks.go
JoeSmithStarkers: While i haven't an issue running that many tasks, i agree for most if won't provide any benefit.

How about 
`parallelTasks` = (runtime.NumCPU() / 4) + 1`

pkg/manager/manager_tasks.go
JoeSmithStarkers: Thanks, will fix.

pkg/manager/manager_tasks.go
JoeSmithStarkers: Thanks, will fix.

pkg/manager/manager_tasks.go
bnkai: I tried it with 4 parallel tasks and the scan process was using ~70% of my server cpu ( dual xeon total 12c/24t). I usually avoid to utilise it that much cpu due to power / noise considerations :-)
Anyway it shouldn't  matter that much ( /4 +1 sounds reasonable to me) since the default is 1 and the user can choose what he wants. Maybe add some more info in the manual  ( https://github.com/stashapp/stash/tree/develop/ui/v2.5/src/docs/en )? Task or Configuration section should be ok.

If you perform tasks remotely (gdrive) insn't it a tradeoff between latency/bw causing cpu underutilisation ( low or 1 parallel task ) and invalidating the cache with many concurrent downloads ( ++ parallel tasks , not sure how rclone handles that though ) ?

pkg/manager/manager_tasks.go
JoeSmithStarkers: Resolved as above.

pkg/manager/manager_tasks.go
JoeSmithStarkers: Resolved as agreed.

Rclone, has three caching modes. 
1. In memory cache. -- Not really caching, but it does do read ahead. 
2. Block based disk cache. -- Good for streaming but can't handle the load of scanning/generation. 
3. File based disk cache. -- Good for scanning but too much latency for browsing/streaming.

I run a large file based cache for scanning (240gb), so it can fit many videos all being scanned/generated concurrently. And i use block based for browsing/streaming.

WithoutPants:
I'll defer this until the images PR is merged since it impacts the scanning code significantly.

com1234475:
When moving a file, the logic to update the files entry seems to be broken.

I get `ERRO[126468] UNIQUE constraint failed: scenes.oshash `

JoeSmithStarkers:
> When moving a file, the logic to update the files entry seems to be broken.
> 
> I get `ERRO[126468] UNIQUE constraint failed: scenes.oshash `

Hi @com1234475 

I tried to reproduced that error by renaming a file but it didn't error.

What I tried:
1. Change scene hash type to ohash.
2. Rename a file.
3. Run scan.

What were the steps you used? Can you include any logs that were produced before the error?

JoeSmithStarkers:
Resolved the conflicts generated by the image pr.

pkg/api/resolver_query_configuration.go
pkg/manager/manager_tasks.go
pkg/manager/task_generate_gallery_thumbs.go
pkg/manager/task_scan.go

com1234475:
Hey @JoeSmithStarkers

My files already have `oshash`, if I rename a file and run the scan i get the following.

```
INFO[175479] /home/***.mp4 not found.  Calculating oshash... 
INFO[175480] /home/***.mp4 doesn't exist.  Creating new item... 
ERRO[175480] UNIQUE constraint failed: scenes.oshash    
```

I had changed:

```
wg := sizedwaitgroup.New(4)
task := ScanTask{FilePath: path, UseFileMetadata: useFileMetadata, fileNamingAlgorithm: fileNamingAlgo, calculateMD5: calculateMD5, GeneratePreview: true, GenerateSprite: true}
```

JoeSmithStarkers:
@com1234475 thank-you for finding that issue. After code review I found the issue, was able to reproduce the bug and have rectified it.

Please could you re-test.

bnkai:
There is actually a race condition when using the scan with more than one threads but needs a lot of conditions to happen so we shoouldn't mind that much.
During scan if you use more than 1 threads (and untick the extra generate stuff),say 4 in example for this case files that are considered as `new` are processed in parallel. If you have say 2 or 3 duplicate files and they happen to be the only new files or close in path ( eg scene.1.mp4 scene.2.mp4 scene.3.mp4 ) to be `lscanned` in parallel then they won't be considered as duplicate because the hash is not yet present in the db. One of them will be added to the DB and the rest will fail with `UNIQUE constraint failed: scenes.oshash` as during the commit stage the first file (could be any of the 3) will be added without problem while the rest will fail due to the first. This isn't a major problem because during the next scan the rest of the files will be correctly identified as duplicates , i mention it here so we know why it happens. We could probably check the error code after the commit and add an extra print but it's not a priority.

Steps to reproduce.

- Click scan ( un tick generate previews,sprites generation)
  Make sure that the scan finishes ok so next time we scan only the sample files will be found as new
- Add a new `sample` video file to your stash media directory.
- Make a couple of copies of it so now you have sample.mp4 sample.1.mp4 sample.2.mp4
- Set number of parallel tasks to 4. Save configuration
- Click scan ( untick generate previews,sprites generation)

you 'll end up with
```
INFO[0801] Scan started with 4 parallel tasks           
INFO[0801] sample.mp4 not found. Calculating oshash... 
INFO[0801] sample.1.mp4 not found. Calculating oshash... 
INFO[0801] sample.2.mp4 not found. Calculating oshash... 
INFO[0802] sample.1.mp4 doesn't exist. Creating new item... 
INFO[0802] sample.mp4 doesn't exist. Creating new item... 
INFO[0802] sample2.mp4 doesn't exist. Creating new item... 
ERRO[0802] UNIQUE constraint failed: scenes.oshash      
ERRO[0802] UNIQUE constraint failed: scenes.oshash  
```

Also purely for cosmetic reasons could you  edit the imports ?
eg
```golang
	"fmt"
	"github.com/remeh/sizedwaitgroup"
	"github.com/stashapp/stash/pkg/logger"
	"github.com/stashapp/stash/pkg/models"
	"github.com/stashapp/stash/pkg/utils"
	"github.com/stashapp/stash/pkg/manager/config"
	"sync"
	"os"
	"runtime"
	"strconv"
	"time"
```
We keep a space between `internal` `external` and `stash` libraries so that the formatter orders them by  name and keeps them tidy
Editing the above like this 
```golang
	"fmt"
	"sync"
	"os"
	"runtime"
	"strconv"
	"time

	"github.com/remeh/sizedwaitgroup"

	"github.com/stashapp/stash/pkg/logger"
	"github.com/stashapp/stash/pkg/models"
	"github.com/stashapp/stash/pkg/utils"
        "github.com/stashapp/stash/pkg/manager/config"
```
the formatter will fix the order (order by name) for every separate group

JoeSmithStarkers:
Import ordering fixed.

I could add better log message/warning for that rare race condition. I don't believe there is much I can do about it. As if you are generating hashs for multiple files in parallel: there is always going to be the possibility that multiple files with the same hash are in progress of being added to the database. Luckily database constraints prevent it from being committed.

WithoutPants (changes requested):

pkg/manager/task_scan.go
```diff
@@ -427,9 +476,10 @@ func (t *ScanTask) scanZipImages(zipGallery *models.Gallery) {
 		subTask.zipGallery = zipGallery
 
 		// run the subtask and wait for it to complete
-		var wg sync.WaitGroup
-		wg.Add(1)
-		subTask.Start(&wg)
+		// tylerj check this.
```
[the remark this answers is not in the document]
JoeSmithStarkers: Checked

pkg/manager/manager_tasks.go
```diff
@@ -432,13 +453,17 @@ func (s *singleton) Generate(input models.GenerateMetadataInput) {
 				continue
 			}
 
-			wg.Add(1)
+			wg.Add()
 			task := GenerateMarkersTask{Marker: marker, Overwrite: overwrite, fileNamingAlgorithm: fileNamingAlgo}
 			go task.Start(&wg)
-			wg.Wait()
+			wg.Wait() // TODO check the safety of this operation in parallel with itself
```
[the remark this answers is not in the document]
JoeSmithStarkers: Done

pkg/manager/manager_tasks.go
```diff
@@ -151,7 +153,15 @@ func (s *singleton) Scan(useFileMetadata bool) {
 			logger.Infof("Starting scan of %d files. %d New files found", *total, *newFiles)
 		}
 
-		var wg sync.WaitGroup
+		start := time.Now()
+		parallelTasks := config.GetParallelTasks()
+		if parallelTasks <= 0 {
+			parallelTasks = (runtime.NumCPU() / 4) + 1
+		}
+
```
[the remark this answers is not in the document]
JoeSmithStarkers: Moved repeated code to new function.

WithoutPants (approved):

PR #824, opened 2020-10-01
Fix ffmpeg/ffprobe downloads

bnkai:
Updated download code to get ffmpeg binaries from different sources since zeranoe's builds are no more available.
It now fetches the 4.3.1 versions from gyan for windows and evermeet for macos.
Tested by disabling the ffmpeg detection and forcing the GOOS switch.
An actual test of the windows/macos build will be needed.


WithoutPants (approved):
Tested new behaviour on Linux with some local code changes.

PR #825, opened 2020-10-02
Example python plugin

WithoutPants:
Adds an example python plugin that works similarly to the raw golang plugin. For illustrative purposes.

Fixes a minor bug where stderr output that was not prefixed would incorrectly be interpreted as a progress level message.

### DIFF
--- a/pkg/plugin/examples/python/pyraw.yml
+++ b/pkg/plugin/examples/python/pyraw.yml
@@ -1,0 +1,29 @@
+# example plugin config
+name: Hawwwwt Tagger (Raw Python edition)
+description: Python Hawwwwt tagging utility (using raw interface).
+version: 1.0
+url: http://www.github.com/stashapp/stash
+exec:
+  - python
+  - {pluginDir}/pyplugin.py
+interface: raw
+tasks:
+  - name: Add hawwwwt tag to random scene
+    description: Creates a "Hawwwwt" tag if not present and adds to a random scene.
+    defaultArgs:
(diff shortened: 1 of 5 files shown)


pkg/plugin/examples/python/pyraw.yml
```diff
@@ -0,0 +1,29 @@
+# example plugin config
+name: Hawwwwt Tagger (Raw Python edition)
+description: Python Hawwwwt tagging utility (using raw interface).
+version: 1.0
+url: http://www.github.com/stashapp/stash
+exec:
+  - python
+  - {pluginDir}/pyplugin.py
```
bnkai, line 8: i had to change that to load the yml without errors
`  - "{pluginDir}/pyplugin.py`

pkg/plugin/examples/python/pyraw.yml
```diff
@@ -0,0 +1,29 @@
+# example plugin config
+name: Hawwwwt Tagger (Raw Python edition)
+description: Python Hawwwwt tagging utility (using raw interface).
+version: 1.0
+url: http://www.github.com/stashapp/stash
+exec:
+  - python
```
bnkai, line 7: users that have python2 also present need to edit `python` to `python3` for the delete tag to work

WithoutPants: Since it's just a proof of concept example, I'll leave it as is.

bnkai (approved):

PR #826, opened 2020-10-02
Video filters and transforms

JoeSmithStarkers:
WIP in progress for review.

I've added css/svg based video filters to the scene player.

The video filters don't work across all browsers. Chrome is the closest to being fully functional, however an issue occurs when mixing an svg filter with a scale/rotate transformation. Firefox on the other hand is laggy with most filters and a decent sized video.

videoFilters


WithoutPants:
While I can't see using this myself, this does cool, and works well based on my cursory tests on Chrome, Firefox and Edge. Haven't yet tried on the iPad. I do think that the Filters tab should probably not be the last tab, since that's the customary location for the Edit tab. The addition of this tab also overflows the tab container when the scene has Movies or a Gallery.

JoeSmithStarkers:
I've re-ordered the panels, and added an aspect ratio correction slider.

Still have to consider the issue of the overflow.

JoeSmithStarkers:
Any suggestions on handling the overflow?

WithoutPants:
```
            <ButtonGroup className="ml-auto">
              <Nav.Item>
                <OCounterButton
                  loading={oLoading}
                  value={scene.o_counter || 0}
                  onIncrement={onIncrementClick}
                  onDecrement={onDecrementClick}
                  onReset={onResetClick}
                />
              </Nav.Item>
              <Nav.Item>{renderOperations()}</Nav.Item>
            </ButtonGroup>
```
This will make it so that the overflow menu and the O-Counter button break together to the next line if needed. That's good enough for me.

ui/v2.5/src/components/Scenes/SceneDetails/SceneVideoFilterPanel.tsx
```diff
@@ -0,0 +1,851 @@
...
+      {renderResetButton()}
+      <div className="row form-group">
+        <span className="col-12">
+          <i>
+            Note: Filters is a beta level feature and all functions may not work
+            in all browsers equally. Chrome seems best but isn&apos;t without
+            issues.
+          </i>
+        </span>
+      </div>
```
WithoutPants, line 847: I don't think this is necessary.

WithoutPants (changes requested):

WithoutPants (approved):

PR #833, opened 2020-10-08
Add scraper list page

WithoutPants:
Adds a Scrapers page to the Settings list. 

image
image

Shows each scraper, the scraping types that it supports, and the URL fragments that it matches. List is divided into scene, performer and movie scrapers. Also has a reload scrapers button.


bnkai:
Haven't tested yet.
Looks great and was needed , we get a  lot of questions `why don't the scene scrapers work` in the channel.
A couple of thoughts.
- Can we get some kind of expandable fields  for the scene urls ?  The full list of urls for all the scrapers is a lot of pages. 

- The links for the URLs on 90% of the cases will be a 404. eg
`<a href="https://passion-hd.com/video/" class="link" target="_blank" rel="noopener noreferrer">https://passion-hd.com/video/</a>`
While it is very useful for the user to see how he should edit the `URL` field to get a match reading the anchor text ,  the `href` could be changed to only the base url
something like
`<a href="https://passion-hd.com" class="link" target="_blank" rel="noopener noreferrer">https://passion-hd.com`/video/</a>

WithoutPants:
I've made URL lists over 5 items long collapsible and changed the hrefs. Long URL lists now have a `more/less` button to expand/collapse the lists.

bnkai (approved):

PR #836, opened 2020-10-09
Prevent invalid date tag in video file from aborting scan

### LINKED ISSUE #731: [Bug Report] Some scanned files give ParseDateStringAsTime Error but no other information
During scanning I get the following error on 13 files:
ParseDateStringAsTime failed: dateString <>

There is no other information given so I cannot find the files giving the error. 

Can you make the error give the file path so I can check the files or remove them?

Stash Version: v0.2.1-53-g634b892

WithoutPants:
Fixes #731 

Some files have a malformed date tag in their metadata. This would cause the ffprobe unmarshal to fail. I have changed `JSONTime` to log a warning instead of returning an error when unmarshalling a malformed date. In addition, I've improved the logging to provide the filename for ffprobe errors, and tidied some of the scan logging to be cleaner and less verbose.

### DIFF
--- a/pkg/ffmpeg/ffprobe.go
+++ b/pkg/ffmpeg/ffprobe.go
@@ -236,7 +236,7 @@ func NewVideoFile(ffprobePath string, videoPath string) (*VideoFile, error) {
 
 	probeJSON := &FFProbeJSON{}
 	if err := json.Unmarshal(out, probeJSON); err != nil {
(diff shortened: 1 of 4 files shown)


bnkai:
Code looks ok and it seems to run ok ( I don't have a sample file to trigger the malformed date issue )

Issue #729, opened 2020-08-12
[Feature] Add Organized flag to scenes


WithoutPants:
**Is your feature request related to a problem? Please describe.**
Users have expressed a desire to exclude scenes from the auto-tagger. It would also be useful information to query on to be able to exclude scenes that are complete from a metadata point of view.

**Describe the solution you'd like**
Add an "Organized" flag to scenes. Scenes with this flag set to true should be excluded from the auto-tagger and should be excluded by default from the filename parser as well. It should be possible query scenes based on the value of this flag.

bnkai:
I think it  was mentioned again in the discord channel and piere-delecto's  script already uses this approach, can't we just use some system reserverd tags ? For the specific case if a scene has a say `stash_metadata_locked` tag applied then no automatic action regarding mutation of metadata should occur ( auto-tagger, list from filename parser ).
If a flag is to be used there should be a filter that makes it easy for the user to find the scenes he flagged.

WithoutPants:
This approach would work if we really wanted to avoid a schema modification, but in my opinion is a misuse/conflation of the tag concept. A flag is a cleaner approach, has a clearer purpose and doesn't pollute the user's tag collection with a system-only entry.

PR #839, opened 2020-10-10
Add organized boolean to scene model (#729)

com1234475:
Add DB migration and graphQL entries.


WithoutPants:
Thanks for the contribution. I have developed this further in #988.

PR #844, opened 2020-10-12
Add logging of duplicates and scan errors

### LINKED ISSUE #837: [Feature] Log potential issues with files and show in dedicated information hub
When scanning files we discard a lot of information that could be used to quickly identify issues or improve a collection.

Examples include:
* Corrupted files
* Wrong filenames (ts container with mp4 extension)
* Inappropriate containers (h264 stream in ts container, needlessly requiring transcode)
* Pointless gallery compression
* Duplicate files

We could save a lot of tech support by logging this info and putting it in an easily searchable view which quickly lets the user know why something does not work as expected.

It could be as simple as a scene_warnings table with id/string that we push to while scanning, and a tab in settings which displays the table, with a more expanded explanation of the issue.

InfiniteTF:
Creates a table for logging arbitrary scene errors. Currently it's used for logging duplicates found during a scan, and ffprobe/ffmpeg errors that occur during scan. Additionally I've made a PR so the perceptual dupe plugin can write to the same table.

For each "error" we can set the relevant scene, error type, when it will reoccur (so we can clear them out before rerunning scan for instance), and whether there are additional related scenes (only used for the perceptual dupe plugin currently.

The results are viewable/searchable in the Settings->Warnings panel.

I would like to add logging of "permanent" errors as well, for instance files requiring transcode and files which have errors during sprite creation, but it requires a bit more thought. At some point I'm planning to add a health check task as well which will scan your collection for errors.

Resolves #837


pkg/ffmpeg/encoder_transcode.go
```diff
@@ -64,7 +64,7 @@ func (e *Encoder) Transcode(probeResult VideoFile, options TranscodeOptions) {
 		"-strict", "-2",
 		options.OutputPath,
 	}
-	_, _ = e.run(probeResult, args)
+	_, _, _ = e.run(probeResult, args)
```
WithoutPants, line 67: If the return values aren't used, just do `e.run(...)`. Multiple instances of this in this file.

WithoutPants (changes requested):

pkg/ffmpeg/encoder_transcode.go
InfiniteTF: Fixed. Rebased and switched migration to 14 as well.

bnkai (approved):

WithoutPants:
I want to defer this PR while I'm working on an overhaul of the scene/file system.

victorhooi:
This is quite useful as we often get video files that simply don't seem to be scanned - so an easy way of grepping for these files in the UI, or seeing them there would be great.

stg-annon:
This sill necessary after the files_refactor branch merged?

WithoutPants:
I think it can be closed. We don't currently have a means to handle ffprobe/ffmpeg errors during scanning, but I think that can be handled in a new PR.

PR #857, opened 2020-10-15
Fix loading issue in galleries and redirect on gallery creation

InfiniteTF:
Added an `<ErrorMessage>` component. Also refactored the gallery loading to use discriminated prop types so we don't have to circumvent the type system. After the change setting isNew=false will guarantee gallery has a value, and vice versa. Using `Partial` and the non-null assertion operator (!) is risky and can frequently lead to bugs since it punches holes in the type system.

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryDetails/Gallery.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryDetails/Gallery.tsx
@@ -22,7 +22,7 @@ export const Gallery: React.FC = () => {
   const history = useHistory();
   const isNew = id === "new";
 
-  const [gallery, setGallery] = useState<Partial<GQL.GalleryDataFragment>>({});
+  const [gallery, setGallery] = useState<GQL.GalleryDataFragment>();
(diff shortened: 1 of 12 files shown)


WithoutPants:
This flashes up the `No gallery with id <i>{id}</i> found.` error message briefly before the gallery is loaded. 

I think `if (!gallery || !data?.findGallery)` shouldn't include `gallery` and should probably check for `isNew` instead.

InfiniteTF:
> This flashes up the `No gallery with id <i>{id}</i> found.` error message briefly before the gallery is loaded.
> 
> I think `if (!gallery || !data?.findGallery)` shouldn't include `gallery` and should probably check for `isNew` instead.

Thanks, the issue was a disconnect between the loading and gallery/image state due to using useEffect which takes effect on the next render. I've cleaned up gallery/image/scene/performers to use the apollo result directly so it's always in sync with the loading state. Made some minor changes to accomodate that, primarily adding an updater to the OMutations so they update the apollo cache automatically and we no longer have to manage it manually.

WithoutPants (approved):

PR #858, opened 2020-10-15
Add shim for String.prototype.replaceAll

InfiniteTF:
replaceAll is still pretty fresh so we should shim it for now. It's a no-op in any new browsers so no risk of breaking things.


WithoutPants:
Is this missing a file or change? Build gives:

```
Failed to compile.

./src/App.tsx
Module not found: Can't resolve 'string.prototype.replaceall' in 'D:\dev\stash\ui\v2.5\src'
```

InfiniteTF:
Duh, I keep adding node packages to the root folder. Fixed now.

WithoutPants (approved):

PR #873, opened 2020-10-20
Add selection and export for all list pages

WithoutPants:
Adds checkbox selection for all list pages, and adds export options to export all or a selection of objects. Includes some refactoring of components.


bnkai:
There seems to be an issue with galleries export. No images get exported if you export a gallery or the other way round also, no galleries get exported when you export the relevant images. I think it may be that the selective import/export code was merged before the image changes.

For the below idk if they were inentional, they should be from the previous selective import/export PRs :
- No scenes get exported when you export a movie, you have to export the specific scenes and then the movie gets exported.
- For performers,studios,tags nothing relevant is exported (relevant images,scenes) when you export them.

WithoutPants:
* Gallery export now includes contained images.
* Image export now includes galleries if related objects is ticked.
* Movie export includes movie scenes if related objects is ticked.

PR #877, opened 2020-10-21
Handle zip file modification

### LINKED ISSUE #469: [Feature] Allow Sorting Scenes By Last Modified Date Of Video File
I want to sort scenes by the modified dates of the video files themselves so that they appear in order from added most recently to least recently. There are three options for sorting by date but none of them appear to do what I want.

It should be possible to sort by the last modified date of the file on the filesystem.

N/A

**Additional context**
N/A

WithoutPants:
Currently, if you modify a zip file after its already been scanned in, then stash doesn't detect the modification and does not detect any new or changed files within the zip file.

This change compares the file modification time against the modification time of the gallery, and if it is newer, then the checksum is recalculated and updated (along with the modification time). This means that the gallery and images can be updated.

Resolves #469


bnkai:
This works provided that the user hasn't edited the gallery afterwards. Then the updated at will be newer and it wouldnt trigger the rescan.
Is adding the file modification time for scanned files (video,image,zip files) to the db an option ? This would allow us to also replace scenes with eg higher quality ones, transcoded ones with the same name without the need to rescrape/edit most metadata.

In the zipped file gallery case what happens if a user removes images from the zip file? Should the user be encouraged to run a clean after the scan?
Also if he updates the zip file with a new `cover.jpg`(same name new image) , would it be updated or because the path exists it will be detected as a dupe? 
(will  actually test the above cases when i test the PR )

WithoutPants:
I have added a schema migration to store the file modification times for scenes, galleries and images. The scan process now checks all of these and refreshes the file data if they are changed. It will also regenerate thumbnails if the image changed, since this is the only candidate where the underlying data is likely to have changed. This should resolve the swapping of cover images issue.

pkg/manager/task_scan.go
```diff
@@ -48,19 +48,33 @@ func (t *ScanTask) scanGallery() {
...
 
-		// if the mod time of the zip file is newer than that of the associated
-		// gallery, then recalculate the checksum
-		modified, err := t.isFileModified(gallery.UpdatedAt)
-		if err != nil {
-			logger.Error(err.Error())
-			return
+		// if file mod time is not set, set it now
+		if !gallery.FileModTime.Valid {
+			t.updateFileModTime(gallery.ID, fileModTime, &qb)
```
bnkai, line 62: This is not enough.
While the existing gallery's mod time gets updated the images included in the zip file don't get their mod times updated and end up with a null file_mod_time after the scan.

pkg/manager/task_scan.go
```diff
@@ -246,7 +291,38 @@ func (t *ScanTask) associateGallery(wg *sync.WaitGroup) {
...
+		logger.Error(err.Error())
+		return
+	}
+
 	if scene != nil {
+		// if file mod time is not set, set it now
+		if !scene.FileModTime.Valid {
+			t.updateFileModTime(scene.ID, fileModTime, &qb)
+
+			// update our copy of the gallery
```
bnkai, line 306: typo 
```golang
			// update our copy of the scene
```

bnkai:
From a quick test ( will make a more thorough test later doday)

Something is wrong when updating images (replace an image with a newer with same name) inside a zip file. While the gallery itself gets its checksum updated the updated image inside doesn't. (For this case i used a new gallery created after the migration as the mod time for images in existing zip galleries won't  get updated as mentioned above)

The rescan scenes needs some more code also. If a scene is modified and the rescan trigerred it should also check if the checksum changed and recreate the screenshot image and delete the old screenshot and generated parts. As it is now the checksum and details are updated but a new scan/clean can't detect the changes needed for the screenshot/gen parts because the path is the same.

A sort by mod_time for the scenes/images can also be useful ( i needed that for the testing )
```diff
diff --git a/ui/v2.5/src/models/list-filter/filter.ts b/ui/v2.5/src/models/list-filter/filter.ts
index de035eff..ab63a179 100644
--- a/ui/v2.5/src/models/list-filter/filter.ts
+++ b/ui/v2.5/src/models/list-filter/filter.ts
@@ -122,6 +122,7 @@ export class ListFilterModel {
           "framerate",
           "bitrate",
           "random",
+          "file_mod_time",
         ];
         this.displayModeOptions = [
           DisplayMode.Grid,
@@ -153,6 +154,7 @@ export class ListFilterModel {
           "o_counter",
           "filesize",
           "random",
+          "file_mod_time",
         ];
         this.displayModeOptions = [DisplayMode.Grid, DisplayMode.Wall];
         this.criterionOptions = [
```

bnkai:
It seems to me that 
```
// if the mod time of the file is newer than that of the associated
		// image, then recalculate the checksum and regenerate the thumbnail
// if the mod time of the file is newer than that of the associated
		// scene, then recalculate the checksum and regenerate the thumbnail
```
this might be interfering with the detection of modified files during the scan.

For plain images and videos is the mod time changed if they get moved or replaced ? Replacing an image with another one the mod time seems to stay the same so maybe just compare to see if the mod date is different for  plain images / videos instead of newer? I had to `touch` an image file after replacing to get it detected as modified.

WithoutPants:
> The rescan scenes needs some more code also. If a scene is modified and the rescan trigerred it should also check if the checksum changed and ~ recreate the screenshot image and ~delete the old screenshot and generated parts. As it is now ~the checksum and file details are updated but a new scan/~clean can't detect the changes needed for the ~screenshot/~gen parts because the path is the same.

I've changed the code so that the local copies of the scene/image are updated when the hash is recalculated, so it should now regenerate the thumbnails as needed. I have elected _not_ to remove the existing generated files in case the original file was moved elsewhere - ie original file was renamed, new file was added with the old file name. This particular use case still causes issues, but I don't think its worth investing the effort for this PR for this particular corner case. It should be addressed at a later point.

All other points should hopefully be addressed now.

bnkai:
With the latest commit most cases work alright.
For the case where we replace an image inside a zip file with a new one with the same name i get this error 
```bash
INFO[1572] Starting scan of 11950 files. 0 New files found 
INFO[1579] /path/reducted/test_gallery.zip has been updated: rescanning 
INFO[1579] Calculating checksum for /path/reducted/test_gallery.zip... 
DEBU[1579] Checksum calculated: 9708935124aa3b8da51444d9f7870d32 
INFO[1579] /path/reducted/test_gallery.zipimage.webp has been updated: rescanning 
INFO[1579] Calculating checksum for /path/reducted/test_gallery.zipimage.webp... 
ERRO[1579] open /path/reducted/test_gallery.zipimage.webp: invalid argument 
INFO[1579] Finished scan                                
INFO[1579] Finished gallery association 
```
and the zip file ( gallery ) info doesn't get updated
Replacing plain scenes/images seems to work ok though.

As for the generated parts not being removed i think the clean function probably needs a second `stage` ( not for this PR obviously ) to do some additionals checks eg the reverse of what we are doing now. Check all the generated files to see if the related scene/image/... file exists in the db

WithoutPants:
Rescanning images in zips should now be fixed.

bnkai:
Everything seems to work fine now.

bnkai (approved):

PR #878, opened 2020-10-21
Query url parameters

WithoutPants:
I found a need to manipulate the values that are fed into `queryURL` when using `sceneByFragment`. This change introduces `queryURLReplace` section which contains a map of placeholder fields (`filename`, `oshash`, `checksum`, `title`) to regex replacement lists. 

I'm using this to enable scraping from filename on the Voyeurweb scraper. It's files are named `xxx-yyy.mp4` and only the `xxx` part is used in the query URL. See the following `sceneByFragment` section:

```
  queryURLReplace:
    filename:
      - regex: ^.*[/\\](\d*)-\d*\..*
        with: $1
```

### DIFF
--- a/pkg/scraper/config.go
+++ b/pkg/scraper/config.go
@@ -114,7 +114,8 @@ type scraperTypeConfig struct {
 	Scraper string        `yaml:"scraper"`
 
 	// for xpath name scraper only
-	QueryURL string `yaml:"queryURL"`
+	QueryURL             string               `yaml:"queryURL"`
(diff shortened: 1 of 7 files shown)


bnkai:
This looks/works ok
It makes it possible to simplify the JavLibrary scraper by using
```yaml
  queryURL: https://www.javlibrary.com/en/vl_searchbyid.php?keyword={filename}
  queryURLReplace:
    filename:
      - regex: (.*[^a-zA-Z0-9])*([a-zA-Z0-9]+-\d+).*
        with: $2
```
to search for movie ids.

PR #879, opened 2020-10-21
Add basic i18n implementation

SpedNSFW:
To be used as a reference point for any future i18n additions for either new languages or more translatable content.

This basic implementation covers the one glaring difference between the correct and incorrect variants of the English language by adding "Favourite" which is used for performer banners.

Ideally, every line should exist in the i18n JSON files to make additions of new languages easier.


ui/v2.5/src/App.tsx
```diff
@@ -40,8 +40,8 @@ const intlFormats = {
 
 export const App: React.FC = () => {
   const config = useConfiguration();
-  const language = config.data?.configuration?.interface?.language ?? "en-US";
-  const messageLanguage = language.slice(0, 2);
+  const language = config.data?.configuration?.interface?.language ?? "en-GB";
```
WithoutPants, line 43: Why are we changing the default language?

PR #881, opened 2020-10-22
Update freeones scraper

WithoutPants:
Updates freeones scraper based on the community scraper PR (https://github.com/stashapp/CommunityScrapers/pull/232). Fixes incorrect name.

### DIFF
--- a/pkg/scraper/freeones.go
+++ b/pkg/scraper/freeones.go
@@ -19,7 +19,7 @@ performerByName:
 performerByURL:
   - action: scrapeXPath
     url:
-      - https://www.freeones.xxx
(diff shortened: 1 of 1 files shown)


bnkai:
We just merged an updated community scraper.
It matches urls from .xxx and .com but uses .com for the actual scraping as it seems that the .com is more `stable` and it now uses the same layout as xxx.
Any performers scraped with old urls ( the very old php with .com , the recent .xxx ) will be given a new with .com if they get rescraped.
I tested a build with the now merged scraper included and it worked fine for new and existing rescraped performer urls.

PR #887, opened 2020-10-24
Don't set default studio image

### LINKED ISSUE #866: [Bug Report] Studio: Is Missing Image filter doesn't work
The "Is Missing: Image" filter for Studios does not work, I assume because it still considers the placeholder image used as an image.
Placeholder

In reality, it should treat the placeholder as though there was no image.

WithoutPants:
When creating studio, it currently sets the default image if one is not provided. This causes incorrect behaviour for the Is Missing Image filter. Changed so that the default image is no longer set if it is not provided. Placeholder image is used for studios without images when viewing them, but they will correctly be filtered for is missing image.

Resolves #866

### DIFF
--- a/pkg/api/resolver_mutation_studio.go
+++ b/pkg/api/resolver_mutation_studio.go
@@ -19,14 +19,12 @@ func (r *mutationResolver) StudioCreate(ctx context.Context, input models.Studio
 	var imageData []byte
 	var err error
 
-	if input.Image == nil {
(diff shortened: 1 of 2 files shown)


bnkai:
Looks/Works ok.
For existing studios should the user  manually edit-> clear the existing default images?
```sql
DELETE FROM studios_image WHERE length(image) = 405 ;
```
seems to work for me. ( I am not sure how probable it is to have another image with that size)

PR #890, opened 2020-10-24
Add Title attribute to Scene cards

### LINKED ISSUE #888: [Feature] Add title attribute to scene card
Because overflow is hidden for scene card titles (`.card-section-title`), you are unable to see the full title if it's too long. I'd like to see the HTML `title` attribute added to either `.scene-card` or `.card-section-title` with the title as the value so hovering over will show the full title.

I'll implement this when I get a free moment if no one else already has by then. This issue is just serving as a reminder.

SpedNSFW:
Closes #888


WithoutPants:
This causes the tooltip to appear over the video preview, which I think will be irritating. I'd move it to the actual title element.

WithoutPants:
Closing this in favour of #932

PR #891, opened 2020-10-24
Cache generated exclude regex

JoeSmithStarkers:
Due to my abuse of the exclude regex to prevent scanning re scanning known duplicates (my config file contains 2580 excludes). I noticed the way the new code works re-compiles/generates the regex for each file scanned, rather than being cached.

This is a simple change to cache the generated exclude regex for each path scanned, rather than being recompiled for every file scanned.

### DIFF
--- a/pkg/manager/exclude_files.go
+++ b/pkg/manager/exclude_files.go
@@ -38,15 +38,20 @@ func excludeFiles(files []string, patterns []string) ([]string, int) {
 	}
 }
 
+func matchFileRegex(file string, fileRegexps []*regexp.Regexp) bool {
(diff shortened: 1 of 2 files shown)


bnkai:
Looks ok. 
If we do that though shouldn't we also take into account the case where a user has a lot of paths like you have excludes and do the regexgenerate once per scan ?

WithoutPants:
> Looks ok.
> If we do that though shouldn't we also take into account the case where a user has a lot of paths like you have excludes and do the regexgenerate once per scan ?

That sounds like a fairly niche corner case, and probably not worth messing with the interface for.

PR #904, opened 2020-10-27
Add page sizes up to 1000

WithoutPants:
Increases the hard limit to 1000 and adds new page size options 250/500/1000.


SpedNSFW:
Should the size limits for SceneFilenameParser also be increased?
https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/components/SceneFilenameParser/ParserInput.tsx#L124

Issue #893, opened 2020-10-24
[Bug Report] No such column: performers.stash_id


peolic:
**Describe the bug**
Error when filtering Performers using `Is missing -> stash_id`

**To Reproduce**
1. Use latest develop
2. Go to Performers tab
3. Filter using: `Is missing -> stash_id`

**Expected behavior**
I assume this is not supposed to be here as the other tabs (scenes & studios) don't have that option.

**Screenshots**
![image](https://user-images.githubusercontent.com/66393006/97084855-65599180-1622-11eb-819b-5dcb23255544.png)

**Stash Version:**
v0.3.0-51-g77da544

**Desktop:**
 - OS: Windows 10 Pro 2004 x64
 - Browser: Firefox
 - Version: 81.0.2

PR #912, opened 2020-10-31
Fix outstanding tagger issues

### LINKED ISSUE #903: [Bug Report] Match fingerprints causes UI crash when >100 scenes shown
**Describe the bug**
UI displays the error message: `Unhandled Rejection (Error): response error: Too many fingerprints.`

**To Reproduce**
Open the scene tagger view, setting page size to 120. Click match fingerprints button.

**Expected behavior**
Any errors from the match fingerprint operation should be handled cleanly.

Ideally where there are more than 100 scenes displayed, the tagger should be performing a series of queries, rather than everything at once.

InfiniteTF:
* Fixes issue where users could sometimes not save due to a state race-condition.
* Fixes duration sorting so closest match is sorted first.
* Adds is_missing stash_id filter for scene/performer/studio.
* Paginates fingerprint lookups so users can search for arbitrary amount of scenes.
* Disable `Match Fingerprint` button after search is performed, and show number of matches.

Resolves #903, #893, #915

### DIFF
--- a/ui/v2.5/src/components/Tagger/Tagger.tsx
+++ b/ui/v2.5/src/components/Tagger/Tagger.tsx
@@ -132,6 +132,7 @@ const TaggerList: React.FC<ITaggerListProps> = ({
     { loading: submittingFingerprints },
   ] = GQL.useSubmitStashBoxFingerprintsMutation({
     onCompleted: (result) => {
+      setFingerprintError("");
       if (result.submitStashBoxFingerprints)
         clearSubmissionQueue(selectedEndpoint.endpoint);
     },
--- a/ui/v2.5/src/components/Tagger/Tagger.tsx
+++ b/ui/v2.5/src/components/Tagger/Tagger.tsx
@@ -115,7 +115,8 @@ const TaggerList: React.FC<ITaggerListProps> = ({
       })
       .catch(() => {
         setLoading(false);
-        const { [sceneID]: undefined, ...results } = searchResults;
+        // Destructure to remove existing result
+        const { [sceneID]: unassign, ...results } = searchResults;
         setSearchResults(results);
         setSearchErrors({
           ...searchErrors,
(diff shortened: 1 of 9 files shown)


WithoutPants (approved):

WithoutPants:
Just discovered: UI is still not handling error conditions when searching scenes by fingerprint or search. 

To reproduce, show the tagger view for some scenes. Kill the stash process, then click Search or Match fingerprints. Causes UI stack trace to show.

InfiniteTF:
> Just discovered: UI is still not handling error conditions when searching scenes by fingerprint or search.

Thanks for the heads up, I forgot to handle network errors. Should be resolved now.

WithoutPants:
I've just done some minor changes to show the network error under the specific scene being searched, rather than at the top.

InfiniteTF:
> I've just done some minor changes to show the network error under the specific scene being searched, rather than at the top.

Looks good. I just tweaked it a bit so scene network errors aren't cleared on fingerprint search, and vice versa.

ui/v2.5/src/components/Tagger/Tagger.tsx
```diff
@@ -168,9 +167,6 @@ const TaggerList: React.FC<ITaggerListProps> = ({
       .filter((s) => s.stash_ids.length === 0)
       .map((s) => s.id);
 
-    // clear search errors
-    setSearchErrors({});
```
WithoutPants: I'm not sure why this was done. With this change, if the text search fails but the fingerprint search does not, then it still shows the error message and does not show the fingerprint matched scene.

InfiniteTF: The issue was that if a fingerprint wasn't found the scene state would flip from `network error` to `no results found` despite no search having been performed. I hadn't considered your scenario though. I've corrected this now so the scene results are wiped if the query fails, and reinstated the search error clearing.

InfiniteTF:
Added a small fix to not abort if the user tries to submit fingerprints for a scene that for some reason doesn't exist. This could easily happen for a number of different reasons, and would basically force the user to wipe the fingerprint cache. Long term the submission state should be moved to the db so it can be properly cleaned up, and persist across browsers.

WithoutPants (approved):

PR #913, opened 2020-11-01
Fix image clean

WithoutPants:
The clean task was using the video exclusion patterns when checking images and galleries. This meant that images and galleries could be scanned into the system, then subsequently cleaned out. Changed so that clean task checks the correct exclusion patterns for the file type.

### DIFF
--- a/pkg/manager/task_clean.go
+++ b/pkg/manager/task_clean.go
@@ -41,9 +41,7 @@ func (t *CleanTask) shouldClean(path string) bool {
 	// use image.FileExists for zip file checking
 	fileExists := image.FileExists(path)
 
-	if fileExists && t.getStashFromPath(path) != nil {
-		logger.Debugf("File Found: %s", path)
-	} else {
+	if !fileExists || t.getStashFromPath(path) == nil {
 		logger.Infof("File not found. Cleaning: \"%s\"", path)
 		return true
 	}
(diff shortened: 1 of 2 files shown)


pkg/manager/task_clean.go
```diff
@@ -43,10 +43,6 @@ func (t *CleanTask) shouldClean(path string) bool {
 
 	if fileExists && t.getStashFromPath(path) != nil {
```
bnkai, line 44: Imho we can also remove the 
`logger.Debugf("File Found: %s", path)`
as it pollutes the log file 
We can then break the if else to a simple if e.g.
```golang
	if !fileExists || t.getStashFromPath(path) ==  nil {
		logger.Infof("File not found. Cleaning: \"%s\"", path)
		return true
	}

File: pkg/manager/task_clean.go [hunk 2]
```diff
@@ -60,6 +56,11 @@ func (t *CleanTask) shouldCleanScene(s *models.Scene) bool {
 		return true
 	}
 
+	if matchFile(s.Path, config.GetExcludes()) {
```
bnkai, line 59: Imo the matchFiles for the scenes,galleries,images should be checked after the checks for Exclude.Video, Exclude.Image, extension.

bnkai (approved):

PR #916, opened 2020-11-01
Migrate PR build storage from transfer.sh to Gofile.io

bnkai:
Since transfer.sh is EOL oshi.at was the most compatible i could find to store the PR builds.


WithoutPants:
The stash-win download link didn't work - file not found.

bnkai:
Switched to `gofile.io` as it seems more stable. Tried also `file.io` but was getting `too many requests` after a  couple of uploads

PR #921, opened 2020-11-03
Add selective cross compile

JoeSmithStarkers:
Modify the cross compile script to allow individual target builds, including the ability to bypass the packr step, with the -fast suffix, eg scripts/cross-compile.sh windows-fast

### DIFF
--- a/scripts/cross-compile.sh
+++ b/scripts/cross-compile.sh
@@ -1,17 +1,64 @@
 #!/bin/sh
 
+# "stashapp/compiler:develop" "stashapp/compiler:4"
+COMPILER_CONTAINER="stashapp/compiler:4"
+
 BUILD_DATE=`go run -mod=vendor scripts/getDate.go`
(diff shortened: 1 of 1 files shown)


bnkai:
Just out of curiousity how long does the packr step take for you @JoeSmithStarkers ? I tried with a fresh checkout and it seems to take ~ 7s for me ( debian linux / nvme  ssd for the build )

JoeSmithStarkers:
> Just out of curiousity how long does the packr step take for you @JoeSmithStarkers ? I tried with a fresh checkout and it seems to take ~ 7s for me ( debian linux / nvme ssd for the build )

Currently it takes me 1 minute 33 seconds (2019 MBP i9 (turbo boost disabled)), but it use to be more of a problem for me because i had goland ide setup incorrectly and it would insist on pulling all dependences into pkg/pkg, which caused the packr scanning to take forever.

PR #928, opened 2020-11-06
Lightbox keybinds

### LINKED ISSUE #293: [Bug Report] FFProbe permission denial when scanning
On Mac Catalina running Chrome. My videos are stored on an external drive. Whenever I scan for videos each video produces the same error in logs, whereas I would expect that the videos would be loaded into stash. I have tried on both currently available builds (2019/12/27).
An example error is shown below:

2019-12-27 12:00:50 Error    FFProbe encountered an error with <[path of video]>.
Error JSON:

Error: fork/exec /Users/[user]/.stash/ffprobe: permission denied

### LINKED ISSUE #923: Image Navigation
**Describe the bug**
Some things are still wrong with the navigation through images. Most outstanding, arrow keys do not work for the picture viewer. When used, the gallery is going to the next result page, but the image stands still. 

**To Reproduce**
1. Go to a Gallery or Images
2. Click on wall mode (to get the image carousel, not the metadata page)
3. Use Arrow keys to navigate carousel

WithoutPants:
Fixes #923 

Disables other key binds while image lightbox is open.

### DIFF
--- a/ui/v2.5/src/components/Images/ImageDetails/Image.tsx
+++ b/ui/v2.5/src/components/Images/ImageDetails/Image.tsx
@@ -10,7 +10,6 @@ import {
 import { ErrorMessage, LoadingIndicator, Icon } from "src/components/Shared";
 import { useToast } from "src/hooks";
 import { TextUtils } from "src/utils";
-import * as Mousetrap from "mousetrap";
 import { OCounterButton } from "src/components/Scenes/SceneDetails/OCounterButton";
 import { ImageFileInfoPanel } from "./ImageFileInfoPanel";
 import { ImageEditPanel } from "./ImageEditPanel";
--- a/ui/v2.5/src/components/Images/ImageDetails/Image.tsx
+++ b/ui/v2.5/src/components/Images/ImageDetails/Image.tsx
@@ -10,6 +10,7 @@ import {
 import { ErrorMessage, LoadingIndicator, Icon } from "src/components/Shared";
 import { useToast } from "src/hooks";
 import { TextUtils } from "src/utils";
+import * as Mousetrap from "mousetrap";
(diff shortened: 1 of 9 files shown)


bnkai:
relevant issue should be #923 not `293`
works ok

ui/v2.5/src/components/MainNavbar.tsx
```diff
@@ -14,6 +14,8 @@ import { SessionUtils } from "src/utils";
 import { Icon } from "src/components/Shared";
 import { Manual } from "./Help/Manual";
 
+const Mousetrap = require("mousetrap-pause")(require("mousetrap"));
+
```
InfiniteTF, line 18: It's not really necessary to use require . You can import the modules normally:
```
import Mousetrap from 'mousetrap';
import MousetrapPause from 'mousetrap-pause';
MousetrapPause(Mousetrap); 
```

You'll also have to add `declare module "mousetrap-pause";` to `src/globals.d.ts`.

ui/v2.5/src/components/Images/ImageDetails/Image.tsx
```diff
@@ -10,7 +10,6 @@ import {
 import { ErrorMessage, LoadingIndicator, Icon } from "src/components/Shared";
 import { useToast } from "src/hooks";
 import { TextUtils } from "src/utils";
-import * as Mousetrap from "mousetrap";
```
InfiniteTF, line 13: I've actually been adding in regular mousetrap imports in #914 since Typescript 4.0 complains otherwise. There's no reason not to import them, MousetrapPause will always run when the page is loaded and affects every import.

InfiniteTF (approved):

PR #932, opened 2020-11-07
Add TruncatedText component

### LINKED ISSUE #767: [Feature] Show full text on hover for truncated text
We use truncate multiple places without any way of seeing the full text. Additionally the truncation is done with different methods and results.

We should create a `<TruncateText>` component which allows you to set the number of lines before truncating, and shows a popover with the full text when necessary.

### LINKED ISSUE #888: [Feature] Add title attribute to scene card

InfiniteTF:
I've created a `TruncatedText` component which truncates text to a configurable line count and shows a tooltip with the complete text when it is hovered for a configurable amount of time. I've switched the majority of cases of `text-truncate` over to using the `TruncatedText` component so we're consistent.

I removed titles from tab panels where necessary since it causes the title to show up anywhere inside the pane when something is hovered.

image

Resolves #767 
Resolves #888


ui/v2.5/src/components/Tagger/Tagger.tsx
```diff
@@ -355,15 +355,12 @@ const TaggerList: React.FC<ITaggerListProps> = ({
...
-                className="scene-link"
-                title={scene.path}
-              >
-                {originalDir}
-                <wbr />
-                {`${file}.${ext}`}
+            <div className="col-md-6 my-1 align-self-center">
+              <Link to={`/scenes/${scene.id}`} className="scene-link">
+                <TruncatedText
+                  text={`${originalDir}{file}.${ext}`}
```
peolic, line 361: I came here to comment on one thing, but I found a second one.
1. The `ext` variable already contains the dot (`.mp4`), so the dot is always duplicated here.
2. One `$` was removed here.
```suggestion
                  text={`${originalDir}${file}${ext}`}
```

InfiniteTF: Thanks, I've noticed the double periods before but somehow I never thought to look for it.

PR #937, opened 2020-11-10
Only set stash from env if not set

### LINKED ISSUE #936: [Bug Report] Exclude Images tickbox is unselecting itself when Stash is restarted/updated (Stash via Docker)
**Describe the bug**
The 'Exclude Image' tickbox in the Configuration screen is not remembering being ticked when Stash is restarted or updated, so I have to keep ticking it and pressing save.

I'm unsure if this also happens with the "Exclude Video" tickbox as well.

**To Reproduce**
1. Go to Configuration, tick the "Exclude Image" tickbox for the folder path (I only have one folder path listed)
2. Click on Save at the bottom of the Configuration page
3. Restart or Update Stash
4. Open Stash, and go into Configuration screen and the tickbox is de-selected.

Stash running via Docker on Synology NAS

Version: | v0.3.0-70-g6bb24d1
Build hash: | 6bb24d1
Build time: | 2020-11-09 01:33:43

 - OS: Windows 10
 - Browser Chrome
 - Version 86.0.4240.183 (Official Build) (64-bit)

WithoutPants:
Fixes #936 

Changes the startup so that it only sets the stash libraries from the `STASH_STASH` environment variable if there is not already a stash library set in the config. Means that `STASH_STASH` will be used when the app is started for the first time, and afterwards will be ignored unless all stash library paths are removed from the config.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -131,10 +131,14 @@ func initEnvs() {
 	viper.BindEnv("host")          // STASH_HOST
 	viper.BindEnv("port")          // STASH_PORT
 	viper.BindEnv("external_host") // STASH_EXTERNAL_HOST
-	viper.BindEnv("stash")         // STASH_STASH
 	viper.BindEnv("generated")     // STASH_GENERATED
 	viper.BindEnv("metadata")      // STASH_METADATA
 	viper.BindEnv("cache")         // STASH_CACHE
+
+	// only set stash config flag if not already set
+	if config.GetStashPaths() == nil {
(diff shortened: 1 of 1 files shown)


bnkai:
Tested, works as intended.
The envs are mostly used to streamline a stash installation without going through the setup page ( envs are recommended over a default config file for docker setups afaik ) . The unraid docker image/setup i think is the best example. Any thoughts on applying the same for the rest of stash envs from #936 ? The path is probably the only one that users will change/add to so it shouldn't matter that much.

bnkai (approved):

PR #939, opened 2020-11-12
Added screenshots/previews to tagger list

JoeSmithStarkers:
Added screenshots and hover over preview videos.

image

More than happy to receive any opinions or ideas of styling....

### DIFF
--- a/graphql/documents/data/gallery-slim.graphql
+++ b/graphql/documents/data/gallery-slim.graphql
@@ -1,0 +1,28 @@
+fragment GallerySlimData on Gallery {
+  id
+  checksum
+  path
+  title
+  date
+  url
+  details
+  rating
+  image_count
+  cover {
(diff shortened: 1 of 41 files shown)


InfiniteTF:
If possible I would recommend exporting the existing ScenePreview component and reusing that here, so we don't duplicate functionality. You should be able to override the styles as necessary.

WithoutPants:
This has broken the layout on mobile devices. Need to bring the preview image and text onto separate rows when on small devices.

image

vs how it looks without the change:

image

I tried fixing it myself but didn't have much luck.

InfiniteTF:
Should work fine now. Looks a bit strange with the tiny previews, but not a lot of ways to show it without taking up all the space.

image

WithoutPants (approved):

Issue #886, opened 2020-10-23
[Feature] Partial scan/autotag based on path parameter


rg9400:
**Is your feature request related to a problem? Please describe.**
After you do an initial scan and have a large library, subsequent scans/autotags to add new files can be incredibly time-consuming. Generate itself is fast since it only runs on new items. For my library of 24k scenes, a scan + generate + autotag can take 30 minutes or more. Additionally, it is not efficient to rescan the entire disk, especially if you use cloud storage. Finally, it makes automation more difficult as if you are running the above steps after a download, they can clog up the downloads which haven't been processed yet.

By specifying a path, it isolates the number of items a task would run on, and I think it could drastically cut down the total time as it would only be spent actually scanning/autotagging/generating vs running on existing files that have already been processed.

**Describe the solution you'd like**
Similarly to Plex/Emby/Jellyfin, give an option to run the scan and autotag processes on a specific subfolder only. Ideally, this option would be available in the web UI, but if it is only available via GraphQL, that can still work.

I know there's some refactoring to give the ability to run all 3 tasks via a single call, so ideally, this would work with that refactor as well.

**Describe alternatives you've considered**
Live with the inefficient scan times, and probably carve that out of any automation I would run. Instead, I would need to scan maybe every 24 hours only, which still results in inefficient scans, but would at least not clog up the download queue.

rg9400:
One option for this feature is to add autotag and clean as parameters for a scan as well like is now available with generate tasks. This way, it would be possible to specify a path for a clean/autotag task by running a scan on it.

That way, if someone wants to develop a tool like [Autoscan ](https://github.com/Cloudbox/autoscan) for Stash (or extend it to Stash), it should be easier as only a single GraphQL call would need to be made for various triggers like inotify or maybe even a future Arr variant for this space. If a file is removed, renamed, added, etc., the impacted path could be passed along while running a scan with generate + clean + autotag on to basically bring that path up to date within Stash.

WithoutPants:
> One option for this feature is to add autotag and clean as parameters for a scan as well like is now available with generate tasks. This way, it would be possible to specify a path for a clean/autotag task by running a scan on it.
> 
[...]

PR #940, opened 2020-11-13
Selective scan

WithoutPants:
Adds an extra button to the tasks page to perform a selective scan:

image

This brings up a dialog allowing the user to select specific folders to scan. It will scan these folders and their subfolders.

image

It displays the stash library paths initially, allowing the user to navigate via their library paths.

A warning is logged during the scan operation if a specified path is not within the library paths.

Partially fulfils #886


bnkai:
Looks  ok and seems  to work ok also.

SpedNSFW:
Works a treat.

PR #943, opened 2020-11-15
Added natural sort for scene and image titles

JoeSmithStarkers:
Implement natural sort for scenes/image titles.


bnkai:
Cant we use <https://github.com/stashapp/stash/blob/develop/pkg/utils/natural_sort.go> the one included already?

bnkai:
Also isn't case insensite better ? doing a strings.ToLower to s and s2 prior to comparing should be enough.

JoeSmithStarkers:
> Cant we use https://github.com/stashapp/stash/blob/develop/pkg/utils/natural_sort.go the one included already?

Oh I didn't know we already had one. However I did find that implementation when I was looking and the reason why I chose fvbommel/sortorder over facette/natsort is performance. facette/natsort uses regex and process the whole string to chunkify it, instead of processing it a character by character with early return. (multi-pass vs single-pass)

https://groups.google.com/g/golang-nuts/c/Puy4TQ20G2w

I plan on doing some performance testing, and i will benchmark the existing solution. 

> Also isn't case insensite better ? doing a strings.ToLower to s and s2 prior to comparing should be enough.

I'm happy to make it case insensitive, but due to performance reason I believe modifying the fvbommel/sortorder to do it character by character is better than processing the wholes strings with tolower(). (Comparing functions should exit early (eg before a single pass of the data) the majority of the time, eg ("abc" vs "bbc"). So anything that requires processing the whole string eg ToLower will be a largish overhead)

bnkai:
Agreed , i just saw all the extra files and was wondering if it was worth it.

PR #951, opened 2020-11-21
Changelog for 0.4 release

WithoutPants:
Added missing changelog entries and added note to verify libraries and rescan.


bnkai:
The only thing that i think is missing is #937. It might be worth it to mention it as it changes the current behaviour of the docker image. It shouldnt break anything but at the very least if someone has a problem they can have a look at the changelog and find the cause.

bnkai:
The Galleries section in the help panel probably needs an update to reflect to the images PR changes.
The cache point should be removed as the relevant task/functionality was removed also.

PR #964, opened 2020-11-25
Replace natural_sort with third party call

WithoutPants:
Removes the `natural_sort.go` file in favour of the third party library added in #943


bnkai:
Just tried it and it works ok for the sprite creation.
I  seem to remember that natural sort was also used to sort the contents of the zip file galleries, it seems that with the image PR this changed though.

bnkai (approved):

PR #965, opened 2020-11-26
Show studio as text in scene cards where studio image isn't set

WithoutPants:
Currently, studios are shown as the default image on scene cards if there is no studio image set. This PR changes that behaviour to show the studio title instead, in the same way as if "Show Studio as text" is checked in the settings.


bnkai:
Looks ok 
Just a reminder for users that had default images on studios before the PR that added separate image tables, that whey would need to clear the studio image and save for this to work.

bnkai (approved):

PR #974, opened 2020-11-30
Implement user customizable menu items

### LINKED ISSUE #970: [Feature] User customizable navbar items
As the project grows, more specialized features are generally added. These features are useful to some people, but can be less useful to others. By letting the user customize parts of the UI to their liking, we can continue to provide powerful features without overwhelming the user.

My suggestion is to let the user choose which items are shown on the navbar. Specifically, the user can show or hide any of these items based on their preference:

image

This becomes especially useful if, in the future, support for more content types is added (#11, #14).

I can't think of any.

**Additional context**
Here is a quick mockup of how the configuration could look:

image

aGlkZGVu:
Closes #970.

## Description

This PR adds a functional interface setting that lets people change what items are shown in the menu.

image

If there is a better place for the `CheckboxGroup` component, please let me know. Currently, it's located next to `SettingsInterfacePanel.tsx`, but it could be used as a generic component. I did find `src/utils/form.tsx` but it didn't really match the other things in that file (those usually render a particular grid layout, which was not suitable for this case). I'm happy to move it anywhere you see fit.

## Future work

The back-end preserves the order of menu items when set, and the navbar respects the order it reads from the config. So it should be relatively simple to allow users to change the order of items through settings. The only thing left to implement for that is are the form controls. I've left this out to reduce complexity of the PR, and because I personally don't need it.


ui/v2.5/src/components/MainNavbar.tsx
```diff
@@ -239,18 +265,20 @@ export const MainNavbar: React.FC = () => {
...
-                <LinkContainer activeClassName="active" exact to={i.href}>
-                  <Button className="minimal w-100">
-                    <Icon icon={i.icon} />
-                    <span>{intl.formatMessage(i.message)}</span>
-                  </Button>
-                </LinkContainer>
-              </Nav.Link>
-            ))}
-          </Nav>
+          <Fade in={!loading}>
```
aGlkZGVu, line 268: This `Fade` was added to avoid a noticeable "pop-in" when rendering the navbar for the first time. This is caused by the fact that we need to fetch the navbar items from the config before we can render the items.

pkg/manager/config/config.go
```diff
@@ -449,6 +450,11 @@ func GetCustomServedFolders() URLMap {
 }
 
 // Interface options
+func GetMenuItems() []string {
+	viper.SetDefault(MenuItems, []string{"scenes", "images", "movies", "markers", "performers", "studios", "tags"})
```
WithoutPants, line 454: Instead of setting a default, I think this should just get the string slice, and return the default if the slice is empty.

WithoutPants (changes requested):

pkg/manager/config/config.go
aGlkZGVu: Do I understand correctly that that would make it impossible to set the option to the empty slice explicitly? That could be a bit surprising behaviour when the user intentionally disables all menu items in settings (not sure why they would want that, but hey). Please let me know what you think, I'm happy to change it.

WithoutPants: Good point. Instead, call `viper.IsSet` to determine if the key is set. If it isn't, return the default list, otherwise return the string slice for the key.

aGlkZGVu: Sure thing, done.

WithoutPants (approved):

PR #976, opened 2020-12-01
Use changesets correctly when updating objects

WithoutPants:
Changes the behaviour of the update mutations so that they no longer require a full object. Adds a translator class which translates fields from the input object into an appropriate model-side value.

If the field is not included in the original request (absent from the client-side request or set to `undefined`), then it is not changed. If the field is included and is `null`, then it is treated as unsetting the value.

There should be no immediate difference in external behaviour, but it allows the client and plugins to only include object fields that are changing.

Prerequisite for the Organized flag changes that I'm working on.


bnkai:
I've gone through the code and used it for a couple of days doing scraping, stash-box tagging and addition removal of performers/studios/scens. It seems ok. I havent used with plugins or scripts though but i guess those would need to update to make better use of this PR.

PR #990, opened 2020-12-10
Fix scene gallery selection

InfiniteTF:
When the user selects a gallery that is not in the initial 20 galleries loaded, the name will get corrupted immediately after selecting. The reason is that the select uses the query data to get the name, and as soon as the gallery is selected the query is reset and no longer contains the selected gallery.

I've resolved this by explicitly saving the entire selected gallery and always showing that rather than using the queried data. The same applies to saved galleries that are not in the initial query, so I've changed it to pass in the entire gallery and use that rather than the queried data.


SpedNSFW (approved):
Works a treat.

PR #994, opened 2020-12-14
Fix integer overflow for scene size on 32bit systems

InfiniteTF:
Integers are 32bits on 32bit systems, for instance raspberry pis. This causes scene stats to overflow above 2GB and produce garbage numbers.

To output proper int64 we would need to manually set up a StatResultType model and matching resolver. Switching to float has the same effect however, and requires no boilerplate.


InfiniteTF:
Turns out file sizes are broken on ingest so we have to wipe all file sizes with a migration.

Ideally the scenes table should be rebuilt with integer size column so we don't have to convert int=>string=>float, but rebuilding scenes is a nightmare and there shouldn't be any risk of overflow anymore.

Tweeticoats:
can you bump the schema to version 17 as #988 has updated the schema to 16.

WithoutPants (approved):

PR #997, opened 2020-12-17
Data layer restructuring

WithoutPants:
This is a huge change and will only really see some proper testing by releasing it to the wild and patching up any issues that arise.

### Summary of changes
* separate the sqlite query builder implementation into a dedicated `sqlite` package, while defining the service interfaces in `models`.
* wrap **all** functions that interact with the database in a transaction, which is handled by the `TransactionManager` interface and is set on the manager singleton.
* clean up the data layer API to eliminate a dedicated joins query builder. 
* refactor sqlite implementation to reduce boilerplate and repeated code

### Transaction management

You will notice that all query-based API functions are now wrapped in `r.withReadTxn`. This function starts a transaction using the configured `TransactionManager` and executes the provided function, providing the `ReaderRepository` which contains functions to access the reader interfaces for all of the object types in the system. The return value of this function determines whether the transaction is committed or rolled back. For `withReadTxn`, this is largely academic, since it should not be possible to modify the database using a reader interface. 

Similarly, mutation functions which touch the database are wrapped in `withTxn`. This works similarly to `withReadTxn` except that the `ReaderWriter` interfaces are exposed, allowing changes to the database.

The manager singleton instance contains the reference to the `TransactionManager` instance, and this is passed onto the API instance, and can also be accessed using `manager.GetInstance().TxnManager`.

### Why?

The benefit of separating the data layer implementation from the definition is that we can now unit test the higher-level API functionality without needing to resort to integration tests. The data layer interfaces can be mocked, and the individual API functions can be unit tested. 

I chose to do this now since it was overdue and because I was looking into further data-layer changes, and the any further changes increase the workload required to perform this refactoring.

All references to the query builders themselves are now removed so that higher level API functions no longer need to be concerned with the implementation of the data layer - it is all accessed via `Reader` and `ReaderWriter` interfaces. Similarly, references to the `database` package have been removed unless necessary.

I am hoping that this restructuring leads to a cleaner structure that is easier to modify without the need for extensive upstream changes, like this very PR.


bnkai:
Something in the code breaks the parallel scanning functionality.
I get a `database is locked` error and scenes are left out / paths not updated when using more than 1 parallel tasks.

WithoutPants:
I've been investigating the locked database issue, and I've discovered that this issue actually exists on the current version. I think sqlite doesn't handle multithreads writing to the database.

bnkai:
I think sqlite doesnt deal well with concurrent writing , multiple reads and a single write is ok
This is the most related https://stackoverflow.com/questions/35804884/sqlite-concurrent-writing-performance issue i found that explains things, i'll try to test what is recommended there

The issue with the errors is the transaction within transaction did you trace that to your code?

WithoutPants:
Force pushed to rebase on existing branch.

I changed the read transactions to not actually start an sqlite transaction, so they are immediate. I don't believe this actually fixes any concurrency issues, since I think it was related to write operations. I may end up changing it back.

I've split up the transactions in the scan task so that it doesn't hold a transaction for too long. This does allow for possible data integrity issues, but this was pretty much how it was prior, so I'll leave it as such for now.

I've changed the sqlite write transactions so that only one write transaction will be allowed to be active at a time. Other transactions will wait a maximum of ten seconds before timing out with an error. A timeout of one second still caused timeout errors when scanning with 8 threads.

InfiniteTF:
Enabling WAL like @bnkai's link suggests, seems ideal. Faster write transactions, and concurrent read/write transactions. The only noticable impact seems to be some additional database files while the db is open.

bnkai:
@WithoutPants  i am trying out the WAL suggestions mentioned in the link i referenced
```diff
diff --git a/pkg/database/database.go b/pkg/database/database.go
index 231fe22..f8ee9f9 100644
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -68,14 +68,14 @@ func Initialize(databasePath string) bool {
 
 func open(databasePath string, disableForeignKeys bool) *sqlx.DB {
        // https://github.com/mattn/go-sqlite3
-       url := "file:" + databasePath
+       url := "file:" + databasePath + "?_journal=WAL"
        if !disableForeignKeys {
-               url += "?_fk=true"
+               url += "&_fk=true"
        }
 
        conn, err := sqlx.Open(sqlite3Driver, url)
-       conn.SetMaxOpenConns(25)
-       conn.SetMaxIdleConns(4)
+       conn.SetMaxOpenConns(1)
+       //      conn.SetMaxIdleConns(4)
        if err != nil {
                logger.Fatalf("db.Open(): %q\n", err)
        }

```
Using the above ( i dont think anything else needs to change)
-we have WAL journal mode ( the only thing you may notice is the extra 2 files in the db folder)
-we limit to 1 the open connections

Making the above changes and doing a scan with 12 parallel tasks or an export  seems to work ok for me
The problerm is i can't reproduce the issue you are having with 8 threads with the current mode ( probably because my db is on an nvme) so i can't tell how much of a difference this makes. In theory with WAL we should have no blocking as reading and writing can proceed concurrently (and thus no need to play with timeouts). Also despite  allowing only 1 connection everything seems to work ok as mentioned in the link i referenced.
Can you try that out?

bnkai:
Even with WAL enabled  i get some database locks when stress testing ( WAL+PR against  WAL+dev branch)

It shouldn't happen to the average user since i used 8, 12 parallel tasks when scanning.
I got 2, 5-6 errors with the database locked message all in cases with duplicate image entries ( batches of 300 video files and a few images).
Probably if i increased the number it would happen to the dev branch also so shouldnt be that big of an issue

Tried scraping, updating entries, browsing during scanning/generation all seem fine. Full import/export also look ok.
I will keep using this build for a couple of days  to make sure it is safe enough to release for public  usage/testing

WithoutPants:
Gallery create was a UI-side issue.

Tag create was not returning the tag object. Now fixed with a unit test.

bnkai:
Both issues are resolved for me.
I can't think of anything else that isn't working atm, it should be ok to merge for more extensive testing.

Just mentioning so that it can be referenced
With WAL enabled if you want to backup your db you  can either use 
`sqlite3 stash-go.sqlite "vacuum into 'backup.sqlite'"`
or do a plain vacuum (the database needs to be closed though so make sure to terminate stash if its running)
`sqlite3 stash-go.sqlite vacuum`
so that you end up with a single sqlite file.

PR #1008, opened 2020-12-20
Add gallery wall view, and new lightbox

### LINKED ISSUE #925: Image Carousel Rotation
**Describe the bug**
When getting to the last image of a page in galleries, it rotates back to the first one instead of continuing the gallery. This is of course mitigated by just increasing the page count to 1000, but it still is a design flaw that should be fixed in my opinion

**To Reproduce**
1. Galleries or images
2. Change to wall mode, reduce the number of results per page to a minimum
3. Click on image, click on right (with mouse, see #923) 
4. Do so until you reach the end of the result page in the carousel and click once more. You're now on the first image again

**Expected behavior**
It should continue with the first image on the next page

InfiniteTF:
Finding a decent lightbox library is surprisingly hard, so I ended up writing one instead. Also added a wall view for galleries with a less jittery wall layout than flexbin.

The lightbox supports has more or less the same feature set as the previous library we used, and also supports pagination for image lists. The logic around that is kinda hairy, but it's working quite well.

Resolves #925


ui/v2.5/src/hooks/Lightbox/Lightbox.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+            <div
+              className={cx(CLASSNAME_CAROUSEL, {
+                [CLASSNAME_INSTANT]: instantTransition,
+              })}
+              style={{ left: `${currentIndex * -100}vw` }}
+              ref={carouselRef}
+            >
+              {images.map((image) => (
+                <div className={CLASSNAME_IMAGE}>
```
WithoutPants, line 298: This needs a `key` prop set to stop the react warnings.

WithoutPants (changes requested):
Looks great. 

Suggest making the arrow buttons in the lightbox a little more subdued. If you can't manipulate the color directly, I'd suggest just lowering the opacity at least.

WithoutPants (approved):

PR #1022, opened 2020-12-25
Strip file extension from scene title when scanning

bnkai:


WithoutPants:
I don't really like this as default, uncontrolled behaviour. File extensions can be stripped post-scan using the filename parser. When imported, the filename is its only identity until it is otherwise renamed, and the extension is part of that identity. At the very least, this should be configurable behaviour.

bnkai:
Agreed on default , added it as optional in the UI (false by default)

For my use case (and maybe a few users from the discord channel) the extension part of the filename was not needed and a thing i had to remove after every scan. Going evey time to the filename parser only for this part wasnt optimal for me ( i even made  a script to batch remove extensions from the db for this reason). Having a properly named scene in many cases  and having the extension to remove was too bothersome. If i ever needed to check the file extension (or something else in the original filename) the file path was still there so no need for the extension to be part of the title (again for my use case).

PR #1036, opened 2020-12-31
extend resolutions

### LINKED ISSUE #1013: [Feature] Scene Resolutions
Pretty simple request. More listed resolutions. Currently Stash only seems to support 240p, 480p, 720p, 1080p, and 4k and only for landscape videos. It would be nice if there were more supported resolutions:

 - 144p (For the super old content)
 - 360p (likewise for older content)
 - 1440p
 - 1920p (VR content)
 - 2880p (5K - VR mostly)

These should all apply to portrait videos too. Just detect when a video is portrait (height > width) and flip the values before labelling.

SpedNSFW:
- Simplifies logic
 - Adds more options including 540p, 1440p, and resolutions common to VR such as 1920p
 - Supports vertical/portrait videos and images

Closes #1013

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -12,11 +12,19 @@ input FindFilterType {
 }
 
 enum ResolutionEnum {
+  "144p", VERY_LOW
   "240p", LOW
+  "360p", R360P
   "480p", STANDARD
+  "540p", WEB_HD
   "720p", STANDARD_HD
   "1080p", FULL_HD
+  "1440p", QUAD_HD
+  "1920p", VR_HD
   "4k", FOUR_K
+  "5k", FIVE_K
+  "6k", SIX_K
+  "8k", EIGHT_K
 }
 
 input PerformerFilterType {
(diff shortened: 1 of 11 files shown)


WithoutPants:
Is `5K` and `6K` necessary? Doesn't seem to match any standard formats that I can find.

These new resolutions need to be added to the filtering options as well.

SpedNSFW:
> Is `5K` and `6K` necessary? Doesn't seem to match any standard formats that I can find.

Those resolutions are found in VR content. If you're not happy supporting it, I'm fine with removing them, though I imagine it's only a couple of years until 2D content starts getting released in higher resolutions that 4K.

> These new resolutions need to be added to the filtering options as well.

Completely forgot that was a thing. Will do.

graphql/schema/types/filters.graphql
```diff
@@ -12,11 +12,19 @@ input FindFilterType {
 }
 
 enum ResolutionEnum {
+  "144p", VERY_LOW
   "240p", LOW
+  "360p", R360P
```
SpedNSFW, line 17: Couldn't think of a decent name for 360p. It was a standard resolution but never really had a label outside of "360p".

PR #1041, opened 2021-01-03
Add hardware acceleration for transcoding

ghost:
Supports vaapi (Intel) on Linux, NVENC (Nvidia) on Windows, and Quick Sync (Intel) on Windows. The stream transcoder tries all available configurations for the platform and falls back to software if none of them work. Should be largely seamless, but is going to require some testing.

For it to work on linux the user running stash needs to be added to the render and video group, if I remember correctly. Also, on my server the default Intel iHD driver doesn't support scaling for whatever reason so I had to switch to the (non-free) i965 driver by running `export LIBVA_DRIVER_NAME=i965` before starting stash. On windows no changes should be necessary.

Performance is significantly better on supported platforms, particularly for HEVC, but it's also very useful for scaling down 4k h264 to lower resolutions.

TODO:
* Add config for OS X.
* Check if raspberry pi is viable (I suspect hevc won't work well above 1080p30, but it's better than nothing if it works).
* Clean up code a bit.

AMD doesn't look like it's possible. FFMPEG supports the h264 AMF encoder, but not the AMF decoder or scaler, so performance is going to middling. I also don't have a Radeon to test with.
EDIT: AMD is supported through vaapi on Linux, but I have no way of checking if it works, unfortunately.


bnkai:
A few tests i did to test hardware the vaapi accelaration decode part  ( amd vaapi encode doesnt want to work fot the specific case  as mentioned above)

First the conclusion is that hw accel is making a difference even for the the decode part only for hevc files. (if the user can not/doesn't want to use the chromium all codec builds)

Hardware used 
AMD Ryzen 5 1600 ( ffmpeg commands limited to 4 threads )
AMD Vega 56 ( open source mesa drivers used)
ffmpeg from debian testing


samples (you can copy paste in a file and then wget -i file to download)
```
https://jell.yfish.us/media/jellyfish-20-mbps-hd-hevc-10bit.mkv
https://jell.yfish.us/media/jellyfish-140-mbps-4k-uhd-hevc-10bit.mkv
https://www.elecard.com/storage/video/Elecard_about_Tomsk_part3_HEVC_1080p.mp4
https://www.elecard.com/storage/video/Elecard_about_Tomsk_part3_HEVC_UHD.mp4
https://www.elecard.com/storage/video/Elecard_about_Tomsk_part2_HEVC_720p.mp4
```

commands used

hw decode and scale  (replace 720 with 1080) for 4k->1080p cases
```bash
ffmpeg -hide_banner -hwaccel vaapi -hwaccel_device /dev/dri/renderD128 -hwaccel_output_format vaapi -threads 4 -i sample_file -vf format=nv12|vaapi,hwupload,scale_vaapi=-8:720,hwdownload -pix_fmt yuv420p -c:v libx264 -preset ultrafast -tune zerolatency -crf 25 -threads 4 -movflags frag_keyframe+empty_moov -an -y -f mp4 test.mp4
```
hw decode and scale with a 10bit input ( notice the extra `format=p010` in the filters, while it isnt needed in sw it is for the hw case)
```bash
ffmpeg -hide_banner -hwaccel vaapi -hwaccel_device /dev/dri/renderD128 -hwaccel_output_format vaapi -threads 4 -i sample_file -vf format=nv12|vaapi,hwupload,scale_vaapi=-8:720,hwdownload,format=p010 -pix_fmt yuv420p -c:v libx264 -preset ultrafast -tune zerolatency -crf 25 -threads 4 -movflags frag_keyframe+empty_moov -an -y -f mp4 test.mp4
```
sw decode and scale
```bash
ffmpeg -hide_banner -threads 4 -i sample_file -vf scale=-8:720 -pix_fmt yuv420p -c:v libx264 -preset ultrafast -tune zerolatency -crf 25 -threads 4 -movflags frag_keyframe+empty_moov -an -y -f mp4 test_sw.mp4
```
plain hw decode / no scale
```bash
ffmpeg -hide_banner -hwaccel vaapi -hwaccel_device /dev/dri/renderD128 -hwaccel_output_format vaapi -threads 4 -i sample_file -vf format=nv12|vaapi,hwupload,hwdownload -pix_fmt yuv420p -c:v libx264 -preset ultrafast -tune zerolatency -crf 25 -threads 4 -movflags frag_keyframe+empty_moov -an -y -f mp4 test.mp4
```
plain hw 10 bit decode / no scale
```bash
ffmpeg -hide_banner -hwaccel vaapi -hwaccel_device /dev/dri/renderD128 -hwaccel_output_format vaapi -threads 4 -i sample_file -vf format=nv12|vaapi,hwupload,hwdownload,format=p010 -pix_fmt yuv420p -c:v libx264 -preset ultrafast -tune zerolatency -crf 25 -threads 4 -movflags frag_keyframe+empty_moov -an -y -f mp4 test.mp4
```
plain sw decode / no scale
```bash
ffmpeg -hide_banner -threads 4 -i sample_file -pix_fmt yuv420p -c:v libx264 -preset ultrafast -tune zerolatency -crf 25 -threads 4 -movflags frag_keyframe+empty_moov -an -y -f mp4 test_sw.mp4
```

the test cases
samples_table

the results
results_table

As shown above for cases where a `resize` is needed the difference in speed / cpu utilization for hw decode is big. In the case of the jelly fish 10bit 4k 140mbit sample it's a difference of streamable vs not.
In cases where no resize is needed the sw only decode is faster but uses more cpu.

While it would be handy to use hwaccel the implementation will be toublesome.
The  above cases are only for linux/amd  vega card/mesa open  gl drivers and DECODE only
If you take into account all the combinations i think the simplest we could do is make some fixed profiles and let the user select them.

- I tried a few ts hevc files but hw decode was not supported mainly because of 1. hevc profile uknown 2. hevc profile not supported
- 10 bit hw decode -> sw encode  needs an extra ffmpeg filter as mentioned ( vaapi / amd specific probably )
- I used the `-threads 4` twice one for ffmpeg and one for the libx264 settings as once wasnt enough ( utilization would go to 800% ). This sumulates usage for low cpu core count / average cpu  cases better
- didnt try any h264 hardware decode / scaling as most users would just  want to directly stream those files ( can try if needed)
- with `-preset ultrafast` and the tuning options used x264 can cope with almost all the cases tested, the resize part though does add a bottleneck especially if we limit the thread/cpu count

praul:
I am really looking forward to this feature. However, it does not work on my machine. 
It states "FMPEG hardware accelerated transcode failed." in log already before I click the play button.
I have added devices and permissions for /dev/dri just like for my jellyfin container.

What changes can I make to get more verbose error messages? I am really eager in getting this to work.

Also, it somehow spawns defunct ffmpegs:
image

Thank you for your work, much appreciated

PR #1042, opened 2021-01-04
Updating travis to build arm based containers as well as x86 containers.

Tweeticoats:
This uses https://github.com/docker/buildx to create a custom builder that is capable of building multiple containers and multiple architectures at once.
This uses binfmt in linux to run a qemu emulator allowing a linux computer to run arm programs so during build the container can start and run the arm version of curl etc during setup.

This also changes the os of the docker container from ubuntu to alpine and includes python, sqlite and python-requests to run most of the scrapers and plugins.
This reduces the size from 118mb to the following:
linux/amd64   83.33 MB
linux/arm/v7  62.32 MB
linux/arm/v6  61.31 MB
linux/arm64   76.2 MB


docker/ci/x86_64/Dockerfile
```diff
@@ -1,25 +1,22 @@
...
+FROM --platform=$BUILDPLATFORM alpine  AS prep
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
+RUN apk add --no-cache curl xz
 WORKDIR /
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-
-RUN curl --http1.1 -o /ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-amd64-static.tar.xz && \
+COPY stash-*  /
+RUN if [ "$TARGETPLATFORM" = "linux/arm/v6" ]; then FFMPEG_URL=https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-armel-static.tar.xz;BIN=stash-pi; elif [ "$TARGETPLATFORM" = "linux/arm/v7" ]; then FFMPEG_URL=https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-armhf-static.tar.xz;BIN=stash-linux-arm32v7; elif [ "$TARGETPLATFORM" = "linux/arm64" ]; then FFMPEG_URL=https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-arm64-static.tar.xz;BIN=stash-linux-arm64v8;elif [ "$TARGETPLATFORM" = "linux/amd64" ]; then FFMPEG_URL=https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-amd64-static.tar.xz;BIN=stash-linux;  fi; \
```
WithoutPants, line 8: You're splitting this command onto the lines below, so split this if/elif line into a more readable format.

Issue #1009, opened 2020-12-21
[Bug Report] Cannot assign galleries to scenes


AshenErn:
After a certain number of galleries, I think after 40
It is no longer possible to choose new galleries in a scene because they are not searched for anymore, only the old ones

AshenErn:
is there any way around this while it is not fixed?

PR #1057, opened 2021-01-10
Decouple galleries from scenes

### LINKED ISSUE #1054: [Feature] Improve and Expand Scene/Gallery Associations
Currently there is no way to assign multiple galleries to a single scene or multiple scenes to a single gallery. Some studios like to separate softcore and hardcore galleries for a single scene. I also have an original copy and a remastered copy of a scene that share the same official gallery. If I try to link the same gallery to a new scene, the association is removed from the old one.
Additionally, the only link to an associated scene while under the Galleries tab is the icon that appears in the Grid view. There is no link listed anywhere when viewing an individual gallery. There is also no way to associate a scene while viewing a gallery.

It would be preferable to be able to choose multiple galleries from a list while editing a scene and vice versa. It sounds like this will require changing how these associations work behind the scenes somehow.

Adding a single gallery to multiple scenes could be accomplished by simply duplicating the gallery. Adding multiple galleries to a single scene could be done by manually combining the galleries as well. However, I think it should be preferable to avoid modifying/duplicating galleries if possible.

InfiniteTF:
Migration adds a scenes_galleries table that supports M:N relations. There's a ton of boilerplate associated with this, but the end product is largely the same as the other entities, with the exception that you can set entities from both sides.

* Change SceneGallery select to a generic multi GallerySelect. Removed the initial 40 results and turned it into a pure search field since there have been a number of users confused by this. Also added title to the fields searched which was missing.
* Added a galleries tab to scenes if it has more than one gallery. If it's just one gallery it's the same.
* Added a SceneSelect field to edit galleries. Inverse of the GallerySelect.
* Added a scenes tab to gallery.
* Removed `ValidGalleriesForScene` query since it's not used anymore.
* Changed performer height in gallery view so aspect ratio matches performer lists.

Note that I haven't tested the import/export stuff yet, but I copied the scenePerformer implementation and the tests are passing.

Resolves #1054, #1009

### DIFF
--- a/graphql/documents/data/gallery-slim.graphql
+++ b/graphql/documents/data/gallery-slim.graphql
@@ -21,7 +21,7 @@ fragment GallerySlimData on Gallery {
   performers {
     ...PerformerData
   }
-  scene {
+  scenes {
     id
     title
     path
--- a/graphql/documents/data/gallery.graphql
+++ b/graphql/documents/data/gallery.graphql
@@ -25,8 +25,6 @@ fragment GalleryData on Gallery {
     ...PerformerData
   }
   scenes {
-    id
-    title
-    path
+    ...SceneData
   }
 }
--- a/graphql/documents/data/gallery.graphql
+++ b/graphql/documents/data/gallery.graphql
@@ -24,7 +24,7 @@ fragment GalleryData on Gallery {
   performers {
     ...PerformerData
   }
-  scene {
+  scenes {
     id
     title
     path
--- a/graphql/documents/data/scene-slim.graphql
+++ b/graphql/documents/data/scene-slim.graphql
@@ -37,7 +37,7 @@ fragment SlimSceneData on Scene {
     seconds
   }
 
-  gallery {
+  galleries {
     id
     path
     title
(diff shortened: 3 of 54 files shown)


bnkai:
The is missing scene filter in galleries is broken
```
Error executing count query with SQL: SELECT COUNT(*) as count FROM (SELECT DISTINCT galleries.id FROM galleries left join performers_galleries as performers_join on performers_join.gallery_id = galleries.id left join scenes_galleries as scenes_join on scenes_join.gallery_id = galleries.id left join studios as studio on studio.id = galleries.studio_id left join galleries_tags as tags_join on tags_join.gallery_id = galleries.id left join galleries_images as images_join on images_join.gallery_id = galleries.id left join images on images_join.image_id = images.id WHERE galleries.scene IS NULL GROUP BY galleries.id ) as temp, args: [], error: no such column: galleries.scene
```

InfiniteTF:
@bnkai Thanks, missed that, it's been fixed now. I had to tweak the migration so you'll have to rerun it.

bnkai:
Everything fixed/seems to work now.
I even tried the full import/export and it looks ok.

I noticed an oversight from enabling WAL ( didnt occur to me when testing that) when doing a full import.
When we reset the DB to do a full import we only delete (Os.remove) the sqlite file and not the 2 new files (-shm -wal)

WithoutPants (approved):

PR #1069, opened 2021-01-20
Add backup database functionality

WithoutPants:
Adds two buttons to the tasks settings page. The first backs up to a file in the same directory as the sqlite file, using the same file naming format as for migrations. The second button backs up the database and downloads the resulting file.

image

### DIFF
--- a/pkg/api/resolver_mutation_metadata.go
+++ b/pkg/api/resolver_mutation_metadata.go
@@ -125,7 +125,9 @@ func (r *mutationResolver) BackupDatabase(ctx context.Context, input models.Back
 
 		ret := baseURL + "/downloads/" + downloadHash + "/" + database.DatabaseBackupPath()
 		return &ret, nil
+	} else {
+		logger.Infof("Successfully backed up database to: %s", backupPath)
 	}
-	
+
 	return nil, nil
 }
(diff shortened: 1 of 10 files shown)


pkg/api/resolver_mutation_metadata.go
```diff
@@ -89,3 +93,41 @@ func (r *mutationResolver) JobStatus(ctx context.Context) (*models.MetadataUpdat
...
+		return nil, err
+	}
+
+	if download {
+		downloadHash := mgr.DownloadStore.RegisterFile(backupPath, "", false)
+		logger.Debugf("Generated backup file %s with hash %s", backupPath, downloadHash)
+
+		baseURL, _ := ctx.Value(BaseURLCtxKey).(string)
+
+		ret := baseURL + "/downloads/" + downloadHash + "/" + database.DatabaseBackupPath()
```
bnkai, line 126: While the plain  backup works the download one doesnt seem to.
Something seems wrong with the translation of the filename to the download path
```bash
INFO[0775] Backing up database into: /opt/nvme/stash/meta/downloads/backup199153937.sqlite 
DEBU[0780] Generated backup file /opt/nvme/stash/meta/downloads/backup199153937.sqlite with hash 9cb3394b 
```
the backup is correctly stored in my downloads folder but 
`http://localhost:9998/downloads/9cb3394b//home/username/.stash/stash-go.sqlite.17.20210120_144604`
is the link that is generated and 404s. 

i think in
`ret := baseURL + "/downloads/" + downloadHash + "/" + database.DatabaseBackupPath()`
the  `database.DatabaseBackupPath()` feels a little off, shouldnt it be `backupPath` instead?

PR #1073, opened 2021-01-22
Force reload of custom css if it's updated

InfiniteTF:
Seems like the only way to get the browser to reload the css is to bust the cache and then reload the page. Not a big fan of reloading like that, but I figure since it's a setting that isn't modified frequently it shouldn't be a huge issue.

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v050.md
+++ b/ui/v2.5/src/components/Changelog/versions/v050.md
@@ -7,6 +7,7 @@
 * Allow configuration of visible navbar items.
 
 ### 🎨 Improvements
+* Refresh UI when changing custom CSS options.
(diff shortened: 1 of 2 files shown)


bnkai:
Works ok except of the case of ticking/unticking the `Custom CSS enabled`.
Enabling / disabling the custom css should probably count as updating the CSS.

InfiniteTF:
Good point, I hadn't considered that. Updated.

bnkai:
Yeah looks fine.
As a sidenote the first time i did the upgrade/test for this version i still had to do a force refresh to get the UI updated.
From then on though a simple tick/untick/change of the CSS and save work fine.

WithoutPants (approved):

WithoutPants:
Worked from the start for me.

PR #1079, opened 2021-01-26
Make file upload limits configurable

InfiniteTF:
Sets the default to 1GB. ~~I'm not entirely sure what the impact of maxuploadmemory is. According to the gqlgen documentation it should spill to disk if the upload requires more memory than is allowed, but I didn't have much luck when I didn't set it.~~


pkg/manager/config/config.go
```diff
@@ -579,6 +583,25 @@ func GetLogAccess() bool {
 	return ret
 }
 
+// Max allowed graphql upload size in megabytes
+func GetMaxUploadSize() int64 {
+	ret := int64(200)
+	if viper.IsSet(MaxUploadSize) {
+		ret = viper.GetInt64(MaxUploadSize)
+	}
+	return ret * 1000000
```
bnkai, line 592: to be accurate if it is megabytes (MB) it should be `1048576` (1024*1024) instead of `1000000`
sa,e with memory below

InfiniteTF: Good point, fixed.

PR #1082, opened 2021-01-29
String regex filter criteria and selective autotag

### LINKED ISSUE #865: [Feature] Allow RegEx in search fields
Title should be self explanatory.

An example would be if I want to find all performers whose name starts with the letter B, I could enter `^B` into the field.

### LINKED ISSUE #886: [Feature] Partial scan/autotag based on path parameter

WithoutPants:
Functionality changes:
* added regex matches/not matches string criteria (see known issue)
* added paths input to auto tag task
* made 0 value for `per_page` in `FindFilterType` to mean get all results (no changes to UI to support this)

Bit of refactoring involved. Removed the scene query by regex functions in favour of using the `Query` interface with new matches regex filter criteria.

Known issues:

The regex filtering does not work for images in zips due to the null character separator in the path field. This will need to be migrated to use a different control character separator. Until then, regex filters on path will not work correctly (the function receives only the zip part of the path, not the file).

~~The selective auto tag does not filter out scenes outside the provided paths in the database query. It performs a post-filter to eliminate scenes. This is by necessity until we alter the filtering engine to allow multiple criteria on the same field. As a result, selective auto-tagging is unlikely to see much performance benefit compared to running a full auto-tag.~~

Closes #886 
Closes #865


bnkai:
For cases like this where we only need to filter by a criterion that is avaliable in a single table isn't it a waste to use the Query that uses 6-7 joins? Instead of using scene query wouldn't it be more efficient to make a simple customised query ?
eg `SELECT * from scenes WHERE path LIKE "/path/goes/here/%" OR path LIKE  "/path2/% ORDER BY path`
I was thinking of how to implement the same filtering for a selective clean feature
A function that constructs multiple path clauses ( `path LIKE "escaped/path%"` ) and string joins them using `OR` and then executes the query was what i was planning to implement.

WithoutPants:
> For cases like this where we only need to filter by a criterion that is avaliable in a single table isn't it a waste to use the Query that uses 6-7 joins? 

That's more an issue for the Query function than anything. It should include all those joins unless they are actually used.

> Instead of using scene query wouldn't it be more efficient to make a simple customised query ?
> eg `SELECT * from scenes WHERE path LIKE "/path/goes/here/%" OR path LIKE "/path2/% ORDER BY path`
> I was thinking of how to implement the same filtering for a selective clean feature
> A function that constructs multiple path clauses ( `path LIKE "escaped/path%"` ) and string joins them using `OR` and then executes the query was what i was planning to implement.

I think improving the Query function to support AND/OR filters (and minimise superfluous joins) would suit better than adding/keeping redundant query functions. As it is, we have a bunch of `FindBy` methods that should probably be consolidated in the `Query` function instead.

bnkai:
Regex expressions are not validated/checked before being executed
Trying a simple  not valid regex eg `*.mp4` fills the screen with the error message.

Some characters cannot be used for the regex  eg the `+`  turns to ` ` (a plain ws char) and the `&`  breaks the page

PR #1103, opened 2021-02-03
fix check version

bnkai:
On rare occasions `git rev-parse --short HEAD` generates a longer than the default 7 characters short hash
I adjusted the version check code to take that into account

I added also an extra `debug` print for  statistical reasons in case this occurs
eg
```bash
DEBU[0000] Short Hash:0dd2e269 Length:8 differs from default 7
INFO[0000] Version: (0dd2e269) is already the latest released.
```

to force the issue for testing you can either export the needed > 7 chars `GITHASH`  or just  do a `GITHASH=0dd2e269 make pre-ui && GITHASH=0dd2e269 make release`
`0dd2e269` is used to match the current (as of testing) dev build/commit hash

### DIFF
--- a/pkg/api/check_version.go
+++ b/pkg/api/check_version.go
@@ -251,7 +251,6 @@ func getShaFromTags(shortHash bool, name string) string {
 					last = len(gitShort)
 					logger.Debugf("Short Hash:%s Length:%d differs from default %d\n", gitShort, last, defaultSHLength)
 				}
-				// default length of git short hash
 				return tag.Commit.Sha[0:last]
 			}
 
--- a/pkg/api/check_version.go
+++ b/pkg/api/check_version.go
@@ -18,6 +18,7 @@ const apiReleases string = "https://api.github.com/repos/stashapp/stash/releases
 const apiTags string = "https://api.github.com/repos/stashapp/stash/tags"
 const apiAcceptHeader string = "application/vnd.github.v3+json"
 const developmentTag string = "latest_develop"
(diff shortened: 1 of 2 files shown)


WithoutPants:
Is there much point doing all the length checking and logging? Can't we just use the length of the current version short hash as the end point for the sha substring (with some sanity checking I guess)?

bnkai:
I have removed the debug prints and tweaked a  bit the sanity check.
For the checks we need at least one check to make sure no out of bounds exception exists ( check index < legth of sliced string) 
I 've included the second one as a sanity one ( default length 7 ) as this is the default minimal  that git uses and that happens to be the same (fixed this time) that github uses.Using that makes sure we have a fallback with the 7 first digits of the hash.

what remains is like this
```golang
        if shortHash {
                last := defaultSHLength 
                _, gitShort, _ := GetVersion() 
                if len(gitShort) > last && len(gitShort) < shaLength { 
                        last = len(gitShort)
                }
                return release.Target_commitish[0:last]
        }

```
setting by default to the length of the gitShort hash would give something like that
```golang
if shortHash {
   last := len(gitShort)
  if last < shaLength { // index check
            if last < defaultSha { //sanity check
            last = defaultSHLength
            }
   return release.Target_commitish[0:last]
} 
```
If we partially skip the sanity check
we can just do
```golang
if shortHash {
   last := len(gitShort)
  if last > shaLength { // index check
               last = defaultSHLength 
                }
   return release.Target_commitish[0:last]
} 
```
which imo isnt a big difference
I would prefer one of the two first options or some other more efficient if possible (i couldnt think of anything else)

PR #1105, opened 2021-02-04
Gallery view persistence

InfiniteTF:
Adds display persistence to the gallery view by adding a `VIEW` persistence mode which only reads and writes the displayMode, but leaves the rest of the persisted state intact.

Also changes the gallery filter to only show folder name if the gallery doesn't have a title, rather than the entire path. This matches how we now display it elsewhere.

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v060.md
+++ b/ui/v2.5/src/components/Changelog/versions/v060.md
@@ -1,4 +1,5 @@
 ### 🎨 Improvements
+* Remember gallery images view mode.
 * Add option to skip checking of insecure SSL certificates when scraping.
 * Auto-play video previews on mobile devices.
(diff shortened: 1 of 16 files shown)


WithoutPants:
I don't really feel that this is a good fix for what I imagine is the original issue (gallery images view state is not persisted), since setting the view in the gallery images view overwrites the view in the top-level Images view. Gallery images view should have its own entry in the view state.

InfiniteTF:
I've added a prop to set a custom key for the gallery images so they're stored separately from the main image view. There's no getting around the need for a limited state persistance however, since sublists don't display the same datasets, so persisting filters and pagination would lead to empty pages.

WithoutPants (approved):

PR #1108, opened 2021-02-06
[WIP] ETA (time remaining for progress bar)

reduych:
This should show a time remaining under the progress bar:

image

Work in progress! Doesn't work quite as expected...

The aim is to have a `started` variable (ideally in Go backend) that can be used to calculate the remaining time (based on current progress). 

This is my first time working in Go, but I've worked on React loads. 

So my first approach was to simply store a `started` variable within the React app itself. But obviously that wouldn't work when page is refreshed or something. It had to be in Go app. That's where I'm struggling a bit... 

Am I going in the right direction? I've tried to make the `started` variable in Go but..

1. Not sure if it's in all the right places
2. Can't get it through the GraphQL API call over to the frontend. 

Any help?

### DIFF
--- a/ui/v2.5/src/components/Settings/SettingsTasksPanel/SettingsTasksPanel.tsx
+++ b/ui/v2.5/src/components/Settings/SettingsTasksPanel/SettingsTasksPanel.tsx
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from "react";
 import { Button, Form, ProgressBar } from "react-bootstrap";
+import ms from "ms";
 import { Link } from "react-router-dom";
 import {
   useJobStatus,
@@ -56,6 +57,7 @@ export const SettingsTasksPanel: React.FC = () => {
   const [autoTagPerformers, setAutoTagPerformers] = useState<boolean>(true);
   const [autoTagStudios, setAutoTagStudios] = useState<boolean>(true);
   const [autoTagTags, setAutoTagTags] = useState<boolean>(true);
+  const [started, setStarted] = useState<number>(+new Date());
 
   const jobStatus = useJobStatus();
   const metadataUpdate = useMetadataUpdate();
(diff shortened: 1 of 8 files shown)


ui/v2.5/src/components/Settings/SettingsTasksPanel/SettingsTasksPanel.tsx
```diff
@@ -91,8 +93,11 @@ export const SettingsTasksPanel: React.FC = () => {
     if (jobStatus?.data?.jobStatus) {
       setStatus(statusToText(jobStatus.data.jobStatus.status));
       const newProgress = jobStatus.data.jobStatus.progress;
+      console.log('jobStatus:', jobStatus);
```
reduych, line 96: Can't seem to get the `started` variable here

InfiniteTF: You need to add the started field to `graphql/documents/queries/settings/metadata.graphql`. All frontend queries are defined in the `graphql/documents` folder. When you then run `make generate`, or `yarn gqlgen` in the ui folder, this populates the types.

InfiniteTF:
This is a great idea! One note on implementation however.

When the task is first started stash will enumerate all files in the paths to figure which files needs to be processed. The end result could be anything between 0 files or all of them, it can also take a few minutes. Crucially, this enumeration process is not representative of the time the actual task will take, so the started timestamp should not be set until counting is done, otherwise it will throw off estimates.

bnkai:
A couple of issues that make the ETA a little difficult to calculate

- The scan or generate task dont always count the needed jobs to be done. There is a timeout of 90 seconds that they give up and we have an infinite status (no percentage also when this occurs). This was mainly to account for cases with remote mounted fs like google drive mounts where the calculations would take too much time. The elapsed time could probably be shown instead of the ETA in this case.
- The duration of each task during a scan/generate varies a lot.
 eg during scan if the file already exists (nothing needs to be done) or if a file is renamed(only path is updated) or if the file is new (then the time to finish is exponentially larger). The progress as it is used now is not that useful for ETA calculation due to this reason.  A possible workaround would be to calculate some kind of rolling average of the progress over a fixed interval and use that for the ETA calculation. That could probably be easier implemented in the backend though.

WithoutPants:
Closing due to inactivity.

PR #1112, opened 2021-02-08
Fix scene edit state resetting when scene is updated

### LINKED ISSUE #1110: [Bug Report] Setting Organised State Reverts Data to Last Saved State
**Describe the bug**
Clicking on the organised button for a scene will revert the scene data back to the state it was at its last save.

**To Reproduce**
1. Go to a scene and scrape (or just enter some data)
2. Click on the organised button
3. All scraped/entered data will be gone.

**Expected behavior**
Any data entered should not be affected by the change of Organised state.

InfiniteTF:
Replaces useEffect form initialization with initialization directly in the state variables, so state is only ever initialized once. This avoids the issue where updating organized/ocounter state would reinitialize the form.

Resolves #1110.

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v050.md
+++ b/ui/v2.5/src/components/Changelog/versions/v050.md
@@ -29,10 +29,11 @@
 * Support configurable number of threads for scanning and generation.
 
 ### 🐛 Bug fixes
+* Fix edit data being lost when clicking the O-Counter, Organized or Favorite buttons.
 * Exclude media in `generated` directory from the library.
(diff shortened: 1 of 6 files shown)


WithoutPants:
This works, but could you also cover the Image, Gallery, and Performer pages which have the same bug?

InfiniteTF:
I've fixed the gallery and image pages. The performer page really needs a refactor, but I fixed the favorite button resetting state by removing the loading state, which I think is a better experience anyway.

PR #1118, opened 2021-02-10
Exclude media in generated directory

### LINKED ISSUE #1102: [Bug Report] Stash folders should be excluded from scan
**Describe the bug**
Stash data folders like generated and cache are included in the scan if the parent folder is a content folder, causing recursive behavior.

**Expected behavior**
When scanning, folders that include the path of any of stash data folders should be skipped. Should be an easy fix.

PS: haven't checked in code that this is the case, but I'm assuming this is what's happening for me.

WithoutPants:
Resolves #1102 

Excludes any media within the configured `generated` directory from being added when scanning, and removes any such items during clean.

### DIFF
--- a/pkg/manager/task_clean.go
+++ b/pkg/manager/task_clean.go
@@ -41,7 +41,9 @@ func (t *CleanTask) shouldClean(path string) bool {
 	// use image.FileExists for zip file checking
 	fileExists := image.FileExists(path)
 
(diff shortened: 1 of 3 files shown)


bnkai:
Looks and works ok.
 I didn't even know about `filepath.SkipDir`, that makes it a very `clean` fix.
Should this be included to 0.5 instead ?

bnkai (approved):

PR #1120, opened 2021-02-12
Skip insecure certificates check when scraping

bnkai:
The http client we use to do the scraping and download images is a little picky with certificates. That causes an error when trying to scrape from sites that have certificate issues.
example  `https://www.sslshopper.com/ssl-checker.html#hostname=https://dreamtranny.com`

If needed i can add this as an option to the scraper config section instead of hardcoding it.


WithoutPants:
Given that this is a security matter, I think it would be better as a configurable flag.

PR #1122, opened 2021-02-13
Random strings for cookie values

SpedNSFW:
If a scraper requires cookies and any random value can be used, we can now set the value to `"random"` for a random string to be generated and used.


bnkai:
I would prefer not to introduce the functionality like this. 
It would be better suited as an extra attribute/parameter in the scraping/cookies section like `value_random: 20` for example
It should also support the CDP part not only the native.

bnkai (approved):

PR #1130, opened 2021-02-18
Add full timestamp for console/file logging

### LINKED ISSUE #1128: [Feature] Add timestamp to logfile on disk
The logfile on disk doesn't contain a timestamp.

bnkai:
<del>Adds a new config option that enables by default a full timestamp `[2006-01-02 15:04:05] ` when logging to console or the logfile.
The current `[seconds]` from program start is left as an option for users that may not want more prints to their logfile/console.
</del>
Console and logfile prints now use a full timestamp `[2006-01-02 15:04:05]` instead of the old `[seconds]` from program start one

closes #1128

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,155 @@
+name: Build
+
+on:
+  push:
+    branches: [ develop, master ]
+  pull_request:
+    branches: [ develop ]
+  release:
+    types: [ published ]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Checkout
+      run: git fetch --prune --unshallow --tags
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.13
+
+    - name: Set up Node
+      uses: actions/setup-node@v2
+      with:
+        node-version: '12'
+    
+    - name: Cache node modules
+      uses: actions/cache@v2
+      env:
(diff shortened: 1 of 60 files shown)


InfiniteTF:
Seems like a pretty sensible default to me. Do we need a setting for it?

bnkai:
I would also prefer not to add an option and use that as a default (less options is sometimes better). I wan't sure if anyone has issues with that (removing the option is a few seconds :-) ).

WithoutPants:
I believe this will make logging consistent across platforms. On Windows, the old logging looked like this:
```
time="2021-03-01T11:55:24+11:00" level=info msg="using config file: D:\\dev\\stash_data\\sfw\\config.yml"
time="2021-03-01T11:55:24+11:00" level=debug msg="Reading plugin configs from D:\\dev\\stash_data\\sfw\\plugins"
time="2021-03-01T11:55:24+11:00" level=info msg="stash is listening on 0.0.0.0:9999"
time="2021-03-01T11:55:24+11:00" level=info msg="stash is running at http://localhost:9999/"
```

It now looks like this:
```
DEBU[2021-03-01 11:58:28] Reading plugin configs from D:\dev\stash_data\sfw\plugins 
DEBU[2021-03-01 11:58:28] Reading scraper configs from D:\dev\stash_data\sfw\scrapers 
INFO[2021-03-01 11:58:28] stash is listening on 0.0.0.0:9999
INFO[2021-03-01 11:58:28] stash is running at http://localhost:9999/
```

WithoutPants (approved):

PR #1132, opened 2021-02-19
Performer tags

### LINKED ISSUE #120: Performer tags
### Performer tags
In a nutshell, it would let the user add tags to performers. 
- Performer tags could act as a filter (as demonstrated in the attached concept image below)
- Performer tags would be another way of implementing "performer details" per #53 

(IIRC this might be another feature request that wasn't ported from the Ruby version, or I'm confusing it with Pornganizer functionality, or both.)

performer tags
(the tags in the image are repeating because I'm lazy af)

#### Concept image
The above front end ui #108  concept shows how performer tags might be implemented in a container on every url containing `/performers*`.
- Clicking a performer tag would have the equivalent functionality of filtering the performer list
- I would suggest display the performer tag container on the mobile version of Stash too. xHamster has something similar, and it lets you swipe through the tag container left or right to scroll through the list of tags.

#### Auto populate scene tags with performer tags
- Performer tags would automatically populate the scene tags of any scene the performer is added to. This might be an efficient way to add a mass of tags to any given scene.
- Some users might just want to use performer tags as a way to filter their performer list. Therefore, add the option to enable/disable auto tag population in `/settings?tab=tags` > `Performer tags` > `Populate scene with performer tags: Enable/disable`.
- Let the user specify tag exceptions that do not get auto populated in scenes. There would be some tags, like a performers country, that would probably be irrelevant to the scene. In `/settings?tab=tags` > `Performer tags` > `Exceptions (add any performer tags that you do not want to auto populate scenes): {input tag here}`.

performer tags settings

#### Usage
Performer tags should probably be innate descriptions of the performer, as in qualities about them that don't vary from scene to scene in your Stash library.

For example, if you have scenes of a blonde performer that never (or only rarely) changes their hair colour, then _blonde_ would be a good performer tag if that's something you care about. However, if they do change their hair color often, a performer tag based on the colour of their hair would be an inaccurate performer tag. But this is just a guideline of how you could use performer tags, obviously the use-case could vary.

Real performer tag examples would be things like: _amateur_; _bbw_; _pawg_; _mature_; _Germany_; _Czech_; _hispanic_; _black_; _freckles_; _bdsm_ (if the performer is exclusive to that genre); _large breasts_; _ginger_

#### Relationship with scraping 
Would it be a good idea to automatically generate performer tags based on scraped data?
- Specifically I'm thinking of things like _ethnicity_ and _country_.
- _Fake tits_ is an iffy performer tag as it's a quality that isn't necessarily a permanent quality in the performers create, and could create inaccurate information.
- Of course, there could be a bunch of tick boxes in a `/settings?tab=tags` page that lets the user select which scrapped data to add as a performer tag, but it's more complexity for the devs.
- Some scraped tags would be so redundant I wouldn't even want to see them, eg. If _performer country = United States_ or if _ethnicity = white_. Most performers for many of us would be from white Americans, so we wouldn't ever care about filtering or seeing that.
- I'm proposing an exceptions option in `/settings?tab=interface` to hide a user provided list of performer tags. This could function exactly like adding or removing scene tags to an input box from the scene edit panel

### LINKED ISSUE #1148: [Feature] [Cosmetic] Add performer scene count icon to performer cards
With the introduction of performer tags and their presentation on performer cards, I think the performer cards UI should be made to be consistent with the scene cards. That is, displaying metadata on the bottom of the card using a svg icon and corresponding integer.

- Move the line `Stars in n scenes` adjacent to the performer tags icon.
- Change the string `Stars in n scenes` to `[icon] n`, where `n` is the total count of scenes and `[icon]` is an svg that corresponds to scenes.

**Additional context**
- Beyond the scope of this issue, but I think the same should be applied to other grid view cards, studio, images, etc.

WithoutPants:
Adds tags to performers.

Tags may be set in the create/edit performer page. Scraping of performer tags is now available using the `Tags` field. Scraped tags may be created in the same way as scraped scene tags.

Tags are visible in performer cards in the same way as scene tags, and performers may be filtered by their tags in the same way as scenes.

Scenes, Images and Galleries may be filtered by performer tags. This is distinct from the tags field, which continues to filter only on [scene/image/gallery] tags.

Tag card now has a performer count popover.

Possible extensions of this:
* add a filter criterion that is the union of [scene/image/gallery] tags and performer tags.
* add an option to include performer tags in the [scene/image/gallery] tag popovers/count.
* add a performer tags field to the scene details panel.
* add an option to separate performer tags from scene/image/gallery tags, so that these do not appear in the tag drop downs.

These extensions can be raised as separate issues as needed, according to usage.

Resolves #120 
Resolves #1148


WithoutPants:
Feedback from Discord:

Two observations after building a performer tags release:

1. It would be great if I could mass-edit these performer tags
2. When editing the tags on the performers page, after I click on tag field something seems to cut off the field with all the available tags that pop up.
Either only three rows are displayed in the popup-menu (when the performer has a stash-id) or only 1.9 rows which isn't a whole lot when you have hundreds of tags.
After I add 
```
.css-2b097c-container {margin-bottom: 50px}
```
 to the Custom CSS more rows are displayed

Jimbo4041:
In addition to that there are two more things that I just noticed:

3. The addition of a performer register card when you click on a tag on the Tag page. You already have Scenes, Images & Markers, so it only seems natural to also add Performers.

And 4. being able to sort tags by performer_count on the Tags page.

GsMumbo:
Does this, or could it also include performer star ratings?

WithoutPants:
Performer star ratings are a separate issue and will not be included in this particular PR.

WithoutPants:
Thanks for the feedback. I've added the following improvements:
* performer tag dropdown no longer drops off the edge of the screen
* added a bulk performer edit dialog with the option to set tags and the favourite flag
* added a Performers tab to the Tag page
* added count filters and sort options for Images, Galleries and Performers in the Tag filter
* changed the scene count in the Performer card to an icon per other cards

bnkai (approved):

bnkai:
Everything seems fine except from a small regression with the performers scraping.
When you use the babepedia / freeones scraper to search for a performer and then select the performer from the list you get an error `code 422` whan doing the scraper. Using the URL directly works fine ( the URLs from the search seem also to be fine)

WithoutPants:
> Everything seems fine except from a small regression with the performers scraping.
> When you use the babepedia / freeones scraper to search for a performer and then select the performer from the list you get an error `code 422` when doing the scrape. Using the URL directly works fine ( the URLs from the search seem also to be fine)

Thanks. `ScrapedPerformerInput` does not accept tags, so removed from the fragment that the UI sends. Should be fixed now.

Jimbo4041:
A bug(?) and a request:

When creating a new performer, tags aren't getting saved, only after the performer already exists.


I think it would be helpful if there would be an additional field where you can select what type the tag is -- is it a performer tag, is it a scene/image tag, or is it both.
- Only tags which are selected as performer tags or both are gonna be listed when adding a performer tag. 
- On the Tag page you can also filter to only show performer tags, scene tags, both or where that info is missing.

PR #1144, opened 2021-02-23
Detect cover images in subdirectories

WithoutPants:
Fixes issue where `cover.jpg` image files would not be detected as a cover image if the file's path was in a subdirectory.


bnkai:
Looks ok.
A small regression probably from the image transition  though.
`If an filename of an image in the gallery zip file ends with cover.jpg` that was till we used zip files only(and is still in the help section) , afterwards it changed to `cover.jpg` . Either the documentation/help section should change or a `strings.HasSuffix` should be used instead of `==`

WithoutPants:
Good pickup. Changed so that it's checks for suffix instead of equality.

bnkai:
Works ok. 
The `filepath.Base(fn)` shouldnt be needed now that the HasSuffix is used. (works without it too)

PR #1149, opened 2021-02-24
Fix hacky CSS for scene card video preview

SpedNSFW:
The positioning for previews was rather bad and extremely hacky. Using CSS visibility is a more appropriate solution that won't cause issues.

### DIFF
--- a/ui/v2.5/src/components/Scenes/styles.scss
+++ b/ui/v2.5/src/components/Scenes/styles.scss
@@ -196,6 +196,9 @@ textarea.scene-description {
 
     &-video {
       position: absolute;
+      top: -9999px;
+      transition: top 0s;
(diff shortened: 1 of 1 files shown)


SpedNSFW:
Explanation copied from Discord:

> Basically, scene previews are moved 9999px above the scene image. Because there's an overflow: hidden, that preview is not shown. In a custom CSS theme, if you override the overflow, previews will appear.
> That was an issue I was having at the time I made the PR.
> The PR just makes the previews invisible in addition to moving them 9999px above.

WithoutPants:
With this change in place, the hover preview functionality doesn't work at all for me on Chrome.

WithoutPants:
Closing due to inactivity.

PR #1156, opened 2021-02-27
Find correct python executable

SpedNSFW:
For script scrapers using python, both python and python3 are valid depending on the OS and running environment. To save users from having any issues, this change will find the correct executable for them.

### DIFF
--- a/pkg/scraper/script.go
+++ b/pkg/scraper/script.go
@@ -193,12 +193,10 @@ func (s *scriptScraper) scrapeMovieByURL(url string) (*models.ScrapedMovie, erro
 }
 
 func findPythonExecutable() (string, error) {
-	cmd := exec.Command("python3", "--version")
-	_, err := cmd.Output()
+	_, err := exec.LookPath("python3")
 
 	if err != nil {
-		cmd := exec.Command("python", "--version")
-		_, err := cmd.Output()
+		_, err = exec.LookPath("python")
 
 		if err != nil {
 			return "", err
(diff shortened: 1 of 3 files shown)


pkg/scraper/script.go
```diff
@@ -30,6 +30,13 @@ func newScriptScraper(scraper scraperTypeConfig, config config, globalConfig Glo
 func (s *scriptScraper) runScraperScript(inString string, out interface{}) error {
 	command := s.scraper.Script
 
+	if command[0] == "python" {
```
bnkai, line 33: this should work both ways (some users have already edited the scraper's yaml file to use python3)
```golang
	if command[0] == "python" || command[0] == "python3" {
```

ui/v2.5/src/docs/en/Scraping.md
```diff
@@ -85,7 +85,7 @@ script:
   - query
 ```
 
-This configuration would execute `python iafdScrape.py query`.
+Stash will find the correct python executable for your system, either `python` or `python3`. So for example. this configuration could execute `python iafdScrape.py query` or `python3 iafdScrape.py query`.
```
WithoutPants, line 88: This should be more explicit about what it does. From what I understand of `findPythonExecutable`, this will search for `python3` first, then fall back to use `python` if the former isn't found.

WithoutPants (changes requested):

bnkai (approved):

WithoutPants (approved):

PR #1159, opened 2021-02-28
Fix SQL error in 8K resolution filter

gitgiggety:
The 8K resolution filter is currently broken both for scenes and images. This fixes the SQL error by removing the redundant open parentheses (same as the `VERY_LOW` resolution doesn't use parentheses).


WithoutPants (approved):
Thanks for fixing :)

PR #1168, opened 2021-03-02
Performer UI improvements

WithoutPants:
Significant rewrite/separation of performer details and new/edit pages.

image

Performer details are now shown as field pairs as shown above.

The edit page now uses Formik. Page prompts if the user leaves the page after modifying a value without saving. The edit page now uses columns instead of a table:

image
image

The URL scrape button is now attached to the field. The below screenshot shows when it is enabled:

image

This PR also includes an upgrade to `react-scripts` 4.0.3, and adds an environment variable to stop the eslint causing the app to not initialise.


bnkai:
In the edit performer page the scrape button appears but doesnt seem to work.

bnkai (approved):

PR #1185, opened 2021-03-07
Updated to reflect folder changing workflow

jeremymeyers:


WithoutPants:
I think this is better suited as an FAQ - either in the Wiki or as a new manual page.

WithoutPants:
Closing due to inactivity.

PR #1190, opened 2021-03-10
Setup and migration UI refactor

### LINKED ISSUE #540: [Feature] Improve initial setup experience

WithoutPants:
Resolves #540 

This PR includes the following changes:
* refactor `manager/config` to use an instance rather than bare methods. Some clients have been modified to accept a `config.Instance` object, but others are using `config.GetInstance`. Eventually we want to move toward using the former.
* startup process has been changed, to use the following process:
  * if no config file is found, the system continues to start up, but does not initialise the database. Functions that use the database will return a `database not initialized` error when called, until the system is initialised.
  * the exception to this is if the new `STASH_CONFIG_FILE` environment variable is not empty, in which case, an empty config file will be created and filled in with any applicable config values from the environment. This is to maintain docker container compatibility.
  * if a migration is required, the database is not initialised until the migration has been executed.
* added a `systemStatus` graphql query that responds with the status (`SETUP`, `NEEDS_MIGRATION`, or `OK`), the path to the database file, and the app and database schema versions.
  * `SETUP` is returned if no config file was found.
* added `setup` and `migrate` graphql mutations.
* made `stash`, `cache` and `metadata` config fields non-mandatory. `cache` currently has no use, and `metadata` is only needed for full export/import - so an applicable error message is returned in those functions if it is not set. `database` and `generated` are mandatory and a panic will be raised if the config is missing these fields.
* added setup and migration wizards to the UI, with code to redirect to these pages if required after checking the system status.
* added `STASH_CONFIG_FILE` environment variable to the docker files.

Screenshots:
Setup:
image
image
image
image

Migration:
image


bnkai:
Doing a `clean setup` with default settings in Linux errors out.
Using the $HOME/.stash button i get
all-default(homepath)

And then after the confirm
error-home

Going back and selecting the cwd button results in the same error (probably related to the below error as a config file is created this time)

If i restart stash and go with option  use current workind directory
all-default(current)
All seems fine and setup completes after the confirm BUT when i reach the config and try to save i get 
error-cwd

jeremymeyers:
Would suggest that the "success" dialog make some note about the difference between scanning (adding files to the database, pulling metadata from filenames when possible) and scraping (pulling full scene and performer info from external sources) since it seems that people get confused and frustrated when they set up a scan and theres still a lot of missing data.

bnkai:
The error when trying to save  from the configuratioin page appears to be from the  `Cache` directory not being set (if we follow the default installations i described).
With the last commit the setup completes for both cases. In the case of the home directory default install  though it uses `/home/username` instead of `/home/username/.stash` as a default path and so the config file/db are saved in  that directory instead. That means that when stash is restarted the config.file won't be found and the installation is launched again.

bnkai:
Previous errors are fixed.
New issue found though.
Using the cwd option during setup we end up with realtive paths.
gen

While the database is ok the generated folder cannot be used.
Ffmpeg during scan errors out as it doesnt try to resolve the relative path.

```bash
ffmpeg error when running command </usr/bin/ffmpeg -v error -ss 17.068 -y -i /reducted/Videos/thomas_mandelid--clips_of_the_Aurora.mp4 -vframes 1 -q:v 2 -vf scale=1280:-1 -f image2 generated/screenshots/c37eb1fac6c9a706.jpg>: [image2 @ 0x5555556ae780] Could not open file : generated/screenshots/c37eb1fac6c9a706.jpg

bnkai (approved):

PR #1193, opened 2021-03-10
Upload Image from url

### LINKED ISSUE #479: [Feature] Upload Image From URL
It creates a few extra steps for each image I add for a performer or studio having to download the image to my computer and then upload it to stash.

It would be a lot easier if images could be added through the direct image URL from any website.

There isn't really a current alternative to this.

**Additional context**
N/A

WithoutPants:
Resolves #479 

Changes `Browse for Image...` buttons to `Set Image` dropdown button. This gives two options - `From File...` and `From URL...`. The former option is the same as existing behaviour. The second opens a modal dialog allowing the input of a URL to an image. Entering the URL and confirming sets the image URL in the form. The create/update backend code has been changed to accept either a base64 encoded image, or a URL.

image

### DIFF
--- a/pkg/api/resolver_mutation_movie.go
+++ b/pkg/api/resolver_mutation_movie.go
@@ -26,7 +26,7 @@ func (r *mutationResolver) MovieCreate(ctx context.Context, input models.MovieCr
 
 	// Process the base 64 encoded image string
 	if input.FrontImage != nil {
-		_, frontimageData, err = utils.ProcessBase64Image(*input.FrontImage)
+		frontimageData, err = utils.ProcessImageInput(*input.FrontImage)
 		if err != nil {
 			return nil, err
 		}
@@ -34,7 +34,7 @@ func (r *mutationResolver) MovieCreate(ctx context.Context, input models.MovieCr
 
 	// Process the base 64 encoded image string
 	if input.BackImage != nil {
-		_, backimageData, err = utils.ProcessBase64Image(*input.BackImage)
+		backimageData, err = utils.ProcessImageInput(*input.BackImage)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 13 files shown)


bnkai:
Works ok.
A couple of  (possibly not needed) tweaks
- Should the http client use the same settings as the scraper/image one? I dont think user agent checks would be used for plain images but it doesnt hurt to have the functionality. The same for insecure certs. Given that this is in `utils` we could use a fixed ua and a skip check by default for the certs (this would be only for the images so it shouldnt be a security issue).
- Given that this is mostly user triggered should the timeout be increased to 60 or 90 secs? That would also work as a failover if the scraping fails due  to slow/non responsive sites. (The user can manualy set the image and wait a bit more than the 30secs of the scraping timeout. The UI in the scraper dialog always gets the  image while the backend sometimes timeouts)

bnkai (approved):

WithoutPants:
> * Should the http client use the same settings as the scraper/image one? I dont think user agent checks would be used for plain images but it doesnt hurt to have the functionality. The same for insecure certs. Given that this is in `utils` we could use a fixed ua and a skip check by default for the certs (this would be only for the images so it shouldnt be a security issue).

I left these out because they were named as scraper configuration options, so I didn't want to silently conflate them. Possible follow-up PR to sanitise the whole thing.

> * Given that this is mostly user triggered should the timeout be increased to 60 or 90 secs? That would also work as a failover if the scraping fails due  to slow/non responsive sites. (The user can manualy set the image and wait a bit more than the 30secs of the scraping timeout. The UI in the scraper dialog always gets the  image while the backend sometimes timeouts)

I'm happy to increase it. 60 seconds seems reasonable, 90 starts getting too long.

Issue #944, opened 2020-11-16
[Bug Report] Performer scrapers


hassaffe:
If you are looking for an actress and the performer scrapers. If there are several actresses with the same name, several names are displayed in the list. But no matter which entry in the list you select, the same actress is always loaded.

**Desktop (please complete the following information):**

 - OS: Linux
 - Browser Firefox, Chrome

![StashError](https://user-images.githubusercontent.com/55273555/99297856-1e3d7580-2849-11eb-9010-e72451b338bd.jpg)

halorrr:
I was actually going to be bringing this up as a feature request https://github.com/stashapp/CommunityScrapers/pull/268#issuecomment-725230104

From a UX perspective it would be nice to give the pop up a max height and make it scrollable in my head.

From a function perspective as someone who downloads ever community scraper, it would be nice in settings, especially now that we have a scraper list to turn off search providers you don't want to show in the search list. (Example if you only use a particular scraper for the scene function, but it is taking up space in your performer list)

bnkai:
The main issue mentioned here (the UX one is still valid) looks more like a specific scraper issue than a stash issue ( maybe scraper doesnt return correct name/url pairs). @hassaffe can you make an issue in <https://github.com/stashapp/CommunityScrapers/issues> for the specific scraper you used ?

hassaffe:
yes I do

bnkai:
Ok verified.
The graphql query seems to return correct results.
<https://pastebin.com/Qf8QT6qA>
If you search the query for `"barbie"` you'll see that all the results are there.
No matter which `barbie` you choose from the UI though the first with url `http://www.iafd.com/person.asp?perfid=barbie_rus_10&gender=f&barbie.htm` will be selected

peolic:
Perhaps the value (which is later used to determine which item was selected) shouldn't use `name` in this case:
https://github.com/stashapp/stash/blob/21806decca8cb192ded92543d3a53553bae3bfbc/ui/v2.5/src/components/Shared/Select.tsx#L147
Maybe URL, as it will most likely be the unique value.

PR #1198, opened 2021-03-11
Improve performer scrape search modal

### LINKED ISSUE #775: [Feature] Performer name should be set in the scrape by name search field if available.
When scraping for a performer by name, where the name field is already populated, it is pretty depressing to copy paste it in the search field of the scraper and feels like a bad UX.

Populate the editable search field with the performer name when it's initialized if available.

As an alternative or even better UX the "Scrape with" button could be moved next to the name field to be consistent with scrape by URL. The pop up is not needed in this case, where the name field acts as search field as well.

InfiniteTF:
Removes select in favor of a search field and a scrollable list. Initializes to performer name when scraping existing performer.

image

Resolves #775 and #944 .

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v060.md
+++ b/ui/v2.5/src/components/Changelog/versions/v060.md
@@ -2,7 +2,7 @@
 * Added Performer tags.
 
 ### 🎨 Improvements
-* Improve performer scraper search modal
+* Improve performer scraper search modal.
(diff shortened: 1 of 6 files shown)


bnkai (approved):

jeremymeyers:
Would it be possible to add an 'open in external window' icon in the dropdown? The use case would be for if there are performers with the same name in the dropdown making it impossible to determine the correct one to select without additional information

PR #1211, opened 2021-03-17
Add write mutex and max connection lifetime

WithoutPants:
Adds a database write mutex so that only one write operation may occur at once. This should eliminate any database locked issues.

Adds a connection timeout of 30 seconds for the database connection. This will mean that after 30 seconds of inactivity, the `-wal` and `-shm` files will be deleted, hopefully mitigating some of the risk if users try to restore a database file without removing these files.

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


bnkai:
Works ok.
The timeout seems to do the job although sometimes i have to stop and restart stash to actually `merge` the wal and shm file.

bnkai (approved):

PR #1214, opened 2021-03-19
Scene queuing

### LINKED ISSUE #309: [Feature] Playlist Type Functionality.
Filters and then being able to have what I guess is what I'd call a play list on your filters.  

So lets say you do a filter on all 5 star scenes, upon going to the scene player those scenes are in a playlist so you can basically go to next video, or use the random play function to randomly go through to another scene in said playlist.

I could see this being useful for many watch activity and also cataloging / tagging scenes.  Typical workflow would to set a filter on unwatched scenes of a performer.  Play random could then be used to randomly select scenes.  Conversely for cataloging / organizing you could set a filter and then roll through those videos adding tags / studios and what not.

### LINKED ISSUE #557: Add Play random button to sceneplayer
add a play random button in the `sceneplayer` to change the scene randomly without having to come back to navigate on scene or scene marker

WithoutPants:
Includes fix for `FindScenes` which wasn't handling `sceneID` is passed.
Includes fix for `GetSceneStreamPaths` where failing to get the container caused the entire operation to fail. Now ignores, meaning it will fallback to transcoding if an error occurs while trying to determine the container.

Adds scene queuing. When clicking a scene from the scene list, a scene queue will be populated from the current filter. The scene queue is visible as a tab on the scene details page.

image

It is possible to click on any of the scenes in the queue and the clicked scene will be played. Also has previous, next and random controls. 

The Play Random operation in the scenes page now also populates the queue and sorts by random.

Added a new "Play selected" operation in the scenes page overflow menu, which puts all selected scenes into a queue and plays them in order.

When autoplay is enabled (ie when clicking a queue item, or Play Random or Play Selected operations), the next scene will be played automatically when the current scene ends.

Current annoyances/things to improve:
* the o-counter and organized buttons probably need to be moved off the tabs and placed near the title instead.
* I'd like to have the queue content scrollable, rather than the current scrolling which puts the controls off screen.
* it would be nice to have the play next/previous buttons in the scene player, but it's probably not worth the effort if we plan on moving away from jwplayer.
* situating the queue in the tabs isn't the ideal place, but I struggle to find a good place. Perhaps want a button (probably on the player) that overlays the content or something.

Resolves #309 
Resolves #557


bnkai:
Seems to work ok as described.

Some additional issues for the list
- minor annoyance: the user cannot opt out of the queue functionality. Idk if instead of clicking a scene, a new button or menu item should be clicked instead to populate the queue. This is mostly for cases  when you are browsing your library and clicking one of the scenes populates the queue with all available scenes.
- functionality: remove an item from the queue or change its position in it.

WithoutPants:
> * minor annoyance: the user cannot opt out of the queue functionality. Idk if instead of clicking a scene, a new button or menu item should be clicked instead to populate the queue. This is mostly for cases  when you are browsing your library and clicking one of the scenes populates the queue with all available scenes.

I'm not sure why this matters. When clicking on a scene in this way, the queue is populated, but other than the presence of the queue tab, there's no change in functionality. The queue should not continue to play when navigated from this, so the queue is there if needed, but is otherwise unobtrusively available.

bnkai (approved):

PR #1221, opened 2021-03-22
Correct tag link for gallery and images tags

### LINKED ISSUE #1216: [Bug Report] Gallery Tag selection leading to Scene
**Describe the bug**
When tagging a gallery, one can see the tags in the gallery view. On Click, those tags should filter by this tag, which they do, however they lead to a filter on the scenes page. This does not seem like it is intended and is rather annoying.

**To Reproduce**
1. Go to Galleries
2. Tag a Gallery
3. Hover over the Tag icon in Galleries, click the tag
4. You get the filtered page in Scenes

**Expected behavior**
You should get the selected Tag filter in Galleries. For now I didn't try it with images, but it should be a general rule that a Filter should be applied to the type of items the tag was associated with.

v0.5.0-32-g53cb9a1b

**Desktop:**
 - OS: Linux
 - Browser: Chromium
 - Version: 87.0.4280.141

WithoutPants:
Fixes #1216

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryCard.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryCard.tsx
@@ -47,7 +47,7 @@ export const GalleryCard: React.FC<IProps> = (props) => {
     if (props.gallery.tags.length <= 0) return;
 
     const popoverContent = props.gallery.tags.map((tag) => (
-      <TagLink key={tag.id} tag={tag} />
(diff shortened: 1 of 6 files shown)


bnkai:
The TagLink should also be updated in `GalleryDetailPanel.tsx` and `ImageDetailPanel.tsx`

bnkai (approved):

PR #1224, opened 2021-03-23
[Feature] Added slideshow to gallery in wall display mode

elad-lachmi:
# Feature: Add slideshow to the gallery wall view mode

Added a slideshow option to the "more" menu for galleries in wall view mode. 
This feature reuses the existing Lightbox. It adds controls for pause/resume of the slideshow, as well as a control for the delay between transitions in seconds.

I also added a section to the interface configuration for setting the default slideshow delay, incl. updated GQL schema and go code to save/load this configuration setting. 
I myself wanted such a feature, so I went ahead and implemented it. Hope that's ok :) 

This is my first contribution. I tried to follow the guidelines as best I could. 
I'd be more than happy to fix any issues you find in CR. Please let me know if anything isn't in line with the guidelines. 

## QA Notes
I've tested this with several galleries. I've also tested with an empty gallery. 
Everything I tested was working as expected. Admittedly, I don't yet know the software well enough to say my own testing was exhaustive.


WithoutPants:
Thanks for the contribution. 

Why was the formatting of `config.graphql` changed?

elad-lachmi:
It was actually auto-formatted by prettier on save, since I forgot to disable it globally in vscode. 
However, I believe that is the correct format for descriptions according to the [spec](http://spec.graphql.org/June2018/#sec-Descriptions), so I didn't revert it.

WithoutPants:
Yes please, since it is now inconsistent with the rest of the graphql files. We'll probably want to be a pass through the graphql files in future with prettier.

elad-lachmi:
Good point. Done. 
I don't want to mix it in with this feature, but I don't mind doing a chore commit for adding prettier (or whatever formatter) for graphql files in the `fmt` step and align all the existing files to that.

bnkai:
Imo the format already used is easier to work with. The `spec` compliant one from prettier (especially the enum part) looks more complicated/bloated.

elad-lachmi:
Yes, I see what you're saying. The way it's formatted right now is more succinct.

ui/v2.5/src/hooks/Lightbox/Lightbox.tsx
```diff
@@ -260,6 +304,38 @@ export const LightboxComponent: React.FC<IProps> = ({
...
+                        value={(slideshowInterval ?? 0) / SECONDS_TO_MS}
+                        onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+                          handleSlideshowIntervalChange(
+                            Number.parseInt(e.currentTarget.value, 10) *
+                              SECONDS_TO_MS
+                          );
+                        }}
+                        size="sm"
+                        disabled={slideshowInterval === null}
+                      />
```
WithoutPants, line 326: Please add `className="text-input"` so that the text field styling is consistent with fields.

WithoutPants (changes requested):
I've run into a few issues:
* clicking left/right button should reset the interval - at the moment you can wait four seconds then click left/right, then a second later the image goes to the next one automatically.
* if I try to highlight the text in the delay field by clicking and dragging in the text field, then the slideshow ends
* setting the delay to 0 makes the slideshow go bonkers :) The delay should have a minimum of 1. I noticed that I could set it to a negative number.
* there's some weird behaviour when moving to a different tab in Chrome, waiting a few intervals and then clicking back - the transitions between all the images happen at once

ui/v2.5/src/hooks/Lightbox/Lightbox.tsx
elad-lachmi: Added className + set a min value of 1.

- re: left/right click - I have a fix in mind. Will implement later today.
- re: 0/negative number - Does go a bit bonkers :) the min setting on the input should address that
- re: drag input - that works only on the spinner itself. If you click on the spinner and drag up/down, the number will change as expected. Dragging the value, I believe, is not built into number inputs and is a whole other implementation. Not sure it's worth getting into. WDYT?
- re: move between tabs - this is because the interval is allowed to run in the background, while the UI update is not. This can be solved using the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API). Pause on lost visibility and restart on the visible callback

I'll take care of the other issues later today.
Thanks for the feedback!

elad-lachmi: Done and done.

WithoutPants: I can still type in `0` or a negative number and it goes bonkers, so `min` won't be enough.

My suggestion would be the following:
* have the slideshow be the default behaviour when clicking an image in the wall view and remove the `Start Slideshow` menu item. Clicking the image shows the slideshow but starts paused.
* delay should always be shown, even if paused. It currently sets the delay text field to 0. It should keep the existing value. It's not necessary to disable the text field.
* `useInterval` should ensure that the interval value is not less than some safe value. It should either error out or choose a safe value if an invalid value is provided. The slideshow client should also set the minimum value as `1` so that even if the user types in a `0` or negative value, it will still use a delay of `1`.
* the styling needs to be fixed up on smaller screens, compare non-slideshow view vs slideshow view on viewports < 800px wide:

image

In terms of the drag input, I found that if I highlight the current value and don't drag outside the textfield, then it will highlight the input value correctly, if I let go outside the textfield then the slideshow ends. To be clear, all I'm trying to do is select the current value to type in a new value (which can also be done by double clicking on the field). I'm not sure if there's a fix or not.

elad-lachmi: I was initially trying to make the slideshow feature as isolated as possible, but I'm on board with what you're suggesting.  
I'll also add better handling of invalid values both on the input and useInterval and look into < 800px viewport width.

Regarding the drag behavior, if I understand the use-case you're describing correctly, this is due to how the Lightbox handles mouse events and not specifically the input. I'll see how much of an effort it'll be to fix this and go from there.

elad-lachmi: - I implemented your suggestion for how the slideshow should work

- I added checks for invalid values both on the input itself and the useInterval hook

- Space comes at a premium for < 800px, so I created a CSS breakpoint and replaced the inline form with an icon and a popover form. I also refactored the HTML, so the entire header isn't so much based on margins anymore

- Regarding the input, I fixed it so that if the mousedown event happens within the input, it's not going to close the lightbox. It's not perfect, but I think it's much better (and less annoying) than it was

WithoutPants: > * Regarding the input, I fixed it so that if the mousedown event happens within the input, it's not going to close the lightbox. It's not perfect, but I think it's much better (and less annoying) than it was

I can no longer click on the field to edit it. Clicking on the field shows the cursor briefly, then it defocuses.

Everything else looks good.

elad-lachmi: Sorry, had a super busy few days.
I'll take a look at the input fix.

WithoutPants: All good. I've fixed the issue.

elad-lachmi: Great! btw: got BE and FE going fully native on Apple M1, FYI :)

PR #1227, opened 2021-03-24
Movie UI refresh

WithoutPants:
Cleans up and improves the Movie page.

Details and Edit panels have been separated. Details are now shown in the same style as performers:

image

If no back cover image is set, then it is now omitted from display: 

image

Edit page now uses formik and prompts if navigating away after changing data:

image


bnkai:
Something is wrong with the movie date when you switch from view to edit mode.
Even if the date is populated it shows as null.
The issue is present even in the screenshots above 
`Date:11 September 2003` -> `Date:`

bnkai (approved):

PR #1228, opened 2021-03-24
Add avif to the compatible Image Formats

yoshnopa:
Avif is a new open image format which is very likely to be the next big format. It is supported by Chrome (and Chromium-derivates) as well as Firefox, although there is a config variable you have to set (should be gone any time now, it was announced to be standard at the end of February).

### DIFF
--- a/pkg/manager/config/config.go
+++ b/pkg/manager/config/config.go
@@ -37,7 +37,7 @@ var defaultVideoExtensions = []string{"m4v", "mp4", "mov", "wmv", "avi", "mpg", 
 
 const ImageExtensions = "image_extensions"
 
-var defaultImageExtensions = []string{"png", "jpg", "jpeg", "gif", "webp"}
+var defaultImageExtensions = []string{"png", "jpg", "jpeg", "gif", "webp", "avif"}
 
 const GalleryExtensions = "gallery_extensions"
 


yoshnopa:
What do we need the decoder for? I tested it and everything worked fine.

yoshnopa:
To elaborate on that, when searching for "gif" in the repo, there is only one occurance in the Code which is only triggered on creating a thumbnail for a scene (which in turn is done by taking a shot from the video). There is no occurance associated with the actual image object (from the image tab) beside in the array of allowed filetypes, the line that was modified in this request.

bnkai:
You missed the `SetFileDetails` part probably.
Having a decoder is needed for the thumbnail creation as you mentioned and to get the dimensions width/height  and calculate the resolution.
I dont think we should add the extension to the `default` ones till we have a  working (pure-go) decoder.
Is there a specific reason it had to be in the default anyway? Any user can add whatever extension they want if the y can live with the limited usability.

bnkai:
Sadly avif/heiff are already requested for a couple of years in the ffmpeg trac. I wouldnt expect them soon.

yoshnopa:
Thank you for your patience, your right I didn't find the `SetFileDetals`.
Also, I wasn't aware that you can set the allowed file endings in the settings, maybe I should check that before trying to compile a new version :sweat_smile: 
Anyway, thank you for your patience, I agree that this might be far to soon and would suggest compatibility where there is none.

PR #1234, opened 2021-03-25
Add scene sorting by tag count

mrbrdo:
Hey,

I added scene sorting and filtering by tag count. Especially useful when organizing newly imported files. I had to do a few minor changes to the code to make it possible.
As it's my first time writing Go, please check that I am not violating some code style laws or something like that :)
Due to my editor (Atom) trimming trailing spaces, there were a few extra lines or so that were changed just because of the trailing spaces being removed. If it's a big deal I can fix that up (or you can at merge if u want). One of those was a multiline sql string so you might want to check if that trailing space was needed, although I don't think so as newline counts just as well.

PS: I can also confirm that stash builds fine on Apple M1 arm64.

Cheers!

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -47,7 +47,7 @@ input PerformerFilterType {
   measurements: StringCriterionInput
   """Filter by fake tits value"""
   fake_tits: StringCriterionInput
-  """Filter by career length"""  
+  """Filter by career length"""
   career_length: StringCriterionInput
   """Filter by tattoos"""
   tattoos: StringCriterionInput
@@ -80,7 +80,7 @@ input SceneFilterType {
   AND: SceneFilterType
   OR: SceneFilterType
   NOT: SceneFilterType
-  
+
   """Filter by path"""
   path: StringCriterionInput
   """Filter by rating"""
@@ -103,6 +103,8 @@ input SceneFilterType {
   movies: MultiCriterionInput
   """Filter to only include scenes with these tags"""
   tags: MultiCriterionInput
+  """Filter by tag count"""
+  tag_count: IntCriterionInput
   """Filter to only include scenes with performers with these tags"""
   performer_tags: MultiCriterionInput
   """Filter to only include scenes with these performers"""
--- a/pkg/sqlite/scene.go
+++ b/pkg/sqlite/scene.go
@@ -348,6 +348,7 @@ func (qb *sceneQueryBuilder) makeFilter(sceneFilter *models.SceneFilterType) *fi
 	query.handleCriterionFunc(sceneIsMissingCriterionHandler(qb, sceneFilter.IsMissing))
 
 	query.handleCriterionFunc(sceneTagsCriterionHandler(qb, sceneFilter.Tags))
+	query.handleCriterionFunc(sceneTagCountCriterionHandler(qb, sceneFilter.TagCount))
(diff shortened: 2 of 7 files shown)


WithoutPants:
Thanks for the contribution. I've done some testing and found that filtering by `tag_count = 0` doesn't work correctly.

I'm thinking that the count functionality can be rolled into `MultiCriterionInput` (with `has count`, `has not count`, `has less than`, `has more than`), and utilise a `HAVING` clause rather than the current joining logic (which is what causes the `0` criterion to not work). This allows us to roll out this functionality to more associations and means we don't need to add another filter criterion (`tag_count`).

This PR should either be updated with this new approach, or I can raise an alternative PR and close this once it's complete.

BTW, this needed a unit test. Here's one I made to test it.

```
func TestSceneQueryTagCount(t *testing.T) {
	const tagCount = 1
	tagCountCriterion := models.IntCriterionInput{
		Value:    tagCount,
		Modifier: models.CriterionModifierEquals,
	}

	verifyScenesTagCount(t, tagCountCriterion)

	tagCountCriterion.Modifier = models.CriterionModifierNotEquals
	verifyScenesTagCount(t, tagCountCriterion)

	tagCountCriterion.Modifier = models.CriterionModifierGreaterThan
	verifyScenesTagCount(t, tagCountCriterion)

	tagCountCriterion.Modifier = models.CriterionModifierLessThan
	verifyScenesTagCount(t, tagCountCriterion)
}

func verifyScenesTagCount(t *testing.T, tagCountCriterion models.IntCriterionInput) {
	withTxn(func(r models.Repository) error {
		sqb := r.Scene()
		sceneFilter := models.SceneFilterType{
			TagCount: &tagCountCriterion,
		}

		scenes := queryScene(t, sqb, &sceneFilter, nil)
		assert.Greater(t, len(scenes), 0)

		for _, scene := range scenes {
			ids, err := sqb.GetTagIDs(scene.ID)
			if err != nil {
				return err
			}
			verifyInt(t, len(ids), tagCountCriterion)
		}

		return nil
	})
}
```

mrbrdo:
Sounds good, personally I don't like to use GROUP/HAVING with complex queries, in my experience it can be a real hassle (like breaking certain clauses, different way the results are given, the specifications can be very different between DBMS and so on). And many times performance can even be worse than with a virtual table like this (although my experience is mostly with Postgres and maybe with Sqlite, things are different).
Generally I agree with your proposition, but maybe compare HAVING and virtual table performance first, if it's the same then personally I would still go with the virtual table :)

I could give it a shot with MultiCriterionInput, but not sure if it would be as you imagined, as I said I never worked with Go before and am not familiar with code style preferences.

As far as count 0 not filtering correctly, if addJoin makes a normal or inner join then that is the issue. Should just be changed to do a left join instead and I think that would solve the issue. Just guessing because I didn't have time to check rn.

Cheers!

mrbrdo:
Also worth looking into regarding performance would be a correlated subquery like this:
`SELECT scenes.id, ..., (
        SELECT  COUNT(*) AS count
        FROM scenes_tags WHERE scene_id = scenes.id
        ) FROM ...`

Is there a simple way to enable logging executed SQL queries from stashapp? I can check the different query types.

WithoutPants:
Thanks for the submission. I've extended it in #1254 so I'll close this PR. You'll be marked as co-author in that PR. Feel free to offer feedback on the changes.

mrbrdo:
Thanks. Too bad cause I really liked the number of the PR 1234. But sadly I have to agree.

PR #1241, opened 2021-03-29
Api key

WithoutPants:
Adds the option to generate an API key. Adds a `generateAPIKey` graphql interface which generates and sets an API key in the configuration. API key can also be cleared.

If the correct API key is present in the header for unauthenticated requests, then the request is treated as if it were authenticated with the configured user. 

Adds an API key field to the configuration page:

image

Tested against a custom build of the kodi plugin.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/antchfx/htmlquery v1.2.3
 	github.com/chromedp/cdproto v0.0.0-20200608134039-8a80cdaf865c
 	github.com/chromedp/chromedp v0.5.3
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/disintegration/imaging v1.6.0
 	github.com/fvbommel/sortorder v1.0.2
 	github.com/go-chi/chi v4.0.2+incompatible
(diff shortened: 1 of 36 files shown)


bnkai:
Seems to work ok
Can we add something to the documentation/help for it ? ( I'll update the wiki with an example as soon as it is merged )
I had to look at the source to confirm the actual header `APIKey`
Working example (after you set the API key)
```
curl -X POST -H "ApiKey: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1aWQiOiJiaWxsIiwiaWF0IjoxNjE3MDkxOTAzLCJzdWIiOiJBUElLZXkifQ.z7ccsVhd__DI7N2yYHkTAkidCKnBtKuccTy_dB1-fTs" -H "Content-Type: application/json" --data '{ "query": "mutation { generateAPIKey (input:{clear: false})}" }' localhost:9999/graphql
curl -X POST -H "ApiKey: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1aWQiOiJiaWxsIiwiaWF0IjoxNjE3MDkzMDYwLCJzdWIiOiJBUElLZXkifQ.WhUyvmnVeW8wGV5fkVyje3xLfz5A97HFwyZy-4i8Q-I" -H "Content-Type: application/json" --data '{ "query": "mutation { metadataScan (input:{useFileMetadata: false})}" }' localhost:9998/graphql
```
The only thing that is a bit annoying is  that we don't get a response if the authentication has failed eg wrong api key

bnkai (approved):

PR #1249, opened 2021-03-31
Fix "Clear Image" button

peolic:
This PR fixes two scenarios:
1. **Editing a performer/studio/tag:**
   _Before:_ When clearing, the image remains visible (but when saving the performer/studio/tag, it is removed).
   _Now:_ The default performer/studio/tag image replaces the cleared image.
2. **New performer:**
   _Before:_ When clearing, the image source changes to `undefined?default=true` [making the ` alt` value appear](https://user-images.githubusercontent.com/66393006/113194337-b3299400-9269-11eb-9672-ad1ee2b0522b.png).
   _Now:_ The image is hidden.

The code in the second commit feels ugly, but this is all I could come up with (suggestions are welcome).

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v070.md
+++ b/ui/v2.5/src/components/Changelog/versions/v070.md
@@ -10,5 +10,6 @@
 * Change performer text query to search by name and alias only.
 
 ### 🐛 Bug fixes
+* Fix `Clear Image` button not updating image preview.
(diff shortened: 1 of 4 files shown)


bnkai:
Can you apply the fix to the studios and tags also?

WithoutPants (approved):

PR #1251, opened 2021-03-31
Fix performer age timezone issues

peolic:
Continuing #1237, and as discussed on Discord:
The scene date / birthdate were both being created as UTC, which meant they are not at the user's midnight (as it should have been), causing the age to be incorrect, at times.

This could lead to issues if at any point later on a full date-time string is passed to this function,
because obviously the time part will be reset.
But this is function is specific for calculating the age, and I believe the comment makes it clear.

### DIFF
--- a/ui/v2.5/src/utils/text.ts
+++ b/ui/v2.5/src/utils/text.ts
@@ -72,11 +72,14 @@ const fileNameFromPath = (path: string) => {
 const getAge = (dateString?: string | null, fromDateString?: string) => {
   if (!dateString) return 0;
 
+  const birthdate = new Date(dateString);
+  const fromDate = fromDateString ? new Date(fromDateString) : new Date();
+
   // Set time to midnight to force system timezone instead of UTC
-  const birthdate = new Date(`${dateString} 00:00:00`);
-  const fromDate = fromDateString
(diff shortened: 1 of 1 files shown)


WithoutPants:
I don't think this is the correct solution. The crux of the matter is that the two dates that we're comparing are at different timezones, UTC and local. I'm not convinced that setting the UTC hours to midnight will necessarily fix the issue, but I haven't tested to make sure. I would think that if your timezone is before UTC, then this will set the date to the day prior. 

I think the better solution would be to parse and create the `Date` object correctly using local time (creating from string treats the date as UTC time). The date objects can then be compared correctly. I would add a separate function `stringToDate` to create the date in local time from the standard date string (`YYYY-MM-DD`), and change the `getAge` function to use the new function to create the `birthdate` and `fromDate` `Date` objects.

peolic:
I was writing a whole response, but halfway through I figured I we had the same idea, that using `new Date()` to parse a string is not the way to go. I'll work on `stringToDate`.

peolic:
Please let me know if you feel this could be improved upon.
I've tested this, works as intended.

WithoutPants (approved):

PR #1254, opened 2021-04-01
Join count filter criteria

WithoutPants:
Extension of #1234 

* Adds `tag_count` criterion and sorting option to Performer, Scene, Gallery and Image filters.
* Adds `performer_count` criterion and sorting option to Scene, Gallery and Image filters.
* Adds `scene_count`, `image_count`, `gallery_count` criteria to Performer filter.

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -61,6 +61,14 @@ input PerformerFilterType {
   is_missing: String
   """Filter to only include performers with these tags"""
   tags: MultiCriterionInput
+  """Filter by tag count"""
+  tag_count: IntCriterionInput
+  """Filter by scene count"""
+  scene_count: IntCriterionInput
+  """Filter by scene count"""
+  image_count: IntCriterionInput
+  """Filter by scene count"""
+  gallery_count: IntCriterionInput
   """Filter by StashID"""
   stash_id: String
 }
@@ -109,6 +117,8 @@ input SceneFilterType {
   performer_tags: MultiCriterionInput
   """Filter to only include scenes with these performers"""
   performers: MultiCriterionInput
+  """Filter by performer count"""
+  performer_count: IntCriterionInput
   """Filter by StashID"""
   stash_id: String
 }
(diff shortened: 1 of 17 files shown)


graphql/schema/types/filters.graphql
```diff
@@ -61,6 +61,14 @@ input PerformerFilterType {
...
   """Filter to only include performers with these tags"""
   tags: MultiCriterionInput
+  """Filter by tag count"""
+  tag_count: IntCriterionInput
+  """Filter by scene count"""
+  scene_count: IntCriterionInput
+  """Filter by scene count"""
+  image_count: IntCriterionInput
+  """Filter by scene count"""
+  gallery_count: IntCriterionInput
```
peolic, line 71:
```suggestion
  """Filter by image count"""
  image_count: IntCriterionInput
  """Filter by gallery count"""
  gallery_count: IntCriterionInput
```

peolic (changes requested):
Seems to work well (have only tested scenes & performers (+ performer tags), no gallery or images)

pkg/sqlite/sql.go
```diff
@@ -218,6 +223,11 @@ func getMultiCriterionClause(primaryTable, foreignTable, joinTable, primaryFK, f
 	return whereClause, havingClause
 }
 
+func getCountCriterionClause(primaryTable, joinTable, primaryFK string, criterion models.IntCriterionInput) (string, int) {
+	lhs := fmt.Sprintf("(SELECT COUNT(*) FROM %s s WHERE s.%s = %s.id)", joinTable, primaryFK, primaryTable)
```
mrbrdo, line 227: I believe you are missing the `IFNULL(..., 0)` around the query, as I had originally. Without that I believe it won't work properly for count = 0 condition.

mrbrdo (changes requested):
I believe that filter query needs the IFNULL addition.

bnkai:
- the tag_count is missing from the UI's sort order in the performers page. (manually editing the URL and using it seems to work ok)
- the IS NULL / IS NOT NULL doesnt make sense/work for the _count filters as mentioned in the above post. Is Null returns nothing while Is not Null returns all entries. 
- you can probably ignore this. The  `/scenes?sortby=tag_count&sortdir=desc` was giving me 4 irrelevant scenes as first with fewer tags. I edited->saved those scenes and the sort order is fine now. Not sure why that happened.

bnkai (approved):

PR #1263, opened 2021-04-02
Whitespace is not trimmed from the end of query strings

### LINKED ISSUE #1260: [Bug Report] Whitespace is not trimmed from the end of query strings
**Describe the bug**
If any whitespace characters follow the final character of a query, the entire Stash DB is returned as a result.

**To Reproduce**
1. Given that "MacAndCheese" is the name of a file, enter a query of "MacAndCheese " 
2. Observe that the entirety of the Stash DB is returned from the query rather than just "MacAndCheese"

**Expected behavior**
1. Given that "MacAndCheese" is the name of a file, enter a query of "MacAndCheese " 
2. Stash should return a result as if "MacAndCheese" has been entered, regardless of how much whitespace is entered at the end.

 - OS: Windows 10 20H2
 - Browser: Google Chrome 89.x
 - Stash Version: v0.6.0

julien0221:
I added a trim on the query strings so now it doesn't returns all the search if there are one or more space at the end of query.
Resolves #1260


InfiniteTF:
We should trim the query string on the backend as well so any scripts/external apps don't run into the same issue. Also, filtering out zero length clauses, which is presumably what happens when there are multiple sequential spaces in the middle of the query.

julien0221:
sure, I added a fix to trim it on backend too

pkg/sqlite/sql.go
```diff
@@ -108,7 +108,8 @@ func getSearchBinding(columns []string, q string, not bool) (string, []interface
 	}
 
 	queryWords := strings.Split(q, " ")
-	trimmedQuery := strings.Trim(q, "\"")
+	trimmedQuery := strings.Trim(q, " \"")
```
WithoutPants, line 111: This will result in inconsistent behaviour:

Search `aaa bbb`, searches `aaa` or `bbb`.
Search ` aaa bbb` (leading whitespace), will search `aaa bbb` because `trimmedQuery != q`.

If we're improving this code, then we should address another issue:
Search `aaa  bbb` (double-space), will return everything, due to one of the `queryWords` being `""`.

Suggest the following:
* trim whitespace first
* if string starts and ends with `"`, then search exact query
* otherwise, split using `" "`, remove any empty strings, and search for any of the words

This also needs to have unit tests added.

WithoutPants (changes requested):

pkg/sqlite/sql.go
julien0221: ok I changed the way to deal with double space and whitespace. And I added unit tests.

pkg/sqlite/sql.go
```diff
@@ -106,9 +107,11 @@ func getSearchBinding(columns []string, q string, not bool) (string, []interface
 		notStr = " NOT"
 		binaryType = " AND "
 	}
-
+	q = strings.TrimSpace(q)
+	q = regexp.MustCompile(`\s+`).ReplaceAllString(q, " ")
 	queryWords := strings.Split(q, " ")
 	trimmedQuery := strings.Trim(q, "\"")
+
 	if trimmedQuery == q {
```
bnkai, line 115: If we relax a bit on the "exact" query this is ok.
If "exact" though includes spaces then this is not ok and test `q = "   \"Girl     does   threesome\" "` (multiple spaces in the quoted query vs single in the title) should fail
If the later is what we want then the 
```golang
q = regexp.MustCompile(`\s+`).ReplaceAllString(q, " ")
queryWords := strings.Split(q, " ")
```
should be moved inside the 
```golang
if trimmedQuery == q {
```

pkg/sqlite/scene_test.go
```diff
@@ -1376,6 +1376,60 @@ func TestSceneStashIDs(t *testing.T) {
...
+			Q: &q,
+		}
+
+		scenes = queryScene(t, qb, nil, &findFilter)
+		assert.Len(t, scenes, 1)
+		assert.Equal(t, scenes[0].ID, created.ID)
+
+		findFilter.Q = nil
+		scenes = queryScene(t, qb, nil, &findFilter)
+		assert.NotEqual(t, 0, len(scenes))
```
bnkai, line 1425: rewritting the test a bit makes it easier to add some more queries imho
```golang
                type test struct {
                        query string
                        id    int
                        count int
                }
                tests := []test{
                        {query: " Girl    does    ", id: created.ID, count: 1},
                        {query: "   \"Girl     does   threesome\" ", id: created.ID, count: 1},
                        {query: "threesome", id: created.ID, count: 1},
                }

                for _, tst := range tests {
                        f := models.FindFilterType{
                                Q: &tst.query,
                        }
                        scenes := queryScene(t, qb, nil, &f)

                        assert.Len(t, scenes, tst.count)
                        assert.Equal(t, tst.id, scenes[0].ID)
                }

                findFilter := models.FindFilterType{}
                scenes := queryScene(t, qb, nil, &findFilter)
                assert.NotEqual(t, 0, len(scenes))
```

pkg/sqlite/sql.go
julien0221: You are right, exact should be exactly what the user want to type.

pkg/sqlite/scene_test.go
julien0221: Awesome, thank you for the change, I am still new with Golang :)

WithoutPants (approved):

PR #1264, opened 2021-04-03
Handle /healthz for liveness checks.

stashist:
This is pre-authentication as it exposes no info and is far cheaper than
e.g. transmitting the login page.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -124,6 +124,7 @@ func Start() {
 
 	r := chi.NewRouter()
 
+	r.Use(middleware.Heartbeat("/healthz"))
 	r.Use(authenticateHandler())
 	r.Use(middleware.Recoverer)
 
--- a/ui/v2.5/src/components/Changelog/versions/v070.md
+++ b/ui/v2.5/src/components/Changelog/versions/v070.md
@@ -3,6 +3,7 @@
 * Added scene queue.
 
 ### 🎨 Improvements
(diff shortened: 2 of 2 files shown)


Tweeticoats:
If we are adding this we might as well update the docker container to run a check this url to see if the container is running.
ie run curl / wget and download the url and see if this returns a valid response.
To keep size down we could use busybox and the wget applet to add 743k to the size of the container.

You can then add the following to the Dockerfile:
HEALTHCHECK --interval=12s --timeout=12s --start-period=30s  CMD busybox wget http://localhost:9999/healthz

stashist:
Definitely. That's what I'm doing now internally.

That said, my intent is to propose a followup PR for discussion where I add this functionality to the stash binary itself which is comparatively cheaper (since we're already linking in HTTP get functionality). That would allow me to move to distroless images (and propose those if it works out).

If that get's rejected I'll likely propose the bundled wget/curl instead as you already suggested.

bnkai:
This looks ok.
As for adding the functionality in stash itself i dont think it will be of any practical use  (If i understood correctly what you meant).
When you are troubleshooting you want to test the access to stash from outside eg other pc, wan, outside the container, not from where stash is running.
As for the distro-less image while stash doesnt have any external dependencies helping scripts or scrapers usualy require python or node or something else. It's a lot easier to just use an ubuntu or debian image for example to add those deps instead of a distro-less image.

WithoutPants (approved):

PR #1266, opened 2021-04-03
added an url filter option in scenes

### LINKED ISSUE #1059: [Feature] add a url filter option
- There's no way (that I know of) to search/filter by the content of a scene's url.
- Use case example: I want to determine which urls contain a specific domain so I can tag them accordingly.

- add a `url` option to the filter

**Additional context**
`url` filtering options coulde be:
- includes
- excludes
- equals
- not equals

julien0221:
I added an url filter option into scenes 
Resolves #1059

Do we want to apply the same filter for galleries, movies, performers and/or studios?

### DIFF
--- a/ui/v2.5/src/models/list-filter/criteria/utils.ts
+++ b/ui/v2.5/src/models/list-filter/criteria/utils.ts
@@ -113,5 +113,17 @@ export function makeCriteria(type: CriterionType = "none") {
     case "piercings":
     case "aliases":
       return new StringCriterion(type, type);
+    case "url": {
+      const ret = new StringCriterion(type, type);
+      ret.modifierOptions = [
+        Criterion.getModifierOption(CriterionModifier.Equals),
+        Criterion.getModifierOption(CriterionModifier.NotEquals),
+        Criterion.getModifierOption(CriterionModifier.Includes),
+        Criterion.getModifierOption(CriterionModifier.Excludes),
+        Criterion.getModifierOption(CriterionModifier.MatchesRegex),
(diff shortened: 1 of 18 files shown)


ui/v2.5/src/models/list-filter/criteria/utils.ts
```diff
@@ -113,5 +113,17 @@ export function makeCriteria(type: CriterionType = "none") {
...
+    case "url": {
+      const ret = new StringCriterion(type, type);
+      ret.modifierOptions = [
+        Criterion.getModifierOption(CriterionModifier.Equals),
+        Criterion.getModifierOption(CriterionModifier.NotEquals),
+        Criterion.getModifierOption(CriterionModifier.Includes),
+        Criterion.getModifierOption(CriterionModifier.Excludes),
+        Criterion.getModifierOption(CriterionModifier.MatchesRegex),
+        Criterion.getModifierOption(CriterionModifier.NotMatchesRegex),
+      ];
```
WithoutPants, line 125: Why is this necessary? The standard modifier options (which should include `IS [NOT] NULL`) should be sufficient.

WithoutPants (commented):
Thanks for the contribution. Yes, please extend this for the other objects that have urls.

ui/v2.5/src/models/list-filter/criteria/utils.ts
julien0221: yes you are right, I removed it and use the standard modifier.

julien0221:
Ok I applied the same url filter to movies, galleries, performers and studios. And I used the standard modifier.

WithoutPants (approved):

PR #1268, opened 2021-04-05
Simplify GH build pipeline.

stashist:
The toolchain is already bundled in the stashapp/compiler image.
Rather than introducing a second one via GH actions standardize on that
one instead.

Also
* Clear up what "Cross Compile" actually does
* Still pull stashapp/compiler separately for easier debugability.

Tested locally by manually executing the commands in a clean env ONLY
due to GH permissions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,9 @@ on:
   release:
     types: [ published ]
 
+env:
+  COMPILER_IMAGE: stashapp/compiler:4
+
 jobs:
   build:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


WithoutPants:
You should be able to keep the node caching stuff.

stashist:
@WithoutPants For what purpose do we want to keep it? Are there other users of the cache?

bnkai:
A cache is a cache nothing more. 
In theory it should make the compilation a bit faster. Instead of node downloading the deps every time we compile the UI we cache them and node only needs to download what is not there. The node_modules should be ~ 100MB (zstd compressed). The cache setup phase costs 6 secs so it isnt much.
In practice something seems weird as i cant seem to find a noticable difference in execution time for the pre-install or even compile-ui phase ( if i understood the cache use thats where we should see an improvement ) comparing this PR with others using the node  cache.

@stashist 
You switched the local make targets to ones using docker. Isnt it better to make a new makefile target so that the docker instance is only started once instead of starting and stopping 4 times? Or at least do  a `... -c ""make pre-ui && make generate && make ui-validate fmt-check vet it && make ui-only" so that a single doscker is started for the 4 targets?

WithoutPants:
The existing cache section is incorrect. Please include this in the yml file:

```
    - name: Cache node modules
      uses: actions/cache@v2
      env:
        cache-name: cache-node_modules
      with:
        path: ui/v2.5/node_modules
        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('ui/v2.5/yarn.lock') }}
```

This should shave the 45 seconds or so from the build time.

We can do everything in a single docker container, but we will lose all the steps in the action - it will become a single step. I looked into running everything in a single docker container using the `container` field, but I don't think it will play nicely with `cross-compile.sh`, so we'd need to change the way cross compilation is done as well. The current way of running each step in a separate container isn't ideal, but I doubt we'll see much benefit in executing everything in a single step.

stashist:
Sorry about the confusion folks. For some reason I though the compiler image is also including the node modules. That's obviously not the case nor would it make sense in the first place. Added the cache back using @WithoutPants 's snippet.

.github/workflows/build.yml
```diff
@@ -17,44 +20,32 @@ jobs:
...
-        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-        restore-keys: |
-          ${{ runner.os }}-build-${{ env.cache-name }}-
-          ${{ runner.os }}-build-
-          ${{ runner.os }}-
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('ui/v2.5/yarn.lock') }}
 
     - name: Pre-install
-      run: make pre-ui
+      run: docker run --rm --mount type=bind,source="$(pwd)",target=/stash,consistency=delegated -w /stash $COMPILER_IMAGE /bin/bash -c "make pre-ui"
```
WithoutPants, line 35: Please add the following step prior to this:
```
    - name: Start build container
      run: docker run -d --name build --mount type=bind,source="$(pwd)",target=/stash,consistency=delegated -w /stash $COMPILER_IMAGE tail -f /dev/null
```

Change `run` to the following: 
```
run: docker exec -t build /bin/bash -c "make pre-ui"
```

Likewise, change all subsequent `docker run` to `docker exec -t build ...`

.github/workflows/build.yml
```diff
@@ -17,44 +20,32 @@ jobs:
...
+      run: docker run --rm --mount type=bind,source="$(pwd)",target=/stash,consistency=delegated -w /stash $COMPILER_IMAGE /bin/bash -c "make generate"
     
+    # TODO: Replace with `make validate` once `revive` is bundled in COMPILER_IMAGE
     - name: Validate
-      run: make ui-validate fmt-check vet it
+      run: docker run --rm --mount type=bind,source="$(pwd)",target=/stash,consistency=delegated -w /stash $COMPILER_IMAGE /bin/bash -c "make ui-validate fmt-check vet it"
 
     - name: Build UI
-      run: make ui-only
+      run: docker run --rm --mount type=bind,source="$(pwd)",target=/stash,consistency=delegated -w /stash $COMPILER_IMAGE /bin/bash -c "make ui-only"
```
WithoutPants, line 45: After this step, please add the following step:
```
    - name: Cleanup build container
      run: docker rm -f -v build
```

WithoutPants (changes requested):
I'm unable to make these changes myself, so please make the indicated changes. 

This change makes it so that it starts the container once, and reuses it for the following steps, and finally removes it.

PR #1270, opened 2021-04-05
Added Auto scroll user back to the top when page navigation is clicked

### LINKED ISSUE #1267: [Feature] Auto scroll user back to the top when page number/page navigation is clicked
**Current State** 
If you click First, Previous, <page number>, Next or Last, Stash will open the next page of results, but it will do so without changing where you are in the browser window.

This is particularly unfortunate when the number of thumbnails displayed exceeds the height of your browser window, and you have already scrolled all the way to the bottom of the page-- forcing you to first manually scroll all the way to the top to see the beginning of the new page of results.

**Request:**
When clicking on any of the following (First, Previous, <page number>, Next or Last) auto scroll the user back to the top of the page.

julien0221:
When clicking on any of the following (First, Previous, , Next or Last) auto scroll the user back to the top of the page.
 Resolves #1267

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v070.md
+++ b/ui/v2.5/src/components/Changelog/versions/v070.md
@@ -3,6 +3,7 @@
 * Added scene queue.
 
 ### 🎨 Improvements
+* Scroll to top when changing page number.
 * Add URL filter criteria for scenes, galleries, movies, performers and studios.
 * Add HTTP endpoint for health checking at `/healthz`.
 * Support `today` and `yesterday` for `parseDate` in scrapers.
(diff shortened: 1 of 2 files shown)


WithoutPants (approved):

GsMumbo:
I’ve noticed that an unintended side effect of this is that it scrolls you back to the top when using the browser’s back button. For example if I’m going down a gallery and rating each image one by one, every time I click back I get pulled to the top of the page again. That means I have to scroll back to where I was every single time.

Is there a way to limit this to only trigger when you use the page buttons, and to exclude it when you use your browser’s forward and back buttons?

WithoutPants:
> I’ve noticed that an unintended side effect of this is that it scrolls you back to the top when using the browser’s back button. For example if I’m going down a gallery and rating each image one by one, every time I click back I get pulled to the top of the page again. That means I have to scroll back to where I was every single time.
> 
I can confirm that the issue you describe is _not_ caused by this PR.

PR #1284, opened 2021-04-11
utils: vtt: test + rewrite

stashist:
Please look commit-by-commit for ease of understanding.

### DIFF
--- a/pkg/utils/vtt.go
+++ b/pkg/utils/vtt.go
@@ -1,41 +1,37 @@
 package utils
 
 import (
-	"strconv"
-	"time"
+	"fmt"
+	"math"
 )
 
+// from stdlib's time.go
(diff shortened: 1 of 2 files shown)


bnkai:
Looks ok.
The only difference with the current implementation it that we use a fixed value  of `0` for the ms.
If we want to have the same exact vtts we would need to zero the ms part before the function returns
```golang
	msec = int(fracSeconds * 1000)
	sec, msec = norm(sec, msec, 1000)
	min, sec = norm(min, sec, 60)
	hour, min = norm(hour, min, 60)
	msec = 0
	return fmt.Sprintf("%02d:%02d:%02d.%03d", hour, min, sec, msec)
```
Otherwise we can keep the ms part as is. 
Imo either way is fine since that much accuracy isnt needed for the vtt image display.

PR #1285, opened 2021-04-11
utils: oshash: add tests

stashist:
This is mostly a result of me trying to understand what's going on.

I figured we might as well keep it for documentation.

Didn't want to commit a 12MB file to the repo so the last test is commented out.


pkg/utils/oshash_test.go
```diff
@@ -0,0 +1,60 @@
...
+// 	if err != nil {
+// 		t.Errorf("TestOshasBreakDance: Error from OSHashFromFilePath: %v", err)
+// 	}
+
+// 	want := "8e245d9679d31e12"
+
+// 	if got != want {
+// 		t.Errorf("TestOshasBreakDance: OSHashFromFilePath() = %v; want %v", got, want)
+// 	}
+// }
```
WithoutPants, line 60: I don't think there's any point in keeping this. An alternative would be to generate some file and run oshash against it.

WithoutPants (changes requested):
Since the unit test is not testing the public API, it should be named `oshash_internal_test.go`.

PR #1299, opened 2021-04-14
Separate UI

### LINKED ISSUE #752: [Feature] Allow loading UI from a different path
It would be useful for trying out UI-only changes on live/production instances, if I could load the UI files from a different path (not use the files packed into the executable).
Right now I either have to build stash with go,
or run a dev environment on the same PC I am coding on.

I would like to have the option to specify a different path that stash will load the UI files from.
If this could be a `config.yml` option, great, if not, a CLI parameter is good too.
This is would be a QoL feature for me, and maybe **only me**, so, your call 🙂

Current ways to test UI changes detailed above.
I also tried to point the dev UI to my live instance, but because it is on `https`, it caused more issues...

WithoutPants:
Adds a new configuration field `custom_ui_location`. When not empty, it will serve the UI from the configured location, rather than using the UI embedded in the binary. This will allow (advanced) users to customise their UI to their liking, and potentially allow testing of UI changes (we could zip up the UI as an artifact in PR builds for example).

Requires a restart of stash when changed.

Resolves #752

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -234,9 +234,21 @@ func Start() {
 		})
 	}
 
+	customUILocation := c.GetCustomUILocation()
+
 	// Serve the web app
 	r.HandleFunc("/*", func(w http.ResponseWriter, r *http.Request) {
 		ext := path.Ext(r.URL.Path)
+
+		if customUILocation != "" {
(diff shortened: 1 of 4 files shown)


peolic:
Resolves #752

bnkai (approved):

PR #1301, opened 2021-04-14
Skip validation of existing paths when adding new paths

InfiniteTF:
The current logic prevents users from adding new paths if one of the existing paths is missing. This makes it impossible to swap usb drives unless they have the same pathnames.

This change skips validation of existing paths so we avoid that.


pkg/api/resolver_mutation_configure.go
```diff
@@ -25,11 +25,21 @@ func (r *mutationResolver) Migrate(ctx context.Context, input models.MigrateInpu
...
 	if len(input.Stashes) > 0 {
 		for _, s := range input.Stashes {
-			exists, err := utils.DirExists(s.Path)
-			if !exists {
-				return makeConfigGeneralResult(), err
+			// Only validate existence of new paths
+			isNew := true
+			for _, path := range existingPaths {
+				if path.Path == s.Path {
+					isNew = false
```
bnkai, line 35: It shouldnt matter that much in this case, we could add a `break` here though.

WithoutPants (approved):

PR #1309, opened 2021-04-16
Duplicate checker UI improvements

WithoutPants:
**Changes the behaviour of queries - negative number now returns all results, 0 returns no results (for getting counts).**

* Moved the duplicate checker page into its own page, rather than a settings page. 
* Added Tools settings page with links to the filename parser and duplicate checker.
* Show warning in the duplicate checker if there are ungenerated phashes.
* Added multi-edit button to the duplicate checker UI, and made delete button into icon button.
* Added file path and scene popover buttons (performers, tags, markers, movies, gallery, o-counter, organized) to scene rows.
* Made table responsive.

Screenshots:

image

image

image


bnkai:
Everything seems to work ok.
A minor UI regression in the listing
The previous list used a bit of spacing to separate/group duplicates
before
Now all entries seem the same
after

PR #1320, opened 2021-04-17
Fix creating performer from gallery scrape dialog

peolic:
Apply the same fix #1314 did for scenes.


InfiniteTF:
Ah, missed that it was an issue there as well. Can you break it out into an util function so we can have the same implementation both places?

peolic:
Sure, although I don't know if I put it in the correct place...

InfiniteTF (approved):

PR #1316, opened 2021-04-16
Add in-memory screenshot generation for sprites and phash

InfiniteTF:
Skips writing/reading to disk by encoding to bmp and routing the output to stdout. No performance difference over internal nvme storage, but quite a bit faster than nas storage. Somewhat simpler as well.

PR #1322, opened 2021-04-18
Fix various generate issues

### LINKED ISSUE #1319: [Feature] Add more logging during startup
I had a crash during last night. Afterwards the container didn't start up anymore and the only log entry was 
INFO[0000] using config file: /root/.stash/config.ym

After cleaning the generated/tmp (which was containing a LOT of files), it started again. I needed support to find that out.

Expectation: Logfiles gives all relevant information what's going on, espacially during startup. Maybe a hint that process can take a while and also could be done manually without data loss would be nice.

bnkai:
Fixes a few issues during scan/generation

- tracks and removes intermediate files from preview generation. I didnt touch sprites/phashes as #1316 is a better solution for that
- makes sure that the generated tmp dir is emptied when the user stops a scan/generate task
- fixes an issue where the user could set a very short duration for the preview segment and thus generate invalid video segments ( the fix is hardcoded for now, feedback is welcome if we should also validate in the config )
- added an extra print that should close #1319


pkg/manager/generator_preview.go
```diff
@@ -59,17 +59,22 @@ func (g *PreviewGenerator) Generate() error {
...
+	tmpFiles, err := g.generateConcatFile()
+	if err != nil {
 		return err
 	}
 
 	if g.GenerateVideo {
 		if err := g.generateVideo(&encoder, false); err != nil {
 			logger.Warnf("[generator] failed generating scene preview, trying fallback")
 			if err := g.generateVideo(&encoder, true); err != nil {
+				//remove intermediate files
```
WithoutPants, line 71: Whitespace between `//` and text. Multiple places.

pkg/manager/generator_preview.go
```diff
@@ -152,3 +169,14 @@ func (g *PreviewGenerator) generateImage(encoder *ffmpeg.Encoder) error {
...
+
+func removeFiles(list *[]string) {
+	if list != nil {
+		for _, f := range *list {
+			if err := os.Remove(f); err != nil {
+				logger.Warnf("[generator] Delete error: %s", err)
+			}
+		}
+	}
+
```
WithoutPants, line 181: Remove empty line

pkg/manager/manager.go
```diff
@@ -142,8 +142,10 @@ func (s *singleton) PostInit() error {
 	// clear the downloads and tmp directories
 	// #1021 - only clear these directories if the generated folder is non-empty
 	if s.Config.GetGeneratedPath() != "" {
+		logger.Info("Please wait. Deleting temporary files...")
```
WithoutPants, line 145: I don't think this should be logged on every startup. I would perform the `EmptyDir` actions in a goroutine and wait for it to finish. If it takes more than 5 seconds or so, log a message indicating that it is deleting the temp files.

pkg/manager/manager_tasks.go
```diff
@@ -236,12 +236,20 @@ func (s *singleton) Scan(input models.ScanMetadataInput) {
 
 			if err != nil {
 				logger.Errorf("Error encountered scanning files: %s", err.Error())
+				wg.Wait()                              // running tasks should finish and
+				instance.Paths.Generated.EmptyTmpDir() // tmp emptied
```
WithoutPants, line 240: This feels like the nuclear approach. I think its better to ensure that specific files are removed during the tasks, instead of clearing out the entire `tmp` directory.

pkg/manager/manager_tasks.go
```diff
@@ -616,7 +630,7 @@ func (s *singleton) generateScreenshot(sceneId string, at *float64) {
 
 		wg.Wait()
 
-		logger.Infof("Generate finished")
+		logger.Infof("Generate Screenshot finished")
```
WithoutPants, line 633: use lowercase `s`

pkg/manager/generator_preview.go
```diff
@@ -59,17 +59,22 @@ func (g *PreviewGenerator) Generate() error {
 
 	encoder := ffmpeg.NewEncoder(instance.FFMPEGPath)
 
-	if err := g.generateConcatFile(); err != nil {
+	tmpFiles, err := g.generateConcatFile()
```
WithoutPants, line 62: The concat file is only used in `generateVideo` and `generateConcatFile` is not expensive.

I think the `generateConcatFile` should be moved into `generateVideo`, just before `ScenePreviewVideoChunkCombine`. 

`generateVideo` should have a local `generatedFiles` string slice, and is appended to as it generates each file. Use a deferred function to remove the files. This way, only actually created files will be deleted (you won't end up with delete errors for ungenerated files).

WithoutPants (changes requested):

pkg/manager/manager_tasks.go
bnkai: The scan task already does a `instance.Paths.Generated.EmptyTmpDir()` when its finished.
Doing the same when it exits due to an error  seemed natural to me.
I could just do a 
```golang
	if err != nil {
		logger.Errorf("Error encountered scanning files: %s", err.Error())
		break
		}
```
instead and add an extra if for the print in 
```golang
if s.Status.stopping
```

The curreent one singleton task at a time approach means that we should clear the tmp dir once that task is finished or not?

WithoutPants: The assumption that only one thread will be utilising `generated/tmp` can and probably will change in the future, so I'd rather be a little more surgical in how we remove files from it. As you say however, this is how it's done in other parts of the code, so I'm happy to leave it for now.

WithoutPants (approved):

pkg/manager/manager_tasks.go
bnkai: Its not one thread but one singleton task. Threads or `smaller tasks` are more than one at a time. The weighted waitgroup that was added from the parallel tasks PR and is used in most of the tasks makes that possible. The generated/tmp is only cleaned after all the threads/small tasks are finished and then the scan is finished.

PR #1336, opened 2021-04-23
Fixed 0 for weight when a new performer is created and fixed the search is null

julien0221:
When a new performer is created and if weight is not added, the value 0 is saved in the database for weight instead of null.
And then when you tried to search weight with the criteria is null, it returns an error.

All fixed with this PR

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -76,7 +76,7 @@ input PerformerFilterType {
   """Filter by hair color"""
   hair_color: StringCriterionInput
   """Filter by weight"""
-  weight: StringCriterionInput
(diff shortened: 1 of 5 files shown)


InfiniteTF:
Switching it to a string criterion will prevent greater and less than comparisons, which imo are quite important for a numeric value. I think it's better to fix whatever issue is causing the error when the db value is null.

julien0221:
> Switching it to a string criterion will prevent greater and less than comparisons, which imo are quite important for a numeric value. I think it's better to fix whatever issue is causing the error when the db value is null.

ok you are right, I just changed the code to fix the problem with Int. Now it works. ;)

PR #1338, opened 2021-04-25
Get distinct values from scraper

bnkai:
<del>
Adds a `distinct` boolean fixed post process attibute.
`distinct` is applied last after all the other pp operations

Usage example fragment (for a complete working example replace the below in the scene scraper from https://github.com/stashapp/CommunityScrapers/pull/491 )
```yaml
Tags:
  Name:
    selector: //span[@class="list_attributes"]/a
    distinct: true
```
</del>
Updated last  post

### DIFF
--- a/pkg/scraper/mapped.go
+++ b/pkg/scraper/mapped.go
@@ -6,7 +6,6 @@ import (
 	"math"
 	"reflect"
 	"regexp"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


WithoutPants:
Is a specific post-process action actually necessary? Is there any case where we get a list of strings where we _wouldn't_ want to apply a distinct filter on the result?

bnkai:
I was playing it safe. The actual `distinct` shouldn't have an issue, at least i can't think of one (it should be applied somewhere in the pp function though as in cases where concat is used the results it uses can be duplicates eg for descriptions ). For the one i added i would have to remove the sorting as there are some cases  where we want to keep the first result  by the order  it was scraped and not by name.

WithoutPants:
I think it's better to integrate this behaviour rather than an explicit flag in the scraper config.

bnkai:
Distinct is now applied by default during the pp phase.
The only special case is if a concat is present. In that case  the concat is applied first without a deduplication .( eg multiple a with perfomer names or other data in a description). If a split is applied after a concat the distinct will also be applied after the split.

WithoutPants (approved):

PR #1340, opened 2021-04-26
Revert always show preview videos on small devices

### LINKED ISSUE #1281: [Bug Report] Previews autoplay audio on iOS devices
**Describe the bug**
When on the Scenes tab, a decent chunk of preview videos start automatically playing. These all include sound, regardless of the wall settings you have set. Sometimes you only hear one video, other times they may overlap. This also ends up slowing the entire page down to a crawl if too many try to autoplay at once.

**To Reproduce**
1. Go to the Scenes tab
2. If on an iOS or iPadOS device, they should all start autoplaying without any intervention from you at all.

**Expected behavior**
Previews only autoplay when you hover over them (with iPad’s Magic Keyboard trackpad, or with your finger using touch). When a preview does play, it should respect the setting that disables sound. 

**Screenshots**
Settings: https://imgur.com/a/xibZcnx
Note that not only is sound disabled, it’s set to static image to see if that would stop it.

Screen recording: https://imgur.com/a/x7YAFPj
**Note - Recording has someone shaking their backside. They’re clothed, but may still be considered NSFW.**

**Stash Version: (from Settings -> About):** v0.6.0-21-ga2582047

 - Device: iPad Pro 12.9”, iPhone Mini
 - OS: iPadOS 14.5 (also tested on iOS 14.4.2)
 - Browser: Stock Safari

**Additional context**
I’m able to load it fine on my Galaxy Fold 2’s Chrome browser, so it’s definitely an iOS / WebKit issue causing this.

WithoutPants:
Reverted change from #1104 which automatically showed preview videos on small devices. This had the unintended effect of playing everything on iPad devices.

I considered auto-previewing on very small width devices, but given that sounds of these videos _may_ also play depending on the settings, I think it's better to have this as a user-directed action. 

Note that with this fix, I still had issues when testing on iPad where preview videos would not show correctly if they were rendered off-screen (ie you had to scroll to show the card). I've been unable to fix this issue, but I don't think it was introduced by this change.

Fixes #1281

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v070.md
+++ b/ui/v2.5/src/components/Changelog/versions/v070.md
@@ -30,6 +30,7 @@
 * Change performer text query to search by name and alias only.
 
 ### 🐛 Bug fixes
+* Reverted video previews always playing on small devices.
 * Fixed error when auto-tagging for performers/studios/tags with regex characters in the name.
 * Fix scraped performer image not updating after clearing the current image when creating a new performer.
 * Fix error preventing adding a new library path when an existing library path is missing.
--- a/ui/v2.5/src/components/Scenes/styles.scss
+++ b/ui/v2.5/src/components/Scenes/styles.scss
@@ -214,36 +214,17 @@ textarea.scene-description {
     }
   }
 
-  @media (pointer: fine) {
-    &:hover {
-      .scene-specs-overlay,
-      .rating-banner,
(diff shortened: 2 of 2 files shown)


bnkai:
Seems to work ok. Tested in desktop and an android smartphone.

bnkai (approved):

PR #1344, opened 2021-04-26
Added new filters (date and title) to galleries

### LINKED ISSUE #1262: [Feature] Sort Gallery by title and date
Current version only supports Gallery sorting by path, file mod time (these two, only if zipped), image count, creation time and update time. There should be other natural forms of sorting, using its database attributes, specially release date and title.

### LINKED ISSUE #1318: [Feature] Add Image Count Filter to Galleries
You aren’t currently able to filter galleries based on image count. Thanks to some of the “packs” in my library, there are a bunch of one image galleries. As of now though you aren’t able to filter them out, so they end up cluttering up my library and make it hard to sift through.

Image Count is added to the list of filters in the drop down for Galleries.

julien0221:
Resolves #1262 and Resolves #1318


peolic:
I check and #1318 is missing as well. An image count *filter*, not sort-by option.

julien0221:
> I checked and #1318 is missing as well. An image count _filter_, not sort-by option. Known here as `criterionOptions`.

Yes you are right, I confused both of them. Maybe will add it to this PR. to resolve 1318. Will do that tomorrow.

PR #1345, opened 2021-04-27
Autotag support for images and galleries

### LINKED ISSUE #977: [Bug Report] Auto-tagging does not import tags from images or galleries
Auto-tagging doesn't import metadata (tags, performers, studios, etc) from image file paths.

WithoutPants:
* Adds compound filters (`AND`, `NOT`, `OR`) for image and gallery queries. Refactored the queryBuilders for these accordingly.
* Adds autotag functionality for images and galleries.

Note that images in zips may not tag correctly when autotagging from performer/studio/tag due to the way that regex interacts with the null character in the image paths. This will be addressed when the character for zip paths is changed at a later date.

Resolves #977

### DIFF
--- a/pkg/models/model_image.go
+++ b/pkg/models/model_image.go
@@ -2,6 +2,7 @@ package models
 
 import (
 	"database/sql"
+	"path/filepath"
 )
 
 // Image stores the metadata for a single image.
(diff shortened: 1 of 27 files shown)


bnkai:
A join with the images table for the gallery queries seems to have been removed
```
Error executing find query with SQL: SELECT DISTINCT galleries.id FROM galleries GROUP BY galleries.id ORDER BY COUNT(distinct images.id) ASC LIMIT 40 OFFSET 0 , args: [], error: no such column: images.id
```
This makes the AVG Resolution filter and the sort by Images Count to fail in the `Galleries` page.

pkg/models/model_image.go
```diff
@@ -40,6 +41,16 @@ type ImagePartial struct {
 	UpdatedAt   *SQLiteTimestamp     `db:"updated_at" json:"updated_at"`
 }
 
+// GetTitle returns the title of the scene. If the Title field is empty,
```
bnkai, line 44: typo
```suggestion
// GetTitle returns the title of the image. If the Title field is empty,
```

PR #1364, opened 2021-05-04
DLNA

### LINKED ISSUE #850: [Feature] DLNA
DLNA is a technology that allows you to stream content from one device to any device on the network easily. A DLNA server will show in a list of network devices on supported devices such as but not limited to TVs, phones, and computers. XBVR, a similar service to Stash, uses DLNA to stream content to VR headsets and it works without issue in my experience.

The way XBVR does it you can set the server name and icon (icon only shows on devices that support it), and limit it to certain IPs on your local network. See below image for the available options.
XBVR

WithoutPants:
Resolves #850 

Adds a `DLNA` tab to the Settings page:

image

This allows the user to enable or disable DLNA by default, and to override these with time-based overrides:

image

So if DLNA is disabled by default, then the user can enable it for a specific time, or until server restart. Likewise, it can be disabled for a specific time if it is enabled by default. This temporary behaviour can be cancelled with the cancel button.

By default, the server will reject connections to it and record the IP addresses. These addresses can be allowed temporarily similarly to above, with individual cancel buttons. IP addresses can be whitelisted manually using the text field.

A default whitelist can also be configured.

The DLNA server serves up folders for all scenes, performers, studios, tags, movies and ratings. Here is a screenshot in VLC:

image

Where many scenes are present, it presents pages instead:

image

It gives a prefix for ten entries, since these were expensive to calculate.

I was able to test successfully in VLC and on my LG TV. **I was unable to get it working for Kodi** - and this seems the same for XBVR, rclone and DMS.


bnkai:
Some feedback from the discord channel
```
After several failed attempts to configure my firewall correctly I finally was able to get the DNLA running on my Panasonic TV. 

4 things I noticed: 

1. HEVC files fail to play cause my TV doesn't have the codec and the files aren't transcoded
2. Accessing Performers leads to a server error on my TV
3. The correct number of studios is shown, but only the first 20 studios in alphabetical are displayed and then those repeat over and over again
4. No logs other than the Debug => SQL: Select's
```

From my testing using VLC in linux everything seems fine.
A minor annoyance is that by default the dlna server listens to all interfaces. That can be a bit annoying as  even though no client is accepted the presence of the server can be  leaked.
From what i can see setting the `IfName` in the config of the server  (instead of using "" which is and can stay as the default) should limit the dlna server to a single chosen interface. ( A list of interfaces can be retrieved by `net.Interfaces()` ).

WithoutPants:
Happy to report that Kodi support is now fixed.

> 1. HEVC files fail to play cause my TV doesn't have the codec and the files aren't transcoded

I'm going to defer transcode support for a later PR. It's an extra layer of complexity I don't want to include right now.

> 2. Accessing Performers leads to a server error on my TV

I've not been able to reproduce this locally.

> 3. The correct number of studios is shown, but only the first 20 studios in alphabetical are displayed and then those repeat over and over again

Nor this.

> 4. No logs other than the Debug => SQL: Select's

I've added some trace-level logging of the request and response for now.

> A minor annoyance is that by default the dlna server listens to all interfaces. That can be a bit annoying as even though no client is accepted the presence of the server can be leaked.
> From what i can see setting the `IfName` in the config of the server (instead of using "" which is and can stay as the default) should limit the dlna server to a single chosen interface. ( A list of interfaces can be retrieved by `net.Interfaces()` ).

I've added an interfaces list configuration option. I didn't want to add an extra graphql interface for listing network interfaces, so it's just a plain text list right now.

> Adding/Removing a default IP and then saving seems to work ok but then if you try to leave the page it triggers the `Unsaved changes. Are you sure you want to leave?` popup

This should now be fixed.

pkg/dlna/cds.go
```diff
@@ -333,8 +334,16 @@ func (me *contentDirectoryService) Handle(action string, argsXML []byte, r *http
 				if err != nil {
 					return nil, err
 				}
+
+				// http://upnp.org/specs/av/UPnP-av-ContentDirectory-v1-Service.pdf
+				// maximum update ID is 2**32, then rolls back to 0
+				maxUpdateID := int64(math.Pow(float64(2), float64(32)))
```
bnkai, line 340: using math and power to get 2^32 is not necessary
a simple left binary shift is enough
```golang
var maxUpdateID int64 = 1 << 32
```
[playground snippet](https://play.golang.org/p/eDsz5vkGxty)

pkg/dlna/service.go
```diff
@@ -43,50 +44,66 @@ type Service struct {
...
+	}
+
+	var tmp []net.Interface
+	for _, if_ := range ifs {
+		if if_.Flags&net.FlagUp == 0 || if_.MTU <= 0 {
+			continue
+		}
+		tmp = append(tmp, if_)
+	}
+	ifs = tmp
```
bnkai, line 78: Is a `tmp` interface slice needed to clean the ifs ?
Cant we remove the above code and just adjust line 61?
```golang
if if_ != nil {
  if if_.Flags&net.FlagUp == 0 || if_.MTU <= 0 {
 	continue
  }	
ifs = append(ifs, *if_)
}

			
```

bnkai:
Everything seems to work ok. Selecting an interface works fine and the `unsaved` popup issue is fixed also.

pkg/dlna/service.go
WithoutPants: Line 61 only applies where `ifNames` is not empty.

Tscherno:
Which port is needed in Docker?

peolic:
@Tscherno I can see `1338` in the code, I don't know if any other ports are needed.

bnkai:
The docker  needs the host network to run i would have a look in the discord channel for details <https://discord.com/channels/559159668438728723/642079454054711306/846241919985516570>

PR #1370, opened 2021-05-06
Remove unnecessary graphql fields

### LINKED ISSUE #1368: [Bug Report] Gallery Graphql Slow
**Describe the bug**
I don't know whether this is truly a bug, but for some time the performance of the gallery page has decreased significantly. The bottleneck is the CPU, however this wasn't always the case. Unfortunately I cannot pin down since when this occurs, but I think this might have been an issue since march.
To pin it down, I did the following:

- Look in the network tab -> Graphql request takes between 5 and 10 seconds, everything else is fast
- Look into Task manager of Server -> is maxed out for that time
- Look into Logs of stash (the Tab in Settings) -> The SQL Query is visible and takes a little less than a second

Therefore I am very positive that this is due to some process that takes unnecessarily long, especially because the same amount of processing is needed for loading scenes or images, these requests are more in the range of 200-400 ms.
My benchmarks were took from 40 results per page, the kind of display (list, grid) is irrelevant. 
Maybe I should add that the galleries are big cbzs, up to 250 Images per gallery. This shouldnt impact the performance if everything works fine, but since it doesnt, I thought it might be worth a mention.

**To Reproduce**
1. Go to 'DOMAIN/galleries'
2. Wait
3. Wait some more

**Expected behavior**
The page should load in a reasonable time in LAN, the graphql wait time should take no longer than about 500 ms on a reasonably sized system.

**Stash Version:**
bdac3522

WithoutPants:
Fixes #1368 

Removes fields from graphql queries that are unnecessary. 

Should result in improvements for queries in the following pages:
* galleries
* gallery detail
* image detail
* movie detail
* performers
* performer detail
* scene detail

Also renames `GallerySlimData` to `SlimGalleryData` for consistency.

### DIFF
--- a/graphql/documents/data/gallery-slim.graphql
+++ b/graphql/documents/data/gallery-slim.graphql
@@ -1,4 +1,4 @@
-fragment GallerySlimData on Gallery {
+fragment SlimGalleryData on Gallery {
   id
   checksum
   path
@@ -10,16 +10,31 @@ fragment GallerySlimData on Gallery {
   organized
   image_count
   cover {
-    ...SlimImageData
+    file {
+      size
+      width
+      height
+    }
+    
+    paths {
+      thumbnail
+    }
   }
   studio {
-    ...StudioData
+    id
+    name
+    image_path
   }
   tags {
-    ...TagData
+    id
+    name
   }
   performers {
-    ...PerformerData
+    id
+    name
+    gender
+    favorite
+    image_path
   }
   scenes {
     id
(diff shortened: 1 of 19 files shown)


WithoutPants:
I have optimised a small number of joins to improve performer and tag resolver queries.

With all of these changes, with my contrived database, I got the following page improvements:

galleries - 12.36s -> 81ms
performers - 1.83s -> 641ms
tags - 1.7s -> 156ms

bnkai (approved):

PR #1374, opened 2021-05-08
Discard null values from scraper results

bnkai:


WithoutPants:
This violates the interface for `StrUnique`. Prefer a separate function for filtering empty strings, or handle in the client code.

PR #1379, opened 2021-05-13
Job queueing

WithoutPants:
Major refactor of the job management system.

Jobs are now queued. The Tasks page now shows a job queue. When empty, it looks like so:

image

This empty space is fixed so that adding jobs to the queue does not make the rest of the page jump around with job items. This area is contained so that a scrollbar is shown if there is an overflow.

Here is a shot with the queue filled:

image

The first job has been cancelled and is no longer running. The second job shows the spinning cog icon to indicate that it is running. The latter two jobs are waiting to be run. Not shown here is a finished job shows a green tick icon, and a job that is stopping shows a red spinning cog icon. I'm not married to these icons/colours, so let me know if there's something better to use.

Jobs can show sub-tasks, as shown below:

image

Clicking the buttons in the task page adds jobs to the queue.

The performer tagger has been updated to accomodate this, but there is a limitation that if you navigate away and come back to the page, it won't detect the previously queued job.


bnkai:
Looks ok and tests ok ( merging to dev is necessary for more externsive testing )

It should be mentioned somewhere that tasks that rely on configuration options (eg number of parallel tasks, encoding presets, ...) will use the option set during their execution not addition to the queue
Should the `backup` task be added to the queue? As it is now ( dev + pr ) the backup task can be run even if some other task is running (eg scan). The backuped db will be ok it might include though more or less data (scan vs clean) than what the user expects. This could be deferred to a future PR if needed.

PR #1381, opened 2021-05-13
This adds extra libraries required by scrapers

Tweeticoats:
This adds extra libraries required by the IAFD scraper: requests-toolbelt lxml cloudscraper
This installs pip in the prep environment, uses this to install cloudscraper then copies this to the final application container.
Pip installs this by copying files to /usr/local/lib/python3.8/dist-packages and this directory is copied to the new container.

cloudscraper is pure python so an alternate method would be to just use wget and extract this to that directory but if there are other libraries that are needed it might be useful to have pip available.

### DIFF
--- a/docker/ci/x86_64/Dockerfile
+++ b/docker/ci/x86_64/Dockerfile
@@ -8,9 +8,12 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm/v6" ];   then BIN=stash-pi; \
     elif [ "$TARGETPLATFORM" = "linux/amd64" ];  then BIN=stash-linux; \
     fi; \
     mv $BIN /stash
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt update && apt install -y python3 python-is-python3 python3-requests python3-requests-toolbelt python3-lxml python3-pip && pip3 install cloudscraper
 FROM ubuntu:20.04 as app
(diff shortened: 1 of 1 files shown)


WithoutPants:
Is there any reason we can't just have pip in the final container and do it there, or just remove pip afterwards if needed? Seems odd to install python etc into the prep container and the final one.

Tweeticoats:
Doing it this way will produce a smaller docker container than doing it in the final container as every step you run produces a new file system layer.
Using prep to set up a pip environment and just copying the result means you only need the output and should only need to add ~3mb to the final image size.
Installing pip will install gcc, perl, python headers etc and this will add ~240mb to the final image so I'm trying to avoid that.

Cloudscrape is a pure python package, so all pip is doing is downloading and extracting this to /usr/local/lib/python3.8/dist-packages. As it is pure python we could just run wget and extract it but using pip makes it more repeatable. Other python packages might have parts written in c and these will be compiled to a shared library and put in the same directory as the python files so copying this directory should be all we need from prep container.

Installing python3-requests-toolbelt python3-lxml through apt is personal preference, this could easily be done through pip.
As I wanted to use apt to install these 2 libraries I want the prep to also have these libraries there so pip knows what is installed.

Making this change will make build times a little longer.
Installing ffmpeg in the app container takes ~ 5 minutes, installing pip might take ~6 minutes but as prep and app are built in parallel it should not be that bad.

PR #1392, opened 2021-05-15
Swapped to buttons for UI consistency. (not sure if there was a reason for it to be <form?

jeremymeyers:


bnkai:
I don't think `button` is an appropriate solution. For consistency i think we only use buttons for actions.

PR #1406, opened 2021-05-20
UI filter refactor

WithoutPants:
This is a significant refactor of the UI filter code. The purpose is to try to reduce boilerplate and make it easier to add new filter criteria.

Included are the following changes:
* filter criteria and sort by options are now internationalised and displayed in alphabetical order
* modifier strings are now internationalised as well
* `en-GB` is formalised as the default language, meaning that it will display from `en-GB.json` if the message ID is not found in the chosen language. `en-US.json` should now only contain overrides to values in `en-GB.json`. I've added a `README` file to document this. The json files are now sorted alphabetically as well.


ui/v2.5/src/models/list-filter/criteria/criterion.ts
```diff
@@ -358,18 +255,24 @@ export class NumberCriterion extends Criterion {
...
     this.inputType = "number";
+  }
+}
 
-    if (parameterName) {
-      this.parameterName = parameterName;
-    } else {
-      this.parameterName = type;
-    }
+export abstract class ILabeledIdCriterion extends Criterion<ILabeledId[]> {
```
gitgiggety, line 266: While working on #1397 (which I'm rebasing on top of this branch) I'm wondering if it would be possible to move some methods from the specific criterions to this base class. All seem to have `encodeValue` being identical to:
```ts
  public encodeValue() {
    return this.value.map((o) => {
      return encodeILabeledId(o);
    });
  }
```

And also the fields are pretty common:
```ts
  public modifier = GQL.CriterionModifier.IncludesAll;
  public modifierOptions = [
    Criterion.getModifierOption(GQL.CriterionModifier.IncludesAll),
    Criterion.getModifierOption(GQL.CriterionModifier.Includes),
    Criterion.getModifierOption(GQL.CriterionModifier.Excludes),
  ];
  public options: IOptionType[] = [];
  public value: ILabeledId[] = [];
```
with the only difference being in `IncludeAll` being a possible modifier or not (and thus changing the default as well). So this might be added to `ILabeledIdCriterion` as well by giving it a constructor override and passing it an `includeAll` boolean argument to either add it or not. Descendants are already implementing a default / parameterless constructor anyway and calling to the parents constructor to set the `CriterionOption`, so adding a `true` / `false` argument shouldn't be an issue.

WithoutPants:
Good pickups, thanks!

ui/v2.5/src/models/list-filter/criteria/criterion.ts
```diff
@@ -257,47 +119,69 @@ export abstract class Criterion {
...
+  }
 }
 
-export class CriterionOption implements ICriterionOption {
-  public label: string;
-  public value: CriterionType;
+// used in various places, to eliminate the need to disable eslint rules
+// every time its used.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type CriterionAny = Criterion<any>;
```
InfiniteTF, line 166: You can use `Criterion<CriterionValue>` rather than having to use `any`.

bnkai (approved):

ui/v2.5/src/models/list-filter/criteria/criterion.ts
WithoutPants: Awesome, thanks.

PR #1412, opened 2021-05-22
Tag aliases

### LINKED ISSUE #121: Tag aliases
Per [Tag Improvements on the road map](https://github.com/stashapp/stash/wiki/Roadmap#tag-improvements), "Tags should be able to have sub-tags [...] this still needs to be fleshed out more", and as discussed on #115 and #88 inherited tags.

I don't know if what requires fleshing out is what functionality this would bring to the user or how it would be implemented on the backend.

### What are tag aliases?
As originally discussed from the Ruby project, tag aliases would act as an alternate set of tags for any primary tag [primary meaning any tag that is currently used in Stash]. This feature request was tied to #6 Tagging based on file name. With that original relationship in mind, the requested application of tag aliases was to use similar terms to describe the primary tag as a means to make file-name based tagging a lot more thorough.

But they can also be used for manual tagging with keywords that are used interchangeably to mean the same thing. For example tagging something with the words _tiny_, _petit_, or _small_ is basically the same for many people. We don't need primary tags of all of them if we're using them to mean the same thing. But once you develop a long list of tags it's difficult to remember the exact keyword you have used for something. Tag aliases would alleviate this problem, remove tag clutter, and help with organization.

A description of this in pseudo-code would be,

```
"tags": {
	"primary tag": "car",
		"tag alias": [
			"automobile",
			"vehicle"
		]
	}, {
	"primary tag": "threesome",
		"tag alias": [
			"3some",
			"3-some",
			"menage-a-trois",
			"menage a trois",
			"mmf",
			"fmm",
			"ffm",
			"mff"
		]
	}
}
```
The above `tag alias` contains alternate terms of the `primary tag`.

#### Example functionality
Let's suppose the following filenames that Stash is going to scan and auto-tag (when and if that gets implemented):
```
Scanning new files...
"threesome.mp4" : Primary tag found in filename. Scene auto-tagged with "threesome"
"milf.thresome.mp4" : Primary tag found in filename. Scene auto-tagged with "threesome"
"horny.blondes.ffm" : Tag alias found in filename. Scene auto-tagged with "threesome"
"dp_mmf_mature.mp4" : Tag alias found in filename. Scene auto-tagged with "threesome"
```
Without a `tag alias` Stash would have only picked up half of those filenames. With the tag alias, Stash identifies keywords belonging to the primary tag, and attributes the primary tag only to each filename.

#### Conflicting primary and tag alias names
What if primary tag uses the same name as a tag alias of another primary tag? Or if two different primary tags contain a tag alias of the same name? I gave this _some_ thought and think any tag, whether primary or an alias, should have a unique name.

I suspect there is the potential for many awkward usability issues if allowing duplicate tag alias names that might make the program and code unnecessarily complex, but this isn't my forte so maybe there's a good solution and use-case examples of where duplicates would be desirable.

#### Where does the user create tag aliases?
Obviously tags would require their own page containing their data that can be edited, such as
- primary tag, renaming or deleting
- tag aliases, renaming, creating, and deleting
- maybe it would be a good idea to allow tag type conversion? converting a tag alias into a primary tag and converting a primary tag into a tag alias of _x_.

The creation of a tag edit page seems inevitable since #5 Tag thumbnails are on the roadmap. The UI could be similar to my concept for the Studio edit page here: https://user-images.githubusercontent.com/37937507/63980020-6a579b00-ca88-11e9-89ad-b46f63368c8a.png

#### Relationship between manual scene tagging and tag aliases
If the user is manually tagging a scene and they type the name of a tag that already exists as a tag alias what should the behavior of the program be?

(1) Stash should filter the list of tags based on the name (as it currently does), but should handle matching tag aliases differently to avoid confusion. Therefore, the filtered list should only display the primary tag of a user typed tag alias once there is a complete match of the tag alias rather than a partial one. For example lets say,
- there exists a tag alias named "automobile" associated with a primary tag called "car"
- the user is typing a tag name and so far they have partially typed the string "auto".
- the list will filter any primary tags containing the word "auto", as it currently does
- the filter list should **not** look for partial matches of tag aliases, thus it should not show the primary tag "car" based on a partial match of the word "automobile"
- the filter list should only show the primary tag "car" once there is a complete match for the tag alias "automobile"

My reasoning for this specificity is because otherwise the filter will be bogged down with too much irrelevant partial matches.

(2) Stash should follow the behavior proposed in (1) but cosmetically highlight the match of any associated primary tag in a subtly different colour as a visual indicator that it's different from the partial name matches on the filtered list.

(3) If the user types a tag alias and it's primary tag is applied to the scene, there should be a temporary (like 3 or 5 seconds) visual indicator (such as those already existing green informational popups that Stash uses) informing the user this was the intended behavior of the program, such as: "_{Tag alias name}_ is a tag alias of _{primary tag}_". This will help avoid any confusion since the user is basically typing a keyword, applying it, and then something different than what they typed is being applied.

(4) Or forgot option (1) completely, and just filter the list based on all primary and alias tags, but this should definitely warrant option (2) that cosmetically denotes any tag aliases in the list, and option (3) as once they submit their tag alias it will be converted into its primary tag.

#### Should Stash display tag aliases in any of the tag lists or grid views pages?
My opinion is no, it just increases clutter. Their purpose is analogous to an email alias that catches old, associated, or commonly mistyped addresses into one umbrella. Your email client doesn't need to show you all its aliases outside a settings page.

The only exception I can think of would be if there was a dedicated page to each primary tag analogous to #115 my studio page concept at, where a list of tag aliases could be displayed beneath the main tag heading and thumbnail or something like that.

### LINKED ISSUE #1418: [Bug Report] Editing a tag resets its `created_at` date
_Reported on Discord._
Editing a tag resets its `created_at` date to `0001-01-01T00:00:00Z`.

1. Create a tag.
2. Check its `created_at` date (database) - valid.
3. Edit the tag.
4. Check its `created_at` date again - it is now `0001-01-01T00:00:00Z`.

**Stash Version:** `v0.7.0-14-g33999d3e`
**Desktop:**
 - OS: Windows 10
 - Browser: Firefox 88.0.1

WithoutPants:
Resolves #121 
Fixes #1418 

Adds tag aliases to tags. Tag aliases are unique across all tags.

The tag details/edit pages have been refactored and now use formik.

Tag aliases are displayed in the details page like so:

image

And are edited like so:

image

Aliases show in the tag select if the alias is fully typed in (case-insensitive). This means that you won't see partial matches on tag aliases. Tags that are aliases show in the select like so:

image

Tag aliases are matched during auto-tag, scraping and scene filename parsing operations.


alexhk:
That was fast!

Consider adding "tag links" or sub-tags at some point in the future, so you can add existing tags to tags themselves. Hear me out ;)

Example:

Tag "Japanese Femdom" could be linked to other existing tags: 
- Japanese
- Femdom

If the tag "Japanese" as well has linked tags, such as "Asian", adding only the tag "Japanese Femdom" automatically adds the tags to the Scene:

Japanese Femdom
Japanese
Femdom
Asian

That's just an example. But this would make a library extremely well organized on a very short time.

WithoutPants:
Tag hierarchies should be raised as a separate issue. It was discussed pretty extensively in #121, but it really needs its own issue.

bnkai:
- the (alias) next to the tags seems to appear in `random` and not only for tags that are marked as aliases (both in the plain edit scene/performer/... page  as well as the multi select edit).
- updating a tag to add an alias doesnt change the UpdatedAt field 
- tag aliases dont seem to get `exported`

ui/v2.5/src/components/Shared/Select.tsx
```diff
@@ -454,22 +475,105 @@ export const MovieSelect: React.FC<IFilterProps> = (props) => {
...
+    });
+    setTagAliases(newAliases);
+    setAllAliases(newAll);
+  }, [tags]);
+
+  const TagOption: React.FC<OptionProps<Option, boolean>> = (optionProps) => {
+    const { inputValue } = optionProps.selectProps;
+
+    let thisOptionProps = optionProps;
+    if (inputValue && !optionProps.label.includes(inputValue)) {
```
gitgiggety, line 502: `.includes()` does case sensitive matching. This means that typing in an option in lowercase while the actual option has (some) uppercase letter(s), or the other way around, it won't match and it is assumed to be an alias. In other words: this is the explanation of @bnkai's comment about `(alias)` being added randomly. So both should be converted to either lower or uppercase to make the matching case insensitive (or use some other matching method which isn't case sensitive).

peolic:
From the looks of it (have not had the chance to try it out), this should also fix #1418.

gitgiggety:
I've been wondering if either this PR, or a subsequent PR for the same release / version should add support for merging tags. In other words: open a tag, click merge, select a different tag and update all usages of the current tag (on scenes, images, galleries, performers) to the selected tag, add the current tag (name) as alias to the selected tag, and delete the selected tag.

While testing this branch it is a lot of work to manually do these steps, so having a button for it greatly simplifies it. And when this feature is released and everything is "merged" it still might be a handy feature to have for when having scraped info from a site which created a "duplicate" tag which you want to merge.

Also been wondering about the alias matching in the filter. I guess not showing aliasses as match is a tradeoff between actually finding the alias name vs filtering the list instead of not removing that mutch when you start typing because just one or two letters will most likely match a lot of tags (either name or one of the aliases). So it might also work to match aliases from a certain input length? In other words: match an alias if either the input and alias are an exact match, or it's a partial match and the input length is at least X characters (X being 5 or something alike). Otherwise I think it will be to hard to actually find a tag by alias, because there is a higher chance of making a typo, or having to type in a long name fully. For example having a tag `tiny tits` with alias `small tits`, if you then start typing `small` you won't find it and you still need to type in the full name. While a longer input length obviously reduces the change of "false positive" matches / options.

bnkai:
As mentioned in discord the merge functionality is not included in this PR
```
No merge functionality
Merging would add a fair bit of dev overhead.
```
Can confirm all issues resolved including #1418 that @peolic mentioned above.

bnkai (approved):

WithoutPants:
Merge functionality can be raised as a later issue/PR. I didn't think the process was too onerous to do manually - filter scenes/etc by tag, select all, add new primary tag, delete old tag, then add alias to the primary tag.

Likewise, I'll see how tag alias filtering goes and can adjust it as needed.

gitgiggety:
No problem for not adding it (in this PR). I just meant it would IMHO be a nice feature to have when this is released. So I will start working on this and create a PR when it's ready.

And yes: it does sound simple to manually "merge" tags, but when the tag is used numerous times "Select all" won't work as it obviously won't include those on different pages and I have to increase the item count per page first. Add in having scenes and galleries and I need to do "a lot" of work to just merge a single tag.

PR #1414, opened 2021-05-22
Remove performer "is missing scenes" filter

peolic:
Can use "scene count equals 0" filter instead.
Tested.


pkg/sqlite/performer.go
```diff
@@ -335,9 +335,6 @@ func performerIsMissingCriterionHandler(qb *performerQueryBuilder, isMissing *st
 	return func(f *filterBuilder) {
 		if isMissing != nil && *isMissing != "" {
 			switch *isMissing {
-			case "scenes":
-				f.addJoin(performersScenesTable, "scenes_join", "scenes_join.performer_id = performers.id")
-				f.addWhere("scenes_join.scene_id IS NULL")
```
WithoutPants, line 340: Leave this in, but mark as deprecated instead.

WithoutPants (approved):

PR #1415, opened 2021-05-22
Deprecation Fix: Updated gqlgen server handler initialization

### LINKED ISSUE #1135: Update gqlgen initialization
The way we initialize gqlgen was deprecated in 0.11. https://gqlgen.com/recipes/migration-0.11/

EnameEtavir:
Only changed to the new initialization API.
Tried to make as little changes as possible to minimize risk of regressions.

Resolves #1135

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -14,7 +14,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/99designs/gqlgen/handler"
+	gqlHandler "github.com/99designs/gqlgen/graphql/handler"
+	gqlExtension "github.com/99designs/gqlgen/graphql/handler/extension"
+	gqlLru "github.com/99designs/gqlgen/graphql/handler/lru"
+	gqlTransport "github.com/99designs/gqlgen/graphql/handler/transport"
+	gqlPlayground "github.com/99designs/gqlgen/graphql/playground"
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
 	"github.com/gobuffalo/packr/v2"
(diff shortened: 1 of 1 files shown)


WithoutPants:
This looks to have broken the gql playground for me. Looks like introspection is disabled - you can't see the docs or schema.

The old `handler.GraphQL` function calls `srv.Use(extension.Introspection{})`. You may want to look at the old method and copy through all of the configs that it goes through, to minimise regressions.

I think this is the other difference: `srv.SetQueryCache(lru.New(cfg.cacheSize))` with cache size defaulting to `1000`.

You may also want to consider using `gqlHandler.NewDefaultServer` instead of `New`, since it looks like it uses a lot of the same defaults.

EnameEtavir:
Regarding the missing settings, my bad, I will update those.

Regarding the use of `NewDefaultServer`, I'm not too sure about that. Normally I'd expect to only have default config as part of that, but the way for example `AddTransport` is done, is that those are not unique, and there is no `RemoveTransport` as far as I can see. So I can't exactly overwrite them as I'd expect from other projects where I only get a default "struct", where I only overwrite those parts I want to have different from default.

The other question is if they mark changes to the defaults in there releases good enough so that we don't get surprised. But I have no experience on their releases and don't really stand on either position (for or against usage of the default) in that respect.

EnameEtavir:
I'm not too sure about 
```go
	srv.Use(extension.AutomaticPersistedQuery{
		Cache: lru.New(100),
	})
```
of the `NewDefaultServer` as `GraphQL()` had it like this
```go
	if cfg.apqCache != nil {
		srv.Use(extension.AutomaticPersistedQuery{Cache: apqAdapter{cfg.apqCache}})
	}
```
and I don't see that we had `apqCache` set previously.

WithoutPants:
That's fine, `New` is good enough.

InfiniteTF:
You can ignore AutomaticPersistedQuery. It's a minor optimization for large query strings, but we don't use it on the client so it doesn't do anything.

EnameEtavir:
I amended the commit with the QueryCache and the Introspection extension.

EnameEtavir:
Rebased on latest `develop` to resolve conflict with javascript plugin

WithoutPants (approved):

PR #1419, opened 2021-05-22
Fix: unhandled errors

EnameEtavir:
As these errors where not explicitly ignored with _, I made changes to make sure they bubble up.

### DIFF
--- a/pkg/api/resolver_query_find_image.go
+++ b/pkg/api/resolver_query_find_image.go
@@ -21,6 +21,9 @@ func (r *queryResolver) FindImage(ctx context.Context, id *string, checksum *str
 			}
 
 			image, err = qb.Find(idInt)
+			if err != nil {
+				return err
+			}
 		} else if checksum != nil {
 			image, err = qb.FindByChecksum(*checksum)
 		}
--- a/pkg/api/resolver_query_find_scene.go
+++ b/pkg/api/resolver_query_find_scene.go
@@ -19,6 +19,9 @@ func (r *queryResolver) FindScene(ctx context.Context, id *string, checksum *str
 				return err
 			}
 			scene, err = qb.Find(idInt)
+			if err != nil {
+				return err
+			}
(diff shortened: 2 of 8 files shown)


pkg/manager/task_stash_box_tag.go
```diff
@@ -162,6 +162,9 @@ func (t *StashBoxPerformerTagTask) stashBoxPerformerTag() {
 						return err
 					}
 					err = r.Performer().UpdateImage(t.performer.ID, image)
+					if err != nil {
+						return err
+					}
```
WithoutPants, line 167: This is redundant - error is returned below.

pkg/api/resolver_query_find_scene.go
```diff
@@ -19,6 +19,9 @@ func (r *queryResolver) FindScene(ctx context.Context, id *string, checksum *str
 				return err
 			}
 			scene, err = qb.Find(idInt)
+			if err != nil {
+				return err
+			}
```
WithoutPants, line 24: Redundant - error is returned below.

pkg/api/resolver_query_find_image.go
```diff
@@ -21,6 +21,9 @@ func (r *queryResolver) FindImage(ctx context.Context, id *string, checksum *str
 			}
 
 			image, err = qb.Find(idInt)
+			if err != nil {
+				return err
+			}
```
WithoutPants, line 26: Redundant: returned below.

WithoutPants (changes requested):

EnameEtavir:
I agree, they are redundant, but from a maintainability view they could help you when you refactor the code or add additional steps where you might not see that those errors were not handled in the scope where the originated.

If you still want it removed, I'll do so, just wanted to bring the point up as food for thought.

WithoutPants:
Decent point. I'll merge as is.

PR #1421, opened 2021-05-22
Add `CreatedAt` & `UpdatedAt` to all object resolvers

### LINKED ISSUE #1002: Expose updateAt and createAt in FindScenes graphql query
I have written a number of external helper scripts (eg, to integrate Stash content with Jellyfin / Emby).  It would be useful to be able to identify if/when a scene has been updated, so these scripts only work on scenes that are new or have been modified.  This data is stored in the sqlite database, but there is currently no way to access it via graphql.

peolic:
... and `FileModTime` to supported objects' resolvers.

Closes #1002.

I didn't bother with displaying it in the UI.


graphql/schema/types/gallery.graphql
```diff
@@ -9,6 +9,10 @@ type Gallery {
   details: String
   rating: Int
   organized: Boolean!
+  created_at: Time!
+  updated_at: Time!
+  file_mod_time: Time!
```
WithoutPants, line 14: Some galleries don't have a `file_mod_time` and return nil from the resolver. The graphql displays an error on these objects as a result.

graphql/schema/types/scene.graphql
```diff
@@ -39,6 +39,9 @@ type Scene {
   path: String!
   phash: String
   interactive: Boolean!
+  created_at: Time!
+  updated_at: Time!
+  file_mod_time: Time!
```
WithoutPants, line 44: Similarly, this throws an error if the scene `file_mod_time` is null, which can happen.

graphql/schema/types/image.graphql
```diff
@@ -6,6 +6,9 @@ type Image {
   o_counter: Int
   organized: Boolean!
   path: String!
+  created_at: Time!
+  updated_at: Time!
+  file_mod_time: Time!
```
WithoutPants, line 11: This should be optional as well.

WithoutPants (changes requested):
`FileModTime` is a `NullSQLiteTimestamp`, so these fields should be optional.

ui/v2.5/src/components/Changelog/versions/v080.md
```diff
@@ -9,6 +9,7 @@
 * Add `subtractDays` post-process scraper action. ([#1399](https://github.com/stashapp/stash/pull/1399))
 * Skip scanning directories if path matches image and video exclude patterns. ([#1382](https://github.com/stashapp/stash/pull/1382))
 * Add button to remove studio stash ID. ([#1378](https://github.com/stashapp/stash/pull/1378))
+* Add `CreatedAt` and `UpdatedAt` (and `FileModTime` where applicable) to API objects.
```
WithoutPants, line 12: New changes go at the top of the list.

peolic:
Not sure how I missed this. Thanks for reviewing!

WithoutPants (approved):

Issue #1126, opened 2021-02-15
[Feature] Delete file as config option


Donearm:
Hi,

I'm often in need to clean up images from galleries and clicking "delete" and then checking "delete file" every time can be time consuming. An option in the config file, false by default, to always also delete the corresponding file when clicking the delete button would save quite a lot of clicks.

echo6ix:
if this is added i think there should be a permanent visual notification, like a full-bleed red band under the menu bar that reads `file deletion mode enabled.` or something like that.

the default state of stash has been to prompt the user before a destructive file operation, and if someone forgets they have that enabled because there's no indication that that the prompt has been enabled... well, that would be bad.

Donearm:
That's fine, of course. I often download lots of images that don't need to be kept, so I nearly always want the "delete file" option enabled but I do understand I am sort of an exception here.

Alternatively, it would do also to have the checkbox next to the delete button with a css snippet. I'm bad at css so I couldn't come up with anything to move the checkbox next to the button, do you know how to do it?

sesors:
I'm new to Stash and the first thing I checked was "How fast can I delete files". Currently it needs 4-5 clicks: (Open Sidepanel), Click "Edit", Click "Delete", "Checkbox" and second "Delete". This is way too much.

I would really prefer a one-click method, along with the option to bind a keyboard/mouse5 shortcut.

A trashbin or customized delay would be a plus but I can live with the risk of a missclick deletion.

PR #1424, opened 2021-05-24
Set DeleteFile to default = true

Tscherno:
It just sucks to have to tick it EVERY single time you want to use duplicate checker to delete the duplicates.

### DIFF
--- a/ui/v2.5/src/components/Scenes/DeleteScenesDialog.tsx
+++ b/ui/v2.5/src/components/Scenes/DeleteScenesDialog.tsx
@@ -29,7 +29,7 @@ export const DeleteScenesDialog: React.FC<IDeleteSceneDialogProps> = (
   const messageId = plural ? pluralMessageId : singleMessageId;
   const message = plural ? pluralMessage : singleMessage;
 
-  const [deleteFile, setDeleteFile] = useState<boolean>(false);
+  const [deleteFile, setDeleteFile] = useState<boolean>(true);
   const [deleteGenerated, setDeleteGenerated] = useState<boolean>(true);
 
   const Toast = useToast();


WithoutPants:
Changing the default behaviour to such a destructive action is dangerous imo. The default should be configurable.

bnkai:
Is a configuration option needed for that?
Isn't the multiselect to the left enough for these cases?
The user can tick whatever he wants to delete, click on the recycle bin and only confirm once.
multi-select

WithoutPants:
IMO, yes absolutely. 

There's two issues involved here:

The first is that we are changing the default behaviour. Current users are not accustomed to the Delete button deleting files from the file system. Silently changing the default value of the tick box exposes these users to very serious consequences if they are not careful when deleting. I acknowledge that its most likely that most users are ticking the box to delete the file anyway, but it is not necessarily all users. I also acknowledge that it is clearly communicated with the tick boxes, but it's a change to what may be habitual behaviour.

The second is that destructive filesystem actions should not be taken lightly by the app. For the most part, stash works as a hands-off supplement to a users filesystem, not integrated with it. Deleting files is not easily undone. We should be striving to limit the risk of users accidentally damaging their collections. If a user actually wants to delete their files, we should be making absolutely sure that they want to do it.

My suggestion would be an additional tick box in the dialog - "Use these options by default when deleting" - or something more succinct. Ticking it would set configuration keys for the two delete options and these would be used as the defaults in subsequent delete operations.

bnkai:
What I meant above is that IMO changing the specific default should NOT be allowed even with a config option. There is no reason imo to allow such a destructive action without warning the user ( even if the user wants to skip that)

The user mentions ticking EVERY time, isn't that avoided by using the multiselect ? Is there a use case where that isn't enough?

Tscherno:
Honestly, I don't think deleting without deleting the file makes sense because of scanning will add it again. I think everybody is checking that box. You currently need three clicks every time to that action. I want to reduce it for the most used part to two. So from an UX perspective you still have to click twice and see what you're doing, before really doing it so. You are warned every time!

bnkai:
Removing only the generated makes sense as you might want to keep the file. You can always move it to a different non stash library folder or exclude it from scanning.

Tscherno:
> 
> 
> Removing only the generated makes sense as you might want to keep the file. You can always move it to a different non stash library folder or exclude it from scanning.

Yes but then wouldn't it make more sense to that with Maintenance? To remove everything which is related to the missing files including the generated one?

bnkai:
In my case I don't run a clean unless I need to. I would need to do a dry run first , check if something doesn't look ok and then do the actual clean. Why do that for things I already could delete properly. If you don't do do a dry run first there is a use case where if you use USB drives for your library and the drive isn't mounted properly you'll lose meta
 Imo maintenance is for picking up some leftovers not for doing the actual deletion. I would never run a clean without a dry run first.

Tscherno:
It's just bad UX. Setting an checkbox every time for the most used usecase makes no sense.

WithoutPants:
> What I meant above is that IMO changing the specific default should NOT be allowed even with a config option. There is no reason imo to allow such a destructive action without warning the user ( even if the user wants to skip that)
> 
> The user mentions ticking EVERY time, isn't that avoided by using the multiselect ? Is there a use case where that isn't enough?

The user is still has to confirm the delete, so the user is warned either way. Allowing them the option to opt in to use the more destructive default action is reasonable.

> It's just bad UX. Setting an checkbox every time for the most used usecase makes no sense.

You've been given an alternative use-case. The consequence for someone who does not normally delete the file forgetting to clear the checkbox is far more disastrous than for someone who does normally delete it forgetting to check it. In addition, I've provided an alternative suggestion that should meet your requirements. In its current state, I will not merge this PR.

WithoutPants:
Related to #1126

PR #1429, opened 2021-05-25
Use formik for scene edit

### LINKED ISSUE #1012: [Feature] Alert user when closing the browser window of an unsaved (but modified) scene.
As there's currently no real indication that the edit tab on a scene has been modified but not saved, I propose implementing a JS pop-up stating something along the lines of "You haven't saved your modifications. Are you sure you want to exit?".

This helps to avoid scenarios where you've overlooked that you've added tags, etc and Stash completely fails to let you know what the current status of that edit page is.
Thanks! 🙏

WithoutPants:
* Uses formik for scene edit. 
* Prompts if leaving the scene page without saving. 
* Disables the save button until form is dirtied - this is applied to other formik-enabled forms as well (movie, performer, DLNA settings).
* Fixes bug introduced where it was not possible to unset rating on multiple pages.

Resolves #1012


bnkai:
The movie edit has an issue with the save and images.

- scraping only the image part doesnt `dirty` formik
- you can't clear front/back images and then save

bnkai (approved):

PR #1447, opened 2021-05-27
Add support for merging tags

### LINKED ISSUE #918: [Feature] Merge tags
Sometimes we realize we have multiple redundant tags. Include the ability to merge/consolidate two or more tags into one unified tag.

Navigating to the `tag page` > a tag > `edit` button > there could be a `merge` button.
- A text form field will allow the user to type from a list of existing tags
- The user clicks an `okay` button once they've typed any tags they want merged
- This opens a modal window (comparable to the scene scraper window) to confirm the list of tags and values that are being merged and and prompts the user with the new tag name
  - These values include: default tag name, thumbnail picture, and in the future can include any tag aliases (aka "sub tags")

gitgiggety:
Add support for merging tags. Merging tags means that all usages of the source tag(s) will be updated to use the defined destination tag. Aliases set on the source tag(s) will be added to the destination as well and finally the name of the source(s) will also be added as alias to the destination. Afterwards the source tag(s) will be removed.

Fixes #918

### DIFF
--- a/pkg/sqlite/tag.go
+++ b/pkg/sqlite/tag.go
@@ -536,3 +536,63 @@ func (qb *tagQueryBuilder) GetAliases(tagID int) ([]string, error) {
 func (qb *tagQueryBuilder) UpdateAliases(tagID int, aliases []string) error {
 	return qb.aliasRepository().replace(tagID, aliases)
 }
+
+func (qb *tagQueryBuilder) MergeTags(source []int, destination int) error {
+	inBinding := getInBinding(len(source))
+
+	args := []interface{}{destination}
+	for _, id := range source {
+		if id != destination {
+			args = append(args, id)
+		}
+	}
+
+	if len(args) == 1 {
+		return nil
+	}
+
+	tagTables := map[string]string{
+		scenesTagsTable:      sceneIDColumn,
+		"scene_markers_tags": "scene_marker_id",
+		galleriesTagsTable:   galleryIDColumn,
+		imagesTagsTable:      imageIDColumn,
+		"performers_tags":    "performer_id",
+	}
+
+	tagArgs := append(args, destination)
+	for table, idColumn := range tagTables {
+		_, err := qb.tx.Exec(`UPDATE `+table+`
+SET tag_id = ?
+WHERE tag_id IN `+inBinding+`
+AND NOT EXISTS(SELECT 1 FROM `+table+` o WHERE o.`+idColumn+` = `+table+`.`+idColumn+` AND o.tag_id = ?)`,
+			tagArgs...,
+		)
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err := qb.tx.Exec("UPDATE "+sceneMarkerTable+" SET primary_tag_id = ? WHERE primary_tag_id IN "+inBinding, args...)
+	if err != nil {
+		return err
+	}
+
+	_, err = qb.tx.Exec("INSERT INTO "+tagAliasesTable+" (tag_id, alias) SELECT ?, name FROM "+tagTable+" WHERE id IN "+inBinding, args...)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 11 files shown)


gitgiggety:
I'm opening this PR to 1. Show that I'm working on this feature so there won't be two people implementing it, and 2. most importantly, to discuss the UI part of it.

My current thinking about adding this to the UI is adding a new Operations tab just like the Performers page has. The Auto tag button could then be added to this tab. Furthermore a "Merge" button could then be added, when clicking this button either a model containing a source and destination filter could then be shown, or showing these two fields on the tab itself (ie: clicking the Merge button will show the field either above or below it). After filling in one or more sources and a single destination the could then be merged by making the API call which I already implemented.

I'm thinking about adding the Operations tab as as far as I can see it's not possible to add an additional button next to the Edit / Auto tag button. Adding a merge button to the edit form might be possible (didn't check) but that could lead to odd situations where a user has to click edit, can change values (or add aliases) but without saving first could then start the merge process, which I believe would be buggy. So moving it to a new Operations tab, which also already exits for Performers (only containing the Auto merge button), seems like the best option to me.

stg-annon:
> After filling in one or more sources and a single destination

My initial thought would to be have two separate options to avoid confusion, where one option would open up a multi tag field like that of the scene menu, with all the tags you want to merge into the current tag, and the other option being a single tag selection to merge the currently selected tag to the selection

AdultSun:
If you haven't already, I'd check out how Stash-Box / StashDB does it. You can request edit access through Discord if you need it, just to see what merging tags looks like over there. I don't think Stash's method would need to look identical to Stash-Box but having some consistency between the two may be helpful.

gitgiggety:
Forgot to comment yesterday. I've implemented it as @stg-annon suggested with two individual buttons. "Merge From" and "Merge Into"
So I believe this PR is ready now. Did some testing on it, including fixing duplicate tags when both tags would be set on the same media (so if a scene has tag "source" and tag "destination" and then merging them it would at first apply tag "destination" twice), preventing destination to be (part of the) source, etc.

pkg/sqlite/tag.go
```diff
@@ -536,3 +536,63 @@ func (qb *tagQueryBuilder) GetAliases(tagID int) ([]string, error) {
 func (qb *tagQueryBuilder) UpdateAliases(tagID int, aliases []string) error {
 	return qb.aliasRepository().replace(tagID, aliases)
 }
+
+func (qb *tagQueryBuilder) MergeTags(source []int, destination int) error {
```
WithoutPants, line 540: This function should be moved into the `tag` package and use the Reader/Writer interfaces to perform the operation. The reader/writer interfaces are intended for low-level operations, and this function should be possible to do outside of the data layer.

It should also have an accompanying unit test.

WithoutPants (changes requested):
The GUI change makes it inconsistent with the studio page, and is kind of a mix of styles of both performer and studio (buttons and tabs).

My preference would be - instead of having the operation tab - to add a dropdown button - Merge - to the button bar. The dropdown would have the Merge From/To options. I think `DetailsEditNavbar` needs to be refactored to allow adding custom elements, rather than bloating the input properties. Let me know if you need help doing something like this.

gitgiggety:
To be honest I'm uncertain about the requested changes. It means almost rewriting it from scratch again (as both the UI need to be changed, and the backend needs to be rewritten as well, so the only remaining part would be the GraphQL definition).
And with my limited React knowledge (hadn't used it before) I also don't know if I can get it to work in a reasonable timeframe.

And honestly speaking, as merging tags for me would be more like a one time operation I could just use this code / branch temporarily and then switch back to the normal releases. So I personally don't feel it's worth my effort to write it from scratch again.

So feel free to take over this branch or close the PR and start from scratch.

WithoutPants:
Continued in #1481

PR #1466, opened 2021-06-02
Spanish Translation

kermieisinthehouse:
Spanish translation + Alphabetize the language options


Still34:
I would prefer if the zh-TW option is untouched, otherwise use the standard i18n language tag `繁體中文 (台灣)` instead.

kermieisinthehouse:
> I would prefer if the zh-TW option is untouched, otherwise use the standard i18n language tag `繁體中文 (台灣)` instead.

I've removed it, thanks

Issue #1469, opened 2021-06-02
[Bug Report] Seg fault on new system


Still34:
**Describe the bug**
A seg violation occurs on a system where the config file is yet to be created.

```
WARN[2021-06-02 22:19:48] config file not found. Assuming new system...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1185810]

goroutine 1 [running]:
github.com/stashapp/stash/pkg/plugin.(*Cache).RegisterGQLHandler(...)
        /home/still/stash/pkg/plugin/plugins.go:47
github.com/stashapp/stash/pkg/api.Start()
        /home/still/stash/pkg/api/server.go:187 +0xeb0
main.main()
        /home/still/stash/main.go:19 +0x3c
```

**To Reproduce**
1. Ensure `~/.stash/config.yml` does not exist
2. Start stash (latest `develop` build)
3. Crash

**Expected behavior**
...for it to not crash.

**Screenshots**
N/A

**Stash Version: (from Settings -> About):**
`stash version: v0.7.0-40-g1bb5de12 (1bb5de12) - 2021-06-02 22:11:23`

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Debian 11 (WSL)
 - Browser [e.g. chrome, safari] N/A
 - Version [e.g. 22] N/A

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone6] N/A
 - OS: [e.g. iOS8.1]  N/A
 - Browser [e.g. stock browser, safari] N/A
 - Version [e.g. 22] N/A

**Additional context**
The build was compiled locally on the system.

PR #1470, opened 2021-06-02
Fix PluginCache initialisation

bnkai:
More of a hack/quickfix probably than a proper fix for #1469 
Feel free to edit/close if needed
The plugin reload in the config was added so that after the initial setup the plugins folder gets loaded properly (otherwise you would need to do a reload plugins or restart stash)


WithoutPants:
Closing in favour of #1475

PR #1471, opened 2021-06-02
Incorporate i18n into UI elements

### LINKED ISSUE #566: [Feature] Wrap all text in react-intl for localization
To be able to translate the application, all text has to be wrapped in `<FormattedMessage>` or `formatMessage`. We may have to rethink the way we define messages to avoid problems with extraction. I suspect explicitly calling `defineMessage` for all text centrally in each component is cleaner than defining the text on use, particularly since we can't always use `<FormattedMessage>`.

Still34:
## Summary

This PR aims to incorporate the existing i18n implementation deeper into the UI elements, bringing i18n to the entire frontend. As a proof-of-concept, the zh-TW localization strings are also included in this PR.

## To-do

Elements with i18n implemented:
- [x] Performer (incl. actions, bio, etc.)
- [x] Tags
- [x] Button actions
- [x] Modal dialogs
- [x] Settings
    - [x] General
    - [x] Interface
    - [x] Tasks
    - [x] DLNA
    - [x] Tools
    - [x] Scrapers
    - [x] Plugins
    - [x] Logs
    - [x] About
- [x] Scenes (incl. editing)
- [x] Movies (incl. editing)
- [x] Markers

Resolves #566


ui/v2.5/src/locales/en-GB.json
```diff
@@ -0,0 +1,377 @@
...
+      "interface": "Interface",
+      "logs": "Logs",
+      "plugins": "Plugins",
+      "scrapers": "Scrapers",
+      "tasks": "Tasks",
+      "tools": "Tools"
+    },
+    "general": {
+      "cache_path_head": "Cache Path",
+      "db_path_head": "Databae Path",
```
kermieisinthehouse, line 127: Typo

Still34: Resolved.

WithoutPants (changes requested):
`MovieDetailsPanel` is missing internationalisation.

Apart from the review comments, this appears to test ok.

WithoutPants:
I've added the following to `en-GB`. This can be added to zh-TW json file in another PR:
```
"plugins": {
      "hooks": "Hooks",
      "triggers_on": "Triggers on"    
}
```

WithoutPants (approved):

Issue #1434, opened 2021-05-25
[Feature] Smart Playlists (Saved Filters)


kermieisinthehouse:
**Is your feature request related to a problem? Please describe.**
Content would be more accessible and discoverable if user-defined playlists collected content by filters. Plex has a similar dynamic playlist system that makes it extremely useful to easily see how much content a specific query returns in your library, and to quickly shuffle matches from a playlist. 

**Use cases:**
- You know you just want to shuffle from a collection of 10 studios of at least a certain resolution
- You know you want to repeatedly reference content under a certain group of tags
- You want to know if your collection is under- or over- representing a certain segment of content
- You want to a playlist of 'core' content that doesn't include content you only have for its novelty

**Describe the solution you'd like**
- [ ] On the scenes list page, a new button "Save to smart playlist" would be present, and would open a modal to save the current filters. A new table would save the name, filters object, and cache the current number of matches and total runtime (see performance considerations below). 

> Mockup Modal:
![Mockup Modal](https://user-images.githubusercontent.com/84814418/119572861-f05e5e00-bda2-11eb-9878-01042577df2c.png)

- [ ] A separate 'Playlists' list page would list all of the currently saved playlists and their information. Plex uses four tiled thumbnails from random playlist members, but whatever thumbnail is appropriate to stash's design is acceptable. The playlist item component would display the number of scene matches and their total runtime. Playlists should have a button for "viewing" -> opens up scene list page with filters applied; and a button for "shuffle" -> opens up scene view page to a random matching scene with shuffled queue. This page should also let you edit filter definitions and delete your smart playlists. 

> Plex Prior Art, showing playlists with no filters, a filter for only 1080p matches, 720p matches etc. Note that item counts and edit/delete buttons were on the thumbnail that has been blurred.
![Prior Art](https://user-images.githubusercontent.com/84814418/119573944-7cbd5080-bda4-11eb-94b2-ad5bcfa97fa3.jpg)

- [ ] The DLNA server should make these playlists available under a 'Playlists' heading by name.

**Performance Considerations**
There is a performance consideration on generating the runtime and scene count stats for the dynamic playlists on the fly. Plex's implementation caches these numbers into the smart playlist table at creation time, and updates them only when the playlist is opened or edited. This way, the playlist view page doesn't execute n queries that touch the entire index. A similar system might be implemented, perhaps also updating the playlists after a scan/clean operation, or as a task.

**Describe alternatives you've considered**
[...]

PR #1474, opened 2021-06-02
Saved filters

### LINKED ISSUE #99: [Feature] Allow user to set default scene page parameters in settings

### LINKED ISSUE #730: [Feature] Saved queries
It should be possible to save frequently used queries.

Having set query parameters for a given page, it should be possible to save the query parameters. These saved queries should then be easily accessible from a dropdown menu or something. It should be possible to set a saved query as a default query for a list page, such that if no existing query is in memory, it will use the saved query as the default.

It is possible to use bookmarks to save queries, but this is not a very discreet solution.

WithoutPants:
Includes a further refactoring of the UI filter system. I can separate out the saved filter stuff from the refactoring if necessary.

Adds a "Saved Filters" button to the search text field. Clicking the button shows a dropdown list of saved filters:

image

Typing into the "Filter name" field filters the saved filters displayed. Provided that the entered name is unique, the save button is enabled next to it, allowing the current filter to be saved. Existing filters can also be overwritten or deleted. Clicking on an existing filter applies it.

The Set Default button sets the default filter and view for the page. This is enabled only in the top-level pages (ie where `persistState` is `ALL`). When navigating to the top-level page, if no URL parameters are present (ie from clicking to top navbar buttons), then the default filter is applied. This supercedes the behaviour introduced by #326 that maintained the filter across all visits to the query page. Now, only the display mode is persisted in cookies for components using `persistState` of `VIEW`.

Resolves #99
Resolves #730 
Overlap with #1434


yoshnopa:
I hope this is the right place to propose this, otherwise this could be made an issue.
I think a saved filter is great, but it would be more useful if the sorting and maybe even the view type (grid/list) would be saved that way as well.
depending on the filter and the amount of results, other views or sortings may be necessary. That way, this could be preserved. On the other hand, those barely using views and/or sorting changes won't even notice this enhancement, so no drawback for other use cases.

WithoutPants:
> I hope this is the right place to propose this, otherwise this could be made an issue.
> I think a saved filter is great, but it would be more useful if the sorting and maybe even the view type (grid/list) would be saved that way as well.
> depending on the filter and the amount of results, other views or sortings may be necessary. That way, this could be preserved. On the other hand, those barely using views and/or sorting changes won't even notice this enhancement, so no drawback for other use cases.

Sorting and view type is saved in the saved filter.

kermieisinthehouse:
It seems to reflow strangely on mobile with the new button on the scenes page

WithoutPants:
Made some adjustments to the styling. I'm not super happy with the presentation of the filter UI in general, but I'll save that for another PR.

bnkai (approved):

PR #1481, opened 2021-06-04
Merge tags functionality

### LINKED ISSUE #918: [Feature] Merge tags

WithoutPants:
Supercedes #1447 
Resolves #918 

Adds a dropdown button to the Tag details page, allowing merging from other tags or into a destination tag:

image

image


kermieisinthehouse:
The new button seems to overflow incorrectly when horizontal space is constrained, especially on mobile.

Screenshot_20210607-071925_1

bnkai (approved):

jeremymeyers:
does this address the possiblity of a new scrape wanting to recreate a tag that has previously been merged? or does it just combine the associated entries together?

WithoutPants:
Merging adds the merged tag name as an alias to the destination, so it should address the scenario you describe.

PR #1489, opened 2021-06-07
Changes to Default Performer Images (for Accessibility)

### LINKED ISSUE #1396: [Feature] Accessibility - Ability to Change "No Image" Default Image
**Problem**
The no-image default image (stripes) is a migraine trigger and potentially an epileptic trigger for some users.

Either a plainer default image or the ability to change the default image for performers etc.

There are no workarounds as the images are embedded.

capnrowdy:
Fixes #1396  (some caveats, see below).

Added:
 - Accessibility Form Group created under "Interface"
 - Added Setting for "Use non-striped default performer images", with the requisite backend code.
 - When this setting is enabled, new performers that are created that require a default image will be generated from an accessible box.
 - A new folder for images in `static/performer_female_accessible` has been created with two images (sourced with as free restrictions as I could find as per IMAGE_LICENSES.md, but happy to change / be guided on this one.

Known Issues:
 - This does not retrospectively change existing performer images which do not conform. Not sure how to go about this, as I could not find the code paths responsible for pre-existing images? I suspect DB would be involved in this. Perhaps a task is needed?
 - There is no code path for male accessible default images as (currently) we do not require it, however I would imagine if we want to go for a stylistic consistency perhaps we should have `performer_male_accessible` as well.


WithoutPants:
I think that this is a bit of an over-complicated solution to the original problem.

We only need to provide a mechanism for the user to supply their own images.

My preference would be the following:
* add a configuration setting for "default performer image location". 
* this config setting would be passed to `getRandomPerformerImageUsingName`. If not empty, then it would read a random file from the directory, instead of from the box
* filenames should be read from the directory on the first read, then cached
* configured directory can have subdirectories for each gender, using the directory itself if the subdirectory can not be accessed
* fall back to the box images if an image cannot be found

This change _will_ change performer images for performers created more recently (I don't remember the commit that changed the behaviour) - if a performer does not have an image set, then it uses a random one generated from the name. This is resolved at request time, and is not persisted. Performers created a few versions ago had the image persisted - these will have to have the image cleared manually.

As an aside, we _can_ remove the stripes from the default images fairly easily. It will just take some manual time spent removing them with gimp or whatever. This would resolve the accessibilty issue altogether without the need to implement this feature (although this feature would still be useful). This _would_ mean that everyone's default images would change, but provided that the plain images don't look too bad, I think we can live with this for the purposes of being more accessible. Plus, we could provide the old images that people can use with this feature if necessary.

pkg/api/images.go
```diff
@@ -10,39 +9,43 @@ import (
...
+			// We need to set performerBoxCustom at runtime, as this is a custom path, and store it in a pointer.
+			newBox := packr.New("Custom Performer Box", customPath)
+
+			// only grab from this box if it is not empty
+			if len(newBox.List()) > 0 {
+				performerBoxCustom = newBox
+				box = newBox
+			}
+		}
 	}
```
bnkai, line 36: Shouldnt this be modified and included  in the initialise images function?
I am getting a lot of memory/cpu usage and the custom images not loading which i am guessing is due to boxes being created for every single randomperformer query
Having sorted my performers by recently added and 30+ performers not having images (so random is used) i am getting huge memory usage/leaks when the custom path is not null (be it an existing/correct path or a non exiting one)

WithoutPants:
I've updated packr so that we can utilise `Folder` which prevents trying to pack the custom folder. It now initialises on startup or if the directory changes. One thing that I noticed is that if you select a folder with a heap of files, it takes a very long time to initialise. Under normal usage it should be fine.

bnkai (approved):

PR #1496, opened 2021-06-11
Disable float-on-scroll behavior for jwplayer

### LINKED ISSUE #1440: [Bug Report] Picture in Picture takes over the screen on mobile
**Describe the bug**
When you’re watching a scene and scroll down, picture in picture mode triggers and the scene floats over the page. On mobile devices (especially small ones), it becomes impossible to view or interact with the page behind it.

**To Reproduce + Screenshots**
Example to reproduce the behavior:

In this example I started playing a vertical video. Being vertical it naturally fit the screen almost perfectly. I then scrolled down and the PIP activated, leading to this:

2B83C6A1-4E16-428C-AC90-A51BE7E488F9

Originally it was on the **Info** tab so there wasn’t much for it to cover. I had to move the PIP around the screen and tap the small sliver of the **Edit** tab to expand the screen enough to show the issue. The fact that it was that hard to switch tabs itself shows the issue fairly well though.

**Expected behavior**
When PIP activates it resizes down to a percentage of the screen size. This allows for it to be moved around as you do things like add metadata or handle markers.

**Stash Version: (from Settings -> About):** v0.7.0-33-gc70faa2a

 - Device: iPhone 12 Mini
 - OS: iOS 14.5.1
 - Browser: Safari

**Additional context**
This is most apparent on the iPhone Mini due to the small size of the device, but it happens across the board. Even on my iPad Pro 12.9” I’ve struggled with the PIP box being too large.

ThankGoshForDucks:
Fixes #1440

Narrowed down a possible solution to the bug causing mobile full screen in ScenePlayer to default to a floating window. I have built and tested that this fixes the fullscreen issue, but I'm not sure what this feature was originally enabled for. 

Before (notice the blurred background of floating window behavior):
121628884-f99c2b80-ca2e-11eb-8a13-40043d7f3b71 (1)

After:
image

[Float On Scroll](https://developer.jwplayer.com/jwplayer/docs/jw8-player-configuration-reference#float-on-scroll) configuration reference.

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
+++ b/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
@@ -296,9 +296,10 @@ export class ScenePlayerImpl extends React.Component<
       image: scene.paths.screenshot,
       width: "100%",
       height: "100%",
-      floating: {
-        dismissible: true,
-      },
+      // FIXME: Enabling this in jwplayer 8.11.5 causes full screen in mobile to default to picture-in-picture
+      // floating: {
+      //   dismissible: true,
+      // },
       cast: {},
       primary: "html5",
       autostart:


kermieisinthehouse:
Can confirm this fixes the behavior on mobile Firefox and mobile Chrome.

It seems there is some extra horizontal space in the main scene view page that makes it slightly scrollable, though.

WithoutPants:
This disables the floating player behaviour on _all_ platforms, which is not ideal. 

According to [here](https://developer.jwplayer.com/jwplayer/docs/jw8-css-skin-reference#floating-flag), we should be able to hide the floating player element on mobile with a css adjustment.

WithoutPants:
Closing in favour of #1721.

PR #1505, opened 2021-06-16
Add various filter criteria

### LINKED ISSUE #1092: [Feature] Title/Name Filter
With RegEx filters coming soon at the time of writing this, we need to have filters for Scene, Movie, Image, and Gallery titles as well as Performer, Studio, and Tag names with the option to use RegEx.

The alternative would be to have some sort of checkbox next to the search field to enable RegEx support right there.

WithoutPants:
Resolves #1092 

Adds the following criteria:

Performers: name, details
Scenes: title, details, oshash, checksum, phash
Movies: name, director, synopsis, duration, rating
Studio: name, details
Gallery: title, details, checksum
Image: title, checksum
Tags: name


bnkai:
The scenes  -> phash filter only works with the is null, not null criterion
The regex ones break the page while is/is not, includes/excludes don't work properly. I think it has something to do with how it is saved (compared to oshash/checksum)

bnkai:
Looks ok.
Should the `is not` criterion though include null phashes or not? (`scenes.phash != ? OR scenes.phash IS NULL`)
If you have say 100 scenes with 5 of them not having a phash yet the `is not` phash (for a unique phash ) will return 94 results and not 99?

WithoutPants:
I believe this is consistent with other filter criteria.

PR #1514, opened 2021-06-17
Include server host in plugin input

### LINKED ISSUE #1155: [Bug Report] Plugins should be provided with server host address
**Describe the bug**
Related bug: https://github.com/WithoutPants/stash-plugin-duplicate-finder/issues/4

Plugins are currently given no host address, so it is left to the plugin to determine the correct host address from the `config.yml` file. This should be provided by the `StashServerConnection` struct.

WithoutPants:
Resolves #1155 

Adds `Host` to the `StashServerConnection` structure provided to plugins.

### DIFF
--- a/pkg/plugin/common/msg.go
+++ b/pkg/plugin/common/msg.go
@@ -11,8 +11,8 @@ const (
 type StashServerConnection struct {
 	// http or https
 	Scheme string
-
-	Port int
+	Host   string
+	Port   int
 
 	// Cookie for authentication purposes
 	SessionCookie *http.Cookie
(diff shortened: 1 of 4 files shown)


gitgiggety:
Haven't checked how plugins work etc, but just wanted to drop my thinking about scraper scripts here. What I was thinking about for providing the hostname to scrapers was using an environment variable. This as it wouldn't break the currently defined API and scripts can just check if the environment variable exists and use it. Otherwise it would either need to be made available as some special thingy in the scraper definition so it can be send as argument, or be included in the JSON send using stdin (and thus potentially breaking the current API, and breaking semantics of the posted JSON which wouldn't just contain information about the thing to scrape but also information about the environment).

So possibly such a solution might also work or even be better for plugins as well? (If they're invoked as a script / on command line as well, which I'm unsure of).

bnkai (approved):

PR #1483, opened 2021-06-06
Fix Performer Studio filtering

WithoutPants:
Fixes #1482 

Applied the new hierarchical studio filtering for performer query. I've left the existing implementation in there for `depth = 0` since it seems more performant. I haven't spent a long time trying to improve performance. @gitgiggety if you get the chance could you take a look at the changes? I'll merge this in now so that at least the query should work correctly.

Also fixed issue with the UI studio hook that caused the wrong criterion modifier to be used.

PR #1515, opened 2021-06-17
Optimize studio filter on performers

gitgiggety:
Rewrite studios filter for performers so that it is a lot faster. Standalone queries have been tested by @WithoutPants as part of #1483 and show that instead of taking 2,5, to 3,5 seconds on a generated database it takes less than half a second (possibly even less than a tenth of a second).


gitgiggety:
Didn't have a huge amount of time to properly test it (but obviously did test it quickly) but it should work just fine. The eventual query does differ slightly because I kept the original queries order for FROM / JOIN but I don't think that makes a difference as the join is the same anyway, just the order of the tables changed. Also still changed the `LEFT JOIN` to an `INNER JOIN` as both sides of the join are required anyway.

PR #1517, opened 2021-06-18
DLNA refactor and support browse root object

### LINKED ISSUE #1465: [Bug Report] DLNA error on Samsung TV
**Describe the bug**

I have a Samsung TV, model UE40MU6400.

I can get the Stash DLNA server running fine and can browse/play from a VLC client.

The TV finds the server but then displays only "No Content Found". Whenever I hit the refresh button on the TV I see the following in the Stash log:

```
strconv.Atoi: parsing "/": invalid syntax
```

**To Reproduce**

1. Enable the DLNA server and whitelist the TV's IP address.
2. Go to the "Source" menu on the TV and select the server (which appears just fine)
3. Hit "Refresh" on the screen labelled "No Content Found" (error appears in Stash logs every time the button is pressed).

**Expected behavior**

I should be able to browse my content.

v0.7.0-39-g975343d2

 - OS: Linux
 - Browser: Chrome
 - Version: 91.0.4472.77

I hope that's helpful! I know you don't support these TVs yet but I wondered if maybe it's some incoming data you can skip over rather than crashing?

WithoutPants:
Rearranged CDS code a little.

Only functional change is in the `BrowseMetadata` handler. Code was assuming the object ID would be a path to scene ID. It was not handling the root object (which gets changed to `/`). 

Hoping this will fix #1465. Needs testing on Samsung TVs.


ctyjiuh:
Hi there. Do you need assistance in testing on Samsung TV?

Issue #1463, opened 2021-06-01
[Feature] Allow Tag Aliases to be shared 


stg-annon:
**Is your feature request related to a problem? Please describe.**
It would be useful in some cases for tags to share an alias, When scraping tags you can run into the case where a foreign might represent multiple local tags in stash, in which which case you would want to have that foreign tag to be an alias of multiple local tags, so that all matching local tags are added instead of the foreign tag.

**Describe the solution you'd like**
Allow tags to share aliases

**Describe alternatives you've considered**
Creating a separate tag that represents the combination of two other tags, this is not ideal

Stephan972:
I would see on the tag page, a list of tag aliases jsut like you can see aliases for one performer.

gitgiggety:
@stg-annon would this work for you with #1519 ? Using the tag hierarchy feature you can create a local combined tag and set both of the existing values as parent. When filtering you can then check the "Include sub-tags" option so searching for your existing tag will also show those items tagged with the "combined" tag.

stg-annon:
Yeah it seems effectively the same to me closing

stg-annon:
I've run into this issue again and rereading my comment I realize I did not have a very good explanation.

The solution I'm looking for cases when scraping a site and the tag they have is a combination of multiple tags in stash. Instead of importing the site's tag into stash I would rather the sites tag be mapped to the multiple tags in stash an apply those tags instead, in the same way a tag alias will map a site tag to a stash tag.

example:
say a site has a tag `Tattoos & Piercings` and stash has tags `Tattoos`,  `Piercings` I would like a way to automatically map the tag `Tattoos & Piercings` to both the tags `Tattoos`,  `Piercings`

I'm not sure that my original solution would be the best as just allowing tags to share aliases would probably cause a lot of confusion, I'd imagine there would need to be a way to define a "replacement tag" and then select what it will be replaced with that way you have a clear definition of what is going on

skier233:
Bump on this issue.

skier233:
Potential solution: Allow tags to be a sum of component tags. Then those tags could automatically be chosen when an aliases was found for all component tags. For example, you could have a tag called "wet pussy" and identify that this tag is a sum of the component tags "wet" and "pussy". Then, if you ran into a tag called "dripping vagina" it would still map to "wet pussy" if "dripping" is an alias for "wet" and "vagina" is an alias for "pussy".

stg-annon:
[...]

PR #1519, opened 2021-06-19
Tag hierarchy

gitgiggety:
I'm currently working on adding support for a hierarchy to tags where one tag could have multiple children and/or parents. This feature seems to have been requested a couple of times before, and with 0.8 adding support for tag aliases I might just as well try to add this as well and I guess tags would then have most of the desired features, or at least from management purposes.

As stated this branch is a work-in-progress, so I will list the things which I already implemented:
- [X] DB migration
- [x] Management
  - [x] Adding parent(s) / child(ren)
    - [x] Validation that no loops are created (i.e.: A => B, B => A should fail, and A => B, B => C, C => A should fail as well etc)
  - [X] Listing parent(s) / child(ren)
  - [x] Merging parents/childs when merging tags
- [X] Filters
  - [X] Scenes.Tags
  - [X] Scenes.PerformerTags
  - [x] SceneMarkers.Tags
  - [x] SceneMarkers.SceneTags
  - [X] Images.Tags
  - [X] Images.PerformerTags
  - [X] Galleries.Tags
  - [X] Galleries.PerformerTags
  - [X] Performers.Tags

For the management part (adding parent or child tags) I'm open to suggestions to how to add this to the UI. This applies to both the actual management, as listing the parents and children


pkg/sqlite/scene.go
```diff
@@ -609,56 +615,20 @@ func sceneMoviesCriterionHandler(qb *sceneQueryBuilder, movies *models.MultiCrit
...
-					(select performers_scenes.performer_id from performers_scenes
-						left join performers_tags on performers_tags.performer_id = performers_scenes.performer_id where
-						performers_scenes.scene_id = scenes.id AND
-						performers_tags.tag_id in %s)`, getInBinding(len(performerTagsFilter.Value))), args...)
-			}
-		}
-	}
-}
-
-func handleScenePerformerTagsCriterion(query *queryBuilder, performerTagsFilter *models.MultiCriterionInput) {
```
gitgiggety, line 641: @WithoutPants This method was added by you when you added support for performer tags, but to me it seems like it has been unused all the time, can you confirm this? (Obviously code still compiles fine, but I don't know whether it can be invoked dynamically by means of reflection or something alike).

graphql/schema/types/filters.graphql
```diff
@@ -90,12 +90,10 @@ input PerformerFilterType {
 }
 
 input SceneMarkerFilterType {
-  """Filter to only include scene markers with this tag"""
-  tag_id: ID
   """Filter to only include scene markers with these tags"""
```
gitgiggety, line 93: Removed this as I couldn't find any usages of it, neither by searching for `TagId` nor `tag_id` (with "match word" semantics, otherwise it obviously matches the plural form). And using `tags` instead should yield the same result anyway.

If this is still needed I will readd it.

gitgiggety:
Everything has been implemented except support for merging. But I'm a bit in doubt about this. This as I believe validating no loop will be created after merging is a real pain. So I'm thinking about 2, 3 options for this:
* Leave server code as is, which means relations set on sources will be removed. Always show a confirmation in the UI that all relations of the source will be gone after merging. (I.e. "Are you sure you want to merge these tags? *Note:* all relations set on the source tags will be removed")
* Same as above, but in the UI iterate over the source tags and check if they do have relations to other tags and only show the notification when that's the case. Notification could then also be more informative and list which source tags have relations to others.
* Validate server side that no relations exist on the source tag, and when they do exist return an error. This means the user has to manually remove all relations from the tags to merge.

WithoutPants:
This PR really needs/needed an accompanying RFC to discuss the conceptual design.

Why is it allowing multiple parents? What does this mean conceptually?
This seems to unnecessarily complicate the design and implementation, particularly regarding the schema, and merging of tags.

gitgiggety:
Allowing multiple parents makes sense to me as it allows combined tags. For example having tags `car`, `boat` and `fast` you can create tags `fast car` and `fast boat` which link to both tags. This would also solve #1463 as the remote tag can be created as local tag as well and linked to all (existing) local tags and searching for the "splitted" tag including children would thus also find that tag. I.e.: searching for `fast` including children would thus also find `fast car` and `fast boat`, searching for `car` with children would also find `fast car` etc, and you don't need to apply multiple tags. And this thus eases searching as and/or logic isn't supported. I.e.: with this you can search for "`fast car` or `fast boat`" (using `Includes`) while with only `fast` `car` and `boat` tags you can't search for tags "`fast` and (`car` or `boat`)".

And besides #1463 I do seem to remember I also read some other ticket / comment which requested support for this kind "combined tags" to create more specific tags (`fast car`) and be able to search for them based on the less specific tag (`fast` or `car`). But I can't seem to find that ticket anymore.

With regards to "this makes merging hard / can't be supported" while that's true I don't think the scenario of merging tags with lots of information set will be that common. I believe merging will mostly be used now with aliases being supported and being able to merge (accidental) duplicates / synonyms, and afterwards I do believe it will mostly only occur after accidentally adding new tags based on scraper results. So in these cases tags most likely won't have links to other tags. And besides: when merging tags the image of the source tag(s) are also ignored, and the same would apply to for example the requested feature of adding a description field to tags. So I actually don't think it's that bad to not support merging of links to other tags. (And thus just keeping the links set on the destination tag), as it's just one more property which isn't used in the merged result (just as image already is ignored during merge).

WithoutPants:
I'm guessing that you might be referring to the discussion in #121 

I can see the merit in adding a "tagging of tags" system as discussed there. It does seem needlessly convoluted and niche to me however, with little practical use cases. Anyway..

Does this system allow a tag to have an ascestor tag as its direct parent? ie `A <- B <- C` and `A <- C`, and if so why?

It's entirely reasonable to expect that the merge operation handles relationships to other tags. It's a reasonable expectation that source tag images are not retained when merging into another tag. Relationships can and should be handled.

I'd expect the following behaviour wrt to merging tag relationships:
* all parents of the source tags are added to the destination tag - any source parents that are already ancestors of the destination tag are skipped
* all children of the source tags are added as children to the destination tag - any children that are already descendents of the destination tag are skipped
* any situation which results in a tag being both parent/ascestor and child/descendent results in an error. 

Loop detection logic should be performed outside of the `TagReaderWriter` using its methods. I don't think `DetectLoop` belongs in there, and it should probably provide methods to get child and parent tags, with an option to get all descendents/ancestors. I think this will make it easier to validate post-merge loops.

gitgiggety:
Thank you for your input. You're right, I should rewrite the loop detection to code instead of SQL and that infeed would make it easier to also support the merge scenario.

About the `A <- B <- C` and `A <- C` question. This currently is possible. It isn't a feature per se, but does work just fine and shouldn't lead to any issues. But obviously can be resolved when already iterating over all ancestors anyway.

gitgiggety:
Working on this at the moment, but there are some (not so edge?) cases which I want to make notice of / discuss.

* `A <- B` and then merging `B` into `A` (so rolling up what was a separate tag before into its parent). I believe this should be possible as at some point you might decide that you don't actually need the child tag as it contains not that many items or whatever
* `A <- B`, `A <-C` and merging `C` into `B` (or the other way around). I don't believe this should result in an error as the result would be the same, there is one parent set which is `A`
* `A <- B <-C` and `A <- D` and merging `D` into `C`. Should this really result in an error? It is a strange scenario I guess, but should it lead to an issue / buggy behavior?
* `A <- B <- C` and `A <- D` and merging `D` into `B`. Should work, as setting `A` as parent should work fine / have the same end result. And in case of merging the other way around, `B` into `D`, should obviously result in `A <- D <- C` as well.
* `A <- B <- C` and `D <- B` and merging `B` into `C`. Should work and result in `A <- C`, `D <- C`
* `A <- B <- C` and `B <- D` and merging `B` into `C`. I guess this should work and result in `A <- C <- D`?

kermieisinthehouse:
Just wanted to chime in and say that I've been running with this patch for a few weeks and have had no problems in daily use with a complicated tag hierarchy

WithoutPants:
> Working on this at the moment, but there are some (not so edge?) cases which I want to make notice of / discuss.
> <snip>...

I agree with your findings here.

> * `A <- B <-C` and `A <- D` and merging `D` into `C`. Should this really result in an error? It is a strange scenario I guess, but should it lead to an issue / buggy behavior?

I'd expect this to result in `B <- D` and `A <- D`, which is unusual but assuming we're allowing this weird hierarchy, I suppose that's fine.

graphql/schema/types/filters.graphql
WithoutPants: For compatibility purposes, this should be left in. We can mark it as deprecated if it truly isn't needed, but for now keep it in.

pkg/sqlite/scene_marker.go
```diff
@@ -135,136 +143,96 @@ func (qb *sceneMarkerQueryBuilder) Query(sceneMarkerFilter *models.SceneMarkerFi
...
-		args = append(args, thisArgs...)
-	}
+func sceneMarkerSceneTagsCriterionHandler(qb *sceneMarkerQueryBuilder, tags *models.HierarchicalMultiCriterionInput) criterionHandlerFunc {
+	return func(f *filterBuilder) {
+		if tags != nil && len(tags.Value) > 0 {
+			addHierarchicalWithClause(f, tags.Value, "tags_for_scene", tagTable, "tags_relations", "", tags.Depth)
 
-	if tagID := sceneMarkerFilter.TagID; tagID != nil {
-		whereClauses = append(whereClauses, "(scene_markers.primary_tag_id = "+*tagID+" OR tags.id = "+*tagID+")")
-	}
```
WithoutPants, line 249: Ensure this filter criterion is handled since it should not be removed.

pkg/tag/update.go
```diff
@@ -63,3 +70,163 @@ func EnsureAliasesUnique(id int, aliases []string, qb models.TagReader) error {
 
 	return nil
 }
+
+func EnsureUniqueHierarchy(id int, parentIDs, childIDs []int, qb models.TagReader) error {
```
WithoutPants, line 74: This needs unit tests.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -498,7 +499,7 @@
   "image_count": "Image Count",
   "images": "Images",
   "images-size": "Images size",
-  "include_child_studios": "Include child studios",
```
WithoutPants, line 501: Please remove existing `include_child_studios` from the other localization files, so that we don't end up with vestigial entries.

WithoutPants (changes requested):
Static review only. Not yet tested.

gitgiggety:
Rebased on top of current develop and fixed comments. I will add unit tests over the weekend.

pkg/tag/update.go
gitgiggety: Added.

I did notice some test files have a "marker" for `// +build integration` while others don't, and these also seem to be in `package X_test`. Not sure if I should have added that as well to the test file (I never wrote any unit tests in Go so not sure what the "marker" + different package really means)

WithoutPants: `// +build integration` is used where we need to run against a database. Your tests should run without a database, so it should be fine without.

`_test` package is used to enforce internal API access restrictions, and isn't necessary.

WithoutPants:
Can you please merge in from the latest dev branch changes? I don't want to merge incorrectly.

gitgiggety:
Rebased, only had time for a quick test, but as conflict was only in `AddFilterDialog` it shouldn't have any real impact (and I obviously did test applying a filter sith and without children).

WithoutPants:
I have identified a significant query performance issue with this change, as well as a similar (presumably) existing performance issue with the hierarchical studios change.

Both problems are related to the image count functionality for the tags and studios query pages, although I imagine that this issue will be impacting other count functionality as well.

The problem resides in the `hierarchicalMultiCriterionHandlerBuilder` and `joinedHierarchicalMultiCriterionHandlerBuilder` structs. The issue is that it is performing the recursive `with` queries even if `depth` is set to 0. 

Having run an explain query plan on one of these queries, it looks like it is doing a full table scan of `tag`, `images_tags` and `images`. If there are a lot of images, then this causes a serious performance issue.

As a quick workaround test, I added the following to `imageTagsCriterionHandler`:
```
	if tags != nil && tags.Depth == 0 {
		h := joinedMultiCriterionHandlerBuilder{
			joinTable: imagesTagsTable,
			joinAs:    "tags_join",
			primaryFK: imageIDColumn,
			foreignFK: tagIDColumn,
			addJoinTable: func(f *filterBuilder) {
				qb.tagsRepository().join(f, "tags_join", "images.id")
			},
		}

		return h.handler(&models.MultiCriterionInput{
			Value:    tags.Value,
			Modifier: tags.Modifier,
		})
	}
```

Similarly, for `imageStudioCriterionHandler`:
```
	if studios != nil && studios.Depth == 0 {
		addJoinsFunc := func(f *filterBuilder) {
			f.addJoin(studioTable, "studio", "studio.id = images.studio_id")
		}
		h := qb.getMultiCriterionHandlerBuilder("studio", "", studioIDColumn, addJoinsFunc)

		return h.handler(&models.MultiCriterionInput{
			Value:    studios.Value,
			Modifier: studios.Modifier,
		})
	}
```

I ran a performance comparison on a contrived database with 4m images, 30k scenes, 500 studios and 1500 tags. 

Tags query time without fix:
40 tags/page - 7.12s
1000 tags/page - 2.7min

With fix:
40 tags/page - 6.8ms
1000 tags/page - 618ms

Studios query time without fix:
40 studios/page - 7.94s
500 studios/page - 1.6min

With fix:
40 studios/page - 104ms
500 studios/page - 467ms

I think `hierarchicalMultiCriterionHandlerBuilder` and `joinedHierarchicalMultiCriterionHandlerBuilder` both need to handle zero-depth queries with the same or better performance as the alternative implementations I described above.

kermieisinthehouse:
> I have identified a significant query performance issue with this change, as well as a similar (presumably) existing performance issue with the hierarchical studios change.
> 
> Both problems are related to the image count functionality for the tags and studios query pages, although I imagine that this issue will be impacting other count functionality as well.
> 
> The problem resides in the `hierarchicalMultiCriterionHandlerBuilder` and `joinedHierarchicalMultiCriterionHandlerBuilder` structs. The issue is that it is performing the recursive `with` queries even if `depth` is set to 0.
> 
> Having run an explain query plan on one of these queries, it looks like it is doing a full table scan of `tag`, `images_tags` and `images`. If there are a lot of images, then this causes a serious performance issue.
> 
> As a quick workaround test, I added the following to `imageTagsCriterionHandler`:
> 
> ```
> 	if tags != nil && tags.Depth == 0 {
> 		h := joinedMultiCriterionHandlerBuilder{
> 			joinTable: imagesTagsTable,
> 			joinAs:    "tags_join",
> 			primaryFK: imageIDColumn,
> 			foreignFK: tagIDColumn,
> 			addJoinTable: func(f *filterBuilder) {
> 				qb.tagsRepository().join(f, "tags_join", "images.id")
> 			},
> 		}
> 
> 		return h.handler(&models.MultiCriterionInput{
> 			Value:    tags.Value,
> 			Modifier: tags.Modifier,
> 		})
> 	}
> ```
> 
> Similarly, for `imageStudioCriterionHandler`:
> 
> ```
> 	if studios != nil && studios.Depth == 0 {
> 		addJoinsFunc := func(f *filterBuilder) {
> 			f.addJoin(studioTable, "studio", "studio.id = images.studio_id")
> 		}
> 		h := qb.getMultiCriterionHandlerBuilder("studio", "", studioIDColumn, addJoinsFunc)
> 
> 		return h.handler(&models.MultiCriterionInput{
> 			Value:    studios.Value,
> 			Modifier: studios.Modifier,
> 		})
> 	}
> ```
> 
> I ran a performance comparison on a contrived database with 4m images, 30k scenes, 500 studios and 1500 tags.
> 
> Tags query time without fix:
> 40 tags/page - 7.12s
> 1000 tags/page - 2.7min
> 
> With fix:
> 40 tags/page - 6.8ms
> 1000 tags/page - 618ms
> 
> Studios query time without fix:
> 40 studios/page - 7.94s
> 500 studios/page - 1.6min
> 
> With fix:
> 40 studios/page - 104ms
> 500 studios/page - 467ms
> 
> I think `hierarchicalMultiCriterionHandlerBuilder` and `joinedHierarchicalMultiCriterionHandlerBuilder` both need to handle zero-depth queries with the same or better performance as the alternative implementations I described above.

Goodness this sounds exactly like the performance issue I'm experiencing on my 4.2m image instance, which did pull in this branch. It sounds like mystery solved!

gitgiggety:
Can either, or both, of you post the generated query? If you could include the query plan as well that would be great as well.

@WithoutPants did you generate the database using the DB generator you created / merged a while back? As then I will have to check that out as well to get a big database and test on it for performance issues.

kermieisinthehouse:
Can you provide some instructions on how to do so? I've added some logging to `repository.go` but it spams dozens of queries for a single page load, none that I could find even mention tags.

gitgiggety:
I would have hoped this would be part of the default logging, but it seems like it isn't logged (and only some queries are logged). So I would have to check it myself, but it might take me some time to do so then.

gitgiggety:
Managed to find some time to generate a database (same as @WithoutPants described, 4M images, 30K scenes, 500 studios and 1500 tags). And well, I can confirm it's terribly slow. I extracted a single image count query and on my system that query takes about 4,5 seconds. (Although I'm uncertain whether this applies to all tags, it would mean 4,5 * 40 plus the additional costs for scenes etc).

I managed to quickly test the query with two fixes:
1. Merging both CTEs into a single one improves the performance to just shy off a second (~0,8 seconds), so about 5 to 6 times faster.
2. Removing the actual recursive part when not needed (i.e. dropping the `UNION SELECT ... WHERE depth < 0`) improves the performance to about ~0.06, 0.07 seconds. Without 1. being applied as well, so it might gain a bit of performance from that too).

So I'll try implement both in the upcoming days. Where 2. would obviously only optimize the `depth = 0` case and 1. work for all cases where an additional join table is involved. As I guess applying a tag filter on the image listing will be broken as well and there the depth option could have been set.

gitgiggety:
Created #1643 for the studios performance issue. Going back to the drawing board on this one. The current implemention obviously has a massive performance penalty which needs to be solved. But solving it isn't easy because it needs custom CTEs for every scenario (which means dropping `joinedHierarchicalMultiCriterionHandlerBuilder` as it's kinda useless / breaks the performance).
And it also happens to be that the additional CTE is needed because SQLite lacks support for "nested" joins. In for example MySQL and Postgress you can do the following, but SQLite throws a syntax error:
```sql
FROM foo
LEFT JOIN bar
  INNER JOIN baz ON bar.id = baz.bar_id
ON foo.id = bar.foo_id
```
Which would only join bar if baz also matches. Writing this using two decoupled `LEFT JOIN`s would yield different results as it could then join a bar even when there is no baz.
And in the case of this PR we obviously only want to match for example `images_tags` when the tag is actually allowed / found. That's why I went for the current solution using the CTE. Which apparently leads to this performance issue.

gitgiggety:
Rebased on develop, so the `depth = 0` fix is applied here as well. On the generated DB it still takes about a second to load `/tags` with the default of 40 items per page. But it's at least a lot better than the 50 seconds it took before this fix. With 1000 items per page it does take 20 seconds which is still (way) too long. (Didn't test without the fix, but that should have taken ages (25 * 50 seconds? would have been 20 minutes then)).

Also found another issue. It happens to be that during my daytime job I found out it's actually `WITH RECURSIVE` and not `RECURSIVE x (...)`, where `WITH RECURSIVE` is a hint that recursive CTEs must be enabled (meaning DB might switch implementation), and not a hint that the following CTE is recursive. So in short a syntax error is currently generated when for example applying the studio and tag filter. So `addWith` must have a flag to enable recursion which sets `RECURSION` just once, instead of prefixing the CTEs with `RECURSIVE`.

gitgiggety:
Addes some fixes. On the generated DB it seems to have improved performance on `/images` with a single tag filter (+ children) from 9 seconds to 1.5 second. (Without depth it's still a lot faster though). But there seems to be something strange going on, as the query copy pasted into the SQLite command line tool seems to take only half that time (in other words, only about 0.7, 0.8 seconds). So not sure why it does take so much longer in the end.

I also quickly tested using `INNER JOIN` instead of `LEFT JOIN` when applying "Includes (all)" but that didn't improve the query performance significantly. (Also removed the obvious `WHERE <joined>.<id> IS NOT NULL`)

@kermieisinthehouse if you were running this build before I would appreciate it if you could update to the latest build and compare them and report back any issues.

Still leaving this as WIP as for example "performer tags" filters are not optimized yet.

kermieisinthehouse:
Looking good so far. I can't make a tag filter that doesn't load instantaneously. Studios and tags pages are also instant, partially due to your other fix.

kermieisinthehouse:
I've found a performance regression. I have a saved filter that is: exclude about six tags + children, AND performer count is 0, AND is missing is studio.

In an earlier revision of this branch, it would load in a reasonable amount of time, now it is extremely slow. Disabling "include children" makes it load quickly again. I'd estimate there are about 5-6 child tags in the query.

Ps. Could we rename the UI to "Include sub-tags" or something

gitgiggety:
> I've found a performance regression. I have a saved filter that is: exclude about six tags + children, AND performer count is 0, AND is missing is studio.

I presume the query to generate the listing has regressed then. And these queries should be logged (with trace level) so you could just find it in Settings => Logging
Can you paste it here? And possibly include the query plan as well? You can get the query plan by replacing the args in the query (i.e.: replacing the `?` with the corresponding position of the listed args), opening the database using `sqlite3 /path/to/stash-go.sqlite` and executing `EXPLAIN QUERY PLAN <the query>;` (don't forget the `;` at the end to actually execute it).

If you could also include the old query (+ plan) that would be great.

Do you also now if it regressed with todays fix, or if it possibly already regressed on the fix for `depth = 0`? I presume it's with the changes of today, but can't hurt to ask.

EnameEtavir:
> Ps. Could we rename the UI to "Include sub-tags" or something
I would also vote for sub-tags instead of children.

gitgiggety:
Found some time to look into the "Exclude" issue, and I can confirm it the issue. Checking the query plan SQLite does a full table scan on `images` which obviously hurts the performance big time, while it doesn't do a full table scan when the filter is set to "Include (all)". Unsure how to fix it though, without introducing yet another performance issue.

About the placeholder. It obviously should be fixable, but would need a criterion type based mapping then. So I'll try to get to it, but my first priority is to get the performance sorted.

kermieisinthehouse:
I have bisected the changes and determined that the issue was introduced in the most recent commit, commit 2904bbfa68883a9b02e492cd3f5a1d56ebc403fd.

gitgiggety:
Interesting food for thought. Same query, but without `ORDER BY path ASC` just takes a tenth of a second. While with `ORDER BY` it takes 2 seconds. So I think the problem isn't really in the exclude filter. It's in the sheer number of results. I think it can mostly optimize the CTE and joining it in the main query (as it seems to create an index on `image_tag.image_id` automatically, which it can match against the primary key / index of `images.id` to throw away what isn't needed). But it then has to search every record (which in case of 4M images will most likely be 3M+ results) to get the `path` so it can order on that. I'm guessing that's why the full table scan is used. The CTE results in 3M+ rows, so searching in the id PK is kinda useless as it knows it will match 90%+ of all records, and it still has to read the full record to get the path for sorting.

This also explains why loading the initial `/images` page is so slow. That also triggers a read of every record for ordering.

EnameEtavir:
@gitgiggety maybe you are already using it - but just in case -  to inspect what sqlite is doing for a given statement you can use  `EXPLAIN QUERY PLAN [stmt]`

gitgiggety:
> I have bisected the changes and determined that the issue was introduced in the most recent commit, commit [2904bbf](https://github.com/stashapp/stash/commit/2904bbfa68883a9b02e492cd3f5a1d56ebc403fd).

I think I know where this is coming from. Because of joining in `images_tags` every iteration it doesn't just return a set of unique tags on the depth it is going through, but it also returns the images for those tags. So when you for example have 1 parent tag with 2 child tags, and every child tag has 2 children (so 4 grandchildren) and every tag having 5 unique images (so in total 7 tags, and 35 images). When you then search for the parent including children the old solution would have the recursive CTE first find 1 record, it would then start a recursive iteration to find the children, so search for 1 parent, and find 2 children, and then do another recursive pass to search for the 2 found children as parent, which results in the 4 grandchildren.
The new query however yields very different results. For the first pass it already returns 5 records, all for the parent tag, but it will include the 5 `images_tags` as well, so the second pass will search for the children, but it won't return 2 records, but many more, which I believe to be 50 (5 unique parent records matcht against 5 `images_tags` for the first child tag, resulting in 25 records, and the same for the second child tag). So the third pass, looking for the grandchildren will result in many many many more results (I think 625?). So the main query then has to go over 680, (625 + 50 + 5) non unique, records for joining. And afterwards group by, distinct, and order them. So that would explain the ramification effect of using depth + exclude. While when using an include (all) filter it is able to optimize the query much more.

So IMO the last commit should be reverted, and a different solution should be searched.



> @gitgiggety maybe you are already using it - but just in case - to inspect what sqlite is doing for a given statement you can use `EXPLAIN QUERY PLAN [stmt]`

I know :) And that's what I have been using all along now to come up with these thoughts / conclusions (but it's a shame it doesn't include estimated row counts like Postgres does, as that can lead to very different query plans, and for example doesn't show the current exponential groth of the recursive result)

kermieisinthehouse:
@gitgiggety 
CC @EnameEtavir @Phasetime 

Here's a patch to rename the UI labels to away from 'child/children' - I don't have access to your branch so here's a patchfile, usable with `git am child_ui_strings.patch`

```
From 4a4ae35f892e383c0dc08219b50edc8f58e400d2 Mon Sep 17 00:00:00 2001
From: Kermie <kermie@isinthe.house>
Date: Wed, 18 Aug 2021 20:28:09 -0700
Subject: [PATCH] Remove UI references to child studios and tags

---
 .../List/Filters/HierarchicalLabelValueFilter.tsx  | 14 ++++++++++++--
 .../components/Studios/StudioDetails/Studio.tsx    |  2 +-
 .../components/Tags/TagDetails/TagDetailsPanel.tsx |  2 +-
 .../components/Tags/TagDetails/TagEditPanel.tsx    |  4 ++--
 ui/v2.5/src/locales/de-DE.json                     |  2 +-
 ui/v2.5/src/locales/en-GB.json                     |  7 ++++---
 ui/v2.5/src/locales/pt-BR.json                     |  2 +-
 ui/v2.5/src/locales/zh-CN.json                     |  3 +--
 ui/v2.5/src/locales/zh-TW.json                     |  5 ++---
 9 files changed, 25 insertions(+), 16 deletions(-)

diff --git a/ui/v2.5/src/components/List/Filters/HierarchicalLabelValueFilter.tsx b/ui/v2.5/src/components/List/Filters/HierarchicalLabelValueFilter.tsx
index c157e970..9105beef 100644
--- a/ui/v2.5/src/components/List/Filters/HierarchicalLabelValueFilter.tsx
+++ b/ui/v2.5/src/components/List/Filters/HierarchicalLabelValueFilter.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { Form } from "react-bootstrap";
-import { defineMessages, useIntl } from "react-intl";
+import { defineMessages, MessageDescriptor, useIntl } from "react-intl";
 import { FilterSelect, ValidTypes } from "../../Shared";
 import { Criterion } from "../../../models/list-filter/criteria/criterion";
 import { IHierarchicalLabelValue } from "../../../models/list-filter/types";
@@ -49,6 +49,16 @@ export const HierarchicalLabelValueFilter: React.FC<IHierarchicalLabelValueFilte
     onValueChanged(value);
   }
 
+  function criterionOptionTypeToIncludeUIString(): MessageDescriptor {
+    const optionType =
+      criterion.criterionOption.type === "studios"
+        ? "include_sub_studios"
+        : "include_sub_tags";
+    return {
+      id: optionType,
+    };
+  }
+
   return (
     <>
       <Form.Group>
@@ -63,7 +73,7 @@ export const HierarchicalLabelValueFilter: React.FC<IHierarchicalLabelValueFilte
       <Form.Group>
         <Form.Check
           checked={criterion.value.depth !== 0}
-          label={intl.formatMessage({ id: "include_children" })}
+          label={intl.formatMessage(criterionOptionTypeToIncludeUIString())}
           onChange={() => onDepthChanged(criterion.value.depth !== 0 ? 0 : -1)}
         />
       </Form.Group>
diff --git a/ui/v2.5/src/components/Studios/StudioDetails/Studio.tsx b/ui/v2.5/src/components/Studios/StudioDetails/Studio.tsx
index aaeb8ff4..db05b24e 100644
--- a/ui/v2.5/src/components/Studios/StudioDetails/Studio.tsx
+++ b/ui/v2.5/src/components/Studios/StudioDetails/Studio.tsx
@@ -278,7 +278,7 @@ export const Studio: React.FC = () => {
             </Tab>
             <Tab
               eventKey="childstudios"
-              title={intl.formatMessage({ id: "child_studios" })}
+              title={intl.formatMessage({ id: "subsidiary_studios" })}
             >
               <StudioChildrenPanel studio={studio} />
             </Tab>
diff --git a/ui/v2.5/src/components/Tags/TagDetails/TagDetailsPanel.tsx b/ui/v2.5/src/components/Tags/TagDetails/TagDetailsPanel.tsx
index 2ff9c1ed..18994438 100644
--- a/ui/v2.5/src/components/Tags/TagDetails/TagDetailsPanel.tsx
+++ b/ui/v2.5/src/components/Tags/TagDetails/TagDetailsPanel.tsx
@@ -59,7 +59,7 @@ export const TagDetailsPanel: React.FC<ITagDetails> = ({ tag }) => {
     return (
       <dl className="row">
         <dt className="col-3 col-xl-2">
-          <FormattedMessage id="child_tags" />
+          <FormattedMessage id="sub_tags" />
         </dt>
         <dd className="col-9 col-xl-10">
           {tag.children.map((c) => (
diff --git a/ui/v2.5/src/components/Tags/TagDetails/TagEditPanel.tsx b/ui/v2.5/src/components/Tags/TagDetails/TagEditPanel.tsx
index 24948bfb..86b53558 100644
--- a/ui/v2.5/src/components/Tags/TagDetails/TagEditPanel.tsx
+++ b/ui/v2.5/src/components/Tags/TagDetails/TagEditPanel.tsx
@@ -185,9 +185,9 @@ export const TagEditPanel: React.FC<ITagEditPanel> = ({
           </Col>
         </Form.Group>
 
-        <Form.Group controlId="child_tags" as={Row}>
+        <Form.Group controlId="sub_tags" as={Row}>
           {FormUtils.renderLabel({
-            title: intl.formatMessage({ id: "child_tags" }),
+            title: intl.formatMessage({ id: "sub_tags" }),
             labelProps: {
               column: true,
               sm: 3,
diff --git a/ui/v2.5/src/locales/de-DE.json b/ui/v2.5/src/locales/de-DE.json
index cbb513d7..aadc0651 100644
--- a/ui/v2.5/src/locales/de-DE.json
+++ b/ui/v2.5/src/locales/de-DE.json
@@ -91,7 +91,7 @@
   "birthdate": "Geburtsdatum",
   "bitrate": "Bitrate",
   "career_length": "Länge der Karriere",
-  "child_studios": "Tochterstudios",
+  "subsidiary_studios": "Tochterstudios",
   "component_tagger": {
     "config": {
       "active_instance": "Aktive stash-box Instanz:",
diff --git a/ui/v2.5/src/locales/en-GB.json b/ui/v2.5/src/locales/en-GB.json
index 7a443aa2..8b097f53 100644
--- a/ui/v2.5/src/locales/en-GB.json
+++ b/ui/v2.5/src/locales/en-GB.json
@@ -91,8 +91,8 @@
   "birthdate": "Birthdate",
   "bitrate": "Bit Rate",
   "career_length": "Career Length",
-  "child_studios": "Child Studios",
-  "child_tags": "Child Tags",
+  "subsidiary_studios": "Subsidiary Studios",
+  "sub_tags": "Sub-Tags",
   "component_tagger": {
     "config": {
       "active_instance": "Active stash-box instance:",
@@ -517,7 +517,8 @@
   "image_count": "Image Count",
   "images": "Images",
   "images-size": "Images size",
-  "include_children": "Include children",
+  "include_sub_studios": "Include subsidiary studios",
+  "include_sub_tags": "Include sub-tags",
   "instagram": "Instagram",
   "interactive": "Interactive",
   "isMissing": "Is Missing",
diff --git a/ui/v2.5/src/locales/pt-BR.json b/ui/v2.5/src/locales/pt-BR.json
index dd8cac0a..91a76c23 100644
--- a/ui/v2.5/src/locales/pt-BR.json
+++ b/ui/v2.5/src/locales/pt-BR.json
@@ -91,7 +91,7 @@
   "birthdate": "Data de nascimento",
   "bitrate": "Taxa de bits",
   "career_length": "Duração da carreira",
-  "child_studios": "Estúdios filhos",
+  "subsidiary_studios": "Estúdios filhos",
   "component_tagger": {
     "config": {
       "active_instance": "Ativar stash-box:",
diff --git a/ui/v2.5/src/locales/zh-CN.json b/ui/v2.5/src/locales/zh-CN.json
index 5182f4b8..5e3bc11b 100644
--- a/ui/v2.5/src/locales/zh-CN.json
+++ b/ui/v2.5/src/locales/zh-CN.json
@@ -91,7 +91,7 @@
   "birthdate": "出生日期",
   "bitrate": "比特率",
   "career_length": "活跃年代",
-  "child_studios": "子工作室",
+  "subsidiary_studios": "子工作室",
   "component_tagger": {
     "config": {
       "active_instance": "目前使用的 Stash-box：",
@@ -499,7 +499,6 @@
   "image_count": "图片数量",
   "images": "图片",
   "images-size": "图片大小",
-  "include_child_studios": "包含子工作室",
   "instagram": "Instagram",
   "interactive": "互动",
   "isMissing": "缺失",
diff --git a/ui/v2.5/src/locales/zh-TW.json b/ui/v2.5/src/locales/zh-TW.json
index 935f1ba6..da74548e 100644
--- a/ui/v2.5/src/locales/zh-TW.json
+++ b/ui/v2.5/src/locales/zh-TW.json
@@ -91,7 +91,7 @@
   "birthdate": "出生日期",
   "bitrate": "位元率",
   "career_length": "活躍年代",
-  "child_studios": "子工作室",
+  "subsidiary_studios": "子工作室",
   "component_tagger": {
     "config": {
       "active_instance": "目前使用的 Stash-box：",
@@ -604,6 +604,5 @@
   "filters": "過濾條件",
   "filter": "過濾",
   "filter_name": "過濾條件名稱",
-  "detail": "詳情",
-  "include_child_studios": "包含子工作室"
+  "detail": "詳情"
 }
-- 
2.33.0


```

AnonNSFW:
@kermieisinthehouse If we're concerned about the possibility for misundersanding the meaning of "child" here, we should probably remove localisations and let native speakers ensure that the translations are free of any similar connotations wrt underage vs. subsidiary. For example, the various Chinese localizations are still referring to 子, and while google translate shows the full phrase as "sub studio", 子 is child in Japanese and seems to be (male) child in simplified chinese.

kermieisinthehouse:
My concern isn't necessarily overt misunderstanding, but just some recognition that this for porn and some tact is necessary. I checked the German and Brazilian translations with word reference and didn't see them being overly confusing, even if they share a root word - I am not familiar enough with east Asian languages to check them, but I'd support removing those too. The changes are simple enough I'll leave it to @gitgiggety instead of spamming more patchfile

gitgiggety:
On the performance:
I did a lot of testing/tried lots of different queries and I can't come up with a better solution. The most recent query obviously doesn't work for "Exclude" and has a massive performance impact and can't be used. But this new query is (a lot) faster when using "Include (all)".
So the only solution I can come up with is either reverting to the old query, which works fine when not including sub tags in the result, but is slower when including sub tags (while the new query thus is quicker). Or using two different queries: the new faster one for "Includes (all)" and the old one for "Excludes". Another alternative might be to run a separate query beforehand to resolve the tags to use and build an `IN` based on that result. But I didn't test if this works, let alone if it's as fast as the current query. And in the current code it would be hard to implement as well as filter builders don't have access to the database.
So @WithoutPants do you have any opinion on this? So either reverting to the slower query which is more or less acceptable in all cases (~2 seconds for "includes (all)" and ~5 seconds for "Excludes" when including sub tags). Or complicating the code by using two different queries which keeps the same performance for "Excludes" but speeds up the "Includes (all)". Personally I guess I would go for implementing both and accept the crappy code as a result.

Regarding the "children" discussion. I was questioning the naming as well at the time. But the "parent" "child" naming was already in use when I implemented the studio filtering for "parent / child", by the "Child studios" tab. And as reply to the translations: IMO it's up to the translator to come up with good / unambiguous naming in that case. We obviously don't know the correct wording for every language and can't validate if it would still be an "age related" ambiguous name.

gitgiggety:
Not sure why I didn't test this before. But it seems like the "old new" solution is actually (way) faster than the current code (of develop branch). @kermieisinthehouse can you confirm this? Comparing `develop` branch vs `cf651b35fe6940f6f9c8cb119d2a74f7e9a97bda` that the latter is faster
In a simple test of a single tag exclusion it comes down to 9 seconds for develop and about 3,5 seconds for said commit, on my generated database, and even when setting "include all children" it takes "only" 6 seconds so still 3 seconds faster than the old query.
Also note that not setting any filters already results in a slow query (almost 3 seconds on my database). As already said, this boils down to SQLite having to sort millions of items, and dropping the `ORDER BY` clause results in sub 100 millisecond queries (except when applying a depth, then it takes about 2,5 seconds, in other words: about the same as for "includes" with `ORDER BY` applied, but "includes" only results in a couple thousand matches which have to be ordered). But in this case the query will obviously get slower as you start navigating further (i.e.: applying a `LIMIT 40 OFFSET 3000000` and the query is slow again, as SQLite can't just return the first 40 items which match but has to test over 3000000 items if they do match, just like it has to test all 4 million before it can apply the `ORDER BY`). But IMO this does confirm that this is just too many data for SQLite to handle quickly. And as it looks like right now it does seem like switching from `WHERE NOT EXISTS(...)` to `LEFT JOIN ... WHERE ... IS NULL` already makes applying an "exclude" must faster (just not sub second :()

gitgiggety:
Just another interesting test:
```sql
WITH RECURSIVE tag AS (
    SELECT id as root_id, id as child_id, 0 as depth
    FROM tags
    WHERE id in (1)
    UNION
    SELECT p.root_id, c.child_id, depth + 1
    FROM tags_relations AS c
         INNER JOIN tag as p ON c.parent_id = p.child_id
)
   , image_tag AS (
    SELECT j.*, d.root_id, d.child_id
    FROM images_tags AS j
         INNER JOIN tag AS d ON j.tag_id = d.child_id
)
SELECT DISTINCT images.id
FROM images
     LEFT JOIN image_tag ON image_tag.image_id = images.id
WHERE (image_tag.root_id IS NULL)
GROUP BY images.id
ORDER BY images.path ASC
LIMIT 40 OFFSET 0;
```
This is the 6 second query because it has to exclude tag with id 1 and all it's children. (Note this tag has one child, but having no children doesn't make a difference and also takes 6 seconds to execute)

The first part, is pretty fast, as this only takes about 70milliseconds (so same `WITH RECURSIVE tag AS (...)` but just doing `SELECT * FROM tag`):
```sql
WITH RECURSIVE tag AS (
    SELECT id as root_id, id as child_id, 0 as depth
    FROM tags
    WHERE id in (1)
    UNION
    SELECT p.root_id, c.child_id, depth + 1
    FROM tags_relations AS c
         INNER JOIN tag as p ON c.parent_id = p.child_id
)
SELECT * FROM tag;
```

Converting this into a `VALUES()` thingy results in:
```sql
WITH tag(root_id, child_id) AS (
    VALUES (1, 1), (1, 2)
)
    , image_tag AS (
    SELECT j.*, d.root_id AS root_id, d.child_id AS child_id
    FROM images_tags AS j
             INNER JOIN tag AS d ON j.tag_id = d.child_id
)
SELECT DISTINCT images.id
FROM images
         LEFT JOIN image_tag ON image_tag.image_id = images.id
WHERE (image_tag.root_id IS NULL)
GROUP BY images.id
ORDER BY images.path ASC
LIMIT 40 OFFSET 0;
```
So with `VALUES (1, 1), (1, 2)` to "emulate" root tag id 1 with "items" 1 and 2 (so tag id 2 is a child of tag id 1). And this query (also) takes 3 seconds. So in an ideal world we split this in two queries. One to get the children / items, and then the create a `VALUES()` string out of it to add to the actual query. But this obviously means more "complex" code as the criterion builder would need access to the database which isn't currently the case.

Just as a sidenote. The last query can also be generated as:
```sql
SELECT DISTINCT images.id
FROM images
LEFT JOIN (
        SELECT j.image_id, d.column1 AS root_id
        FROM images_tags AS j
        INNER JOIN (VALUES (1, 1), (1, 2)) AS d ON j.tag_id = d.column2
    ) image_tag ON image_tag.image_id = images.id
WHERE (image_tag.root_id IS NULL)
GROUP BY images.id
ORDER BY images.path ASC
LIMIT 40 OFFSET 0
```
This as `VALUES()` can be used everywhere where `(SELECT ...)` can be used. But I don't think `addJoin` supports this, so keeping a CTE for it might be better. (But on the other hand the current code with `CTE`s doesn't work when you apply the same filter multiple times (using `and` / `or` / `not`) as it obviously will add multiple CTEs using the same name)

gitgiggety:
Don't ask me how, but I actually managed to implement this fix already. So performance should now be much better, both with include and exclude. But this includes a fairly massive rewrite so requires careful retesting.

Beware this also has a known limitation, and that is that the query might get long, possibly too long. So if there are lots of children, and possibly also multiple tags set in the filter, a very long `VALUES(...), (...), (...), (...)` list might be generated. Furthermore it uses the `GROUP_CONCAT` query, and I don't know if SQLite has any limitations on that (like MySQL has: default settings cut of a `GROUP_CONCAT` after I think 1024 chars unless you change some variarble/setting).

kermieisinthehouse:
These last commits are working great! I've tested a few edgecases and couldn't find any performance issues even on my ridiculous database. Please consider merging my UI patch above in some form, but either way, this pull is

Tested-by: kermie@isinthe.house

gitgiggety:
> These last commits are working great! I've tested a few edgecases and couldn't find any performance issues even on my ridiculous database.

That's great to hear 👍

> Please consider merging my UI patch above in some form

Sorry about that. Was a bit to euphoric when I got the fix incorporated late last night. Shouldn't have marked the PR as finished before also updating the wording of "children"

gitgiggety:
@kermieisinthehouse just added your patch.

Didn't really check it. Personally I'm not that fond of just changing translation keys/intention for all translations. This as for example "subsidiary studio" could have a different translation as "child studio". So these changes in intention are only reflected in the English text. But I'll leave it up to @WithoutPants to either leave the translations as is or clear / remove them so the translators have to provide them again.

So I guess that from my point of view this PR is ready now and can be tested / merged.

WithoutPants:
Performance is looking pretty good as far as I can tell. 

There appears to be a bug with the tag hierarchy validation code. I created a hierarchy `A <- B`, `A <- C`, `B <- C`. If I then modify `C` but make no changes and submit, I get the error message: `Parent tag 'A' is already applied`. The error message to the UI probably needs to be improved to indicate exactly how the validation failed, but I can't actually work out why it's failing here.

gitgiggety:
The error is correct? As you have ` A <- C` and `A <- B <- C` so `A` is twice a parent, once directly and once via `B`. This as you more or less requested yourself [here](https://github.com/stashapp/stash/pull/1519#issuecomment-867251064) :) So in that sense it's odd you were able to create this hierarchy in the first place as it shouldn't have been possible.

With regards to improving the error message, that would be really hard. This as currently the list of parents is queried from the database without loading the whole tree. The query just returns that `A` is a parent of `B`, but doesn't include the full path so if you have `A <- X <- Y <- B` the query just returns `A`, `X`, `Y` without the needed info for a detailed error message and it could only state that `A` is already applied via `B` but not if it's a direct or indirect parent of `B`.

So please let me know if you want the validation for duplicate parents or children to be gone altogether, in other words only validating if a parent is a child as well. Or if you want to keep this validation, how you were able to create this hierarchy in the first place :)

WithoutPants:
> Does this system allow a tag to have an ascestor tag as its direct parent? ie `A <- B <- C` and `A <- C`, and if so why?

This was just a question based on my understanding of the change. At the time, I thought that this hierarchy was possible. I was trying to wrap my head around it conceptually. The rest of that comment relates to the merging mechanics, but makes specific reference to loops in the hierarchy (ie `C <- A <- B <- C`). 

You followed up with:

> About the `A <- B <- C` and `A <- C` question. This currently is possible. It isn't a feature per se, but does work just fine and shouldn't lead to any issues. But obviously can be resolved when already iterating over all ancestors anyway.

So, you changed this so that this type of hierarchy is no longer allowed?

I think some clear documentation of what is allowed and disallowed is going to be necessary. I will see if I can reproduce the steps to get into this scenario again.

gitgiggety:
IIRC I indeed did change it when reimplementing the validation. I think an argument can be had for both. I don't see any real use case for allowing it / when it should be needed. But at the same time it shouldn't lead to any issues if it's allowed.

WithoutPants:
Restricting it so that a parent can't appear twice in a hierarchy is fine as far as I can determine, since I still don't see how such a hierarchy would work conceptually.

To reproduce: create `A`, create `C` with parent set to `A`. Create `B`, setting parent to `A` and children to `C`.

WithoutPants:
Import/export code needs to be updated to include tag hierarchy relationships.

gitgiggety:
> To reproduce: create A, create C with parent set to A. Create B, setting parent to A and children to C.

Okay, will test and look into it later.

> Import/export code needs to be updated to include tag hierarchy relationships.

Good catch. Didn't really know about import / export until I "copy pasted" tag aliases to studio aliases.
How would you like this to be stored? Personally I just add a `parents` (or `childs`) field to the tags JSON. So not storing relations separately nor storing both. So either `{"id": 1, "name": "A", "children": [2]}` + `{"id": 2, "name": "B", "children": []}` or the other way around `{"id": 1, "name": "A", "parents": []}` + `{"id": 2, "name": "B", "children": [1]}`. There is IMO no need to store both parents and children as they should be the same. And storing a separate tag relations structure might be cumbersome as well. And as AFAIK you can only export and import everything there can't be any missing tags. Although having the foreign key(s) does mean that all relationships need to be collected and saved after all tags are imported. Otherwise you might run into FK errors. (In case of storing `children` when the sub tag hasn't been imported yet and in case of storing `parents` when the parent hasn't been imported yet).

WithoutPants:
Use the studio hierarchy as the basis. Storing parent tags should be sufficient, and you'll need to ensure that sub-tags are imported after parent tags - which is what the studio code does IIRC.

gitgiggety:
> To reproduce: create A, create C with parent set to A. Create B, setting parent to A and children to C.

Confirmed. I can also keep saving B without any issue. IIRC I'm only validating set parents and all its parents, and set children and all its children. While it should also check the parents of the children (or children of the parents).

Also found another issue. Was testing this in the same tab so everything in the same app instance with one cache. After saving B I used `Sub tags: C` to open C, but this doesn't (re)load the object, as it was already loaded. So it didn't show B as parent and saving thus also removed it as parent. So when a tag is saved the cache of the set parents & sub tags must also be invalidated.
(Luckily there are only two things hard when programming, naming things, cache invalidation, and off-by-one errors).

gitgiggety:
@WithoutPants can you provide any insights in the caching issue? This as I'm new to GraphQL and the Apollo client.

The "first issue" is that currently Apollo already logs a warning when changing the parents / children. This due to the server returning a different array then the client contains, and Apollo not knowing how it should handle the array (should it replace the existing value with the new one? Should it merge the arrays into one?). For which the warning also refers to the documentation: https://www.apollographql.com/docs/react/caching/cache-field-behavior/#the-merge-function . But as I'm new to this and no merge policies being applied in the codebase I'm uncertain whether this is the way to go.

The second issue is that I can't get it to reload the already loaded tag. I've added `GQL.FindTagDocument` to the `tagMutationImpactedQueries` in `StashService`, but this does seems to do nothing. I can see calls being made to reload for example the `AllTagsForFilter`, but it doesn't make a `FindTag` call and when navigating to the other tag it still shows the stale data.
A different approach I tried, which might be preferable (?) is to change the loaded data after mutation to `parents { id children { id } }` and `children { id parents { id } }`. This way Apollo knows the children of the parents, and parents of the children, changed and it does seem to work. But obviously is also impacted by Apollo not really knowing how to handle the new array, going back to the first issue.

gitgiggety:
Hmm, obvious problem in this. When creating `A <- B` it works when fetching `parents { children { id } }` + `children { parents { id } }`. But it doesn't work when removing the relationship. This because when editing `A` and removing `B` as sub tag `B` won't be part of the returned object and thus won't update the cache of `B` either.

So the only solutions would just clearing the cache all together (`refetchQueries` or `update` with the existing `deleteCache` which for some reason doesn't seem to pick up on `FindTagDocument`). Or a manual client side implementation, which I tried using `update` but this suffers the same issue as the old values aren't available (can't be fetched from cache either, as the cache is already updated when `update` is invoked), or a `merge` on the `typePolicies` but this has some issues as well (not having access to the `Tag` object which is updated (i.e.: the object containing the field which merge is invoked for). When deleting tags `merge` obviously won't be applied, etc).

gitgiggety:
Just pushed fixed unique hierarchy detection. Lets hope this one does work.

So currently the only remaining issue is the GraphQL caching. For which I don't think that any `refetchQuery` (or `deleteCache` variant on `update`) will work. This because `refetchQuery` runs the last query of that type once more, but the latest `FindTag` obviously wasn't for the tags which must be updated. And `update: deleteCache` removes the cached values for some queries, but it doesn't actually remove the cached objects. So while at that point there is no cache anymore for `FindTag` or `FindTags` the tags are still loaded into the Apollo cache. So it doesn't "magically" refetch those tags. So the only solution I see is to change `TagEditPanel` to keep track of the old parent & sub tags, compare them to the saved parent & sub tags, and update those which are either added or removed (either by refetching them using `use` or `query` I guess, or manually updating hem in the cache using `cache.modify()`), and the same obviously should be done when tags are deleted (update all tags which the deleted tag had as parent or sub tag).

gitgiggety:
Updated to fix GraphQL issue as well. Relations are now immediately updated locally after a successful create/update/delete.

So this should be ready to go again, unless new issues are found obviously.

WithoutPants:
I think I must have messed up my earlier follow up performance testing, because tag-image query performance is still really slow.

Database was generated using the script in this branch: https://github.com/WithoutPants/stash/tree/generate-markers-tags

As per this reply (https://github.com/stashapp/stash/pull/1519#issuecomment-899267787) getting ~4-5s for tags page with the current code, and under a second with my fix.

gitgiggety:
@WithoutPants are you sure you messed up the previous testing and not the latest one :)

Both on my "old" DB as a new one generated using the script on the branch you mentioned (which did require this branch as well, which was unexpected) it takes under 1 second to load `/tags`, both with 40 items per page as 1000 items per page (40 items per page obviously being a lot quicker as well). But this is by checking the logging of the app/webserver. It takes some more time before the results are actually shown in the browser. But not sure why that is, possibly due to slow rendering of the React components or having to download for example the 1000 images.

WithoutPants (approved):
It does indeed look like a problem at my end.

I think this is ready for wider testing.

gitgiggety:
That sounds strange. Do you mean hardware and/or runtime related, or possibly the database being way off (because it's randomly generated, it might be tilted in having many items under a single tag or whatever)?

WithoutPants:
> That sounds strange. Do you mean hardware and/or runtime related, or possibly the database being way off (because it's randomly generated, it might be tilted in having many items under a single tag or whatever)?

It might be that I didn't have the correct compiled code or something. I basically retried and wasn't able to reproduce.

gitgiggety:
Ah, right, then I misread/misunderstood :)

I thought you meant you were still having issues (but others confirmed it did work fine).

PR #1522, opened 2021-06-22
Add stash-box batch scene tagging

InfiniteTF:
Also:
* Adds field selection for scenes
* Adds phash distance matching, currently controlled by the server.
* Handles performer merging so old merged ids are purged from performers, and the new one added. Refreshing a merged performer is not yet possible since the server doesn't check for redirects yet.

i18n of the new strings is TODO, so is fingerprint submission for batch tagging.


bnkai:
- There seems to be an issue with the `Excluded fields` functionality, nothing seems to be excluded
- Batch Update doesnt seem to populate the URL fields
- Not sure if its expected ... After the batch update is run and finished there is no change in the UI (matched/tagged scenes are not updated). The only way to notice what happened is the log

PR #1526, opened 2021-06-23
Clear scene tagger results on new search

### LINKED ISSUE #1494: [Feature] Hide "No Matches Detected" display on tagger when new search executed
When in tagger view, hitting search with no results shows "No Results Found" in red below.  If you then go in and edit the search and hit search again, the "no results found" stays and may be confusing also due to it not necessarily being visually clear when a search is happenign.

1) clear the "no results found" after hitting the search button a second time
2) Perhaps more largely a popup "Searching..." dialog while finding results for a tagger search, similar to the performer search?

WithoutPants:
Clears tagger results for a scene when beginning a new search.
Fixes #1494 

Also fixes a minor oshash message ID problem.

### DIFF
--- a/ui/v2.5/src/components/Tagger/Tagger.tsx
+++ b/ui/v2.5/src/components/Tagger/Tagger.tsx
@@ -195,7 +195,17 @@ const TaggerList: React.FC<ITaggerListProps> = ({
     inputForm?.current?.reset();
   }, [config.mode, config.blacklist]);
 
+  function clearSceneSearchResult(sceneID: string) {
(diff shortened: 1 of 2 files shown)


JanJastrow:
Hi, I just noticed a minor bug with this, where the tagger result did not get cleared:

I was in tagger view, clicked "match fingerprints",
a scene was matched, but since the match was not correct, I wanted to manually match the scene.

The result was never cleared when I clicked (manual) "search", the match stayed.
I was only able to manually search by reloading the page.

PR #1529, opened 2021-06-23
Simplify tagger results

WithoutPants:
Combines the fingerprint search results with the manual search results, so that we can clear results as needed when handling button clicks. Also gets rid of some repeated code.

Also contains a fix to show the correct message for show/hide unmatched button.


WithoutPants:
Closing this in favour of a larger refactor for 0.9.

Issue #985, opened 2020-12-05
[Feature] Tag Blacklisting


SpedNSFW:
Similar train of thought to #121 but instead of mapping the tag to another name, have a blacklist for tags. Some studios have stupid tags for their scenes like "HD", "Straight", "Pornstar", etc that you don't want to see.

My request is to add an interface where you can enter any tag you wish to be blacklisted and when you scrape a new scene that features those tags, they don't get added (or prompted to be added).

gitgiggety:
@WithoutPants I wanna have a stab at this. But do you have any suggestions for it? Storing the blacklist in a new database table? Of storing it in the config file? Or a separate (txt) file containing a single term per line?
And suggestions for the UI part are welcome as well.

Especially as I do believe such a blacklist might get longish. I.e. might go to 50 or more terms I guess. So for example storing in the config file and adding to the Configuration tab of the settings (below the Scraping header) might not be a good idea.

hyde231:
well, what is a central config file good for, if you don't consequently store the user config in there? 
just my 2 cents

gitgiggety:
@hyde231 It's hard to disagree with you on the fact that user configuration should be stored as such. But for one the user configuration is only configured using the interface, so does it really matter where and how it is stored? And secondly a list of blacklisted tags, which I guess might contain tens if not over 100 items can also be described as a "database" of blacklisted tags, and not necessarily as configuration. So storing it in the database (or flat file as being a flat file database) might be a better solution.

Which obviously is also why I'm asking the question, before developing it one way while it's preferred to implement differently.

WithoutPants:
I don't think it belongs in the database. I think adding it to the configuration file is fine for now. We can look at migrating it out later if we need to.

For UI presentation, I think a collapsible section for it would suit. I like the presentation of the blacklist in the tagger. 

We have a `Scrapers` tab already. Perhaps we should consider renaming it to `Scraping` and move the scraping related configuration to that page. The existing scrapers list can be put into a collapsible section.

gitgiggety:
Thanks for the input. I'll probably split it then. First do the UI refactoring and afterwards adding the tags blacklisting. As both changes can for example be reviewed independent of each other.

gitgiggety:
[...]

Issue #1446, opened 2021-05-27
[Feature] Allow for in-GUI renaming of scrapers, or visible descriptions in lists


GsMumbo:
**Is your feature request related to a problem? Please describe.**
As you start adding more and more scrapers, it becomes very difficult remembering which scrapers focus on what. For example this screenshot is from the Create Performer page:

![B6747C52-101D-4513-B39D-EE71AFF4D68E](https://user-images.githubusercontent.com/77939716/119761069-240bb780-be79-11eb-87ea-810f7f591133.jpeg)

I know some of these focus more on amateurs (OnlyFans for example), while some focus on more commercial stars (Brazzers for example), while some even include celebrities (Freeones and Babepedia for example). I think one focuses on performers found in galleries such as Met-Art but I don’t remember which. A lot of them I’m still unsure of what their focus is.

**Describe the solution you'd like**
One option is to allow for the user to rename scrapers in the GUI. So if I for example start scraping and find that Brazzers has a big focus on commercial stars, I can rename it to `Brazzers (Commercial)` or if I don’t care about the name at all, I can rename it to `Commercial Stars`. Having it in the GUI means you can do it on the fly, even on mobile. If it’s saved in the database then it can even be persisted should the scraper update.

The other option is to allow for a description to be added in the GUI, and for that description to be visible in the scraper selection table. This allows for the name to remain as-is, while still providing the clarity needed to quickly pick the right scraper.

**Additional context**
I know there’s a list of scrapers available in Settings that you can look through to find out what URLs are covered, but it’s not easily referenced in the moment. And in the case of Performers (which is my main focus here), that list doesn’t provide any additional context as it only links to the site itself. You’d have to go to each site, find out what they specialize in, then find a place to note it down so you remember for later. I’ve done that before, and you always end up forgetting if you don’t regularly use a scraper enough.

MyDirtyAccount:
Thinking out loud here - another approach would be to categorize the scrapers. A couple obvious categories might be Stash Box, Commercial site, index (Babepedia), metadata (performer image, stash sqlite). My example category names are poor, but I hope they illustrate the broad categories approach.

I often am looking for just a specific kind of scraper, hence the suggestion - e.g., everything **except** commercial sites.

PR #1548, opened 2021-07-02
Restructure scraping settings

gitgiggety:
As discussed in #985 and [commented](https://github.com/stashapp/stash/issues/985#issuecomment-871083394) by @WithoutPants I moved the scraping configuration from the (general) Configuration tab to the Scrapers tab and renamed it to Scraping. The scrapers are now wrapped in collapsible so hidden by default.

The end result looks like this:
image


jeremymeyers:
This looks great though personally i would move the "Save" above the scrapers section since there is no interaction to save in the scrapers section.

jeremymeyers:
Also might be an opportunity to implement #1446  (if people think its a good idea)

gitgiggety:
During implementation I did think for a second whether the Save button should be in between or at the bottom. But for one all (settings) pages have it at the bottom, so it would be quite strange to put it somewhere else on this tab. Secondly the button isn't really "hidden" IMO as the page obviously is a lot shorter now with the scraper lists being collapsed by default. So on any (desktop) screen the button is in view when you open this tab. And on mobile you would be scrolling anyway, and I guess the save button would be quite quickly into view then too, as the collapsed scraper lists would easily fit onto the screen.

And lastly, the feature request you're linking to would warrant the need for saving the scrappers, so at that point the Save button would still have to be at the bottom anyway.

graphql/documents/data/config.graphql
```diff
@@ -32,9 +32,6 @@ fragment ConfigGeneralData on ConfigGeneralResult {
   galleryExtensions
   excludes
   imageExcludes
-  scraperUserAgent
-  scraperCertCheck
-  scraperCDPPath
```
WithoutPants, line 37: For compatibility purposes, we want to avoid removing these right now. They should be marked as deprecated, but should still function correctly.

graphql/schema/types/config.graphql
```diff
@@ -89,12 +89,6 @@ input ConfigGeneralInput {
   excludes: [String!]
   """Array of file regexp to exclude from Image Scans"""
   imageExcludes: [String!]
-  """Scraper user agent string"""
-  scraperUserAgent: String
-  """Scraper CDP path. Path to chrome executable or remote address"""
-  scraperCDPPath: String
-  """Whether the scraper should check for invalid certificates"""
-  scraperCertCheck: Boolean!
```
WithoutPants, line 97: As above, these should be left in. `scraperCertCheck` should be made optional. For these three fields, the server should only change the configuration value if the fields are present. Mark as deprecated as needed.

graphql/schema/types/config.graphql
```diff
@@ -162,12 +156,6 @@ type ConfigGeneralResult {
   excludes: [String!]!
   """Array of file regexp to exclude from Image Scans"""
   imageExcludes: [String!]!
-  """Scraper user agent string"""
-  scraperUserAgent: String
-  """Scraper CDP path. Path to chrome executable or remote address"""
-  scraperCDPPath: String
-  """Whether the scraper should check for invalid certificates"""
-  scraperCertCheck: Boolean!
```
WithoutPants, line 170: As above - leave in and deprecate.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -240,9 +240,10 @@
       "hooks": "Hooks",
       "triggers_on": "Triggers on"
     },
-    "scrapers": {
+    "scraping": {
```
WithoutPants, line 243: This change should be added to the other locale files as well.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -151,7 +151,7 @@
       "interface": "Interface",
       "logs": "Logs",
       "plugins": "Plugins",
-      "scrapers": "Scrapers",
+      "scraping": "Scraping",
```
WithoutPants, line 154: Please remove `scrapers` from the other locale files.

WithoutPants (changes requested):

graphql/documents/data/config.graphql
gitgiggety: I'm not too familiar with GraphQL and the usage in Stash. But aren't these fragments only used by the client side code? (in other words: used by the code generator for React). So removing these would then be fine I guess as long as the web interface doesn't use them? (which obviously isn't the case anymore)

ui/v2.5/src/locales/en-GB.json
gitgiggety: As internationalization is new I'm wondering how you would want to handle this. Do you want me to remove these lines from the other locale files, or change them into `"scraping": ""`? So updating the key, and making the value an empty string.
Adding (or updating) translations using an empty string IMO has the advantage that translators can find missing translations more easily and all files could be kept identical in format (so every file would have `"scraping"` on line 154 for example) which I think would make it easier to "compare" the files.

gitgiggety:
Rebased on current develop and updated excepted for the two things I commented on / asked a question about.

ui/v2.5/src/locales/en-GB.json
WithoutPants: This way would need some modification to our message loading code. Currently, we merge the chosen locale messages with the default message locale (`en-GB`), using lodash's `merge` function. This combines the chosen and default maps, using the default value if the chosen value resolves to `undefined`. You will probably need to use `mergeWith` supplying a customiser function to return the default value if the chosen value is empty.

I'm also not completely sold on this approach as it makes the `en-US` locale unnecessarily verbose (most of the messages should use the `en-GB` value). I'd prefer not to have to maintain identical locale file structures as well.

However, when adding new messages to add a new placeholder message to the other locales, I think it's not a bad approach to add an empty string placeholder value - provided the above code change is made.

gitgiggety: So if I do understand you correctly you _do_ want me to set the value / translation to `""` and update the translation loading code to ignore these values?

WithoutPants: Sorry, was offering an opinion rather than a directive. :)

If you're happy to do the changes to the message loading code, then go ahead and do it. Otherwise, just remove `scrapers` from the other json locale files.

gitgiggety: That clears it up, I will make the changes then.

I was a bit unsure because the first part of the comment seemed to imply you were totally against using `""` and at the end you seemed to find it a good solution (if the fix to merging is added as well, obviously).

graphql/documents/data/config.graphql
gitgiggety: @WithoutPants any comment on this?

I did accidentally revert it as well, but as stated in aboves comment I'm not entirely sure if it was needed. This because I believe it's only used by the client side code (generator). So if that's the case different clients of the API aren't using this so removing it would be fine as the Stash web UI doesn't need it anymore.

WithoutPants: Oh, yes, you are correct. If it's client side it can be removed assuming it's unused.

ui/v2.5/src/locales/en-GB.json
WithoutPants: Yeah I can understand why it was confusing. What I wasn't keen on was trying to maintain identical file structures for all of the locale files.

gitgiggety:
Rebased on top of current develop branch to include pt-BR as well.

Also added support for language lines to be empty and fall back to the `en-GB` line. But in the end it turned out that all languages already had a translation for `Scraping` which I just reused / copied over.

graphql/documents/data/config.graphql
gitgiggety: > If it's client side

Well, I'm not sure (either) if it is client side only :) But for now I left it in. But feel free to remove it if you think it is unnecessary.

WithoutPants (approved):

PR #1568, opened 2021-07-09
[Feature] Better resolution search

### LINKED ISSUE #1005: [Feature] Resolution Greater Than/Less Than Filters
I'd love to see some additions to the resolution filter for scenes. Currently you can only filter for a set resolution. What I'd like to see is also being able to filter for greater than, less than, greater than or equal to, less than or equal to. The behaviour of these should be pretty straight forward.

1080p with a less than filter will select everything below 1080p; so 240p, 480p, and 720p.  
720p with a greater than or equal to filter will select 720p and everything above; so 720p, 1080p and 4k.

Jekora:
Resolves #1005 

Added the ability to filter resolution by:
- Equals
- Not Equals
- Greater Than
- Less Than

This is my first contribution, and first time writing code in go. I've did my best to follow the guidelines. 
I was looking for this feature, so I tried to implement it myself.
I did struggle with writing the tests, but I think they should be fine. Let me know if there's anything you want me to change.


WithoutPants (approved):
Thanks for the contribution!

Issue #1376, opened 2021-05-10
Funscript implementation for stash compatablity with interactive toys (handy, fleshlight laucnch, keon, etc)


inventiv3919:
**Is your feature request related to a problem? Please describe.**
Stash offers an unparalleled interface for browsing 2d porn collections. [Interactive toys](https://www.thehandy.com) offer an [NSFW unparalleled viewing experience NSFW](https://thumbs2.redgifs.com/BeneficialCalculatingCoral-mobile.mp4) for enjoying 2d porn collections. Stash is currently incompatible with funscripts (small files used to sync interactive toys to videos).

**Describe the solution you'd like**
For complete funscript integration to be added to stash with support for both organisation and playback.
**Describe alternatives you've considered**
Only modifying stash enough for existing funscript playback software to hijack viewing sessions and obtain information about which video file is being played. I can't code so don't know which would be more work but instinct tells me itd be much easier to make stash compatible with one piece of software that can serve scripts to a variety of toys rather than trying to impliment direct support for different toys from different manufaturers.

**Additional context**
[API to manipulate theHandy](https://app.swaggerhub.com/apis/alexandera/Handy/1.0.0) 
[theHandy official subreddit](https://www.reddit.com/r/theHandy/)
[Eroscripts](https://discuss.eroscripts.com) is where most discusion of funscripts takes place, there is a section dedicated specifically to software develpment and releases

rezreal:
Would you consider to support the buttplug.io hardware library?

programotter:
There's also https://xtoys.app/, which has funscript video support and many toys.

vib3coda:
just made a community plugin for autoblow API, if you like to check it out:
https://github.com/vib3coda/CommunityScripts/tree/autoblow_sync/plugins/autoblow_sync

I will open a pull request soon, after running a few more tests, but it's already working great.

DogmaDragon:
The Handy devices now are supported. 

Splitting other device support into individual issues:
- https://github.com/stashapp/stash/issues/3625
- https://github.com/stashapp/stash/issues/6579

PR #1573, opened 2021-07-12
Add script offset / delay to Handy support.

UnluckyChemical765:
Further work on  #1376.

Offsets are added to the current video position, so a positive value leads to earlier motion.  (The most common setting.)

This is needed because most script times have a consistent delay when compared to the video. (Delay from the API calls to the server should be handled by the server offset calculation.)


WithoutPants (changes requested):
`scriptOffset` should be made more specific.

PR #1603, opened 2021-07-31
Actually implement TagFilter.marker_count

gitgiggety:
The marker_count filter/criterion as defined in TagFilterType isn't
actually implemented. This adds an implementation for it.

Do note this implementation _might_ have performance issues because of
using OR (in the join). Another implentation would be to remove both
joins and use:
```SQL
COUNT(
    SELECT id FROM scene_markers WHERE primary_tag_id = tags.id
  UNION
    SELECT scene_marker_id FROM scene_markers_tags WHERE tag_id = tags.id
)
```
Note this doesn't require a DISTINCT as UNION already removes any
duplicate records.


gitgiggety:
Found this issue while working on my [Kodi addon](https://github.com/gitgiggety/plugin.video.stash). So not sure why this wasn't implented but added to the API.

WithoutPants:
It is implemented but is commented out on line 274. It's not documented there unfortunately (but is elsewhere), that it is commented out due to significant performance issues. It should have been commented out in the API as well. I think it should remain disabled until the performance issues are resolved. From memory, I think the issue is because of the `primary_tag_id` column in `scene_markers`. It needs to either be moved as a flag in `scene_markers_tags`, or `scene_markers_tags` should also include the primary tag in it.

gitgiggety:
Hmm, seems like I'm getting blind then, missed the commented out code for this.

Regarding the performance issue: it might be what I described so changing the query to that might solve the issue? Furthermore the performance issue obviously should only occur when the filter is applied, so I don't think it would be that big of an issue? And looking back on the old(er) code it seems like previously all tables where always joined, so that would explain a massive performance penalty as well. But with the conditional joins (only joining when filter is set) that shouldn't apply anymore either.

WithoutPants (approved):
Tested on the problematic databases, and it indeed appears to be performing a lot better. Added in the filter and sorting on the UI as well.

Issue #588, opened 2020-06-01
[Feature] Apply video/image preview changes to marker previews


WithoutPants:
**Is your feature request related to a problem? Please describe.**
#510 changed the preview generation to generate video previews by default, with an extra checkbox for image previews. There is no such option for markers, which are generated in a similar way, and apply to the wall view in the same way.

**Describe the solution you'd like**
Apply the preview changes to markers: change the single checkbox to two: video preview and image preview. 
Have the encoding preset apply to previews and markers, or add a dropdown for marker preset. Ensure it is intuitive for the user.

PR #1604, opened 2021-08-01
Generate screenshot images for markers

gitgiggety:
In some scenarios it might not be possible to use the preview video or
image of markers, i.e. when only static images are supported like in
Kodi. So generate a static screenshot as well.


WithoutPants:
I don't really like the idea of generating artifacts for use exclusively outside of stash. This should probably be implemented as a separate plugin or script.

gitgiggety:
Hmm, unsure if / how this could be implemented as a plugin. Especially since it would need an API change (additional `screenshot` field) and the routing endpoint to actually download the image.

What I didn't investigate further, but might be worthwhile, is that the wall has the (static) image option which at the moment shows the video but doesn't auto play. Obviously this could use the static image as well. But unsure whether downloading a full size image would be faster / better than downloading a part of the preview video and showing the first frame as static image.
So maybe using the image generated by this branch as the static image option of the Wall view would convince you that there is a use case in Stash itself :)

gitgiggety:
@WithoutPants any update on this? Is there a possibility for you to reconsider this? :) As written before, it's impossible to do this in a plugin. And I'm open to making this a separate task so users can choose to generate these images instead of always generating them as part of the video preview generator task.

WithoutPants:
We do indeed have a static image option.

There is a pending issue #588 related to adding the webp generation for markers. If you can add that option and a static image option, then we can have the marker wall view show the static image if it is selected in the interface options (I don't think it's a particularly useful option, but it is there).

gitgiggety:
So splitting the existing task in two (video at top level, webp as sub-option, just like scenes has), and adding the static image task as sub option as well.
And using the image as generated using this PR for the "static image" UI option as well.

👍

gitgiggety:
Updated. Generating markers now has two subitems for animated images and static images. When wall is set to static image the static image will be used.

I made the decision to _not_ generate the webp previews by default because they're also not generated for scene previews by default. But this obviously "breaks" backwards compatibility as previously they were always generated.

WithoutPants (approved):

PR #1609, opened 2021-08-03
Remove stripes and add background colour to default performer images

WithoutPants:
As mentioned in #1396, the stripes in the default performer images are a possible migraine/epilepsy trigger. The images also had no background colour, resulting in a transparent image. Removed the stripes and added a background colour based on the existing stash scheme.

An example of a new image in the performer lightbox:
image

Compared with the old - you can see the transparency issue here:
image

I personally prefer the stripes, and I imagine others might want to use the existing images. A zip containing the old images is attached. This can be used with #1489 once it is merged.

[performer.zip](https://github.com/stashapp/stash/files/6922019/performer.zip)


bnkai (approved):

bnkai:
I also prefer the old images, this looks ok though till 1489 is merged.
Theme devs should probably add a performer pack with appropriate backgrounds after 1489 is merged

PR #1615, opened 2021-08-05
Add target="_blank"  to links.

UncleRoger33:
"Path", "Stream" and "Downloaded From" links in File Info screen will now open in new tab/window. 

pull_request-LINKS

Links in the current version opens in the same tab/window which Stash runs.

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneDetails/SceneFileInfoPanel.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneDetails/SceneFileInfoPanel.tsx
@@ -46,7 +46,7 @@ export const SceneFileInfoPanel: React.FC<ISceneFileInfoPanelProps> = (
         <span className="col-4">
           <FormattedMessage id="path" />
         </span>
-        <a href={`file://${path}`} className="col-8">
+        <a href={`file://${path}`} className="col-8" target="_blank">
           <TruncatedText text={`file://${props.scene.path}`} />
         </a>{" "}
       </div>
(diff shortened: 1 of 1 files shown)


kermieisinthehouse:
Can we get a `rel="noreferrer"` on the outbound link too, to not reveal the existence of stash installs?

UncleRoger33:
> 
> 
> Can we get a `rel="noreferrer"` on the outbound link too, to not reveal the existence of stash installs?

Added rel="noreferrer" to the first commit, but looks like i may have broke something else (_i'm not good at using github_) 😱

gitgiggety:
Would be nice if you can add `rel="noreferrer"` to other outgoing links as well. Like the URL, Twitter, Instagram fields on the performer pages, I believe studio also has a link to the website, galleries too, and images might have as well. Could you check / update those too?

With regard to the GitHub error / failed checks. Before committing you should run `make fmt-ui` (presuming you're running the code locally and not editing the files in GitHubs web editor). The failures are related to formatting issues, which will automatically be corrected when you run `make fmt-ui`.

gitgiggety:
Looked at this a bit better, and seems that for example the performers page uses a custom `URLField` component, which does set `_blank`, `noreferrer` _and_ `noopener`. So `noopener` should be applied here as well (as otherwise the target site might still have some way to communicate with the Stash page).

Not sure though why the `URLField` component isn't applied to for example the scenes page as well. Although the HTML differs "a lot" (scenes page using `<div/>` and `<small/>` while `URLField` uses `<dl/>`, `<dt/>` and `<dd/>`). So I guess it can't be replaced easily to use `URLField` for this File Info tab as well.

gitgiggety:
Based on my previous comment I did start some work to make use of `TextField` and `URLField` but I had to make some more changes. So IMO it's better to use that instead (at the moment I finished `SceneFileInfoPanel`, but need to update the File Info panels for image and gallery as well).

gitgiggety:
I just created #1622 which fixes this as well, and includes some other fixes too.

So sorry for "overruling" your pull request.

WithoutPants:
Closed in favour of #1622.

UncleRoger33:
> 
> 
> I just created #1622 which fixes this as well, and includes some other fixes too.
> 
> So sorry for "overruling" your pull request.

No problem, this is the greatness of open source 👍

PR #1617, opened 2021-08-05
Scrape tag exclusions

### LINKED ISSUE #985: [Feature] Tag Blacklisting

gitgiggety:
Now that the scraping settings have been moved lets continue with the actual feature, tag blacklisting :)

The current implementation has one quirk, which is that the regex is converted to lowercase meaning it doesn't match uppercase tagnames unless explicitly making them case insensitive. In other words: tag `4K` and adding an exclusion with value `4K` won't actually exclude the tag because the regexp is converted to `4k` and tries to match that. This can be "worked around" by setting the exclusion as `(?i)4k` to make the regexp case insensitive. This behavior has been duplicated from the file exclusion feature so can be easily changed when needed.

Fixes #985


pkg/scraper/scrapers.go
```diff
@@ -509,3 +507,36 @@ func (c Cache) ScrapeMovieURL(url string) (*models.ScrapedMovie, error) {
...
+			logger.Errorf("Invalid tag exclusion pattern :%v", err)
+		} else {
+			excludeRegexps = append(excludeRegexps, reg)
+		}
+	}
+
+ScrapeTag:
+	for _, t := range scrapedTags {
+		for _, reg := range excludeRegexps {
+			if reg.MatchString(t.Name) {
```
bnkai, line 529: It should be
```golang
                       if reg.MatchString(strings.ToLower(t.Name)) {
```
Since you convert the regexps to lower the tag names should also be converted to lower (the same is done in the file exclusion code). This would make the matches case insensitive (as they should be imo) without having the user add the `(?i)` in the regex. Only making the regexp lower case doesnt make sense.

I would also add a `logger.Debugf` message with the tag that was skipped due to being blacklisted before the `continue`

gitgiggety: Not sure how I missed the `ToLower` in file exclussions. Anyway, it's added now.

I did add logging as well, but instead of logging every tag I collect them and add a single log entry containing all ignored tags. This message is also logged as info, instead of debug. Because file exclusion also logs as info I saw.

bnkai: Yeah everything seems ok now.

bnkai (approved):

Issue #1485, opened 2021-06-06
[Feature] Locate file in explorer


deliciousnights:
Often when finding a scene in Stash I want to locate it in the File explorer, to either move it or rename it, or even just open it with my player of choice. It would be very convenient to have this option directly in the scene list and in the scene details.

**Describe the solution you'd like**
A button or a link in scene list/details to open the containing directory (and select the corresponding file). Something like:
![image](https://user-images.githubusercontent.com/71196494/120931505-2defaf00-c6f2-11eb-97db-26b92416047d.png)

**Describe alternatives you've considered**
Atm I have to check the file path and then navigate to it manually in the file manager. Which takes much longer than the proposed solution. And this works only from scene details, not from the scene list.

...

One step beyond would be to have the file rename option in Stash itself. But just opening the manager would be already a great help.

gitgiggety:
Webbrowsers don't allow opening local directories, in the file explorer of the OS.

Furthermore Stash might not run on the same computer as that the user is using (I myself run stash on my homeserver) so the paths would also be incorrect.

So even if a workaround is possible (i.e.: letting the Stash backend start the file explorer, which already would be hard because of different OSes, possibly running in Docker, ...) it won't really work in any scenario where Stash is running on a different system.

deliciousnights:
Could it be a plugin? If the plugins are allowed to run code on the server, it could open the file manager, if certain conditions are met.

alexhk:
There already is a link to the file in the file info view. If you use Firefox and install this [addon](https://addons.mozilla.org/en-US/firefox/addon/local-filesystem-links/?utm_content=addons-manager-reviews-link&utm_medium=firefox-browser&utm_source=firefox-browser) and you run Stash on your workstation you should be able to open the file via click.

![2021-06-09_11h28_45](https://user-images.githubusercontent.com/1495412/121335138-9f11ab00-c91a-11eb-89a3-b81c6670c608.jpg)

deliciousnights:
Thanks for the tip, the extension works well! Would be even better if the context menu feature was actually implemented (to use it in the scene list with the tagger view - there the full path is visible), but it's already a great improvement.

For me this solves the issue, albeit by installing extra software. Feel free to close it if you feel Stash doesn't need to add anything here.

jennygreene88:
I've rigged up a little solution to play the files in a native local player even when Stash is running on a different machine. I think it's the simplest working solution possible given the limitations. I:

[...]

PR #1622, opened 2021-08-06
Fix links and styling of File Info tabs

gitgiggety:
This PR kinda supersedes #1615. It rebuilds the File Info tabs / panels so they make use of the `TextField ` and `URLField` components. The later component does automatically set `target="_blank" rel="noopener noreferrer"` on the links so fixes exposing the stash installation to the outside world.

Furthermore the "details" lists like shown on the File Info tabs, but also on the Movie and Performer page now make use of CSS grids. This means that IMO items are better aligned when some longer labels are shown (for example the Measurements label did overlap the value on my phone).

Another small included fix is that I changed the breaking logic for URLs on the file info tabs so that the break can be inserted anywhere. This as previously the break was only applied at spaces, while for example the path and URL to the source can be (too) long and most likely won't contain any spaces. So now the path / URL will fill up the available space and then be eneded with the ellipses.


gitgiggety:
When desired I can update the studio page as well. This currently does stand out as (read-only) view does show the form but with disabled input fields. And besides the design being off, it doesn't even allow clicking the URL to the websiyof the studio.

cseverino:
It looks like this fix also fixed an issue I as having with locating files in the file explorer! This is great!

See this issue:  [https://github.com/stashapp/stash/issues/1485#](https://github.com/stashapp/stash/issues/1485#)

PR #1626, opened 2021-08-09
[WIP, HELP!] Add duration statistics to stats page and scene pagination

FleetingOrchard:
So, first of all, it's a WIP because the code added to `scene.go` slows down the initial load of each page of the scene list _quite_ significantly on queries resulting in lots of results. I gave adjusting `query.executeFind()` to allow it to optionally pull a sum of a given column a red-hot go, but ultimately didn't get anywhere; my golang-fu, in this regard, is shamefully weak, and I invite someone else to pick up the torch and achieve what I could not. I probably won't take another swing at getting that into proper shape myself for a solid, like, week or so, so if you can do this trivally, go for your life, kid. Anyway.

Adds a scene duration to the stats page (and shifts the stats around a bit so that everything still fits nicely), and adds the same to the pagination footer. Here's some screenshots:

image
image


WithoutPants:
I can see some benefit of adding total duration to the stats page, but not so much the query results. Adding yet another return value for `Query` is less than desirable. If I were to implement this, I'd probably refactor the `Query` method to accept some further options which would include whether or not to include the duration, and I'd change the return value to be a struct and error, rather than an object list, count and error. In any case, it would be a non-trivial change. On the query performance side, you really need to include a `sum(duration)` (as required) to the ids query. IMO it's not worth the effort involved just to return the total duration of the queried scenes - the benefit of which appears marginal at best. The query code needs to be refactored first.

FleetingOrchard:
Yeah, that's absolutely what I aimed for at first, I just couldn't get it there when I was putting the changeset together. It's mostly me just not groking Go quite yet.

The benefit of duration in query results, I've found, is mostly about just being aware of the volume of media one has. But it's also useful when using queries as queues/pseudo-playlists, which I see is getting some attention of the hotkey variety in another PR.

FleetingOrchard:
I've rolled this back to _just_ the total scene duration on the stats page, given that's just a much, much more simple change. I'll probably come back to per-query duration at some point in a few months or whatever.

kermieisinthehouse:
Tested-by: kermie@isinthe.house

kermieisinthehouse:
365-day years but 30-day months can make some weird stuff go on, but I think being a little off for 1% of a year isn't so bad, and the alternatives aren't great.

Screenshot_20210823-113759_1

Issue #1625, opened 2021-08-08
[Feature] Studio aliases


echo6ix:
**Is your feature request related to a problem? Please describe.**
Some studios have alternate names or spellings, and some studios have changed their name over the years. Since each studio entry can can only have one name, this limits their auto-tagging scope. The end-user either has to make a duplicate studio entry with the alternate/defunct name, or tag files of said studio manually.

**Describe the solution you'd like**
Create the capability for the end-user to add studio alias names that will be part of the auto-tagging process for studios.

An example use-case is the website studio PornWorld formerly called DDF Networks, and often denoted in filenames as DDF.

- Main studio: PornWorld
  - Alias: ddf
  - Alias: ddfnetwork

gitgiggety:
As an FYI: I started working on this. It's mostly finished but still need to add it to importing and exporting and fixing all unit tests.

PR #1629, opened 2021-08-10
Rebuild Studio page by splitting view and edit

gitgiggety:
As the title says :) Refactor the `Studio` component into a `StudioDetailsPanel` and a `StudioEditComponent` component just like all other of these pages.

After this refactor links are used for both the URL field as the parent studio which where previously not available as the read-only form field was shown. Furthermore it allows for expanding the features like implementing #1625.

### DIFF
--- a/ui/v2.5/src/components/Studios/StudioDetails/StudioDetailsPanel.tsx
+++ b/ui/v2.5/src/components/Studios/StudioDetails/StudioDetailsPanel.tsx
@@ -1,0 +1,57 @@
+import React from "react";
+import { useIntl } from "react-intl";
+import * as GQL from "src/core/generated-graphql";
+import { TextUtils } from "src/utils";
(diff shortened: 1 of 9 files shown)


ui/v2.5/src/components/Studios/StudioDetails/StudioDetailsPanel.tsx
```diff
@@ -45,9 +45,11 @@ export const StudioDetailsPanel: React.FC<IStudioDetailsPanel> = ({
         <TextField id="details" value={studio.details} />
 
         <URLField
-          id="studio"
+          id="parent_studios"
           value={studio.parent_studio?.name}
           url={`/studios/${studio.parent_studio?.id}`}
+          trusted
```
gitgiggety, line 51: I believe these could/should be applied in more places then. Movies for one also uses `URLField` for studio and thus also opens a new tab, and applies `noopener noreferrer`

PR #1635, opened 2021-08-12
Add scene queue keyboard shortcuts

### LINKED ISSUE #1423: [Feature] Keyboard shortcut for next scene in scene view
While I'm in a scene's detail view, (with the video playing on the right side and details on the left) editing a scene I would like a shortcut for the next/previous scene. It would make going through scenes so much more convenient for tagging/marker purposes. It's a bit of a pain needing to back out then click in to the next scene.  I'm sure there's other use cases for such a feature as well. Cheers!

WithoutPants:
Resolves #1423 

Adds shortcut to queue tab (`q`), play next scene in queue (`p n`), previous scene (`p p`), and play random scene in queue (`p r`).

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v090.md
+++ b/ui/v2.5/src/components/Changelog/versions/v090.md
@@ -7,6 +7,7 @@
 * Added not equals/greater than/less than modifiers for resolution criteria. ([#1568](https://github.com/stashapp/stash/pull/1568))
 
 ### 🎨 Improvements
+* Added keyboard shortcuts for scene queue navigation. ([#1635](https://github.com/stashapp/stash/pull/1635))
 * Improve Studio UI. ([#1629](https://github.com/stashapp/stash/pull/1629))
 * Improve link styling and ensure links open in a new tab. ([#1622](https://github.com/stashapp/stash/pull/1622))
 * Added zh-CN language option. ([#1620](https://github.com/stashapp/stash/pull/1620))
--- a/ui/v2.5/src/components/Scenes/SceneDetails/Scene.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneDetails/Scene.tsx
@@ -291,7 +291,9 @@ export const Scene: React.FC = () => {
       const { query } = sceneQueue;
       const pages = Math.ceil(queueTotal / query.itemsPerPage);
       const page = Math.floor(Math.random() * pages) + 1;
-      const index = Math.floor(Math.random() * query.itemsPerPage);
+      const index = Math.floor(
+        Math.random() * Math.min(query.itemsPerPage, queueTotal)
+      );
       const filterCopy = sceneQueue.query.clone();
       filterCopy.currentPage = page;
       const queryResults = await queryFindScenes(filterCopy);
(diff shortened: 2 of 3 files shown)


kermieisinthehouse:
These shortcuts seem a little difficult. What about WASD/HJKL/Arrow keys?

WithoutPants:
Arrow keys are used by the video player. `a` is used to show the scene details tab. `d d` is used to delete a scene if edit tab is shown. `s s` is used to save the scene. `k` is used to show the markers tab. `p r` is used already to navigate to a random item from the query page, so it makes sense to maintain the same shortcut for that. `n` and `p` would be ideal since VLC uses the same shortcuts, but `n` collides with the new marker shortcut if the markers page is shown, and `p` would mean that the `p r` shortcut would not work. 

Of the remaining options, `w` doesn't strike me as intuitive, nor does `h`, `j` or `l`. Other options I'd consider would be `,` and `.` (`<`, `>` for previous and next). The `p` prefix makes sense to me because the play random shortcut uses it.

gitgiggety:
Tested in Firefox and Chrome and works as described. But I have to agree that it isn't very intuitive, although "Play Next", "Play Previous" and "Play Random" are quite easy to remember for `p n`, `p p` and `p r`.

Just to add, MPV uses `<` and `>` by default (so shift is required), but obviously that doesn't give a real option for random. Other options might be `9` `0` for `(` `)` without shift, or `[` `]` / `{` `}`

But I have to agree that VLCs `n` and `p` would be ideal, but if it requires "remapping" others it would obviously be a backwards compatibility breakage which should be carefully thought through. (My personal guess without any knowledge/info: I would think that `n` for new marker isn't used that much, and `n` for next in playlist might be way more popular. So maybe remapping it (and `p r` for random on the overview) _might_ be an option as it's easier to use / remember for a lot of users while possibly only a small group have to "re-calibrate" that "new marker" has a new shortcut)

kermieisinthehouse:
I agree, breaking backward compatibility for old shortcuts to make a more consistent mapping now would be best. 

Since left / right is already mapped to jwplayer, perhaps up and down could be used to change scenes? Betterfap (now defunct) had an excellent keyboard mapping which even included gamepad support. WASD / arrow keys / d-pad were used for next/previous scene and "liking" or "disliking" for their recommender. Perhaps the existing "tab" navigation could go to the numpad, e.g. 1-6 for Scene, Queue, Markers, Edit tabs, etc.

gitgiggety:
Up and down are used by jwplayer as well, to modify the volume. (On the subject of conflicting shortcuts, `f` obviously is fullscreen in jwplayer, but also opens the File Info tab).

Using numpad for the tabs would be odd I guess. This as the same tab might exists for multiple types but be in a different position (i.e. file info and edit are the last tabs, but scenes has way more tabs than image or gallery). So you then would have to remember the tab position per type or count the number of tabs. Prefixing tab shortcuts with `t` might also work, but my guess would be that that would mean a lot of users have to relearn the shortcuts. So if (lots of) shortcuts are changed then an RFC needs to be made to properly design all of them. (Other options would be combinations like `m n` for new marker, `m <X>` to view marker X in the current scene etc. More like prefixing the shortcut with the type you're performing the operation on).

bnkai (approved):

bnkai:
Imho the keyboard shortcuts selected are fine considering the existing ones. If needed i think it is better to create a new RFC issue to discuss the remapping of the shortcuts as a whole (as mentioned above).

WithoutPants:
Good discussion points, but yeah might be better to discuss in a separate RFC issue. I'll merge this so that at least the functionality is in there.

PR #1636, opened 2021-08-12
Custom page size and saved zoom level

### LINKED ISSUE #1532: Grid slider for scenes and images always returns to default position.
Grid slider for scenes and images always returns to default.

When using the slider it would be good for it to stick to what you set it to as it always returns to this position

Screen Shot 2021-06-26 at 3 31 08 PM

### LINKED ISSUE #1556: [Feature] Custom Value for Items to Show
When viewing on various different devices and different orientations I like to change the quantity of scenes on the page to different amounts, but find the current range of values to small and limiting (20,40,60,120,250,500,1000). 
If a custom value could be entered from 1+ items in intervals of 1, i could therefore save with a filter for different devices and achieve the desired look for each device.
Primarily wanted because I can't stand having some lines with different quantities of content than others when I can help it.

Also, the slider that changes the zoom level of scenes could really do with being saved along with the filters, as this also really affects positioning of items and would truly enhance the saved filters giving a more complete preset. Or maybe just the option before saving to include the current zoom level, I definitely can see that help with customising presets to achieve different looks. 

I guess could apply globally to for all filters not just scenes.

WithoutPants:
Resolves #1556 
Resolves #1532 

Adds a `Custom...` item to the page size drop down. Clicking this shows a popover that allows entering a custom page size. The selected page size is shown in the page size drop down menu until another page size is selected.

image
image

Added zoom index to the query parameters as `z`. The zoom level can now be saved in saved filters and default filter.

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v090.md
+++ b/ui/v2.5/src/components/Changelog/versions/v090.md
@@ -1,4 +1,6 @@
 ### ✨ New Features
+* Allow saving query page zoom level in saved and default filters. ([#1636](https://github.com/stashapp/stash/pull/1636))
+* Support custom page sizes in the query page size dropdown. ([#1636](https://github.com/stashapp/stash/pull/1636))
 * Added between/not between modifiers for number criteria. ([#1559](https://github.com/stashapp/stash/pull/1559))
 * Support excluding tag patterns when scraping. ([#1617](https://github.com/stashapp/stash/pull/1617))
 * Support setting a custom directory for default performer images. ([#1489](https://github.com/stashapp/stash/pull/1489))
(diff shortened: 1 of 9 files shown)


gitgiggety:
In Firefox Desktop (both 90.02 and 91.0) I'm unable to set the custom page size. When selecting the option "something" appears but immediately disappears again.

Strange thing is, on Firefox Mobile (on Android) it does work fine. Although on mobile (i.e., not just Firefox, but also when emulating mobile in Chrome desktop) there is a small styling issue which means the button with the checkmark is shown below the input field instead of next to the input field, and the "rectangle" is filled with the gray background:
image

bnkai:
Can confirm the above issue with Firefox (85) and Chrome (91) in linux.
Opera (78) seems to work ok.

WithoutPants:
On Firefox it seemed to be triggering the `rootClose` as part of the dropdown click event. Added a timeout to display the overlay which appears to have fixed it. Changed the overlay to use `InputGroup` which should resolve the wrapping issue on mobile views.

bnkai (approved):

bnkai:
Seems to work ok for me in linux using Firefox(85)/Chrome(91)/Opera(78)
Works also from my phone (chrome android)

Issue #1220, opened 2021-03-21
[Feature] Dupe Checker improvements


scruffynerf:
Per discord discussion with @InfiniteTF putting suggestions for improvements to dupechecker here

Some way to sort the list of duplicates... right now, it's by hash?

Ability to click on Phash in fileinfo, and see a 'similar to this video' list with just matching items (maybe sorted by distance?)  [display would be same as main Dupechecker, but with a single phash input, rather than running thru all phashs]

Ability to merge infos, often I find the smaller file has the right data, and I have to copy to the 'good one' (not really limited to phash dupes)

scruffynerf:
While having the filename showing for all of the dupes in a group is handy, showing the entire path would be handier.  We do that with tagging, so use the same on dupechecker?

PR #1639, opened 2021-08-14
Order sceneDuplicateChecker results by size

kermieisinthehouse:
Right now, the scene duplicate checker results are ordered randomly / by insert order. As the duplicate checker is primarily a space saver tool, this increases its usefulness by ordering results by potential size savings.

Before, small files up front:
Screenshot_20210815-225825_1_1

After, large files up front:
Screenshot_20210815-224758_1

### DIFF
--- a/pkg/sqlite/scene.go
+++ b/pkg/sqlite/scene.go
@@ -68,13 +68,15 @@ SELECT GROUP_CONCAT(id) as ids
 FROM scenes
 WHERE phash IS NOT NULL
 GROUP BY phash
-HAVING COUNT(*) > 1;
+HAVING COUNT(phash) > 1
+ORDER BY SUM(size) DESC;
 `
 
 var findAllPhashesQuery = `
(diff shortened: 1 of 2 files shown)


WithoutPants (approved):

LukeSal88:
For me, this works with search accuracy set to "Exact." Any other setting seems to revert to the old ordering.

kermieisinthehouse:
I can't reproduce this, but they are in fact supposed to return slightly different results due to how phash similarity is calculated. For exact matches, it shows groups with the largest combined size of all the matched scenes, but for inexact matches, it shows groups ordered by their largest scene. It should be very similar in practice. 

More info:
In exact matches, the grouping is done is sqlite, but searching within a window is done application side, so it would need go changes to alter order, instead of just sql.
Stash-box uses a fancy pgsql extension to do difference clustering database-side, but that's not available to us here.

PR #1650, opened 2021-08-18
Fix scene bitrate unit

### LINKED ISSUE #1649: [Bug Report] Bitrate expressed in "Frames per second"
**Describe the bug**
Bitrate is expressed in frames per second instead of something bitty

**To Reproduce**
1. Go to a scene
2. Click on 'File Info'
3. See error

**Expected behavior**
Bitrate expressed in Mbit/second or something sane.

**Stash Version: (from Settings -> About)**
v0.8.0-35-g7b3b2ae9

peolic:
Resolves #1649


WithoutPants:
This should use an internationalisation string instead.

PR #1655, opened 2021-08-21
Improve image scanning performance and thumbnail generation

InfiniteTF:
Speeds up image scanning by making thumbnail generation optional, and removing the full image decode when populating image metadata.

Additionally it adds optional thumb generation with [https://github.com/libvips/libvips](vips), when it's available on the system, or through ffmpeg if not. vips should be available to install on most platforms (libvips-tools on ubuntu), and uses native libraries for all decoding/encoding. 

For my test gallery, timings are:
- Current implementation: 30s to scan gallery
- Current implementation with metadata population fixed: 16s to scan gallery
- ffmpeg: 10s to scan and generate thumbs
- vips: 6s to scan and generate thumbs.
- No thumb generation: ~200ms to scan.

If no thumb is generated they will be generated on demand. For my test library it takes about 2s to load all thumbs completely, vs 500ms when they're pre-generated. In practice it's not super noticable. I've also enabled caching for thumbs, so it needs to happen less frequently.

Things I have not implemented yet are:
* A generate task for thumbs, if the user changes their mind.
* An option to write thumbs to cache when encoding on the fly. This might be useful for users with lower powered machines who'd prefer to only take the hit once.


kermieisinthehouse:
I can't tell from the code change, are on-the-fly thumbnails saved to /generated or just cached in the browser?

InfiniteTF:
> I can't tell from the code change, are on-the-fly thumbnails saved to /generated or just cached in the browser?

Just browser cached atm. Like I mentioned in my previous comment, this should probably be an option.

kermieisinthehouse:
I don't think it should be a user-facing option. The only behavior should be that if a missing thumb is requested, it is generated and saved. I can't think of a usecase where you'd want to generate on the fly on every page view, aside from extremely storage constrained environments - which is not your typical media hoarding stash user.

WithoutPants:
I think a generate task is necessary for this change. It would be very difficult to generate them after scanning without losing metadata otherwise.

gitgiggety:
Just as an FYI: I've just did some comparison between this branch and develop. Reason is because I'm writing a plugin which hooks into creation / scanning and scanning images is terribly slow. (Note the plugin doesn't actually get triggered on `Image.Create.Post`, so there aren't any slowdowns because of it).

Setup:
Scenes: 460 and a bit
Images: 4700 and a bit
Galleries: 32

Between runs I have deleted the images from stash, but unchecked to remove generated files. So a scan operation should only insert the image in the database and associate it with an existing gallery. Thumbnail generation is unchecked on this branch, and shouldn't really happen either on develop because the thumbnail should already exist. Scenes and galleries already exist in the database so should do nothing.

On this branch it takes roughly 8 minutes to scan / import those 4700 images. On develop it takes roughly 40 minutes. So that's an 80% speed increase.

But checking the code changes it does make sense that it's so much quicker. As if I'm reading it correctly then the old code was decoding the full image where the color of every pixel could be fetched etc, just to get the dimensions of the image? While the new code only reads the image "config" which contains the dimensions as well. So it has to read much less data.

For clarification: in my development environment the files are read over an sshfs mount. So if the files were stored locally it might have been quicker as well. And having to read the full file vs just a small portion might explain the big difference too, but decoding the full image might obviously take some time as well, and not just loading the file over the network.

kermieisinthehouse:
You should also add `vips-tools` package to the last section of docker/build/x86_64/Dockerfile

kermieisinthehouse:
Here's a patchset that persists thumbnails to disk just-in-time.

@WithoutPants I've double checked, and re-running a scan with "generate thumbnails" enabled will generate thumbnails for all images, even if they already exist. I don't think we need a dedicated generate task.
```
From 7c9a5f71f59ebddb2c5718d31ac893b0ddb3fcd1 Mon Sep 17 00:00:00 2001
From: Kermie <kermie@isinthe.house>
Date: Wed, 22 Sep 2021 16:15:30 -0700
Subject: [PATCH] Save on-the-fly thumbnails, update dockerfile

---
 docker/build/x86_64/Dockerfile | 2 +-
 pkg/api/routes_image.go        | 8 +++++++-
 2 files changed, 8 insertions(+), 2 deletions(-)

diff --git a/docker/build/x86_64/Dockerfile b/docker/build/x86_64/Dockerfile
index 0452c2b2..2ffcf705 100644
--- a/docker/build/x86_64/Dockerfile
+++ b/docker/build/x86_64/Dockerfile
@@ -36,7 +36,7 @@ RUN make build
 
 # Final Runnable Image
 FROM alpine:latest
-RUN apk add --no-cache ca-certificates
+RUN apk add --no-cache ca-certificates vips-tools
 COPY --from=backend /stash/stash /ffmpeg/ffmpeg /ffmpeg/ffprobe /usr/bin/
 ENV STASH_CONFIG_FILE=/root/.stash/config.yml
 EXPOSE 9999
diff --git a/pkg/api/routes_image.go b/pkg/api/routes_image.go
index f5db78e2..bdb64261 100644
--- a/pkg/api/routes_image.go
+++ b/pkg/api/routes_image.go
@@ -38,7 +38,7 @@ func (rs imageRoutes) Thumbnail(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Add("Cache-Control", "max-age=604800000")
 
-	// if the thumbnail doesn't exist, encode on the fly
+	// if the thumbnail doesn't exist, encode on the fly and save it
 	exists, _ := utils.FileExists(filepath)
 	if exists {
 		http.ServeFile(w, r, filepath)
@@ -49,6 +49,12 @@ func (rs imageRoutes) Thumbnail(w http.ResponseWriter, r *http.Request) {
 			logger.Errorf("error generating thumbnail for image: %s", err.Error())
 			return
 		}
+
+		err = utils.WriteFile(filepath, data)
+		if err != nil {
+			logger.Errorf("error writing thumbnail for image %s: %s", img.Path, err)
+		}
+
 		w.Write(data)
 	}
 }
-- 
2.33.0

```

kermieisinthehouse (approved):

WithoutPants:
I have found one regression in behaviour. webp files don't have thumbnails generated correctly. I haven't tested with vips installed, so I'm not sure if it works with that. 

To be clear, webp's didn't generate thumbnails in the previous version either, _however_ if the thumbnail wasn't generated/found, then the old behaviour was to show the original image instead. On this version, the thumbnail generation fails with an error message (which may get noisy) and a broken image is shown. If I navigate to the image itself then the image is displayed.

kermieisinthehouse:
I can't reproduce this using vips, so I suspect it has webp support. Either way, I think the fix is putting `rs.Image(w, r)` between lines 49 and 50 in routes_image.go, so that it sends the image if thumbnail generation fails

kermieisinthehouse (approved):

PR #1660, opened 2021-08-22
Studio aliases

### LINKED ISSUE #1625: [Feature] Studio aliases

gitgiggety:
Add support for studio aliases. Updates the GrapQL API, auto tagging, filename parsing, importing/exporting, matching during scraping, etc.

Fixes #1625


WithoutPants:
The auto tagger doesn't match on studio aliases when running from `Tasks -> Autotag`. I managed to fix it with this addition in `autotag/studio.go`:
```
	for _, c := range candidates {
		matches := false
		if nameMatchesPath(c.Name.String, path) {
			matches = true
		}

		if !matches {
			aliases, err := reader.GetAliases(c.ID)
			if err != nil {
				return nil, err
			}
			for _, alias := range aliases {
				if nameMatchesPath(alias, path) {
					matches = true
					break
				}
			}
		}

		if matches {
			ret = append(ret, c)
		}
	}
```

It requires an update to the unit test as well, which is why I haven't just done it myself.

gitgiggety:
Hmm, good find. I probably didn't notice because I have the files structured in folders by studio, so to reproduce I had to remove the (full) studio name from path.
_But_ such code isn't in the tag code either (which I copy pasted the code from). So does that work properly or does that have the same bug?

I fixed the unit tests such that they pass. It might be desirable to copy/paste the test method and modify it so the name doesn't match and the alias would match?

Also rebased on top of current develop branch (fixes conflict in `scrapers/matcher.go`)

WithoutPants (approved):

gitgiggety:
@WithoutPants you can be lucky I was using `git push --force-with-lease` :). I just rebased this on develop myself (obviously to fix migration as well). So I almost overwrote your changes / merge.

WithoutPants:
😬 It wouldn't have been a big deal anyway :)

PR #1668, opened 2021-08-23
Fix studio editing

InfiniteTF:


gitgiggety:
My fault strikes again :(

@WithoutPants also already addressed this as part of #1630, see https://github.com/stashapp/stash/pull/1630#issuecomment-902342409 / 381dee2 . But IMO this solution indeed is better, by creating a reusable method to do the conversation.

Issue #1360, opened 2021-04-30
[Feature] Improve Gallery support in Performers page


n0bodyyouknow:
It would be nice to add similar functionality for Galleries which currently exists for Scenes when browsing/sorting in the Performers page.

1. Add a gallery count "button" at the bottom of each performer in grid view, where the scene count and tag count are currently shown.  Clicking the gallery count button would take you to Galleries page, filtered by the performer, just like it does for the scene count button.  (Also would be nice to add a gallery count column to the Performers list view, for UI consistency.)

2. Add a "galleries_count" option to the sort field selector, to sort by number of galleries associated with the performer.

3. Possibly also add a "media_count" option to the sort field selector, to sort by the total number of scenes AND galleries combined that are associated with the performer.

![Untitled](https://user-images.githubusercontent.com/25272063/116752532-b41e2480-a9d3-11eb-96cc-be9245952b67.jpg)

![Untitled 2](https://user-images.githubusercontent.com/25272063/116752970-5b02c080-a9d4-11eb-85c8-d698e3b158c9.jpg)

Issue #987, opened 2020-12-07
[Feature] Gallery Data Support for Performers (Sort Option & Performer Tab)


venturevoom:
**Is your feature request related to a problem? Please describe.**
Gallery Data is not fully shown for Performers at the moment

* Performer Page only states "_Stars in yy scenes_". 
* Sort Options doesn't provide "galleries_count" field

**Describe the solution you'd like**

* Performer Page should state "_Stars in xx scenes and yy Galleries_"
* Sort Options should provide galleries_count similar to scenes_count

![image](https://user-images.githubusercontent.com/72347089/101347837-f4e19980-388a-11eb-9975-84b8d36589ba.png)

![image](https://user-images.githubusercontent.com/72347089/101347661-b2b85800-388a-11eb-942a-5686dacf8317.png)

PR #1671, opened 2021-08-25
Performers listing improvements

gitgiggety:
* Add image and gallery count sort options
* Add image and gallery count to list view
* Use internationalization for list headers

Fixes: #1360 
Fixes: #987

### DIFF
--- a/pkg/sqlite/performer.go
+++ b/pkg/sqlite/performer.go
@@ -500,6 +500,12 @@ func (qb *performerQueryBuilder) getPerformerSort(findFilter *models.FindFilterT
 	if sort == "scenes_count" {
 		return getCountSort(performerTable, performersScenesTable, performerIDColumn, direction)
 	}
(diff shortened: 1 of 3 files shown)


gitgiggety:
Added two more fixes, mostly target at mobile / small screens:
1. The `justify-content: center` on the table container meant the table was actually centered in the container while the table didn't fit the content, leading to the table "overflowing" to both the left and right, but you couldn't scroll to the left so the content was partially inaccessible.
2. In the scenes table the title was truncated, and in the performers table the name. Both are IMO not necessary as both these tables will most likely contain an image which already leads to the row being higher than one or two text lines. Furthermore both tables contain other fields which will most likely also be much longer (for scenes the tags and performers list, and for performer the aliases list) and thus already stretch the row height. So truncating the name after just a single line without any wrapping doesn't make sense to me.

PR #1673, opened 2021-08-25
Add "toolbar" buttons for some list actions

gitgiggety:
Personal pet peeve is having to click twice to start playback of selected scenes (once to open the actions menu and once more for "Play selected". This while the edit and delete buttons always appear right away after selecting a scene while I use those much less. So this fixes it my adding a play button next to the edit and delete buttons.

Also when adding images to a gallery for the first time I really had to search for how to do it. Obviously there is the `Add` tab, but after selecting some images I had to search how to actually save it, and apparently that's also in the `...` menu. But not anymore, as now it's also available in the toolbar. And applied the same for removing images from the gallery. So both are now available as button next to the edit/delete buttons.

This works by adding an optional `icon` property to the "otherOperations" list. When an icon is set the button will be shown. Optionally `buttonVariant` can be provided as well, which is used to make "Remove from gallery" button red (/apply `danger` variant).

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v090.md
+++ b/ui/v2.5/src/components/Changelog/versions/v090.md
@@ -12,6 +12,7 @@
 
 ### 🎨 Improvements
 * Move Play Selected Scenes, and Add/Remove Gallery Image buttons to button toolbar. ([#1673](https://github.com/stashapp/stash/pull/1673))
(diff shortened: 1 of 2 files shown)


gitgiggety:
Forgot to mention. "Play selected" seems to be broken, also on my "production" environment, so not related to these changes.

PR #1678, opened 2021-08-27
Load TLS config files from config path before stash home

### LINKED ISSUE #1552: [Feature Request] consider SSL cert and key in config path/current path
**Describe the bug**
stash.key and stash.cert are recognized when placed in the current path (same as conf, ffmpeg, database file, stash exec present)

**To Reproduce**
place stash.cert and stash.key in current exec path

**Expected behavior**
stash looks to stash.cert and stash.key like config, ffmpeg and start in https

**Screenshots**
None

v0.8.0

 - OS: Windows 10
 - Browser Firefox
 - Version 89

NA

**Additional**
Like to keep stash portable instead of using User Directory (~/.stash)

WithoutPants:
Fixes #1552

Changes TLS code so that it tries to load the `key` and `crt` files from the same directory as `config.yml` file, and falls back to `~/.stash`.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -27,7 +27,6 @@ import (
 	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/manager"
 	"github.com/stashapp/stash/pkg/manager/config"
-	"github.com/stashapp/stash/pkg/manager/paths"
 	"github.com/stashapp/stash/pkg/models"
 	"github.com/stashapp/stash/pkg/session"
 	"github.com/stashapp/stash/pkg/utils"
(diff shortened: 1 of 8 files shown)


pkg/api/server.go
```diff
@@ -328,27 +324,35 @@ func GetVersion() (string, string, string) {
 	return version, githash, buildstamp
 }
 
-func makeTLSConfig() *tls.Config {
-	cert, err := ioutil.ReadFile(paths.GetSSLCert())
+func makeTLSConfig(c *config.Instance) (*tls.Config, error) {
+	c.InitTLS()
+	certFile, keyFile := c.GetTLSFiles()
+
+	if certFile == "" || keyFile == "" {
```
bnkai, line 331: If a user supplies only one of the cert/key do we count that as a `broken` TLS configuration?
Should stash in that case fallback to http as it does now (dev and pr) or panic with an error instead?
```golang
	if certFile == "" && keyFile == "" {
```
for example would fallback to http if **none** of the key/cert file are provided and panic if only one of those is provided or either is a non valid certificate file.

The PR tests and looks ok.

WithoutPants: Changed so that if either SSL file is present, then both must be present, otherwise an error is returned.

PR #1682, opened 2021-08-28
DeoVR support

MaxiJazzer:
Any chance to get DeoVR support?


kermieisinthehouse:
This is not the place to make feature requests, please do not open PRs unless there is a code change attached.

You can use the 'Issues' tab.

PR #1684, opened 2021-08-29
Set NORMAL sync on sqlite3

SmallCoccinelle:
NORMAL sync is safe when using WAL journaliing.

It cuts the amount of sync calls to the disk, resulting in faster write
operation. On power loss, the database will perhaps lose some ongoing
commits, but that is all.

The expectation is that if the database lives on the same disk as the
stash, this could help performance quite a bit under heavier operation.

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -91,7 +91,7 @@ func Initialize(databasePath string) error {
 
 func open(databasePath string, disableForeignKeys bool) *sqlx.DB {
 	// https://github.com/mattn/go-sqlite3
-	url := "file:" + databasePath + "?_journal=WAL"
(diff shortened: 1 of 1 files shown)


SmallCoccinelle:
NORMAL sync avoids an fsync() after each transaction. This means that under power loss, you might have some transactions that get rolled back because the file system didn't synchronize yet. But the lower fsync() count speeds up the disk by quite a lot since it can avoid many fsync calls when there's many transactions in play.

The major concern is how the system handles a massive write, followed by a quick CTRL-C on the application right after. Here, we might lose a transaction or two. But if we close the database handle properly in a future commit, the close operation on the file should force the synchronization.

I've tested this under some load. And I didn't see any problems. Through I'm running this on ZFS, which might protect against some of the pathologies due to its ZIL handling.

PR #1685, opened 2021-08-29
Remove assignments to _

SmallCoccinelle:
It's a no-op if we are not using it, so it can be safely removed.


SmallCoccinelle:
This is more of a test-the-waters commit to figure out if this is something we want in the project or not.

If we want changes like these, we could also consider using the assignment:

```
input, ok := args[updateInputField]
if ok {
  ret = input.(map...)
  return ret
}

return make(map[...])
}
```

To check availability in the input map, rather than relying on the null-value default of input.

WithoutPants (approved):

PR #1686, opened 2021-08-29
When stopping, close the database

SmallCoccinelle:
This patch is likely to cause errornous behavior in the application.
This is due to the fact that the application doesn't gracefully shut
down, but is forcefully terminated.

However, the purpose is to uncover what needs to be done, to make
it a more graceful shutdown.


SmallCoccinelle:
Another test-the-waters patch.

In principle I'd like to follow up with the following:

* Graceful close of the HTTP API first, through use of the httpservers `Shutdown(...)` functionality.
* Handle each of the managers services and allow them to gracefully close. Of these, the job manager seems like where patches should go at a first glance.
* Start threading a proper termination context around in the system, together with a waitgroup for the services.

The longer term idea is that on a termination, we set a relatively short deadline, 15s, to stop jobs and clean up. If it can't happen in that timeframe, or we get another signal, we just terminate hard. Thus, the database can be left in a known good state, and we can finish up on half-done jobs upon termination, lowering the risk of data corruption on more odd configurations.

PR #1687, opened 2021-08-29
Heatmap and speed sorting/filtering for interactive scenes

agentfisk:
Generates heatmaps for interactive scenes and displays them in grid view

Screenshot from 2021-08-29 18-48-37

Generator also stores median stroke speed on the scenes record to display in grid view and allows to sort and filter scenes by speed.

Apologies for the one big commit, first time developing in years and contributing was an after thought.


WithoutPants:
From a static code review, it looks mostly ok. I think `heatmap` needs to be renamed to something more scoped to interactive scenes, since heatmaps can have other non-interactive types (like what Pornhub does, for example), and I don't want to paint us into a corner if we decide to add some similar functionality.

PR #1689, opened 2021-08-30
Add container_name to docker compose

Tweeticoats:
Setting container_name will set the container name to "stash" instead of the stack and container name ie "stash-stash-1".
For DLNA support we need to change from port mapping to host networking for this to work with docker.

### DIFF
--- a/docker/production/docker-compose.yml
+++ b/docker/production/docker-compose.yml
@@ -6,7 +6,11 @@ services:
     image: stashapp/stash:latest
     container_name: stash
     restart: unless-stopped
-    network_mode: host
+    ## the container's port must be the same with the STASH_PORT in the environment section
+    ports:
+      - "9999:9999"
+    ## If you intend to use stash's DLNA functionality uncomment the below network mode and comment out the above ports section
+    # network_mode: host
     logging:
       driver: "json-file"
       options:
--- a/docker/production/docker-compose.yml
+++ b/docker/production/docker-compose.yml
@@ -4,9 +4,9 @@ version: '3.4'
 services:
   stash:
     image: stashapp/stash:latest
+    container_name: stash
     restart: unless-stopped
-    ports:
-      - "9999:9999"
+    network_mode: host
     logging:
       driver: "json-file"
       options:
--- a/docker/production/docker-compose.yml
+++ b/docker/production/docker-compose.yml
@@ -17,6 +17,8 @@ services:
       - STASH_GENERATED=/generated/
       - STASH_METADATA=/metadata/
       - STASH_CACHE=/cache/
+      ## Adjust below to change default port (9999)
+      - STASH_PORT=9999
     volumes:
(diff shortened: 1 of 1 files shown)


bnkai:
As mentioned in discord since using host networking might lead to problems with other apps using the same port (9999) as stash, we could add to the env section something like the below
```yaml
    environment:
      - STASH_STASH=/data/
      - STASH_GENERATED=/generated/
      - STASH_METADATA=/metadata/
      - STASH_CACHE=/cache/
      ## Adjust below if needed to change stash's listen port
      - STASH_PORT=9999
```

kermieisinthehouse:
I don't think host mode networking should be the default, many people have docker networks firewalled, logged, etc. and this would just override that. A commented saying that it's necessary for UPNP and you can uncomment for that functionality would be fine though

bnkai:
@kermieisinthehouse 
So we would need something like this?
```yaml
# APPNICENAME=Stash
# APPDESCRIPTION=An organizer for your porn, written in Go
version: '3.4'
services:
  stash:
    image: stashapp/stash:latest
    container_name: stash
    restart: unless-stopped
    ## the container's port must be the same with the STASH_PORT in the environment section
    ports:
      - "9999:9999"
    ## If you intend to use stash's DLNA functionality uncomment the below  network mode and comment out the above ports section
    #network_mode: host
    logging:
      driver: "json-file"
      options:
        max-file: "10"
        max-size: "2m"
    environment:
      - STASH_STASH=/data/
      - STASH_GENERATED=/generated/
      - STASH_METADATA=/metadata/
      - STASH_CACHE=/cache/
      ## Adjust below to change default port (9999)
      - STASH_PORT=9999
    volumes:
      - /etc/localtime:/etc/localtime:ro
      ## Adjust below paths (the left part) to your liking.
      ## E.g. you can change ./config:/root/.stash to ./stash:/root/.stash
      
      ## Keep configs here.
      - ./config:/root/.stash
      ## Point this at your collection.
      - ./data:/data
      ## This is where your stash's metadata lives
      - ./metadata:/metadata
      ## Any other cache content.
      - ./cache:/cache
      ## Where to store generated content (screenshots,previews,transcodes,sprites)
      - ./generated:/generated
```

Also if we want to address the `generated` issue for now as mentioned in the discord channel we should probably remove 
the
```yaml
      - STASH_GENERATED=/generated/
```
from enviroment:
and the 
```yaml
      ## Where to store generated content (screenshots,previews,transcodes,sprites)
      - ./generated:/generated
```
from the volumes section
Existing users shouldnt be affected since they would have their own compose file or an existing config file

kermieisinthehouse:
I think the file as quoted is acceptable to merge, and the bug with ignoring STASH_GENERATED is a separate issue. (Which will require a UI change at minimum)

kermieisinthehouse (approved):

Issue #601, opened 2020-06-07
[Feature] Update path field via SceneUpdate


bugs181:
**Is your feature request related to a problem? Please describe.**
Being able to update the path field via `SceneUpdate` would allow third party utils to move the files around and update the media item without having to rescan files and re-add metadata.

Further investigation appears that the graphql `SceneUpdate` is part of `query` in the JSON. So adding `path` field to `SceneUpdateInput` would most likely allow us to use it.

**Describe the solution you'd like**
I'm new to how grapql works but I believe adding the path field to [line 15](https://github.com/stashapp/stash/blob/d886012d742beb558b96f401b23240323baf7ac4/graphql/documents/mutations/scene.graphql#L15) would be required.

**Describe alternatives you've considered**
The only alternative I could think of would be to create a new exposed API that would allow updating whatever info we wanted.

**Additional context**
This feature would be used in a future tool that I'm developing to auto-download files and add them to StashApp.

ConstantMoo:
Hi there, I'd also greatly appreciate having the `path` field available in `sceneUpdate` for my third party tool.

Since I'm pretty much clueless in Graphql and Golang, I naively tried to poke around `SceneUpdateInput` in `\graphql\schema\types\scene.graphql` and `sceneUpdate()` in `\pkg\api\resolver_mutation_scene.go`  but got compilation errors that I couldn't understand:

```
func (r *mutationResolver) sceneUpdate(input models.SceneUpdateInput, translator changesetTranslator, repo models.Repository) (*models.Scene, error) {
[...]

	updatedScene.Title = translator.nullString(input.Title, "title")
	updatedScene.Details = translator.nullString(input.Details, "details")
	updatedScene.URL = translator.nullString(input.URL, "url")
	updatedScene.Date = translator.sqliteDate(input.Date, "date")
[...]
//attempt to add path field
	updatedScene.Path = translator.nullString(input.Path, "path")
```

```
input SceneUpdateInput {
  clientMutationId: String
  id: ID!
  title: String
  details: String
  url: String
  date: String
  rating: Int
  organized: Boolean
  studio_id: ID
  gallery_ids: [ID!]
  performer_ids: [ID!]
  movies: [SceneMovieInput!]
  tag_ids: [ID!]
  """This should be a URL or a base64 encoded data URL"""
  cover_image: String
  stash_ids: [StashIDInput!]
//attempt to add path field
  path: String
}
```

Error:
```
go generate -mod=vendor
[...]

PR #1692, opened 2021-09-01
Update path field via SceneUpdate

ConstantMoo:
Hi there, this is my first PR for this project, so please be kind! 😄 

This PR adds the `Path` field to the `sceneUpdate()` and `BulkSceneUpdate()` API functions as requested in #601.

Following @stg-annon advice, there is a check if the new path value is either null or empty, pretty much like the CoverImage field. However there is no check if the new path is valid or not due to the following reasons:

* We can't know if the third party tool moving the video file is moving it before or after the API call is made.
* Changing the path field of a scene is a rather important change, it should be on the third party tool responsibility to make sure the video file was correctly moved.

**Known issue (help needed)**

* In the API playground doc, all the functions' input argument tabs auto-close by themselves after a short second, making them unreadable. Needs to be fixed before merging this PR (help appreciated!).

### DIFF
--- a/graphql/schema/types/scene.graphql
+++ b/graphql/schema/types/scene.graphql
@@ -70,6 +70,7 @@ input SceneUpdateInput {
   rating: Int
   organized: Boolean
   studio_id: ID
+  path: String
   gallery_ids: [ID!]
   performer_ids: [ID!]
   movies: [SceneMovieInput!]
@@ -100,6 +101,7 @@ input BulkSceneUpdateInput {
   rating: Int
   organized: Boolean
   studio_id: ID
+  path: String
   gallery_ids: BulkUpdateIds
   performer_ids: BulkUpdateIds
   tag_ids: BulkUpdateIds
--- a/pkg/api/resolver_mutation_scene.go
+++ b/pkg/api/resolver_mutation_scene.go
@@ -257,6 +257,10 @@ func (r *mutationResolver) BulkSceneUpdate(ctx context.Context, input models.Bul
 	updatedScene.StudioID = translator.nullInt64FromString(input.StudioID, "studio_id")
 	updatedScene.Organized = input.Organized
 
+	if input.Path != nil && *input.Path != "" {
+		updatedScene.Path = input.Path
+	}
+
 	ret := []*models.Scene{}
 
 	// Start the transaction and save the scene marker
--- a/pkg/api/resolver_mutation_scene.go
+++ b/pkg/api/resolver_mutation_scene.go
@@ -108,6 +108,10 @@ func (r *mutationResolver) sceneUpdate(input models.SceneUpdateInput, translator
 	updatedScene.StudioID = translator.nullInt64FromString(input.StudioID, "studio_id")
 	updatedScene.Organized = input.Organized
 
+	if input.Path != nil && *input.Path != "" {
+		updatedScene.Path = input.Path
+	}
+
 	if input.CoverImage != nil && *input.CoverImage != "" {
 		var err error
 		coverImageData, err = utils.ProcessImageInput(*input.CoverImage)


gitgiggety:
Could you explain why you added it to `BulkSceneUpdate` as well? This as the bulk update is AFAIK used to for example add / remove / set multiple tags on an item, update the studio, or performers. But there will be more items which share this (i.e.: you can select multiple scenes and set the same studio on them, or add a shared tag or performer). But as scenes are based on the video files you shouldn't be able to have multiple scenes which share the same path. So IMO it doesn't make sense to add this to the bulk update, as every scene should require a unique path.

> However there is no check if the new path is valid or not due to the following reasons:
>
>    We can't know if the third party tool moving the video file is moving it before or after the API call is made.

IMO it should still be validated and the tool must ve required to move the file beforehand. Otherwise the DB will be broken as well if the rename in the tool fails for whatever reason. And I guess it's more likely that a move / rename will fail than that the API call will fail.

Furthermore as @stg-annon already pointed out requiring the file to be available already at the new path also allows an `oshash` check to validate if it actually is the same file.

bnkai:
IMHO this functionality shouldnt be available in stash. For the same use case something like the rename plugin (https://github.com/stashapp/CommunityScripts/tree/main/plugins/renamerOnUpdate) or a simliar type of plugin is better suited

The current implementation updates only the path. The user from what i understand should rename/move the file himself and then go to stash and paste/edit the path? 
A simple selective scan would update the path correctly eg a user renames some files in a folder from the explorer then goes to stash and does a selective scan in the folder

If we had to add this functionality (again not a good idea IMHO) i would expect something like the below 
- user enters a new path value
- user should check an option or accept a popup window warning
- stash validates the new path (checks that its in the library paths)
- stash moves the file 
- if the move above was completed without error stash updates the path

ConstantMoo:
> Could you explain why you added it to BulkSceneUpdate as well? This as the bulk update is AFAIK used to for example add / remove / set multiple tags on an item, update the studio, or performers. But there will be more items which share this (i.e.: you can select multiple scenes and set the same studio on them, or add a shared tag or performer). But as scenes are based on the video files you shouldn't be able to have multiple scenes which share the same path. So IMO it doesn't make sense to add this to the bulk update, as every scene should require a unique path.

You are right, I modified `BulkSceneUpdate` where I thought I modified `scenesUpdate`, which by the way calls `sceneUpdate(*scene, translator, repo)` in a loop of the scene array it receives. My mistake, I'll revert this.

---

As for your concerns about the updated path check: fair enough, I'll try to work on that.

stg-annon:
I would agree with @bnkai that users don't need to be able to edit a scenes path from within stash's UI, however I do see the need to do it programmatically so the user is not always required to do some cleanup/organization task for things after a plugin/hook might rename a file.

First I think it would be best to avoid editing the database directly and only make edits to the DB via the API just to avoid issues, so ideally there would be a go to API endpoint for plugin devs to use in the case where a plugin might rename a file instead of doing what the [renamerOnUpdate](https://github.com/stashapp/CommunityScripts/blob/main/plugins/renamerOnUpdate/renamerOnUpdate.py#L469) does accessing the DB directly. 

Existing functionality can allow you to start an automatic scan of the parent folder of the new path of a renamed file which will automatically update the paths of any renamed files
e.x.
```json
{
    "operationName": "MetadataScan",
    "variables": {
        "input": {
            "paths": [
                "path\to\renamed\file1",
                "path\to\renamed\file2"
            ],
            "useFileMetadata": false,
            "stripFileExtension": false,
            "scanGeneratePreviews": false,
            "scanGenerateImagePreviews": false,
            "scanGenerateSprites": false,
            "scanGeneratePhashes": false
        }
    },
    "query": "mutation MetadataScan($input: ScanMetadataInput!) { metadataScan(input: $input)}"
}
```
The downside of this is that I see is that it's a one way operation that you would fire off at the end of the whatever operation the plugin is doing and would not get meaningful feedback, you would just hope that it updated all the paths.


The benefit that I see for adding the functionality of updating the paths of individual scenes from **sceneUpdate** is that you can then get feedback in the form of GQL errors for an individual file where you could then revert a rename operation if stash came back to you with an error for whatever reason (bad/inaccessible path, wrong hash, etc...)

WithoutPants:
I agree with @bnkai 's struck-through comments. I don't believe this functionality should be added to `SceneUpdate` at all. The internal stash system maintains it's view of the file paths, and we shouldn't allow external manipulation of these values. In the future, the way that files are maintained will be decoupled from the scenes as well, which would make it problematic to maintain this functionality.

As already mentioned, we already have the mechanism to rescan post-move using `MetadataScan`. If a synchronous version of this interface is required, then we can add that instead.

ConstantMoo:
@WithoutPants Thanks for your input. As suggested I tried `metadataScan` in my third-party tool and managed to update the path of my scenes as I wanted initially.

Considering the concerns raised by adding the path field to `sceneUpdate`, I'd suggest to close this PR and #601. If anyone needs to manually update the path of a scene this is the way to go:

1. Move the file with your third-party tool
2. Call `medataScan` providing the new path of your file (see @stg-annon comment above)
3. The scan identifies the moved video file with its hash and updates the scene path as intended.

PR #1693, opened 2021-09-01
Add new zh-tw localizations & move non-legacy localization keys

Still34:
# Summary

This PR updates the localization table for zh-tw, as well as fixing inconsistencies in the existing tables. Most notably, moving stats exclusive keys under the stats category, as opposed to leaving them in the root. Other similar entries are kept in the root as they are considered strings that are shared across the entire UI (e.g., "Scenes" may be used on the stats screen, on the navbar, etc.) Additionally, key naming inconsistencies are also corrected, where some key names used the kebab-case naming conventions instead of the snake-case one used by the rest of the table.

### DIFF
--- a/ui/v2.5/src/locales/en-GB.json
+++ b/ui/v2.5/src/locales/en-GB.json
@@ -237,7 +237,9 @@
       "sqlite_location": "File location for the SQLite database (requires restart)",
       "video_ext_desc": "Comma-delimited list of file extensions that will be identified as videos.",
       "video_ext_head": "Video Extensions",
-      "video_head": "Video"
+      "video_head": "Video",
+      "include_audio_head": "Include audio",
+      "include_audio_desc": "Includes audio stream when generating previews."
     },
     "logs": {
       "log_level": "Log Level"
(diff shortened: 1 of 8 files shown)


WithoutPants:
There's been a few localisation updates in the meantime, so this will need to be updated.

Still34:
Updated to the latest develop branch and made fixes from the conflicts.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -237,7 +237,9 @@
       "sqlite_location": "File location for the SQLite database (requires restart)",
       "video_ext_desc": "Comma-delimited list of file extensions that will be identified as videos.",
       "video_ext_head": "Video Extensions",
-      "video_head": "Video"
+      "video_head": "Video",
+      "include_audio_head": "Include audio",
+      "include_audio_desc": "Includes audio stream when generating previews."
```
WithoutPants, line 242: What are these? I don't see any reference in the tsx files.

Still34: Huh, it's supposed to be under the generated content settings. I thought I replaced the string in there. I'll fix it when I'm home, or you could make a commit. Either is fine by me.

WithoutPants (approved):

PR #1714, opened 2021-09-09
Fix studio scene count sort

WithoutPants:
Fixes SQL error introduced in #1660


gitgiggety:
I managed to break sorting without touching the sorting code? :(

Had to check the code what happened. Didn't know fallback `getSort` has some special logic for `*_count` colums. Which did break because the scenes table wasn't joined anymore.

PR #1715, opened 2021-09-09
Errcheck phase 1

SmallCoccinelle:
This stack of patches goes through unhandled errors in stash, and does the following:

* If there's an obvious error-return already in place, the error is returned.
* If not, a `logger.Warnf(..)` is issued with the error. My rationale is that the program has worked well up until this point, so it's not an error that will suddenly make the program fail.
* Wrap errors if there is more contextual data to the error in a few places.
* If we already have logging higher up the call stack, propagate the error.

in `func RunMigrations`, just defer the close of the migration data to the function epilogue through a `defer`.

in `config_concurrency_test.go`: prepare to check errors. Rename `t` to `k` so `t *testing.T` is available, pass the worker explicitly.

Unhandled errors tend to hide faults in programs. By logging these, we get a heads-up on those faults. It can also improve support on a program because the errors are fed back to the user in the log, so many users can fix their own problems.

There's a `phase 2` errcheck handling coming up as well. So this isn't all the unchecked errors in the software. But by now, this stack is large enough I'd like to get that in first if possible.

### DIFF
--- a/pkg/api/resolver_mutation_configure.go
+++ b/pkg/api/resolver_mutation_configure.go
@@ -289,8 +289,7 @@ func (r *mutationResolver) ConfigureDlna(ctx context.Context, input models.Confi
 		if !*input.Enabled && dlnaService.IsRunning() {
 			dlnaService.Stop(nil)
 		} else if *input.Enabled && !dlnaService.IsRunning() {
-			err := dlnaService.Start(nil)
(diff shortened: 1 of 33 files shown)


SmallCoccinelle:
Went ahead and simplified most of the warning handling to one-line initializers.

PR #1716, opened 2021-09-09
Make IHierarchicalLabeledIdCriterion backwards compatible

gitgiggety:
Add some form of backwards compatibility in the UI for
`IHierarchicalLabeledIdCriterion`. With this backwards compatibility
it's possible to recall saved filters which use the tags filter.
Otherwise stuff would blow up because the saved filter has a different
structure than what the `IHierarchicalLabeledIdCriterion` expects.


gitgiggety:
I'm guessing this is a nice to have to not annoy users with having to recreate every saved filter which contains a tag filter

PR #1718, opened 2021-09-10
[Bug Fix] Fix scene markers editor mobile UI

liquid-flow:
# Scope

Scene -> Markers

This PR fixes the mobile view of the scene markers editor on 320px and 576px breakpoints. It also includes some minor refactoring and UI improvements for the components in scope.

## Testing

1. Launch the server: `./stash`
2. Launch UI dev server: `make ui-start`
3. Navigate to any scene
4. Select `Markers` tab
5. Check the layout on the following breakpoints: 320px, 575px, 576px, 767px, 768px, 1023px, 1024px, 1199px, 1200px

### 320px
| Before | After |
| --- | --- |
|before-320|after-320|

### 576px
| Before | After |
| --- | --- |
|before-576|after-576|

### 768px
| Before | After |
| --- | --- |
|before-768|after-768|

### 1024
| Before | After |
| --- | --- |
|before-1024|after-1024|

### 1200px
| Before | After |
| --- | --- |
|before-1200|after-1200|

### DIFF
--- a/ui/v2.5/src/components/Shared/DurationInput.tsx
+++ b/ui/v2.5/src/components/Shared/DurationInput.tsx
@@ -93,7 +93,7 @@ export const DurationInput: React.FC<IProps> = (props: IProps) => {
   }
 
   return (
-    <Form.Group className={`duration-input ${props.className}`}>
+    <div className={`duration-input ${props.className}`}>
       <InputGroup>
         <Form.Control
           className="duration-control text-input"
@@ -122,6 +122,6 @@ export const DurationInput: React.FC<IProps> = (props: IProps) => {
           {renderButtons()}
         </InputGroup.Append>
       </InputGroup>
-    </Form.Group>
+    </div>
   );
 };
--- a/ui/v2.5/src/index.scss
+++ b/ui/v2.5/src/index.scss
@@ -217,6 +217,7 @@ div.react-select__control {
   border-color: $secondary;
   color: $text-color;
   cursor: pointer;
+  white-space: nowrap;
 
   .react-select__single-value,
   .react-select__input {
--- a/ui/v2.5/src/styles/_theme.scss
+++ b/ui/v2.5/src/styles/_theme.scss
@@ -78,7 +78,7 @@ hr {
 .nav-tabs {
   border: none;
   margin: auto;
-  margin-bottom: 0.5rem;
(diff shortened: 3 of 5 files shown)


gitgiggety:
On a phone in landscape the `Time` input is already shown on the same line as `Primary tag` but the label doesn't fit and is wrapped to `Tim` + `e` (or at least on my phone, and when emulating iPhone 6/7/8 (Plus)). Would be nice if that was fixed as well (in other words: keep showing the `Time` input on it's own line until the screen is (much?) wider)

ui/v2.5/src/components/Shared/DurationInput.tsx
```diff
@@ -93,7 +93,7 @@ export const DurationInput: React.FC<IProps> = (props: IProps) => {
   }
 
   return (
-    <Form.Group className={`duration-input ${props.className}`}>
+    <div className={`duration-input ${props.className}`}>
```
liquid-flow, line 96: This way we avoid adding the `form-group` class, which adds some margin at the bottom and thus makes all the forms containing `DurationInput` look inconsistent with the rest of the form inputs. Check the input:
1. Settings -> Interface -> Maximum loop duration
2. Settings -> DLNA -> Enable temporarily
3. Movies -> New -> Duration
4. Scenes -> Filter -> Duration

ui/v2.5/src/index.scss
```diff
@@ -217,6 +217,7 @@ div.react-select__control {
   border-color: $secondary;
   color: $text-color;
   cursor: pointer;
+  white-space: nowrap;
```
liquid-flow, line 220: To avoid long placeholder text bloating text inputs

ui/v2.5/src/styles/_theme.scss
```diff
@@ -78,7 +78,7 @@ hr {
 .nav-tabs {
   border: none;
   margin: auto;
-  margin-bottom: 0.5rem;
+  margin-bottom: 1.5rem;
```
liquid-flow, line 81: Render some space between the tabs and the form to separate them visually.

liquid-flow (commented):
@gitgiggety Good catch 👍 I did some further improvements. Please, check when you have time.

WithoutPants (approved):
Looks good. Thanks!

PR #1719, opened 2021-09-10
Support subpaths when serving stash through a reverse proxy

InfiniteTF:
Uses the `X-Forwarded-Prefix` header from the proxy to rewrite the basepath. It works by injecting the path into the index file. A more elegant approach might be to use a proper index template and instead fetch the manifest to pull in the bundles, but this approach works fine from what I can tell, and requires less changes.

Tested through nginx, directly on port 9999, and in dev mode, everyting *should* work fine.


WithoutPants:
This will need testing from someone with a reverse proxy to play with. We should also add some documentation of this behaviour to the README file. I'm not well versed with how this works, otherwise I'd do it myself. The reverse proxy wiki page already gives a more comprehensive set of setup instructions, so the README only needs to cover the behaviour of `X-Forwarded-Prefix` and `X-Forwarded-Port`.

I tried some manual queries with postman. When I used the following headers:
```
X-Forwarded-Prefix: /foo
X-Forwarded-Port: 6969
```
I ended up with the following URL in the `sceneStreams` query: `http://localhost:9999:6969/foo/scene/90/stream`

InfiniteTF:
> I tried some manual queries with postman. When I used the following headers:
> 
> ```
> X-Forwarded-Prefix: /foo
> X-Forwarded-Port: 6969
> ```
> 
> I ended up with the following URL in the `sceneStreams` query: `http://localhost:9999:6969/foo/scene/90/stream`

I think the issue is you're defining `X-Forwarded-Port` but not `Host`. Something is defaulting the host value to `host+port` so the port header would need to be dropped. I don't think it's worth handling this in stash, I assume anyone who is interested in using a reverse proxy can tweak their config themselves.

PR #1721, opened 2021-09-11
[Bug Fix] Disable float-on-scroll JWPlayer feature on mobile

liquid-flow:
This is a follow-up PR for https://github.com/stashapp/stash/pull/1496 (I couldn't figure out how to contribute to the original one).

It also updates JWPlayer to the latest version (8.22.0)

# Scope

UI: JWPlayer behavior on mobile

## Closes/Fixes Issues

Closes https://github.com/stashapp/stash/issues/1440

### Prerequisites
1. Start the server `./stash`
1. Start UI in development mode: `make ui-start`
1. Navigate to the app in mobile browser: `http://<ui_server_ip_address>:3000`

### Mobile
1. Select any scene, start playing, and trigger the full-screen mode
1. The video should go full-screen as expected

### Tablet/Desktop
1. Open any scene on tablet/desktop (ensure the one-column layout mode with the controls below the video, not on the left)
1. Select `Filters` tab, start playing the video, and scroll to the bottom of the page
1. The video should float to the bottom right of the page as usual


ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -314,6 +317,7 @@ export class ScenePlayerImpl extends React.Component<
       getDurationHook,
       seekHook,
       getCurrentTimeHook,
+      ...(!ScreenUtils.isMobile() && extras),
```
gitgiggety, line 320: IMO it's kinda odd to drop `extras` when on mobile (as it's not clear from the variable name). Personally I would change it to always include `extras` but either make it an empty object on mobile or explicitly enable/disable the floating player.

So either:
```js
const extras = {};
if (!ScreenUtils.isMobile()) {
  extras.floating = {
    dismissable: true,
  };
```

Or just change original and utilize `mode`:
```js
      height: "100%",
      floating: {
        dismissible: true,
        mode: ScreenUtils.isMobile() ? 'never', 'notVisible',
      },
      cast: {},
```

Note that the latter requires JWPlayer 8.17+ and I don't know which version is being used at the moment.

ui/v2.5/src/utils/screen.ts
```diff
@@ -0,0 +1,6 @@
+const isMobile = () =>
+  window.matchMedia("only screen and (max-width: 768px)").matches;
```
gitgiggety, line 2: Shouldn't `768px` be `767px`? IIRC both `min` and `max` are inclusive. So if CSS is using a mobile first approach and using `min-width: 768px` then at a screen of `768px` both the `min-width` and `max-width` will be hit.

gitgiggety (commented):
Nice that someone (finally) looks into this.

Checking the API docs myself now I believe it can be disabled a bit "cleaner".
Furthermore I believe that the switch from mobile / phone to tablet is inconsistent on screens which are exactly 768px wide.

See comments for explanation. But IMO neither are deal breakers, but would be nice if you could look into it.

liquid-flow:
@gitgiggety Thanks for the review and great suggestions. I adusted the PR and also updated JWPlayer to the latest version 8.22.0. Please have a look when you have time.

gitgiggety:
The upgrade does need some good regression testing though.

Following seems to work fine:
* Preview images on progress / seek bar
* Seeking using the scrubber (I guess that's the name used) at the bottom
* Scrubber being updated during playback
* Markers being shown in progress / seek bar
* Starting markers
* Changing playback rates (although most likely just internal to JW Player so can't really regress unless JW Player is broken)

Following does seem to be broken:
* Transcoding
  * Current position is lost (i.e.: video starts from the beginning again)
  * Video duration etc is lost. So during playback (/download of video) the duration keeps increasing from just a couple of seconds up to the entire length when the full file is downloaded

It seems like previously the request was, for example, `stream.webm?resolution=LOW&start=750.214595` while after the update it's just `stream.webm?resolution=LOW`, so it doesn't include the starting position anymore. But the question is: is this URL generated in stash, or in JW Player? In other words, it is a regression in JW Player, or a breakage on our end due to the upgrade.

liquid-flow:
@gitgiggety Reverted back to using `extras`.

I tried to hunt down where JWPlayer sets the `start` parameter in its codebase, but to no avail. We don't seem to be doing it in our code.

Moreover, when I built JWPlayer of the same version that we're currently using, it also lacked the `start` parameter.

We can have a follow-up ticket to investigate that and upgrade JWPlayer.

InfiniteTF:
> I tried to hunt down where JWPlayer sets the `start` parameter in its codebase, but to no avail. We don't seem to be doing it in our code.
> 
> Moreover, when I built JWPlayer of the same version that we're currently using, it also lacked the `start` parameter.

IIRC we use a modified build of JWPlayer. I'm not sure if this is documented anywhere. To upgrade JWPlayer you would have to port over the modifications.

gitgiggety:
> IIRC we use a modified build of JWPlayer.

In that case I would propose to leave it as is. AFAIK there also is the intention to switch to Video.js. So seems a bit too much to hunt down this issue/regression if JWPlayer is gonna be dropped in the future anyway. So just leaving JWPlayer in some form of "maintenance mode" and only fixing issues for now (for example fixing this annoying floating player on mobile :))

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -237,6 +237,7 @@ export class ScenePlayerImpl extends React.Component<
         },
       ],
       sources: this.props.sceneStreams.map((s) => {
+        console.log(s.url);
```
gitgiggety, line 240: I'm guessing this is some accidental leftover from debugging and thus can be removed?

WithoutPants:
The modified code lives here: https://github.com/WithoutPants/jwplayer/tree/hooks

It's documented in one of the earlier PRs: https://github.com/stashapp/stash/pull/268

It should probably be documented in the dev notes.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
liquid-flow: @gitgiggety Good catch. Removed.

WithoutPants (approved):

PR #1731, opened 2021-09-15
Execute Gallery.Create.Post plugin hook during scan

gitgiggety:
Fix issue where Gallery.Create.Post hook is not executed when a new
gallery is created during scan, when the gallery is created based on the
folder.

### DIFF
--- a/pkg/manager/task_scan.go
+++ b/pkg/manager/task_scan.go
@@ -1141,11 +1141,19 @@ func (t *ScanTask) scanImage() {
 		} else if config.GetInstance().GetCreateGalleriesFromFolders() {
 			// create gallery from folder or associate with existing gallery
 			logger.Infof("Associating image %s with folder gallery", i.Path)
+			var galleryID int
+			var isNewGallery bool
 			if err := t.TxnManager.WithTxn(context.TODO(), func(r models.Repository) error {
-				return t.associateImageWithFolderGallery(i.ID, r.Gallery())
+				var err error
+				galleryID, isNewGallery, err = t.associateImageWithFolderGallery(i.ID, r.Gallery())
+				return err
 			}); err != nil {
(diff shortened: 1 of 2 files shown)


gitgiggety:
Discovered this while working on a plugin which hooks into `Gallery.Create.Post`, but it wasn't getting called if a gallery is created based on a directory. Didn't check other methods to create a gallery, but based on code it seems like it should work correctly when the gallery is created based on a zip or manually.

Implementation is like this to allow the plugin to access the gallery. If the hook is invoked by `associateImageWithFolderGallery` then the transaction isn't committed yet which means the plugin can't access the gallery (as the transaction to create the gallery thus isn't committed yet, and a `findGallery()` call in the plugin thus won't find the gallery as the GraphQL call obviously uses a different database transaction).

gitgiggety:
After checking code once more I found that the ZIP scanning also falls for the same issue as I had. In other words: the hook being invoked while the transaction wasn't committed yet. So changed that as well.

pkg/manager/task_scan.go
```diff
@@ -1208,14 +1221,15 @@ func (t *ScanTask) rescanImage(i *models.Image, fileModTime time.Time) (*models.
 	return ret, nil
 }
 
-func (t *ScanTask) associateImageWithFolderGallery(imageID int, qb models.GalleryReaderWriter) error {
+func (t *ScanTask) associateImageWithFolderGallery(imageID int, qb models.GalleryReaderWriter) (int, bool, error) {
```
bnkai, line 1224: Just for readability issues, should the function be adjusted so that the return values are named (eg `galleryID int, isNew bool`)?

gitgiggety: Good suggestion, updated 👍

WithoutPants (approved):

PR #1733, opened 2021-09-16
Make hierarchical criterion depth input optional

WithoutPants:
Fixes backward compatibility with clients (now) using hierarchical multi criterion inputs. Makes depth optional, defaulting to 0.


WithoutPants:
@gitgiggety tagging you into this for review since you're most familiar with the recent changes.

gitgiggety:
Based on review: looks good to me. Thought I found some issue / oddity in the tests, where `...Criterion.Depth = X` assignments where replaced with `depth`. But those obviously work due to the reference being used.

PR #1737, opened 2021-09-17
Fix scene edit panel layout

liquid-flow:
# Scope

UI: SceneEditPanel

This PR fixes the layout on 320 and 1024 viewports (labels for Galleries, Studios, Movies/Scenes)

# How to test

1. Run the UI
2. Navigate to any scene, click the `Edit` tab
3. Check that the layout is consistent on the following breakpoints: 320, 576, 768, 1024, 1200

# Screenshots

## 320px

| Before | After |
|---|---|
|before-320|after-320|

## 1024px
| Before | After |
|---|---|
|before-1024|after-1024|


WithoutPants (approved):
Thanks, I hadn't tested on 320px width viewports.

FYI, changelog entries are added at the top of the list, I've amended the changelog to this effect.

PR #1740, opened 2021-09-18
Add indexes for path and checksum to images

gitgiggety:
The scenes table has unique indexes/constraints on path and checksum colums. The images table doesn't, which doesn't really make sense, as scanning uses these colums extensively which warrents an index, and both should be unique as well.

Adding these indexes thus heavily improves the scanning tasks performance. On a database containing 4700 images a (re)scan of those 4700 files, which thus shouldn't do anything, took 1.2 seconds, with the indexes added this only takes 0.4 seconds. Taking the same test on a generated database containing 4M images + the actual 4700 images took 26 minutes for a rescan, and with the index existing also only takes 0.4 seconds.

### DIFF
--- a/pkg/database/custom_migrations.go
+++ b/pkg/database/custom_migrations.go
@@ -1,0 +1,69 @@
+package database
+
+import (
+	"database/sql"
+	"strings"
+
+	"github.com/jmoiron/sqlx"
+	"github.com/stashapp/stash/pkg/logger"
+)
+
+func runCustomMigrations() error {
+	if err := createImagesChecksumIndex(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func createImagesChecksumIndex() error {
+	return WithTxn(func(tx *sqlx.Tx) error {
+		row := tx.QueryRow("SELECT 1 AS found FROM sqlite_master WHERE type = 'index' AND name = 'images_checksum_unique'")
+		err := row.Err()
+		if err != nil && err != sql.ErrNoRows {
+			return err
+		}
+
+		if err == nil {
+			var found bool
+			row.Scan(&found)
+			if found {
+				return nil
+			}
+		}
+
+		_, err = tx.Exec("CREATE UNIQUE INDEX images_checksum_unique ON images (checksum)")
+		if err == nil {
+			_, err = tx.Exec("DROP INDEX IF EXISTS index_images_checksum")
+			if err != nil {
+				logger.Errorf("Failed to remove surrogate images.checksum index: %s", err)
+			}
+			logger.Info("Created unique constraint on images table")
+			return nil
+		}
+
+		_, err = tx.Exec("CREATE INDEX IF NOT EXISTS index_images_checksum ON images (checksum)")
+		if err != nil {
+			logger.Errorf("Unable to create index on images.checksum: %s", err)
+		}
+
+		var result []struct {
+			Checksum string `db:"checksum"`
+		}
+
+		err = tx.Select(&result, "SELECT checksum FROM images GROUP BY checksum HAVING COUNT(1) > 1")
+		if err != nil && err != sql.ErrNoRows {
+			logger.Errorf("Unable to determine non-unique image checksums: %s", err)
+			return nil
+		}
+
+		checksums := make([]string, len(result))
+		for i, res := range result {
+			checksums[i] = res.Checksum
+		}
+
+		logger.Warnf("The following duplicate image checksums have been found. Please remove the duplicates and restart. %s", strings.Join(checksums, ", "))
+
+		return nil
+	})
+}
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -23,7 +23,7 @@ import (
 var DB *sqlx.DB
 var WriteMu *sync.Mutex
 var dbPath string
-var appSchemaVersion uint = 27
+var appSchemaVersion uint = 28
 var databaseSchemaVersion uint
 
 var (
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -86,6 +86,10 @@ func Initialize(databasePath string) error {
 	DB = open(databasePath, disableForeignKeys)
 	WriteMu = &sync.Mutex{}
 
+	if err := runCustomMigrations(); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/pkg/database/migrations/28_images_indexes.up.sql
+++ b/pkg/database/migrations/28_images_indexes.up.sql
@@ -1,5 +1,3 @@
 DROP INDEX IF EXISTS `images_path_unique`;
-DROP INDEX IF EXISTS `images_checksum_unique`;
 
 CREATE UNIQUE INDEX `images_path_unique` ON `images` (`path`);
-CREATE UNIQUE INDEX `images_checksum_unique` ON `images` (`checksum`);
--- a/pkg/database/migrations/28_images_indexes.up.sql
+++ b/pkg/database/migrations/28_images_indexes.up.sql
@@ -1,0 +1,5 @@
+DROP INDEX IF EXISTS `images_path_unique`;
+DROP INDEX IF EXISTS `images_checksum_unique`;
+
+CREATE UNIQUE INDEX `images_path_unique` ON `images` (`path`);
+CREATE UNIQUE INDEX `images_checksum_unique` ON `images` (`checksum`);
--- a/ui/v2.5/src/components/Changelog/versions/v0100.md
+++ b/ui/v2.5/src/components/Changelog/versions/v0100.md
@@ -1,3 +1,5 @@
+#### 💥 Note: Please check your logs after migrating to this release. A log warning will be generated on startup if duplicate image checksums exist in your system. Search for the images using the logged checksums, and remove the unwanted ones.
+
 ### ✨ New Features
 * Added options to generate webp and static preview files for markers. ([#1604](https://github.com/stashapp/stash/pull/1604))
 * Added sort by option for gallery rating. ([#1720](https://github.com/stashapp/stash/pull/1720))
@@ -10,6 +12,7 @@
 * Support filtering Movies by Performers. ([#1675](https://github.com/stashapp/stash/pull/1675))
 
 ### 🎨 Improvements
+* Added missing image table indexes, resulting in a significant performance improvement. ([#1740](https://github.com/stashapp/stash/pull/1740))
 * Support scraper script logging to specific log levels. ([#1648](https://github.com/stashapp/stash/pull/1648))
 * Added sv-SE language option. ([#1691](https://github.com/stashapp/stash/pull/1691))
 


gitgiggety:
@kermieisinthehouse can you test this :)

If I remember correctly you have a database with 4M actual files? So if it took 26 minutes on my system to check if 4700 files / paths are already in the database, and this index reduces it to under half a second than I'm guessing it took literally ages for you to check if 4M paths are already in the database, while with these changes I guess it will take a couple of minutes tops.

(You should obviously be able to just run the two `CREATE UNIQUE INDEX` statements on the `sqlite3` CLI without having to check out and run this branch)

SmallCoccinelle:
LGTM.

The speedup is really great. Even if this takes up more index space, it is likely to be warranted. Also, both columns are `NOT NULL` so there's no shenanigans that require partial indexes.

gitgiggety:
Just for fun reverted to the original database containing the 4M dummy images and ran a scan to get the 4700 images included as well. On develop this took about 1 hour and 25 minutes. With these added indexes it takes just 30 minutes. So with the indexes being added the actual importing scan is about as fast as it is without index to do a scan which doesn't do anything.

The reason why it's an hour slower without index is also quite simple. When a file is new there is the slow query to check if the path is found in the database, and because it isn't found the MD5 hash / checksum is calculated and a second query is run to check if the checksum is found (or the duplicate / move check). And both these queries will obviously be slow / way slower without index.

kermieisinthehouse:
This is fantastic! I can't believe something this simple was missed for so long! I can't wait to test this out, but I ran into another problem: I can't create the indexes because my image table includes duplicate checksums! Looking through the matches, the appear to be genuine duplicates. I don't understand how this happened, as I figured Stash would simply update the path upon finding one.

Is this a bug in stash? Is this expected behavior? Is my database wonked? Either way I'm sure I'm not the only one in this situation. If it helps for debugging purposes, my fork includes #1655 but I've skimmed through it and couldn't find the cause there.

If it is a bug, we should both fix it (so we don't just fight sqlite during a scan) and include a workaround for the people affected. `DELETE FROM images WHERE checksum IN (SELECT checksum FROM images GROUP BY checksum HAVING COUNT(checksum) > 1);` perhaps in the migration? This would lose any changes done manually to the image though...

kermieisinthehouse:
I ran that directive to finish the migration, and it's much faster! I'll get exact measurements soon, but it's several times faster at least.

My error log however is plagued with:
```
2021-09-18 11:00:22Error   FOREIGN KEY constraint failed
2021-09-18 11:00:22Error   FOREIGN KEY constraint failed
2021-09-18 11:00:22Error   FOREIGN KEY constraint failed
2021-09-18 11:00:22Error   FOREIGN KEY constraint failed
2021-09-18 11:00:20Error   FOREIGN KEY constraint failed
2021-09-18 11:00:20Error   FOREIGN KEY constraint failed
2021-09-18 11:00:20Error   FOREIGN KEY constraint failed
2021-09-18 11:00:20Error   FOREIGN KEY constraint failed
```
It seems image scanning does not handle duplicates correctly.

gitgiggety:
Right... Based on the code it indeed shouldn't be possible to have duplicate files (/files with the same index) in the database. This as when the path isn't found the checksum is calculated and validated if it doesn't exist (if if does, it's validated if the path for the existing item is still valid, ehen that's the case the item being processed is ignored for being a duplicate, if the path doesn't exist anymore the existing item is updated with the new path, as it being a file move).

Checking the code I do however believe there are two possible scenarios to trigger this:
1. Parallel scanning being used and the duplicates being imported at the same time. This as every query is ran in it's own transaction. This means that at the time of validating whether the checksum already exists for path A it isn't found, but by the time that A is inserted in parallel duplicate file A' is being processed as well. But when A' ia checked the checksum also doesn't exist (yet) but at time of inserting A' that A also is inserted already
2. Based on reading the code a bit it does seem like that modified files aren't validated for being a duplicate. So if you first have file A and file B they are both successfully, but if you then copy A over B, it only detects that file B is changed (modification time of the file differs from what's stored in the database), after which it calculates the checksum of the (current) file and stores it. But there doesn't seem any validation being done to see if the checksum of the new file B isn't a duplicate

Do any of these two possibilities seem likely to you? And maybe @WithoutPants has some other ideas about how you are able to create this database with duplicate checksums.
(Maybe you can backup the database and create a new one and run an import on that to see if it's reproducible on an empty database and a single scan being done).

gitgiggety:
> ERRO UNIQUE constraint failed: images.checksum    

So super easy to reproduce and 2. is valid.

1. Make sure you have images imported
2. Delete one of the images
3. Copy one of the remaining files and give it the name of the deleted file
4. Run a new import
And then it will thus fail (either insert duplicate checksum or fail because of the unique constraint). @kermieisinthehouse can you confirm that this is most likely the reason you're having these duplicates? So copying files to names which are still in the database.

kermieisinthehouse:
@gitgiggety I don't think this is the reason. My filenames and paths change often, but almost never to things that have previously existed. Plus, when I did the first check, it was full of items that were genuine duplicates, and both existed "far away" from each other. As in:
Studios/StudioA/Galleries/ModelC/GalleryB
Models/ModelC/Galleries/StudioA/GalleryB

which are far away enough in the tree that no threads would touch both at the same time. They both have the same oshash, but never shared a path with anything else.

SmallCoccinelle:
This is something to check as well: `sha256sum <fileA>` and `sha256sum <fileB>` to figure out if they are really the same file, or if they just by accident match on the `oshash` (which I find unlikely, but better verify through a full hash)

gitgiggety:
@kermieisinthehouse then I would have no clue where the issue is coming from. But if any of you could try to reimport everything into a clean database that might give a clue if it's related to rescanning breaking things (possibly still related to changed files not checking whether the new file/contents has an already existing hash) or whether it's the initial import already failing (possibly still related to parallel tasks being used and a race condition between checking whether the hash is duplicate and actually storing the image).

@SmallCoccinelle `oshash` apparently isn't applied to images. Those only (/always) use md5 for hashing. And if I understand @kermieisinthehouse correctly they are actually duplicate files which should have been ignored as such. (But otherwise it can't hurt to do a sha256 comparison either. If the files do differ than it's highly unlikely that two hashing algorithms have a has collision).

SmallCoccinelle:
Did a couple of `sha256sum` on-disk comparisons. The files are the same.

I had the usual case where you have a file saved twice from a browser: `fn.jpg` and `fn(1).jpg` are thus the file names. The files are the same checksum-wise, but occurs twice in the database, probably because they were scanned as such.

kermieisinthehouse:
I can confirm the entries that try to get inserted with the same checksum are exact duplicates. I also created a fresh stash instance with an empty database, and performed a scan. It inserted a duplicate checksum within a few minutes, which I confirmed were actual file duplicates. I tried reading through the scan task but my go is not up to par to debug.

kermieisinthehouse:
Also this makes it orders of magnitude faster. I'm expecting to have my collection scanned under a week, when without it would've been around 3 months.

gitgiggety:
So it also happens during an import into a clean database...

Could you then check / answer the following:
* Do you have parallel scanning enabled? (This is logged as well: `Scan started with X parallel tasks`)
* The files which are imported while being duplicates, are those the same as were imported as dupes on your actual database? (If you're using parallel scanning and the issue is because two threads check the same checksum at the same time then it might be that it slips through at that time but is detected another time)
* Can you check when the dupes are imported? Logging contains the timestamp, and path, so you should be able to see if both paths are processed at the same time.
* Are any dupes detected at all? (As for me, I apparently have a duplicate from the source, and that's detected just fine)

To answer these questions you might want to enable logging to a file (Settings => Configuration => Logging => Log file)

(BTW: I do believe it's better to create a new issue for this. So this PR can be used to discuss adding the index and possibly to discuss a "migration" path to remove the duplicates / maybe just create a "normal" index and not a unique index / constraint etc. This also because I don't know whether I can fix the issue you're facing at all, and if we can one way or another do add the index then duplicates will be prevented anyway. And it might even be that the issue you're facing is in the scenes code as well but goes undetected because the query just fails. As at least the "rescanning" looks the same for scenes and images. So I believe that suffers from the same issue as I described above: if you copy a file using a name which is already in the database the duplicate isn't detected).

kermieisinthehouse:
I do have parallel scanning enabled. It happens with both 9 threads and 64 threads. I can test the other stuff in a bit, those logs move by very quickly.

Ultimately it's not a problem for this PR, which is a HUGE fix. We can simply drop the duplicate entries or keep all but one. The sql for the first is above, the sql for the second sounds convoluted (is it even possible in pure sql in this scenario?). I'm fine with the first, the duplicate image error rate seems much less that 1% of 1% for the two cases we've seen

WithoutPants:
Agree that the issue surrounding the possible addition of duplicate images belongs in a separate issue.

My gut feel is that there's not going to be a simple way out of this for those that already have duplicate entries: they are going to need to do some manual SQL work to remove duplicate entries if the migration does not work. I'm inclined to merge this in and accept that some people may need to manually fix their databases.

kermieisinthehouse:
It's probably for the best to avoid just deleting people's data.

We're ultimately going to get a flood of support requests asking why migrations are failing.

Diagnose the issue:
`SELECT path FROM images GROUP BY checksum HAVING COUNT(checksum) > 1; `
Just delete the bad entries:
`DELETE FROM images WHERE checksum IN (SELECT checksum FROM images GROUP BY checksum HAVING COUNT(checksum) > 1);`

gitgiggety:
> I can test the other stuff in a bit, those logs move by very quickly.

That's why I described how you can enable file logging :)

> is it even possible in pure sql in this scenario?

It should be possible. Just select the `MIN` or `MAX` id in the group as well and remove those where the path matches but it's not that id. (Luckily we're not using MySQL, as that does trigger an error when you try to run a subquery on the same table as you're performing a mutation (insert / update / delete) on...)
But obviously any SQL solution would keep a "random" item which might not have any data set while one of the duplicates does have data set.
Query would thus be something aling the lines of:
```sql
DELETE FROM images
WHERE id IN (
    SELECT id FROM images
    LEFT JOIN (
        SELECT path, MIN(id) AS id
        FROM images
        GROUP BY path HAVING COUNT(1) > 1
    ) d ON i.path = d.path AND i.id != d id
)
```

> they are going to need to do some manual SQL work to remove duplicate entries if the migration does not work

Wouldn't a problem be that not everyone is a user with a technical background and thus might not know "how to fix it"?
And can some form of documentation be added somewhere? Previously the changelog sometimes contained some message that for example a rescan was required. But if the migration is triggered the user won't see the changelog until afterwards. But the migration page also can't show a message, correct? You'll just get an error the migration failed and that's it? No way to add some extra information like "might be you have duplicate images, see this page for how to fix it".

WithoutPants:
An alternative might be to not tie this change with a new schema version:
- amend the original image migration sql (13) to include the missing index - this means that new systems will have the correct index from the beginning
- add a check on startup to check if the index exists - if not, try to add it. If it fails, then log a detailed error (ideally listing duplicate image entries), but don't prevent start up
- add a note in the changelog indicating the changes, to direct users to the log file

gitgiggety:
Great idea 👍 

_But_ doing such a check on startup might be a bit lame? Especially as it will need to be supported till the end of time. But I think I saw something about a post migration hook as well in the code? So can we add an empty migration for schema version 28, and post migrate to 28 check if there are no duplicates and run the query? And otherwise log the error and query the user has to run.
Then the check to add the index will be part of the migration and not trigger on every startup.

WithoutPants:
The problem is that we want the user to be able to remove the problem images within the app (not via some third-party tool). The user needs to be able to use the system to remove the said images, then some mechanism needs to be in place to re-run the migration when it is able to do so.

I do agree that doing a check on every startup isn't ideal. Another alternative is to add a manual migration button, but that is probably over engineering things as well.

kermieisinthehouse:
I have to disagree on that approach. We have to consider user friction and adoption of this performance boost. Very few people know how to touch sqlite and many people complain of slow image scans. Telling them to either get their hands dirty or suffer slow scans is a bad approach.

I think deleting all but one of the duplicates as part of the migration is the way to go here. No risk of breaking a migration, everyone gets the speedup, no need to touch sqlite for end users, and very little chance of dropping important data - it should all exist in a duplicate image's entry anyway.

gitgiggety:
> The problem is that we want the user to be able to remove the problem images within the app (not via some third-party tool). The user needs to be able to use the system to remove the said images, then some mechanism needs to be in place to re-run the migration when it is able to do so.

Right, true as well. But in the log it would still be hard to explain which images are involved. Only listing ids doesn't help them in finding them. Listing by path would generate a massive log entry I guess (especially in case of long paths / many nested directories).

> many people complain of slow image scans

Which will most likely be partially fixed by #1655, as currently the images are fully decoded on import to get the size instead of just reading the metadata (so much more io is involved).
Furthermore @WithoutPants proposal is to add the index if it doesn't exist yet (and is possible to add it). So the only users affected would be those who already have duplicates in the database, as for those the index can't be created. And optionally we might even choose to create a normal index in that case, instead of a unique index / constraint. You will then still get the performance benefit without need to remove the duplicates.

I'll try to get something like this implemented today (which I believe would be "tomorrow" for you 😃) . Most likely including the fallback to a normal index. (And on startup check if the unique index exists, and if it doesn't check if there are duplicates and create the unique index and remove the fallback if there aren't any).

kermieisinthehouse:
No need for application side handling / conditional migrations / startup checks. I tested the code below to successfully prepare the database to accept the new unique indexes, you can just put it at the beginning of the migration:

```
DELETE FROM images 
WHERE checksum IN ( 
    SELECT checksum FROM images 
    GROUP BY checksum 
    HAVING COUNT(1) > 1
)
AND id NOT IN (
    SELECT min(id) FROM images 
    GROUP BY checksum 
    HAVING COUNT(1) > 1
);
```

PS. I've narrowed down the duplicate inserts to almost certainly being a concurrency problem. I haven't been able to reproduce running 1 thread.

WithoutPants:
> Right, true as well. But in the log it would still be hard to explain which images are involved. Only listing ids doesn't help them in finding them. Listing by path would generate a massive log entry I guess (especially in case of long paths / many nested directories).

Images can be searched by entering the checksum into the query text field, so listing checksums would be sufficient.

> I think deleting all but one of the duplicates as part of the migration is the way to go here. No risk of breaking a migration, everyone gets the speedup, no need to touch sqlite for end users, and very little chance of dropping important data - it should all exist in a duplicate image's entry anyway.

I'd rather migrate as many users as possible, and give the remaining users a clear path to fix the issue than delete random data from their database.

gitgiggety:
> Images can be searched by entering the checksum into the query text field, so listing checksums would be sufficient.

That's great and would save a great deal 👍 

Just thinking. I guess the migration / startup script could be as simple as:
```
if err := <run query>("CREATE UNIQUE INDEX IF NOT EXISTS images_checksum_unique ON images (checksum)"); err == nil {
    return;
}

<run query>("CREATE INDEX IF NOT EXISTS index_images_checksum ON images (checksum)")

<run query>("SELECT checksum FROM images GROUP BY checksum HAVING COUNT(1) > 1");
<log warning (or error or info?) for results of previous query>
```

No need to check if the index exists, as I guess the `CREATE UNIQUE INDEX IF NOT EXISTS` will still be successful when it already exists, just like when it is successfully created.
If the unique index doesn't exist create a normal / non unique index to get the performance gains, and log the duplicates.

SmallCoccinelle:
> Checking the code I do however believe there are two possible scenarios to trigger this:
> 
> 1. Parallel scanning being used and the duplicates being imported at the same time. This as every query is ran in it's own transaction. This means that at the time of validating whether the checksum already exists for path A it isn't found, but by the time that A is inserted in parallel duplicate file A' is being processed as well. But when A' is checked the checksum also doesn't exist (yet) but at time of inserting A' that A also is inserted already

This is definitely possible:

* The read transactions on checksum find are run separately from the write transaction. So anything can really happen in between.
* Since we have WAL mode enabled, the database uses *snapshot isolation* so we'll hit the problem anyway, even if we coalesced the transactions, because the update to the checksum doesn't have a unique index and thus won't conflict.

If there's a unique-index on the checksums, then we can detect the problem and abort the write.

For files close in the file system it is very likely that this happens. The database will have everything in memory and be fast-access, whereas reading from disk to compute the checksum will be slow. So any goroutine will be using the slow computation of a checksum as a soft-sync point. Once the file cache has the file, all checksum computation will be fast, putting pressure on the write window to have a conflicting write.

This doesn't explain files far away from each other in the file system, so I'm inclined to guess there are multiple faults in the code base, and this is just one of them.

If we want early-detection of this, we'll need to have the goroutines enter checksums they are working on in a map protected by a mutex, or by using sync.Map. But it's probably easier to let the database detect the update conflict.

I suggest we enable the unique-constraint, and then figure out where the errors occur in the program. This should explain where the conflicts are.

gitgiggety:
Updated. As discussed during startup there is a check if the unique constraint exists. In case it doesn't it tries to create it. When creating the unique index fails a normal index is created (so those users get the performance benefit as well), and the duplicate checksums are queried and written as warning to the log.
When the unique constraint is created successfully the surrogate will be removed.

Do note that I choose to ignore any actual errors in the process, as everyone works "fine" without indexes anyway (as it's the same as the current code). So IMO startup shouldn't fail on this.

@WithoutPants feel free to change the (log) messages, and I suggest you write warning in the changelog as well :)

PR #1648, opened 2021-08-17
Support scraper logging to specific log levels

stg-annon:
Resolves #1348

I don't have any experience with Go most of this was pulled directly from parts of how plugins handle logging

PR #1741, opened 2021-09-18
Scraper log improvements

gitgiggety:
This makes two fixes / changes:
1. It fixes logs not being shown in the web UI as reported in #1648 (`logger` in the `logger` package is the logrus instance, not a reference to the package)
2. It prefixes the log lines not just with `[Scrape]` / `[Plugin]` but includes the name of the scraper / plugin as well so it's easier to distinguish them

@WithoutPants I might have found a third unrelated issue. `logger.Trace()` doesn't add the log to the list of "UI logs" so is never shown in the UI, while `Tracef` is, as seen here:
https://github.com/stashapp/stash/blob/66f92c5dccaed02ef0a6caa2669ad0e890dcf929/pkg/logger/logger.go#L183-L185
(Compare `Trace` with `Tracef`, `Debug` etc).
Do you want me to fix this as well, or is there some reason to not append unformatted trace messages to the log? (Possibly because of spamming a lot or whatever).

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -182,6 +182,11 @@ func Progressf(format string, args ...interface{}) {
 
 func Trace(args ...interface{}) {
 	logger.Trace(args...)
+	l := &LogItem{
+		Type:    "trace",
+		Message: fmt.Sprint(args...),
+	}
+	addLogItem(l)
 }
 
 func Tracef(format string, args ...interface{}) {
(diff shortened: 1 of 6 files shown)


stg-annon:
I mentioned this in discord and it probably more suitable as a different PR but I thought I'd mention it here

Would anyone else see the benefit to being able to filter by log type just like how we filter by log level currently. I.E. no only could you pick debug level but you could choose from [combined, stash, plugin, scraper] to select out one of the logging types

I mention this without knowing how difficult a task this would actually be but it seems like the logs are already categorized by the prefix tag on the log line so that it should be possible to distinguish them easily

WithoutPants (approved):
Missing call in `Trace` looks like an oversight to me. Fixed.

PR #1749, opened 2021-09-21
Only close DB if not nil

WithoutPants:
Fixes panic when stopping the stash process before the database is initialised (eg during new system or pending migration).


SmallCoccinelle:
LGTM. I've hit this as well once, and it's a nice fix.

Issue #1638, opened 2021-08-13
[Feature] Movie gridview cosmetic consistency


echo6ix:
**Is your feature request related to a problem? Please describe.**
Cosmetically the movies page isn't consistent with the rest of Stash's gridview pages

**Describe the solution you'd like**
- Display the scene count as a button
- Display the date beneath the movie title

gitgiggety:
Started working on this. Currently the "Scene number" is shown on the Movies tab of a scene. Question is how to display this after the changes. When adding "everything" (date, synopsis, popover button for scene count) I don't think it really fits in anymore.
At first I thought to show it on the scenes count button / popover, for example `2/5` for scene 2 out of 5. But the number / index is arbitrary, so you can have a movie with 5 scenes, but only have 2 available in stash, for example scene 3 and 4, so the button would then show `3/2` which is really odd.

echo6ix:
> When adding "everything" (date, synopsis, popover button for scene count) I don't think it really fits in anymore.

Don't add everything?

I think the scene count number and date is enough at this point. Adding the number/index is confusing; it breaks the continuity from the rest of Stash of having scene popover icon only display a total scene count.

Is it /really/ that useful to show the synopsis in the movie cards view or is that overkill that begins to clutter the cards ui with too much information?  displaying the synopsis in the card (if my assumption is correct) breaks the simple visual symmetry of the movie poster grid view. 

Down the road it might be cool to add a performer popover and image count popover, and perhaps even an organized popover icon.

PR #1758, opened 2021-09-23
Movie card visual consistency

gitgiggety:
Style the movie card the same as the scene card.
* Show date
* Show synopsis
* Move scene count to "button" which shows the scene names on hover
* Show scene number / index as "button"

Fixes: #1638


WithoutPants:
Here is what a Movie card now looks like in the Movies page:
image

When viewed from a scene, it looks like this (scene index is 1):
image

I think the scene index display is not intuitive at all. It's not clear to me that that number is the scene number in the movie. I think a variation on the original display should probably be used. I had a quick crack, but it probably could be better:

image

gitgiggety:
The scene number / index obviously was a "hack" and I didn't know how to properly show it either. Adding it as just a `Scene #x` is a possibility as well .

But I myself don't use / have any movies in Stash, so I don't really care how it's shown :) So let me know how you want it to look and I'll change it. (Or if you've got the time or want to try it yourself, just change it).

WithoutPants (approved):

PR #1759, opened 2021-09-23
Add golangci-lint workflow

SmallCoccinelle:
This patch adds a `golangci-lint` action workflow in parallel to the normal build. The rationale for picking a parallel run is that it doesn't impede the build times of the existing build, and we don't mess with the existing build as well.

* Use our docker image for `go generate`. This makes sure the generated source stays consistent to our normal build.
* Use `golangci-lint`'s action to run the build.
* Prefer `-mod=vendor` as this is what happens in the rest of the code base.

There are currently 2 failing errchecks due to unchecked buffers. Everything else has been addressed by earlier patches to the project.

Once up and running, it should provide annotations directly to PRs with lint-problems it finds. False positives can be handled by a `//nolint:..` comment, as per `golangci-lint` documentation.


SmallCoccinelle:
I've seen it timeout a `1m` before, especially on a first run, so lets bump to `3m` and see if that helps.

SmallCoccinelle:
The `3m` timeout seem to have worked nicely.

We can matrix it to OSX/Windows on top if we want, but I think it's fine to do this on Linux only.

SmallCoccinelle:
(Oh, and you can see the errors in "files changed")

WithoutPants:
Looks good. Can we get a make target to run the linter locally, and maybe add instructions to install the linter added to `README.md`?

SmallCoccinelle:
Excellent suggestions. Added.

WithoutPants:
I think (with some modification), this can be made a replacement for most of our existing validation.

Specifically: `fmt-check`, `vet` and `lint`. All of these can be enabled with `gofmt`, `govet` (which is already enabled by default), and `revive`.

As such, I think these should be enabled in the `golint-ci` settings, and the relevent commands/targets removed from the Makefile and build.yml. The `README.md` can also be amended to remove the mention of `revive`. We can also remove the script for `fmt-check`.

SmallCoccinelle:
This makes sense.

Do you want to push this before or after such a change? It's definitely going to happen, it's more about if you want the CI lint in with said change, or if we want to get the CI linter in first, then lift the existing checks on top of it.

SmallCoccinelle:
With the current patches, we have the additive parts done and golangci-lint runs fmt checks and revive.

Once we are happy that this is the direction we want, we can go ahead and remove the tools we don't need anymore.

SmallCoccinelle:
This should clean up the older ways of doing things, and use golangci-lint as the meta-linter for everything.

Also, the `new-from-rev` issue setting should clean up most of the validation errors we have from before the introduction of golangci-lint. I intend to fix them later, but don't want other people to be impeded going forward.

Makefile
```diff
@@ -133,23 +133,16 @@ generate-stash-box-client:
...
-	go vet -mod=vendor ./...
-
 .PHONY: lint
-lint:
-	revive -config revive.toml -exclude ./vendor/...  ./...
+lint: golangci-lint
+
+.PHONY: golangci-lint
+golangci-lint:
+	golangci-lint run
```
WithoutPants, line 141: This is redundant. Just merge the `golangci-lint` target into `lint`.

README.md
```diff
@@ -125,17 +125,16 @@ NOTE: The `make` command in Windows will be `mingw32-make` with MingW.
 * `make pre-ui` - Installs the UI dependencies. Only needs to be run once before building the UI for the first time, or if the dependencies are updated
 * `make fmt-ui` - Formats the UI source code
 * `make ui` - Builds the frontend
-* `make vet` - Run `go vet`
-* `make lint` - Run the linter
+* `make lint` - Run the meta-linter on the backend
+* `make golangci-lint` - Run the golangci meta-linter
```
WithoutPants, line 129: As above, get rid of `golangci-lint`.

WithoutPants (changes requested):

Makefile
SmallCoccinelle: Yeah, I was thinking the same when I wrote it, but decided against it if more linters were incoming. But I guess we could just coalesce it for now. I'll push a change.

PR #1761, opened 2021-09-24
Disallow access in publicly exposed services

kermieisinthehouse:
Several in-the-wild unprotected stash instances have been found, disclosing their contents on the public internet.

This patch tries to prevent that from being possible. When a request is received from an IP that is not a private network and authentication is not enabled, it logs details, blows a tripwire, and shuts down the server to protect the data. Stash will not start again until a config option, `security_tripwire_accessed_from_public_internet` is changed.

Users of external auth can set `dangerous_allow_public_without_auth` to `true` to continue using their setups. 

Also created for reference: https://github.com/stashapp/stash/wiki/Authentication-Required-When-Accessing-Stash-From-the-Internet


pkg/manager/config/config.go
```diff
@@ -138,6 +138,12 @@ const SlideshowDelay = "slideshow_delay"
 const HandyKey = "handy_key"
 const FunscriptOffset = "funscript_offset"
 
+// Security
+const DangerousAllowPublicWithoutAuth = "dangerous_allow_public_without_auth"
+const DangerousAllowPublicWithoutAuthDefault = "false"
+const SecurityTripwireAccessedFromPublicInternet = "security_tripwire_accessed_from_public_internet"
+const SecurityTripwireAccessedFromPublicInternetDefault = "false"
```
WithoutPants, line 145: These aren't accessed in other packages so can be made package-local.

pkg/api/server.go
```diff
@@ -64,25 +68,54 @@ func authenticateHandler() func(http.Handler) http.Handler {
...
+						logger.Error("Stash has been accessed from the internet, without authentication. \n" +
+							"This is extremely dangerous! The whole world can see your stash page and browse your files! \n" +
+							"You probably forwarded a port from your router. At the very least, add a password to stash in the settings. \n" +
+							"Stash will not start again until you edit config.yml and change security_tripwire_accessed_from_public_internet to false. \n" +
+							"More information is available at https://github.com/stashapp/stash/wiki/Authentication-Required-When-Accessing-Stash-From-the-Internet")
+						c.Set(config.SecurityTripwireAccessedFromPublicInternet, true)
+						w.WriteHeader(http.StatusForbidden)
+						w.Write([]byte("You have attempted to access stash over the internet, and authentication is not enabled." +
+							"This is extremely dangerous! The whole world can see your your stash page and browse your files!" +
+							"Please read the log entry or visit https://github.com/stashapp/stash/wiki/Authentication-Required-When-Accessing-Stash-From-the-Internet"))
```
WithoutPants, line 113: Should this mention that the system is shutting down?

WithoutPants (changes requested):

kermieisinthehouse:
Fixed those notes. I also can't get the server to start with this patch. It looks like cfg.GetSecurityTripwire(...) is always true. Can't figure that one out. Also, the server doesn't seem to exit, simply hang and not answer requests. Which has me thinking, should we just not shut down the server, but instead 403 every request after the tripwire is blown?

kermieisinthehouse:
I've stopped the server from shutting down, now it just closes the database and returns every request with a little snippet about what's happening. I figure this will stop support requests for "My server crashed for no reason and I can't be bothered to read the logs"

I still have the problem with it being triggered all of the time. I don't know how the config options work well enough to fix this. Also it seems that if the config file is missing, it triggers too.

kermieisinthehouse:
I've fixed the last few bugs and tested it. It seems all good now. The protection is not applied to the first-time setup, but I think that's fine.

pkg/api/server.go
```diff
@@ -61,28 +73,59 @@ func authenticateHandler() func(http.Handler) http.Handler {
...
 				}
-
-				// otherwise redirect to the login page
-				u := url.URL{
-					Path: "/login",
+			} else {
+				//authentication is not required
+				//security fix: traffic from the public internet with no auth is disallowed
+				if !c.GetDangerousAllowPublicWithoutAuth() && !c.IsNewSystem() {
+					requestIPString := r.Header.Get("X-FORWARDED-FOR")
```
gitgiggety, line 103: Using `X-FORWARDED-FOR` without any checks doesn't make it secure. An attacker can just include the `X-FORWARDED-FOR` header in the request with a private IP address in it and the request will be accepted, right?

This needs a "trusted proxies" settings so the remote IP is validated to be a proxy which may be used, in that case `X-FORWARDED-FOR` may be accepted, otherwise the request should still be denied (/trip the wire). The "trusted proxies" could then still be set to a default of the private IP ranges for example.

gitgiggety: Personally I would most likely implement it as:
1. Check if `X-FORWARDED-FOR` is set
2. If it's set validate whether the remote IP of the request complies with a "trusted proxies" setting
3. If the remote IP doesn't comply deny the request (but don't trip the wire)
4. If `X-FORWARDED-FOR` is set it can now safely be used

This would require users of a proxy to set up this trusted proxies setting, but IMO that should be fine and at least makes it secure by default.
If that's not desirable I would allow private subnets by default until the user sets it.

gitgiggety: Another comment about `X-FORWARDED-FOR`. As far as I know this header can contain multiple addresses. This in case that a request passes through multiple proxies. But I'm unsure what it's value is in that case (comma separated? colon separated?). But I guess that all IP addresses should be validated in that case. All except the last (or first?) must be validated as trusted proxy, and the remaining should in that case thus comply with a private subnet.

kermieisinthehouse: Good catch!

Although now that I'm thinking about it, I'm tempted to just remove the header check. It's a de-facto non-standard that can contain anything and everything, so it would be very hard to validate all possible values. Also, what would the usecase be where you are accessing stash from a proxy outside its local network? If you are, you probably know enough to not expose it to the internet or to set up your own auth if you are.

gitgiggety: Doesn't removing it circumvent this fix?

If I'd be running a reverse proxy in my home, which does expose Stash (for example putting Traefik in front of it and using a Lets encrypt certificate for it) then Stash could be accessible from the internet again. Because the proxies IP address would be private and the IP of the actual requestor isn't validated anymore.

kermieisinthehouse: I think that if you know how to set up traefik or even know what a reverse proxy is, you aren't the target of this patch. This is for the 50 or so (mostly windows gaming machines) that have all of their ports forwarded. There are some docker installs publicly accessible too, but still.

gitgiggety: I do get your point, but at the same time the same IMO applies to those Windows users which just add a port forward for all available ports.

In both cases an ignorant user applies a faulty configuration which makes them vulnerable. Whether it's a user which just applies a port forward for all ports or whether it's a user which has either Traefik running with the configuration to expose all containers it does find, the user manually exposing it to Traefik and having to make it publicly accessible for the HTTP ACME challenge and forgetting to set up authentication or whatever. (And there most likely are much more scenarios, using a different reverse proxy which might always expose all services it finds. The computer not being behind a NAT-router so exposing all ports to the internet as well, running Stash on a server / VPS, ...).

In both cases it is because an ignorant user applies the "wrong" configuration which exposes their Stash instance. And only protecting a small group of users thus feels "wrong" IMHO.

kermieisinthehouse:
@gitgiggety moving to comments for visibility.

I agree. I've added a trusted proxies section, and am allowing proxy chains from the local network by default. They still don't allow access from endpoints on the public internet. This should prevent the cases you've mentioned, such as leaving a reverse proxy exposed accidentally (provided it send X-FORWARDED-FOR). It also prevents spoofing x-forwarded-for to bypass the protection.

@WithoutPants when this is merged, is it possible to make a backported 0.9.1 release with just this? I don't have exact numbers but a decent chunk, maybe most, of affected instances ran the latest stable, so it's hopeful to assume they'd take a point release too, protecting them.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -193,7 +193,9 @@
         "password_desc": "Password to access Stash. Leave blank to disable user authentication",
         "stash-box_integration": "Stash-box integration",
         "username": "Username",
-        "username_desc": "Username to access Stash. Leave blank to disable user authentication"
+        "username_desc": "Username to access Stash. Leave blank to disable user authentication",
+        "trusted_proxies": "Trusted proxies",
```
gitgiggety, line 197: I believe the intention is to sort these by on the keys. So it should be above:
```
"stash-box_integration": ...,
"trusted_proxies": ...,
"trusted_proxies_desc": ...,
"username": ...,
```

pkg/api/resolver_mutation_configure.go
```diff
@@ -137,6 +137,10 @@ func (r *mutationResolver) ConfigureGeneral(ctx context.Context, input models.Co
 		c.Set(config.MaxSessionAge, *input.MaxSessionAge)
 	}
 
+	if input.TrustedProxies != nil {
+		c.Set(config.TrustedProxies, *input.MaxSessionAge)
```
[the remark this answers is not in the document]
kermieisinthehouse: Good eye! fixed

ui/v2.5/src/locales/en-GB.json
kermieisinthehouse: Reordered these

kermieisinthehouse:
Fixed an issue where access from 127.0.0.1 was blocked.

Here's my testcases for this PR, I used curl to simulate the proxy headers.
```
# access from local IP
2021/09/28 19:27:17 "GET http://192.168.1.119:9999/ HTTP/1.1" from 192.168.1.5:41988 - 200 3124B in 48.22µs
# access from CGNAT IP
2021/09/28 19:29:05 "GET http://100.72.140.103:9999/ HTTP/1.1" from 100.68.156.57:40400 - 200 3124B in 34.715µs
# access from loopback
2021/09/28 19:33:16 "GET http://127.0.0.1:9999/ HTTP/1.1" from 127.0.0.1:52602 - 200 3124B in 43.811µs
# access from "localhost"
2021/09/28 19:38:14 "GET http://localhost:9999/ HTTP/1.1" from 127.0.0.1:52604 - 200 3124B in 68.538µs
#curl --header "X-FORWARDED-FOR: 192.168.1.1" 192.168.1.119:9999
2021/09/28 19:35:30 "GET http://192.168.1.119:9999/ HTTP/1.1" from 192.168.1.119:33302 - 200 3124B in 44.153µs
#curl --header "X-FORWARDED-FOR: 192.168.1.1, 192.168.1.2" 192.168.1.119:9999
2021/09/28 19:36:57 "GET http://192.168.1.119:9999/ HTTP/1.1" from 192.168.1.119:33304 - 200 3124B in 37.39µs
#curl --header "X-FORWARDED-FOR: 8.8.8.8" 192.168.1.119:9999
Trips the wire!
#curl --header "X-FORWARDED-FOR: MALFORMED" 192.168.1.119:9999
Trips the wire!
#curl --header "X-FORWARDED-FOR: 8.8.8.8, 192.168.1.1" 192.168.1.119
Trips the wire!
```

WithoutPants:
I've done a bit of refactoring on the new code:
- moved the authentication handler code into a separate file in `api`
- moved the external access checking logic into the `session` package
- changed the `trustedProxies` setting to accept a list of strings, rather than a comma-separated string
- changed the `security_tripwire_accessed_from_public_internet` to store the IP address that triggered the tripwire. This IP address is printed as part of the log message when the tripwire is activated
- added a check for the tripwire activation at startup. If the tripwire is active, then it logs the same message as when it was originally tripped

I have updated the wiki page to be consistent with the new behaviour.

kermieisinthehouse (commented):
New changes staticly reviewed reviewed but not tested, approved.

WithoutPants (approved):

bradydjohnson:
Where in the config.yml file should the `security_tripwire_accessed_from_public_internet` line go? Does anyone have an example config.yml file they can share?

I tried to access my stash remotely using Tailscale as recommended [here](https://github.com/stashapp/stash/wiki/Authentication-Required-When-Accessing-Stash-From-the-Internet), but it didn't work.  Then I think I removed `security_tripwire_accessed_from_public_internet` and now I am unable to access my Stash even on my home private network.

PR #1764, opened 2021-09-25
Fix video player sizing

InfiniteTF:
By default `object-fit` is set to `fit` which cuts off the sides to avoid black bars.


kermieisinthehouse:
This isn't affected by this change, but I noticed while testing that very small black bars appear on the top and the bottom of the scene player on mobile for any video ratio, both on Chrome and Firefox for Android. If we're fixing black bars, could you take a look to see if this is a quick fix?

WithoutPants:
FYI - I was only able to reproduce the original bug on a fresh chrome instance with no extensions. Clicking play would set the style to include `object-fit: fit`. Clicking on one of the scene tabs would cause the style to be cleared. See also https://github.com/micnews/react-jw-player/issues/94

WithoutPants (approved):

PR #1765, opened 2021-09-25
Fix list state navigation

### LINKED ISSUE #1762: [Bug Report] Sort Dir is reset when dynamic filter
**Describe the bug**
When you click on a Perf/Tag/Movie during Grid scene, the filter is correctly apply but sortdir is set ASC.

Default view:
`http://localhost:9999/scenes?perPage=20&sortby=created_at&sortdir=desc`
When you apply a tag:
`http://localhost:9999/scenes?c={"type":"tags","value":{"items":[{"id":"24","label":"Big%20Tits"}],"depth":0},"modifier":"INCLUDES_ALL"}&perPage=20&sortby=created_at`

The url forget the `sortdir`.

**To Reproduce**
1. Go to Scenes (maybe you need to set a Default Saved Filter with `sortdir=desc`)
2. Click on Tag/Performer within a card
3. See error, `sortdir` change to ASC

**Expected behavior**
Don't change the current direction.

<html><body>

Version: | v0.9.0-52-gb14d5c56
-- | --
Build hash: | b14d5c56
Build time: | 2021-09-24 02:02:53

</body>
</html>

 - OS: Win10
 - Browser: Firefox 92.0

**Additional thing**
When you applied the filter and press Back on your browser, the filter is still apply. I think the good behavior would be to remove the filter. Maybe it's worth a other issue ?

gitgiggety:
This fixes two bugs with lists:
* When order is set to DESC and clicking a tag or performer in the popover the sort column is maintained, but the sort order is lost
* When using the back button to a URL where no criterion is applied the criterion(s) aren't removed, in other words the criterion(s) are still applied
* Show the first page when clicking on tag / performer / ... while on a different page

Fixes #1762

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v0100.md
+++ b/ui/v2.5/src/components/Changelog/versions/v0100.md
@@ -19,6 +19,8 @@
 * Added sv-SE language option. ([#1691](https://github.com/stashapp/stash/pull/1691))
 
 ### 🐛 Bug fixes
+* Fix criteria being incorrectly applied when clicking back button. ([#1765](https://github.com/stashapp/stash/pull/1765))
+* Show first page and fix order direction not being maintained when clicking on card popover button. ([#1765](https://github.com/stashapp/stash/pull/1765))
 * Fix panic in autotagger when backslash character present in tag/performer/studio name. ([#1753](https://github.com/stashapp/stash/pull/1753))
 * Fix Scene Player CLS issue ([#1739](https://github.com/stashapp/stash/pull/1739))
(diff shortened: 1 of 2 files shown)


gitgiggety:
@Belleyy can you confirm this fixes the issues you reported? (And test it doesn't introduce any new issues).

Belleyy:
It work great! Don't find any bug.

gitgiggety:
It's fixed as well now :)

It's all part of the same method in the code and thus more or less the same logic. So not worth creating a pull request for every change.

WithoutPants (approved):

PR #1771, opened 2021-09-25
[WIP] Scan checksum concurrency fix

bnkai:
Marked as WIP as i can't force the below issue during scanning even using one 1k copies of an image and a lot of threads.

When using a lot of workers (parallel tasks) and a lot of files (images is easier to check) that are duplicated in `close` paths then you could have multiple workers competing for the same entry and failing due to the checksum's unique constraint.

A checksum manager is used to only allow a single active worker with a given checksum.
Any following workers with the same checksum will be blocked till the previous one(with the same checksum) is finished.
Instead of blocking the worker we could just return a duplicate message and quit the worker but if we want to allow/track duplicates in the database in the future the way it is implemented  is more compatible.

From my test performance doesnt seem to be affected even when using a lot of copies and parallel jobs (not easy to happen in real conditions)


SmallCoccinelle (commented):
I think my solution would be to store a `map[checksum]chan struct{}{}` inside a 1-element channel. The map is a token you pull out to operate on the map. If there is no checksum there, you store a channel `done` in the map and put it back. Once you are done, you close the channel.

If there's already an entry, you grab the `done` channel and put the map back. Once the channel is closed, you "race" to grab the map and put your entry in.

This avoids a lot of the communication channels, and it might be cleaner code-wise, but I haven't really tried it.

Issue #1775, opened 2021-09-27
[Feature] Show stats / totals for list results


kermieisinthehouse:
List results pages should include details about results, such as duration and filesize (for scenes), and megapixels and filesize (for images). This would ideally be under the pagination, and the pagination should be visible both on top and below the list results. 

Examples (obviously the number rounding is incorrect)

![Screenshot_20210926-171511_1](https://user-images.githubusercontent.com/84814418/134829532-f44fdaf2-7e74-40ef-934f-03f9e4ab01d5.jpg)
![Screenshot_20210926-172050_1](https://user-images.githubusercontent.com/84814418/134829535-95ace9a8-6424-4b1f-b822-8de9fdd212bd.jpg)

The formula for megapixels is (height*width/1_000_000). Also, the Query() function needs some refactoring for performance, so that it returns a struct of stats that are specifically requested instead of running all of them.

I have a half-baked [branch](https://github.com/kermieisinthehouse/stash/tree/show-duration-and-filesize-in-results) if it's useful for some reference, but I can't finish it, it has some issues I'm happy to discuss if it matters.

WithoutPants:
Bounty placed for $35

ALonelyJuicebox:
Can I just ask that this data get its own ID (throw it in a <span> tag or something) so that we can customize how its viewed using CSS?

Belleyy:
> 
> 
> Can I just ask that this data get its own ID [...] Not everyone may want this.

Can confirm, i would personally like having the filesize but not the duration.

PR #1776, opened 2021-09-27
Show duration and filesize in results

kermieisinthehouse:
Re: https://github.com/stashapp/stash/issues/1775

This PR has some issues. Off the top of my head:

1) the frontend stuff can be ignored. It's a mess and I'm sure I'm doing stuff wrong
2) the scenes query doesn't work if it requires a join (probably simple)
3) the images query is untested and doesn't work
4) For performance, Query () may need to be refactored into something along the lines of:
```
type SceneQueryStatsRequest struct {
    count bool
    durationSeconds bool
    filesizeBytes bool
}

type SceneQueryStats struct {
    count int
    durationSeconds float64
    filesizeBytes int
}

func Query(query queryBuilder ..., statsToReturn SceneQueryStatsRequested) (scenes ..., stats SceneQueryStats) {
...
}
```

So that it will conditionally run the queries.


pkg/models/image.go
```diff
@@ -16,7 +16,7 @@ type ImageReader interface {
 	// CountByStudioID(studioID int) (int, error)
 	// CountByTagID(tagID int) (int, error)
 	All() ([]*Image, error)
-	Query(imageFilter *ImageFilterType, findFilter *FindFilterType) ([]*Image, int, error)
+	Query(imageFilter *ImageFilterType, findFilter *FindFilterType) ([]*Image, int, float64, int, error)
```
WithoutPants, line 19: I think that we're going to need to refactor how we do this interface. Returning this many values is ugly. 

I'm guessing we probably want a `ImageQueryResult` struct that returns the image objects and other stuff. I also think that we should probably accept a query options object so that we don't always have to use potentially slower queries.

It also might be a good opportunity to return ids only and have the object-resolving code pushed out to the caller.

PR #1782, opened 2021-09-27
Allow setting metadata directory from UI

WithoutPants:
Currently not able to set the `metadata` directory config setting from the UI, which means the user needs to manually edit `config.yml` before they can perform a full import/export. Added the ability to set `metadata` directory in the UI.


bnkai:
Seems ok.
A minor issue, in new setups the default value for metadata is empty ( the cache is the same but i dont think its used anywhere anymore?)

while the generated and db options are populated with the default.

WithoutPants:
That's expected basically, and unrelated to this change. The metadata directory is an optional thing, so doesn't need to be set a setup.

PR #1783, opened 2021-09-27
Remove new-from-rev from the linter

SmallCoccinelle:
With the linter in place, we can now "backflush" the project for the older errors we didn't fix before enabling the linter. This small patch stack does that, and cleans up the remaining revive and errcheck warnings in the project.

Once done, the `new-from-rev` setting is removed.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,10 +3,23 @@ run:
   timeout: 3m
   modules-download-mode: vendor
 
-issues:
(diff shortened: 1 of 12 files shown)


SmallCoccinelle:
This isn't ready for merge. It needs #1784 as it fails local tests if you run `make validate` (which it should, it's wrong)

SmallCoccinelle:
Should be all good now.

PR #1785, opened 2021-09-27
Support Is (not) null for all multi criterions

### LINKED ISSUE #1398: [Feature] Add NULL/Not NULL to Studios filter
The URL filter has a option for "Is NULL" "Not NULL". I would really like this added to the Studios filter as well. Thank you!

gitgiggety:
Add support for the Is null and Is not null modifiers for all cases of the MultiCriterionInput and HierarchicalMultiCriterionInput. This
partially overlaps the "X Count" filter which sometimes is available (because it would be the same as "X Count equals 0" and "X Count greater than 0") but this also enables it for other criterions like the "Parent Studio" filter for studios or just the "Studios" filter for scenes / images / galleries, the "Movies" filter for scenes etc.

Fixes #1398


WithoutPants:
This needs an integration test suite.

I started writing some unit tests and found a failure when filtering on galleries -> performer tags. Is NULL still returns galleries with performers with tags.

This is the code that I added to discover the failure in `TestGalleryQueryPerformerTags`:

```
		tagCriterion = models.HierarchicalMultiCriterionInput{
			Modifier: models.CriterionModifierIsNull,
		}
		q = getGalleryStringValue(galleryIdx1WithImage, titleField)

		galleries = queryGallery(t, sqb, &galleryFilter, &findFilter)
		assert.Len(t, galleries, 1)
		assert.Equal(t, galleryIDs[galleryIdx1WithImage], galleries[0].ID)

		tagCriterion.Modifier = models.CriterionModifierNotNull
		q = getGalleryStringValue(galleryIdxWithPerformerTag, titleField)

		galleries = queryGallery(t, sqb, &galleryFilter, &findFilter)
		assert.Len(t, galleries, 1)
		assert.Equal(t, galleryIDs[galleryIdxWithPerformerTag], galleries[0].ID)
```

kermieisinthehouse:
This doesn't seem correct. The "sub tags is null" filter returns tags that are parents, for example

WithoutPants:
Should now be fixed. 

I noticed non-obvious behaviour for the `performer tags` criterion: for `is NULL`, it will return any object that has a performer without tags - ie scenes with performers with performer tags will be returned. An object will be excluded only if _all_ performers have tags. `not NULL` returns only objects where a single performer has tags.

Probably ok with leaving that as a known issue and fixing it later.

PR #1793, opened 2021-10-02
Fix Studios -> Stash ID filter

bnkai:


WithoutPants (approved):
Introduced in dev, so leaving out of changlog.

PR #1796, opened 2021-10-03
Errorlint sweep

SmallCoccinelle:
This patch stack provides the necessary rewrites to enable the "errorlint" linter. This linter mostly centers around Go 1.13 style errors and wrapping them. In particular, rewrites prefers the use of `errors.Is(..)` and `errors.As(..)` to check for errors, such that you can wrap errors and still check their presence.

The patch stack also rewrites `err.Error()` into `err` in a large set of formatting strings: `fmt` is aware of errors and can format them with `%v`. Furthermore, if one wants to change into a wrapped error, `fmt.Errorf("...%w", err)` wraps the underlying error, so it can be used as part of the API. The rewrite sets us up for such things in the future.

Remove the `goconst` linter from the set of "interesting" linters since it has a very large false positive rate.

Tell `rowserrcheck` that `moiron/sqlx` is a SQL package as well, so it will check into that.

Each type of change gets it's own patch in the stack, so we can easily drop them if need be.

Patch has seen *some* testing on my end, but haven't run it for too long.


kermieisinthehouse (approved):

WithoutPants:
This is causing a panic when starting a new system.

```
panic: error initializing configuration: Config File "config" Not Found in "[D:\\dev\\stash_data\\fresh C:\\Users\\<snip>\\.stash]"

goroutine 1 [running]:
github.com/stashapp/stash/pkg/manager.Initialize.func1()
        D:/dev/stash/pkg/manager/manager.go:64 +0x5cb
sync.(*Once).doSlow(0xc000072000, 0xfd3aa5)
        C:/Program Files/Go/src/sync/once.go:68 +0xd2
sync.(*Once).Do(...)
        C:/Program Files/Go/src/sync/once.go:59
github.com/stashapp/stash/pkg/manager.Initialize()
        D:/dev/stash/pkg/manager/manager.go:60 +0x31
main.main()
        D:/dev/stash/main.go:26 +0x2e
```

SmallCoccinelle:
Oh!

I'll try to figure out what's wrong. Probably the error handling where I have something to learn.

SmallCoccinelle:
It's because the `viper` subsystem  returns a non-pointer version of the error rather than a pointer, whereas most of errors of this kind is passed by address such as `&PathError{...}`. I'll push a fix when I'm not too tired.

SmallCoccinelle:
This should fix all of the `errors.As(..)` problems by reflecting if the error is passed by (pointer)reference or not. While here, rebase to current `develop` and handle the type switch in the new authentication module.

PR #1801, opened 2021-10-04
Make generate args optional

WithoutPants:
This specifically makes `markerScreenshots` optional in `GenerateMetadataInput` to maintain backward compatibility with external clients. It also incidentally makes the other boolean options optional so that clients don't need to include irrelevant options.

### DIFF
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -1,15 +1,15 @@
 scalar Upload
 
 input GenerateMetadataInput {
-  sprites: Boolean!
-  previews: Boolean!
-  imagePreviews: Boolean!
+  sprites: Boolean
+  previews: Boolean
+  imagePreviews: Boolean
   previewOptions: GeneratePreviewOptionsInput
-  markers: Boolean!
-  markerImagePreviews: Boolean!
-  markerScreenshots: Boolean!
-  transcodes: Boolean!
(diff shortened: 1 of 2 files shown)


graphql/schema/types/metadata.graphql
```diff
@@ -1,15 +1,15 @@
 scalar Upload
 
 input GenerateMetadataInput {
-  sprites: Boolean!
-  previews: Boolean!
-  imagePreviews: Boolean!
+  sprites: Boolean
```
SmallCoccinelle, line 4: You could also go with something like:

```
    sprites: Boolean! = False
```

This makes it optional on the client side, because there's a default value the server uses. And it avoids the trouble of having to check for nullability on the server side, while maintaining backwards compatibility.

The only problem with this approach is that the front-end clients need to understand this schema construction, and unfortunately, not all of them do.

SmallCoccinelle:
I had a patch which went the way of the comment. More as a FYI of what is involved:

https://github.com/SmallCoccinelle/stash/commit/93df004c2f384017cb472a8388b525bda58cf33b for `ScanMetadataInput`. The problem is that you need to provide the types in the typescript code, since the generator isn't clever enough to figure out the fields are optional when there is a default value.

WithoutPants:
Using `false` instead of `False` seems to work. Thanks!

PR #1805, opened 2021-10-05
Refactor clean task

WithoutPants:
The existing clean task code found _all_ scenes, images and galleries. This caused a massive amount of memory usage. The clean task has been refactored to perform clean assessments in query batches of 1000. This means that (unfortunately) items must be removed after assessing all of the items of that type, otherwise the batch query will be impacted.


pkg/manager/task_clean.go
```diff
@@ -2,147 +2,367 @@ package manager
...
-	fileNamingAlgorithm models.HashAlgorithm
+type cleanJob struct {
+	txnManager models.TransactionManager
+	input      models.CleanMetadataInput
+	scanSubs   *subscriptionManager
 }
 
-func (t *CleanTask) Start(wg *sync.WaitGroup, dryRun bool) {
-	defer wg.Done()
+func (j *cleanJob) Execute(ctx context.Context, progress *job.Progress) {
```
bnkai, line 24: A
```golang
        logger.Infof("Starting cleaning of tracked files")
        if j.input.DryRun {
                logger.Infof("Running in Dry Mode")
        }
```
should be added somewhere here ?  (it was removed due to the refactoring but not re-added)

bnkai:
Code looks ok.
It seems to work ok also from a few quick tests.

A further optimization (out of scope of the PR) would be to not use the Query function.
While the Query function is ok for general usage, imho for maintenance or batch tasks that are very specific it may be too generic.
For example a simple `SELECT path,id from scenes LIMIT 1000 OFFSET ?` query would be a lot faster and lighter in this case.
For clean tasks we only need the `id` and `path` from every file. Using the Query function means that we get all data (and with a lot more db requests), which for the scenes (long descriptions) could be a considerable amount of bytes per batch.

PR #1807, opened 2021-10-05
Update zh-tw Localization & Fix i18n Grammar Issue

Still34:
## Summary

This PR is your usual periodic zh-tw localization update. 

## Issues Before Merging

Before this can be merged, however, one issue needs to be addressed. `Parent of`/`Sub-tag of` was added in one of the merges recently. As previously mentioned in one of the localization updates before, this type of string can be problematic for some languages. The current approach in the implementation of these string is as follows,

```tsx

  function maybeRenderChildren() {
    if (tag.children.length > 1) {
      return (
        <div className="tag-sub-tags">
          <FormattedMessage id="parent_of" />
          &nbsp;
          <Link to={NavUtils.makeChildTagsUrl(tag)}>
            {tag.children.length}&nbsp;
            <FormattedMessage
              id="countables.tags"
              values={{ count: tag.children.length }}
            />
          </Link>
        </div>
      );
    }
```

Ideally, this needs to be re-written as something that would allow the translator to move the tag name around in the localization string, e.g., `Parent of <x>`/`<x> 的母標籤`.

### DIFF
--- a/ui/v2.5/src/locales/zh-TW.json
+++ b/ui/v2.5/src/locales/zh-TW.json
@@ -192,7 +192,9 @@
         "password_desc": "使用 Stash 時所需的密碼，留空以關閉身份驗證",
         "stash-box_integration": "整合 Stash-box",
         "username": "用戶名",
-        "username_desc": "使用 Stash 時所需的用戶名，留空以關閉身份驗證"
+        "username_desc": "使用 Stash 時所需的用戶名，留空以關閉身份驗證",
+        "trusted_proxies": "已信任代理伺服器",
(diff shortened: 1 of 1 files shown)


WithoutPants:
Ok. How about we change `parent_of` to be `Parent of {children}` (in `en-GB`), then `children` will be `Link` element. Will that work?

WithoutPants:
I've added `part_of`, and changed `parent_of` and `sub_tag_of` to include arguments.

Still34:
Alright thanks I'll take a look at the rest when available later tonight.

PR #1815, opened 2021-10-06
Fix exclude filter query performance

kermieisinthehouse:
Use IN instead of EXISTS to dramatically increase query performance in exclude filters.

Excluding a performer would hit a 2 minute timeout before, now loads in sub-second time.


kermieisinthehouse:
Should we sneak this one into the 0.10 release? It's a performance fix with no behavior change.

You can see the speedup by loading the scenes list with "filter > performer > exclude > random performer"

PR #1816, opened 2021-10-07
Scan refactor

WithoutPants:
- create new `ffmpeg.FFProbe` type and replace `manager.FFProbePath` with an instance of it. `NewVideoFile` is now a function on `FFProbe` and no longer needs the ffprobe path.
- replace `manager.FFMPEGPath` with an instance of `ffmpeg.Encoder` and using this instead of creating a new encoder object each time.
- move ffmpeg image thumbnail generation code into `ffmpeg.Encoder`.
- refactor vips image thumbnail generation code to be similar to ffmpeg generation
- create a `file` package for common operations for file scanning.
- move zip related code into `file` package.
- add a `models.File` type to consolidate file metadata into a single struct. This will form the basis of the file tracking in a later iteration.
- move scanning code into the specific packages (`scene`, `gallery` and `image`).
- refactor scanning code to use `file.Scanner` which performs the hashing and file modtime checking.
- moved remaining scanning code in `manager` into separate files based on type.
- reimplements #1771 to use a new `utils.MutexManager` type. Implementation is based on @SmallCoccinelle's comments, and imo is a far simpler way of solving the issue.

This refactor is a necessary step to shift to storing file information for media in the database. I don't imagine there will be significant performance improvements, but it should demonstrate some slight improvements.

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v0110.md
+++ b/ui/v2.5/src/components/Changelog/versions/v0110.md
@@ -3,6 +3,9 @@
 * Added built-in `Auto Tag` scene scraper to match performers, studio and tags from filename - using AutoTag logic. ([#1817](https://github.com/stashapp/stash/pull/1817))
 * Added interface options to disable creating performers/studios/tags from dropdown selectors. ([#1814](https://github.com/stashapp/stash/pull/1814))
 
+### 🎨 Improvements
+* Optimised scanning process. ([#1816](https://github.com/stashapp/stash/pull/1816))
+
 ### 🐛 Bug fixes
 * Sort directory listings using case sensitive collation. ([#1823](https://github.com/stashapp/stash/pull/1823))
 * Fix auto-tag logic for names which have single-letter words. ([#1817](https://github.com/stashapp/stash/pull/1817))


WithoutPants:
> Can we take this opportunity to remove the 90 second file count delay? On any large instance it's just a 90 second spinlock.

I was going to do this in a separate PR, but it touches on affected code, so I've implemented it here.

The scan process is refactored so that it counts files _and_sends the file details to a channel that the scan tasks now draw from and process. A happy side-effect of this is that the files are potentially touched one fewer times. I've set the channel size to 200,000. From my testing, this seems like a decent buffer without blowing out memory too much. The buffer size affects at what proportion the counting function will return. Where the total files is under 200,000, it will be unaffected. Where the total files is 400,000, the earliest that the progress meter will be shown is at 50%. For 1m files, it won't show up until the final 80%.

I've set the number of counting threads to be the same number as the scan threads. This would ideally prevent the scan threads from starving. In my testing however, if there is nothing for the scan threads to actually do (no changes), the scan threads starve any way and the progress meter doesn't show up at all. This is actually good since it means that previously, it would have taken as long to count (up to the cutoff duration) as it did to scan.

This pattern should be applied in future to the generate task as well.

One other very noticeable change is that the progress throttling has been changed from once per second to once every 100ms. This results in a far more responsive looking job manager when jobs are being processed quickly. It increase the amount of bytes being fed into the websocket - from ~400 bytes per second to potentially ~3.9 kib/s.

PR #1817, opened 2021-10-07
Autotag scraper

WithoutPants:
- Moved autotag matching logic, and scraped object matching logic into the new common package `match`.
- Refactored the scraper code a little to make it easier to add the new scraper implementation.
- Added new built-in scraper `Auto Tag`. This scraper finds performers, studios and tags only and is supported for scene and gallery scrapes. It operates using the same logic as the auto-tagger.


kermieisinthehouse:
✓ Code reviewed
✓ Tested scraping auto-tag from scene
✓ Tested auto-tag a performer
✓ Tested auto-tag task, and selective auto-tag

Auto-tagging just one scene / image is a nice touch!

PR #1823, opened 2021-10-10
Add collation to directory listings

### LINKED ISSUE #1806: [Bug Report] Directory listing case sensitive alphabetical order
**Describe the bug**
When I selective scan, the directory listing displays the directories in a case sensitive alphabetical order.

**Current behavior**
The directory listing displays:
* Asomething
* Bsomething
* Csomething
* afoo
* bbar
* ctest

**Expected behavior**
* afoo
* Asomething
* Bbar
* Bsomething
* Csomething
* ctest

**Stash Version: (from Settings -> About):** pre-release 0.10.0, build hash a5ca8fc6

SmallCoccinelle:
## Commentary

Patch message at the end.

#### Lifting language detection:

We can write something like

```
func fromRequest(r *http.Request) *collate.Collator {
	lang, _ := r.Cookie("lang")
	accept := r.Header.Get("Accept-Language")
	tag, _ := language.MatchStrings(matcher, lang.String(), accept)

	return collate.New(tag)
}
```

and pull this into a more general language tagging, where we then use the `locale` as an override if given. Basically, we use cookies and/or `Accept-Language` headers to control the locale for resolvers in general with this solution. Often the tag is stored in the resolvers context.

#### The field-arg comments in GraphQL

Our `/playground` system doesn't do the correct documentation, and doesn't pull arg docstrings into the Docs interface. But it's being parsed correctly by the backend. Witness by the following graphql query:

```
query Q {
  __schema {
    types {
      name
      fields {
        name
        args {
          name
          description
          defaultValue
        }
      }
    }
  }
}
```

And look for the `directory` field on `Query`. I've added the docstring anyway, because the `/playground` is likely to catch up at some point.

## Patch message
Introduce a new `locale` arg to the `Query.directory` field. Set "en"
as the default for the field for backward compatibility. Use the given
locale, sending it through a language matcher, and use `x/text` as the
collation engine for the matched language.

Augment the file `ListDirs` call to optionally take a Collator. If the
Collator is given, sort file listings according to the collators rules.

While here, document the GraphQL schema a bit more.

Add matchers by looking at the current front-end locales, and make sure
each of these occur in the matcher list.

Resolves #1806

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v0110.md
+++ b/ui/v2.5/src/components/Changelog/versions/v0110.md
@@ -4,5 +4,6 @@
 * Added interface options to disable creating performers/studios/tags from dropdown selectors. ([#1814](https://github.com/stashapp/stash/pull/1814))
 
 ### 🐛 Bug fixes
+* Sort directory listings using case sensitive collation. ([#1823](https://github.com/stashapp/stash/pull/1823))
 * Fix auto-tag logic for names which have single-letter words. ([#1817](https://github.com/stashapp/stash/pull/1817))
 * Fix huge memory usage spike during clean task. ([#1805](https://github.com/stashapp/stash/pull/1805))


SmallCoccinelle:
Binary file size increase: ~47 -> ~48 Megabyte.

kermieisinthehouse:
I have a hard time reconciling 80k lines of code and 1MB of binary size with just alphabetizing file paths. Surely we can just limit this to utf-8 or something? Or perhaps do a case-insensitive sort to resolve the original issue?

PR #1832, opened 2021-10-12
Start browser on server start

kermieisinthehouse:
This will automatically start the user's default browser on any platform, opened to stash except when running under docker, as root, or when a user has passed `--nobrowser`.

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v0110.md
+++ b/ui/v2.5/src/components/Changelog/versions/v0110.md
@@ -1,4 +1,5 @@
 ### ✨ New Features
+* Optionally open browser on startup (enabled by default for new systems). ([#1832](https://github.com/stashapp/stash/pull/1832))
 * Support setting defaults for Delete File and Delete Generated Files in the Interface Settings. ([#1852](https://github.com/stashapp/stash/pull/1852))
 * Added Identify task to automatically identify scenes from stash-box/scraper sources. See manual entry for details. ([#1839](https://github.com/stashapp/stash/pull/1839))
 * Added support for matching scenes using perceptual hashes when querying stash-box. ([#1858](https://github.com/stashapp/stash/pull/1858))


WithoutPants:
I personally hate this idea, but I understand why it's desirable. If we're to add it, there needs to be a config setting to override it so the user doesn't need to run `--nobrowser` every time they invoke it.

kermieisinthehouse:
I disagree actually - I think this is useful to the majority of users, which is why I made it an opt-out. I suspect that the prime stash use-case is starting it up "as needed" to reference, or just manually double clicking an executable in your downloads folder after you log in-- situations in which a browser open would be useful. 

I also suspect the people who have set up service files or docker (which are excluded automatically from running) are much more likely to be able to investigate and turn it off. Either way, I'll defer to your decision

WithoutPants:
Just to clarify - this is about the _config_ key, not the executable flag.

For new users and new systems, it absolutely makes sense to open the browser when they first start it. It will reduce confusion for initial users, and will make it a more app-like experience.

It is for existing systems that I am concerned about.

Imagine a scenario where someone has set up stash to run at startup on their desktop machine. I imagine that they'd be somewhat surprised and likely a bit miffed if a browser unexpectedly appeared with their porn collection when they start their machine.

In my opinion, it's bad practice change the startup behaviour like this for existing users. Existing users are accustomed to the existing behaviour, and can easily change this behaviour if they wish. Performing this behaviour by default for existing users not only submits them to this new behaviour, but puts the onus on them to _fix_ it, which I don't think is good UX design.

Ultimately, this is what I think is the appropriate behaviour, but I'm happy to discuss further:
- `GetNoBrowserFlag` should set the default value to true if there is no existing value
- when there is no config - ie it is a new system - explicitly set the `nobrowser` flag to true and open the browser

This means that systems created _after_ this change will open the browser when stash is started. For systems _before_ this change, the existing behaviour is maintained, and the user can opt-in as needed.

I think also that the way that the configuration interacts with the flags needs to be looked at as well. From memory, when a flag is provided, it persists that flag in the configuration, which we don't want it to do.

PR #1833, opened 2021-10-12
Docker+Build optimizations

kermieisinthehouse:
Assorted changes to our builds:
- Switches the release docker image to Alpine Linux, and optimizes the build quite a bit. `docker pull`s should be half the size, and merges to `develop` should shave 5-8 minutes off the waiting time. Supports all of our architectures, and I've been running stash in an alpine image for > 1 month already, no issues.

- Added a concept of 'Official builds' - tests whether or not a specific binary was produced by our CI pipeline. This is not a security feature, the use-case is for informing a future update notification / auto-update mechanism if it should run.
**This also requires a compiler image update+push**

- Pass the git-hash directly to docker/build images instead of the entire .git folder, to reduce build cache misses considerably. Always build with latest ffmpeg.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,17 +52,12 @@ jobs:
         path: .go-cache
         key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/go.sum') }}
 
-    - name: Start build container (PR Preview Build)
-      if: ${{ github.event_name == 'pull_request' }}
+    - name: Start build container
+      env:
+        official-build: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/develop') || (github.event_name == 'release' && github.ref != 'refs/tags/latest_develop') }}
       run: |
         mkdir -p .go-cache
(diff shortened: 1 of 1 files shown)


kermieisinthehouse:
Updated with fixes and added binary compression to the builds.

WithoutPants:
Just had a thought: would it not be easier to move the official build stuff into the Github action rather than embedding it in the compiler image?

kermieisinthehouse:
> Just had a thought: would it not be easier to move the official build stuff into the Github action rather than embedding it in the compiler image?

If you can find a mechanism for it. The actual build happens in the compiler so any build args have to be present there. We can pass build args into the compiler container from the runner, but we still need to push a new image with ARG statements for it to take effect. It's easier to just declare it in there.

WithoutPants:
So the current implementation only validates that the binaries were created using our stash compiler image. I think it would make more sense to set `OFFICIAL_BUILD` to true in `build.yml` when building a develop or master release (ie the same logic as the `Development Release` and `Master release` actions). The environment variable can be passed into the `cross-compile` actions using `-e`. As such, modifying the stash compiler image is unnecessary.

kermieisinthehouse:
Updated, compiler image rebuild no longer necessary

WithoutPants:
Shouldn't the environment variable only be set when building a dev or master release? As it is, I think it's going to mark PR artifacts as official releases.

kermieisinthehouse:
Updated, should be all set. I assumed all builds will would suffice- but you're right, we don't want a PR preview to auto-update.

PR #1839, opened 2021-10-13
Identify task

### LINKED ISSUE #1353: How to scrape multiple videos
#### TL;DR; In short: How to avoid having to paste the URL for each video in order to scrape a gazillion videos?

I want each video in my library to have information pulled from the appropriate site using a community scraper.

As far as I can tell, for each video I have click on the video -> Click edit -> Paste the URL for for the video -> click on the scrape button -> Choose which tags I want to add -> Save.

This would take a tremendous amount of time to do for, say a hundred videos.

Surely there must be a better way? I have read the docs, but I could not find anything that would let me scrape multiple videos automatically.

### LINKED ISSUE #1589: [Feature] Bulk/Automate Scrap from stashdb or other scraper
Searching the scene and Scraping metadata for each scene one by one

To automate the process by scraping next one after other.
Like the scan process but for scraping from stashdb or other scraper

None

**Additional context**
None

WithoutPants:
As discussed in #1802.

Resolves #1353 (alongside #1812).
Resolves #1589.

Supercedes #1522. 

Adds an Identify task to the Tasks page, grouped with Scan and Auto Tag in the Library section:

image

Opens a dialog as follows:

image

By default, only the first stash-box is added as a source. Other sources can be added, but must either be a stash-box instance or a scene fragment scraper. 

Allows setting default options which can be overridden in the individual sources. Includes field options:

image

This allows specific field behaviour to be changed. Strategy can be "ignore", "merge", or "overwrite". For single value fieds, merge ~~is functionally the same as Overwrite~~ only sets if no value is already set on the field. For multivalue fields (performers, tags), it means that existing values are kept while adding to them. For performers, tags and studios, a create missing option is available that creates any missing objects as needed. This defaults to false.

It is possible to set the defaults for the task. This pattern will be applied to the other tasks, alongside a general refactor of the Tasks pages, in another PR.

The Identify task itself tries to identify the scene starting from the first source. If it finds a result, then it sets the scene fields in accordance with the field settings (preferring source-specific settings, then defaulting to the global values. Where a scene field isn't specified, its strategy is set to "merge" and not to create missing objects.

The other sources are only checked if the previous ones had no results.

The task logs if a scene was identified or not.

This PR still has some pending things to do, but it's worth giving it a test as it is.

TODO:
- [x] identify from scene selection
- [x] support Identify using folder selection
- [x] documentation
- [x] unit tests

Future iterations:
- add identify task as scan option
- add filename filtering to scraper configs
- add some sort of throttling for scraper requests

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v0110.md
+++ b/ui/v2.5/src/components/Changelog/versions/v0110.md
@@ -1,4 +1,5 @@
 ### ✨ New Features
+* Added Identify task to automatically identify scenes from stash-box/scraper sources. See manual entry for details. ([#1839](https://github.com/stashapp/stash/pull/1839))
 * Added support for matching scenes using perceptual hashes when querying stash-box. ([#1858](https://github.com/stashapp/stash/pull/1858))
 * Generalised Tagger view to support tagging using supported scene scrapers. ([#1812](https://github.com/stashapp/stash/pull/1812))
 * Added built-in `Auto Tag` scene scraper to match performers, studio and tags from filename - using AutoTag logic. ([#1817](https://github.com/stashapp/stash/pull/1817))


WithoutPants:
I did an extensive refactor of the backend code and moved it into a new `identify` package. The aim was to make it idiomatic and actually testable. It includes a high-coverage suite of unit tests, and I managed to squash some undetected bugs while creating them.

WithoutPants:
I've added the ability to select folders to identify, and added an Identify action to the overflow menu on the Scenes page.

This includes a minor improvement to the folder select control so that requests are debounced a little when typing in the folder text field.

Added an entry to the manual, and tried to make the description on the Identify dialog more clear.

This should now be ready for testing and review.

SmallCoccinelle:
Went over this PR as well. Managed to push the wrong button so I reviewed some outdated stuff as well. But I'm sure you can manage to find if it has moved elsewhere.

Issue #1744, opened 2021-09-19
[Feature] Move ratings select from 'edit' to main view


CosimoVestri:
**Is your feature request related to a problem? Please describe.**
At the moment, one must go to the edit page for changing item rating.

**Describe the solution you'd like**
Allow changing the rating without going to 'edit' and then saving the edit. I would like to see this as a changeable item from the main item page. Babepedia's implementation contains a direction that is phrased 'rollover to rate.' My request is for similar functionality. Click on a star to set rating to that number. Click on current star to remove rating. Without hover or click, current rating would be shown.

**Describe alternatives you've considered**
Marking this issue as 'wontfix'

**Additional context**
One additional side effect of the current implementation is that the 'updated at' is modified when rating is changed. I'd rather it not be affected by only a rating change. This affect would be similar to incrementing the O-Counter in nature.

oikmeg:
Fully agree with this one. Having to go to edit and click save to rate is not intuitive.

PR #1844, opened 2021-10-13
Add performer star ratings to main view

7dJx1qP:
~~Resolves~~ #1744 (only resolves for performer page)

image


kermieisinthehouse:
The stars are a little big compared to the surrounding elements.

Otherwise tested and reviewed

WithoutPants:
@CosimoVestri I'm an idiot and completely misunderstood both the original issue and this PR.

For the record, I thought this was about moving the stars on the cards in the query view. My apologies to you and to @7dJx1qP 

Feedback from a quick test: if we have rating at the top of the page, we don't need it in the details tab or the edit tab.

PR #1848, opened 2021-10-14
Enable gocritic

SmallCoccinelle:
#### Gocritic

General simplification rewrites, except for the `appendAssign` category, which we handle in the next patch stack. Also disabled `commentFormatting` since it's quite strict and the code base isn't really amenable for it. At least not yet.

* Avoid capitalizing local variables
* Use `re.MustCompile` for static regexes
* else-if block simplification
* Turn a lot of if-else chains into `switch` statements
* Use `strings.ReplaceAll` if applicable
* Use `+=`, `*=` and `++` for consistency

Finally, enable `gocritic` with these settings.

### DIFF
--- a/graphql/schema/types/scraper.graphql
+++ b/graphql/schema/types/scraper.graphql
@@ -175,10 +175,16 @@ type StashBoxFingerprint {
   duration: Int!
 }
 
+"""If neither performer_ids nor performer_names are set, tag all performers"""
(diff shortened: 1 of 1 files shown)


SmallCoccinelle:
Rebased to current develop. Handle all conflicts so it applies cleanly again.

PR #1851, opened 2021-10-14
Significantly speed up UI build by leveraging esbuild

kermieisinthehouse:
In my local testing, UI building time is down to 0:16 from 1:09.

Esbuild is a replacement for an entire javascript build chain, written in golang and much much faster than the traditional tools. It doesn't support every webpack feature we use, so this only leverages esbuild to replace babel and terser in our build, thanks to a craco framework (create-react-app-config-override) plugin designed just for that.

Also, resolved that annoying browserlist out of date error.

### DIFF
--- a/ui/v2.5/craco.config.js
+++ b/ui/v2.5/craco.config.js
@@ -1,5 +1,4 @@
 const CracoEsbuildPlugin = require('craco-esbuild');
-
 module.exports = {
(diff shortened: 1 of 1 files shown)


WithoutPants:
Should the `start` script be similarly changed?

WithoutPants:
Cool. I just tried it with `craco start` and it gave an error about the Lightbox element.

PR #1852, opened 2021-10-15
Add delete file and generated files by default config options

### LINKED ISSUE #1126: [Feature] Delete file as config option

7dJx1qP:
Resolves #1126 

Initial attempt at https://github.com/stashapp/stash/issues/1126 and https://github.com/stashapp/stash/pull/1424
Updates the delete scene, image, and gallery dialogs and adds settings to Interface. I'm sure these changes can be styled and worded better.
image
image

One of the comments mentions a permanent visual notification, but I haven't figured out how exactly it should look.


WithoutPants:
I would change `Images` to be `files` so that it's explicit that it's files that are being deleted, but otherwise that looks pretty good.

PR #1855, opened 2021-10-15
Cache and reuse the scraper HTTP client

### LINKED ISSUE #1850: [Bug Report] Request not being closed or fully read in scraper code
**Describe the bug**
(Reported from discord) Using ThePornDB scraper in the new tagging view exposed an existing bug:
Running a sufficiently large 'Scrape All' job will result in the error `dial tcp ... connect: cannot assign requested address`. This is local port exhaustion, probably because we are not releasing connections back into the connection pool. Somewhere, there is an http.Get() (or similar) call, where we are not doing request.body.Close() or io.ReadAll(request.body).

SmallCoccinelle:
This PR is an early draft for fixing #1850.

The core idea is to reuse an *http.Client. We put such a client onto the scraper cache since it's really a "cache" of alive HTTP requests, and as such naturally lives there. I tried threading the client through the code, but this quickly gets very unwieldy since the scraper code has lots of interfaces and also lots of wrapper structs. Instead, I opted to piggyback on how the txnManager is handled in the code base and use that as the guide on where to push the *http.Client. It's safe for concurrent use, so as long as you aren't operating on a stale one, everything is fine even when multiple goroutines use the same one.

To get handling of cookies, the rewrites into request-oriented HTTP client processing helps: we make the request, and then we add cookies to the request as a one-shot cookie. This retains the old behavior: the client and jar were thrown away anyway after each request, so we don't keep the cookies around for later requests.

We still create a cookie jar and populate it from the configuration. We then use it to extract the cookies for the request url. This also retains existing behavior, reuses the stdlib jar code, and executes roughly like the current code.

This isn't the full patch however, and it hasn't seen much testing by me yet:

* There are still references to one-shot *http.Client's in the code for setting images, among other things. We ought to handle those as well.
* It looks like the GraphQL client has the same behavior, and we should probably also handle that case, perhaps in a later patch.

Right now, I'm more interested in if this is the right direction. Or if you have a better idea for a solution. Keeping the clients as injected clients rather than globals tend to be a boon for testing: you can inject a mock client in tests.

Fixes #1850

### DIFF
--- a/pkg/scraper/stashbox/stash_box.go
+++ b/pkg/scraper/stashbox/stash_box.go
@@ -39,6 +39,10 @@ func NewClient(box models.StashBox, txnManager models.TransactionManager) *Clien
 	}
 }
 
+func (c Client) getHTTPClient() *http.Client {
+	return c.client.Client.Client
+}
+
 // QueryStashBoxScene queries stash-box for scenes using a query string.
 func (c Client) QueryStashBoxScene(ctx context.Context, queryStr string) ([]*models.ScrapedScene, error) {
 	scenes, err := c.client.SearchScene(ctx, queryStr)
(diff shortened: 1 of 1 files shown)


SmallCoccinelle:
I took care of the `getImage(..)` call inside the scraper code as well.

SmallCoccinelle:
This needs some more testing in the hands of others. I've tested it against a couple of sites myself, which requires cookies in the scrapers, but I haven't run it in bulk to check if the problem really goes away. I'm more in the area of: "I didn't break any new stuff with the patches."

bnkai:
Looks ok. 
Tested it mainly with theporndb scraper (json) in the new tagger view and with various sites in edit scene mode.
Seems to work ok with cookies and headers
I also couldn't reproduce the original issue so not sure about that :-(

PR #1859, opened 2021-10-16
Add phash and parent_id to "is missing" filter

7dJx1qP:
Adds phash to scene is missing filter option
Adds parent_id to studio is missing filter option


WithoutPants:
The Is Missing filter criterion is largely deprecated in favour of `is null`. It should only be used where it's not otherwise possible to filter using `is null`. In this instance, `phash is missing` is the same as `phash is null` and is therefore redundant.
 #1785 adds `is null` as an option to multi-criterion types, which should cover the `parent_studio` scenario.

PR #1860, opened 2021-10-16
Fix saving using performer tagger

peolic:
Fix save button not working on performer tagger view (reported on Discord).

https://user-images.githubusercontent.com/66393006/137599262-c4403c97-3337-4080-a4d4-6b7e04d03269.mp4

 
Issue is only on develop (#1812), no need for a change log entry.


peolic:
More issues:
> when saving the stashID is not updated
> just noticed this also
>

PR #1892, opened 2021-10-25
Fail if attempting to delete a file that is undeletable

kermieisinthehouse:
Currently, if attempting to delete a scene/image that is unable to be deleted (immutable bit set, read-only filesystem, insufficient permissions, etc.), Stash will delete the database entry and related files and claim that the file was deleted. On a subsequent scan, the file will magically reappear. 

This PR tries to delete the file first (if file deletion is set), and then carries on with the other actions if it succeeds. If it fails, the UI is notified.

### DIFF
--- a/pkg/api/resolver_mutation_gallery.go
+++ b/pkg/api/resolver_mutation_gallery.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/manager"
 	"github.com/stashapp/stash/pkg/models"
 	"github.com/stashapp/stash/pkg/plugin"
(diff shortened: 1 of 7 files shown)


SmallCoccinelle (commented):
This looks good to me as it stops the problem.

Usually, the way you get a database and a file system to agree is that you choose one side to be authoritative, commonly the db side. You delete by marking for deletion in the db. If the db transaction fails, you roll it back, and nothing has happened to the file system.

To make the file system agree, you run a (background) cleanup job: find files marked for deletion in the db, and delete them. Then update the database. This essentially makes the file system a follower of the authoritative db. If files can't be deleted, you report it to the user. The deletion from disk is idempotent: if the file has already been deleted, we can just update the db.

WithoutPants:
I have started on an alternative approach - see here: https://github.com/WithoutPants/stash/tree/file-deleter

Adds `file.Deleter` and `scene.FileDeleter`. Scene destroy code has been refactored to use these. Untested and still WIP. Gallery and Image code still to do.

This consolidates the file deletion code into a single point. This would be my preferred approach.

kermieisinthehouse:
Closing in favor of #1954

PR #1893, opened 2021-10-25
Refactor scraper top half

SmallCoccinelle:
This PR refactors the top half of the scraper code, and simplifies it.

### Not addressed by this PR

As a scope limitation, and because the PR is already too large for its own good:

* Mapped Scrapers. There are definitely cleanups possible in this area.
* ScraperActionImpl. This interface currently results in some big switches in the code, but I have a hunch once we work on it, those switches will disappear.
* storing the *http.Client in a structure. It's currently passed around, which is fine. But once things settles it probably lives inside the scraper code itself.
* Scraper lookup optimizations. Fast partial URL matches could be nice to have. I.e., a function which given a url, gives back scrapers which support said URL. Without having to test url matching on each scraper in turn.
* Elimination of `context.TODO()` in the scraper package.
* Introduction of a coalescing `Query.scrape(..): ScrapeResult` fields in the graph. Instead we use the existing fields and morph outputs.

### Goals of this PR

The immediate goal of this PR is to make the scraper code less reliant on large specialized interfaces.

The key idea to achieve this is to introduce a `union ScrapedContent = ...` in the GraphQL Schema. `gqlgen` then generates

```
type ScrapedContent interface {
    IsScrapedContent()
}
```

And implements the function:

```
func (p ScrapedPerformer) IsScrapedContent() {}
func (s ScrapedScene) IsScrapedContent() {}
...
```

hence making `ScrapedContent` into a phantom interface we use as `tagging`. Once this is in, we have output variance so we don't need separate function calls for each type we produce. We just return `ScrapedContent` instead. In the resolver code, we have marshal functions which can convert the interface to its underlying representation, so we can support all the old API endpoints. In the future, we can use GraphQLs union type and write queries using fragment expansions and the spread (`...`) operator.

Once we return `ScrapedContent`, all the code paths collapse unto a few main ones, and a lot of code is eliminated as a result.

The scraper cache sees some changes:

* Use a map, indexed by scraper ID rather than a slice to store scrapers. Speeds up lookup.
* All postprocessing code is moved to it's own .go file.
* Postprocessing now switches on `ScrapedContent` and runs the correct postprocessing depending on content.
* As a result, there's way more sharing in the scraper cache and the interface is way smaller. This is also the interface used by the remainder of the application, as it doesn't need to know about scraper internals.

A new `scraper` interface is introduced, used by the scraper cache to store scrapers. The core allows for probing of scrapers.
```
// scraper is the generic interface to the scraper subsystems
type scraper interface {
	// spec returns the scraper specification, suitable for graphql
	spec() models.Scraper
	// supports tests if the scraper supports a given content type
	supports(models.ScrapeContentType) bool
	// supportsURL tests if the scraper supports scrapes of a given url, producing a given content type
	supportsURL(url string, ty models.ScrapeContentType) bool
}
```

The scraper cache type converts this into expanded interfaces, here for urls,

```
// urlScraper is the interface of scrapers supporting url loads
type urlScraper interface {
	scraper

	loadByURL(ctx context.Context, client *http.Client, url string, ty models.ScrapeContentType) (models.ScrapedContent, error)
}
```

when it needs a scraper to support additional functionality. This scraper interface is implemented either directly by scraper code, or through the structure `group` which is a group of scrapers, essentially an `Yaml` scraper file. As a result, the config_scraper interfaces can be removed.

The end result is that a lot of the code paths are coalesced and use the same path through the code, and only varies at the lower levels. This variance can be tackled in a later PR, as we push the flow further down into the scraper subsystem.

### Smaller stuff

* Pass scraperCache in the resolver state
* Eliminate some `context.TODO()` pathing. Tie it to the request context instead.

### TODO

There are a number of things left to do in this PR:

- [x] Re-read all of the code and do a fixup run.
- [x] Rename stuff for consistency, especially around the interface names.
- [x] Rearrange functions so they live in better localized environments.
- [x] Remove the fixup commit and fold it into the right commit (rebasing messed it up)
- [x] Test `Query.list...Scrapers`
- [x] Test `Query.scrapeSingleScene(..)`
- [x] Test `Query.scrapeMultiScenes(..)`
- [x] Test `Query.scrapeSinglePerformer(..)`
- [x] Test `Query.scrapeMultiPerformers(..)`
- [x] Test `Query.scrapeSingleGallery(..)`
- [x] Test `Query.scrapeSingleMovie(..)`
- [x] Test `Query.scrapePerformerList(..)`
- [x] Test `Query.scrapePerformer(..)`
- [x] Test `Query.scrapeScene(..)`
- [x] Test `Query.scrapeGallery(..)`
- [x] Test `Query.scrapeFreeones(..)`
- [x] Test `Query.scrapeFreeonesPerformerList(..)`
- [x] Test `Query.queryStashBoxScene(..)`
- [x] Test `Query.queryStashBoxPerformer(..)`


SmallCoccinelle:
Here's what I've done.

I went over each use of `fmt.Errorf(..)` in the scraper code that I have touched in this PR.

Error's wrapping a static `errors.New(..)` now generally follow the form: `fmt.Errorf("%w: <details>", Err)` which I agree is reading far nicer. The error is a contextual thing, and the details is what you are interested in.

Error's which sit higher up the chain generally follow the form: `fmt.Errorf("<context>: %w", Err)`. They add context they know at their level of the call stack.

I've hesitated to use the word "error" as a prefixing on a `fmt.Errorf(..)`:

* You get stuttering if something wraps the error: "error context1: error context2: error while foo"
* It's an error. It's already implicit something is wrong. These generally bubble into the GraphQL world. Suppose we have:

```graphql
query Q {
  scrapeSingleScene(source: {scraper_id: "i-dont-exist"}, input: {query:"i-dont-exist"}) {
    title
  }
}
```

Then the return value is going to be:

```json
{
  "errors": [
    {
      "message": "scraper not found: id i-dont-exist",
      "path": [
        "scrapeSingleScene"
      ]
    }
  ],
  "data": null
}
```

i.e., the return JSON contains a key errors, so it's already implicit that message is the error message. The same is true in a log line of a severity. It's given that the logged thing is a warning, error, emergency, ...

At least that's the extent of my reasoning here. I've tried to be consistent to those rules.

pkg/api/resolver_query_scraper.go
```diff
@@ -6,114 +6,178 @@ import (
...
+				m[s.ID] = s
+			}
+		}
+	}
+
+	var ret []*models.Scraper
+	for _, v := range m {
+		ret = append(ret, v)
+	}
+
```
bnkai, line 58: After scrapers are deduped from the map the order is lost. The current dev version prints the scrapers in the order they were loaded ( coincides with a by name sort in most cases). I think we could sort the `ret` here by the `Name` field before returning it. This makes the lists look better either in the settings page and in the scrape with actions.

pkg/api/scraped_content.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// marshalScrapedPerformer will marshal a single performer
+func marshalScrapedPerformer(content models.ScrapedContent) (*models.ScrapedPerformer, error) {
+	p, err := marshalScrapedPerformers([]models.ScrapedContent{content})
+	if err != nil {
+		return nil, err
+	}
+
+	return p[0], nil
```
bnkai, line 96: In theory we could have a `p` slice here with the first element being `nil` and then some more `actual` performers.
Do we want/need to check for that? The same goes for all the marshalScrape single functions

pkg/api/resolver_query_scraper.go
SmallCoccinelle: Good call, I agree.

pkg/api/scraped_content.go
SmallCoccinelle: In practice this function is to facilitate the existing GraphQL fields. Many of those don't do any kind of multiplicity and only ever returns a single value. If the first value is nil, we want to return nil.

For the multiple case, if I query for [X, Y, Z] and Y fails, I expect [{...Xres...}, null, {...Yres...}] in the GraphQL. I.e., that the order is preserved. This works for the multiplicity-variants.

But it is arguably stronger just to return `[ScrapedContent]` in the GraphQL and use GraphQL to do this conversion for us. It's just that it won't work for existing fields.

pkg/scraper/cache.go
```diff
@@ -0,0 +1,296 @@
...
+// ListScrapers returns scrapers matching a given kind
+func (c Cache) ListScrapers(k models.ScrapeContentType) []*models.Scraper {
+	var ret []*models.Scraper
+	for _, s := range c.scrapers {
+		if s.supports(k) {
+			spec := s.spec()
+			ret = append(ret, &spec)
+		}
+	}
+
```
bnkai, line 163: It sees that the `order` changes from the latest commit need to be replicated here for the `deprecated` listSceneScrapers/listPerformerScrapers/... functions to return a sorted list

SmallCoccinelle: Yep, because the underlying cache now uses a map. I'll get those sorted as well.

bnkai (approved):

pkg/api/server.go
```diff
@@ -69,6 +69,7 @@ func Start(uiBox embed.FS, loginUIBox embed.FS) {
 	txnManager := manager.GetInstance().TxnManager
 	pluginCache := manager.GetInstance().PluginCache
 	resolver := &Resolver{
+		scraperCache: manager.GetInstance().ScraperCache,
```
WithoutPants, line 72: `manager.singleton::RefreshScraperCache` creates a new instance of `scraper.Cache` and assigns it to a member on the singleton. The resolver's copy will likewise need to be refreshed.

pkg/manager/task_identify.go
```diff
@@ -236,7 +236,16 @@ type scraperSource struct {
 }
 
 func (s scraperSource) ScrapeScene(sceneID int) (*models.ScrapedScene, error) {
-	return s.cache.ScrapeScene(s.scraperID, sceneID)
+	content, err := s.cache.ScrapeID(context.TODO(), s.scraperID, sceneID, models.ScrapeContentTypeScene)
```
WithoutPants, line 239: I think the `ScrapeScene` methods should be changed to accept a context and pass that through instead.

pkg/scraper/cache.go
```diff
@@ -0,0 +1,305 @@
...
+	c.scrapers = scrapers
+	return nil
+}
+
+// TODO - don't think this is needed
+// UpdateConfig updates the global config for the cache. If the scraper path
+// has changed, ReloadScrapers will need to be called separately.
+func (c *Cache) UpdateConfig(globalConfig GlobalConfig) {
+	c.globalConfig = globalConfig
+}
```
WithoutPants, line 153: Remove

WithoutPants (changes requested):

pkg/scraper/cache.go
SmallCoccinelle: Done.

pkg/manager/task_identify.go
SmallCoccinelle: Yes, this sounds like the right move to me. Done.

pkg/api/server.go
SmallCoccinelle: Solved this by adding a function to the resolver which reads it from the singleton. If the pointed-to object isn't stable, I'd rather have a single point where it needs to be updated than having multiple points you need to remember to update.

However, this then uncovers a pile-on problem: the singleton instance needs a mutex to protect access to the field as it's being updated, possibly by another goroutine. I'm not going to do that in this patch, however since it is a new problem and not one created by this PR.

WithoutPants (approved):

PR #1898, opened 2021-10-26
Improve flag and environment config overrides

### LINKED ISSUE #1887: [Bug Report] STASH_GENERATED envvar leads to broken configs
**Describe the bug**
Not absolutely certain of the exact bug pathology, but I suspect that the envvar is ignored and overwritten during new system setup, but that subsequent server starts after that use the envvar-provided path, leading to two generated folders being created / content being lost if it was generated in the meantime.

**Expected behavior**
The envvar should always be respected and changing it should be disallowed in the UI if set.

WithoutPants:
Fixes #1887 

Currently, if the generated folder is changed while running and the `STASH_GENERATED` environment variable is set (such as when running with the standard `docker-compose` configuration), then the generated folder is reset and overridden the next time stash is run. This has caused some problems where users have changed this value, then scanned or generated content, then when stash is next run their generated content is nowhere to be found, or somewhere unexpected.

Environment variables and flags should act like overrides and not change the underlying configuration. Currently, if you want to run stash on a different port temporarily and run `stash --port 9876`, then the port value is set in the configuration.

This PR refactors the config system by having two separate `viper` instances - one backed by the config file and the other containing only the environment overrides (variables and flags). This means that environment variables and flags passed into stash are maintained only for that execution. 

The system will now return an error if the user attempts to change configuration values that are currently overridden.

Changed the setup wizard so that it will hide the generated folder section if the generated value is already populated (by the environment). The setup wizard will also now pre-populate the library list with the current library value if set. This effectively means that when running with the standard docker-compose configuration, the wizard will prepopulate the library with `/data` (assuming the `docker-compose.yml` file hasn't been modified) and the generated section will be hidden (but set to `/generated` by default).


kermieisinthehouse:
Tested - a few notes
If the user specifies no stashes, the onboarding tool doesn't complain. We should probably warn about this.
If a user runs from the default docker-compose, the only thing they are asked is for the config path. The default will be useful for 99% of people, so we should probably de-emphasize the prompt or emphasize that it is optional.

kermieisinthehouse:
I also can't get the original setup question to show up, the one with two buttons for setting up in the home screen or in the PWD. This makes "portable" installations easy and shouldn't be removed I think

WithoutPants:
> I also can't get the original setup question to show up, the one with two buttons for setting up in the home screen or in the PWD. This makes "portable" installations easy and shouldn't be removed I think

I have not been able to reproduce this. Can you please give explicit instructions on how to reproduce?

kermieisinthehouse:
If you do `make docker-build`, and then edit `docker/production/docker-compose.yml`'s `image` to point to `stash/build:latest`, and then remove the entire `environment` section, so that no envvars are set, you are presented with "Stash will use the following configuration file path: /root/.stash/config.yml", which I suppose is correct, but how does it _know_ that?

WithoutPants:
The docker images include `ENV STASH_CONFIG_FILE=/root/.stash/config.yml`. The wizard detect this environment variable and don't present the choice if it is present.

kermieisinthehouse:
Whoops, that's on me, I didn't know it was built-in.

kermieisinthehouse (approved):

WithoutPants:
> If the user specifies no stashes, the onboarding tool doesn't complain. We should probably warn about this.

I will add this in a separate PR.

PR #1900, opened 2021-10-26
Vite-based frontend builds

kermieisinthehouse:
Vite is a project from the Vue founder to speed up JS builds. It uses esbuild internally, and rollup for the final bundle. I've removed create-react-app from the dependencies. I've also enabled gzipping static assets as part of the UI build, and enabled sending the gzipped assets directly to browsers as-is, freeing up CPU while serving requests.

I think the numbers speak for themselves:
| X | Current Develop | This Branch | Difference Improvement |
| --- | --- | --- | --- |
| UI Build Time| 33s | 24s | 27% |
| UI Bundle Size | 21MB | 3MB |  85% |
| Release Executable Size | 59MB | 42MB | 28% |


WithoutPants:
Going to push this into the next version so that it gets a cycle worth of testing.

kermieisinthehouse:
Probably a good idea. I've been running it in my instance trying to suss out bugs but nothing's come up, and I don't expect any.

The only thing I've left untested is the subpath stuff, because I don't have a setup for it. However, I've allowed for it by leaving the index.html unchanged and uncompressed so that the string replacement would work. So I suspect it's exactly the same

WithoutPants:
On Windows, I get a blank page and the following error in the browser:
```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.
```

WithoutPants:
Restarting the browser appears to have fixed the issue.

PR #1917, opened 2021-10-27
Draft: Support direct streaming of audio files. 

### LINKED ISSUE #1258: [Feature] Audio Stories
there is a issue for this but as im putting a bounty on it thought i should make mine new.
old issue #11
how i see this working is,
is to play a mp3,ogg,wav,m4a,etc in the Brower
metadata i like to see is 
name
author
Performer
url for source
website of source
date
language 
cover front
cover back
Rating
Tags
Details

rezreal:
/closes #1258

More or less open for discussion. This was the quickest attempt to make audio files workable for me.
I think audio can be treated as any other video - so there is no new backend machinery required. You could make "movies" out of them - which I think would be highly desirable for multipart hypnosis sessions.

The thumbnail scraper may need to be adjusted to return a static dummy image for audio or simply a black screen. A proper cover image could still be provided as manual thumbnail.

### DIFF
--- a/pkg/ffmpeg/ffprobe.go
+++ b/pkg/ffmpeg/ffprobe.go
@@ -53,6 +53,9 @@ const (
 	MimeMp4            string     = "video/mp4"
 	MimeHLS            string     = "application/vnd.apple.mpegurl"
 	MimeMpegts         string     = "video/MP2T"
+	MimeAudioMpeg      string     = "audio/mpeg"
+	MimeAudioOgg       string     = "audio/ogg"
+	MimeAudioWav       string     = "audio/wav"
 )
 
 // only support H264 by default, since Safari does not support VP8/VP9
@@ -141,6 +144,21 @@ func IsValidAudioForContainer(audio AudioCodec, format Container) bool {
 		return IsValidAudio(audio, validAudioForWebm)
 	case Mp4:
 		return IsValidAudio(audio, validAudioForMp4)
+	}
+	return false
+
+}
+
+func IsAudioContainer(format Container) bool {
+	switch format {
+	case "mp3":
+		return true
+	case "flac":
+		return true
+	case "ogg":
+		return true
+	case "wav":
+		return true
 	}
 	return false
 
--- a/pkg/manager/scene.go
+++ b/pkg/manager/scene.go
@@ -264,6 +264,14 @@ func GetSceneStreamPaths(scene *models.Scene, directStreamURL string, maxStreami
 			MimeType: &mimeMp4,
 			Label:    &label,
 		})
+	} else if GetAudioMimeType(scene.Format.String) != "" {
+		label := "Direct audio stream"
+		mimetype := GetAudioMimeType(scene.Format.String)
+		ret = append(ret, &models.SceneStreamEndpoint{
+			URL:      directStreamURL,
+			MimeType: &mimetype,
+			Label:    &label,
+		})
 	}
 
 	// only add mkv stream endpoint if the scene container is an mkv already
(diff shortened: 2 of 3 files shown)


kermieisinthehouse:
I always pictured audio as a separate tab, but perhaps putting them in the scenes table and UI lists isn't actually too bad. This feature is requested /extremely/ often.

This would need a little more refinement though. We need a way to filter in/out audio-only scenes, and we need to decide what to fill into the database for video attributes like height and width (probably nulls). They should be displayed sllightly differently in scene lists to denote they do not contain a video stream, and we probably don't want or need to load a full jwplayer session (and redundancies like color filters) - an audio specific player will do. The built-in browser embedded audio player is probably fine.

We also need some handling around generate tasks, to either skip for some jobs (like phash) or special handling (waveform for sprite, perhaps? Also for thumbnail)

Lastly, I ran `find . -type f | sed -n 's/..*\.//p' | sort | uniq -ic | sort -nr` over my audio archive, and found these filetypes:
```
m4a
mp3
mp4
webm
opus
wav
aac
mkv
mpeg
aif
```

WithoutPants:
There are enough distinctions between video and audio that forcing audio into the video concept seems like a bad workaround at best. I understand the desire for supporting audio, but I'd want to implement it properly. My concern with this approach is that it may become difficult to maintain features specific to audio while adding features for video. An example is in this PR already - it blithely removes code concerned with ios device compatibility.

rezreal:
Thanks for your feedback. I agree that treating audio as "video" comes with weird tradeoffs that feel hacky. But from a "business" perspective I'd argue that treating "audio" as scenes (or maybe "clips"?) is not too far off - especially because it  shares a vast amount of features of it (playback, positions, markers, playback speed (jw specific?), haptics, previews with audio and the whole metadata and tagging/scraping infrastructure).  From my understanding, phashing would also work for audio files. I would not like to miss any of that in case audio becomes a dedicated feature. 
Tbh I am not advocating for anything because I do not know the desired architecture at all yet - but I'd be happy to help to push this feature somewhat further if there is agreement on the direction.

SmallCoccinelle:
The core idea in a phash is to run a 2d-DCT over the incoming image after it's been resized to a small block. The DCT reduces out frequencies, if you pick the low coefficients, and after some more manipulation, you have a 64bit hash. That won't work out of the box for audio since it isn't 2d. What could work would be an 1d-DCT over the audio stream, but I'm pretty sure that hasn't really been defined in the context of phashes. Spotify has some blog posts on how to attack audio streams, and FFT/DCT are definitely part of the puzzle. But it requires a different approach since video and audio are different.

rezreal:
TBH I am not so deep in  the details, just saw that (if this is the correct lib), there seem to be something like a audiophash: https://github.com/aetilius/pHash/blob/master/src/audiophash.h

SmallCoccinelle:
We use a go library for this, and it doesn't have that function at the moment.

SmallCoccinelle:
It's pretty nice they defined phashes for audio streams as well, and it looks like they use an approach like the one I sketched. But it does require some more work to get phashes for audio. Apart from the go code, you probably also need to figure out how to read the audio data. In the code you linked, they can use mpg123 for this, but there are other formats and you probably want to support reading those as well.

I think audio is something that needs support. It's awesome that it can work! What I'm hesitant about is to fold it into video. It is conceptually different from Video. So you end up having people populate their database with audio data mixed with video data. If we then later on add some true audio support, you run into trouble with all the existing data sets users have in their databases. Splitting this apart in a way that works is often much harder than having them separate and glue'ing them together.

QxxxGit:
Random thought: Should audio eventually be playable in the background? 

So let's say you go to an audio clip. You might navigate to another page while its playing. Then the bottom of the screen would display controls and a seek bar, maybe a loop button too.

I'm thinking this will be requested at some point in the future. It's probably common for people to listen to audio while viewing other media such as images. For now, multiple tabs should work; but I figured I'd bring it up just in case.

rezreal:
@QxxxGit Same question for video with pip :)

kermieisinthehouse:
Audio files is going to require a big effort to add to Stash, and some architectural planning. Closing this for now, but thanks for participating!

PR #1922, opened 2021-10-28
Change performer country value to be ISO code

InfiniteTF:
Countries are currently stored as text fields, which are parsed on the frontend to show flags. Any typos, or unrecognized country name variations break the flags. It's also similarly hard to translate the names, and filter by countries.

By switching to ISO codes instead we can easily show the correct flag and localized name. I've also added a country selector so users no longer need to type it in.

 The migration and scraper resolving is based on the existing iso detection logic on the frontend. It has a 100% match rate on the performers in my db that have a country name set.


SmallCoccinelle:
Another idea. Implement country as a GraphQL scalar: https://gqlgen.com/reference/scalars/

I have a hunch this will wrap up a lot of code into a single neat local package, where most of the Go code is going to rely on it, including the sql database.

WithoutPants:
We definitely should avoid erasing user data.

My suggestion is to rename the existing `country` column to something else and repurpose it - or just rename it to `countryLegacy` or something. The new `country` column will be set according to the mappings already present in the migration. If the value is correctly mapped (exactly), then the legacy field can be wiped.

HijackHornet:
I would suggest an auto-fill input area with a dropdown option.
By doing that, you avoid spelling mistakes while also providing a way to enter free text. Example :
https://ej2.syncfusion.com/react/documentation/combo-box/how-to/autofill/ <- click preview to test

WithoutPants:
Rebased and changed the migration so that it tries to convert countries to ISO codes, and falls back to the original value otherwise. Changed the implementation in the UI so that the user can enter a custom country, in addition to selecting country codes.

Issue #1754, opened 2021-09-22
[Bug Report] Tag Hierarchy loop detection is bugged


kermieisinthehouse:
**Describe the bug**
The loop detection doesn't allow you to add some sub-tags, when they are valid.

**To Reproduce**
1. Create four tags, A,B,C,D
2. Under A, list B and C as sub-tags.
3. Under D, list B as a sub-tag, and save.
4. Under D, attempt to list C as a sub-tag.
5. The operation fails with "Cannot apply tag "C" as it is linked to "A" which already is a child"

**Stash Version: (from Settings -> About):**
Develop branch after 0.9.0

@gitgiggety

gitgiggety:
If I understand you correctly you end up with `A <- C`, + `A <- B <- D <- C`, correct? In that case the error fact there is an error is correct, although the message should be "already is a parent". Because when applying this `C` would thus have `A` as parent twice, once directly and once via `D -> B`. And I'm guessing that if you try to edit `C` and set `A` and `D` as parent it will be more clear what's going on. (or at least state "parent" instead of child)

_But_ this limitation shouldn't be necessary. And in case there is a valid use case to allow it it could be removed anyway.

kermieisinthehouse:
Sorry, there seems to be a misunderstanding. The hierarchy is only one layer deep. I created this structure earlier in the history of the patch, so it used to be possible, but now I cannot modify it without it throwing errors, or recreate it. Here's a diagram and a usecase:

![Screenshot_20210921-231008_1](https://user-images.githubusercontent.com/84814418/134291993-bdd2dde9-1697-4600-b915-88af0b79b96d.jpg)

Reproduction steps, same as above:
![Screenshot_20210921-231413_1_1](https://user-images.githubusercontent.com/84814418/134292535-9efb11ce-3df1-4c1d-974c-97a4bd091cda.jpg)
![Screenshot_20210921-231439_1](https://user-images.githubusercontent.com/84814418/134292551-0a6ea8da-daa0-4690-84be-6834a122e473.jpg)
![Screenshot_20210921-231432_1](https://user-images.githubusercontent.com/84814418/134292558-420e5e2b-a1b4-489d-a915-ee50a170c40f.jpg)

gitgiggety:
Sorry, you're right, I did misread but the image clears it up.

[...]

PR #1926, opened 2021-10-29
Fix tag hierarchy validation

7dJx1qP:
An attempt to fix https://github.com/stashapp/stash/issues/1754

Updates were made based on a conversation with @SmallCoccinelle in Discord.

The first thing I noticed while investigating the tag hierarchy was that in `tag.go` the queries for FindAllAncestors and FindAllDescendants were returning tags that weren't ancestors/descendents of the tag being checked. I think it was returning all descendents of all ancestors and all ancestors of all descendents. Removing the last union select from each query seems to address that. This change alone actually fixes the bugged scenario described in the issue https://github.com/stashapp/stash/issues/1754#issuecomment-924620432

I then noticed that taking the test scenario and attempting to add the D-> A relationship was returning an error. The rest of the code changes are to address that.

The end result is intended to just allow for any tag hierarchy that is a DAG.

The changes break the current tests and I haven't check yet if MergeHierarchy works properly.


7dJx1qP:
I updated the recursive query to include the tag path so it can be displayed in the error message. It should help users get an idea of why an update is rejected.
image

For this change I created a TagPath model which is just a Tag model with another Path field, but it breaks a bunch of tests. Is there a better way to implement returning a tag model + path from the query?

7dJx1qP:
Added logging the save and merge errors
image

SmallCoccinelle:
This is pushing in the right direction I think.

7dJx1qP:
I don't think I'll be adding to this unless something needs fixing. The tag UI could use improvement but I think that's outside the scope of this PR. And I don't know if I could successfully implement a more significant change anyway.

WithoutPants:
> One of the key things in a fix here is to establish the rules of the tag hierarchy. Those rules define a good test set, and also defines if a future issue is a bug or not. Some observations:

The only rule established was that there can be no cycles. We allow tags to have multiple parents.

WithoutPants:
There are a few issues in this PR and in the validation code - not least that it's not being triggered when creating tags. I'm going to spend some time refactoring and testing and will push to this branch.

7dJx1qP:
It looks like no validation on tag creation is a bug that affects the current release as well. I knew that the tag hierarchy described in https://github.com/stashapp/stash/issues/1754 worked when creating tags, but I just incorrectly assumed it was due to the validation working properly when tags were created and not working when editing. Good catch.

This means users might have already introduced loops into their hierarchy without knowing it and would only find out if they went back to edit the tag. Maybe on migration or startup there should be a validation run on the whole hierarchy to log/alert the user of any existing loops?

WithoutPants:
I'm content to leave it as a latent issue for the moment. It will only come up when the user tries to edit a tag in the hiearchy, at which point they will need to correct it, but I think it's likely to be a rare issue and I'm not keen on adding a heap of code to detect it.

PR #1927, opened 2021-10-29
WIP: Autotag Improvements

kermieisinthehouse:
WIP, it probably doesn't even build yet.

Final goal is to have:
- A checkbox for tags to enable / disable eligibility to autotag. This is useful now that we have hierarchies, to avoid tagging Categories-of-Categories and the like. Might extend to studios/performers depending.
- Memoization of compiled RE objects during an autotag run to save CPU
- A dedicated query function for the autotag task that reduces necessary n+1 queries and sorts by ROWID instead of path, which is where 99% of the CPU goes during an autotag run
- Indexes for that query so that autotag can scream

Current time to beat is about 2 weeks to autotag images on my instance.


ui/v2.5/src/components/Tags/TagDetails/Tag.tsx
```diff
@@ -166,9 +166,12 @@ export const Tag: React.FC = () => {
 
   async function onAutoTag() {
     if (!tag?.id) return;
+    if (!tag?.allow_autotag) {
+      Toast.error({content: intl.formatMessage({ id: "toast.autotag_not_allowed" })})
+    }
     try {
       await mutateMetadataAutoTag({ tags: [tag.id] });
-      Toast.success({ content: "Started auto tagging" });
+      Toast.success({content: intl.formatMessage({ id: "toast.started_auto_tagging" })});
```
kermieisinthehouse, line 174: Beat me to it https://github.com/stashapp/stash/pull/1927/files#diff-74e92961bf92654d420b28e6c909729ee5c396b44af0eaa9b3b44c8137bbf526R24

PR #1932, opened 2021-10-30
Suppress a benign warning

SmallCoccinelle:
On serving studio images, we can get broken pipe errors. They are relatively common, and can safely be suppressed as they are entirely benign. Haven't seen much testing yet, but should be safe.

### DIFF
--- a/pkg/api/routes_studio.go
+++ b/pkg/api/routes_studio.go
@@ -2,8 +2,10 @@ package api
 
 import (
 	"context"
+	"errors"
 	"net/http"
 	"strconv"
+	"syscall"
 
 	"github.com/go-chi/chi"
 	"github.com/stashapp/stash/pkg/logger"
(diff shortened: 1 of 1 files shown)


pkg/api/routes_studio.go
```diff
@@ -47,7 +49,12 @@ func (rs studioRoutes) Image(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := utils.ServeImage(image, w, r); err != nil {
-		logger.Warnf("error serving studio image: %v", err)
+		// Broken pipe errors are common when serving images and the remote
+		// connection closes the connection. Filter them out of the error
+		// messages, as they are benign.
+		if errors.Is(err, syscall.EPIPE) {
```
kermieisinthehouse, line 55: Shouldn't this be the inverse?

SmallCoccinelle: Yeah, it is late :) Fixed in the next push here.

kermieisinthehouse (approved):

PR #1941, opened 2021-11-01
Small docker cleanup

kermieisinthehouse:
- Remove useless pip cache from docker releases, saving 600kb from each image
- Update docker/build readme.
- In the docker/production readme, move instructions above the explanation of what docker is - If you clicked on docker install, you probably already know what it is
- Delete unmaintaned and outdated docker/production/x86_64, it is just going to confuse users as it is in the same directory as the docker instructions


docker/build/x86_64/README.md
```diff
@@ -1,55 +1,18 @@
...
 
 From the top-level directory (should contain `main.go` file):
 
 ```
-docker build -t stash/build -f ./docker/build/x86_64/Dockerfile .
+make docker-build
 
 ```
 
-# Running the docker container
```
WithoutPants, line 14: Why was this section removed?

kermieisinthehouse: I thought that stuff was 'obvious' if you've deployed to docker before, but I've restored it

PR #1951, opened 2021-11-03
Support a minModTime input on metadata scans.

SmallCoccinelle:
Extend the GraphQL world with a Timestamp scalar. This scalar can parse a number of
different input formats, among those RFC3339 and relative time entries.

Extend Mutation.metadataScan(input: $input) to support a new optional
value, minModTime. If set, the scanner will exit early if the file it
is looking at has an mtime older than the cut-off point of minModTime.

This speeds up scanning in the case where the user knows how old the
changes on disk are, by exiting the scan early if that is the case.


SmallCoccinelle:
Tests the waters for some of the FS Automation QoL changes we could make. This is one of them.

WithoutPants:
I think accepting a modtime parameter would be more useful for typical use cases.

SmallCoccinelle:
Yeah, that is also a possibility. Something like `afterModTime: Time` is what you have in mind?

WithoutPants:
`minModTime` parses better to me, but yeah.

SmallCoccinelle:
I'll hack that solution up then. It shouldn't be much different.

SmallCoccinelle:
It now accepts both. Give it an RFC3339 date and it parses as a date. Give it '<4h' and it parses that as "4 hours in the past".

SmallCoccinelle:
Often, relative times are better ergonomics for manual invocation. You don't have a full RFC3339 date, and the parser for it is actually not very lenient: it doesn't accept e.g., "2021-10-04" but wants the full "2021-10-04T00:00:00Z". This is awesome for consistency, but a human user would probably rather like to say "<24h" instead.

SmallCoccinelle:
Seems like the CI system is finding some nice timezone problems.

pkg/manager/task_scan.go
```diff
@@ -47,9 +47,14 @@ func (j *ScanJob) Execute(ctx context.Context, progress *job.Progress) {
 
 	logger.Infof("Scan started with %d parallel tasks", parallelTasks)
 
+	var minModTime time.Time
+	if input.MinModTime != nil {
+		minModTime = *input.MinModTime
+	}
+
```
bnkai, line 54: Does this need to be here?
Can't it be inside the queueFiles function?
That way we can get the input directly eg `j.input.MinModTime` and not use an extra argument in the queueFiles function

SmallCoccinelle: You are right, it needs to be pushed down.

WithoutPants:
We should probably consider some sort of file filter input and put the min mod time in there. Are there any other filter criteria that would be suitable for such a thing? More specific file filtering, for example?

WithoutPants:
Is there any issue with making this a little more lenient for the date input? Could we add `YYYY-MM-DD` as an accepted format? From an implementation point of view, we would probably just try the full RFC3339 format first, then fallback to the date-only format.

WithoutPants:
We actually have `utils.ParseDateStringAsTime` that essentially supports RFC3339, `YYYY-MM-DD`, `YYYY-MM-DD HH:MM:SS`, and `YYYY-MM-DD HH:MM:SS TZ`. We should probably just use that, in addition to the relative formats.

SmallCoccinelle:
Ah, this should be done indeed! I'll make it happen.

WithoutPants:
I'm not going to add a changelog for this (yet) since it's not something most users will be able to use easily. I think we can integrate something into the UI in a later change though.

PR #1954, opened 2021-11-04
Refactor file deletion

### LINKED ISSUE #1313: [Bug Report] Gallery Cleanup leaves empty entries in Gallery View
**Describe the bug**

I (accidentally) excluded Images from a Directory and run a Cleanup. This cleanup removed the Image-Galleries that were as zips in the directories. 

The Cleanup removed the Image Entries as expected, but left the Gallery-Entries still there (see screenshot)

**Expected behavior**

I expect a Cleanup of cleaning up the gallery-entries as well

**Screenshots**
image

v0.6.0-34-ga5e9e7ab

WithoutPants:
Supercedes #1892.

Adds `file.Deleter` concept. This object exposes `Files` and `Dirs` methods which mark files and directories, respectively, for deletion. The marking process involves renaming the marked items to add a `.delete` suffix. If the rename fails, then an error is returned. When a transaction is committed, the `Commit` method should be called which deletes the marked items. If the transaction is rolled back, then the `Rollback` method should be called which restores all marked items back to their original states.

Destroy mutations and clean code is updated to use the deleter concept. The major change in behaviour here is that if any files marked for deletion fail to be marked, then the transaction will be rolled back, leaving the database and filesystem in their original states. 

Also includes a fix to delete empty folder-based galleries during clean.

Fixes #1313


kermieisinthehouse:
I discovered and fixed in the other PR that galleries could be created in the main stash path, e.g. /media is your Stash path, and it's just full of images. We should not attempt to delete the gallery file in this case

bnkai:
> I discovered and fixed in the other PR that galleries could be created in the main stash path, e.g. /media is your Stash path, and it's just full of images. We should not attempt to delete the gallery file in this case

Isn't that already taken into account? os.Remove will not remove non empty directories...
```go
	for _, gallery := range galleries {
		if utils.IsTrue(input.DeleteFile) && !gallery.Zip && gallery.Path.Valid {
			// try to remove the folder - it is possible that it is not empty
			// so swallow the error if present
			_ = os.Remove(gallery.Path.String)
```

kermieisinthehouse:
The stash path might be empty, if it contained only images and you're deleting it. We should preserve it or remove it from Stashes, and I think preserving it is the least surprising to the user

bnkai:
Oh now i get what you mean.... If the gallery path is the same with a root `library` path.... Yeah agreed

kermieisinthehouse:
Deleting an undeletable scene on this branch does not show an error on the UI, you need to return false somewhere

WithoutPants:
> Deleting an undeletable scene on this branch does not show an error on the UI, you need to return false somewhere

I haven't been able to reproduce this. Can you provide instructions to reproduce?

kermieisinthehouse:
That problems seems to have resolved, I can no longer reproduce.

I have found another issue: you can't delete stuff that has already been deleted.

1) Scan a scene
2) Delete the backing file
3) Try to delete the stash entry
4) The rename fails

bnkai:
> That problems seems to have resolved, I can no longer reproduce.
> 
> I have found another issue: you can't delete stuff that has already been deleted.
> 
>     1. Scan a scene
> 
>     2. Delete the backing file
> 
>     3. Try to delete the stash entry
> 
>     4. The rename fails

Was 3 with the `delete file` option ticked or not? 
For cases like this shouldnt the user have to run a clean first ?  2 delete file 2.5 run a clean .... In cases were the file is missing that means either the user renamed the file (so he should run a new scan) or deleted the file (so he should run a clean)

WithoutPants:
Need to establish what the correct expected behaviour is when trying to delete a file that does not exist.

We can either fail (current behaviour), or we can check if the file exists and if not then return silently. The latter could be problematic if the file is merely temporarily unavailable.

kermieisinthehouse:
I think if the user intention is to remove the file and its metadata, I think we should remove the metadata and just silently pass. 

The worst outcome would be the file shows up again during a scan.

kermieisinthehouse (approved):

PR #1957, opened 2021-11-04
Set us up for Search Engine experiments - Introduce an event dispatcher

SmallCoccinelle:
# Background

A search engine would like to receive online updates as there are changes to data. This change enables that functionality.

# PR

Introduce package event with a dispatcher subsystem. Add this to the
manager and to the plugin subsystem.

Whenever the plugin subsystem execute a PostHook, we dispatch an
Change event on the event dispatcher bus. This currently has no
effect, but allows us to register subsystems on the event bus for
further processing. In particular, search.

By design, we opt to hook the plugin system and pass to the event bus
for now. One, it makes it easier to remove again, and two, the context
handling inside the plugin subsystem doesn't want to live on the
other side of an event bus.

While here, write a test for the dispatcher code.

### DIFF
--- a/pkg/event/dispatcher.go
+++ b/pkg/event/dispatcher.go
@@ -1,0 +1,105 @@
+// package event dispatches change-events in stash
+package event
+
+import (
+	"context"
+	"sync"
+)
+
+// ChangeType defines what has changed
+type ChangeType int
+
+const (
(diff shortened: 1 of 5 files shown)


Gykes:

Issue #1868, opened 2021-10-20
[Feature] Expanded support for external links for Stash Entities


stg-annon:
cross posting this from [stash-box](https://github.com/stashapp/stash-box/issues/113) to track progress in Stash

**Is your feature request related to a problem? Please describe.**
Currently many entities only support a single external URL, it would be nice in some cases to have a structured way of adding links to these entities, for example performers may have many links to external sites for socials and other bios, movies may have multiple links as well

**Describe the solution you'd like**
A solution similar to #1790 and [Stash-Box PR#140](https://github.com/stashapp/stash-box/pull/140) for stash with the ability to have multiple URLs for entities where it makes sense to have them.

e.x.
a performer may have multiple external URLs that could be scraped for metadata, a stash user may want to store all these URLs in stash instead of just one

xx790:
Progress so far:

- Scenes
  - PR #3852 - already shipped ✔
- Images
  - PR #4000 - already shipped ✔
- Galleries
  - PR #4114 - already shipped ✔
- Movies / Groups
  - PR #4900 - shipped in 0.27 ✔
- Performers
  - PR #4958 - shipped in 0.27 ✔
- Studios
  - PR #6223
- Tags
  - no URLs at all
  - (not sure for everyone, but I'm missing those or description markup)

omfgntsa:
Requesting for additional url fields for studios as well, to allow for patreon links, socials like twitter etc.

feederbox826:
https://github.com/stashapp/stash/pull/6223 Studios have multiple URLs now

just missing tags. If there's a usecase for tags, please open a new issue for it

PR #1959, opened 2021-11-05
Additional performer urls

7dJx1qP:
Basic support for additional performer urls, partial solution to https://github.com/stashapp/stash/issues/1868. Adds a performers_urls table (performer_id, url). It doesn't replace the performer url field and I haven't made an attempt to integrate the additional urls with any other features like scraping or filtering. For now it's just a simple way to store additional performer urls.

Edit tab
image

Details tab
image

### DIFF
--- a/graphql/documents/data/performer-slim.graphql
+++ b/graphql/documents/data/performer-slim.graphql
@@ -13,4 +13,5 @@ fragment SlimPerformerData on Performer {
     stash_id
   }
   rating
+  urls
 }
--- a/graphql/documents/data/performer.graphql
+++ b/graphql/documents/data/performer.graphql
@@ -37,4 +37,5 @@ fragment PerformerData on Performer {
   death_date
   hair_color
   weight
+  urls
 }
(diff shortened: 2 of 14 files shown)


7dJx1qP:
Refactored the input based on the stash-box integration input but with a scrape button added in. This removes the need for an edit button, because the urls are edited in place.
image

WithoutPants:
I couldn't find where this was documented, but the intent was to generalise our existing URL fields into a type/url table.

That is, where we currently have URL, Twitter, Instagram, we would instead have three URL entries in the table: one for website, one for twitter and one for instagram. The UI would be a pair of text fields. We could then use the entries to populate the top links bar by looking for hard-coded strings.

reapier:
@7dJx1qP Any reason why this PR was closed? Mind if if open a similar PR to try to get this in?

reapier:
@WithoutPants should this PR be update to match this change: https://github.com/stashapp/stash/pull/3852/commits

PR #1973, opened 2021-11-08
Add forward seeking buttons to jwplayer

pornulus:
Unfortunately this feature is missing from jwplayer, but someone posted a jQuery-based workaround in the jwplayer bug tracker and I've ported it to vanilla js/typescript. (c.f. https://github.com/jwplayer/jwplayer/issues/3894)

This has a little quirk: because it's mirroring the rewind button it's also mirroring the `10` which looks a little odd. 
On the pro side: it adds a feature I badly want and keeps this PR simple and self contained because no additional asset needs to be added (that would possibly violate copyrights if we took the original icon and edited it)
Hopefully it's added to jwplayer upstream at some point.

Screenshot_20211108_150753
Screenshot_20211108_151126


kermieisinthehouse:
This un-centered the controls on mobile, shifting them to the top left corner. You can actually see the effect in the image in the top post.

PR #1977, opened 2021-11-09
Fix performer tagger field updating

InfiniteTF:
Removed `url/twitter/instagram/details/death_date/weight` fields from the performer tagger since the fields are not populated by stash-box, and not present in the field selector which would allow users to unselect them.

I've also added `hair color` which for some reason was missing.

### DIFF
--- a/pkg/scraper/stashbox/stash_box.go
+++ b/pkg/scraper/stashbox/stash_box.go
@@ -600,6 +600,10 @@ func performerFragmentToScrapedScenePerformer(p graphql.PerformerFragment) *mode
 		sp.EyeColor = enumToStringPtr(p.EyeColor, true)
 	}
 
+	if p.HairColor != nil {
+		sp.HairColor = enumToStringPtr(p.HairColor, true)
+	}
+
 	if p.BreastType != nil {
 		sp.FakeTits = enumToStringPtr(p.BreastType, true)
 	}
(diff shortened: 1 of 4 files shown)


kermieisinthehouse (approved):

WithoutPants:
`PerformerModal` is also used when scraping from non-stash-box sources. All of those fields _may_ be provided by scene scrapes from non stash-box sources. We should also consider that the performer tagger could potentially be used for no stash-box sources in future as well. Is there a particular bug arising out of the extra fields inclusion?

InfiniteTF:
> `PerformerModal` is also used when scraping from non-stash-box sources. All of those fields _may_ be provided by scene scrapes from non stash-box sources. We should also consider that the performer tagger could potentially be used for no stash-box sources in future as well. Is there a particular bug arising out of the extra fields inclusion?

Right, I keep forgetting it's used for other sources as well. I've readded the fields and instead filter the data that is set based on whether it's nullable or not. This matches the logic that determines whether a field is rendered in the modal, so if a field isn't shown or is unchecked it's not updated. This does mean that if for instance the `tattoo` field is the empty string, it won't overwrite the field, but I think that's preferable to overwriting a field that was never shown to the user.

I've also added the death_date, etc. fields to the excluded fields dialog so users can set defaults for them.

WithoutPants (approved):

PR #1979, opened 2021-11-09
Fix identify defaults always returning nil

Releck:
Fixes a bug introduced in 49b2860.

Steps to reproduce:
1. Navigate to `/settings?tab=tasks`
2. Click `Identify` and change any default value, click "Set as default".
3. Reload the page

Expected:
When clicking identify again, the values will be the same ones when "Set as default" was clicked.

Actual:
Identify has reverted to its stock value of autotagger only.

Cause:
`ConfigDefaultSettingsResult` from graphql always returns null. This is because `GetDefaultIdentifySettings()` searches a viper instance that has a nil config.

Tested on:
Windows 10
Go v.1.17.1
Fresh config


kermieisinthehouse (approved):
Tested and reviewed, thanks for the first contribution!

PR #1982, opened 2021-11-09
Add search string parsing

### LINKED ISSUE #1210: [Feature] Improved Search Functionality
Hello,

I think the current search functionality is unintuitive.  For example, when I search "red head kitty", I expect to find the scene titled "red head kitty" as the first result.  Instead, I am presented with every single scene I have that has the words "red", "head", or "kitty" in the title or description.  As someone who has descriptions for almost every scene, this presents a large problem for searching.  I know that I can enter that title in quotes and get the scene I want but there are other examples of this type of search causing an issue.  For example, when I search "skye blue", I typically want just the scenes with Skye Blue, not all the scenes with the word "blue" somewhere in the title or description.  I know that there are workarounds for these issues like searching by performer or even using regex but I am not proficient in regex.  No other porn site has these same search issues.  Apologies if this sounds a bit nitpicky but it greatly affects my regular use of Stash.  I would be happy to provide more examples if needed.  Thanks!

WithoutPants:
Resolves #1210 

Changes the query string parsing behaviour to the following:
* all words are required by default, this means that for a query `foo bar`, the results must include both `foo` and `bar`. Previously, the results would include either `foo` or `bar`.
* the keyword `or` (case-insenstive) can be used to allow for either terms. For example: `foo or bar` would return results with `foo` or `bar`, like the previous behaviour. 
* the `-` prefix can be added to exclude terms. For example: `foo -bar` would return results with `foo` and exclude any results with `bar`.

Phrase parsing is unchanged and can be used in combination with the or and not modifiers. Quotes can also be used to escape the `or` and `-` keywords.

The `or` keyword _cannot_ be used directly in combination with the `-` symbol. What this means is that the search `-foo or bar` will return results with `-foo` or `bar` - it does not exclude `foo` _or_ include `bar`.

`or` sets are anded together. This means that `foo or bar baz or qux` will return results that must include one of `foo` and `bar`, _and_ one of `baz` and `qux`.

This hasn't been performance tuned yet. We should also consider reducing the columns on which these queries are run, since they may cause unexpected results. For example, string queries on scenes uses `title`, `details`, `path`, `oshash`, `checksum` and (weirdly) `scene_marker.title`. The first three make sense, the next two have some limited utility and have been convenient for me personally in the past, and I don't think scene markers belong there at all.

The other remaining task is documentation. The browsing page in general probably needs a manual page. I also considered a mouseover tooltip that documents the query syntax, but I felt this might be overly obtrusive.


kermieisinthehouse:
Huge! Perhaps a replacement for search by hash is creating a new path for scenes, like scene/hash/{HASH}/. That way if you want a specific scene by hash, you can still locate it in the UI - or just get rid of it. I can't think of a usecase for finding a specific oshash in your library

WithoutPants:
> Huge! Perhaps a replacement for search by hash is creating a new path for scenes, like scene/hash/{HASH}/. That way if you want a specific scene by hash, you can still locate it in the UI - or just get rid of it. I can't think of a usecase for finding a specific oshash in your library

Mostly maintenance tasks. Finding a scene from it's preview file for example. The old scene duplicate finder plugin also output the hashes. We can already filter on checksum and hash, so it would still be possible, just a little more onerous.

kermieisinthehouse:
Did some testing. Wow, search is useful now.

Failing tests:
- I cannot get the NOT operator to work by itself, e.g. `-X`, I suspect a performance issue.
- `-X or -Y` returns the results for `X or Y`
- Any 1-2 character term times out the search (>2 minutes)

PASS:
- `X or Y or Z` works correctly
- `-X X` returns no results
- `X -Y` works
- Have not found any operator order issues

SmallCoccinelle (commented):
From a quick glance, this seems like solid work.

bnkai:
From a quick test everything seems to work fine.

@kermieisinthehouse  1 and 3 from your failing work fine with me so it must be performance related as you suspect 
the `-X or -Y returns the results for X or Y` is supposed to work like `"-X" or "-Y"` according to the first post and it does work fine for me

stg-annon:
I would think that there would be an issue when someone goes to search for something with the `or` keyword in it. To avoid that I would suggest only accepting `OR` or `|` as keywords for the or operation

WithoutPants:
> Failing tests:
> 
> * I cannot get the NOT operator to work by itself, e.g. `-X`, I suspect a performance issue.
> * Any 1-2 character term times out the search (>2 minutes)

I haven't been able to reproduce this, even with my contrived generated database.

bnkai:
Update: After using for a week i dont think i have noticed anything wrong or  a performance related issue. The only thing missing i guess is the help/instructions section

kermieisinthehouse (approved):

silembir:
Woo!! Thanks everyone for the great work!  I'm glad my bounty has gone to good use :)

PR #1983, opened 2021-11-10
Confirm empty stash paths

WithoutPants:
Small change to the setup wizard. If no library paths are selected, then an alert pops up asking the user to confirm before continuing.


kermieisinthehouse (approved):
Tested, looks good!

PR #2007, opened 2021-11-14
Fix trace loglevel

7dJx1qP:
This fixes switching between trace and debug log levels not doing anything, because the trace logs were displaying as debug logs in the `Settings -> Logs` page.


kermieisinthehouse:
This has been an issue for so long, I can't believe it was that easy

kermieisinthehouse (approved):

PR #2015, opened 2021-11-16
Add py3-regex to Docker builds

kermieisinthehouse:
Needed for some py-based scrapers


peolic:
Realistically speaking, I must say the `regex` package is likely not even needed.
The standard library `re` is fully-fledged and from the looks of it `IFeelMyself.py` (the only scraper currently using `regex`), can be adapted to use `re` over `regex` with minimal effort.

PR #2052, opened 2021-11-22
Validate SQLiteDate

### LINKED ISSUE #1976: [Bug Report] No indication given to user when unsupported date format is rejected
**Describe the bug**
If you enter the date for a scene/gallery/performer that is not of the expected "YYYY-MM-DD" format (e.g. if you enter just "YYYY" or "YYYY-MM"), and then save the scene/gallery/performer, you will see the green notification message "Updated scene" or "Updated gallery" indicating that all entered fields have been saved.  There is no indication that there was any problem saving the date value. 
 However, if you look at the scene/gallery details you will see that the date was actually not saved.  Looking at the debug log also shows:

`sqlite date conversion error: ParseDateStringAsFormat failed: dateString <2013>, format <2006-01-02>`

**Expected behavior**
If the value entered into the date field is going to be rejected, there should be some kind of warning/error notification message displayed to the user informing them of this.

v0.10.0-71-gdbfd92f9

WithoutPants:
Fixes #1976 

Changes `SQLiteDate.Value` to return an error if parsing fails, instead of logging a warning.

Also trims any leading and trailing whitespace before parsing the date string.

Tested by trying to set an invalid date on scenes, which now emits an error. Given that this is a low-level function, this may have unintended side effects elsewhere.

### DIFF
--- a/pkg/models/sqlite_date.go
+++ b/pkg/models/sqlite_date.go
@@ -2,9 +2,10 @@ package models
 
 import (
 	"database/sql/driver"
+	"fmt"
+	"strings"
 	"time"
 
-	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/utils"
 )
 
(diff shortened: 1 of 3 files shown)


kermieisinthehouse:
The error message is a little arcane but the fix works

kermieisinthehouse (approved):

PR #2059, opened 2021-11-24
Skip cleaning for search by name scrape queries

bnkai:
Skips postprocessing(mostly dedup) for xpath and json elements when doing a (Performers|Scenes)ByName scrape query
<del>Still a WIP as </del>

1. <del>There is a minor edge case with the htmlquery package we use ( reported it here <https://github.com/antchfx/htmlquery/issues/49> ). For example the scraper from  <https://github.com/stashapp/CommunityScrapers/pull/782> can have a couple of duplicate images missing when searching. JSON scrapers should work fine though like adding a date field in theporndb scraper mentioned here <https://github.com/stashapp/CommunityScrapers/pull/790></del> <https://github.com/stashapp/CommunityScrapers/pull/782> and <https://github.com/stashapp/CommunityScrapers/pull/790> seem to work fine with this PR
2. I am not sure if we want to skip pp in ByFragment queries or somewhere else as well


SmallCoccinelle:
I have a PRish thing which coalesces the two scraper types into one, which should make this PR even easier to pull off: https://github.com/SmallCoccinelle/stash/tree/scraper-refactor-3 ... I'm just waiting for scraper-refactor-2 to be merged before I push that one.

The approach used here would mesh nicely with that branch I think.

bnkai:
> I have a PRish thing which coalesces the two scraper types into one, which should make this PR even easier to pull off: https://github.com/SmallCoccinelle/stash/tree/scraper-refactor-3 ... I'm just waiting for scraper-refactor-2 to be merged before I push that one.
> 
> The approach used here would mesh nicely with that branch I think.

Yeah i had a feeling you would have something in the works :-)
I had a quick look at the scraper-refactor-2 betfore making this PR and tried to do the least conflicting changes i could. I hope it wont have many conflicts when merging

Update: The upstream bug in `htmlquery` was fixed so i vendored the latest version and it seems to work ok

SmallCoccinelle:
The conflicts should mostly go down in count because we don't have to separately handle json and xpath scrapers. They are both instances of "document scraping" in which you somehow download a document, then scrape the contents of it.

PR #1949, opened 2021-11-03
Tasks page refactor

WithoutPants:
Resolves #908

Moves the Scan, Generate and Clean tasks behind dialogs and reorders the tasks page:

image
image
image
image
image

Folder selection is integrated in the dialogs if the selective method was available prior.

Note that the preview generation settings and the overwrite flag have been rolled into the generate dialog so that they are available both on the tasks page and from scene selection.

Todo:
- [x]  save defaults for all tasks

The following should be implemented after this PR is merged:
- add path selection for clean per #1074 
- add path selection for generate
- provide force parameter for transcodes per #1683

PR #2061, opened 2021-11-25
Refactor tasks follow up

WithoutPants:
Moves the tasks page into a separate top-level page with it's own navbar icon.
image

When tasks are running, an animated spinner is shown for the button icon:
image

The tasks page is divided into sections: Library, Data management and Plugins. Library has the scan, clean, autotag, identify and generate tasks. Data management has the import/export, backup and migration tasks.

Following user feedback, the options have been moved out of the dialogs, and the selective tasks buttons have been restored. The system now saves the selected options when running the tasks instead of the user explicitly setting the defaults. Identify now always brings up a dialog.
image

A future iteration will be to show the job queue when mousing over the tasks button.


WithoutPants:
Reverted previous changes. Tasks page is back to a sub-page of the settings page and mostly restored to the way it was prior to #1949. I have moved the animated button functionality into the settings button. It now spins while tasks are running.

image

I will revisit the design of the Task UI at a later date with further consultation with users.

bnkai (approved):

PR #2064, opened 2021-11-25
Symwalk files

SmallCoccinelle:
As per a Discord discussion, we don't seem to be following symlinked files. This patch set should remedy that, by following symlinks for files.

Not tested at all yet, but this is a start.

### DIFF
--- a/pkg/manager/task_scan.go
+++ b/pkg/manager/task_scan.go
@@ -378,6 +378,16 @@ func walkFilesToScan(s *models.StashConfig, f filepath.WalkFunc) error {
 			return nil
 		}
 
+		// If the path is a symlink, follow it through
+		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
+			finalPath, err := filepath.EvalSymlinks(path)
+			if err != nil {
+				logger.Warnf("error expanding symlink %v: %v", path, err)
+			}
+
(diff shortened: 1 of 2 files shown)


bnkai:
If I understand correctly for files that are symlinks you store the expanded path instead.
This has issues:
- We may end up with a path outside the library and thus either the file won't be added to the library or can be removed during a clean from stash
- If a user deletes a scene + file he would then delete the original file and not the symlink
- Filtering/tagging and anything that works with paths will also have issues

I think we should instead make the code that processes those files symlink aware. Eg any generate using ffmpeg should work. Oshash/md5 should be checked. Not sure about phash. It will be a lot more complex to solve but imho that would be worth it

SmallCoccinelle:
Yes, your points are good. It's probably better to push this to various subsystems. Having slept on it, I don't understand why that's not happening right now though. `os.Open` doesn't set `O_NOFOLLOW` or `O_PATH`, so I don't get why it doesn't just opens the contents the link points to in this case.

More investigation is needed in any case.

SmallCoccinelle:
This is definitely not the problem, as I'm wrong on several accounts here. `os.Open(..)` follows symlinks, so no special handling is needed. Closing this, and going back to investigation.

PR #2068, opened 2021-11-26
Fix setting images

SmallCoccinelle:
This correct the mistake from the scraper-refactor-patch where images were not being set correctly. They should have been passed by reference to the setter functions, but wasn't because of a mistake on my part. This has been tested with performers, but I'm pretty sure every other part will work as well, since the fix is entirely mechanical.

### DIFF
--- a/pkg/scraper/image.go
+++ b/pkg/scraper/image.go
@@ -11,7 +11,7 @@ import (
 	"github.com/stashapp/stash/pkg/utils"
 )
 
-func setPerformerImage(ctx context.Context, client *http.Client, p models.ScrapedPerformer, globalConfig GlobalConfig) error {
+func setPerformerImage(ctx context.Context, client *http.Client, p *models.ScrapedPerformer, globalConfig GlobalConfig) error {
 	if p.Image == nil || !strings.HasPrefix(*p.Image, "http") {
 		// nothing to do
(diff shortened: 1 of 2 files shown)


pkg/scraper/image.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/stashapp/stash/pkg/utils"
 )
 
-func setPerformerImage(ctx context.Context, client *http.Client, p models.ScrapedPerformer, globalConfig GlobalConfig) error {
+func setPerformerImage(ctx context.Context, client *http.Client, p *models.ScrapedPerformer, globalConfig GlobalConfig) error {
 	if p.Image == nil || !strings.HasPrefix(*p.Image, "http") {
```
peolic, line 15: This (and in the other functions) used to have a nil check `p == nil || ...`, is it needed here again?

SmallCoccinelle: It depends.

If you call the function from some random part of the code, then having a nil-check would be necessary. However, these functions are only called from a function where the value cannot be nil, which is the underlying reason to make the change in the first place.

A good way to follow up here is likely to get rid of the functions completely. The key operating function is `getImage(..)`, and if you hoist that into the caller, you avoid the pointered type altogether.

PR #2069, opened 2021-11-26
Scraper refactor 3

SmallCoccinelle:
Hack the lower half of the scrapers

### handlerFuncs

Introduce a type for being a handlerFunc, which is modeled after `net/http`'s handlerFunc interface. Use a middleware chain to build up a post process action chain. Using the chain is now a simple funcall.

While here, eliminate a large set of types by using a closure to construct the handlerFuncs.

Note: since we build the chain in reverse, some traversals have been reversed as well as to obtain the same behavior we already had.

### merge of json and xpath scrapers.

Scrapers for JSON and XPath are largely the same, except in a few places:

* How we look up the mapped scraper.
* How we treat a document loaded from an URL.

Introduce `docScraper` as a struct for both with a type switch in it. Introduce `docQueryer` which is the old mapped query interface in new clothes. Retain `xpathQuery` and `jsonQuery` as structs implementing `docQueryer`. This is enough to allow scraping to vary over the kind of document we are querying.

End result:

* One less code path, since json/xpath is now running the same path.
* ~200 lines removed as a consequence.


Gykes:
Closing as stale.

PR #2071, opened 2021-11-27
Fix identify ScrapedScene conversion

Releck:
Minor fix for the identify task which was broken after 4089fcf. Not sure if there's a more idiomatic way of doing this post-refactor.


SmallCoccinelle:
LGTM. Thanks for fixing my bugs.

kermieisinthehouse (approved):

Issue #2049, opened 2021-11-22
[RFC] Request for Submissions: Stash Logo


kermieisinthehouse:
# Stash Logo

## Scope and Requirements
I would like to request candidates for a replacement Stash logo. Our current logo is used inconsistently and is easily confused with other meanings and projects. We want to refine our logo so that we have a single, definitive, high-resolution or vector copy of the logo from which we can automatically derive all other resolutions and formats. 

We shouldn't stray TOO far from the current 'box' concept, and we definitely don't want to convey any adult themes with our logo. Your creations should be AGPL licensed so that they are usable by the project.

It should be something that scales well and is appropriate in a variety of contexts, listed below:

### We have application distribution needs:  
- Windows .exe icon
- Macos .app icon
- Multi-platform taskbar / status bar icon (future use)

|Current macOS | Current Windows|
:----:|:---:|
|![Screenshot from 2021-11-21 17-18-36](https://user-images.githubusercontent.com/84814418/142788364-ee399516-4bea-4f20-82ff-8ff68f36f6b4.png)|![windows-current-exe](https://user-images.githubusercontent.com/84814418/142788367-a45c7c7e-1284-4bae-bdc1-03bb1ab0148f.png)|

### Web app needs:
- favicon
- Single-page application icons (home screen icon, used differently depending on platform)

|Current Favicon|
|---|
|![Screenshot from 2021-11-21 17-18-56](https://user-images.githubusercontent.com/84814418/142788512-d2b8133b-d3aa-41d0-a095-6cc20ab9ab86.png)|

### Community needs:
- Discord server
- [Stashapp](https://stashapp.cc) discovery page
- Any materials we distribute in the future

|Current Discord|
|---|
|![Screenshot from 2021-11-21 17-19-14](https://user-images.githubusercontent.com/84814418/142788554-99800130-a65c-4b99-a74e-48be073483bd.png)|

## Examples
Current Box SVG:
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->
[...]

Issue #1174, opened 2021-03-04
[RFC] Can't open on mac


puddingsoda:
I'm not able to open this on mac. I downloaded the osx release, `cd`'d to its folder, and tried running `./stash-osx`. It says `zsh: permission denied: ./stash-osx`. So I tried using sudo, so `sudo ./stash-osx`. I got the error `sudo: ./stash-osx: command not found`. What should I do now?

bnkai:
Does it have exec permissions? 
`chmod +x stash-osx` should do it, if thats the case.

puddingsoda:
Thank you, my bad, that was the issue.

Issue #980, opened 2020-12-01
[Feature] Stash minimize to tray


Mercury9981:
First of all: you're all doing a wonderful job. I've finally found the perfect way to organize my stash! Thanks a million!

**Is your feature request related to a problem? Please describe.**
I always like to work on a clean desktop and taskbar. Currently the StashApp is an ugly (no offense) app residing in my taskbar.

**Describe the solution you'd like**
A way to minimize the working part of the app to the background, preferrably to the tray

**Describe alternatives you've considered**
I've tried to start StashApp as a service in windows, but this didn't seem to work properly.

bnkai:
I think its a lot easir to run it in the background .... <https://superuser.com/questions/198525/how-can-i-execute-a-windows-command-line-in-background>

WithoutPants:
Closing since there are other ways to do this already, and I think we'd struggle to do this in a platform-independent manner. Let us know if you need further help achieving this.

PR #2073, opened 2021-11-28
Desktop integration

### LINKED ISSUE #594: [Feature] Generic filename and icon (Windows version)
This could be a minor issue, but Windows version of StashApp (stash-win.exe) has the generic icon and the generic filename as seen below:

image

image

Unfortunately, most of malwares, viruses (or whatever you call them) has the same generic icon and basic names. 

If you look at the Task Manager in Windows, you'll see this:

image

This could be confusing for most of end users. If someone uses StashApp rarely and sees this, he/she can think "_what was that stash thing? and what is it doing right now?_"

1) Adding the favicon (https://github.com/stashapp/stash/blob/develop/ui/v2.5/public/favicon.ico) to the exe file.
2) Instead of using a generic name like "stash-win.exe" and an empty description, something like this could be more understandable for PC illiterates like me:

> filename: "StashApp-server.exe"
> description: "Stash is a Go app which organizes and serves your porn."

None.

**Additional context**
None.

kermieisinthehouse:
### ✨ System Tray Integration
(Windows / MacOS) The Stash server now lives in the system tray by default, instead of opening a shell window. This should be less off-putting to nontechnical people, and a little bit of a cleaner experience. This will detect when it should and shouldn't enable, and will keep old behavior if run from a terminal. `./stash` and  `>stash-win.exe` will continue to work exactly the same. The tray icon allows quick access to a few parts of the interface, and allows you to close the server. On the first run of this version, a notification will inform the user of the new location. Adding shortcuts is easy, future state can include a "Scan" button, etc.

<sub>This is not available on Linux because the relevant libraries do not support static linking, and we distribute static executables- it is otherwise functional when dynamically linked.</sub>

| Mac | Windows |
|:---:|:---:|
|Screenshot from 2021-11-27 00-19-11|Screenshot from 2021-11-27 00-58-03|

### ✨ Notifications
(Windows / MacOS / Linux) A desktop notification system has been set up, for sending information to the user. There is an easy opt-out, and as of this patch, it will only notify the user when a task completes. In the future, this could be exposed to plugins / expanded.

Screenshot from 2021-11-27 18-48-59

Capture

### ✨ MacOS Bundling
(MacOS) The build chain has been updated so that Mac executables are shipped as .app bundles, which is the idiomatic way of distributing applications. We gain (sort-of, we have signing issues) one-click launching, decorations, and access to more APIs by shipping like this. No more "chmod" for mac users, just download and run. There is now a single universal bundle for both MacOS architectures.
***Like many of my PRs, this requires a compiler image update: We are missing the `zip` package***

### ✨ Iconography
(Windows / MacOS) The executables are now decorated with an icon! There is a new script that allows updating every instance of the icon across the entire codebase from one master render. Supply it with a 1024x1024 png image to update the Stash logo programmatically in every place it appears. For now I've put on the glowing box concept, but this can be easily changed pending the results of #2049

| Mac | Windows |
|:---:|:---:|
|Screenshot from 2021-11-27 18-19-07|Screenshot from 2021-11-27 00-58-50

This PR has been tested in VMs of all of our supported platforms.

Resolves #594.


kermieisinthehouse:
Closes #594, #980, and forever stops issues like #1174

Some more details:
The tray menu during setup is just the Open/Close buttons, afterwards it uses the UI tabs that are enabled in settings.

Still TODO: 
Notifications do not feature an icon. Most of the APIs take a path to a file, so I'm thinking about (over-)writing an icon to the stash config folder on start, and passing a path to that. Perhaps using /tmp on platforms that have that.

I also want to add a channel to Config.go that alerts when a "large" change has been made, and that the systray has to re-render, e.g. the system setup is finalized, the UI tabs are changed, and (in the future) when an update is available.

kermieisinthehouse:
Notifications have icons now, and it passes the CI pipeline. I have kicked down the road systray updates, as this is buggy on windows and mac. I'll need to hack on the underlying library a little bit more to get that solved. Also, fixed an unrelated ffmpeg download bug on windows.

This is otherwise functional, systray works on mac / windows correctly. Notifications work on all platforms.

kermieisinthehouse:
Known issue: Windows does not have output when started from console. Exploring remediations for this

kermieisinthehouse:
All reported issues resolved.

- Windows shell behavior is now correct
- Console windows do not pop up when executing.
- Merged develop

kermieisinthehouse:
Thanks to @Belleyy for testing.

- Fixed a performance regression in windows ffmpeg (thanks @InfiniteTF )
- Fixed macos broken notifications
- Fixed logging http on windows

pkg/api/server.go
```diff
@@ -245,38 +249,32 @@ func Start(uiBox embed.FS, loginUIBox embed.FS) {
...
-		// This can be done before actually starting the server, as modern browsers will
-		// automatically reload the page if a local port is closed at page load and then opened.
-		if !c.GetNoBrowser() && manager.GetInstance().IsDesktop() {
-			err = browser.OpenURL(displayAddress)
-			if err != nil {
-				logger.Error("Could not open browser: " + err.Error())
-			}
-		}
+	printVersion()
+	printLatestVersion(context.TODO())
```
bnkai, line 253: this was run in a go function so that we dont need to wait for the response/timeout/network error and thus delay the start

pkg/desktop/desktop.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+}
+
+func IsDesktop() bool {
+	// Check if running under root
+	if os.Getuid() == 0 {
+		return false
+	}
+	// Check if stdin is a terminal
+	if term.IsTerminal(int(os.Stdin.Fd())) {
```
bnkai, line 59: Just double checking for this one and linux systems
In the current dev this wasn't checked so stash could open a browser window when run from a terminal
With this applied only if you double click and execute stash from an explorer the browser window and the notifications are working

kermieisinthehouse: That's intended. I got annoyed with Stash taking over my terminal with a `links` window when I was testing a build over ssh

pkg/desktop/desktop.go
```diff
@@ -0,0 +1,156 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+func Shutdown() {
+	database.Close()
+	os.Exit(0)
```
bnkai, line 128: shouldnt we check the error and log it and os.Exit(1)  if its != nil ?

pkg/manager/manager.go
```diff
@@ -458,8 +427,12 @@ func (s *singleton) GetSystemStatus() *models.SystemStatus {
 }
 
 // Shutdown gracefully stops the manager
-func (s *singleton) Shutdown() error {
+func (s *singleton) Shutdown(code int) {
```
bnkai, line 430: Do we really need to force the exit code? 
If everything is ok it should be `0` otherwise `1`
In this case the only thing  we do is close the db, so if the err != nil we should then os.Exit(1) , not only log the error

pkg/api/authentication.go
```diff
@@ -135,9 +135,4 @@ func securityActivateTripwireAccessedFromInternetWithoutAuth(c *config.Instance,
 	if err != nil {
 		logger.Error(err)
 	}
-
-	err = manager.GetInstance().Shutdown()
-	if err != nil {
-		logger.Error(err)
-	}
```
bnkai, line 142: just double checking, shouldnt we shutdown here?

kermieisinthehouse: No, it serves a short message about the misconfigured network. This feature is aimed toward people that don't read logs :)

pkg/manager/manager.go
bnkai: **Not intended for the scope of this PR** 
It seems that the only way to reach the shutdown before this PR was for the user to either ctrl-c in the stash terminal or kill the stash process. With this PR this can be done  easily from the systray. We should find a proper way to shutdown  and wait for the manager (eg context cancelation) so that any running tasks are properly terminated before attempting to close the DB.

bnkai:
Tests ok in linux (debian + xfce as a wm ) ( the notifications :-) )

pkg/manager/manager.go
kermieisinthehouse: Yeah, that's been on the to-do list a while.

I made it take an exit code so that a function that's bailing out can drop a code in. I can probably do an override if the parameter is 0 and it runs into an error closing the database, return a 1 instead.

pkg/desktop/desktop_platform_linux.go
```diff
@@ -0,0 +1,43 @@
...
+
+func hideExecShell(cmd *exec.Cmd) {
+
+}
+
+func sendNotification(notificationTitle string, notificationText string) {
+	err := notificator.New(notificator.Options{
+		AppName:     "Stash",
+		DefaultIcon: getIconPath(),
+	}).Push(notificationTitle, notificationText, "", notificator.UR_NORMAL)
```
bnkai, line 39: If its only for linux and only used here the included `notificator` library only does  the below and could be omitted
```golang
err := exec.Command("notify-send", "-i", getIconPath(), notificationTitle, notificationText, "-a", "Stash").Run()
```

kermieisinthehouse: I bounced around the notification libraries a lot while testing, this might be true. I'll take a peek at this

bnkai: It mentions Kdialog in the readme but it was removed in some point, only `notify-send` is used for linux

pkg/manager/manager.go
bnkai: Yeah that  sounds good.

pkg/desktop/desktop_platform_linux.go
kermieisinthehouse: It seems that's correct, I'll just call it directly then and save some vendor changes. I used that library for all three platforms but had to switch after the problems showed up, but it worked on Linux and it stayed.

pkg/manager/manager.go
kermieisinthehouse: Added a check to see return error code if necessary

pkg/desktop/desktop.go
kermieisinthehouse: Fixed in latest

pkg/api/server.go
kermieisinthehouse: Fixed in latest

pkg/desktop/desktop_platform_linux.go
kermieisinthehouse: Removed the library, thanks for the tip on this

Makefile
```diff
@@ -46,9 +46,13 @@ ifndef OFFICIAL_BUILD
 endif
 
 build: pre-build
+ifdef IS_WIN
+PLATFORM_SPECIFIC_LDFLAGS := -H windowsgui
+endif
```
WithoutPants, line 51: `IS_WIN` is intended for shell behaviours, not for detecting the windows platform. I had to manually include the windows flags in my local version because it didn't detect windows from my environment (which uses git bash).

pkg/desktop/desktop.go
```diff
@@ -0,0 +1,160 @@
...
+func Start() {
+	if IsDesktop() {
+		OpenURLInBrowser(false, "")
+		writeStashIcon()
+		startSystray()
+	}
+}
+
+// OpenURLInBrowser opens a browser to the Stash UI. Path can be an empty string for main page.
+func OpenURLInBrowser(force bool, path string) {
```
WithoutPants, line 30: I don't think this needs to be public.

pkg/desktop/desktop.go
```diff
@@ -0,0 +1,160 @@
...
+	// This can be done before actually starting the server, as modern browsers will
+	// automatically reload the page if a local port is closed at page load and then opened.
+	c := config.GetInstance()
+	if force || (!c.GetNoBrowser() && IsDesktop()) {
```
WithoutPants, line 34: I think the `force` parameter should be removed and the other conditions should be moved into `Start` so that `OpenURLInBrowser` is only called under these conditions.

pkg/manager/config/config.go
```diff
@@ -271,7 +282,22 @@ func (i *Instance) GetNoBrowser() bool {
...
 
+func (i *Instance) GetNotificationsEnabled() bool {
+	return i.getBool(NotificationsEnabled)
+}
+
+func (i *Instance) GetConfigUpdatesChannel() chan int {
+	return i.configUpdates
+}
+
+func (i *Instance) GetShowOneTimeMovedNotification() bool {
```
WithoutPants, line 293: This needs explanation.

pkg/manager/config/config.go
```diff
@@ -271,7 +282,22 @@ func (i *Instance) GetNoBrowser() bool {
 	return i.getBool(NoBrowser)
 }
 
+func (i *Instance) GetNotificationsEnabled() bool {
+	return i.getBool(NotificationsEnabled)
+}
+
+func (i *Instance) GetConfigUpdatesChannel() chan int {
```
WithoutPants, line 289: Is this used? The channel doesn't appear to be used at all.

pkg/desktop/desktop.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func Shutdown() {
+	err := database.Close()
+	if err != nil {
+		logger.Errorf("Error closing database: %s", err)
+		os.Exit(1)
+	}
+	os.Exit(0)
+}
```
WithoutPants, line 133: This is (approximately) a duplicate of `manager.Shutdown` and I don't think it belongs here.

WithoutPants (changes requested):
- running from shell does not output logger or http access messages to the shell (running via git bash)
- the favicon file should be accessible via the embedded `uiBox`. I don't understand why we're converting it into a go file.
- can we please use the existing stash icon until a decision is made to change it?
- there's a few commented out bits of code that either need to be removed or documented.

Makefile
kermieisinthehouse: Fixed by adding a new variable for detecting the platform

pkg/desktop/desktop.go
kermieisinthehouse: Refactored to remove the force option and check before calling

pkg/manager/config/config.go
kermieisinthehouse: Added a comment

pkg/manager/config/config.go
kermieisinthehouse: This is for updating the systray, which isn't working yet. It's a requirement for auto-update, so I've commented it out for now.

pkg/desktop/desktop.go
kermieisinthehouse: This was here because of an import cycle between manager and desktop. I've moved shutdown() from manager to config to avoid the cycle, and the duplicated code.

pkg/desktop/desktop.go
kermieisinthehouse: Made private

kermieisinthehouse:
>  running from shell does not output logger or http access messages to the shell (running via git bash)

Looking into this, the windows shell has been painful. I tested against cmd.exe and powershell, I'm not sure how git bash differs.
>    the favicon file should be accessible via the embedded `uiBox`. I don't understand why we're converting it into a go file.

That is still present and works as usual. It is a go file here so that notifications and the systray can be passed image data directly. Also, Windows and *nixes need the image in different formats. 

>    can we please use the existing stash icon until a decision is made to change it?

Changed. Again, the script included in this can swap it out super easily.

>   there's a few commented out bits of code that either need to be removed or documented.

I've added some more comments. Many of these will get used in the auto-update PR I'm working on, I started concurrently with this.

kermieisinthehouse:
@WithoutPants I'm having no issues with git bash, can you retest on the latest?

Screenshot from 2022-01-07 17-03-17

WithoutPants:
This bug appears to be limited to the git bash terminal within Visual Studio Code. Running stash via git bash externally logs as normal, as does running stash in cmd in Code.

WithoutPants:
I think there's also a potential issue with how this handles startup errors. Without any terminal to log to, if startup fails it just doesn't do anything at all. Probably not a show-stopper but it's a potential issue that we should try to resolve at some point.

pkg/manager/config/init.go
```diff
@@ -159,3 +161,17 @@ func makeOverrideConfig() *viper.Viper {
...
+	// for now, we just close the database.
+	err := database.Close()
+	if err != nil {
+		logger.Errorf("Error closing database: %s", err)
+		if code == 0 {
+			os.Exit(1)
+		}
+	}
+	os.Exit(code)
+}
```
WithoutPants, line 177: This does not belong in `config` - it belongs in `manager`.

To resolve the import cycle, I would suggest adding a new interface in `desktop`:

```
type ShutdownHandler interface {
	Shutdown(code int)
}
```

`desktop.Start` should be changed to accept a `ShutdownHandler` parameter. `manager` can create a small internal type that fulfils this interface by calling `Shutdown`.

WithoutPants (changes requested):

WithoutPants:
> > the favicon file should be accessible via the embedded `uiBox`. I don't understand why we're converting it into a go file.
> 
> That is still present and works as usual. It is a go file here so that notifications and the systray can be passed image data directly. Also, Windows and *nixes need the image in different formats.

Why can't this be passed the image data directly using the embedded data? It seems redundant.

pkg/manager/config/init.go
WithoutPants: I've implemented the suggested changes.

WithoutPants:
> Why can't this be passed the image data directly using the embedded data? It seems redundant.

I've implemented the changes required for this, adding a png file to the ui directory. The generate script is amended so that it no longer generates go files. There is one pending issue with the script. If we're generating from some master png file, that png file should be committed somewhere in the code base, and the script should reference that file directly. I wasn't sure which original source file was used.

kermieisinthehouse:
Oh damn, thanks, I was only halfway done on my branch.

The current logo I used is https://github.com/kermieisinthehouse/stash-graphics/blob/main/stash-plain-box-padded-to-sqaure.png which I've pushed to this branch and re-ran the script.

Changes approved!

PR #2077, opened 2021-11-29
Introduce a CLI framework for Stash

SmallCoccinelle:
Use Cobra to handle the execution of stash. This has a number of immediate consequences:

* Autocompletion
* Ability to have multiple commands: i.e., `stash scan` becomes possible
* Cleaner separation of override configuration from file configuration
* Separation between configuration and manager
* A natural place for log initialization and cpu profiling

Hasn't seen lots of testing yet, and is likely to be rebased on top of Kermie's system tray PR.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -3,6 +3,7 @@ package cmd
 
 import (
 	"embed"
+	"fmt"
 	"net"
 	"os"
 	"os/signal"
@@ -10,6 +11,7 @@ import (
 	"syscall"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"github.com/stashapp/stash/pkg/api"
 	"github.com/stashapp/stash/pkg/logger"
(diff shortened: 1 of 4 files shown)


kermieisinthehouse:
Some thoughts:

How do we handle somebody running `./stash subcommand` when there is already a running instance? Do we detect the running instance and communicate the command to it? What if there are multiple instances running on different ports, for example?

If there is no running instance, should we start the server and queue up the command?
If it's an informational command and no server is running, like `./stash getstats scenes`, should we spin up the server and keep it open, or just return the info and close?

Lots of edge cases to be fleshed out here

SmallCoccinelle:
I was more or less going for the "communicate" path: issue GraphQL commands to an existing instance, fail if it isn't currently available. This makes a command like `stash scan` a client of the server, much like the current front-end is. If there are multiple instances, surely you need to supply a host/port or config option to target the right instance.

My current goal is to support a relatively common use case of mine: copy stash data, then scan. I'm usually sitting in a terminal, and having a way to easily tell stash it can scan would be nice. `pwait(1)`-like functionality would also be nice to have. And it would support all those "periodic scan" requests we get all the time with a good solution.

bnkai:
While having something like cobra handle the cmd line args is ok i think its a lot easier to restrict those args for configuration like we have it now. Imho it is better to have a separate utility ( [transmission-remote](https://linux.die.net/man/1/transmission-remote) is the closest example i can  think of) that performs the actions/commands. That way we dont have to handle any of the issues @kermieisinthehouse  mentioned above and we would only need to implement a graphql client which sends a few specific requests to the stash server instance. This would work well with the autocompletion and the separation between configuration and manager

WithoutPants:
Agree with bnkai, and I think this is stale. Closing.

PR #2079, opened 2021-11-29
Update Tasks documentation

kermieisinthehouse:
It was noted in the chat that this is outdated.

### DIFF
--- a/ui/v2.5/src/docs/en/Tasks.md
+++ b/ui/v2.5/src/docs/en/Tasks.md
@@ -8,7 +8,7 @@ The scan function walks through the stash directories you have configured for ne
 
 Stash currently identifies files by performing a quick file hash. This means that if the file is renamed for moved elsewhere within your configured stash directories, then the scan will detect this and update its database accordingly.
 
-Stash currently ignores duplicate files. If a file is detected with the same hash as a file already in the database (and that file still exists on the filesystem), only the last copy is used.
+Stash currently ignores duplicate files. If two files contain identical content, only the last one it comes across is used.
 
 The "Set name, data, details from metadata" option will parse the files metadata (where supported) and set the scene attributes accordingly. It has previously been noted that this information is frequently incorrect, so only use this option where you are certain that the metadata is correct in the files.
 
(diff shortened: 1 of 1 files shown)


bnkai:
Since you are updating the file could you also add the 
- Perceptual hashes (for deduplication) 
in the `Generated Content` section

and update the `Exporting and Importing section`
with the 
- Full Import
- Import from file 

options? We used to have only the Full import that was destructive. (it must be over a year  :-))

ui/v2.5/src/docs/en/Tasks.md
```diff
@@ -2,15 +2,13 @@
...
-
 # Scanning
 
 The scan function walks through the stash directories you have configured for new and moved files. 
 
-Stash currently identifies files by performing a full MD5 hash on them. This means that if the file is renamed for moved elsewhere within your configured stash directories, then the scan will detect this and update its database accordingly.
+Stash currently identifies files by performing a quick file hash. This means that if the file is renamed for moved elsewhere within your configured stash directories, then the scan will detect this and update its database accordingly.
 
-Stash currently ignores duplicate files. If a file is detected with the same hash as a file already in the database (and that file still exists on the filesystem), then the duplicate file is ignored.
+Stash currently ignores duplicate files. If a file is detected with the same hash as a file already in the database (and that file still exists on the filesystem), only the last copy is used.
```
WithoutPants, line 11: I don't actually understand the new wording.

kermieisinthehouse: Reworded

ui/v2.5/src/docs/en/Tasks.md
```diff
@@ -8,7 +8,7 @@ The scan function walks through the stash directories you have configured for ne
 
 Stash currently identifies files by performing a quick file hash. This means that if the file is renamed for moved elsewhere within your configured stash directories, then the scan will detect this and update its database accordingly.
 
-Stash currently ignores duplicate files. If a file is detected with the same hash as a file already in the database (and that file still exists on the filesystem), only the last copy is used.
+Stash currently ignores duplicate files. If two files contain identical content, only the last one it comes across is used.
```
bnkai, line 11: `If two files contain identical content, only the last one it comes across is used.`
Isnt it the opposite? In case of multiple files with the same hash the first file will be added during scan (if its not in the db) and all the rest will be considered as duplicates and be ignored.

kermieisinthehouse: Had to consult the code, but it seems you are correct. Fixed.

bnkai (approved):

Issue #2076, opened 2021-11-29
[Bug Report] Improving the gallery view performance for large collections


esanchezm:
I have a big collection, including more than 7M images in 14k galleries. I'm having issues with the gallery view as it takes a long time to fetch the images belonging to a gallery. The operation takes around 15 seconds so I get a timeout and nothing shows.

I tried to debug the SQL query and it's basically performing a full scan on the `images` table. That's obviously fine in most of the cases but with a 7M rows table it isn't.

```
explain query plan SELECT DISTINCT images.id FROM images LEFT JOIN galleries_images AS galleries_join ON galleries_join.image_id = images.id LEFT JOIN galleries ON galleries_join.gallery_id = galleries.id WHERE (galleries.id IN (14664)) GROUP BY images.id  HAVING (count(distinct galleries.id) IS 1) ORDER BY images.path ASC LIMIT 40 OFFSET 0;
QUERY PLAN
|--SCAN TABLE images
|--SEARCH TABLE galleries_images AS galleries_join USING INDEX index_galleries_images_on_image_id (image_id=?)
|--SEARCH TABLE galleries USING INTEGER PRIMARY KEY (rowid=?)
|--USE TEMP B-TREE FOR DISTINCT
`--USE TEMP B-TREE FOR ORDER BY
```

So, I think this is something that could be fixed by using `INNER JOIN` instead of `LEFT JOIN`:

```
sqlite> explain query plan SELECT DISTINCT images.id FROM images INNER JOIN galleries_images AS galleries_join ON galleries_join.image_id = images.id INNER JOIN galleries ON galleries_join.gallery_id = galleries.id WHERE (galleries.id IN (14664)) GROUP BY images.id  HAVING (count(distinct galleries.id) IS 1) ORDER BY images.path ASC LIMIT 40 OFFSET 0;
QUERY PLAN
|--SEARCH TABLE galleries USING INTEGER PRIMARY KEY (rowid=?)
|--SEARCH TABLE galleries_images AS galleries_join USING INDEX index_galleries_images_on_gallery_id (gallery_id=?)
|--SEARCH TABLE images USING INTEGER PRIMARY KEY (rowid=?)
|--USE TEMP B-TREE FOR GROUP BY
|--USE TEMP B-TREE FOR DISTINCT
`--USE TEMP B-TREE FOR ORDER BY
```

I suppose there are good reasons behind using a `LEFT JOIN`, but I think this should be fixed so Stash is usable for large/huge collections.

[...]

PR #2083, opened 2021-11-30
Use inner join when getting images in a gallery

esanchezm:
This is an attempt to fix my own issue #2076 that's basically using an `INNER JOIN` instead of a `LEFT JOIN` when fetching images belonging to a gallery.

Please, be gentle with my code as it's the first time I'm writing Go :sweat_smile:  I'll be happy to address your feedback and comments and if you help me writing tests, I'll do it :+1:

### DIFF
--- a/pkg/sqlite/filter.go
+++ b/pkg/sqlite/filter.go
@@ -159,12 +159,12 @@ func (f *filterBuilder) not(n *filterBuilder) {
 	f.subFilterOp = notOp
 }
 
-// addJoin adds a join to the filter. The join is expressed in SQL as:
+// addLeftJoin adds a left join to the filter. The join is expressed in SQL as:
 // LEFT JOIN <table> [AS <as>] ON <onClause>
 // The AS is omitted if as is empty.
 // This method does not add a join if it its alias/table name is already
 // present in another existing join.
-func (f *filterBuilder) addJoin(table, as, onClause string) {
+func (f *filterBuilder) addLeftJoin(table, as, onClause string) {
 	newJoin := join{
 		table:    table,
 		as:       as,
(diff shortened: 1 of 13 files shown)


pkg/sqlite/filter.go
```diff
@@ -164,6 +169,23 @@ func (f *filterBuilder) addJoin(table, as, onClause string) {
 		table:    table,
 		as:       as,
 		onClause: onClause,
+		joinType: "LEFT",
```
7dJx1qP, line 172: Just for naming clarity, I think `addJoin` should be renamed to `addLeftJoin`, because I would expect that a join without a left or right is an inner join by default.

pkg/sqlite/query.go
```diff
@@ -127,6 +127,18 @@ func (qb *queryBuilder) join(table, as, onClause string) {
 		table:    table,
 		as:       as,
 		onClause: onClause,
+		joinType: "LEFT",
```
7dJx1qP, line 130: Same change as above, I would rename `join` to `leftJoin` for clarity.

7dJx1qP (changes requested):

pkg/sqlite/query.go
```diff
@@ -127,6 +127,18 @@ func (qb *queryBuilder) join(table, as, onClause string) {
...
+
+	qb.joins.add(newJoin)
+}
+
+func (qb *queryBuilder) innerJoin(table, as, onClause string) {
+	newJoin := join{
+		table:    table,
+		as:       as,
+		onClause: onClause,
+		joinType: "RIGHT",
```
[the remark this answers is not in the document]
esanchezm: Oh, I'm not really sure why that file got passed, it was part of my _testing_. I'll fix it now

pkg/sqlite/query.go
esanchezm: Same answer too. I'll work on that

esanchezm:
@7dJx1qP all yours :smile:

WithoutPants (approved):
Thanks for your contribution!

kermieisinthehouse:
Wow, galleries load instantaneously for me now!

esanchezm:
It would be great if we could identify other places where `INNER JOIN` should be used :blush:

Issue #1549, opened 2021-07-02
[RFC] Proposal For Configuration Section Refactor


jeremymeyers:
*### Scope*
Rework the config page for better organization.

##  Long Form
As  #1548 is moving scrapers to its own page, I think it would be good to take a look at the configuration section overall.

My suggestions are as follows:

# Separate pages into collapsible sections, expanded by default 
(like this: https://www.codehim.com/bootstrap/bootstrap-vertical-menu-with-submenu-on-click/ )
![Screenshot 2021-07-02 16 57 22](https://user-images.githubusercontent.com/3514095/124327906-a3fefe80-db56-11eb-9277-d6a288870716.png)

# Move stuff around
(in no particular order)

## Content Section
The purpose of the Library section is to show configuration settings related to WHICH CONTENT IS INDEXED OR NOT INDEXED
### Library
- Video Extensions
- Image Extensions
- Gallery Zip Extensions
- Excluded Video Patterns
- Excluded Image/Gallery Patterns
### Tasks
- Auto Tag (Rename "Detect Tags From Filenames")
- Scan (Rename "Add To Library....").  Also rename "Scan" and "Scan Selected" to "Scan All" and "Scan Folders..."
- Generated Content (Rename "Generate Interface Files....")

## System Section
The purpose of the System section is to show configuation options and settings related to BACK-END SETTINGS
### Paths Menu
- Database Path
- Generated Path
- Cache Path

### Logs Menu
- Log File
- Log Level
- Current contents of "Log" page

### Maintenance Menu
- Clean Database
- Backup
- Metadata Export

## Indexing Section
The purpose of the System section is to show configuation options and settings related to THE INGESTION OF CONTENT
- Hashing
- Parallel Scan/Generation
- Preview Generation

## Playback Section
The purpose of the Playback section is to show configuration options and settings related to VIDEO PLAYBACK
### Playback Menu
- Auto-Start Video
- Maximum Loop Duration

### Conversions Menu
- Transcoding options
### DLNA Menu
- Existing DLNA Menu
(Possible future additions: Global playback speed, video player selection, etc)

## Users Section
The purpose of the Users section is to show configuration options and settings related to USER LOGIN
 - (contents of Authentication section)
 - Handy Connection Key

## Appearance Section
The purpose of the APPEARANCE section is to show configuration options and settings related to USER INTERFACE
### Interface Menu
- Language
- Menu Items (Rename "Enable/Disable sections")
- Wall Settings
- Slideshow Delay
[...]

Issue #1173, opened 2021-03-03
[Feature] Add "Edit Config.yml" button to settings area


jeremymeyers:
As there are a bunch of features not surfaced into the configuration menus, a button to launch config.yml in the users default editor would be helpful (probably with a warning about "advanced settings that could break your install".

WithoutPants:
The fact that stash is a webapp makes this not really feasible. There's no guarantee that the client is the same machine as the server (and in many cases won't be). Perhaps an in-app editor would be more suitable (like the custom css editor)?

jeremymeyers:
yeah i was kinda thinking that? though i think at some point that might tie into a discussion about reworking the layout of the settings page because the Configuration tab is messy and not optimally organized

jeremymeyers:
@WithoutPants what about this in an "advanced" submenu of config?

PR #2086, opened 2021-12-01
Settings UI refactor

WithoutPants:
* makes all settings pages auto-save when changing options
* made settings menu fixed to the left edge of the screen and centred content panel
* restructured and reordered pages based on discussion in #1549 
* change checkboxes to switches

image

Still to be done:

- [x] add a saving/saved indicator instead of using Toast notification
- [x] restructure and restyle Tasks page
- [x] readd some missing help links


jeremymeyers:
- love the switch to switches (heh). 
- I still think the left nav should be split up into collapsible (expanded by default) sections because right now its in no particular order (or at least no particular visually distinct order) and makes it hard to find things
- Still feel strongly that the swiches should be "Include Video" "Include Images" rather than "Exclude" ... opt-in system is more natural even if they are "on" by default.  Actually this may be better served by a switch that has "Video" "Images" "Both" as options


Possible Additions
- This is a longshot wishlist item but would it be possible to have the extensions boxes function like how the tag interface functions? (i.e. in little rounded-corner boxes that are created when you press comma)
- Maybe the "About" content could be moved to the help system and the configuration menu thing could be "Updates"

7dJx1qP:
I don't think splitting up the left nav menu is necessary. There's not many settings and I don't see them falling into any clear grouping. Looking at Chrome settings, there's no grouping or clear order to them either (maybe they're ordered by importance/most used). The one section that does exist is an "Advanced" section that is collapsed by default, but I don't think that's needed here, unless you want to shorten the list and put a few of the items like Services, System, Plugins, and Logs under it.

jeremymeyers:
I would put:
## Tasks
- Library
-  Interface
- Metadata Providers

## Tools
- Plugins
- Services

## System
-  Logs
-  Security
-  About and Update

7dJx1qP:
Noticing some weird behavior with the switches. If you flip a lot of them in a short period of time, some of them will flip back and forth in weird ways. The best way to reproduce would be the Interface > Menu items set of switches and just go down the list and toggle the switches. The first few toggles work fine until the first "Updated" toast notification appears. But if you pause for a second and then continue toggling switches and let the toast notifications stack up, some of the previous switches will toggle randomly. It's like the saving can't keep up with the changes so some of the switch states get reverted.

WithoutPants:
> Noticing some weird behavior with the switches...
Thanks for the report. Probably the debounce causing some issues. I'll look into it when I work on the save notification stuff.

7dJx1qP:
The setup wizard redirects to `http://localhost:9999/settings?tab=configuration` which doesn't exist anymore so you just see a blank settings page.

WithoutPants:
Refactored the tasks page (again).

image

I'm not really happy with the way that the collapse buttons displace the task buttons on the right side, so I think there's some more iterating to be done there.

image

The options themselves should be reworded to be more terse and the explanations should be put into the manual.

When changing options, the loading indicator is now shown (at the bottom by default - on top when the navbar is at the bottom) until the configuration is successfully saved, after which is shows a green check mark and fades out - as shown below:
image
image

I have also (hopefully) addressed the issue that caused the state to be broken when editing multiple settings in quick succession.

WithoutPants:
> * I still think the left nav should be split up into collapsible (expanded by default) sections because right now its in no particular order (or at least no particular visually distinct order) and makes it hard to find things

I agree with @7dJx1qP on this, and I don't think the posted category splits are a very good representation of their meanings.   I do think that we could condense it a bit as I agree that there's quite a few menu items there. 

> * Maybe the "About" content could be moved to the help system and the configuration menu thing could be "Updates"

I think the `About` content is better suited at the top of the `System` section. I think the `Security` could probably be rolled into `System` as well. `Tools` should be moved out eventually because it doesn't really belong there at all, but that will be a change for another PR.

> * Still feel strongly that the swiches should be "Include Video" "Include Images" rather than "Exclude" ... opt-in system is more natural even if they are "on" by default.  Actually this may be better served by a switch that has "Video" "Images" "Both" as options

A switch is a bad idea when we have other potential content types in the pipeline (eg audio and text). I'm ambivalent about the include/exclude thing. They are the current way because that's how they are worded in the backend.

Belleyy:
For **Plugin Tasks** in Task:

It look like when you have a plugin with only 1 task, the menu is collapse automatically. 

I would personally prefer that every task need to be collapse manually or all automatically. 
I have many plugins that have 1 task but i don't really use them often, so i kind dislike having them showing instead of a plugin with many task that i use daily being hidden by default.

Best thing would be that Stash remember which need to be show or not, because you have some plugin that you will use once per month and some that you can use daily. But probably worth a new Feature Request/PR.

7dJx1qP:
I think the plugin tasks should all be uncollapsed and visible by default, regardless of whether a plugin has 1 or more tasks.

Stash remembering the collapse state would be nice as well. That feels like something that would be saved to a cookie rather than the config file.

WithoutPants:
* Changed the collapse buttons to use a more minimal up/down chevron. Up/down appears to be the appropriate directions to use according to my UI reading
* Added message when no tasks are running
* Added background color to tasks table
* Added manual links for tasks
* Added help tooltips to scan and generate options
* Made the task queue stay at the top of the screen on larger viewports
* Attempted to make plugins, tools and about pages consistent with the rest of the settings pages

image
image
image
image
image

WithoutPants:
From Discord:

> Regarding the library page:
> 1) The library section is another fine example for a disimprovement. 
> I'm all for spacing, but at this moment my library includes 30 directories (and 22 Excluded Video Patterns).
> 
> Instead of being able to see the list on one single page it would now probably take approx. somewhere between 5 to 8 pages.
> 2) Again, why don't you let me do stuff right away on this very page and force me into an extra step?! You can throw out all Edit buttons, cause all you need is one Save button for the Library section and one for the Media Content Extensions section.
> Regarding the Tasks page: Make Scan, Identify and Auto Tag (just like Generated Content) into their own sections and DO NOT COLLAPSE THEM (by default). These are the most important and most used options in all the Settings. Make them stand out. 
> If you feel it's too large remove the white lines (cause they are visual overkill anyway), make the font smaller, and squish everything just a little bit.

WithoutPants:
Generate dialog page is currently broken looking.

Belleyy:
- Switch are broken, you can't interact with them.
- Would be better if the help button (❓)  go directly to the correct part, clicking Cleaning help should go to Task > Cleaning.
- I think it would be nice to be able to collapse by just pressing anywhere in the card than just a button.

7dJx1qP:
> 2) Again, why don't you let me do stuff right away on this very page and force me into an extra step?! You can throw out all Edit buttons, cause all you need is one Save button for the Library section and one for the Media Content Extensions section.

I disagree with this. I don't think these kinds of settings are something where speed should be valued. Sure it's more annoying for an experienced power-user but I think that's something they'll just have to get used to.

For these settings pages, we should favor accuracy, especially for new users who are not familiar with what they do. Edit buttons for each option that require a free-form user input is good. Compartmentalizing each option, making it require extra steps, and requiring more conscious effort on the user is a benefit, not a con, because it reduces the chance for a user to mess up their settings as opposed to letting them do a bunch of stuff right away and then grouping them all behind a single save.

> Regarding the Tasks page: Make Scan, Identify and Auto Tag (just like Generated Content) into their own sections and DO NOT COLLAPSE THEM (by default). These are the most important and most used options in all the Settings. Make them stand out.

I wouldn't mind seeing what it would look like to have these tasks in their own section. I agree that the options should be open by default.

---

For the tasks page I don't like the scrollbar inside the tasks container. I think the whole page should scroll like the other settings pages. If the idea is to keep the Task Queue section on top always visible when scrolling through the tasks, that could be done by fix positioning it like the left menu.

echo6ix:
> If you feel it's too large remove the white lines (cause they are visual overkill anyway), make the font smaller, and squish everything just a little bit.

I disagree with all his feedback, except for his comment about the white lines. However, I do not think they should be removed because they're consistent with the rest of the design.[^lines] But in my opinion the white divider lines should be the same color as all the other divider lines, so not white.

[^lines]: In my experience it is par for how indented sub-sections with form elements are separated. See [example here](https://i.imgur.com/bPjF0qx.png), where the spell check languages are separated by an indented divider line that is the same colour as the full bleed divider lines.

WithoutPants:
> For the tasks page I don't like the scrollbar inside the tasks container. I think the whole page should scroll like the other settings pages. If the idea is to keep the Task Queue section on top always visible when scrolling through the tasks, that could be done by fix positioning it like the left menu.

I've reverted the change, but I haven't been able to successfully fix it to the top of the page.

WithoutPants:
This is what it looks like when expanded. I think this is a problem. 
image

> I disagree with this. I don't think these kinds of settings are something where speed should be valued. Sure it's more annoying for an experienced power-user but I think that's something they'll just have to get used to.

I disagree pretty strongly with this, and I don't think telling power-users to just get used to it is a sign of good UX design.
 
> ...requiring more conscious effort on the user is a benefit, not a con, because it reduces the chance for a user to mess up their settings as opposed to letting them do a bunch of stuff right away and then grouping them all behind a single save.

I disagree pretty strongly with this as well. Accuracy is important yes, but I don't think it should come at the expense of convenience. I think where it conceptually makes sense to group items we should do so. 

The exclusions section is one that I think should be grouped, allowing a user to modify both exclusion fields in a single operation instead of having to click into multiple dialogs. The extensions can probably stay as is. They aren't likely to be changed very often, and aren't necessarily related to eachother.

I think the library directory editing process needs to be a lot smoother overall, and as a result I think belongs behind a dialog, using a variant of the previous library table.

WithoutPants:
This is what it looks like separated with black borders:
image

7dJx1qP:
> I've reverted the change, but I haven't been able to successfully fix it to the top of the page.

I came up with this to get the task queue to stick where it is as you scroll down. It should probably be in a media query so it doesn't apply during mobile widths though.

```css
    position: sticky;
    top: 3rem;
    z-index: 2;
    background-color: #202b33;
    padding-top: 1rem;
    margin-top: -1rem;
    padding-bottom: 1rem;
```

echo6ix:
> This is what it looks like when expanded. I think this is a problem.

@WithoutPants How about a version like this, which might be a good compromise between old and new:

Library sources are displayed in a "data grid" type view again. Less frequently used elements, like option to delete or edit (or other future options) are hidden in a vertical ellipses on each line.

wwuZqrkJCk9u9UjwYyK9e2

That's how pretty much how TrueNAS Core's dashboard lists all the vdevs

echo6ix:
Btw. Re: Library resources. Wouldn't it be more intuitive to make the toggle switches _include_ video or images than exclude? IDK, maybe it's just me but I always found checking a box, and now a switch, to _exclude_ an item to be counter-intuitive.

ON = yes we want videos/images. OFF = no don't include videos/images.

jeremymeyers:
Why not, for the scan section:
New Wireframe 1

jeremymeyers:
Why not this for the genrate section?

New Wireframe 1

jeremymeyers:
Also, should "Replace Existing Files" be a switch for each individual section?  If you have everything enabled and "replace existing files" enabled and hit "scan", is it going to regenerate the entire database?

WithoutPants:
> Why not this for the genrate section?

I think this will need to be part of another PR, since it will probably want some refactoring on the generate interface as well.

> Also, should "Replace Existing Files" be a switch for each individual section? 

I don't think we really need that degree of control over what to replace or not.

> If you have everything enabled and "replace existing files" enabled and hit "scan", is it going to regenerate the entire database?

That's correct. The "replace existing files" is intended for surgical replacement typically (it was originally only in the generate dialog). It's here for convenience only.

7dJx1qP:
image

The edit/delete menu gets cut off when there's only a single path. Looks like it's coming from one of the containing divs with a `card` class which has an `overflow: hidden;` rule

WithoutPants:
Thanks for the feedback. Disabled styling has been adjusted per the screenshot below:
image

Menu cutoff issue should hopefully now be resolved also.

metaltermite:
@WithoutPants Wooo it is almost there. Final nitpick as far as I can see. For every item that is indented the divider line should also be indented. So, in your final screen shot, the divider lines following those disabled items should be indented.

UncleRoger33:
>  
> 
> image
> 

It would be great if you squeeze the ability to set "Scrapers" directory in the UI, perhaps under the "Tools" category on the left. Like you did in this PR https://github.com/stashapp/stash/pull/1782#issue-1007779560

  


If you use a custom folder and then change the path of that custom folder, you can modify new paths for "Database", "Generated", "Metadata" and "Cache" but not "Scrapers" in (v0.11.0). In my case, i changed my custom path from **D:\stash** to **D:\Portable\Stash** and pointed new paths to `config.yml` from the UI.

Settings1

But i have to manually edit the `config.yml` for new scrapers path, it remembers the original path (**D:\stash**) during the install.

 Settings3

This whole thing is somehow related to https://github.com/stashapp/stash/issues/1173#issue-821378318

WithoutPants:
@UncleRoger33 I've opened #2097 to track this - I won't include it in this PR since it requires backend and graphql schema changes, but I'll get it in for 0.12.

UncleRoger33:
> @UncleRoger33 I've opened #2097 to track this - I won't include it in this PR since it requires backend and graphql schema changes, but I'll get it in for 0.12.

Great, take your time! Rome wasn't built in a day after all :)

kermieisinthehouse:
Some notes:

- Toggles can't be toggled until the green checkmark animation completes, which makes toggling options "rate-limited" and slow.

- This list is half the screen on mobile and pushes the settings below the fold. Perhaps making it expandable, only showing the current selection?
Screenshot_20211209-160643_1_1

- These buttons on the right took me 10 tries to click on mobile
Screenshot_20211209-155405_1

- I can see that these are vertically aligned to the center of the row, but it just makes it inconsistently spaced. The drop down arrows overflow strangely.
Screenshot_20211209-155154_1_1

- This long "preview generation options" string shouldn't be repeated. The button should just have an ellipsis or something. (Btw, you don't need to i18n an ellipsis)
Screenshot_20211209-155228_1

echo6ix:
> (also, word break on the headers)



Considering those are relatively short paths, the only presentation I can think that would fit a table-view like that on a small viewport is to give each "column" a minimum fixed size that prevents the word break; and to implement a horizonal slider above the `Add directory` button and below the last path entry (when the viewport reaches a particular minimum size).

It's probably possible to also use the previous ui iteration for this section explicitly for mobile viewports, but that might be a lot more work than the first suggestion.

If those were the only two solutions I'd probably choose the first.

If someone has a ton of paths it might be annoying to only have one horizonal slider on the bottom of the card versus a horizonal slider below each path entry. I'm not sure which experience would be more favorable though as I can't test thisbuild out.

WithoutPants:
> Some notes:
> 
> * Toggles can't be toggled until the green checkmark animation completes, which makes toggling options "rate-limited" and slow.

An error in the css was causing the loading indicator boundary to stretch to the entire vertical space of the viewport. Should now be fixed.

> * This list is half the screen on mobile and pushes the settings below the fold. Perhaps making it expandable, only showing the current selection?

I'm going to defer this as a problem for another day.

> * These buttons on the right took me 10 tries to click on mobile (also, word break on the headers)

This is related to the first point. Should be fixed now.

> * I can see that these are vertically aligned to the center of the row, but it just makes it inconsistently spaced. The drop down arrows overflow strangely.

Restyled so that it's a little less gross looking.

image


> * This long "preview generation options" string shouldn't be repeated. The button should just have an ellipsis or something. (Btw, you don't need to i18n an ellipsis)

Removed the redundant text and added an ellipsis.

Thanks for the feedback.

kermieisinthehouse:
Just want to confirm these buttons are supposed to look like this:

Screenshot_20211213-171641_1

If they are, that's fine, just confirming content is missing or something. Everything else looks great!

WithoutPants:
> If they are, that's fine, just confirming content is missing or something. Everything else looks great!

No that was a stuff up on my part. Should now be fixed.

kermieisinthehouse (approved):

WithoutPants:
$50 bounty placed for this PR (thanks!)

PR #2091, opened 2021-12-07
Add performer aliases to stash-box tagging/scraping

InfiniteTF:
For some reason it was never added, it seems.


kermieisinthehouse (approved):

bnkai:
Works ok, just a small glitch when displaying a lot of aliases
alias

PR #2100, opened 2021-12-09
Replace JW Player with video.js

### LINKED ISSUE #659: Move from JW Player to open-source video.js?
What do you think of replacing JW Player with the opensource project video.js?

It's open-source, under very active development ([Github repo](https://github.com/videojs/video.js
)), and the licensing is much more permissive.

@WithoutPants mentioned before one issue was the theming (i.e. out of the box, video.js did not look as good as JWPlayer before), however, I checked and they now offer four themes out of the box - they don't look that bad?

https://videojs.com/fantasy

Thoughts?

InfiniteTF:
JW Player is proprietary software that is hard to extend. We're currently relying on hooks we've hacked into the source, which makes upgrading difficult.

video.js is completely open source and has an extensive number of plugins. It doesn't have quite the same feature set as JW out of the box, but everything should be possible to add through plugins, plus further functionality like casting, and possibly VR.

TODO:
- [ ] Support multiple streams with fallbacks.
- [ ] Improve styling.
- [ ] Keyboard shortcuts.
- [ ] Check funscript.

Resolves #659


philpw99:
I wish we can have a choice: JW Player, VideoJS or none. Currently I don't use JW Player at all, and I still love using Stash !

kermieisinthehouse:
Shaping up nicely, just a few things:
- On a transcoded scene, seeking produces an error, perhaps caused by a timeout?
- The "restart video" button at the end of a scene is the same button as the playlist navigation
- don't forget about looping short videos, we have a setting for it

QxxxGit:
> Shaping up nicely, just a few things:
> 
>     * On a transcoded scene, seeking produces an error, perhaps caused by a timeout?

I brought this up in the discord on November 12 in the help channel regarding the JW Player (couldn't get the link to the message on mobile for some reason - sorry).

My error: `error while deregistering ffmpeg stream: exit status 1`

Is that the error you're seeing @kermieisinthehouse?

WithoutPants response
> yeah that's a side-effect of the way the live transcoding code works. When you seek, it fires off a new ffmpeg process from the seek point you clicked. Rejigging the transcode stuff is probably going to be necessary for the switch to videojs (based on my prototyping), so I'll probably try to improve it in the medium term.

kermieisinthehouse:
> Is that the error you're seeing @kermieisinthehouse?

No, that one is expected, and it works fine in jwplayer. I mean that seeking creates an error message on top of the video player about having no content left in the playlist.

ui/v2.5/index.html
```diff
@@ -13,7 +13,7 @@
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
-    <link rel="manifest" href="/%BASE_URL%/manifest.json" />
+    <link rel="manifest" href="manifest.json" />
```
kermieisinthehouse, line 16: I believe this will break accessing the manifest when a subpath proxy is configured. However, if it does, we might just want to get rid of the manifest / meta tags for A2HS. It's not set up right anyway

kermieisinthehouse:
Definitely shaping up! Some things I've noted:
- If a file requires transcoding, it never plays. I haven't found a workaround 
- The autoplay setting is not respected, it always autoplays
- The play bar is a little hard to click on mobile, but this may just be the default styling

JaseNZC:
I compiled this a couple of time on a linux system and both time just got sound and no video. Didn't see any error's so yeah dunno what was happening really.

JaseNZC:
Also no gui at all on Safari even from the home page its all empty.

JaseNZC:
Compiled perfectly and worked just as well.
Awesome work this is a big change.

WithoutPants:
This is almost ready. Interactive functionality is still untested.

I am unable to get the custom controls to appear on iPad. In every case, it shows the native video player. In addition to what is already committed, I have tried adding `html5: { nativeControlsForTouch: false }` to the player options, and I have tried adding `playsInline` and `controls` to the `<video>` element. Neither made any difference. This will be a blocker until we can resolve it.

WithoutPants:
iPad and mobile device display is now fixed. This is ready for proper testing.

kermieisinthehouse (approved):

bnkai:
Everything seems ok. Both issues mentioned above are now resolved.

victorhooi:
This sounds awesome - would there be any way to test this in Develop builds?

Also, would this unlock being able to view 3D videos on normal 2D monitors?

bnkai:
> This sounds awesome - would there be any way to test this in Develop builds?
> 
> Also, would this unlock being able to view 3D videos on normal 2D monitors?

This afaik is just the first step. Some existing functionality is still missing but everything else `seems` ok. No added functionality compared to existing jwplayer implentation. 
If you want to test it you can get the binary from the build artifacts of this PR <https://github.com/stashapp/stash/actions/runs/2049749077>

victorhooi:
This sounds awesome - would there be any way to test this in Develop builds?

Also, would this unlock being able to view 3D videos on normal 2D monitors?

Issue #318, opened 2020-01-14
[Feature] vr/360 video support


iamjen023:
Are there any plans to support vr videos I'm using xbvr for now but would like it to be all in one place

quantumbeep:
Looks like JWPlayer has native VR playback capability of some kind:
https://www.jwplayer.com/developers/web-player-demos/360-video/

Leopere:
That would be a cool value add.  Perhaps in the next major release.

overloadedStasher:
I'm definitely following this issue, as VR Support is a big missing point. 

JWPlayer just deprecated the vr/360 support of it's player, so it's no use.

I found [Video.js](https://videojs.com) that is open source and with much support. It also has a [VR plugin ](https://github.com/videojs/videojs-vr)

I guess migrating to Video.js would be a huge undertaking, but it might help in other areas?

As for supporting VR on Stash itselft, we could have a dropdown list to specify the VR type of the video, in the Scene's edit tab. Depending on the type, it would load the plugin and set the variables accordingly.

I have some experience in TS but not in Go, if some people are willing to help I can try to build a quick demo.

brumouta:
It would be great to integrate with DeoVR player too.

Tweeticoats:
There seems to be support for 360 video support in three.js.
https://threejs.org/examples/?q=vide#webxr_vr_video
https://github.com/mrdoob/three.js/

With some testing I could get stash to play back a video served from stash with the following problems:
- the demo assumes a 360 video, most vr video are 180 so we need to work out how to project on half of the sphere.
- webxr requires ssl so you need to configure ssl and self signed certificate to use the api in you browser

As it is all based on browser technology you just need a <video> tag and the java script to run and accepts the http streaming ie /scene/3030/stream

DampToast:
> It would be great to integrate with DeoVR player too.

https://deovr.com/doc

This would offload playback to DeoVR and for the most part looks like Stash would just have to make a JSON file that could be served to DeoVR with information Stash already has.  It does look like it requires HTTPS though.

yoshnopa:
Just in case someone wants to work with deovr (which would be a great solution in my opinion), here my experiences with the features:

The Web-Player is very easy to implement, when the style and the script are given in the html (both can be integrated or downloaded from sources referenced here: https://deovr.com/app/doc/webvr-video). 
Simply create the video with the element
```
<deo-video>
[...]

PR #2105, opened 2021-12-11
DeoVR Support

kermieisinthehouse:
This creates a button under scenes to open the scene in the DeoVR player, only on Android. This should complement nicely if we ever decide to do in-browser VR with [videojs-vr](https://github.com/videojs/videojs-vr), as we could have a vr button in the embedded video player, and have the json endpoints available for external apps to consume.

TODO:

- [x] Make config option to disable
- [x] Create a root /deovr endpoint so that you can browse your entire library natively

Caveats:

- Requires HTTPS
- Won't support funscript

Closes https://github.com/stashapp/stash/discussions/1735
Closes https://github.com/stashapp/stash/issues/318
Sort-of supercedes https://github.com/Tweeticoats/stash-deovr @Tweeticoats


Releck:
Thanks for working on this, really excited for this feature. I made a small change [here](https://github.com/Releck/stash/commit/4b33fe688b55977bb848c5764ca17793eed742c3) so I could browse scenes from within the player. Thumbnails and hover previews both work.

1. The button from the stash web UI needs stash to run with https but browsing within the player does not.
2. Serving all the scenes to the player is a bit overwhelming. For 18k scenes this meant 4.5MBs of json. DeoVR paginates 12 at a time so this meant navigating over 1,500 pages.
3. By default the deovr button was showing up on scenes. Turning the option on in settings disabled the button.
4. This works in DeoVR but doesn't in HereSpere (separate player that also implements the DeoVR API) because `video_url` uses `deovr://`. [Changing it](https://github.com/stashapp/stash/commit/3cc1d8fbc5d4d2035e915e5f2144fbda27ad2712) to `http://` makes it work in HereSphere as well.

kermieisinthehouse:
I identified the broken JSON but you beat me to the fix! Thanks a bunch, I'll merge this in today.

To my knowledge there's no working around sending the entire library, the API doesn't have pagination. My library comes out to 150MB, 30 seconds load time. I feel that's not unreasonable for a connection that's also expected to stream VR video.

Releck:
Yes you're right there's no way around sending that much data unless it were pre-filtered. Possibly something to tie-in with saved filters but that seems out of scope for this PR.

kermieisinthehouse:
Thanks for the review! Fixes pushed.

Releck:
Re-tested. Button config works as expected. DeoVR and HereSphere work out of the box. Looking good.

HereIam80:
Are you using the DeoVR API or browsing stash via the browser? The API works for me as well but it is mostly useless for me as I have much more 2D stuff than VR stuff in my stash and there is no way of filtering them. If you are using the stash site, have you set up https?

That being said I tried HereSphere and I am impressed. It looks like a very great player by itself. Much better than DeoVR imho.
At least here you can search the DeoVR API by name (and tags if you set it up??).

Releck:
Launching DeoVR via the button works but I haven't configured stash to use https so the browser in DeoVR loads and shows an error screen. Seems like any `deovr://` url is interpreted as https by default and doesn't do anything when launching with HereSphere. I've been testing by navigating to the stash homepage inside the VR player.

With DeoVR navigation is going to be difficult as you'll have to dig through all of your scenes. HereSphere lets you search by title. If `actors` and `categories` were added to `FullDeoScene` that would enable tag searching too.

Caleripo:
I think the best way to solve the lack of filters in DeoVR is to display saved filters in stash as lists in DeoVR. That way people can display only VR videos or specific studios in DeoVR.

kermieisinthehouse:
So, we currently don't have a way of rendering those saved filters server-side. 

I can make lists, but since the user has no input, I can't just send every possible filterable attribute. Ultimately, I can send a lit of performers, a list of tags, a list of studios, etc. Or just keep it as-is and send everything

Releck:
If it isn't possible to use saved filters could there be some opinionated lists? I'd propose a "Favorites" list for scenes rated 5 stars and one for scenes tagged "VR" if the tag exists.

If by list of performers you mean a permutation for every performer I don't think that's going to be viable. There's limited real estate to display lists so only a small amount make sense.

.

Tweeticoats:
Could you add interactive support if there is a script attached to the scene.
To get this to work you need to add the following to the scene json

"isScripted":true,
"fleshlight":[{"title":"something.funscript","url":"http://ip:9999/scene/1676/funscript"}],

kermieisinthehouse:
Sounds good, but where did you find this format? I can't find any docs around this

Tweeticoats:
Copied this from xbvr.
I implemented it in my stash-vr-companion and it was tested by a user.
I don't have a device to test myself.

kermieisinthehouse:
I'm planning on scoping the returned results to a tag called "VR" and all subtags, if one exists.

I'm also planning on adding the funscript changes. All next week.

mmavx:
Could we get a config for serving the `/deovr` route as well? On the DeoVR app for Quest it always navigates to this even when a sub-path (e.g. `/scenes`) was entered in the URL input. I would hate to lose the stash UI for that DeoVR list.
Also it would be great to know if there is any way I can help this PR along, be it testing or coding.

FYI: it seems like DeoVR (at least in their latest update) support the multi-selection way (with `http://something.local/deovr`) now also with non-https URLs. Maybe we can get them to drop this requirement for the deep links as well.

devloch:
If you get this branch merged to be compatible with version 0.15, I can help test it a bit. (Or I can build an earlier stash DB, but Github doesn't have the artifacts anymore from the previous builds).

pickleahead:
@WithoutPants Would like to replace the bounty (https://github.com/stashapp/stash/pull/2708#issuecomment-1269083743) on DeoVR-Support ($ 100). I can open a separate issue if needed. #2384 is close, but not very specific.

WithoutPants:
> @WithoutPants Would like to replace the bounty ([#2708 (comment)](https://github.com/stashapp/stash/pull/2708#issuecomment-1269083743)) on DeoVR-Support ($ 100). I can open a separate issue if needed. #2384 is close, but not very specific.

A new issue might be ideal, just so that we have a concrete set of requirements.

skier233:
What would be required to make this work for windows VR devices (oculus, Valve index, etc)?

kermieisinthehouse:
I'm closing this as stale, but it's a good reference (as far as the structs and JSON tags go) for somebody that wants to finish this up. I never had VR hardware, so I probably wasn't the person to see this through.

PR #2107, opened 2021-12-11
Revert "Use inner join when getting images in a gallery"

kermieisinthehouse:
Reverts stashapp/stash#2083

This introduced two issues: A gallery's "Add" tab excluded all images that did not have an associated gallery, and the image filter "Is missing > Gallery" return no results.

### DIFF
--- a/pkg/sqlite/filter.go
+++ b/pkg/sqlite/filter.go
@@ -39,7 +39,6 @@ type join struct {
 	table    string
 	as       string
 	onClause string
-	joinType string
 }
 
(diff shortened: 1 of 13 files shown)


InfiniteTF:
Couldn't we change the join type back to left join on the affected query rather than reverting the entire PR?

kermieisinthehouse:
Absolutely, I just have no idea which ones they are. We probably have to QA every query here.

kermieisinthehouse:
Closing so I can come up with an alternative workaround

Issue #2114, opened 2021-12-13
[Bug Report] "Error loading image." when tagging performers


DefinitelyNotSpicy:
**Describe the bug**
When trying to tag performers using StashDB it show's "Error loading image." instead of the performer image.

**To Reproduce**
1. Go to 'Performers' and open the 'Tagger" view
2. Click on 'Refresh' on an already tagged performer
3. Wait for the modal update page to open

**Expected behavior**
The image from stashdb should be shown.

**Screenshots**
If applicable, add screenshots to help explain your problem please ensure that your screenshots are SFW or at least appropriately censored.
![image](https://user-images.githubusercontent.com/96064475/145828551-e606aa02-59f7-48bd-8036-4f0ce068636c.png)

**Stash Version: (from Settings -> About):**
v0.11.0-49-g98e836fd

**Desktop (please complete the following information):**
 - OS: Windows 11 x64
 - Browser Firefox Developer Edition
 - Version 96.0b3 (64-bit)

**Additional context**
It looks like Content-Security-Policy introduced in #2108 is the problem here:
![image](https://user-images.githubusercontent.com/96064475/145828932-ce044c89-63c6-4605-afaa-246947d673b4.png)

DefinitelyNotSpicy:
The problem also occurs with tagging scenes with ThePornDB.
`Content Security Policy: The page’s settings blocked the loading of a resource at https://cdn.metadataapi.net/scene/44/29/10/f5218020b9c510b08f21f1dd670badc/background/bg-all-fine-girls-didnt-wait-for-me.jpg (“default-src”).`

![image](https://user-images.githubusercontent.com/96064475/145856106-4705a040-d128-46d0-95ec-e308ac5b89a8.png)

We don't know which domains these images will be shown from, right? They are different for every scraper.

PR #2115, opened 2021-12-13
Fix images loading from stashdb

kermieisinthehouse:
One more security exclusion: We remotely load images from stashdb directly

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -352,7 +352,7 @@ func SecurityHeadersMiddleware(next http.Handler) http.Handler {
 		}
 		connectableOrigins += "; "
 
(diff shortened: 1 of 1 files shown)


kermieisinthehouse:
Fixes #2114

DefinitelyNotSpicy:
Does this work if someone selfhosts stashbox?

InfiniteTF:
I'm not sure whitelisting every potential image source is a great solution. I don't see a big problem with allowing images from arbitrary remote servers. If someone wants to track users they can already do that when stash hits whatever server is hosting the data.

WithoutPants:
Agreed. Made a change to allow any images for now. We can tighten it up later if we need to, but I don't personally see the need.

PR #2119, opened 2021-12-14
Fix image querying

InfiniteTF:
is_missing -> gallery, and the add image to gallery screen currently does not work correctly. This reverts the two inner joins to left joins so images without galleries are returned.


kermieisinthehouse:
Thanks for getting to this!

kermieisinthehouse (approved):

PR #2124, opened 2021-12-15
Support setting scrapers path in UI

### LINKED ISSUE #2097: [Feature] Allow setting scraper directory from UI
(from https://github.com/stashapp/stash/pull/2086#issuecomment-988747633)

It would be great if you squeeze the ability to set "Scrapers" directory in the UI, perhaps under the "Tools" category on the left. Like you did in this PR #1782 (comment)

If you use a custom folder and then change the path of that custom folder, you can modify new paths for "Database", "Generated", "Metadata" and "Cache" but not "Scrapers" in (v0.11.0). In my case, i changed my custom path from D:\stash to D:\Portable\Stash and pointed new paths to config.yml from the UI.

WithoutPants:
Resolves #2097 

Adds setting in System setting page to set scrapers path.


bnkai:
There is an issue when the UI tries to set the scraper path and fails (same goes for the rest of the paths not only for scrapers)
setscrapers

You need to reload the page so that the UI stops trying to set the path

kermieisinthehouse (approved):

WithoutPants:
> There is an issue when the UI tries to set the scraper path and fails (same goes for the rest of the paths not only for scrapers)

Fixed in #2146

PR #2125, opened 2021-12-15
Selective clean

### LINKED ISSUE #1074: [Feature] Selective Clean
Similar to Selective Scan, but for Cleaning. When you know that the only content that needs cleaning is in a specific directory, going through your entire database can be a little time consuming, especially if it's rather large. It gets to be rather unnecessary and I would rather it didn't occur.

WithoutPants:
Adds selective clean button and accompanying path selection dialog.

Resolves #1074.

### DIFF
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -106,6 +106,8 @@ type ScanMetadataOptions {
 }
 
 input CleanMetadataInput {
+  paths: [String!]
+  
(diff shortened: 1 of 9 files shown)


kermieisinthehouse:
Tested, no functional issues.

Putting the warning in the file picker is a little strange, but I'm not sure we can chain modals to show it afterward.

kermieisinthehouse (approved):

PR #2126, opened 2021-12-15
Add force transcode option

### LINKED ISSUE #1683: [Bug Report] Transcode file formats doesn't line up exactly
**Describe the bug**

Stash allows me to generate an MP4 transcode for unsupported file formats to avoid transcoding to webm on the fly (something my GPU-less server can't handle in real-time).

There is at least one file format (it always happens for files from "PureTaboo", for example) where if I ask stash to generate a transcode it refuses (presumably it believes the file format to be supported) but when I try to play the file, it always tries to generate a transcode to webm in realtime.

**To Reproduce**
1. Add a file in this format to the stash library
2. Go to the scene player for that file
3. Try to play it (it generates the transcode on the fly with "[stream] transcoding video file to video/webm" in the log)
4. Click "Generate", select "Transcodes" and "Overwrite"
5. Stash does not generate a transcode (with "Generating 0 sprites 0 previews 0 image previews 0 markers 0 transcodes 0 phashes" in the log)

**Expected behavior**
The transcode is generated and it is used when I try to play the video.

**Stash Version: (from Settings -> About):**  v0.8.0-49-g50cb6a9c

 - OS: Linux (Arch, latest)
 - Browser: Chrome
 - Version: 93.0.4577.58

**Additional context**

Video info from ffprobe of one of the affected files:

```
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '<filename redacted>':
  Metadata:
    major_brand     : isom
    minor_version   : 512
    compatible_brands: isomiso2avc1mp41
    encoder         : Lavf58.29.100
  Duration: 00:40:50.48, start: 0.000000, bitrate: 3467 kb/s
    Stream #0:0(und): Video: h264 (High) (avc1 / 0x31637661), yuv420p(tv, bt709), 1920x1080 [SAR 1:1 DAR 16:9], 3399 kb/s, 23.98 fps, 23.98 tbr, 90k tbn, 47.95 tbc (default)
    Metadata:
      handler_name    : VideoHandler
    Stream #0:1(eng): Audio: aac (HE-AACv2) (mp4a / 0x6134706D), 48000 Hz, stereo, fltp, 62 kb/s (default)
    Metadata:
      handler_name    : SoundHandler
```

WithoutPants:
Adds an option to the Generate dialog to force a transcode to be generated even if the video file appears to be supported for browser streaming.

image

This option is not available in the tasks page, since I believe it to be potentially hazardous and shouldn't be required most of the time.

Also removed the "Set as default" button in the Generate dialog, since it acts contradictorily to the tasks page behaviour (which saves the last options as the default).

Related to #1683


jeremymeyers:
There was a request elsewhere about forcing transcodes for files that are part of a scan but not streaming compatible (and therefore not ingested into the database).  Would that be addressed by this?

WithoutPants:
> There was a request elsewhere about forcing transcodes for files that are part of a scan but not streaming compatible (and therefore not ingested into the database). Would that be addressed by this?

No. This does not change the scan process at all.

bnkai:
Seems to work ok.
Can we change the text though? It seems a bit misleading
I dont know how it would look better maybe a plain 
`Force transcode` 
or add an `even` ?
`Generate Transcodes even for browser-supported videos` 
The help popup explains things but at first glance a lot of users will get confused imho

An extra case to check out, probably outside this scope is when you select the above option then force the encoder to do a proper encode and not try to remux  the file ( eg if the video codec is h264 the video stream will be stream copied).  A plain remux may or may not fix the [related issue](#1683), a full transcode (reencode both streams) has a better chance of doing this

Issue #2141, opened 2021-12-17
[Bug Report] Scene and Image detail pages are missing dates


kermieisinthehouse:
Scene detail page and Image detail page don't make available created_at or updated_at, even though you can filter for them

Reported in Discord

PR #2145, opened 2021-12-19
Creation and update dates visible for images and scenes

meed69:


meed69:
Can't figure out how to link it but this is directly related to #2141

kermieisinthehouse:
We need to change "Created At" and "Updated At" to be i18n strings, and surprisingly it looks like ImageDetailsPanel isn't translated at all! That will probably need to be a separate effort, but we can start on these strings we're adding.

Otherwise works great, 🎉 congrats on your first PR!

WithoutPants (approved):
I added the fields to the gallery page and internationalised the strings. Looks good. Thanks!

PR #2146, opened 2021-12-20
Fix endless toast errors in Settings

WithoutPants:
Fixes issue in the settings pages where if setting failed, then the Toast error messages would flood the page.

Now, if a save operation fails, the error toast is showed once, and the loading indicator changes to a cross and stays on the page until the operation succeeds again (once the user corrects the error).

image

### DIFF
--- a/ui/v2.5/src/components/Settings/context.tsx
+++ b/ui/v2.5/src/components/Settings/context.tsx
@@ -92,9 +92,23 @@ export const SettingsContext: React.FC = ({ children }) => {
   >();
   const [updateDLNAConfig] = useConfigureDLNA();
 
-  const [updateSuccess, setUpdateSuccess] = useState(false);
+  const [updateSuccess, setUpdateSuccess] = useState<boolean | undefined>();
 
   const [apiKey, setApiKey] = useState("");
+
+  // cannot use Toast.error directly with the debounce functions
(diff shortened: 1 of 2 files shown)


kermieisinthehouse (approved):

kermieisinthehouse:
Tested, but not code reviewed.

bnkai:
It seems to work ok.
A minor glitch is that after you update a config path with a faulty one while the value is not updated and fails the UI still has the path shown. You need to refresh the page to see the old data

eg 
update the cache path to an invalid option (/test/) (no access error)
hit confirm
the path shown in the ui is still /test/ although it failed
refresh page and you get the old (correct path)

WithoutPants:
> It seems to work ok. A minor glitch is that after you update a config path with a faulty one while the value is not updated and fails, the UI still has the path shown. You need to refresh the page to see the old data
> 
> eg update the cache path to an invalid option (/test/) (no access error) hit confirm the path shown in the ui is still /test/ although it failed refresh page and you get the old (correct path)

I spent some time changing this around so that it resets the state back to the original, but I think that it results in a worse experience for the user. With the current implementation, it allows the user to correct their mistake, instead of having to re-type in whatever they got wrong with no indication of their previous input.

PR #2150, opened 2021-12-20
Update styling of folder select error state

InfiniteTF:
Moved the error message below the search box so the search box isn't pushed around.

Before:
image

After:
image


bnkai (approved):

kermieisinthehouse (approved):
Reviewed and tested

Issue #2122, opened 2021-12-14
[RFC] Filter Functionality UI/UX Refactor Discussion


jeremymeyers:
# Reworking Filtering
This will be a discussion around making adjustments and improvements to the interface for filtering the content that Stash displays.

## Abstract
It is generally accepted that the current iteration of the Filter UI, while functional, can be confusing and overwhelming.  Any changes would need to accomplish the following things
- Make the selection of filter criteria more visual and intuitive
- Organize the filter criteria in a way that is easy to understand, from a layout and word choice perspective
- Make the ability to save and recall 'favorite' filter combinations more intuitive

Additionally, we should ensure clarity around how the **Filters** functionality interacts visually and in workflow with the **Search** functionality as well as the **Sort** functionality.  We can start by thinking about the default Scenes display as a "Search with no criteria", after which the user can narrow down the display.

My thought is to combine these functionalities (at least in terms of how they are thought of) into one larger "Find And Display" process.  This will require a refactoring of what the search box does and the longer-form interface to build a filter query

## Organization
I suggest this be split into a few sections
- Filtering Look and Feel - How the overall filter interface is displayed, as well as how selected filters look
- Filter Results Look And Feel - How a page with filtered results is displayed, and how changes can be made
- Saved Filters Look And Feel - I think this should be part of a separate PR, honestly.
- Workflow - How one adds a filter criteria to the current view, how one saves filters, how one recalls filters

## Specific Feature Suggestions
My own wishlist to be included as part of the refactor would be
- Using #1545 as a starting point, categorize filters into easier-to-understand groupings
- A cleaner and more intuitive bucket for currently active filter criteria
- Visual differentiators (icons, colors) for filter types to make at-a-glance easier.

## References:
I like the concept of a horizontal filter bar that lives above the content (possibly with an alphabet filter bar below (#1129 ).  See: https://baymard.com/blog/horizontal-filtering-sorting-design

![horizontal-filtering-sorting-design-07-crate-and-barrel-da6026a6f283a3d7a148f2b9569b8b54](https://user-images.githubusercontent.com/3514095/146092516-9a4653f3-fe97-4fbc-8c2f-f84381c88972.jpg)

- Specifically the Crate and Barrel example above spoke to me, since it has 
1) A bunch of different criteria, mostly explicitly displayed with a "More..." button for rarer filter categories
2) A combination of dropdowns with text fields and checkboxes for each category which seems like it would translate well (other than an OR vs AND)
[...]

PR #2172, opened 2021-12-27
Emphasize default filter option

kermieisinthehouse:
"How do I save a sort" may be the #1 most asked question on our support, so this emphasizes the button significantly. I've also switched the "bookmark" for a "floppy" as I feel people will be more likely to click on it.

Before:
Screenshot from 2021-12-27 15-34-38

After:
Screenshot from 2021-12-27 15-30-52


WithoutPants:
I don't really like the aesthetics of either of these changes.

kermieisinthehouse:
Just an experiment, closing. We should figure out a way to make it more discoverable sometime though

WithoutPants:
I'd suggest refactoring this as part of #2122

PR #2173, opened 2021-12-28
Add stash-box credentials validation

InfiniteTF:
image


kermieisinthehouse:
Tested working

bnkai:
Looks and works ok.

Minor issue.
The test button can be clicked even if either of the endpoint/api key input is left blank resulting in the following errors

`No response from server`/`Unknown error: response error: Not authorized`

Issue #2078, opened 2021-11-29
[Feature] Improve and add more classnames to elements in card-sections


echo6ix:
**Is your feature request related to a problem? Please describe.**
Custom CSS styling could be made easier if more elements had class names.

**Describe the solution you'd like**
Add class names to elements in the `card-section`'s.

- make it so all the classes in the card-section can be ordered using the [css order property](https://www.w3schools.com/csSref/css3_pr_order.asp)
- add `card-section-date` class - for dates
- add `card-section-details` class - for scene page descriptions; studio page details; performer page age; gallery page description; etc 
- add `card-section-performers` class - if a user wants to display performer names on the scene page. this class would be `display: none` by default.
- add `card-section-tags` class - if a user wants to display tag names. this class would be `display: none` by default.

Example:

```
<div class="scene-card">
     <div class="video-section">...</div>
     <div class="card-section">
          <h5 class="card-section-title">Scene title goes here</h5>
          <div class="card-section-date">2021-11-29</div>
          <div class="card-section-performers"><a href="..">Performer name</a></div>
          <div class="card-section-details">Scene description here</div>
          <hr>
          <div class="card-popovers">...</div>
     </div>
</div>
```

This would probably break existing custom CSS "themes", but such is the risk with custom CSS injection.

ALonelyJuicebox:
Huge fan of customizability. I think this would be a net improvement.

Issue #1652, opened 2021-08-20
[Feature] [UI] Improve/modernize consistency of formatting in scene detail


echo6ix:
**Is your feature request related to a problem? Please describe.**
The cosmetic formatting of the details lacks consistency: The date section is larger than the other values and is the only value that lacks a heading; the headings that do exist are all above their values, except the resolution which is to the left. 

**Describe the solution you'd like**
Streamline the way the details are presented ala Material Design or the Discord settings UI
- Make each section in the details (date, details, performers, tags, etc) a unique flex item with an [order](https://www.w3schools.com/csSref/css3_pr_order.asp) property. This will be useful for css customization if the user wants to reorder the sections.
- Give each section a heading and it's value with consistent formatting. (See attached pic, specifically the red arrows)
- The title probably doesn't need to be so big but it should still stand as the largest text

**Additional context**
![SPOILER_Screenshot_2021-08-20_015408](https://user-images.githubusercontent.com/37937507/130193953-3d33c22a-a37d-45b5-91d6-051be7d5b281.png)

PR #2188, opened 2021-12-30
Change row orders in SceneDetailPanel.tsx

UncleRoger33:
Moved "**Created At:**" and "**Updated At:**" rows to the bottom. `SceneDetailPanel.tsx` still needs a UI revamp and these two rows made that tab look worse after being added to v0.12.0. 

**Current:**

1_current

**Proposed:**

2_suggested

**Related issues:**
https://github.com/stashapp/stash/issues/1652#issue-975333906
https://github.com/stashapp/stash/issues/2078#issue-1066400001

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneDetails/SceneDetailPanel.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneDetails/SceneDetailPanel.tsx
@@ -114,14 +114,6 @@ export const SceneDetailPanel: React.FC<ISceneDetailProps> = (props) => {
               )}
             </h6>
           )}
-          <h6>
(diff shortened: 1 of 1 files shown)


kermieisinthehouse:
I believe this came up in the discord before it was finalized. I personally don't like it on the bottom and think it looks strange. Perhaps moving it to the file info tab would be a good compromise? Although semantically neither field has much to do with the file

UncleRoger33:
Yep, these two in "File info" tab will look better . I think creation/update dates aren't related to the detail of the scene itself.

Also those two rows doesn't mean anything other than a confusion to the user. There is already a release date on the details tab.

PR #2191, opened 2021-12-31
Fix conversion of multi word stash-box enums

InfiniteTF:
Ensures that `MIDDLE_EASTERN` translates to `Middle Eastern`.


Belleyy:
If the gender is Transgender, the gender no longer appear in the UI
image

InfiniteTF:
Gender translation was relying on the naive string conversion. I've added proper gender enum translation to resolve that.

Issue #2218, opened 2022-01-08
[Feature] Group icons together on performer page


echo6ix:
**Is your feature request related to a problem? Please describe.**
Various performer related icons are presented on the performer page ranging from country flag, gender icon, star ratings, and url related icons. There presentation lacks consistency in sizing and location. There's a lot of wasted horizontal space. Grouping all the icons together adjacent to the name will save a few lines (on mobile views the name and icons can be on separate lines)

**Describe the solution you'd like**
- Standardize the height of all the icons
- Group all the icons together in one location
- Have some logical order to the icons
  1. Interactive icons first (fav icon, rating stars)
  2. Informative icons second (gender)
  3. Link related icons last

Attached image for an example

![d1yTMdHkWMggkiuskuBqGQ](https://user-images.githubusercontent.com/37937507/148626270-f8992a45-2cde-40d9-bbb1-f2b6c647852d.png)

kermieisinthehouse:
I'm working on getting all of the heights lined up, but I'm not sure if they should all be grouped together. They all seem to blend together like that.

PR #2220, opened 2022-01-08
Fix button spacing, fix gender/country height

kermieisinthehouse:
Per Discord feedback, this increases the margin underneath the new performer buttons, and it gives the gender / country flag icons the same size and lines them up. 

This also adds a small tombstone emoji next to the age if a death date is listed. (Existing behavior is to list either current age or age at time of death)

CC #2218 

|Before|After|
|---|---|
|Screenshot from 2022-01-07 22-24-15|Screenshot from 2022-01-07 22-25-14|

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
@@ -259,7 +259,6 @@ const PerformerPage: React.FC<IProps> = ({ performer }) => {
       // provided by the server
       return (
         <div>
-          {performer.death_date ? "🪦 " : ""}
           <span className="age">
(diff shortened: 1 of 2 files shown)


WithoutPants:
Tombstone icon isn't working for me:
image

kermieisinthehouse:
These are emojis, are other emojis working on your system?

WithoutPants:
As far as I can determine, Windows 10 does not support emoji 13.0.

InfiniteTF:
Unfortunately emojis aren't really reliable for this kind of thing due to mixed platform support and rendering differences. Shouldn't be too hard to find a suitable svg though.

kermieisinthehouse:
I've removed the tombstones. A little too macabre anyway

Issue #2224, opened 2022-01-09
[Bug Report] Reverse proxy error


mbhkoay:
Describe the bug
Videos do not load while browsing from reverse proxy url

To Reproduce
1. works as expected while browsing from xxx.xxx.xxx.xxx:port
2. Able to access into stash, but when browsing from scenes, all videos do not load, showing error 224003 
3. Logs show: error writing to UDP socket: write udp 0.0.0.0:1900->239.255.255.250:1900: sendto: required key not available

Stash Version: (from Settings -> About):
v0.12.0

OS: Unraid
Browser: Both Chrome & Edge
Version 0.12

Additional info:
1. Reverse proxy is using Nginx Proxy Manager
2. dangerous_allow_public_without_auth: true
3. security_tripwire_accessed_from_public_internet: ""

![image](https://user-images.githubusercontent.com/29805565/148667983-a6e513e9-0068-4828-83dd-cbfd02344804.png)
![image](https://user-images.githubusercontent.com/29805565/148668011-cdddebb4-ed0f-409b-986c-c966d3584d3b.png)
![image](https://user-images.githubusercontent.com/29805565/148668019-1528a90c-ca1d-4933-aafb-697f3f816d26.png)

kermieisinthehouse:
This is a known issue with nginx proxy manager on a non-standard port. Regular nginx with plain configuration is unaffected. A fix will be proposed soon.

Known temporary workaround, courtesy of GsMumbo:
"Update - reading over the chat and looking at that code, I decided to add a custom location for / and set the X-Forwarded-Port to 443. Mainly because I saw it was one of the values that was set to be ignored in the code."

![IMG_3450](https://user-images.githubusercontent.com/84814418/148668303-6bc9d5d6-0632-410d-ab87-b7676ea17055.png)

mbhkoay:
Thanks kermieisinthehouse! 
This is sufficient for the moment, and has solved the issue for me.

Issue #2211, opened 2022-01-06
Organizr and localhost version Broken by CSP hardening


tamndhzd:
This issue was introduced by the following change: https://github.com/stashapp/stash/pull/2108/files

* setting X-Frame-Options to DENY prevents displaying stash in an iframe, hence it breaks Organizr completely. There should be a user override here in `.stash/config.yml`
* The new Content Security Policy directives seem to prevent images and media from loading when viewed from http://localhost:PORT rather than through an nginx reverse proxy.
* External theming via nginx subfiltering is now broken, again due to CSP changes.

kermieisinthehouse:
>     * setting X-Frame-Options to DENY prevents displaying stash in an iframe, hence it breaks Organizr completely. There should be a user override here in `.stash/config.yml`

I'll probably just remove this completely. A "real" site would send this, but for something self hosted it's probably more of a pain 

>     * The new Content Security Policy directives seem to prevent images and media from loading when viewed from http://localhost:PORT rather than through an nginx reverse proxy.

This is fully intended behavior. You should serve all of your content from the same origin. Removing this would defeat the security guarantees of the CSP.

>     * External theming via nginx subfiltering is now broken, again due to CSP changes.

You can use the built-in custom CSS mechanism for theming, or if your custom themes are extensive, you can load the files into `.stash/custom` and have nginx rewrite to those paths, as they will be served from Stash directly

PR #2228, opened 2022-01-09
Deprecate x-forwarded-port

kermieisinthehouse:
This was never necessary. Host headers from modern browsers include the port number, unless it can be inferred from the protocol (80, 443).

This is causing serious problems with proxy users. At least, any assets that use baseurl (practically all media) are routed around the reverse proxy, and from external networks are inaccessible. 

Fixes about a half dozen affected users on discord, and #2224 #2211

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -385,13 +385,7 @@ func BaseURLMiddleware(next http.Handler) http.Handler {
 		}
 		prefix := getProxyPrefix(r.Header)
 
-		port := ""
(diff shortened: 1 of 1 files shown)


kermieisinthehouse:
An affected user on Discord has confirmed that this fixes the issue. Thanks Ollie!

kermieisinthehouse:
To anyone who was affected by this issue, you should have 

`proxy_set_header Host $http_host;` 

in your nginx.conf, NOT

`proxy_set_header Host $host;`, which was a typo in our documentation.

PR #2229, opened 2022-01-10
Remove trusted proxies

kermieisinthehouse:
Trusted proxies in its current form is near useless, and perhaps encourages bad practices.

It allows a connection from a proxy on the public internet to forward requests into stash. 99% of reverse proxies are running on the same host, and the other 1% are a load balancer or bastion host running on the same network or VPC. I can't figure a use case for this, and I thought the same when I was implementing it. It only came up because of a 'what if' on code review, but I think we should thin it down.


WithoutPants:
This is a nuclear approach. Normally, we'd mark it as deprecated and warn users about the pending removal. How certain are you that this is not being used? I think we're better off giving users the tools they need to tailor security to their needs.

kermieisinthehouse:
I'm fairly convinced that the few people that have set the option don't understand why they did or are doing so needlessly.

Most people set up a reverse proxy like this, where the reverse proxy is on the same network as the stash host:


```
+-----------------------------+
|                             |          +-----------------------------+
|       Public Internet       |          |                             |
|                             |          |        Private Network      |
|                             |          |                             |
|       +-------------+       |          |                             |
|       |             |       |     +----+----+         +---------+    |
|       |             |       |     |    |    |         |         |    |
|       |   Clients   +-------+----->  Proxy  +--------->  Stash  |    |
|       |             |       |     |    |    |         |         |    |
|       |             |       |     +----+----+         +---------+    |
|       +-------------+       |          |                             |
|                             |          |                             |
|                             |          |                             |
|                             |          +-----------------------------+
+-----------------------------+
```

However, the trusted proxies setting is to allow people to create setups like this, where the proxy is accessing stash through the internet. This is an exotic setup, tricky to get right, and requires you to set up firewall rules to allow the proxy to do this:

```
+--------------------------------------------------------------------------+
|       Public Internet                                                    |
|                                                                          |
|                                                                          |
|        +-------------+                                                   |
|        |             |                                                   |
|        |             |                                                   |
|        |   Clients   |                                                   |
|        |             |                                                   |
|        |             |                                                   |
|        +-----+-------+                                                   |   
|              |                                                           |                                       
|              |                     +---------------------------------+   |
|              |                     |   Firewall that only allows     |   |
|              |                     |    connections from Proxy's IP  |   |
|              |                     |                                 |   |
|         +----v----+                |          +---------+            |   |
|         |         |                |          |         |            |   |
|         |  Proxy  +--------------------------->  Stash  |            |   |
|         |         |                |          |         |            |   |
|         +---------+                |          +---------+            |   |
|                                    |                                 |   |
|                                    |                                 |   |
|                                    |                                 |   |
|                                    |                                 |   |
|                                    +---------------------------------+   |
|                                                                          |
+--------------------------------------------------------------------------+
```

kermieisinthehouse:
FYI, if somebody does have this setup, they can simply not pass X-forwarded-for, and retain the old functionality as a workaround

PR #2231, opened 2022-01-11
[WIP] Fix Identify accepting empty results

bnkai:
The identify task accepts empty results from a scraper and thus doesnt move on to the next scraper in line.
I added an extra check of the scraped scene so that if it is empty the next scraper will be used.
Also  when a scraper errors out  just try the next one instead of returning out

Steps to reproduce the original issue/bug
- Select a few scenes to identify ( we need scenes that the first source won't have) and add some sources
sources
- Hit identify
- Even though the first scraper doesnt return any results for some of the scenes it wont continue to the next source

### DIFF
--- a/pkg/identify/identify.go
+++ b/pkg/identify/identify.go
@@ -68,7 +68,7 @@ func (t *SceneIdentifier) scrapeScene(ctx context.Context, scene *models.Scene) 
 		}
 
 		// if results were found then return
-		if isValidScrape(scraped) {
+		if scraped != nil {
 			return &scrapeResult{
 				result: scraped,
 				source: source,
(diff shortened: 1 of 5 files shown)


bnkai:
I removed the 2 scrape error tests from the `identify_test` because of the `when a scraper errors out just try the next one instead of returning out` change. Imho if a scraper source errors out we should just skip to the next not error out. If thats not ok let me know to revert the specific change and the tests.

WithoutPants:
I think that this checking of a valid scrape should be an expected behaviour of the `source.Scraper.ScrapeScene` method. That is, `ScrapeScene` should not be returning empty scenes.

Also, I don't think that the error tests should be removed, but should be rewritten to accomodate the new behaviour.

bnkai:
> I think that this checking of a valid scrape should be an expected behaviour of the `source.Scraper.ScrapeScene` method. That is, `ScrapeScene` should not be returning empty scenes.
> 
> Also, I don't think that the error tests should be removed, but should be rewritten to accomodate the new behaviour.

Moved the validation so that ScrapeScene returns nil if the scrape result is empty (i assume thats what should not be returning empty scenes meant ?)

Tests were adjusted and re-added with the following logic
When a scraper (source) errors out in an identify task the scrape error is ignored so we can skip to the next source. That is an error in scrape doesnt mean an error in identify

bnkai:
Closing. Superceded by  #2375

Issue #2116, opened 2021-12-13
[Feature] Only one step away from full DeoVR support.


philpw99:
**Is your feature request related to a problem? Please describe.**
 Need to have a stream link with file extension like .mp4 or .mkv so DeoVR can play the scene.

**Describe the solution you'd like**
Add a stream link with file extension, like "http://192.168.1.10/scene/123/stream.mp4".

**Describe alternatives you've considered**
A "Launch File" button? Which will open the local file directly, not through the web interface.

**Additional context**
Stash has been my favorite porn organizer since I discovered it. I wrote an app to help me launch the DeoVR player to play a scene I selected from the browser. It works well for me.

The reason I want this feature, because I see right now Stash has come so close to fully support the DeoVR playback. Just a little modification, and DeoVR will be working great with Stash, either with Quest 2 remotely, or with other HMD locally.

Right now in DeoVR, I can browse the movies like this:
![s1](https://user-images.githubusercontent.com/22040708/145867539-80fcf7e4-ad40-41d2-a084-b41869b2d91f.png)

Then view the movie poster like this:
![s2](https://user-images.githubusercontent.com/22040708/145867624-4088a005-19ac-4e89-a428-96541066cc54.png)

Then go in to the scene like this:
![s3](https://user-images.githubusercontent.com/22040708/145867693-b5529b8a-3b3e-45a9-8317-0d64fc153715.png)

It's almost perfect, except the stream link doesn't work yet. ( The file link doesn't work either because web browser cannot access local files.)
DeoVR just need a link with file extension, like "http://192.168.1.10:9999/scene/123/stream.mp4", and it's done. No need to re-encoding or remuxing, just serve the original file with "stream.mp4" or "stream.mkv". DeoVR player will happily take over from there.

philpw99:
Also if we implement this, there is no need for DLNA serving for DeoVR users any more. We can just view and play the files directly from the browser.

kermieisinthehouse:
#2105

philpw99:
> #2105

This is different than the JSON approach. I tried JSON before, and it requires a HTTPS connection. Yet if we use a direct file name, suddenly it works without https.

philpw99:
@Caleripo  I don't know about other VR players. For me DeoVR comes very close to be able to fully functional in its own browser.
Skybox doesn't work with a gamepad or keyboard/mouse and it was abandoned long time ago. So I don't use Skybox. I don't see a web browser in Skybox either.
[...]

PR #2239, opened 2022-01-16
Allow direct file streams and enable opening scene videos to external players in Windows/MacOS

philpw99:
When playing video files by URLs like
`http://192.168.1.10:9999/scene/101/stream.mp4`
The streaming is extremely slow and hosting machine suffers huge performance impact, while using the regular format:
`http://192.168.1.10:9999/scene/101/stream`
Didn't have any impact at all. The first one will cause CPU runs at 100% load and the video still stutters. The latter one just play smoothly and CPU is running at 2%.
Therefore I propose the following changes to improve the performance of streaming, also it enables DeoVR and HereSphere users to be able to watch VR videos directly from Stash:
* For all video files, instead of `http://192.168.1.10:9999/scene/101/stream.xxx` , use 
`http://192.168.1.10:9999/scene/101/stream.org.xxx` to stream the original mp4 file.
* Funscript will also load at address `http://192.168.1.10:9999/scene/101/stream.org.funscript` , but it will be handled by the original funscript routine.

~~If someone makes a mistake and using: `http://localhost:9999/scene/1031/stream.org.mkv` An error page will show in plain text "not a mkv file". This will rarely happen because Stash itself will not offer such a link.~~

This way JWPlayer continues to work, and we can finally directly stream the file with the proper extension. DeoVR, HereSphere and other media players will be greatly benefited from this.

Related Issue #2116 , #1735  ( Yeah I did request this feature like 6 months ago as a newbie. I am very glad that I have learned so much in the past few months. )

**Update: Now you can launch the scene with an external media player in Windows/MacOS/DeoVR/HereSphere.**
For Windows and MacOS, you can only drag the links to the media players to play. There is no other easy route for this. There are some browser plug-ins that allows you to open a media stream directly, but it will constantly monitoring your website activities, also require a "small" program to run in the background, which gives me no confidence for it. Therefore, dragging the video link to the media player seems a reasonable compromise.
For DeoVR player, this link works great! You can open its built-in browser and look around, choose the scene you want to watch, then click on the "Launch / Drag Me" text and the video will play. Funscript should work with this method as well, given you place the .funscript file next to the video file with the same name. DeoVR running in Quest 2 natively should have no problems any more. You don't need the clumsy DLNA. 😃 All you need is a Stash in a docker.
Also graphql scheme is changed to reflect this modification. You can query each scene's original file stream name by running a query:
`{ findScene(id: 101) { paths{ stream_org } }`
The result will be either:
`http://192.168.1.10:9999/scene/101/stream.org.xxx` , and xxx is whatever the extension in the "scene.path" string except funscript.


philpw99:
I have to say, after witnessing all 2900 lines of output for compiling and building processes. I was at awe !
This program is like a state-of-art exhibition ! Truly appreciate everyone's effort to make this gigantic project possible !
Compared to this, my little contribution here today is nothing. It's just a tiny code snippet based on all other people's hard work.

**Kudos for all the programmers/designers/maintainers/contributors in this project !!** 👍

bnkai:
If i understand correctly the original issue was that the /stream route doesnt include an extension and thus cannot be recognised by DeoVR ?
If that is the case then why cant we just have a new `/serve.mp4` for example instead of adding orgs, using that route  with a StreamDirect target? Afaik most players check the headers first once a media file is opened (filename extension doesnt matter)  so there is probably no need for separate mp4,mkv,webm cases. Also the new Stream targets are a copy of the StreamDirect with an extra print and a check, they shouldnt be needed. StreamDirect should be enough.

philpw99:
I am so glad that you asked. At least someone cares.
I don't use `/serve.mp4` approach, because the structure of .mp4 containers and .mkv containers are very different. Yes, most media players will check the header, but I am concerned about the ones that are not smart enough and gives out an error message, telling us we are opening a wrong type instead of just playing it.
I'd rather playing it safe, than presuming every media players are smart and able to correct the problem automatically.

The reason for a check, is that I am worried someone will play with the url and put
`http://192.168.1.10/scene/stream.org.mkv` instead of a `mp4` 
Of course the chance of this is very slim, but I think it doesn't harm to put a little extra check before streaming a 20GB file to the user. If the video file itself was corrupted, or wrongly name, or deleted by accident. We can catch the error before streaming the whole thing.

bnkai:
```
	// Set streamOrgPath by extension.
	streamOrgPath := ""
	ext := filepath.Ext(obj.Path)
	streamOrgPath = builder.GetStreamOrgURL(ext)
```
If you plan to check for media containers anyway then you would need to do it a the place above
The extension could be wrong or modified by the user, there are a lot of `.ts` files named as `.mp4` and some old avi<->mp4
Checking the container there means you will have a correct path from the start ( mkv -> mkv, mp4->mp4, webm->webm) and dont need to recheck that in the StreamDirect so again no need for the StreamOrgDirects just use the routes with the existing StreamDirect target

Edit: you have added a `.vs` directory with this PR probably my mistake?

pkg/api/routes_scene.go
```diff
@@ -28,18 +28,20 @@ func (rs sceneRoutes) Routes() chi.Router {
...
 		r.Get("/stream.ts", rs.StreamTS)
 		r.Get("/stream.mp4", rs.StreamMp4)
-
+		r.Get("/stream.org.mp4", rs.StreamOrgMp4)
 		r.Get("/screenshot", rs.Screenshot)
 		r.Get("/preview", rs.Preview)
 		r.Get("/webp", rs.Webp)
 		r.Get("/vtt/chapter", rs.ChapterVtt)
 		r.Get("/funscript", rs.Funscript)
+		r.Get("/steam.org.funscript", rs.Funscript)
```
bnkai, line 43: typo

bnkai:
> For the path building, I have a different opinion. If the user makes a mistake by changing the extension from ".ts" to ".mp4". That's their fault. We are not the one who gives them the wrong extension. I don't want to slow down the Stash process when showing the scene details, by checking what kind of video file it actually is. Supposed a user open a scene just to see who's the performer. It's not ideal to make Stash call the whole ffmpeg.exe or ffprobe.exe just to get a link of org file.
> 
> On the other hand, "scene.path" is just a variable in the memory. Using it won't slow down the process a bit.

The container is stored in the database along with the path so the above is not an issue
Changing
`streamOrgPath := builder.GetStreamOrgURL(filepath.Ext(obj.Path))`
to
`streamOrgPath := builder.GetStreamOrgURL(obj.Format.String)` for example lets you have the correct container(for example `matroska`,`webm`,`mp4`) from the start and generate the correct path regardless of the extension.
Again imho the extra code (the StreamOrgXXX) isnt needed if you generate the proper path(using the container info) from the start

Another thing i noticed is that your `GetStreamOrgURL` doesn't check/validate at all the paths it creates. You only created routes for mp4,mkv,webm you should only return one of those and for the rest probably return an empty one ?(and hide/not show the icon in the UI ? ). As it is now for say `wmv` files you get a non valid (no route) URL that 404s

Not a UI expert but i think the `Launch Me/Drag Me` text is not needed. The icon is enough imho

philpw99:
@bnkai , I think you are right. Just MP4, MKV and WEBM is not enough. If we check the files, then there are other various video formats that we need to consider, like:

mpegts files: .ts, .mts, .m2ts | mpeg files: .mpg, .mpeg | Quicktime files: .mov, .qt | Real Media : .rm .rmvb ...

There are just too many video formats that we cannot just check one after another. So the method "filepath.Ext(obj.Path)" is actually the only way that we can make sure to get the extension right. In this case, checking the format becomes too much a burden, it doesn't really do us any good. So I switch all the serving back to the direct streams. Like I said, it was an extra mile, but now it seems to be extra 10 miles.

============
I hope someone with UI expertise can tell me what to do. In DeoVR the original button becomes too tiny. It's hard to see and hard to click on it. Plus making the text longer and obvious will prompt user to notice this important feature. They need to know besides using JWPlayer, you can also drag the link to the external players. If they don't see drag, probably in Windows they don't realize that they are NOT supposed to click that button.

PR #2246, opened 2022-01-18
Add Tasks table to Logs Page

kermieisinthehouse:
I often switch between these, as the current task informs what log lines to look for. I figure putting them in the same place just makes sense

Screenshot from 2022-01-18 15-57-59


bnkai:
> I often switch between these, as the current task informs what log lines to look for. I figure putting them in the same place just makes sense
I do the same, till now i had two split tabs just for this.
Looks and tests ok to me

bnkai (approved):

PR #2254, opened 2022-01-22
Fix preview video stopping media on Android

dumdum7:
If you have media playing in the background on Android (eg. Spotify), it stops when playing a scene preview video.
If you have the "Enable sound" setting set to off, this should not happen.

This fixes it by using the HTML \<video\> "muted" attribute when you have Scene / Marker Wall sound disabled in settings.
This also removes the annoying media notification you get every time you play a preview video.


kermieisinthehouse (approved):
LGTM

PR #2257, opened 2022-01-23
Add Several Media Performer Detail Filters, Scene Filters and Sort

kermieisinthehouse:
- Scenes, images, and galleries can now be filtered by "features a favorited performer" (or none)

- Scenes and galleries can now be filtered by performer age at time of production

- There is now a filter in scenes to only show / not show scenes with a phash exact duplicate. I also added a phash sort; when used in combination, they enable you to browse your duplicates side by side and apply additional filters.

No performance issues identified in my testing.


graphql/schema/types/filters.graphql
```diff
@@ -148,6 +150,8 @@ input SceneFilterType {
   tag_count: IntCriterionInput
   """Filter to only include scenes with performers with these tags"""
   performer_tags: HierarchicalMultiCriterionInput
+  """Filter scenes that have perfromers that have been favorited"""
```
bnkai, line 153:
```suggestion
  """Filter scenes that have performers that have been favorited"""
```
typo

graphql/schema/types/filters.graphql
```diff
@@ -243,6 +247,8 @@ input GalleryFilterType {
   performers: MultiCriterionInput
   """Filter by performer count"""
   performer_count: IntCriterionInput
+  """Filter galleries that have perfromers that have been favorited"""
```
bnkai, line 250:
```suggestion
  """Filter galleries that have performers that have been favorited"""
```
same

graphql/schema/types/filters.graphql
```diff
@@ -324,6 +330,8 @@ input ImageFilterType {
   performers: MultiCriterionInput
   """Filter by performer count"""
   performer_count: IntCriterionInput
+  """Filter images that have perfromers that have been favorited"""
```
bnkai, line 333:
```suggestion
  """Filter images that have performers that have been favorited"""
```

bnkai (approved):

kermieisinthehouse:
Added "performer age at time of production" filter to scenes and galleries.

bnkai:
Tests ok, also no performance issues from what i can see.

WithoutPants:
I had an unfinished feature branch to filter on phash duplicates. Can I make the suggestion that `duplicated` is changed from a `boolean` to a first class criterion input?

In my branch, I had a `DuplicateCriterionInput` which contained a `distance` member. This wouldn't work for the negative scenario like yours, but would it be possible to do something similar so that we can add support in the future? My branch was unfinished because for non-zero distances, a temporary table was required to be created. 

The UI can be left as is, and distances > 0 would be unsupported in the backend for now, but at least the graphql schema will be capable of doing this in the future, rather than us having to change the type around.

kermieisinthehouse:
Yeah I can give it its own type with a distance parameter, but actually implementing it sounds like a pain and a half

kermieisinthehouse:
I've killed a few hours trying to untangle how the filter criteria work on the frontend, with no luck. Typescript is not my forte, it seems. 

I've adjusted the graphql and backend code to a bespoke criterion, but left the frontend unchanged, needing an implementation

WithoutPants (approved):

PR #2258, opened 2022-01-23
Reuse static server for multiple requests

vearutop:
Hello, `statigz.FileServer` function indexes and hashes provided file system to optimize performance of serving, it is intended to be initialized before serving requests to avoid extra work for each request.

This PR moves file server setup outside of http handler to improve performance.

Thanks.


kermieisinthehouse (approved):

kermieisinthehouse:
Thanks for popping in!

PR #2265, opened 2022-01-26
Delete funscripts while deleting scene

kermieisinthehouse:
(Reported in Discord)

Deleting a scene file would not delete the associated funscript. This changes that behavior.


bnkai:
Should the UI dialog  be updated also so that the user  is informed before deleting?
```
Delete file
Delete generated supporting files
```
I dont think we need to add an extra check option , just update the `Delete file` to include the Funscript file (if available)?

kermieisinthehouse:
I agree we should change the copy, but any ideas to what? I don't want to make "delete file" too long, maybe "Delete File (+ funscript)"

PR #2275, opened 2022-01-31
Add lightbox preview button to image card

WithoutPants:
Adds a button to the centre of image cards. This button shows the image in the lightbox, similar to the wall view. The button only appears when mousing over it.

image


kermieisinthehouse:
On mobile, this causes taps on the center of the image to open to the lightbox, but taps elsewhere to open to the detail page. Maybe we should put the search icon in the corner, always shown on mobile or something?

WithoutPants:
Changed styling so that touch devices always show the preview button in the bottom right corner, as pictured.

image

kermieisinthehouse (approved):

Issue #1408, opened 2021-05-21
[Feature] Enhance bulk editing of Performers and Scenes to edit every allowed field


hyde231:
**Is your feature request related to a problem? Please describe.**
When importing metadata via fingerprints from stashbox, there might be different terms used than what exists in the local Stash DB. For example, I have already hundreds of performers listed with the country "Russia", however, stashbox uses "Russian Federation". So I end up with different countries. 

**Describe the solution you'd like**
When filtering for performes with country "Russian Federaton", I'd like to be able to bulk edit the country entry to "Russia". The current edit dialogue does not allow to edit other fileds than tags, favorite and rating. I could imagin additional fileds in this bulk edit dialog: A dropdown for the field names, and a text box for the desired value to write into the fields for all selected performes

**Describe alternatives you've considered**
One alternative would be to translate entry values so that "Russian Federation" gets autmatically translated. 
Another solution would be to introduce synonyms so that every search includes "Russia" if "Russian Federation" is selected.
Both solutions would come with a performance impact, so changing these entries witht he bulk edit seems o be much easier.

**Additional context**
This enhanced bulk edit feature would also be useful for other entities, such as scenes, images or galleries.

WithoutPants:
$40 bounty placed.

discourse-stashapp:
This issue has been mentioned on **Stash**. There might be relevant details there:

https://discourse.stashapp.cc/t/request-for-additional-interface-functions/4866/2

Iceshen87:
Hi! 👋 I'd love to work on enhancing the bulk editing feature. I have experience with Go and React development. Please assign this to me!

PR #2282, opened 2022-02-03
Add additional fields to bulk performer dialog

mmavx:
In regards to #1408 this PR adds additional fields to the bulk performer editing dialog.

## Questions
- [x] Some of the fields seem nonsensical to edit in bulk, e.g. URLs, dates. Are these to be changeable just for the consistency or should the dialog be limited to "useful" fields?
- [x] Is it necessary to prefill the fields with the common/undefined value as with favorite or tags? AFAICT this would mean having to fetch this data from all selected performers as it is not included in the `SlimPerformer` set. 

## ToDos
- [x] Gender field
- [x] ~Weight field~

## Preview
image

### DIFF
--- a/ui/v2.5/src/components/Performers/EditPerformersDialog.tsx
+++ b/ui/v2.5/src/components/Performers/EditPerformersDialog.tsx
@@ -9,10 +9,7 @@ import { useToast } from "src/hooks";
 import { FormUtils } from "src/utils";
 import MultiSet from "../Shared/MultiSet";
 import { RatingStars } from "../Scenes/SceneDetails/RatingStars";
-import {
-  genderStrings,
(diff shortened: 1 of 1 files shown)


kermieisinthehouse:
I'd argue that height, measurements, tattoos, and piercings are individual enough that they don't make sense to bulk edit. To answer your question, URL and dates too. 

I think autofilling is more of a wishlist item. We should probably have an endpoint return the tag intersection for the remove tags field at some point, at least.

mmavx:
For tattoos and piercings one could image a yes/no value where bulk edit might come in  handy, for the others I agree and will remove them

kermieisinthehouse (approved):

Issue #1819, opened 2021-10-09
[Bug Report] Saved filter does not URL decode values when used


7dJx1qP:
**Describe the bug**
Saved filters with values that have URL encoded characters don't get URL decoded when the saved filter is used. This causes the filter to use a different value than the one it was saved with.

**To Reproduce**
1. Go to Scenes
2. Save a "Title includes +" filter
3. Select the saved filter
4. The filter will display "Title includes %2B" and scenes will be filtered by 2b rather than +

**Expected behavior**
Loading a saved filter should show and use the same values that it was saved with.

**Screenshots**
Original filter ![image](https://user-images.githubusercontent.com/38586902/136665570-e5a34ec0-d442-4fc3-95f3-43be49aef07b.png)
Saved filter ![image](https://user-images.githubusercontent.com/38586902/136665580-32c54a70-c916-44ab-90f6-a2da72a955d4.png)

**Stash Version: (from Settings -> About):**
Version: | v0.9.0
Build hash: | a73c99a6

**Additional context**
Refreshing the page applies the filter correctly and shows the correct value so the problem is just with the initial load of the saved filter.

kermieisinthehouse:
Can reproduce

kermieisinthehouse:
This looks like two separate bugs, but I haven't completely investigated yet. Deserializing the filters from the url (Models/list-fitler/Filters.tsx) is parsed differently than deserializing them from a saved query (components/List/SavedFilterList.tsx), but both are incorrect. 

This also caught my eye:
```
public makeQueryParameters(): string {
    return queryString.stringify(this.getQueryParameters(), { encode: false });
  }
```

DampToast:
I have a fix for this locally and will submit it tomorrow via a PR.

PR #2301, opened 2022-02-08
Make changes to fix special characters in Criterion labels (#1819)

DampToast:
Reverse the '&' and '+' replacement done on StringCriterion
Decodes special characters in IHierarchicalLabeledIdCriterion

Issue #1819 

209-212 just undo what was done during encodeValue().  The downside is that if somebody has %2B in their string it will decode that to a '+' instead of showing '%2B' if that was desired.  Likewise if they have an encoded '&' that they purposefully want.  But that seems rare. This only affects the display of this data.

I also noticed that a more widespread issue was happening with the IHierarchicalLabeledIdCriterion where it was URI encoded in the database but not decoded for display.  So you would see strings like "Bob%20Ross" doing similar steps but with something like filtering by studio.  This fixes that as well.  Again it only alters the display value, nothing in the URL or data passed around for the application.

### DIFF
--- a/ui/v2.5/src/models/list-filter/criteria/criterion.ts
+++ b/ui/v2.5/src/models/list-filter/criteria/criterion.ts
@@ -206,7 +206,10 @@ export function createStringCriterionOption(
 
 export class StringCriterion extends Criterion<string> {
   public getLabelValue() {
-    return this.value;
+    let ret = this.value;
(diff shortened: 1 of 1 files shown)


kermieisinthehouse (approved):
Code reviewed and tested, approved.

Thank you for your first contribution!

kermieisinthehouse (approved):

PR #2309, opened 2022-02-10
Allow custom UI theme with config file

Teda1:
Currently, browsers that support menu bar colors use the default Stashblue color. When users modify the styling with their own CSS the same blue Stashmenu bar remains present which gets in the way of custom styling. This pr introduces a `ui.properties` file which at the moment can be used to defined a custom theme color in place of the default Stashblue color. The `ui.properties` file will currently be created in the Stash config directory. The hard coded #202b33 color was moved to that file so it can be overridden. In the future, I imagine this file could be used for other things such as allowing users to provide their own mobile icon or favcon for Stash.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -58,11 +58,6 @@ func Start(uiBox embed.FS, loginUIBox embed.FS) {
 		})
 		r.Use(httplog.RequestLogger(httpLogger))
 	}
-	uiPropPath := c.GetUIPropertiesPath()
-	exists, _ := utils.FileExists(uiPropPath)
-	if !exists {
-		c.CreateUIProperties()
-	}
 	r.Use(SecurityHeadersMiddleware)
 	r.Use(middleware.DefaultCompress)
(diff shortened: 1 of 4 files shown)


Teda1:
Before:
IMG-1229

After:
IMG-1255

WithoutPants:
This needs an explanation and probably some documentation in the manual.

Teda1:
I went ahead and added a description explaining the change. Sorry I assumed Kermie would be looking at this. The pr was created as a follow-up to a discussion we had in the discord channel. I'll need to look into how to update the Stash manual.

Teda1:
Seems like updating the Wiki should be pretty straightforward. It probably makes sense to update the wiki after this change has been accepted.

WithoutPants:
Do we really need to introduce an entirely new properties file for this? Can this not be put into the `config.yml` file?

Teda1:
No problem

Teda1:
Moved my changes to https://github.com/stashapp/stash/pull/2365

PR #2317, opened 2022-02-15
Scan now uses exclude regex on gallery zips

### LINKED ISSUE #2308: [Bug Report] "Excluded Image/Gallery Patterns" does not exclude files in zips
**Describe the bug**
If you keep all your galleries in zips, it appears that the Excluded Image/Gallery Patterns regexs only look at the zip filename and not the filenames inside the zips.

**To Reproduce**
1.  Create two images, one called willshow-demo.jpg and wontshow-demo.jpg.  Make sure they are different.  Put willshow-demo.jpg inside a zip.
2. Add a regex to ignore "-demo.jpg" such as    ^.*-demo\\.jpg$
3.  Scan the folder that has both of these files.
4. willshow-demo.jpg will appear as it is inside a zip, wontshow-demo.jpg will not show up.

**Expected behavior**
I would expect filenames within a zip to be ignored as well.

v0.12.0-43-g47670d1e

**Additional context**

DampToast:
Fixes #2308 

Previously scanning was not using the regex exclusions during a scan.  Now when it walks the gallery hierarchy it uses the regex exclusions.  It constructs the filename as it does everywhere else with the full zip path, null character, then the file name in the zip.

I added 2 imports to bring in the config to get the exclusions, and the file import to construct the zip name.
On line 19 I get the image exclusions
On line 34 I do the regex check

The other changes are due to having to rename a variable from 'file' to 'f', although that might just be my complete lack of knowledge of Go.  Otherwise I couldn't use file.ZipFile.


DampToast:
My commit is so bad I broke Docker's will to live.

Issue #2288, opened 2022-02-05
[Bug Report] Performer images wall view broken (pre-release)


QxxxGit:
There's something funky going on with the wall view for performer images. It's the similar issue of clicking any of the images opens up the first image, and for some reason it won't let me switch back to grid view either. I can't quite nail it down as to what causes it because it happens frequently and rarely it doesn't happen.

In my case, if I go Performers -> add a filter -> select performer -> images tab -> wall view, and select one of the images it only opens up the first, and it also won't allow me to go back to grid view.

In another user's case, they raised that it only happened to them if the performer had more than one page of images. I haven't tested this yet.

kermieisinthehouse:
I can reproduce what I believe is the same issue.

Performer Detail Page > Images Tab > Attempt to set a filter locks ups up the main thread in both chrome and firefox for me.

QxxxGit:
Did more testing and found that this issue also occurs on grid view, and I'll verify it only seems to happen when there are more than 1 page of images.

kermieisinthehouse:
Quoting you:

* Performer Detail Page -> images tab -> on either grid/wall view, clicking an image that isn't the first image to open the lightbox (either by the magnifying glass on grid view or by clicking the image on wall view) and there are more than 1 page of images, the first image opens in the lightbox instead of the one you clicked on.

* Performer Detail Page -> images tab ->  clicking saved filters goes to a "Something went wrong" error page with a link for reactjs error decoder # 185 - maximum update depth exceeded. this can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

* Performer Detail Page -> images tab -> attempt to set filter. When I select a filter from the dropdown, it's like the state doesnt change so it's impossible to actually add a filter

PR #2321, opened 2022-02-16
Attempted spinlock fix for Images view

WithoutPants:
Attempted fix for #2288 

I haven't been able to reproduce the issues locally, but I have seem the "maximum update depth exceeded" error message.

There is an underlying problem somewhere where the `filter` parameter keeps getting updated, but it's a bit of a mess upstream, so for now I'm attempting to address it here. Needs testing.

### DIFF
--- a/ui/v2.5/src/components/Dialogs/SubmitDraft.tsx
+++ b/ui/v2.5/src/components/Dialogs/SubmitDraft.tsx
@@ -59,8 +59,6 @@ export const SubmitStashBoxDraft: React.FC<IProps> = ({
   const handleSelectBox = (e: React.ChangeEvent<HTMLSelectElement>) =>
     setSelectedBox(Number.parseInt(e.currentTarget.value) ?? 0);
 
-  console.log(data);
-
   return (
     <Modal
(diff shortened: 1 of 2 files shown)


bnkai:
Tested against dev version 0.12.54
The following from #2288
-    Performer Detail Page -> images tab -> on either grid/wall view, clicking an image that isn't the first image to open the lightbox (either by the magnifying glass on grid view or by clicking the image on wall view) and there are more than 1 page of images, the first image opens in the lightbox instead of the one you clicked on.

 -   Performer Detail Page -> images tab -> clicking saved filters goes to a "Something went wrong" error page with a link for reactjs error decoder # 185 - maximum update depth exceeded. this can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

  -  Performer Detail Page -> images tab -> attempt to set filter. When I select a filter from the dropdown, it's like the state doesnt change so it's impossible to actually add a filter

were reprooduced in dev version and seem fixed in this one (using Firefox/Chrome in Linux)

PR #2322, opened 2022-02-17
Overhaul HLS streaming

WithoutPants:
Introduces a stream manager to the `ffmpeg` package which manages HLS streaming. It is intended to be a replacement of the live transcoding code we have, ~~but we will not be able to do that until we have a video player that supports HLS (ie videojs). For now, the new HLS implementation will only be exercised on iOS devices.~~ 

The HLS implementation works by generating and serving chunk files using the `cache` directory. This makes the `cache` directory mandatory for HLS streaming (and eventually live transcoding in general). If a chunk is requested that is ahead of the currently generated chunks, then the ffmpeg transcoded is stopped and restarted at the requested time. If the transcoding process gets far enough ahead of the requested chunks, then the process is stopped until a missing chunk is requested again. If no chunks for a video are requested for ten seconds, then all chunks are deleted from the cache.

~~This should be considered a prerequisite for #2100. The duration hacks will no longer be necessary when using HLS and we can remove the other live transcoding implementation.~~


bnkai:
Code looks ok. I have no way of testing it though.
Should it be merged so that ios users can test it?

WithoutPants:
Deferring until ffmpeg refactor and videojs PRs are merged.

WithoutPants:
This has been picked up by others, so I think it can be safely closed.

PR #2323, opened 2022-02-17
Performer tagger modal will load country name instead of code

### LINKED ISSUE #1567: [Bug Report] Batch Scraped performer Country data from stash-box is only two-lettered
**Describe the bug**
When the data for performers is scraped from stashbox, it sets country names with only two letters (US,CO,CA, etc.) instead of United States, Colombia, Canada, and so on. This results in stash not recognizing the country name of the performer.

**To Reproduce**
1. Go to Performers tab
2. Click on Tagger button
3. Click on Batch Add Performers button
4. Enter performer names and click on Add Performers
4. The scraped data for the new performers will contain the Country names in short forms, which messes up stash's country recognition

**Expected behavior**
The country names fetched should be expanded, not in their short forms

**Stash Version: (from Settings -> About):** v0.8.0

 - OS: Windows 10 Pro
 - Browser: Microsoft Edge (Chromium)
 - Version: 91.0.864.64 (Official build) (64-bit)
 
**Additional context**
If the data for performers is scraped from their individual page, then the country names are added properly.

DampToast:
When using the performer tagger, it would pull the 2 character country code from StashDB and store that.  It now shows the full country name and stores the full country name instead.  Also verified that it causes no issue with a performer that has no country set on StashDB.

### DIFF
--- a/ui/v2.5/src/components/Tagger/PerformerModal.tsx
+++ b/ui/v2.5/src/components/Tagger/PerformerModal.tsx
@@ -12,7 +12,7 @@ import {
 } from "src/components/Shared";
 import * as GQL from "src/core/generated-graphql";
 import { stringToGender } from "src/utils/gender";
-import country, { getCountryByISO } from "src/utils/country";
+import { getCountryByISO } from "src/utils/country";
 
 interface IPerformerModalProps {
   performer: GQL.ScrapedScenePerformerDataFragment;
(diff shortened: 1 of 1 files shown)


WithoutPants:
This fixes the issue when using the tagger view for a single performer, but it doesn't address the batch performer scrape operation as described in #1567. This is still good and ready to merge, but I'll unlink it from the issue.

DampToast:
Makes sense.  I'll see if I can fix that issue as well tomorrow.

PR #2326, opened 2022-02-17
Fix countries with batch performer add/update

### LINKED ISSUE #1567: [Bug Report] Batch Scraped performer Country data from stash-box is only two-lettered

DampToast:
I had to add a new dependency. I don't know anything about GoLang so hopefully this is the right way.  Also the license looks okay but of course double check to see if it works.  https://github.com/biter777/countries

This should fix #1567.  When doing a batch performer add or update, it converts the 2 character code to the full country name and stores it that way.  I also tested this with a performer that had no country in StashDB to ensure it didn't cause issues.

This also fixes another issue I found which was assigning the "Tattoos" data from StashDB to the Twitter field.  I figured it would be okay to include with this PR though.


WithoutPants:
This is going to collide/be obsoleted by #1922.

DampToast:
Ok.  Do you want me to close this and resubmit just the fix for Tattoos going into the Twitter field?

WithoutPants:
Obsoleted by #1922.

PR #2335, opened 2022-02-20
Add filesize on scene cards, invisibly

kermieisinthehouse:
This puts filesize on the scene overlay, next to resolution. I find this extremely useful, but the popular sentiment on the discord was low at best, so this is invisible by default. 

To actually use it, you have to add to your custom CSS:

```
.overlay-filesize {
 display: inline
}
```


bnkai (approved):

kermieisinthehouse:
I changed the classname to be .extra-scene-info to adjust

PR #2336, opened 2022-02-21
Handle unicode characters in autotag

### LINKED ISSUE #2293: [Bug Report] Non-ASCII performers fail to be tagged
**Describe the bug**
Non-ASCII performers appear to invalid to the auto taggers and cause the tagger to ignore the paths - specifically Japanese/Chinese in my case. Other non-ASCII characters were not tested.

**To Reproduce**
1. Create a directory containing non-ASCII character of the performer's name and move a valid media file to the directory.
2. Create a new performer with the non-ASCII character (e.g., `伏字`).
3. Enqueue auto-tagging tasks.

**Expected behavior**
The media within the performer's directory should be tagged with the performer's metadata, but it is not. 
e.g., `/media/伏字/hello-world.mp4` should be tagged with the performer `伏字` that was just created.

v0.12.0-40-gc857e301

 - OS: Both Windows and Linux

WithoutPants:
Fixes #2293 

Adds `\p{L}` to regex strings to handle unicode letter characters. Fixes `getPathWords` to correctly split words with unicode characters in them.


pkg/sqlite/tag.go
```diff
@@ -236,7 +236,6 @@ func (qb *tagQueryBuilder) QueryForAutoTag(words []string) ([]*models.Tag, error
 	var args []interface{}
 
 	// always include names that begin with a single character
-	singleFirstCharacterRegex := "^[\\w][.\\-_ ]"
```
kermieisinthehouse, line 239: Why remove these?

WithoutPants: This is replaced with `const singleFirstCharacterRegex = `^[\w\p{L}][.\-_ ]` in `performer.go`.

kermieisinthehouse (approved):

bnkai:
For me this build seems to cause a big impact in performance
- Create a new tag
- Save
- Hit autotag (from the  newly created tag not the task)
Repeat the same for performers

Dev version 3-5 secs
This Build 35 secs

kermieisinthehouse:
Maybe we can determine if any of the characters in the tag fall outside of ASCII, then apply the special regex?

bnkai:
Even if unicode characters are in the tag/performer/.... they get matched fine in the dev build (tried with a japanese name). I think the issue was if they are mixed with some plain characters in the path splitting code (betwen delimiters) . Not sure how we can optimize that though. 
```
Maybe we can determine if any of the characters in the tag fall outside of ASCII, then apply the special regex?
```
This could be a  good starting point i guess?

WithoutPants:
I've reverted the change in `getPathQueryRegex`. It means that more scenes/images/galleries will be returned in situations where the object name is bounded by unicode letters. For example, for performer `伏字`, a file named `/字伏字/foo.mp4` will be returned. However, this will be post-filtered out in the call to `nameMatchesPath`, which has the `\p{L}` included in its regex. The query performance improvement ought to outweight the performance hit of dealing with the extra results. Will need a retest though.

bnkai:
> I've reverted the change in `getPathQueryRegex`. It means that more scenes/images/galleries will be returned in situations where the object name is bounded by unicode letters. For example, for performer `伏字`, a file named `/字伏字/foo.mp4` will be returned. However, this will be post-filtered out in the call to `nameMatchesPath`, which has the `\p{L}` included in its regex. The query performance improvement ought to outweight the performance hit of dealing with the extra results. Will need a retest though.

Seems ok to me.
Repeating the tests with the latest changes gives the same running time with current dev, so no impact in peformance

WithoutPants:
> For example the below 2 files `/path/to/字伏字/foo.mp4`,`/path/to/foo/字伏字.mp4` Creating a performer/tag named `字伏字` and then autotagging from there only matches the second file The second file `/path/to/foo/字伏字.mp4` gets matched/tagged even like this `/path/to/foo/test_字伏字_720p.mp4` while the first not at all (the same thing happens in the latest dev version)

I haven't been able to reproduce this. I tried both filename variants and auto tagging from the performer/tag and tasks pages. Scenes were successfully tagged in all scenarios.

bnkai (approved):

bnkai:
> > For example the below 2 files `/path/to/字伏字/foo.mp4`,`/path/to/foo/字伏字.mp4` Creating a performer/tag named `字伏字` and then autotagging from there only matches the second file The second file `/path/to/foo/字伏字.mp4` gets matched/tagged even like this `/path/to/foo/test_字伏字_720p.mp4` while the first not at all (the same thing happens in the latest dev version)
> 
> I haven't been able to reproduce this. I tried both filename variants and auto tagging from the performer/tag and tasks pages. Scenes were successfully tagged in all scenarios.

Yeah i found the issue :sweat_smile: , one of the scenes i used was actually `organized` ... Tests fine now

MC1R-13:
> For me this build seems to cause a big impact in performance
> 
> * Create a new tag
> * Save
> * Hit autotag (from the  newly created tag not the task)
>   Repeat the same for performers
> 
> Dev version 3-5 secs This Build 35 secs

I'm having this exact issue with the autotag task. It's fine with newly created tags.

pkg/match/path.go
```diff
@@ -36,7 +43,7 @@ func getPathWords(path string) []string {
 	}
 
 	// handle path separators
-	const separator = `(?:_|[^\w\d])+`
+	const separator = `(?:_|[^\p{L}\w\d])+`
```
DampToast, line 46: It probably won't give a speed increase but I think the \w can be removed since \p{L} and \d already encompass it.  Same on line 84 I think.

PR #2356, opened 2022-03-03
Restructure go project

WithoutPants:
This is a significant restructuring and refactoring of the go project structure. The main purpose of this restructuring is to separate reusable packages from stash internal packages, and to reduce the rather extensive coupling between packages.

This is not an exhaustive restructuring, and there is still plenty of non-reusable packages in `pkg`. This is mostly due to their reliance on `models`, which needs to have further refactoring done.

These changes are made to make the project more in line with the [standard golang project layout](https://github.com/golang-standards/project-layout).

The main impetus for these changes were to make the file tracking implementation easier.

List of changes:
- moved main code into `cmd/stash` 
- move `api`, `autotag`, `desktop`, `dlna`, `identify`, `manager`, and `static` into the new `internal` folder.
- added `ui` package to embed ui files in, instead of in the root package.
- moved various `utils` functions and data structures into client packages where such things were only in use by one package.
- moved MD5 hash functions into `hash/md5` and oshash functions into `hash/oshash`.
- moved other hash functions into `hash`.
- moved file-system related functions in `utils` into `fsutil`.
- refactored `logger` to use a generic interface, and move stash-specific logging implementation into `internal/log`. This removes the dependence of `logger` on `logrus`. It contains a global field `Logger`, but in future we should be treating `LoggerImpl` as a dependency and passing it, rather than calling the package functions.
- moved int slice utility functions from `utils` into `sliceutil/intslice` and string slice utility functions into `sliceutil/stringslice`.
- moved `utils.SliceSame` into `sliceutil`.
- removed md5 return value from `utils.ProcessBase64Image` since this was unused in all calling code
- moved `utils.GetFunscriptPath` into `scene`.
- moved `desktop.Start` from `api.Start` to `manager.singleton.PostInit`.
- moved code to write icon file to `manager` instead of `desktop`.
- removed `desktop.HideExecShell` in favour of new package function `exec.Command` which automatically performs the required functionality. This eliminates the `desktop` dependency on many packages.
- decoupled `dlna`, `plugin`, `session` and `scraper` from `config`.
- changed `GenerateRandomKey` to return an error rather than logging a warning.
- moved `manager/jsonschema` and `manager/paths` to `models/jsonschema` and `models/paths` respectively.
- moved desktop notification code from `job` to `manager` to eliminate dependency of `job` on `desktop`.
-


WithoutPants:
Rebased to resolve conflicts.

bnkai (approved):

bnkai:
From a quick pass over the code everything looks ok.
Basic functionality test ok also.

PR #2363, opened 2022-03-06
Remove single unicode character from autotag query

WithoutPants:
Fix for slow autotag performance. 

The inclusion of `\p{L}` in the performer/tag/studio autotag query slows it significantly. Reverted to use `\w` only.

This reopens #2293 with a known issue that autotag will now miss performer/tag/studios including a single unicode character word in the title. It's a very minor corner case, but an issue nonetheless.

The eventual solution will be to query for single character objects up front in the autotag task and cache them, instead of querying for them every time.

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,9 +35,10 @@ func main() {
 	manager.Initialize()
 	api.Start(uiBox, loginUIBox)
 
-	// stop any profiling at exit
-	defer pprof.StopCPUProfile()
 	blockForever()
+
+	// stop any profiling at exit
+	pprof.StopCPUProfile()
 
 	manager.GetInstance().Shutdown(0)
 }
--- a/pkg/match/path.go
+++ b/pkg/match/path.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"unicode"
 
 	"github.com/stashapp/stash/pkg/gallery"
 	"github.com/stashapp/stash/pkg/image"
@@ -12,7 +13,12 @@ import (
 	"github.com/stashapp/stash/pkg/scene"
 )
 
-const separatorChars = `.\-_ `
+const (
+	separatorChars = `.\-_ `
+
+	reNotLetterWordUnicode = `[^\p{L}\w\d]`
+	reNotLetterWord        = `[^\w\d]`
+)
 
 func getPathQueryRegex(name string) string {
 	// escape specific regex characters
@@ -68,10 +74,22 @@ func getPathWords(path string) []string {
 	return ret
 }
 
+// https://stackoverflow.com/a/53069799
+func allASCII(s string) bool {
+	for i := 0; i < len(s); i++ {
+		if s[i] > unicode.MaxASCII {
+			return false
+		}
+	}
+	return true
+}
+
 // nameMatchesPath returns the index in the path for the right-most match.
 // Returns -1 if not found.
 func nameMatchesPath(name, path string) int {
-	re := nameToRegexp(name)
+	// #2363 - optimisation: only use unicode character regexp if path contains
+	// unicode characters
+	re := nameToRegexp(name, !allASCII(path))
 	found := re.FindAllStringIndex(path, -1)
 
 	if found == nil {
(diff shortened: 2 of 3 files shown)


kermieisinthehouse:
Why not restore the original regex instead, so it will match the first letter in non-unicode strings, and we can narrow the hole to just single first letter non-unicode strings?

WithoutPants:
> Why not restore the original regex instead, so it will match the first letter in non-unicode strings, and we can narrow the hole to just single first letter unicode strings?

It is the original regex. It's changed to a static constant, and the definition uses ``` characters instead of `"` so that the escape characters are no necessary.

bnkai:
Is it supposed to revert all the changes from #2336? (the 'p{L}' is still included in the nameMatchesPath function)

A minor optimization for the autotag specific part like
```diff
diff --git a/pkg/match/path.go b/pkg/match/path.go
index e80601fc..b8543658 100644
--- a/pkg/match/path.go
+++ b/pkg/match/path.go
@@ -93,6 +93,31 @@ func nameMatchesPath(name, path string) int {
 	return found[len(found)-1][0]
 }
 
+func nameToRegexp(name string) *regexp.Regexp {
+	// escape specific regex characters
+	name = regexp.QuoteMeta(name)
+
+	name = strings.ToLower(name)
+
+	// handle path separators
+	const separator = `[` + separatorChars + `]`
+
+	reStr := strings.ReplaceAll(name, " ", separator+"*")
+	reStr = `(?:^|_|[^\p{L}\w\d])` + reStr + `(?:$|_|[^\p{L}\w\d])`
+
+	re := regexp.MustCompile(reStr)
+	return re
+}
+
+func regexpMatchesPath(r *regexp.Regexp, path string) int {
+	path = strings.ToLower(path)
+	found := r.FindAllStringIndex(path, -1)
+	if found == nil {
+		return -1
+	}
+	return found[len(found)-1][0]
+}
+
 func PathToPerformers(path string, performerReader models.PerformerReader) ([]*models.Performer, error) {
 	words := getPathWords(path)
 	performers, err := performerReader.QueryForAutoTag(words)
@@ -208,8 +233,9 @@ func PathToScenes(name string, paths []string, sceneReader models.SceneReader) (
 	}
 
 	var ret []*models.Scene
+	r := nameToRegexp(name)
 	for _, p := range scenes {
-		if nameMatchesPath(name, p.Path) != -1 {
+		if regexpMatchesPath(r, p.Path) != -1 {
 			ret = append(ret, p)
 		}
 	}
@@ -240,12 +266,12 @@ func PathToImages(name string, paths []string, imageReader models.ImageReader) (
 	}
 
 	var ret []*models.Image
+	r := nameToRegexp(name)
 	for _, p := range images {
-		if nameMatchesPath(name, p.Path) != -1 {
+		if regexpMatchesPath(r, p.Path) != -1 {
 			ret = append(ret, p)
 		}
 	}
-
 	return ret, nil
 }
 
@@ -272,8 +298,9 @@ func PathToGalleries(name string, paths []string, galleryReader models.GalleryRe
 	}
 
 	var ret []*models.Gallery
+	r := nameToRegexp(name)
 	for _, p := range gallerys {
-		if nameMatchesPath(name, p.Path.String) != -1 {
+		if regexpMatchesPath(r, p.Path.String) != -1 {
 			ret = append(ret, p)
 		}
 	}
```
removes from the pprof report the PathToScenes/Images/Tags(and seems to reduce from 14s to 12s the autotag test for me). Not a lot can be done for the the other case though. The nameMatchesPath and the regex db function/query seem to generate a lot of regexp compiles that seem to take most of the time for me when running the autotag from the tasks.

WithoutPants:
This is only intended as a performance optimisation at the expense of a regression on a corner case. It is not intended as a full revert of #2336.

Having done some further profiling, it appears that the next hotspot is the usage of `\p{L}` in `nameMatchesPath`. Removing this almost halved the execution time on a production database - after putting in the above optimisation. This would however cause a full regression of #2293 which I won't do. Instead, I have added an optimisation originally suggested by @kermieisinthehouse: where we know the path used, compile a non-unicode pattern if there are no unicode characters in the path. This appears to restore performance to its original state when used with files without unicode characters. It still appears to address #2293.

kermieisinthehouse (approved):

kermieisinthehouse:
Can confirm that it's roughy 50% faster than current develop

Issue #2366, opened 2022-03-06
[Feature] Autotag Optimizations


kermieisinthehouse:
Autotag performance currently leaves me unable to run autotag on my image library (just under 14 million images).

I'm requesting some specific optimizations so that autotag can be greatly sped up. Note that the old bulk autotag implementation (sqlite regex based), was much faster, but was less configurable and used more memory.

In scope:
- A checkbox for tags / performers / studios that enables / disables eligibility to autotag. This can reduce the search space considerably AND allow people to run autotag without worrying about pollution of e.g. single name performers.
- Memoization of compiled RE objects during an autotag run to save CPU
- Folding all alias regexes into a single regex: all aliases match the same underlying ID, and assuming they are somewhat similar, we can quickly check them all in one optimized call to regexp. We can generate all of the matching regexes to all aliases, then concat them into a master regex of "REGEX1 | REGEX2 | REGEX3 | ...". The final compiled object will optimize this pretty well, especially if the tags share substrings, like "example tag phrase" and "example tag phrases".
- A dedicated query function for the autotag task that doesn't use a sort by title. On large collections, sqlite sorts use crazy amounts of CPU time, when the order doesn't matter. The internal ROWID sort is consistent enough.
- Indexes for images / scenes tables as necessary for new query

Not in scope: I am unsure of how much time the tag narrowing strategy currently saves. Is tokenizing the string and querying the database really faster than just naively doing all regex comparisons, especially if we only compile them once during task lifetime?

This was partially started in #1927, but it is not necessary to reuse any of it.

I am willing to put a decent bounty on this issue, as it is somewhat large and very useful to me.

WithoutPants:
Bounty placed for $251.

PR #2368, opened 2022-03-07
Autotag optimisation

WithoutPants:
Follow up to #2293, #2363, #2336. 

Partial resolution to #2366.

- Adds `match.Cache`. For now, this caches performers/studios/tags that start with a single letter word. A query is used to get these objects explicitly, and the applicable regex condition is taken out of `QueryForAutoTag` query.
- Adds an LRU cache to the `regexp` sqlite function. This means that a regex pattern should only be compiled once per query.
- Changed the behaviour of `Query` for scenes, images and galleries to _not_ sort the results if no sort parameter is passed. 
- Moved regex pattern compilation out of `getPathWords` and out as a static variable.

From my testing on a production database, in a scenario where nothing needed to be updated, the autotag operation would complete in 35% of the original, unoptimised code time.

Analysis of the CPU profile for the autotag task now indicates that most of the time is spent querying the database, and not compiling regex patterns.


pkg/match/cache.go
```diff
@@ -0,0 +1,120 @@
...
+			c.singleCharPerformers = make([]*models.Performer, 0)
+		} else {
+			c.singleCharPerformers = performers
+		}
+	}
+
+	return c.singleCharPerformers, nil
+}
+
+// getSingleLetterStudios returns all performers with names that start with single character words.
```
bnkai, line 50: typo performers -> studios ?

pkg/match/cache.go
```diff
@@ -0,0 +1,120 @@
...
+// See getSingleLetterPerformers for details.
+func getSingleLetterStudios(c *Cache, reader models.StudioReader) ([]*models.Studio, error) {
+	if c == nil {
+		c = &Cache{}
+	}
+
+	if c.singleCharPerformers == nil {
```
bnkai, line 57: c.singleCharStudios ?

pkg/match/cache.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// getSingleLetterTags returns all performers with names that start with single character words.
+// See getSingleLetterPerformers for details.
+func getSingleLetterTags(c *Cache, reader models.TagReader) ([]*models.Tag, error) {
+	if c == nil {
+		c = &Cache{}
+	}
+
+	if c.singleCharPerformers == nil {
```
bnkai, line 90: c.singleCharTags ?

pkg/match/cache.go
```diff
@@ -0,0 +1,120 @@
...
+			c.singleCharStudios = make([]*models.Studio, 0)
+		} else {
+			c.singleCharStudios = studios
+		}
+	}
+
+	return c.singleCharStudios, nil
+}
+
+// getSingleLetterTags returns all performers with names that start with single character words.
```
bnkai, line 83: typo performers -> tags

pkg/database/regex.go
```diff
@@ -0,0 +1,36 @@
+package database
+
+import (
+	"regexp"
+
+	lru "github.com/hashicorp/golang-lru"
+)
+
+const regexCacheSize = 10
```
bnkai, line 9: add a comment for the size? is it in MB? (the docs in the library are not mentioning anything)

pkg/manager/task_autotag.go
```diff
@@ -586,6 +595,7 @@ func (t *autoTagFilesTask) processGalleries(r models.ReaderRepository) error {
 }
 
 func (t *autoTagFilesTask) process() {
+	begin := time.Now()
```
bnkai, line 598: should this and the relevant log print be run in 
`func (j *autoTagJob) Execute(ctx context.Context, progress *job.Progress) {`  instead ?
When doing a specific autotask no time stats are printed.

kermieisinthehouse:
I've been running autotag from this branch for 45 minutes, and I have no log entries after "starting auto tagging". Is this normal? It's using a ton of CPU so I assume it's busy, but I can't confirm it's actually working.

bnkai:
> I've been running autotag from this branch for 45 minutes, and I have no log entries after "starting auto tagging". Is this normal? It's using a ton of CPU so I assume it's busy, but I can't confirm it's actually working.

Its working fine and it is way faster than before
A single performer autotag takes <1 sec compared to 4 secs that it did before
I finished a full autotag task (performers only though) in ~30 minutes which  i never run before because it was too slow
As for CPU usage it is CPU bound for me also. I do have the stash db in an ssd and a lot of ram though.

kermieisinthehouse:
I think my issue is that because it goes by insert order now, if you've already autotagged some of it, it just spins with no log output for the majority of the time, since there are no new matches. All of the "useful" work is all toward the end, the media that you've added since your last run (unless you've added new tags).

Some progress logging would be nice, perhaps every 10,000 media steps, or when it switches from scenes to images or galleries?

bnkai:
For my case i just saw the percentage increasing in the task status.....
I only have <100k scenes+images though so i guess that wouldnt update that fast for you.
A per fixed media steps or time interval log print could be handy though.

pkg/database/regex.go
WithoutPants: I'm flexible on the actual number used. I chose a small number because it's most likely use is for a single query - this function gets called for every row in the (filtered) results. It's likely to only need no more than 1 or 2 in any given query. After that point, it's just sitting in the cache and is unlikely to be used again. I think increasing the number is probably going to give negligible performance improvements at the cost of more memory usage. I have no idea how much memory a `regexp.Regexp` instance takes up.

bnkai (approved):

pkg/database/regex.go
kermieisinthehouse: I've increased the cache size to 4096 and didn't see a change in memory usage (unless it was under 5MB), so there may be another issue somewhere...

bnkai: I just added a print when the cache is hit and it seems to me that it is like withoutpants mentioned above it is only once per query ( size =1 )
I did the below change(just for testing)and left it running some more time and it doesnt seem to get more than that

```diff
diff --git a/pkg/database/regex.go b/pkg/database/regex.go
index dc7b5feb..cdce9766 100644
--- a/pkg/database/regex.go
+++ b/pkg/database/regex.go
@@ -1,6 +1,7 @@
 package database
 
 import (
+       "fmt"
        "regexp"
 
        lru "github.com/hashicorp/golang-lru"
@@ -12,7 +13,7 @@ import (
 // results. It's likely to only need no more than 1 or 2 in any given query.
 // After that point, it's just sitting in the cache and is unlikely to be used
 // again.
-const regexCacheSize = 10
+const regexCacheSize = 100
 
 var regexCache *lru.Cache
 
@@ -36,6 +37,10 @@ func regexFn(re, s string) (bool, error) {
                regexCache.Add(re, compiled)
        } else {
                compiled = entry.(*regexp.Regexp)
+               length := regexCache.Len()
+               if length > 1 {
+                       fmt.Printf("cache usage %d", length)
+               }
        }
 
        return compiled.MatchString(s), nil
```

PR #2375, opened 2022-03-09
Fix identify and script scraper bugs

WithoutPants:
Supercedes #2231 

* continue checking sources if a scraper source fails during identify task
* ensure nil values are returned from `scraperSource.ScrapeScene` and `scraper.Cache.ScrapeID`
* convert scraper output to object pointers instead of concrete objects so that null json values are handled correctly

Fixes issue where autotag scraper would output the following if nothing could be tagged: `error scraping from scraper builtin_autotag: could not convert content to scene`

Fixes issue where a script scraper returning no results would be interpreted as a found scene despite being empty.

### DIFF
--- a/pkg/api/resolver_query_scraper.go
+++ b/pkg/api/resolver_query_scraper.go
@@ -221,10 +221,14 @@ func (r *queryResolver) ScrapeSingleScene(ctx context.Context, source models.Scr
 				return nil, fmt.Errorf("%w: sceneID is not an integer: '%s'", ErrInput, *input.SceneID)
 			}
 			c, err = r.scraperCache().ScrapeID(ctx, *source.ScraperID, sceneID, models.ScrapeContentTypeScene)
-			content = []models.ScrapedContent{c}
+			if c != nil {
+				content = []models.ScrapedContent{c}
+			}
 		case input.SceneInput != nil:
 			c, err = r.scraperCache().ScrapeFragment(ctx, *source.ScraperID, scraper.Input{Scene: input.SceneInput})
-			content = []models.ScrapedContent{c}
+			if c != nil {
+				content = []models.ScrapedContent{c}
+			}
 		case input.Query != nil:
 			content, err = r.scraperCache().ScrapeName(ctx, *source.ScraperID, *input.Query, models.ScrapeContentTypeScene)
 		default:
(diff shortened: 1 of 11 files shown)


bnkai:
Fixes above seem to work ok.
There is still an issue with json scrapers and identify
I tested the below in identify ThePornDB(json scraper)->stashdb->Traxxx(script scraper)
When there is no matching result from ThePornDB we get the below
```bash
[33mWARN[0m[2022-03-13 12:48:22] key 'Details': could not find json path 'data.0.description' in json object 
[33mWARN[0m[2022-03-13 12:48:22] key 'Image': could not find json path 'data.0.background.small' in json object 
[33mWARN[0m[2022-03-13 12:48:22] key 'Title': could not find json path 'data.0.title' in json object 
[33mWARN[0m[2022-03-13 12:48:22] key 'URL': could not find json path 'data.0.url' in json object 
[33mWARN[0m[2022-03-13 12:48:22] key 'Date': could not find json path 'data.0.date' in json object 
[37mDEBU[0m[2022-03-13 12:48:22] Nothing to set for xxxxxxxx.mp4
```
and the identify doesnt skip to the next scraper (it only does if there is an error during scrape, timeout for example)
A sample response (no match) from tpdb that causes the above is 
```json
{"data":[],"links":{"first":"https:\/\/api.metadataapi.net\/scenes?query=&page=1","last":"https:\/\/api.metadataapi.net\/scenes?query=&page=1","prev":null,"next":null},"meta":{"current_page":1,"from":null,"last_page":1,"links":[{"url":null,"label":"&laquo; Previous","active":false},{"url":"https:\/\/api.metadataapi.net\/scenes?query=&page=1","label":"1","active":true},{"url":null,"label":"Next &raquo;","active":false}],"path":"https:\/\/api.metadataapi.net\/scenes","per_page":25,"to":null,"total":0}
```
The above behaviour means that json scrapers can not be used above others in the scraper list as they will only transition to the next scraper due to an error

Transitions from script scrapers (Traxxx) or stash-box scrapers (stashdb) to the next in list seem to work fine (Either due to an error or no match )

bnkai (approved):

bnkai:
Looks ok and tests ok now.

PR #2390, opened 2022-03-17
Order gallery images by path

### LINKED ISSUE #2376: [Bug Report] Gallery Order gets messed up in wall view and lightbox
**Describe the bug**
In the most current version v0.13.0-2-ga0595282 the gallery wall view is not in the order that one would expect. I didn't even find a pattern what exactly happens here. For quick info, I am only using .cbz files for galleries, so maybe this is for those files only. 
This messed up order (and cover image as well) is also continued when clicking on a wall item to show the light box. Interestingly enough, when going into the specific gallery (the view you get into by clicking in the grid view) the order is still respected and also the lightbox started here with the wall view respects the galleries order.
The front page (cover) is also wrong in the grid and list view, and consistently so (between views or sorting orders).
About my .cbz files, they contain their images numbered pretty consistently, so basically from 001.jpg to 999.jpg for example. I have never found a program that wasn't able to figure out the order of that.
I was able to pin down that it was down after version v0.12.0-82-g9ef30604, so in terms of commits it has to come from one of those:
https://github.com/stashapp/stash/pull/2353
https://github.com/stashapp/stash/pull/2368
https://github.com/stashapp/stash/pull/2372
https://github.com/stashapp/stash/pull/2363

**To Reproduce**
1. Go into Galleries
2. Choose the Wall option
3. Open lightbox
4. Open other view-options in Galleries

**Expected behavior**
Front cover being the first image in the .cbz, further images in the lightbox being in the alphanumerical order of the file names if not defined otherwise.

v0.13.0-2-ga0595282

 - OS: Ubuntu 21.10
 - Browser Chromium
 - Version 98.0.4758.102

WithoutPants:
Changes `gallery.Images` resolver to query using the `Query` function and to sort by path.

Fixes #2376

### DIFF
--- a/internal/api/resolver_model_gallery.go
+++ b/internal/api/resolver_model_gallery.go
@@ -26,7 +26,10 @@ func (r *galleryResolver) Title(ctx context.Context, obj *models.Gallery) (*stri
 func (r *galleryResolver) Images(ctx context.Context, obj *models.Gallery) (ret []*models.Image, err error) {
 	if err := r.withReadTxn(ctx, func(repo models.ReaderRepository) error {
 		var err error
-		ret, err = repo.Image().FindByGalleryID(obj.ID)
+
+		// #2376 - sort images by path
+		ret, err = image.FindByGalleryID(repo.Image(), obj.ID, "path", models.SortDirectionEnumAsc)
+
 		return err
 	}); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


bnkai (approved):

WithoutPants:
This was introduced in the dev branch so no changelog entry required.

PR #2391, opened 2022-03-17
Move tag exclusion to scrape query resolver

### LINKED ISSUE #2379: [Bug Report] Excluded Tag Patterns don't work for scraping from stashdb.org
"Excluded Tag Patterns" is not working when scraping metadata from stashdb.org

**To Reproduce**
1. set tags you want to filter in the "excluded tag patterns"-list (Settings -> metadata provider -> excluded tag patterns)
2. go to a scene you want to scrape from stashdb.org with at least one tag on stashdb.org from the list
3. scrape from stashdb.org
4. the tags that should be ignored, will be added to the scene 

**Expected behavior**
The tags from the "excluded tag patterns"-list should be ignored while scraping from stashdb.org

v0.13.0-2

 - OS: Windows 10
 - Browser Firefox

WithoutPants:
Fixes #2379 

Moves the tag exclusion code into the scrape query resolver so that all code paths are covered by the tag exclusion code.


internal/api/resolver_query_scraper.go
```diff
@@ -113,7 +123,52 @@ func (r *queryResolver) ScrapeScene(ctx context.Context, scraperID string, scene
...
+		return
+	}
+
+	for _, s := range scenes {
+		var newTags []*models.ScrapedTag
+		for _, t := range s.Tags {
+			ignore := false
+			for _, reg := range excludeRegexps {
+				if reg.MatchString(strings.ToLower(t.Name)) {
+					ignore = true
```
bnkai, line 160: In the removed code the ignored tags (t.Name) are also added to a list and then log printed. Should we readd that code?

bnkai (approved):

PR #2392, opened 2022-03-17
Restructure ffmpeg

WithoutPants:
Significant refactor of the `ffmpeg` package. Focus here was on making a more consistent and reusable package interface.

Changes include:
* removed the `stripExt` parameter from `NewVideoFile` since this was used during scanning only
* introduced `fsutil.ReadLockManager` to replace running stream checking code. Allows clients to "lock" a particular file for reading. When an operation requires to delete a file, it runs `Cancel` on all of the running lock contexts before returning to allow deletion.
* eliminates `ffprobe` execution when stream transcoding files. Should improve performance.
* moved frame rate calculation code into `ffmpeg`
* moved most scene generation code into new `scene/generate` package
* added `ffmpeg/transcoder` package to expose basic video and image manipulation functions: `Transcode` (re-encodes or copies a video with optional start time and duration), `Splice` (combines video files using a concat file), `Screenshot` (takes a screenshot at a given position), `ImageThumbnail` (makes a thumbnail of an image).
* moved `JSONTime` into `models/json` package so that `ffmpeg` does not depend on `models`.
* moved video phash generation code into new `hash/videophash` package. This should help formalise the algorithm.
* added functions to `ffmpeg` for argument building

Most public interfaces are now documented according to godoc specs. The only thing that's still in a halfway state is the sprite generation stuff, which will need another refactor pass in future.

Biggest things to consider with this change are ensuring phash generation is consistent with the existing behaviour, ensuring all generation operations work correctly, ensuring live transcoding continues to work and that scene deletion follows existing behaviour with respect to running streams.


bnkai:
PR seems ok and tests ok.
---
Checked
- [x] previews generation
- [x] screenshots generation
- [x] markers generation
- [x] phashes generation (phash values seem to remain the same as before for a few scenes i checked)
- [ ] live transcoding seems to work ok (Needs to be rechecked against videojs when dev gets pulled).
 Cancelation also seems to work if you try to delete a scene that is live transcoded (also needs a check against videojs). An extra debug print would be useful though.
---
Todo ?
- #2458  needs to be re-applied (it is probably in the conflicting files)
- According to #2488  context should not be used in structs, so the lock_manager code may need some refactoring?

internal/manager/task_transcode.go
```diff
@@ -111,14 +108,14 @@ func (t *GenerateTranscodeTask) isTranscodeNeeded() bool {
...
-		audioCodec = ffmpeg.AudioCodec(t.Scene.AudioCodec.String)
+		audioCodec = ffmpeg.ProbeAudioCodec(t.Scene.AudioCodec.String)
 	}
 
 	if t.Scene.Format.Valid {
 		container = t.Scene.Format.String
 	}
 
-	if ffmpeg.IsStreamable(videoCodec, audioCodec, ffmpeg.Container(container)) {
+	if ffmpeg.IsStreamable(videoCodec, audioCodec, ffmpeg.Container(container)) != nil {
```
bnkai, line 118: If file `is` streamable no transcode is needed
```suggestion
	if ffmpeg.IsStreamable(videoCodec, audioCodec, ffmpeg.Container(container)) == nil {
```

PR #2422, opened 2022-03-24
Improve performance of gallery and image queries

### LINKED ISSUE #2347: findImages seems to have some performance issues
A gallery can take almost a minute trying to load in from the gallery page but loads almost instantly in the scene page which even loads all the images at once.

User DampToast mentioned `The quick load uses findGallery, the slow one uses findImages.  I mentioned it a few days back.  If you use wall view on the galleries page, they load faster as well, but if you go into the individual gallery (even wall view in there) it is slow`. It seems like findImages likely has some performance issues since both methods are essentially doing the same thing in this case yet one method is significantly faster

The discussion around this issue in the Discord channel can be found at https://discord.com/channels/559159668438728723/559159910550732809/945479464169660496.

WithoutPants:
Fixes #2347 

Reverts to use `FindByGalleryID` for gallery image queries. Changed the query string to use an inner join, which is faster.

Optimised the `getMultiCriterionClause` to utilise the join table is present. 

This addresses a performance regression introduced in the dev branch, and a latent performance issue for `findImages`.

On my contrived database with 4m images, it reduced the load time for `findImages` from 5.5s to 123ms. `findGalleries` wasn't returning at all.

### DIFF
--- a/internal/api/resolver_model_gallery.go
+++ b/internal/api/resolver_model_gallery.go
@@ -28,7 +28,11 @@ func (r *galleryResolver) Images(ctx context.Context, obj *models.Gallery) (ret 
 		var err error
 
 		// #2376 - sort images by path
-		ret, err = image.FindByGalleryID(repo.Image(), obj.ID, "path", models.SortDirectionEnumAsc)
+		// doing this via Query is really slow, so stick with FindByGalleryID
+		ret, err = repo.Image().FindByGalleryID(obj.ID)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


kermieisinthehouse:
Galleries load faster on my database, from many seconds to under one second. Accessing the images view still takes ~15 seconds, regardless of sort

kermieisinthehouse (approved):

bnkai (approved):

PR #2430, opened 2022-03-26
Enable FreeBSD support

### LINKED ISSUE #2429: [Bug Report] Build stash on FreeBSD
**Describe the bug**
Compiling stash on FreeBSD leads to an error.

**To Reproduce**
Run gmake in the stash source folder.

**Expected behavior**
The compilation should succeed.

**Screenshots**
[jmt@clu ~/src/stash]$ gmake
go generate -mod=vendor ./cmd/stash
validation failed: packages.Load: /usr/home/jmt/src/stash/vendor/github.com/kermieisinthehouse/systray/systray_nonwindows.go:13:8: could not import C (no metadata for C)
exit status 1
cmd/stash/main.go:1: running "go": exit status 1
gmake: *** [Makefile:165: generate-backend] Error 1

This error comes with v0.13.0. The version v0.12.0 compiles without any problems.

 - Device: PC
 - OS: FreeBSD 13.0-RELEASE-p10

**Additional context**
The problem comes with the systray library. I managed to disable systray and work arround this problem for the FreeBSD os. But I thing the thrown error is not systray-specific.

isnogudus:
This code enable FreeBSD support. It is mainly inspired by @bnkai on issue #2429. But I also had to edit the systray library for making it compilable on FreeBSD.

Fixes #2429

### DIFF
--- a/internal/desktop/desktop_platform_freebsd.go
+++ b/internal/desktop/desktop_platform_freebsd.go
@@ -1,0 +1,39 @@
+//go:build freebsd
+// +build freebsd
+
+package desktop
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/stashapp/stash/pkg/logger"
+)
+
+// isService checks if started by init, e.g. stash is a *nix systemd service
+func isService() bool {
+	return os.Getppid() == 1
+}
+
+func isServerDockerized() bool {
+	_, dockerEnvErr := os.Stat("/.dockerenv")
+	cgroups, _ := ioutil.ReadFile("/proc/self/cgroup")
+	if !os.IsNotExist(dockerEnvErr) || strings.Contains(string(cgroups), "docker") {
+		return true
+	}
+
+	return false
+}
+
+func sendNotification(notificationTitle string, notificationText string) {
+	err := exec.Command("notify-send", "-i", getIconPath(), notificationTitle, notificationText, "-a", "Stash").Run()
+	if err != nil {
+		logger.Errorf("Error sending notification on Linux: %s", err.Error())
+	}
+}
+
+func revealInFileManager(path string) {
+
+}
--- a/internal/desktop/desktop_platform_freebsd.go
+++ b/internal/desktop/desktop_platform_freebsd.go
@@ -14,23 +14,19 @@ import (
 
 // isService checks if started by init, e.g. stash is a *nix systemd service
 func isService() bool {
-	return os.Getppid() == 1
+	// at the moment there is no desktop integration for FreeBSD
+	return true
 }
 
+// there is no docker on FreeBSD
 func isServerDockerized() bool {
-	_, dockerEnvErr := os.Stat("/.dockerenv")
-	cgroups, _ := ioutil.ReadFile("/proc/self/cgroup")
-	if !os.IsNotExist(dockerEnvErr) || strings.Contains(string(cgroups), "docker") {
-		return true
-	}
-
 	return false
 }
 
 func sendNotification(notificationTitle string, notificationText string) {
 	err := exec.Command("notify-send", "-i", getIconPath(), notificationTitle, notificationText, "-a", "Stash").Run()
 	if err != nil {
-		logger.Errorf("Error sending notification on Linux: %s", err.Error())
+		logger.Errorf("Error sending notification on FreeBSD: %s", err.Error())
 	}
 }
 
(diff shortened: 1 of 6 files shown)


bnkai:
I am not sure if editing the vendored files is best practise here, especially since the library was forked so that it can get included in stash  ? Maybe @kermieisinthehouse  can update the fork and then we would just do a `go mod vendor` ?

isnogudus:
I totally agree with you. But I did not find a way to send @kermieisinthehouse an issue to his fork.

isnogudus:
I do not understand the error in the windows cross system build. There is no "!" in the mentioned file. !?!?!? I seem to oversee something.

kermieisinthehouse:
I'll patch the fork, and create a separate PR that implements it. In general, vendor files aren't editable. Also, we can add freebsd to the build system, there are like 10 users. Really up to @WithoutPants if we want to include it.

isnogudus:
The diff shows, that the "!" in the build-line of `internal/desktop/systray_nonlinux.go` is gone in the freebsd branch. Where comes the error from and what am I missing. Please help.

internal/desktop/systray_nonlinux.go
```diff
@@ -1,5 +1,5 @@
-//go:build windows || darwin || !linux
-// +build windows darwin !linux
+//go:build windows || darwin
+// +build windows darwin
```
bnkai, line 2: I am not sure if it is related to the build error, why didnt you just add `!freebsd` instead of removing the !linux part ?
eg
```go
//go:build windows || darwin || !linux || !freebsd
// +build windows darwin !linux !freebsd
```

internal/desktop/desktop_platform_freebsd.go
```diff
@@ -0,0 +1,39 @@
...
+
+func isServerDockerized() bool {
+	_, dockerEnvErr := os.Stat("/.dockerenv")
+	cgroups, _ := ioutil.ReadFile("/proc/self/cgroup")
+	if !os.IsNotExist(dockerEnvErr) || strings.Contains(string(cgroups), "docker") {
+		return true
+	}
+
+	return false
+}
```
bnkai, line 28: Those 2 functions are linux specific. 
Does freebsd use pid 1 for init or if an app is run a service ?
Does a docker container running in freebsd have the relevant files?
Both of those are used to check if stash is running in  desktop mode (not 100% sure about that) or not so either a correct way to detect that in freebsd is needed or just say that desktop mode is not supported and return a fixed value instead (probably `return true`?)

bnkai:
> I'll patch the fork, and create a separate PR that implements it. In general, vendor files aren't editable. Also, we can add freebsd to the build system, there are like 10 users. Really up to @WithoutPants if we want to include it.

Imo it is not necessary to add to the build system just make it easier/possible (if its easy enough and doesnt break anythin) to compile for freebsd

internal/desktop/systray_nonlinux.go
isnogudus: This does not compile at my system. So I decided to make an explicit selection. This makes only a difference, if there are more than these 4 systems (windows darwin linux freebsd).

internal/desktop/desktop_platform_freebsd.go
isnogudus: To be honest, I just copied it from the linux-file, sorry. `isService` should return true to disable desktop and `isServerDockerized` should return false. There is no docker on FreeBSD as far as I knew. FreeBSD uses PID 1 for init, but you can run a service without having `PPID==1`. There is runit for example.

isnogudus: I pushed a new and hopefully fixed version.

isnogudus: And now I reverted the previous fix and applied it to the freebsd-file, not the linux-file. It's to late for me ;-)

internal/desktop/systray_nonlinux.go
bnkai: I think it somehow interferes with the windows cross build, not sure though

vendor/github.com/kermieisinthehouse/systray/systray_nonwindows.go
```diff
@@ -1,5 +1,5 @@
-//go:build !windows
-// +build !windows
+//go:build !windows !freebsd
```
bnkai, line 1: thats probably where the error is from
didnt notice that the cross compile was using the vendor file
should be
`//go:build !windows || !freebsd`
at least for the syntax

WithoutPants:
I'll patch this branch once the systray library changes are done.

WithoutPants:
Closing as this is superceded by #2477

PR #2439, opened 2022-03-29
Add ignore autotag flag

### LINKED ISSUE #2366: [Feature] Autotag Optimizations

WithoutPants:
Resolves #2366 

Adds "Ignore Auto Tag" flag to Performers, Studios and Tags. When true, the object is ignored for autotagging from the tasks page. This will still allow auto-tagging for the object if the Auto Tag button is pressed on the object's page.

Note that the flag is visible on the Edit panel, but not currently on the Details panel. I'm open to changing this.

The flag can be filtered on for queries and has been added to the Performers edit dialog.


bnkai (approved):

kermieisinthehouse (approved):
Tested, thanks!!

aaron-typo:
At least for Tags this does not work: I set the ignore flag to a specific tag and it still gets assigned to scenes when I run autotag from the tasks page.

WithoutPants:
> At least for Tags this does not work: I set the ignore flag to a specific tag and it still gets assigned to scenes when I run autotag from the tasks page.

It's working ok for me. If you can reproduce the issue please raise a bug report issue.

aaron-typo:
> It's working ok for me.

Are aliases of the ignored tag also getting ignored for you?

PR #2443, opened 2022-03-29
Support Go 1.18: Upgrade gqlgen to v0.17.2

SmallCoccinelle:
This enables builds on Go 1.18. github.com/vektah/gqlparser is upgraded
to the newest version too.

Getting this to work is a bit of a hazzle. I had to first remove
vendoring from the repository, perform the upgrade and then re-introduce
the vendor directory. I think gqlgens analysis went wrong for some
reason on the upgrade. It would seem a clean-room installation fixed it.


SmallCoccinelle:
Ok, lets see how the CI om-nom-noms this.

SmallCoccinelle:
Also while here, I've eliminated a linter warning on the use of `strings.Title(..)` which isn't unicode aware. This should bring us into phase with the new linter, for now. There are some new nice linters we could enable as well, but I'll refrain from doing so until we have this PR in.

kermieisinthehouse:
Approved. We should bump the compiler version number and push it to the registry before merging this.

bnkai:
> Approved. We should bump the compiler version number and push it to the registry before merging this.

Can't we just revert the dockerfile changes and leave them for a separate PR? I dont think anything in this PR `requires` 1.18 it is just suppported right? The go.mod versions is still 1.17 anyway.

PR #2457, opened 2022-03-31
Fix lightbox issues

### LINKED ISSUE #2451: [Bug Report] lightbox: Play button not working, value changing, and input focus issue
Using commit f3355f3

# Issues

* Lightbox's play button stopped working.

* Changing the delay value in the lightbox options seems to multiply it by 1000 as if converting from seconds to milliseconds, even though the input asks for seconds.

* By default, my `Settings -> Interface -> Slideshow Delay` value was set to 5000. By changing it to 5, the play button started working normally again.

* When attempting to type double digits in the lightbox delay setting, the focus to the input is lost. So if I attempt to type 10, it stops at 1 and I have to reclick the input to add the 0

# Steps

Weird seconds to milliseconds conversion issue in lightbox options:

* Go to any list of images, use grid or wall view
* Open up the lightbox
* Click the gear
* Change the value to a single digit number like 3. Hit play, nothing happens.
* Close and reopen the light box
* Click the gear
* The value will now say 3000.

# Thoughts

WP thinks the interface setting used to be in milliseconds. If there was a change to this, users may need their values updated behind the scenes to account for this. Otherwise their lightbox will be 5000 seconds. May need more investigation to see if this is true.

I also think the `Settings -> Interface -> Slideshow Delay` option should specify seconds or milliseconds in the dialog window for clarity to the end user.

### LINKED ISSUE #2469: [Bug Report] scroll bar is disabled after going to image page info from lightbox view
**Describe the bug**
open any image in lightbox view than Click on image name after that go to any other page (scenes, performers, ..etc.) after that you can see there is no scroll bar and you can't scroll using the muse to fix this you have to refresh the page.

**To Reproduce**
1. open any image in lightbox view
2. Click on image name 
Screenshot 2022-04-02 131533

3. after that go to any other page (scenes, performers, ..etc)
Screenshot 2022-04-02 133030

4 see there is no scroll bar and you can't scroll
Screenshot 2022-04-02 133047

v0.13.1-28-ga2c611f9

 - OS: Windows 10 Home 21H2
 - Browser chrome
 - Version 99.0

WithoutPants:
Fixes #2451 
Fixes #2469 

- this makes the delay setting in interface settings in seconds rather than milliseconds, to be consistent with the lightbox interface. I will add a note in the changelog notifying users of this, since no conversion of existing values will be done.
- fixed slideshow delay setting to use the correct interval value
- fixed issue that caused focus to be lost on the delay input field when typing

Also changes the localforage hook to store objects directly instead of converting to and from json strings.

### DIFF
--- a/ui/v2.5/src/components/Changelog/versions/v0140.md
+++ b/ui/v2.5/src/components/Changelog/versions/v0140.md
@@ -1,3 +1,5 @@
+##### 💥 Note: Image Slideshow Delay (in Interface Settings) is now in seconds rather than milliseconds and has not been converted. Please adjust your settings as needed.
+
 ### ✨ New Features
 * Add python location in System Settings for script scrapers and plugins. ([#2409](https://github.com/stashapp/stash/pull/2409))
 
--- a/ui/v2.5/src/components/Images/ImageList.tsx
+++ b/ui/v2.5/src/components/Images/ImageList.tsx
@@ -65,6 +65,8 @@ interface IImageListImages {
   onChangePage: (page: number) => void;
   pageCount: number;
   onSelectChange: (id: string, selected: boolean, shiftKey: boolean) => void;
+  slideshowRunning: boolean;
+  setSlideshowRunning: (running: boolean) => void;
 }
 
 const ImageListImages: React.FC<IImageListImages> = ({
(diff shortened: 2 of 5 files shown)


bnkai:
The duration setting in the interface settings page and the actual lightbox has an issue now.
Setting a 3 secs duration in the interface settings results in 0.003 in the lightbox
interface-settings

lightbox-settings

bnkai (approved):

PR #2465, opened 2022-04-01
Change seek behavior

stg-annon:
changes up/down arrows to seek 1 minute forward/backward per suggestion in discord

moves volume shortcut to plus/minus on top row as well as numpad

feel free to reject this if it does not have much support, more of a suggestion


WithoutPants:
Deferring to next release to get more feedback.

WithoutPants:
It doesn't appear to have popular support as far as I can ascertain. Closing for now.

PR #2466, opened 2022-04-02
Mobile scene details page revamp

Teda1:
This pull request improves the user experience on the scene details page. The video and nav tabs will now stay in view as you scroll through scene details, markers, filters, etc. See https://discord.com/channels/559159668438728723/644934273459290145/959594477079629965 for Demo.

### DIFF
--- a/ui/v2.5/src/index.scss
+++ b/ui/v2.5/src/index.scss
@@ -134,10 +134,10 @@ textarea.text-input {
     display: grid;
     grid-auto-flow: column;
     height: 3rem;
-    margin: 0;
-    margin-bottom: 0rem;
-    max-height: 3rem;
     left: 0;
+    margin: 0;
+    margin-bottom: 0;
+    max-height: 3rem;
     padding-top: 0.1rem;
     position: fixed;
     right: 0;
@@ -203,7 +203,6 @@ textarea.text-input {
     padding-top: 0.5rem;
   }
 }
-
 @media (min-width: 576px) {
   .zoom-0 {
     width: 240px;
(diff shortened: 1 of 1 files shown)


WithoutPants:
The o-counter, organized and overflow menu buttons are no longer visible.

image
(old)
image

Teda1:
Ideally, I'd like to bring that section out of the tablist and into one of the tabs. But if I was going to make this change with just CSS I think the best option would be to float those items down to the navbar.

Screenshot 2022-04-02 093846

Teda1:
How about this:
Screenshot 2022-04-02 145234
Screenshot 2022-04-02 145219

WithoutPants:
You've still lost the organized flag. Can you not put the controls on the next row? We can always hide it in custom css.

Teda1:
I'll give that a go.

Teda1:
This isn't too bad.
Screenshot 2022-04-02 151609
Screenshot 2022-04-02 151710
Screenshot 2022-04-02 151627

WithoutPants:
I think it would look better right-justified.

Teda1:
Yeah, I fixed that after posting. I'll push something up soon.

Teda1:
Results after the change I just pushed.
Screenshot 2022-04-02 163013

WithoutPants:
The overflow menu appears under the edit buttons:

image

Teda1:
That problem should be gone now.
Screenshot 2022-04-02 232038

WithoutPants (approved):

PR #2468, opened 2022-04-02
Do not display tabs with no content

### LINKED ISSUE #2432: [Feature] [UI] Do not display tabs with 0 count pill badges
The Stash UI is becoming very busy with redundant/unnecessary UI elements. One such element is the performer tab badge (see image below) when the badge count reads 0.

Screenshot 2022-03-27 155212

Do not display the badge when the count is 0.

***Reasoning***
0 count = NULL = not displaying the badge. These are all relaying the same information to the user, except the latter reduces the amount of clutter on the UI. Your email client, SMS client, Discord client, etc. have visual indicator counters for every folder/contact/channel that has unread mail/messages, it doesn't show 0 beside each item that doesn't meet the unread criteria. the same principle applies here.

Pills/badges in general should not appear when their count is 0. Their purpose as a UI element is to notify the user of a >0 integer.

Give 0 count badges a unique class name so they can be hidden using CSS injection.

WithoutPants:
Resolves #2432 

Hides tabs in the performer/studio/tag pages where the count is zero.


kermieisinthehouse (approved):
This was my preferred solution to this, although I didn't follow the discussion all the way. Awesome!

PR #2472, opened 2022-04-04
Update v0140.md file with scene detail page update

Teda1:
I never updated the v0140.md file with my changes from https://github.com/stashapp/stash/pull/2466.


WithoutPants:
I think it should probably be `re-styled` rather than `revamped`, and it should specify for mobile devices.

Teda1:
I'll create a new pr.

PR #2503, opened 2022-04-12
Separate graphql API from rest of the system

WithoutPants:
Currently, the `models` package is our dumping ground for any data structs used by multiple packages, plus all of the generated graphql types and methods. The latter means that anything using the `models` package must run the generate command for it to compile. This is really not good for code reuse. There's also a whole heap of types and functions that are only used in the `api` package. It makes no sense to have these in the `models` package.

This change moves the generated graphql code into the `api` package. Common structs that are used across the system are left or added to the `models` package. Where a type is used by only one package other than `api`, the type is added to that package and a binding is added to graphql.

Eventually, some of these types should be moved out of the `models` package where it is appropriate to do so.

There is no functional change to the system, and provided that it compiles, I don't anticipate any particular regressions.


WithoutPants:
Going to iterate on this further.

PR #2509, opened 2022-04-14
Update stash-box fingerprint query to fully support distance matching

### LINKED ISSUE #2269: [Bug] "Scrape All" does not return matches if remote PHash was distance-matched
**Stash Version:**
`v0.12.0-31-g184117ea`

**When called on a Stash-Box source (StashDB),
the "Scrape All" tagger button does not return matches when the PHash was matched via hamming-distance matching,
and is not equal to the original PHash sent.**
Maybe not really a bug, but it's something that caused some confusion on Discord, and looks to me like an oversight.

**To reproduce, what I did was:**
1. Take a random PHash fingerprint from StashDB
2. Modify it to be within a hamming distance of 2 from the original (make sure the modified hash does not exist on the scene).
3. Convert hex to int and insert into the database as the PHash value for any scene not matched with with a Stash-Box source.
4. Tagger view, make sure scene is visible, then Scrape All. **No match found for the scene**
5. Scrape by Fragment on the scene itself, **Match found by PHash.**

**I debugged this a little, and I believe the reason for this is:**
Scrape all (ultimately) calls: https://github.com/stashapp/stash/blob/184117ea39132ceb8ee95f4ce44c537261abc578/pkg/scraper/stashbox/stash_box.go#L72

The fingerprints are sent correctly and scenes are received correctly,
however, when matching the submitted hashes to the matched scenes' hashes,
the PHash is only checked for an exact fingerprint match on any of the returned scenes,
which means the distance match (for example, of 2) is never linked to the local scene and the result(s) is never shown.

https://github.com/stashapp/stash/blob/184117ea39132ceb8ee95f4ce44c537261abc578/pkg/scraper/stashbox/stash_box.go#L135-L143

This caused a confusion on Discord, where Scrape All did not match the scene/fingerprint, but Scrape by Fragment did.
That's because the latter only has to "deal" with a single scene result - either matched or not matched,
and does not need to perform a match based on the fingerprint hash.

InfiniteTF:
The updated query accepts an array of hashes grouped by scene, and returns the results in the same order. This means stash no longer needs to manage the results.

I've also removed the old, deprecated queries. Additionally I've changed the input value from `ID` to `Int`. ID is mostly intended for globally unique identifiers, and using it for an integer sequence just complicates things since it's mapped to string by default.

Resolves #2269.

### DIFF
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -114,12 +114,6 @@ type Query {
   """Scrape a list of performers from a query"""
   scrapeFreeonesPerformerList(query: String!): [String!]! @deprecated(reason: "use scrapeSinglePerformer with scraper_id = builtin_freeones")
 
-  """Query StashBox for scenes"""
-  queryStashBoxScene(input: StashBoxSceneQueryInput!): [ScrapedScene!]! @deprecated(reason: "use scrapeSingleScene or scrapeMultiScenes")
-  """Query StashBox for performers"""
-  queryStashBoxPerformer(input: StashBoxPerformerQueryInput!): [StashBoxPerformerQueryResult!]! @deprecated(reason: "use scrapeSinglePerformer or scrapeMultiPerformers")
-  # === end deprecated methods ===
-
   # Plugins
   """List loaded plugins"""
   plugins: [Plugin!]
--- a/graphql/schema/types/scraper.graphql
+++ b/graphql/schema/types/scraper.graphql
@@ -133,14 +133,14 @@ input ScrapeSingleSceneInput {
   """Instructs to query by string"""
   query: String
   """Instructs to query by scene fingerprints"""
-  scene_id: ID
+  scene_id: Int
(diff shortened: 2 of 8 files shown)


kermieisinthehouse (approved):
Reviewed but not tested

WithoutPants:
> Additionally I've changed the input value from `ID` to `Int`. ID is mostly intended for globally unique identifiers, and using it for an integer sequence just complicates things since it's mapped to string by default.

The inputs to these interfaces _are_ globally unique identifiers. We may eventually want to change the `ID` scalar to be int in the backend, but these should still accept `ID`.

WithoutPants (approved):

kermieisinthehouse:
I'm seeing the tagger only hitting exact matches now, but it's unclear to me (or impossible) how to adjust distance

floydcg:
could this work like the scene duplicate checker tool where you have a choice of exact, high, medium, and low based off of what distance you select? that would allow it to be more granular but still give the same end result.

PR #2543, opened 2022-05-02
Improve Windows stash behaviour

WithoutPants:
Addresses some issues with the Windows version of stash:
- removes the `windowsgui` flag from the build. This was causing interoperability issues when running from the console, and as far as I can tell was only used to prevent the console window from appearing.
- changes the `IsDesktop` function to return true if the executable was launched from Windows, rather than invoked from a console. It does this via `isDoubleClickLaunched`.
- if `IsDesktop` returns true, then hides the console window that appears when launching via `hideConsole`.
- if `IsDesktop` is true, and an error occurs on startup, or a panic occurs, a dialog window will be displayed with the error message, instead of crashing silently.

Also makes `manager.instance` public and renames to `manager.Manager`.


kermieisinthehouse (approved):
Reviewed but not tested.

Looks good to me, generally. Have you tested getting rid of fixconsole? I think if you got rid of the windowsgui flag, you probably don't need it to fix the shell when running from a console. Otherwise looks good, and having some error handling is nice

PR #2555, opened 2022-05-04
Improve Handy integration

### LINKED ISSUE #2539: [Bug Report] Handy responsiveness poor since Handy Server side update to APIv2
This began after theHandy updated to APIv2.
When loading a new fun script to the Handy, the upload time is significantly increased. 
When playing a funscript video, starting a new video has a large delay before the previous stops, then a large delay before the new script is uploaded and begins to play
There is a slight increase in latency when scrubbing around a video, but it is a moderate increase in latency to pre APIv2

Since pause/play control responsiveness is still fairly seamless, I believe the issue is in:

    await this._handy.getServerTimeOffset();

or

await this._handy.setMode(HandyMode.hssp);

    this._connected = await this._handy
      .setHsspSetup(tempURL)
      .then((result) => result === HsspSetupResult.downloaded);
  }

Stash Version
v0.14.0-9-g0c2dc17e

There may be a newer method of connecting to the Handy servers? Not sure.

WithoutPants:
- Now syncs with a Handy device when the UI is first loaded, and caches the server offset time.
- Added more information for Handy integration to the `Interactive Options` section in the Interface Settings:
image

When not connected, the connect button allows attempting to reconnection.
When connected, there is a Sync button which calculates the server offset.

- Adds a status overlay to interactive scenes when the player is paused/not started:
image

This indicates when an interactive script is still being uploaded. It is only shown when a Handy key is configured and when the scene is interactive.

Resolves #2539 - the initial lead time was due to recalculating the server offset time before uploading the script, which is time consuming and mostly unnecessary. 

Other issues that I identified that are yet to be addressed in this PR:
- ~~does not signal to stop the device when navigating to another page~~
- ~~needs to handle looping videos~~
- ~~needs to handle next/previous navigation~~
- ~~when `autoplay` is true and the scene is interactive, it should finish uploading the funscript before starting playback~~


Morderath:
First impression:

- Responsiveness is once again very good in general, and the other issues you identified covers the remaining sources of giving the perception of latency. I'm stoked I was right that it was the Sync section of the code that was causing those delays
- UI element on the video is very good - Can you make it a bit brighter though actually? I dig the idea of keeping it out of the way, it's JUST on the wrong side of visible though, doesn't need much (I'm about 7ft from a older model flatscreen 58" making that observation)
- Not sure why, but there is now a delay when deleting videos - It takes a while to think. To expand on this -> I noticed that in the previous version where I had the other issues, I started getting a new error - I apologize, I did not capture it, but it was a large, large box of context text, essentially saying "File is in use". Refreshing the screen to try again succeeds, pausing video had no effect -> Made me think the Handy integration was causing some sort of filelocking when using the delete function that includes the "Delete Funscript also" feature that was added a few months ago.

If I notice anything else I'll let you know but well done!

inventiv3919:
I haven't had a chance to test this pr yet (tomorrow I'm going to try and finally learn how to create docker images from GitHub repositories) but I figured it might help to know that 

>  noticed that in the previous version where I had the other issues, I started getting a new error - I apologize, I did not capture it, but it was a large, large box of context text, essentially saying "File is in use". Refreshing the screen to try again succeeds, pausing video had no effect -> Made me think the Handy integration was causing some sort of filelocking when using the delete function that includes the "Delete Funscript also" feature that was added a few months 

I've been deleting videos + funscripts ever since the pr hit docker without encountering this error. I also tested it out just now by connecting my handy, playing a scripted video in one tab then deleting it from another tab and it didn't throw up any errors.

WithoutPants:
I've updated to address the leftover issues I found. I've also bumped up the opacity and the state will also stay on-screen until the device is ready, in the event that the player is started before the script is uploaded.

PR #2565, opened 2022-05-05
Log reason for not direct streaming files

### LINKED ISSUE #2557: Transcode or direct stream
This shouldn't be a bug, but it's a bit troubling

all my media files are h264 ，however, some videos are forced to be encoded during playback,
I found that frame rate determines whether or not to encode  （60fps encoded && 30fps direct)
Based on my poor host performance, I can't afford to encode playback

Whether the direct Stream can be forcibly enabled

WithoutPants:
Fixes #2557 

The `sceneStreams` graphql type now includes `disabled` and `disabled_reason`. `sceneStreams` now includes the direct stream, but sets `disabled` to true if the scene cannot be directly streamed and includes the reason.

Adds PCM 16 big endian and little endian as supported audio codecs for mp4 files.

Changes the behaviour of the direct scene route to explicitly set the mime type based on the scene container, instead of relying on the file extension.


bnkai:
Tested (linux)
- works ok on chromium/chrome
- firefox (99.00) plays the scene directly without audio though. Manualy forcing a transcode profile is required to have correct audio playback
- opera (86.0.4363.50, 88.0.4401.0) fails to play directly with console error `(MEDIA_ERR_SRC_NOT_SUPPORTED) The media could not be loaded, either because the server or network failed or because the format is not supported)` and switches to live transcoding



I used a sample.mov (h264/pcm16le) video file converted from a h264/aac mp4 using ffmpeg
```
ffmpeg -i sample.mp4 -c:v copy -c:a pcm_s16le sample_pcm16.mov
```
as ffmpeg only muxes pcm16 in mov containers and not mp4. (for stash it doesn't matter as both  mov/mp4 will be detected as mp4)

WithoutPants:
Verified that Firefox does not correctly decode audio for files encoded with PCM. There is no easy way to detect if the audio track failed to initialise. I think for now I might close this and consider other alternatives.

Issue #1653, opened 2021-08-20
[RFC] Proposed Stash Home Page Layout


jeremymeyers:
# Creating a useful and interesting home page for Stash instances

### Scope
One of the things currently missing from Stash is a specifically useful home page for new installs (currently it goes to a changelog)

My thoughts are that a home page should serve the following goals (in this case, "Content" means any scene, movie, image, gallery, tag or performer":

- Show an overview of the users library, with ways to navigate deeper
- Display content previews in an attractive way
- Make newly added content immediately reachable
- Allow user to continue content they may have not finished previously
- Find interesting ways to present content the user may not have engaged with recently, and show combinations of content that would be useful.

##  Details

Please note that this is intended mostly as a starting point for a conversation rather than a locked-down specific recommendation.

To this end, I have sketched (very loosely) out a few ideas:

(if anyone wants to mess about with the wireframes,  [stash redesign.zip](https://github.com/stashapp/stash/files/7022948/stash.redesign.zip) they are in[ Balsamiq ](https://balsamiq.com/)format)

### A proposed home page layout.
![Home Page](https://user-images.githubusercontent.com/3514095/130264685-76b52c99-a963-47e9-b9dd-b21acc9d5874.png)

Some features:
- Customizable sliders for recent, favorite, tags, etc
- Hero image rotator for featured content
- A way to quickly see most-used tags, as well as favorite performers, scenes, etc
- A redesigned "Tag" box with more information
- The addition of breadcrumb navigation ( further discussion is in #1297 ) as well as high-level content stats

### A "Discover" page more specifically for finding buried content.  Obviously this could be rolled into the home page.
![Discover](https://user-images.githubusercontent.com/3514095/130264725-2846ce92-6a70-4a46-894c-4d51efe844e0.png)

### A reworked "Content Info Box" (currently its a Scene one) more applicable to this design
![Scene Info Box](https://user-images.githubusercontent.com/3514095/130264813-159f2ed1-f193-47cf-8bb1-896418d84a4b.png)

### A reworked "Tag Box" more applicable to this design
![Tag Box](https://user-images.githubusercontent.com/3514095/130264846-2e4766b7-84db-4dba-80dc-01671a494dbd.png)

## Examples
Other content management apps do some version of a good job at this.  Here's the home page from the HTPC app i use, Emby.

[...]

Issue #633, opened 2020-06-24
[Feature] Play Counter


velzy00:
**Is your feature request related to a problem? Please describe.**
I'd love for Stash to track the play count for scenes, so I can get a better sense of what's getting watched and what isn't. 

**Describe the solution you'd like**
I would suggest presenting the play counter next to the o_counter in the Details tab of the Scene page and on the Scene card. On the Edit tab, the functionality to decrease or reset the count would work the same as the o_counter. Instead of manually tracking this count though, Stash would automatically track every time someone plays a scene. Hit play on a scene and the count goes up.

If we wanted to get fancy, there could also be a setting for the minimum play duration required for a play to be counted, but that may be more complex than is.

**Describe alternatives you've considered**
I think the the o_counter is the best model for the to be presented. This could also be shown as a section in the details tab, with the playcount and last play date/time show but that the o_counter style seems cleaner. 

**Additional context**
It can be hard to keep track of so many scenes, particularly when continuously adding new more. Having access to more data to use in pruning the collection is so helpful for making decisions to manage the size of the collection.

honeybeaver:
+1 and would be nice to also track play history, and less important, play progress in videos. Thanks.

kermieisinthehouse:
Good idea but please make this togglable. I often reference videos to tag them / compare duplicates and it would easily pollute the counters.

Caleripo:
Also tracking how many times a scene was played would enable really useful features in the future. 
For example you could filter by the most watched scene, display trending scenes, etc.

Stephan972:
IMO, this feature should appear in a porn theater rather than a porn organizer. The theater would count plays , views or build other analytics the organizer could consume later.

0xb0af:
> IMO, this feature should appear in a porn theater rather than a porn organizer. The theater would count plays , views or build other analytics the organizer could consume later.

This makes sense in theory but how would it actually map to the current architecture of Stash (being both an organizer, and a player, and backed by a single database)? 

Having a play count would also enable filtering by watched/unwatched (play_counter > 0)

WeedLordVegeta420:
While number of times played is a decent metric, the better metric for porn in my opinion is minutes/seconds watched. Unlike a standard TV show of film, there's less of expectation with porn that when you click play, you're going to actually watch it all the way through to the end without skipping ahead/back.

[...]

PR #2571, opened 2022-05-09
Recommendation home page

Teda1:
Related to #1653

This pull request updates the stash home page to a recommendation page similar to what can be seen on other sites or server media applications. Basic filters are provided for users across all relevant content types on this home page. The previous stats home screen has been moved to the left of the settings icons on the navbar.

Desktop Page:
RecommendationDesktop

Mobile Page:
RecommendationMobile


kermieisinthehouse:
Looks good! Amazing that this didn't require any backend changes to get this far.

My first impressions are that this needs some way to reorder or dismiss the categories. Some people don't use some of the features and it ends up unnecessarily cluttered for them. For this patch, you can probably get away with just hiding the headers if there are no items returned.

I also don't like how you can scroll left at the immediate page load? It seemed like a bug at first until I realized how it was populating. I'd rather the carousel only go right, and then zip to the beginning of you go to length + 1.

Making the stats page secondary opens up some possibilities to make it more into a dashboard, since it doesn't have to do double duty as the homepage.

Overall, awesome.

ui/v2.5/src/components/Recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+  const galleryfilter = new ListFilterModel(GQL.FilterMode.Galleries);
+  galleryfilter.sortBy = "date";
+  galleryfilter.sortDirection = GQL.SortDirectionEnum.Desc;
+  galleryfilter.itemsPerPage = itemsPerPage;
+  const galleryResult = useFindGalleries(galleryfilter);
+  const hasGalleries = galleryResult.data && galleryResult.data.findGalleries;
+
+  const messages = defineMessages({
+    emptyServer: {
```
kermieisinthehouse, line 66: These should be added to the en-gb.json file so that they can be localized.

ui/v2.5/src/components/Recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,218 @@
...
+            </div>
+          )}
+
+          {hasStudios && (
+            <div className="recommendation-row">
+              <div className="recommendation-row-head">
+                <div>
+                  <h2>{intl.formatMessage(messages.mostActiveStudios)}</h2>
+                </div>
+                <a href="/studios?sortby=scenes_count">View all</a>
```
kermieisinthehouse, line 145: I think this is missing a sortdir=desc

ui/v2.5/src/components/Recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,218 @@
...
+  studiofilter.itemsPerPage = itemsPerPage;
+  const studioResult = useFindStudios(studiofilter);
+  const hasStudios = studioResult.data && studioResult.data.findStudios;
+
+  const moviefilter = new ListFilterModel(GQL.FilterMode.Movies);
+  moviefilter.sortBy = "date";
+  moviefilter.sortDirection = GQL.SortDirectionEnum.Desc;
+  moviefilter.itemsPerPage = itemsPerPage;
+  const movieResult = useFindMovies(moviefilter);
+  const hasMovies = movieResult.data && movieResult.data.findMovies;
```
kermieisinthehouse, line 48: This doesn't work for me, I didn't investigate.

Teda1: Yeah, upon testing, it's not for me either. I can push up a fix. I intended only to display the rows if content exists.

Teda1:
I'm glad you like it!

> I also don't like how you can scroll left at the immediate page load? It seemed like a bug at first until I realized how it was populating. I'd rather the carousel only go right, and then zip to the beginning of you go to length + 1.

I fought with this for some time yesterday. It is a feature of the infinite scroll that is being used. Turning infinite scroll off would disable the prev button at the start of the list and the next button when at the end as you'd expect. The issue with this is it could lead to some significant gapping on the screen at the end of the list, as seen in the image below:

Screenshot 2022-05-09 125318

The gap issue can be fixed by forcing the elements a set number of items to be shown on screen at a time, but this would yield card stretching to ensure the set items continue to fit, which wouldn't always resize properly. The infinite scroll and variable-width solution were the better of these evils.

kermieisinthehouse (approved):
I'd get rid of the "You must be new here" verbiage, but otherwise I think this is ready to go. Reviewed, built and tested on my phone, so I can't speak for the layout in desktop widths, but mobile looks great.

Teda1:
I've gone ahead and updated the message. I also optimistically mentioned this feature in the `v0150.md` file.

Teda1:
The `always keep show slidesToShow at 1 on touch` message meant `slidesToScroll`. I've done some tests on my phone, tablet, and desktop are various window widths. I shouldn't need to commit anything else, assuming my build passes.

Caleripo:
Really cool! I would love to see stuff like „most viewed scenes“ and „trending scenes“ on this. But that probably needs some backend changes like https://github.com/stashapp/stash/issues/633

jeremymeyers:
This looks like it's gonna end up great! Glad my workup was useful.

PR #2578, opened 2022-05-11
Recommendations page bug fixes and refactoring

Teda1:
This pull request addresses the centered reflection issue that occurs when the card count is less than the number of cards displayed. I went ahead and included the refactoring in this pull request as I realized this fix would benefit from that refactor.

Before:
Center Reflection

After:
Screenshot 2022-05-11 003956

This pull request also contains a fix for the loading pop-in that could previously be seen on this page.


Teda1:
I updated the "Most Active Studios" row to show "Latest Studios", aka latest created studios, to be consistent with the other defaults.

ui/v2.5/src/components/Galleries/GalleryRecommendationRow.tsx
```diff
@@ -0,0 +1,88 @@
...
+      {
+        breakpoint: 801,
+        settings: {
+          slidesToShow: 1,
+          slidesToScroll: 1,
+          dots: false,
+        },
+      },
+    ],
+  };
```
InfiniteTF, line 69: Unless I'm mistaken you can break this out into a shared function that accepts `cardCount` and `isCount`.

kermieisinthehouse:
Tested, looks good

Issue #2290, opened 2022-02-05
[Bug Report] Python Scrapers and plugins broken


DefinitelyNotSpicy:
**Describe the bug**  
Whenever I try to use something in Stash that starts a python script it opens about 5 or 6 Windows terminal windows open and close one after another except for the last one which stays open. This problem started to occur after the Desktop integration of stash. It seems the open scraper window is waiting on the fragment input. If I manually paste the fragment into the open window and CTRL+Z the Scraper continues but doesn't return the data to stash.

**To Reproduce**

1. Open a performer edit view
2. Click on "Scrape with"
3. Choose a python scraper like IAFD
4. See Terminal windows open and close until one stays open and stash keeps showing the loading spinner.
5. Close the Terminal Window to see error

**Expected behavior**
There should be one or less terminal windows opening and the scraper should actually finish. 

**Screenshots**  
Terminal windows with keep open configured:  
![image](https://user-images.githubusercontent.com/96064475/152633719-9842ad8b-2e59-4aab-b432-1557c54aff1e.png)

Process tree when waiting for input:  
![image](https://user-images.githubusercontent.com/96064475/152634076-6eb16aef-39e4-45f5-97e3-f82714279785.png)

**Stash Version: (from Settings -> About):** v0.12.0-40-gc857e301

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Firefox
 - Version 97.0b9

**Additional context**  
This error occurs with regdardless of how I start stash. I tried it with just doubleclicking, per Windows terminal (powershell) and with git bash.

I also tried using a different Python version (tested on 3.9.9, 3.10.1, 3.10.2). I'm using pyenv-win to manage my python version so maybe that's the problem but it did work before.

kermieisinthehouse:
Somebody reported having this issue with python installed through scoop. How is python installed?

DefinitelyNotSpicy:
Python is installed through [pyenv-win](https://pyenv-win.github.io/pyenv-win/) which does use a shim afaik.  
Is there a way to tell stash which python executable to use?

kermieisinthehouse:
Not currently, but we might have to look at implementing that, since this probably affects more people.

DefinitelyNotSpicy:
[...]

PR #2590, opened 2022-05-15
Allow the use of fully qualified paths for Python.

dsully:


kermieisinthehouse (approved):

peolic (changes requested):
Haven't tested, but I believe this change would break the current way script scrapers (and possibly raw plugins) use this.
In any case (instead of this change), you should be able to use Python Path setting, which should be in the System tab of the settings, or `python_path` in the `config.yml`.

dsully:
@peolic That's the problem though. If a fully qualified Python path is specified, such as:

`/home/linuxbrew/.linuxbrew/bin/python3`

Then `isPythonCommand` will return false, since it's comparing that fully qualified path against `"python" || "python3"`

peolic:
I'll explain again.

In the scraper/plugin you define, **you use either `python` or `python3` as the first argument**.
Basically makes it so the files are shareable with minimal to no modification necessary, like on [CommunityScrapers](https://github.com/stashapp/CommunityScrapers) and [CommunityPlugins](https://github.com/stashapp/CommunityScripts).

In [__*your own* Stash configuration__](https://user-images.githubusercontent.com/66393006/168477008-7a44ff83-4e8b-499d-ae6c-fbd3576bc212.png),
you can set a fully qualified path to the Python executable, so it uses it whenever it detects a request to run Python (using `isPythonCommand`),
or leave it blank for Stash to find it from `PATH`.

For example, the code for script scraper execution:
https://github.com/stashapp/stash/blob/v0.14.0/pkg/scraper/script.go#L35-L53

```go
if python.IsPythonCommand(command[0]) {
```
would be true when the first argument of the script scraper definition is `python` || `python3`

next,
```go
		pythonPath := s.globalConfig.GetPythonPath()
		var p *python.Python
		if pythonPath != "" {
			p = python.New(pythonPath)
```
will get the Python Path from your config, and use that if it's defined.

related: #2290 #2409

dsully:
That's not the behavior I'm seeing. If I set Python to a fully qualified path in settings, scripts are not launched properly.

WithoutPants:
peolic is correct. `IsPythonCommand` is intended to be used for the plugin and scraper code to identify if the script is using python. It's entire purpose is to avoid having to put in the full path to python in the plugin/scraper config in the first place. If python isn't being resolved correctly after setting the python path in your stash configuration, then please raise a bug issue for it. This isn't the correct solution.

PR #2603, opened 2022-05-20
React code splitting

WithoutPants:
Adds lazy loading in many places so that the entire UI isn't loaded when the page is loaded.

This means that there is a small loading delay when navigating to any of the top-level pages for the first time.

It also moves the internationalisation strings into lazily loaded components so that we're not loading internationalisation strings for unused languages.

Changed the FontAwesome loading so that we explicitly specify which icons to load.

These are the largest chunks post-change, as output by vite:
```
build/assets/index.faa9decc.js                    147.98 KiB / gzip: 38.33 KiB
build/assets/MarkdownPage.1eb0aa4a.js             132.79 KiB / gzip: 42.68 KiB
build/assets/index.77af60c4.css                   288.29 KiB / gzip: 55.29 KiB
build/assets/Scene.a07eecc7.js                    587.60 KiB / gzip: 167.00 KiB
build/assets/vendor.9dc089bd.js                   897.38 KiB / gzip: 284.20 KiB
```

By comparison, this is the output pre-change:
```
build/assets/en-GB.54073393.js                37.96 KiB / gzip: 8.36 KiB
build/assets/index.846f18c0.js                54.92 KiB / gzip: 20.89 KiB
build/assets/index.8724a24c.css               289.38 KiB / gzip: 55.48 KiB
build/assets/index.6c718696.js                1538.16 KiB / gzip: 434.75 KiB
build/assets/vendor.18b73bcf.js               2400.31 KiB / gzip: 748.36 KiB
```

The vendor chunk could be divided up further - probably along React, lodash and FontAwesome - but it requires messing with the rollup config which I'm not too familiar with.

Potential risks and regressions:
* if the `Suspense` elements aren't well placed, then the page might fallback to showing a general loading page instead of restricting it to the parent component. Not exactly a big issue.
* it's possible that I've missed adding some font awesome icons to the library. It's not terribly easy to check.


InfiniteTF:
You should be able to shave off a bunch from lodash by switching to lodash-es. It's the same thing, but supports tree shaking so only the functions that are used get bundled.

Similarly for fontawesome you can get tree shaking by dropping the library import and instead importing the icons as required in each consuming component. It's tedious to rewrite all the places where it's used, but it does avoid having to maintain the huge library import.

I did both of these in stash-box a while ago:
https://github.com/stashapp/stash-box/pull/147

WithoutPants:
Moving to lodash-es took off 50kb:

```
build/assets/index.1a06e199.js                    147.58 KiB / gzip: 38.26 KiB
build/assets/MarkdownPage.a7c8cdcf.js             132.79 KiB / gzip: 42.68 KiB
build/assets/index.77af60c4.css                   288.29 KiB / gzip: 55.29 KiB
build/assets/Scene.625d4a6d.js                    587.58 KiB / gzip: 166.97 KiB
build/assets/vendor.6131cec8.js                   846.13 KiB / gzip: 266.12 KiB
```

WithoutPants:
I think what I was doing must have been sufficient, since moving to individual icon imports actually made the vendor chunk size go up a little. However, individual icon imports are less error-prone, so I think it's worth putting in anyway.

peolic:
For `lodash-es` you should be able to import like `import { isEqual } from "lodash-es"`, I believe tree-shaking works (you don't need to `import isEqual from lodash-es/isEqual`).

PR #2659, opened 2022-06-08
Support for videojs-vr

### LINKED ISSUE #2504: [Feature] Support for videojs-vr
Now that Stash is using video.js, it would be great to get support in the browser for VR with [videojs-vr](https://github.com/videojs/videojs-vr) 

Perhaps the easiest implementation would just be client-side, as a dropdown in the UI to change to projection to 180/360 degrees? If I get a chance I'll have a look into it myself now the hard work has been done moving to video.js

Teda1:
Fixes #2504 

Notes;
- video.js doesn't provide a drop-down to enable vr playback. So short of creating a button somewhere outside of the video player, auto-detection seemed like the cleanest route. Looking at other sites, they seem to be doing the same. 
- The VR player will be enabled on videos that use the `Virtual Reality`, `VR`, or `VR Porn` tags. This is the Tag and Aliases used in StashDB for VR Videos. The check for these tags is case insensitive to be more flexible.
- videojs-vr currently only supports 360 projections. Trying other projection types just yields a black screen. 360 projection works well on 180 videos though, so this is not a problem for those formats.
- videojs-vr does not play well with devices that provide their own VR playback options. Enabling any VR options on these devices while 360 projection is active yields a very poor experience, so projection will be kept at NONE on these devices.

### DIFF
--- a/ui/v2.5/package.json
+++ b/ui/v2.5/package.json
@@ -75,6 +75,7 @@
     "video.js": "^7.17.0",
     "videojs-landscape-fullscreen": "^11.33.0",
     "videojs-seek-buttons": "^2.2.0",
+    "videojs-vr": "^1.8.0",
     "videojs-vtt-thumbnails-freetube": "^0.0.15",
     "vite": "^2.7.1",
     "vite-plugin-compression": "^0.3.5",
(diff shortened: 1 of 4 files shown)


Teda1:
On top of figuring out what is going on with the yarn.lock file, I noticed that when projecting in 360, the video play glitches out when you click on any of the tabs in the `scene-tabs` area. That will require more investigation as well.

Teda1:
Closing this pull request in place of https://github.com/stashapp/stash/pull/2660.

PR #2660, opened 2022-06-09
Support for videojs-vr with new fields for screen projection and stereo mode and tag based fallback

### LINKED ISSUE #2504: [Feature] Support for videojs-vr

pickleahead:
Closes #2504 

Instead of using tags like in #2659, this approach introduces new fields for screen projection (Flat, Dome, Sphere, ...) and stereo mode (Mono, top bottom and left right).

- Using a single tag is too vague to properly support various types of projections.
- Projection and stereo mode have enums and are expandable
- Projection and stereo mode are set in the scene edit panel
- The new fields can also be used for general VR support (e.g. DeoVR api) or images in the future.
- Also includes a fix for applying image filters. Without, the filter function unhides the video element hidden by videojs-vr. The filter is now applied on the canvas in this case.

Things that could be improved:
- Updating a scene does not reload the player. Changes are not applied instantly.
- Maybe change the fields in the edit panel or move them to another tab
- Bulk editing
- Add support for scrapers


ui/v2.5/src/components/ScenePlayer/vrmode.ts
```diff
@@ -0,0 +1,46 @@
...
+        default:
+          player.vr({ projection: "360" });
+      }
+      break;
+    case "DOME":
+    case "MKX200":
+    case "RF52":
+      switch (scene.stereo_mode) {
+        case "LEFT_RIGHT":
+          player.vr({ projection: "180_LR" });
```
Teda1, line 27: Have you tested these different projections to see if they work? `videojs-vr` only claims support for 360 projections. From my experience, other projection types just display a black screen?

Teda1:
I'm happy to close my branch for yours, but the main issue with this approach is that users with 1000+ VR files would have to go back and manually update each of the scenes one by one to use these new fields you've added. Even users who are adding new scenes in bulk to their stash would likely find manually going through each new scene for these fields frustrating. I agree that this approach is more accurate, but it is also currently less accessible.

Your change needs to take into account users who view their stash using a VR headset. For these users, you will need to disable the projection setting so they can apply the VR options from their headset instead. Otherwise, the stash settings and headset settings will conflict with each other

Teda1:
> I agree that this approach is more accurate, but it is also currently less accessible.

You could mitigate this by writing a scrapper users could use to populate these new fields based on their existing tags. The scrapper could allow users to specify which tags translate to which screen projection and stereo mode combination.

> Your change also needs to take into account users who view their stash using a VR headset. For these users, you will need to disable the projection setting so they can apply the VR options from their headset instead. Otherwise, the stash settings and headset settings will conflict with each other

This issue would need to be addressed in this pull request.

pickleahead:
> I'm happy to close my branch for yours, but the main issue with this approach is that users with 1000+ VR files would have to go back and manually update each of the scenes one by one to use these new fields you've added. Even users who are adding new scenes in bulk to their stash would likely find manually going through each new scene for these fields frustrating. I agree that this approach is more accurate, but it is also currently less accessible.

Thank you! I see the issue. I added a tag based fallback for more convenience. So if a scene does not have projection and stereo mode set **but has a tag like "VR", "Virtual Reality"**, etc. the player falls back to a 180° side-by-side projection. This should be correct for most VR content out there.
Additionally Tags like "180", "180°", "180_LR", "360", "360°", "360_TB", "360_TOP_BOTTOM", etc. control the projection too.

Tags like "vr", "3d", "180" and "360" are very common. I guess for users who scraped and tagged their vr content before, the player will work just fine out of the box.

Teda1:
No worries about the scrapper, it should be pretty easy for most in the community to write and share.

> I disabled videojs-vr for certain user agents like in https://github.com/stashapp/stash/pull/2659. @Teda1 Can you explain the issue more (Just for better understanding)?

Thanks. On VR headsets applications will usually have options that will allow the user to select what projection and stereo mode they want to watch a video at. These options are available for any video you watch on the device, so the options can also be applied to non-VR videos. If you are browsing Stash on a headset and Stash has already applied projection and stereo modes, any mode you attempt to apply from the headset would be added on top of the modes Stash has already applied. This distorts the video and causes severe lag. On the Oculus Browser, you can't enter proper full-screen mode for VR videos until you have selected one of its VR options.

devloch:
If there's a development build available I could test this a bit - I have a lot of files with the default download filename that includes things like "LR_180" or "TB_360", so could also use auto-tag to test fallback.

While tags are kind of a clunky way to set the image, I'm not sure how feasible manually setting it on large collections would be - I feel like it would probably be possible for a complicated script to identify if an image is top bottom or left right, but unless there's consistent ratios figuring out the nuances of 360 LR/etc probably aren't feasible.

pickleahead:
I feel like unless the fields will be used #2105 or stash-box the issue of updating existing libraries should not be dealt in core for now. There is simply no benefit in updating every scene yet. A tag like "vr" would already work most of the time. This is also true for tags like "180" and "360", which are commonly used by sites. However, in cases, when the projection is off, one can easily set the correct values.

Detecting the projection by analysing the image data would be very complicated and slow. In most cases relying on tags and filenames should be enough to set the fields.

Teda1:
I think the current state of this pull request is good enough. If users want the option to set these new fields based on tags, I or someone else in the community can write a scrapper to do that.

Teda1 (approved):
Approving to get the wheels spinning on this change.

scruffynerf:
since I just wrote a quick vertical/mobile tagger for https://github.com/stashapp/stash/issues/2677
I realized the VR stuff is mostly the same.  Writing a plugin that on hook_scene_create, looks at the scene and does best guess field setting seems pretty simple.  So +1, and we can have a plugin that 'autotags' most of the content as it's added, and it's always tweakable that way.  I'll submit something soon.  so +1 for this patch from me.

WithoutPants:
With the files-refactor changes now in place, we have a separate concept of scenes and video files. My feeling is that the projection and stereo mode would be _file_-specific and not _scene_-specific fields. In other words, it might be possible for a scene to have two separate files, one that is not VR-enabled, or it might have a different projection field than the other file.

If my assumption is correct, then projection and stereo mode fields need to be on the file and not the scene. The difficulty there is that we don't currently have the API to edit file objects. This may need to be deferred until we have such functionality in place, _or_ we need some way of setting these details on scene files in the API.

pickleahead:
Not sure, how common this would be. The only scenario I can think of is for non vr 3D scenes. Someone might store a 2D version for performance reasons.

This would also mess with tags like „VR“, „3D“, etc. We could only rely on file names then.

Storing the data for files is better though.

scruffynerf:
Yeah, it's possible... I have a few examples of 2Ded 3D scenes, or same scene but done in different projections.  It's rare but not unknown.

It's rare though.  Like a handful out of 10s of thousands of scenes.

The idea that there are scene tags (describing the content), and file tags (describing the file properties) makes sense.  2 files of the same scene might be encoded differently (which we track), have different resolutions (which we track), and yes, could be different forms of VR (SBS versus TB), or even 2D.

So yes, fields might make more sense than tags, but I don't see a easy way to auto detect this, and it would need to be editable if wrong.  It also might be a combo of things (oh this file is SBS, 190 degrees not 180).

At that point, I do wonder if tags are better here.

Can we just do this with file specific tags instead of fields?

pickleahead:
Guessing all this information from tags is prone to errors and it gets worse with more supported tags and tag formats. Users also wouldn't know, which tags are supported (190? 190 degrees? 190_deg? ...). Support for common tags is very convenient, but in general tags are not reliable.

scruffynerf:
> in general tags are not reliable.

While I agree... How are you going to determine this as a field?  If wrong, how does the user edit it?  In other words, fields are just as problematic and worse we have no way to edit or change them, so you are required to invent that.  Tags have issues, but we use tags now.

scruffynerf:
Also, we now have two competitive VR Stash "docker" solutions that both use tags for these properties, so Tags is doable, and with a 3rd for a built in videojs solution, a consensus as to supported tags could be reached, including aliases etc.  Installing aliases and a set of "supported" tags would be trivial, compared to the field solution.  Granted, we still need to have tags on files, not just on scenes, but seems like far less of a stretch.

pickleahead:
I am going to push some improvements soon. In general it is a lot easier to update the data through scene updates for now. We can either update the primary file or all scene files.

I guess an api for file edits is likely to happen at some point (File browser/management, management of scene and gallery files, …). So the forms can be moved to file specific forms later.

I am not opposing file tags. But it is a feature, that should be dealt on its own. There are also other open questions. E.g. would file tags be merged with scene tags?
Supporting them in this PR is not an issue. It already works for scene tags.

It also looks like setting the data for files makes auto detection by filename easier. This could be done in file (re-)scanning.

WithoutPants:
Files aren't really something that should be user-editable in general. Is there a problem with inferring this information from the filename the way that xbvr does it? https://github.com/xbapps/xbvr/blob/b7e6781ecbe81838555f2b470fc9af7edce879a3/pkg/tasks/volume.go#L224

scruffynerf:
File edits changes the hashes.  A bad thing.  Filename is fine, of course.

WithoutPants:
> File edits changes the hashes. A bad thing. Filename is fine, of course.

Sorry, I meant file entries in the database, not the files themselves.

Tweeticoats:
My preference would be to have these extra fields as you may want to be explicit in what you want with auto as the default and fall through to file name and tag detection.
Detection based on file names is good and that is what xbvr does and it works fine but as people use the renamer plugins they can potentially mess with the file names to stop this working.
Tags also works well, that is what my tool and fl0w's tool does.
Stashdb does have tags for these and a lot of the scenes have these tags [180°](https://stashdb.org/tags/103a1f16-83e1-4b9f-ab14-e85e0405e715), [190°](https://stashdb.org/tags/e3338b1d-63f8-42ea-b617-37b0acf0eea5), [200°](https://stashdb.org/tags/823219a4-41a6-48f0-ab1b-de9ed82927c3), [220°](https://stashdb.org/tags/eb6e1816-36b7-459c-9881-fb2d315d1d08), [360°](https://stashdb.org/tags/99985259-c544-412a-870c-e51e26a46cda), [Virtual Reality](https://stashdb.org/tags/6458e5cf-4f65-400b-9067-582141e2a329)

There are a few different types of projections commonly used but the pull request covers these.
On top of projection Heresphere does support some lens correction, this can be set in the player so we can probably ignore this.
[HereSphere_JSON_API_Version_1(5).txt](https://github.com/stashapp/stash/files/9762148/HereSphere_JSON_API_Version_1.5.txt)

pickleahead:
In my current state both fields are resolved in the scene resolver (database -> tags -> filename). Without any other changes in the code the result would be written into the database (video_files) on the next scene update. Basically the projection is always detected by tags and filenames for unedited scenes. The form fields are prefilled with the result. The fields can be reset, because they are nullable.

I‘d like to have your opinion first. 

Database fields could be omitted in my opinion. Managing the projection would be more complicated on user side though (filenames and correct tagging).

WithoutPants:
> In my current local state both fields are resolved in the scene resolver (database -> tags -> filename). Without any other changes in the code the result would be written into the database (video_files) on the next scene update. Basically the projection is always detected by tags and filenames for unedited scenes. The form fields are prefilled with the result. The fields can be reset, because they are nullable.
> 
> I‘d like to have your opinion first.
> 
> Database fields could be omitted in my opinion. Managing the projection would be more complicated on user side though (filenames and correct tagging).

File-based fields shouldn't be resolved/set via the scene resolvers. I think there needs to be support added for modifying file values separately before we can do this functionality effectively. Is it feasible to have the first iteration to be read-only and generated from filename/tag first, with the intent to add support for modification of files in a later PR?

pickleahead:
I fully agree on waiting for file data modification support. This data needs to be editable.

I feel like this approach got way more complicated for videojs-vr only and there are  too many things left unclear or things that could be an issue or PR on their own. Detection by filename is less reliable than tags and (scene) tags would be sufficient in most cases for videojs-vr.

Maybe #2659 could be reopened and merged instead. It is way more easy to shift to fields than to rewrite file specific fields and detection later. This way we wouldn‘t rush file mutations and keep the videojs-vr implementation simple.

Anon247:
@WithoutPants what needs to be done to get this feature released? Personally, I think a basic implementation doesn't even need the software to have an understanding of what content is VR. Just a toggle in the the video player would be more than adequate that we have to set each time the page is viewed. What are your thoughts?

WithoutPants:
A manual VR selection button is definitely a more simple and acceptable solution in the shorter term, yes.

WithoutPants:
@pickleahead @Teda1 Please see https://github.com/WithoutPants/stash/tree/vr-proto for an initial prototype for a VR button. I wasn't able to successfully test it, so if you guys want to check it out and add your input, that might get things moving.

Teda1:
I’ll find some time this week to check it out.

Teda1:
@WithoutPants do you want to put up the pull request so we can leave comments there? I'm running your `vr-proto` now and it's not clear how it's supposed to work. I don't see any VR options.

WithoutPants:
Closing in favour of #3636

PR #2664, opened 2022-06-12
Fix Synopsis json string in movie jsonschema

### LINKED ISSUE #2663: "Synopsis" key is misspelled in export of Movie records to json file.. The key is misspelled as "sypnopsis".
**Describe the bug**
"Synopsis" key is misspelled in export of Movie records to json file.. The key is misspelled as "sypnopsis".

**To Reproduce**
1. Under Settings, select Tasks -> Metadata -> Full Export
2. Open any .json file in movies directory in export folder.

**Expected behavior**
I expected the key name to be spelled "synopsis"

**Screenshots**
n/a

v0.15.0-1-ga142ec22

 - OS: Windows 10
 - Browser: Chrome
 - Version: 102.0.5005.115

n/a

**Additional context**
n/a

bnkai:
Resolves #2663

### DIFF
--- a/pkg/models/jsonschema/movie.go
+++ b/pkg/models/jsonschema/movie.go
@@ -15,7 +15,7 @@ type Movie struct {
 	Date       string        `json:"date,omitempty"`
 	Rating     int           `json:"rating,omitempty"`
 	Director   string        `json:"director,omitempty"`
-	Synopsis   string        `json:"sypnopsis,omitempty"`
+	Synopsis   string        `json:"synopsis,omitempty"`
 	FrontImage string        `json:"front_image,omitempty"`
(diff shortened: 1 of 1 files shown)


kermieisinthehouse:
Not so fast - this would break compatibility with existing movie exports... Although I'm not sure on what the right course of action is

bnkai:
> Not so fast - this would break compatibility with existing movie exports... Although I'm not sure on what the right course of action is

Thanks @kermieisinthehouse . I always export and then import so forgot about that one
With the latest commit if a movie doesnt get a synopsis during the import then we try to get the value from the old (mispelled) key.

kermieisinthehouse (approved):
Code reviewed but not tested

PR #2668, opened 2022-06-14
Separate filter buttons from query field

### LINKED ISSUE #2634: [Feature] Modify UI for Filters/Saved Filters
The current state of the Filters/Saved Filters UI has caused a considerable amount of confusion among newcomers. The Saved Filters icon in particular is positioned in such a way that it almost feels as though it's not a button.

As a feature request, I suggest that we **group the Filter/Saved Filters together** so as to isolate them from the search query area as seen in the example below.

**Current State**
currentstate

**Requested State**
1isolatedfilter

Thanks! :)

WithoutPants:
Resolves #2634 

Separates the saved and add filter buttons from the text field input control.

image


Teda1 (approved):
looks good

PR #2681, opened 2022-06-18
Add ignore option to scan task

cc1234475:
Check if the directory had a .ignore file, if so skip it.

### DIFF
--- a/internal/manager/task_scan.go
+++ b/internal/manager/task_scan.go
@@ -377,12 +377,12 @@ func walkFilesToScan(s *models.StashConfig, f filepath.WalkFunc) error {
 			if fsutil.IsPathInDir(generatedPath, path) {
 				return filepath.SkipDir
 			}
-	
+
 			ignore_exists, _ := fsutil.FileExists(filepath.Join(path, ".ignore"))
 			if ignore_exists {
 				return filepath.SkipDir
 			}
-			
+
 			// shortcut: skip the directory entirely if it matches both exclusion patterns
 			// add a trailing separator so that it correctly matches against patterns like path/.*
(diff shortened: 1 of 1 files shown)


internal/manager/task_scan.go
```diff
@@ -378,6 +378,11 @@ func walkFilesToScan(s *models.StashConfig, f filepath.WalkFunc) error {
 				return filepath.SkipDir
 			}
 
+			ignore_exists, _ := fsutil.FileExists(filepath.Join(path, ".ignore"))
+			if ignore_exists {
+				return filepath.SkipDir
```
kermieisinthehouse, line 383: Perhaps we should settle on a .stashignore name or something, if this isn't going to be configurable.

kermieisinthehouse (approved):

WithoutPants:
I think this should be configurable. Under `Exclusions` settings, something like "Exclude directories where this filename is found".

kermieisinthehouse:
I personally think it's fine - Plex doesn't make `.plexignore` configurable, for example. I'm not sure that we have the span yet that `.stashignore` shows up in archives and such - the only ones on your system will be ones you create

WithoutPants:
I think it's more about avoiding redundancy. You might already have `.ignore` or `.plexignore` files in your directory, and it would be nice to have it play nice with those instead of having to rename or add a bunch of new files.

kermieisinthehouse:
I didn't think of that usecase. Agreed.

WithoutPants:
Closing this for inactivity. This will need to be redone for the new scanning engine anyway, and it needs to be configurable.

PR #2689, opened 2022-06-22
feat: Scrapers - Path in input

TgSeed:
Hello,
This pull request makes stash to send scene's path the scraper script.

Generally, it gives several abilities to scrapers.
I ended up making this change because I made a scraper and had 'Date' field option Strategy set to 'Merge', so I don't know how and why, but at the end I had a thousand scenes with incorrect Date and no URL in scene info.
- The no URL in scene info was my mistake in coding the scraper.
- The incorrect date was my fault/unawareness while setting up the identifier task.
I wanted to fix it all in one step, but my scraper cannot make it with it just having the correct title and no url!
This is how I am here.

([The scraper](https://github.com/stashapp/CommunityScrapers/pull/1035))


WithoutPants:
We don't want to add `Path` to `SceneUpdateInput` because it's not actually read or used for this purpose.

Both `sceneToUpdateInput` and `galleryToUpdateInput` are only used by `scriptScraper`. They should be moved to `script.go` instead of `stash.go`, they should use their own type instead of `models.SceneUpdateInput`/`models.GalleryUpdateInput`, and they should be renamed to something more suitable for their use inside script scrapers. This means we can add whatever we need to the new types instead of polluting the update input types.

Tweeticoats:
The work around is to look up the scene and get the metadata that way.
Stash parses the scene id to the scraper, and from this you can call the graphql api to get the path and other metadata.
This is what lots of other scrapers use and there is python code to get this info in the py_common folder for scrapers to use.

Adding this to stash might be useful but it is not needed.

TgSeed:
> The work around is to look up the scene and get the metadata that way. Stash parses the scene id to the scraper, and from this you can call the graphql api to get the path and other metadata. This is what lots of other scrapers use and there is python code to get this info in the py_common folder for scrapers to use.

Unfortunately, I didn't know of that file and functions, so I thought it might be easier for me to simply add the path to the input data.

> Adding this to stash might be useful but it is not needed.
I guess only if the graphql api calls are not lightweight and taking some time, which is not the case here because it's a local network call, and we're not running hundreds of scripts all at once.

WithoutPants:
Closing as stale and unnecessary.

PR #2701, opened 2022-06-26
Performer/Studio/Tag/Gallery Create compnent has default name as search query

### LINKED ISSUE #2618: [Feature] New Performer/Studio/Tag/Gallery based on search input
I often search for performers, studios and tags and find zero results. It's a bit annoying copying or retype the same thing in the name field. I would like to click on the new button and the search input gets copied to the name field automatically.
Another solution would be a second new-button which contains the input string as button title.

Example searching for "High Heels":
[New "High Heels"] [New]
The first one creates with pre-filled and the original new an empty form.

TgSeed:
feat: Closes #2618
fix: New button is now available even if location.pathname ends with '/'


WithoutPants:
Can you please use more descriptive titles for your PRs.

PR #2710, opened 2022-07-03
Fix broken link in Tasks docs

m6man:
Fix "JSON specification" link which is broken


WithoutPants:
I'm rejecting this since it will break the in-app manual link. Manual links to other manual pages are in the form of `/help/<manual page>.md`. I realiase this results in broken links when following these links on the github page. We have an interceptor set up in `Manual.tsx` that converts `/help` links accordingly. Perhaps this needs to be changed to intercept relative links. That way, I think, it will work in both environments.

PR #2726, opened 2022-07-07
Release notes dialog

WithoutPants:
The front page no longer includes the changelog. Important information is no longer immediately visible when loading stash. This PR adds a release notes dialog for important messages after upgrading stash. 

The stats page is no longer an appropriate place for the changelog, so this has been moved into the settings page. The first release note prompt notifies to this effect:
image

Stash keeps track of the last release note seen and only shows the dialog if there are new notes to show.

This PR also includes migration notes, so that the user can be informed about specific issues prior to migrating. This and the release notes will be important for the file refactor PR. No migration notes are currently implemented.

This PR also includes some housekeeping, moving manual markdown files to a subdirectory of `docs/en`, and moving changelog files to the `Changelog` directory under `docs/en`.


bnkai:
There is an issue when first switching to this version. You need a couple of force-refreshes for the main page to load.
Also an issue when trying to access by middle clicking (open in a new tab) any of the stash/scene/galleries/.../tags

To reproduce open a new private window and try to access stash directly from the urls eg
<http://localhost:9999/>
<http://localhost:9999/scenes>
<http://localhost:9999/images>
<http://localhost:9999/scenes/markers>
...
Reloading or force-reloading eventually loads the page
Browsing stash normally works ok

PR #2743, opened 2022-07-14
Files refactor fixes

WithoutPants:
Target: `files-refactor`

- fixes issue where deleting a zip gallery would not remove the file entry from the database, preventing the gallery from being rescanned
- re-added `minModTime` scan filter option
- deprecated `useFileMetadata` and `stripFileExtension` scan options. These are probably better suited as scrapers, plugins or scripts.

### DIFF
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -71,10 +71,19 @@ input ScanMetaDataFilterInput {
 input ScanMetadataInput {
   paths: [String!]
 
+  # useFileMetadata is deprecated with the new file management system
+  # if this functionality is desired, then we can make a built in scraper instead.
+
   """Set name, date, details from metadata (if present)"""
-  useFileMetadata: Boolean
+  useFileMetadata: Boolean @deprecated(reason: "Not implemented")
+
+  # stripFileExtension is deprecated since we no longer set the title from the 
+  # filename - it is automatically returned if the object has no title. If this
+  # functionality is desired, then we could make this an option to not include
+  # the extension in the auto-generated title.
+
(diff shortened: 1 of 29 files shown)


WithoutPants:
Optimised post migration to remove `offset` usage. Improves performance for larger systems.

bnkai:
- [x] Zip galleries can now removed and rescanned

- [ ] Havent tested the migration optimization as i am still testing with a clean db

useFilemetadata was mostly misleading, a separate on scene/image/... create plugin would be more useful (with potential to extract exif data for images for example)
stripFileExtension while useful is very easy to move to a on create hook plugin (including a simple date parser that was requested a few times)

A couple of cosmetic bugs with the galleries (for files-refactor)
- Adding  an image in a gallery the toast message  is `Added {entity}`
- Removing an image from a gallery we have `Added images` as the message

WithoutPants:
> A couple of cosmetic bugs with the galleries (for files-refactor)
> 
> * Adding  an image in a gallery the toast message  is `Added {entity}`
> * Removing an image from a gallery we have `Added images` as the message

These exist on the `develop` branch as well. I'll make a fix there.

WithoutPants:
* fixed error when submitting scene edit draft to stash-box
* fixed issue where files with missing metadata (like image format) weren't repopulated if they were moved.

I've changed the behaviour of the file scanning and storage system:
* Where multiple candidates are possible for a file rename/move, it will choose the first as the rename source, instead of ignoring and just adding a new file.
* When deleting a scene/image/gallery, it will no longer destroy the file entry from the database unless the file is being deleted as well. The scan system has been modified to re-create scenes/images/galleries where a file has no associated object.

PR #2748, opened 2022-07-15
Fix messages when adding/removing gallery image

WithoutPants:
Corrects messages when adding/removing images to galleries.


ui/v2.5/src/components/Galleries/GalleryDetails/GalleryImagesPanel.tsx
```diff
@@ -65,10 +65,15 @@ export const GalleryImagesPanel: React.FC<IGalleryDetailsProps> = ({
...
         image_ids: Array.from(selectedIds.values()),
       });
+
       Toast.success({
         content: intl.formatMessage(
-          { id: "toast.added_entity" },
-          { entity: intl.formatMessage({ id: "images" }) }
+          { id: "toast.removed_entity" },
+          {
+            count: selectedIds.size,
```
bnkai, line 73: Minor cosmetic issue

It seems that for some reason the selectedIds/size arent reset after a successful delete.
- Remove an image from a gallery -> count = 1  ->  OK
- Remove one more image from the gallery  -> count = 2 not 1 
So once you remove an image from the gallery all the next remove messages for the gallery will be `Removed Images` , the plural one.
The remove  operations seem to complete fine. 

The same behavior is present  when adding image/s to a gallery

WithoutPants:
This is related to a latent bug in the ListHook code. The selected ids are not being reset after the add/remove operation, so are being re-sent in subsequent operations. It is going to require a deeper dive into the ListHook code (and likely some refactoring). I won't include it in this PR.

PR #2752, opened 2022-07-18
Fix non-default video stream from ffprobe result

### LINKED ISSUE #2751: Stash states wrong resolution
**Describe the bug**
When I import a 4K (3840x2160) video Stash sees it as a 720P video, the reason is that Namer picks up the embedded poster/cover art in the video file

**To Reproduce**
1. Go open a scene
2. Click on file info
3. See error

**Expected behavior**
That Stash reports the file as a 4K video

**Screenshots**
Stash:
https://media.discordapp.net/attachments/559159910550732809/998584986313703584/Stash.png
MediaInfo:
https://media.discordapp.net/attachments/559159910550732809/998585286172872775/MediaInfo.png?width=388&height=701

v0.16.0
Build hash:
351472f6

**Additional context**
[ffprobe.txt](https://github.com/stashapp/stash/files/9133667/ffprobe.txt)

We were discussing the problem on Discord:
https://discord.com/channels/559159668438728723/559159910550732809/998584952281108592

peolic:
~Should fix~ Fixes #2751.
As far as I can tell, ffmpeg/ffprobe sets the disposition for each stream.
https://github.com/FFmpeg/FFmpeg/blob/HEAD@%7B2022-04-28%7D/fftools/ffmpeg_opt.c#L2229-L2230


bnkai:
This needs a bit more testing.
Isn't the referenced code only when copying metadata through ffmpeg?
I think we need to get some more ffprobe meta from various containers / files to be sure.
From a quick look (from my phone atm) at my library I think some mpeg files have no streams with default streams ( according to `ffprobe file.mpg -show_streams ` ). If that is the case I think the index function should be changed so that it returns a default stream index if the default disposition is set but use the index it returns now if no default is set.

PR #2763, opened 2022-07-21
[Files Refactor] Import export fixup

WithoutPants:
Re-adds the import/export implementation and rewrites to include support for the file/object separation.

Related files and folders are now written to the `files` subdirectory and scenes, images and galleries now reference files by their full path (or gallery title for user-created galleries).

The schema and general layout is changed and is incompatible with the previous layout and schema. The `mappings.json` file is removed. Instead, all files will be read from the subdirectories.


bnkai:
After a very short test.


```json
  "created_at": null,
  "updated_at": null,
```
that is the same for all `files`. If it is not needed shouldnt it be skipped and not exported?

 
```json
    {
      "type": "phash",
      "fingerprint": -659148031732383170
    }
```
phash is exported as a number instead of a string ?


Manual created galleries are exported using only their name. If you create a couple of couple of galleries with the same name then only one gets exported ( or the second overwrites the first ? havent checked the code yet).
The folder galleries include the gallerry id so there should be no problem. The zip galleries include the checksum, what happens though if we have duplicate zip files with the same name in different directories?

internal/manager/task_export.go
```diff
@@ -124,36 +138,44 @@ func (t *ExportTask) Start(ctx context.Context, wg *sync.WaitGroup) {
...
+	if t.baseDir == "" {
+		logger.Errorf("baseDir must not be empty")
+		return
+	}
+
 	t.json = jsonUtils{
 		json: *paths.GetJSONPaths(t.baseDir),
 	}
 
+	_ = fsutil.EmptyDir(t.baseDir)
```
bnkai, line 150: This is a little dangerous. What if a user chooses a non empty dir as his stash metadata path? Or what if he selects the same directory with the generated path? On first full export everything will be deleted...
We could warn the user not to select the same dir (some users may have done that already though) or just create a new function that deletes only the json related dirs(if they are present) and do something like `paths.EmptyJSONDirs(t.baseDir)`

WithoutPants:
Thanks for the feedback.

> After a very short test.
> 
> ```json
>   "created_at": null,
>   "updated_at": null,
> ```
> 
> that is the same for all `files`. If it is not needed shouldnt it be skipped and not exported?

Fixed. Note that this requires the 32 schema migration to be re-run, or the query views to be manually recreated.

> 
> ```json
>     {
>       "type": "phash",
>       "fingerprint": -659148031732383170
>     }
> ```
> 
> phash is exported as a number instead of a string ?

Fingerprints are exported in whatever the json marshaller marshals them as, since fingerprint values are stored as a generic `interface{}` values. I don't believe this should be an issue.

> Manual created galleries are exported using only their name. If you create a couple of galleries with the same name then only one gets exported ( or the second overwrites the first ? havent checked the code yet). The folder galleries include the gallerry id so there should be no problem. The zip galleries include the checksum, what happens though if we have duplicate zip files with the same name in different directories?

All gallery filenames now include the id number.

The code no longer calls `EmptyDir` on the metadata directory. It now runs it on the individual subdirectories as suggested.

PR #2767, opened 2022-07-22
Add Create and Tagger sections to the Tasks page

7dJx1qP:
image

This adds two new task groups, "Create" and "Tagger", under the Library section after the Scan task. The create group is a collection of the "New" buttons one finds on the respective object pages, but all collected in one place. The tagger group has buttons that take you to the scene tagger view and the performer tagger view.

The idea behind this came about from a thread on the discord about improving the newbie experience. A common issue is new users will Scan and then try to Identify or Autotag and get frustrated when nothing happens. They don't know they should be creating objects first or will never discover where the Tagger is, because it's hidden on the scene and performer pages. Allowing new users to discover the create and tagger capabilities from the tasks page and placing it right after the scan section makes for a much more logical progression since those are the things we would recommend new users do after a scan.

### DIFF
--- a/ui/v2.5/src/components/Settings/Tasks/LibraryTasks.tsx
+++ b/ui/v2.5/src/components/Settings/Tasks/LibraryTasks.tsx
@@ -375,19 +375,34 @@ export const LibraryTasks: React.FC = () => {
         <Setting
           heading={
             <>
-              <FormattedMessage id="config.tasks.tagger.heading" />
+              <FormattedMessage id="config.tasks.scene_tagger.heading" />
               <ManualLink tab="Tagger">
                 <Icon icon={faQuestionCircle} />
               </ManualLink>
             </>
           }
-          subHeadingID="config.tasks.tagger.description"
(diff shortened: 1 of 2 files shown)


7dJx1qP:
image

Split the scene and performer taggers so they can have their own subheaders. Our help docs could also use a page for the Performer Tagger. Right now I just have it bringing up the scene tagger help doc.

HijackHornet:
If i may, i find that it was better when scene and performer tagger where grouped.
The settings pages are getting bigger and bigger, and i'm not sure adding shortcuts to what would take 2 clicks to reach, should take so much space?

WithoutPants:
The tasks page is meant for long-running tasks, not for shortcuts to other pages. Having these shortcuts dilutes the intent of the Tasks page, and makes the button behaviour inconsistent - the user won't know if a button will start a task or navigate to another page. I understand that there's an issue to be solved here, but I don't think this is the correct approach.

PR #2781, opened 2022-07-28
#1810 Truncate large numbers on buttons

### LINKED ISSUE #1810: [Feature] Truncate large numbers on buttons
The numbers displayed in the button class of cards that display the count of scenes, images, galleries, movies, etc can become large and crowd the card.

Add an option in the interface settings to display truncated numbers larger than 999 to one decimal place and attach K, M, and B accordingly to denote thousands (K), millions (M), and billions (B). Similar to https://stackoverflow.com/posts/35329932/revisions

Examples:

Number | Shortened Value
--- | ---
100 | 100
1,000 | 1K
1,100 | 1.1K
23,300 | 23.3K

JoeScylla:
Resolves #1810.

- Added a configuration setting to format counts or not
- Added utility function to format counts into an object
- Added React.FC Counter to render counts
- Edited studio, performer and tag details page to use the new React.FC Counter component

Disclaimer: first time with Go and React. Its working, but i am not sure if everything was done the right way / based on best practice.


coscosname:
I don't know if it is planned, but the names of numbers should be localizable because not all languages use the Latin alphabet. In addition, some countries use the short scale and others the long scale, which also affects the names.

WithoutPants (approved):

PR #2789, opened 2022-07-31
[Feature] Exif auto tagging

HijackHornet:
# Scope
Add exif reading for compatible file types (tiff, jpg, jpeg...) when autotagging.
It is lightweight and only adds 2 functions.

It went through unit tests and lint.


pkg/match/path.go
```diff
@@ -15,14 +22,16 @@ import (
 )
 
 const (
-	separatorChars   = `.\-_ `
-	separatorPattern = `(?:_|[^\p{L}\w\d])+`
+	separatorChars       = `.\-_ `
+	separatorPattern     = `(?:_|[^\p{L}\w\d])+`
+	separatorPatternExif = `(?:_|[^\p{L}\w\d\,])+`
```
HijackHornet, line 27: Slightly different pattern because tags are separated by comma so we must not remove them

.gitignore
```diff
@@ -58,3 +59,5 @@ node_modules
 /stash
 dist
 .DS_Store
+
+
```
HijackHornet, line 63: Idk why they kept being added... Sorry for the 2 extra line return at the end of the file

bnkai:
From a quick look.

I am not sure if it the right place to add the exif code.

Anyway even if that is not the issue there are issues...
- Tagging with exif should be user configurable (not all users have useful tags in their images)
- The way you add the tags will slow down the tagging procedure a lot.
You are opening every file and rely on the exif library to error out if its not a supported file or no tags are present. Even a plain file open takes some time.
You could for example check the path first to see if the file is a supported one (jpg,jpeg,tiff) and only then open and decode
- What happens with images in zip files?

HijackHornet:
>     * The way you add the tags will slow down the tagging procedure a lot.
>       You are opening every file and rely on the exif library to error out if its not a supported file or no tags are present. Even a plain file open takes some time.
>       You could for example check the path first to see if the file is a supported one (jpg,jpeg,tiff) and only then open and decode
Done in 458e8b
>     * What happens with images in zip files?
I'm not sure.. It autotagging applied to images in zip files ? I'm only adding a step to the existing system. If the current implementation of autotagging doesnt apply to files in zip, exif wont do either. 
> Tagging with exif should be user configurable (not all users have useful tags in their images)
That i totally agree. Though you have to understand that i'm already not a Go developper and had to learn it to do this PR.
For the config part, again, i never worked with React. And while i'm fluent with Angular, i dont understand a thing to the ui code. So ideally, someone here could step in and add the checkbox ? Isnt it what collaborative work means ?

Finally, as for where the code is placed. I originaly created its own file, but then someone mentioned that Autotagging was soon to be moved out to a plugin. So i went in favor of simplicity, in order not to get denied on the terms of maintainability.

HijackHornet:
Lets close this PR for now

MyDirtyAccount:
What about showing EXIF data alongside images in the image view, under the _File Info_ tab?

PR #2795, opened 2022-08-02
Prevent constantly switching sources on scrub

gerit1a:


gerit1a:
new pr to fix hls streaming coming soon which obsoletes this

PR #2803, opened 2022-08-04
Fix HLS and adapt mobile better for touches

gerit1a:
Backend:
- Fix HLS so connections are terminated correctly
- Improve transcoding for on the fly transcoding
- Fix csp so hls plays correctly

Frontend:
- Touches on video start preview now
- Thumbnails now follow with the touch
- Thumbnail is on top of other elements now
- Thumbnail disappears when no longer touching

Help:
Need to figure out how to best remove css effects from non touch elements so they don't conflict
Is there a better way to fix the thumbnail library rather than embedding it directly?


WithoutPants:
I would suggest breaking this PR into smaller pieces; there's a hell of a lot going on here to review.

WithoutPants:
I've made changes to #2322 to integrate the fixes you made in this branch. Please suggest changes to that branch or make a PR against that branch instead of re-creating the changes in a separate PR.

WithoutPants:
As mentioned above, if you want these improvements merged, then please raise more focussed PRs.

PR #2157, opened 2021-12-21
Image rater

archtasse:
Implements #2156

PR #2830, opened 2022-08-16
Stash rating system

### LINKED ISSUE #1962: [Feature] Half Stars / 0-10 Rating System
I think it's self explanatory, below you'll find an explanation for my reasoning of why this should be a thing.

Your personal collection is very likely centered around the things you like, or as the app name implies, it might have a lot of stuff that you stashed and haven't properly looked at yet. 

If we follow the first line of thought, given it's already centered around things you like, and if you belong to the subset of people that rate entries, then you're very likely to keep rating stuff probably at a 3/5 or above. 
And if we follow the second line of thought, if you were to find something stashed in your collection you're thinking on rating a 1/5, you're also likely thinking about deleting it altogether, that's why the delete option is there after all.

Under this pretense, I don't find it far fetched to assume that the people that rate entries, are likely rarely using the full spectrum of star ratings, keeping it all in the 3/5-5/5 realm, and that in a sense is limiting, because you're likely going to have to blend your "9/10"s and "10/10"s together as 5*. 

Sure, you could adopt a standard for yourself where you keep 1* as a "6/10" and 5* as a "10/10", but it's far from elegant.
Personally, if I'm going to rate things like movies or games, I prefer to have the option to give even half points (i.e. 8.5/10), but quarter stars would look a bit silly, and changing the feature idea to support different systems you can pick from (i.e. 5 stars, 0/10 full numbers, 0/10 with a single decimal) would be quite the step up in work required from half stars.

This might look a bit silly of a request, but if effort was put into the ability to give ratings, might as well go the extra mile.

skier233:
Doesn't need to merge into this branch (can wait until files-refactor goes in) but this allows the pr to be reviewed now.

EDIT:
No longer relevant now

Resolves #1962


ui/v2.5/src/components/Scenes/SceneDetails/RatingSystem.tsx
```diff
@@ -0,0 +1,312 @@
...
+import StarBorderIcon from "@mui/icons-material/StarOutline";
+
+export interface IRatingSystemProps {
+  value?: number;
+  onSetRating?: (value?: number) => void;
+  disabled?: boolean;
+}
+
+export interface IRatingStarsProps {
+  value?: number;
```
skier233, line 15: There's technically now a bug here where if you change the rating in the edit menu and switch to the details menu, the details menu will show the old value unless the page is refreshed. This doesn't show up for the number rating but it does for the star rating which is strange. Not currently sure what could cause that.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -447,6 +447,18 @@
           "description": "Remove the ability to create new objects from the dropdown selectors",
           "heading": "Disable dropdown create"
         },
+        "rating_system": {
```
skier233, line 450: other languages will need to be added.

HijackHornet:
Can you please describe the features being added ?
Also, is this similar to https://github.com/stashapp/stash/pull/2157 ?

HijackHornet:
I did a few tests and found some issue :
- In numerical point system, there is some missing Css 
image
- In `Five start system` (default), it still show the `Five stars with half increament` frontend.
- There is no option to turn off the rating system. Just adding a "None" in the config option that would hide the rating from the front end would be enough.

skier233:
> I did a few tests and found some issue :
> 
> * In numerical point system, there is some missing Css
>   image
> * In `Five start system` (default), it still show the `Five stars with half increament` frontend.
> * There is no option to turn off the rating system. Just adding a "None" in the config option that would hide the rating from the front end would be enough.

I'm not seeing this bug about the five star system. Can you describe exactly what the issue is? When you select 5 stars then it shows half star increments on the stars? I don't seem to see that issue.

HijackHornet:
Exactly that yes.

skier233:
> Exactly that yes.

Hmm. I'm not able to repro that.

skier233:
> I did a few tests and found some issue :
> 
> * In numerical point system, there is some missing Css
>   image
> * In `Five start system` (default), it still show the `Five stars with half increament` frontend.
> * There is no option to turn off the rating system. Just adding a "None" in the config option that would hide the rating from the front end would be enough.

Fixed the css. I'm not sure about disabling the rating system altogether since that would need to occur on each caller of the rating system. However, if people don't want to use the rating system they can just not rate anything and that should work fine.

skier233:
> Exactly that yes.

Wasn't able to repro but I think I found why you're seeing it. Should be fixed now if you want to try again.

HijackHornet:
I still would prefer if we could just add the rating from the views if the rating system is set to "None".
I dont like rating systems, and i can see it be used by some, but it would be better if toggleable. At least on the front side.

skier233:
> I still would prefer if we could just add the rating from the views if the rating system is set to "None". I dont like rating systems, and i can see it be used by some, but it would be better if toggleable. At least on the front side.

Added none rating system.

ui/v2.5/src/components/Scenes/SceneDetails/RatingSystem.tsx
```diff
@@ -0,0 +1,363 @@
...
+    if (!useValidation && props.onSetRating != null) {
+      e.target.value = Number(val).toFixed(1);
+      setInput(Number(val).toFixed(1));
+      setPrevious(Number(val).toFixed(1));
+      let tempVal = Number(val) * 10;
+      props.onSetRating(tempVal != 0 ? tempVal : undefined);
+      setValidation(true);
+      return;
+    }
+    const match = /(\d{0,1})(\d{0,1})(.?)((\d{0,1})?)/g.exec(val);
```
Teda1, line 57: Seems like this could be simplified to `(\d)(\d)(.?)((\d)?)`

ui/v2.5/src/components/Scenes/SceneDetails/RatingSystem.tsx
```diff
@@ -0,0 +1,363 @@
...
+    const matchOld = /(\d{0,1})(\d{0,1})(.?)((\d{0,2})?)/g.exec(previous ?? "");
```
Teda1, line 58: This one simplified to `(\d)(\d)(.?)((\d{0,2})?)`

Teda1: On a side note, it seems like it might have been easier to just convert the string to a number and validate the number instead of the string. This would also help the readability of the code.

Teda1 (commented):
This looks great. I didn't finish going through all the code yet, but here are some comments on what I've seen so far.

I was not able to build the code initially. Unfortunately, I did not capture the error I was receiving, but running `yarn add @mui/material` and ``yarn add @mui/icons-material` bumped up versions, resolving the issue. I think you may have to rerun those commands.

Looking through the pull request, I saw that the code guards against rating numbers larger than 100 on the front end, but it is still possible for anyone to specify values larger than 100 using the graphql endpoints. Though I don't think this is a regression. But it might be worth specifying a range for these values in the database table.

ui/v2.5/src/components/Scenes/SceneDetails/RatingSystem.tsx
skier233: changed

ui/v2.5/src/components/Scenes/SceneDetails/RatingSystem.tsx
skier233: I had thought about this and originally was doing something like this. But, I'm not just validating the number here, I'm doing some fancy stuff to move the cursor around and change the input based on where the user input a number. So, to do it with a number means I'd have to do a bunch of mod division to get the value of certain digit/decimal places and honestly I think that would be less readable but definitely still open to other options on this one. I agree its not super readable atm but just struggling to think of a much better way.

skier233: > This one simplified to `(\d)(\d)(.?)((\d{0,2})?)`

Simplified

skier233:
> This looks great. I didn't finish going through all the code yet, but here are some comments on what I've seen so far.
> 
> I was not able to build the code initially. Unfortunately, I did not capture the error I was receiving, but running `yarn add @mui/material` and ``yarn add @mui/icons-material` bumped up versions, resolving the issue. I think you may have to rerun those commands.
> 
> Looking through the pull request, I saw that the code guards against rating numbers larger than 100 on the front end, but it is still possible for anyone to specify values larger than 100 using the graphql endpoints. Though I don't think this is a regression. But it might be worth specifying a range for these values in the database table.

Good find. I updated all the sql tables to not allow values <0 or >100.

Teda1:
The latest force push makes it a bit harder to identify what code belongs to this pull request. I test the code branch, and things appear to be working fine after the push.

If possible, I would leave the CSS for the rating ribbon alone since it's unnecessary for this work. The change will likely create more work for people who already have their own CSS.

Also, I think it would be good to migrate people's existing ratings to fit into this new 100 point system. People who have already rated their content on the old system likely won't be too happy to fix their ratings for the new system.

skier233:
> The latest force push makes it a bit harder to identify what code belongs to this pull request. I test the code branch, and things appear to be working fine after the push.
> 
> If possible, I would leave the CSS for the rating ribbon alone since it's unnecessary for this work. The change will likely create more work for people who already have their own CSS.
> 
> Also, I think it would be good to migrate people's existing ratings to fit into this new 100 point system. People who have already rated their content on the old system likely won't be too happy to fix their ratings for the new system.

Yea. The migration script should be in there working correctly to migrate existing ratings. I can pull the css changes out. I thought it made it more readable but I can just make that a local custom css or something.

Teda1:
Hmm, a few hours ago I migrated a database with 4 star rating in the old system. After the database migration the scene still had a rating of 4 instead of 80. I can try again later to ensure I didn’t miss something.

Teda1:
Personally the issue I see with the ribbon CSS update is that it now uses the stash blue color scheme which essentially requires an update for any custom theme.

skier233:
> Personally the issue I see with the ribbon CSS update is that it now uses the stash blue color scheme which essentially requires an update for any custom theme.

I can change the color to a more generic one. I just chose that color blue because it had the most contrast of any color on the image backgrounds. That doesn’t matter as much though when you have the ribbon background.

skier233:
> Hmm, a few hours ago I migrated a database with 4 star rating in the old system. After the database migration the scene still had a rating of 4 instead of 80. I can try again later to ensure I didn’t miss something.

Hmm. If so that seems like a bug so if you are able to reproduce this then that would be super useful.

skier233:
> Personally the issue I see with the ribbon CSS update is that it now uses the stash blue color scheme which essentially requires an update for any custom theme.

Changed the rating back to white to be generic across themes.

Teda1:
Ahh, I see what was throwing me off. We had rating-5, rating-4, rating-3, rating-2, and rating-1 CSS classes which provided different colors depending on the rating. Under your change, those colors were missing for obvious reasons. We should try to restore that behavior in your change. So the rating-1 equivalent would be rating from 0-20, rating 2 20-40, etc. Those classes make it easier to identify better-rated scenes, galleries, or performers at a glance.

Also, I was able to test the rating migration and confirm your conversion SQL does work. I may I migrated a bad database when I tested yesterday.

skier233:
> Ahh, I see what was throwing me off. We had rating-5, rating-4, rating-3, rating-2, and rating-1 CSS classes which provided different colors depending on the rating. Under your change, those colors were missing for obvious reasons. We should try to restore that behavior in your change. So the rating-1 equivalent would be rating from 0-20, rating 2 20-40, etc. Those classes make it easier to identify better-rated scenes, galleries, or performers at a glance.
> 
> Also, I was able to test the rating migration and confirm your conversion SQL does work. I may I migrated a bad database when I tested yesterday.

Thats good on the sql migration part. I’ll look into the css styling tomorrow. Ideally, it’d be nice to get 10 different colors for 0-10,10-20,20-30, etc. But yea, I’ll look into this.

Teda1:
10 different colors would be even better 👌 .

AdultSun:
I'm remembering a request now the last time rating system improvements were brought up, regarding a change from the banner to stars. The idea I remembered getting some positive reception was color-coding the stars using metallics instead of the current colors. So five stars could be gold, four stars silver, three stars bronze, not sure what 1 and 2 would be. Can't quite get to 10 different shades that way, but it's another option.

As long as the colors can be easily changed with CSS I guess the specifics don't matter as much.

skier233:
> > 10 different colors would be even better ok_hand .
> 
> If we're going to have so many colors, please have the color shades make sense, like a cool-to-hot heatmap.... Rather than having 1 be yellow and 2 be blue and 3 be red, etc.

Yea. Sounds like a plan. I’ll look into some nice colors tomorrow.

Teda1:
+1 I was thinking the colors would still fall in the same range as the current system. At 10 options instead of 5, it would just have more depth to the transitions. The color range is:
```
.rating-5 {
  background: #ff2f39;
}

.rating-4 {
  background: $red1;
}

.rating-3 {
  background: $orange1;
}

.rating-2 {
  background: $sepia1;
}

.rating-1 {
  background: $dark-gray5;
}
```

AdultSun:
> +1 I was thinking the colors would still fall in the same range as the current system. At 10 options instead of 5, it would just have more depth to the transitions.

That level of contrast between options makes sense when there's only 5 of them, but if we're doubling it to 10 then we may need to increase the range so they're still visibly different at a glance.

Teda1:
I'm not sure a 10 needs to be that visibly different from a 20. To me, what's important is that a 0-20, 20-40, 40-60, 60-80, and 80-100 are distinguishable. A wider range could be done though. I'm just unsure what color outside of the spectrum we currently have makes sense.

AdultSun:
Hard to say without seeing it in action, but like I said the specifics probably aren't too important if it's all exposed for easy CSS overrides. For both themes and personal customization, obviously. For the default, less obtrusive is likely better.

skier233:
Thoughts on this coloring system?

<details>
Capture

skier233:
> > Thoughts on this coloring system?
> > Capture
> 
> I think a red to green progression without the blues would be better. red = bad, green = good. Something like:
> 
> image
> 
> or
> 
> image
> 
> or your colors but reversed, with the logic being a progression from "cold" colors (blues, low rating) to "hot" (red, high rating)
> 
> image

The issue is thats a whole lot of different shades of yellow in the red to green progression. 
I like the idea of red being the lowest since usually red is associated with "bad" (error messages in red, stop lights, etc)

skier233:
> > > Thoughts on this coloring system?
> > > Capture
> > 
> > 
> > I think a red to green progression without the blues would be better. red = bad, green = good. Something like:
> > image
> > or
> > image
> > or your colors but reversed, with the logic being a progression from "cold" colors (blues, low rating) to "hot" (red, high rating)
> > image
> 
> The issue is thats a whole lot of different shades of yellow in the red to green progression. I like the idea of red being the lowest since usually red is associated with "bad" (error messages in red, stop lights, etc)

@Teda1, @AdultSun, @sleetx, thoughts?

AdultSun:
Typically, I'd say you're right that the warmer colors would read as "bad" and cooler colors read as "good", but in this case the reverse makes more sense to me. "Red hot" in our context would mean good and scenes that leave you "cold" would be a turn-off. But if the range is just green to red, I think the temperature context isn't there anymore. Now we're just using traffic light colors, essentially, so green is good and red is bad.

Though it feels weird to say, if the range is blue to red I'd want red as the high ratings, but if the range is green to red I'd want green as the high rating.


As for which color scheme, I'm not sure which I'd prefer. Red-Green feels more obvious in it's meaning (green good, red bad) but doesn't give us as large a range for 10 distinct shades as Blue-Red. Another option would be Blue-Red with shades of purple in between instead of greens and yellows, but I doubt that would be better than the other two options. You'd have the same limited range as Red-Green (probably more-so actually) and I'm not sure it's any clearer in it's meaning (blue bad, red good), but it would bring back the idea of temperature instead of traffic lights.

I'm leaning towards the same Blue-Red range you have above, but reversed with red as the high ratings. The broader range makes it easier to read at a glance.

skier233:
> I agree with @AdultSun on both points -
> 
> If you're using traffic light colors, then red-to-green is fine. The excess of yellow isn't a huge deal because it just signifies the rating is somewhere in the middle.
> 
> If you decide on a wider "temperature"-based color spectrum, blue and green should be "cold" low-rated colors and orange/red should be "hot" high-rated colors. I prefer this option.
> 
> Also I would recommend making the transparent rating colors opaque. Some colors, particularly yellows, seem to clash with the scene background, but the clashing color will be variable depending on the scene.

It's a bit hard to see from the image but I already did make them all opaque. I think it looks better than solid backgrounds.

skier233:
What about these colors then:
Capture

DogmaDragon:
You should censor all NSFW stuff on Github.

skier233:
> You should censor all NSFW stuff on Github.

It doesn't let you put an image inside a censor. I tried :/

Teda1:
You can add a CSS filter to blur the image before you capture the screenshot. I'm personally not a fan of the transparency that was added to the banner. I'd still lean toward keeping it as it was and let others modify the CSS past that if they want to. In terms of the transition, I would also say to leave the dull gray to vibrant red that we had before. The bright purple doesn't suggest a bad rating to me.

Teda1:
Here is an image of how the current colors flow:
Screenshot 2022-09-08 173531

skier233:
> Here is an image of how the current colors flow: Screenshot 2022-09-08 173531

yea but this is kind of difficult to turn into 10 colors. It isn't actually following any sensible gradient except for the orange to red.

Teda1:
The transition in between the current colors we have can be subtle. Can just just pick a color code that falls in between the 2.

Teda1:
I think the main point of having the 10 CSS classes is that it will give people more options for custom CSS.

skier233:
What about this?
Capture

Teda1:
> What about this? Capture

This looks better

skier233:
> This looks better

Using this site which lets you make gradients:
https://cssgradient.io/

Teda1:
Thanks for the link. The site could come in handy later.

skier233:
Here is what seems pretty close to a final version. Since its so easy to do, I added a different color for the every 5 instead of every 10 so its a pretty true gradient now:
Capture

WithoutPants (changes requested):
I haven't exhaustively tested and reviewed all of this, but these are my notes so far.

- the rating should not be shown if it is unset in the details pages. Currently it is showing 5 empty stars
- the ratings stars are not vertically aligned to the middle on the edit pages
- the detail tab doesn't appear to update the rating after submitting a rating change in the edit scene tab
- there's something wrong with the display of the rating system. Even though config is returning `five_pointfive_stars`, it is displayed as `Five star system`
- I rated something 3.5 stars, and it's displayed the rating banner as `70`.
- I don't think it's worth rebuilding the object tables just to add a check for the rating values. This can and should be validated in the software anyway.

I'm not sold on the current overall design. It's one thing to modify the schema to accept rating values from 0 to 100, but this is changing meaning of the `rating` fields of the _API_. This means that any external scripts etc which expected a rating value of 1-5, now need to be rewritten to accomodate the new values. This applies to filters as well. For the existing meaning of `rating`, the value _must_ be a value between 1 and 5. If we want to be accepting a using values between 1 and 100, then it should probably be a separate field, and the legacy field should translate to/from the new field.

I think we need to scrutinise _every single option_ for rating systems that we add. Every single system needs to be maintained once it is added. I am really skeptical that we need this many options. 

I don't understand the purpose of adding a None value. If we don't want to show the ratings field, there are ways to do this in custom css, and it adds nothing otherwise.

Is it absolutely necessary to add new UI dependencies for this functionality?

WithoutPants:
Also, please include a full description of the changes in your pull requests. I shouldn't need to hunt through other issues or the changed files to get a handle on what has changed.

skier233:
> Also, please include a full description of the changes in your pull requests. I shouldn't need to hunt through other issues or the changed files to get a handle on what has changed.

It was at the top in one of the comments: https://github.com/stashapp/stash/pull/2830#issuecomment-1220620303

skier233:
> I haven't exhaustively tested and reviewed all of this, but these are my notes so far.
> 
> * the rating should not be shown if it is unset in the details pages. Currently it is showing 5 empty stars
> * the ratings stars are not vertically aligned to the middle on the edit pages
> * the detail tab doesn't appear to update the rating after submitting a rating change in the edit scene tab
> * there's something wrong with the display of the rating system. Even though config is returning `five_pointfive_stars`, it is displayed as `Five star system`
> * I rated something 3.5 stars, and it's displayed the rating banner as `70`.
> * I don't think it's worth rebuilding the object tables just to add a check for the rating values. This can and should be validated in the software anyway.
> 
> I'm not sold on the current overall design. It's one thing to modify the schema to accept rating values from 0 to 100, but this is changing meaning of the `rating` fields of the _API_. This means that any external scripts etc which expected a rating value of 1-5, now need to be rewritten to accomodate the new values. This applies to filters as well. For the existing meaning of `rating`, the value _must_ be a value between 1 and 5. If we want to be accepting a using values between 1 and 100, then it should probably be a separate field, and the legacy field should translate to/from the new field.
> 
> I think we need to scrutinise _every single option_ for rating systems that we add. Every single system needs to be maintained once it is added. I am really skeptical that we need this many options.
> 
> I don't understand the purpose of adding a None value. If we don't want to show the ratings field, there are ways to do this in custom css, and it adds nothing otherwise.
> 
> Is it absolutely necessary to add new UI dependencies for this functionality?

-This is how it worked before. Why would we want to change that behavior?
-I can take a look at this
-It may require a page refresh but it will be there
-That's odd. I haven't really seen that issue. I'll investigate.
-Yes. That is the expected behavior. The banner should show the 100 point system.
-If we don't want to add a check in sql that's fine. Can be removed. It does help ensure the db can't ever get in a bad state though.
I updated filters. The reality is this is how it has to be done to add this functionality. You can't map a system with more detail (0-100) onto one with less detail (1-5) so it has to be changed in the backend.
In terms of maintenance, these different options are really just a few different configurations of only 2 different options so there really isnt much maintenance here.
I wasn't going to add "None" but many many people wanted this functionality so I added it. See comments above.
The new UI dependencies were needed to do the half stars.

.gitignore
```diff
@@ -58,3 +58,14 @@ node_modules
...
+*.yml
+/ffmpeg-release-essentials.zip
+/icon.png
+/package-lock.json
+/package.json
+/stash-go.sqlite
+/stash-go.sqlite-wal
+/stash-go.sqlite-shm
+/yarn.lock
+/generated/thumbnails
```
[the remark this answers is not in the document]
skier233: I've always used the idealogy that its better to have all the generated files be in the gitignore to ensure they dont get pushed but if you don't want them here I can remove them.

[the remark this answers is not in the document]
skier233: updated

internal/manager/config/config.go
```diff
@@ -161,6 +161,8 @@ const (
 	DisableDropdownCreatePerformer = "disable_dropdown_create.performer"
 	DisableDropdownCreateStudio    = "disable_dropdown_create.studio"
 	DisableDropdownCreateTag       = "disable_dropdown_create.tag"
+	RatingSystem                   = "rating_system"
+	defaultRatingSystem            = "five_pointfive_stars"
```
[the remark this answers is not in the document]
7dJx1qP: We shouldn't assume all existing users would see this as an improvement. It should be up to the user whether they want to change the rating system from what it currently is.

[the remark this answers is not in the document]
skier233: Well what could possibly be the downside of this? If you don't want to use the half stars then you just dont click them. The reality is many users wont be aware that it's even possible to change the rating system so if we don't change it by default then many users will never see the benefits of these changes.

7dJx1qP:
There's an inconsistency with the default rating system shown in the settings dropdown and the actual default rating system used when you go to edit a scene's ratings.

The UI shows five star system, but when I go to edit a scene's ratings it's using the half star UI.
image

skier233:
> There's an inconsistency with the default rating system shown in the settings dropdown and the actual default rating system used when you go to edit a scene's ratings.
> 
> The UI shows five star system, but when I go to edit a scene's ratings it's using the half star UI. image
> 
> If I switch the rating system setting to something else and then back to 5 star the rating system UI uses the correct whole star UI, so it looks like it's just an inconsistency with the initial dropdown value. I tested using a fresh install.

I think WithoutPants saw this too. Useful to know that it only happens on the fresh install. I'll look into this.

7dJx1qP:
> -Yes. That is the expected behavior. The banner should show the 100 point system.

The banner should show the values corresponding to the selected rating system. The legacy system shows 1-5 because it's a 1-5 star, but if someone changes to 1-10 they would expect to see 1-10. There's no reason to expose users to the underlying 1-100 values.

And speaking of 1-100, it's odd to not have it as one of the rating options. I see there's a 10 point decimal which is equivalent to 1-100, but I find 1-100 is easier to input values for than decimal point values.

skier233:
> > -Yes. That is the expected behavior. The banner should show the 100 point system.
> 
> The banner should show the values corresponding to the selected rating system. The legacy system shows 1-5 because it's a 1-5 star, but if someone changes to 1-10 they would expect to see 1-10. There's no reason to expose users to the underlying 1-100 values.
> 
> And speaking of 1-100, it's odd to not have it as one of the rating options. I see there's a 10 point decimal which is equivalent to 1-100, but I find 1-100 is easier to input values for than decimal point values.

Regarding the banners, these should use the same number system as filters. Filters will need to use 0-100 because they go directly to the backend which is why this should probably stay as it is. 

It would need a different code system to do 1-100 which is more to maintain as @WithoutPants mentioned. The 0.0-10.0 system is probably more intuitive for most people so I think this should be preferred. A 0-100 system could always be added later though.

Teda1:
When I was investigating this work initially, I planned on changing the rating datatype from an int to a datatype that supports decimal numbers. Doing so would have allowed for finner ratings while also keeping the 1-5 range intact and preventing some headaches. I personally wouldn't see a need for a 1-10 star system when we can get the precision we need within the 1-5 system by adjusting the data type. To me, more stars just equal more clutter. But it doesn't hurt to have that option if you are up for the work.

7dJx1qP:
> Regarding the banners, these should use the same number system as filters. Filters will need to use 0-100 because they go directly to the backend which is why this should probably stay as it is.

Ideally, filters should use the selected rating system and it would be translated to 0-100 before going to the backend, so it's a consistent experience to the user. I can see this being an area of confusion for people looking to filter by ratings. The banners using a 0-100 gives some hint, but it's still not clear and users shouldn't be exposed to the 0-100 whether it's the banner or the filter.

skier233:
> > Regarding the banners, these should use the same number system as filters. Filters will need to use 0-100 because they go directly to the backend which is why this should probably stay as it is.
> 
> Ideally, filters should use the selected rating system and it would be translated to 0-100 before going to the backend, so it's a consistent experience to the user. I can see this being an area of confusion for people looking to filter by ratings. The banners using a 0-100 gives some hint, but it's still not clear and users shouldn't be exposed to the 0-100 whether it's the banner or the filter.

I don't think this is really possible as filters are implemented. Also, this would completely break any saved filters if you switch rating systems. If you have an idea of how this can be implemented, I'm not opposed to the idea if it could be implemented well but I looked into this for quite some time and it doesn't look possible with the current filter system.

Teda1:
> Well its already all implemented as is. I personally think using a 1-100 system under the hood is by far the best internal implementation and I've heard many people agree. Decimals are messy and unclear. For example, doing a 1-100 system with decimals would be super messy with a 1-5 decimal approach.

I agree that the 0-100 system is cleaner, but it also limits precision compared to using a decimal value. The decimal values could be cleaned up in the UI in various ways. I'm okay with the 0-100 system. A decimal 1-5 star system is just a suggestion to alleviate WithoutPants concerns.

skier233:
> > Well its already all implemented as is. I personally think using a 1-100 system under the hood is by far the best internal implementation and I've heard many people agree. Decimals are messy and unclear. For example, doing a 1-100 system with decimals would be super messy with a 1-5 decimal approach.
> 
> I agree that the 0-100 system is cleaner, but it also limits precision compared to using a decimal value. The decimal values could be cleaned up in the UI in various ways. I'm okay with the 0-100 system. A decimal 1-5 star system is just a suggestion to alleviate WithoutPants concerns.

I agree there is technically less precision but it would only be ever off by less 1 out of 100 which seems like a small enough increment to never really matter.

internal/manager/config/config.go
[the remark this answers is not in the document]
skier233: It looks like over 3x people prefer 5 star with half star increments over any other option so this should probably stay as is.

graphql/schema/types/config.graphql
```diff
@@ -299,6 +313,9 @@ type ConfigInterfaceResult {
   """Wall playback type"""
   wallPlayback: String
 
+  """Rating System Type"""
+  ratingSystem: String
```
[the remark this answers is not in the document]
skier233: Made changes but my local branch is now broken trying to migrate to one of the new changes in develop. More details here: https://discord.com/channels/559159668438728723/559169859351740418/1025419641142652948

Teda1:
I have no issue with a 0-10 decimal system to store the ratings. My issue was with a 10-star system to present the data if multiple UI options were too much to maintain. I agree a 0-10 decimal system for the data would be preferable.

skier233:
> I have no issue with a 0-10 decimal system to store the ratings. My issue was with a 10-star system to present the data if multiple UI options were too much to maintain. I agree a 0-10 decimal system for the data would be preferable.

I agree that the 0-10 star system is a bit big on the UI albeit not terrible. I don't see a problem keeping it because there is no difference in code for a 5 star system vs a 10 star system. They use the exact same code under the hood. It's just an extra enum value mapping to 10 instead of 5 so there isn't anything more to maintain so I figure its better to have the option for anyone that might want it. One use-case is people that want more precision than 5 half stars but still want to use a star system. Then it could make sense.

skier233:
> There's an inconsistency with the default rating system shown in the settings dropdown and the actual default rating system used when you go to edit a scene's ratings.
> 
> The UI shows five star system, but when I go to edit a scene's ratings it's using the half star UI. image
> 
> If I switch the rating system setting to something else and then back to 5 star the rating system UI uses the correct whole star UI, so it looks like it's just an inconsistency with the initial dropdown value. I tested using a fresh install.

Fixed this.

graphql/schema/types/config.graphql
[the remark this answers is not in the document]
skier233: Updated (using a fresh db to get around the migration failure)

graphql/schema/types/config.graphql
```diff
@@ -270,6 +281,9 @@ input ConfigInterfaceInput {
   
   """Set to true to disable creating new objects via the dropdown menus"""
   disableDropdownCreate: ConfigDisableDropdownCreateInput
+
+  """Rating System Type"""
+  ratingSystem: String
```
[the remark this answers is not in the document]
skier233: Updated

skier233:
Giving up on this PR.
The changes needed to add a duplicate record for a new rating as well as keeping the old record for backwards compatibility and mapping them to each other and doing that for every different type of ratable content seems insurmountable for me while working a full time job. Looks like it will need changes in over 100 places not to mention the testing effort to ensure that system works. If someone else wants to pick up the PR they can or else I'll just let it die and use my branch locally.

WithoutPants:
> The banner should show the values corresponding to the selected rating system. The legacy system shows 1-5 because it's a 1-5 star, but if someone changes to 1-10 they would expect to see 1-10. There's no reason to expose users to the underlying 1-100 values.

Agree, this needs to be changed.

> Regarding the banners, these should use the same number system as filters. Filters will need to use 0-100 because they go directly to the backend which is why this should probably stay as it is.

That the rating banners match the rating system is a pretty integral and reasonable requirement. We will get pushback from users if their rating banners go from 1-5 to 20/40/60/80/100.

WithoutPants:
I've opened a pull request against your branch: https://github.com/skier233/stash/pull/2

It should give you a start on the backend implementation. Happy to offer further guidance as needed.

HijackHornet:
Hello
Sorry i havent checked for a while, but it seems like the rating system is in the latest develop but not the option to turn it off / switching it to other counting method ?!

skier233:
> Hello Sorry i havent checked for a while, but it seems like the rating system is in the latest develop but not the option to turn it off / switching it to other counting method ?!

No. It hasn't merged

HijackHornet:
Well it must be partially merged then because i do have parts of it

skier233:
> Well it must be partially merged then because i do have parts of it

No. None of it is in develop.

HijackHornet:
image

skier233:
> image

That's the old system.

skier233:
Note to self: There's a bug with performer rating's not getting set correctly from the ui on the performer page

skier233:
> Note to self: There's a bug with performer rating's not getting set correctly from the ui on the performer page

fixed

pkg/sqlite/gallery.go
```diff
@@ -51,7 +52,15 @@ func (r *galleryRow) fromGallery(o models.Gallery) {
...
 	r.Details = zero.StringFrom(o.Details)
-	r.Rating = intFromPtr(o.Rating)
+
+	// prefer rating100 over rating
+	if o.Rating100 != nil {
+		r.Rating = intFromPtr(o.Rating100)
+	} else if o.Rating != nil {
+		rating100 := models.Rating5To100(*o.Rating)
+		r.Rating = null.IntFrom(int64(rating100))
+	}
```
WithoutPants, line 62: This can be reverted. `Rating` should already be 1-100.

pkg/sqlite/gallery.go
```diff
@@ -75,7 +84,7 @@ func (r *galleryQueryRow) resolve() *models.Gallery {
 		URL:           r.URL.String,
 		Date:          r.Date.DatePtr(),
 		Details:       r.Details.String,
-		Rating:        nullIntPtr(r.Rating),
+		Rating100:     nullIntPtr(r.Rating),
```
WithoutPants, line 87: This should be reverted to use `Rating`.

pkg/sqlite/gallery.go
```diff
@@ -90,6 +99,11 @@ func (r *galleryQueryRow) resolve() *models.Gallery {
 		ret.Path = r.FolderPath.String
 	}
 
+	if r.Rating.Valid {
+		rating5 := models.Rating100To5(int(r.Rating.Int64))
+		ret.Rating = &rating5
+	}
+
```
WithoutPants, line 106: This can be removed.

pkg/sqlite/gallery.go
```diff
@@ -102,7 +116,18 @@ func (r *galleryRowRecord) fromPartial(o models.GalleryPartial) {
...
+	// prefer rating100 over rating	
+	if o.Rating100.Set {
+		r.setNullInt("rating", o.Rating100)
+	} else if o.Rating.Set {
+		v := o.Rating
+		if v.Value != 0 {
+			v.Value = models.Rating5To100(v.Value)
+		}
+		r.setNullInt("rating", v)
+	}
```
WithoutPants, line 129: This should be able to be reverted. `Rating` should already be expressed as 1-100.

pkg/sqlite/image.go
```diff
@@ -40,7 +42,15 @@ type imageRow struct {
...
 	r.Title = zero.StringFrom(i.Title)
-	r.Rating = intFromPtr(i.Rating)
+
+	// prefer rating100 over rating
+	if i.Rating100 != nil {
+		r.Rating = intFromPtr(i.Rating100)
+	} else if i.Rating != nil {
+		rating100 := models.Rating5To100(*i.Rating)
+		r.Rating = null.IntFrom(int64(rating100))
+	}
```
WithoutPants, line 52: This can be reverted. `Rating` should already be 1-100.

pkg/sqlite/image.go
```diff
@@ -60,7 +70,7 @@ func (r *imageQueryRow) resolve() *models.Image {
 	ret := &models.Image{
 		ID:        r.ID,
 		Title:     r.Title.String,
-		Rating:    nullIntPtr(r.Rating),
+		Rating100:    nullIntPtr(r.Rating),
```
WithoutPants, line 73: This can be reverted as well.

pkg/sqlite/image.go
```diff
@@ -76,6 +86,11 @@ func (r *imageQueryRow) resolve() *models.Image {
 		ret.Path = filepath.Join(r.PrimaryFileFolderPath.String, r.PrimaryFileBasename.String)
 	}
 
+	if r.Rating.Valid {
+		rating5 := models.Rating100To5(int(r.Rating.Int64))
+		ret.Rating = &rating5
+	}
```
WithoutPants, line 92: This isn't necessary.

pkg/sqlite/image.go
```diff
@@ -85,7 +100,18 @@ type imageRowRecord struct {
...
+	if i.Rating100.Set {
+		r.setNullInt("rating", i.Rating100)
+	} else if i.Rating.Set {
+		v := i.Rating
+		if v.Value != 0 {
+			v.Value = models.Rating5To100(v.Value)
+		}
+		r.setNullInt("rating", v)
+	}
+
```
WithoutPants, line 114: This can be reverted. `Rating` is already 1-100.

pkg/sqlite/scene.go
```diff
@@ -73,7 +74,15 @@ func (r *sceneRow) fromScene(o models.Scene) {
...
-	r.Rating = intFromPtr(o.Rating)
+
+	// prefer rating100 over rating
+	if o.Rating100 != nil {
+		r.Rating = intFromPtr(o.Rating100)
+	} else if o.Rating != nil {
+		rating100 := models.Rating5To100(*o.Rating)
+		r.Rating = null.IntFrom(int64(rating100))
+	}
+
```
WithoutPants, line 85: This can be reverted. `Rating` is already 1-100.

pkg/sqlite/scene.go
```diff
@@ -97,7 +106,7 @@ func (r *sceneQueryRow) resolve() *models.Scene {
 		Details:   r.Details.String,
 		URL:       r.URL.String,
 		Date:      r.Date.DatePtr(),
-		Rating:    nullIntPtr(r.Rating),
+		Rating100: nullIntPtr(r.Rating),
```
WithoutPants, line 109: This can be reverted.

pkg/sqlite/scene.go
```diff
@@ -114,6 +123,11 @@ func (r *sceneQueryRow) resolve() *models.Scene {
 		ret.Path = filepath.Join(r.PrimaryFileFolderPath.String, r.PrimaryFileBasename.String)
 	}
 
+	if r.Rating.Valid {
+		rating5 := models.Rating100To5(int(r.Rating.Int64))
+		ret.Rating = &rating5
+	}
+
```
WithoutPants, line 130: This is unnecessary.

pkg/sqlite/scene.go
```diff
@@ -126,7 +140,18 @@ func (r *sceneRowRecord) fromPartial(o models.ScenePartial) {
...
+	if o.Rating100.Set {
+		r.setNullInt("rating", o.Rating100)
+	} else if o.Rating.Set {
+		v := o.Rating
+		if v.Value != 0 {
+			v.Value = models.Rating5To100(v.Value)
+		}
+		r.setNullInt("rating", v)
+	}
+
```
WithoutPants, line 154: This can be reverted.

pkg/sqlite/setup_test.go
```diff
@@ -967,11 +967,18 @@ func makeScene(i int) *models.Scene {
 		}
 	}
 
+	rating100 := getRating(i)
+	rating5 := rating100
+	if rating5.Valid {
+		rating5.Int64 = int64(models.Rating100To5(int(rating5.Int64)))
+	}
+
```
WithoutPants, line 975: This isn't needed.

pkg/sqlite/setup_test.go
```diff
@@ -967,11 +967,18 @@ func makeScene(i int) *models.Scene {
...
 	return &models.Scene{
 		Title:        title,
 		Details:      details,
 		URL:          getSceneEmptyString(i, urlField),
-		Rating:       getIntPtr(getRating(i)),
+		Rating100:    getIntPtr(rating100),
+		Rating:       getIntPtr(rating5),
```
WithoutPants, line 981: This can be reverted.

pkg/sqlite/setup_test.go
```diff
@@ -1046,9 +1053,16 @@ func makeImage(i int) *models.Image {
 	pids := indexesToIDs(performerIDs, imagePerformers[i])
 	tids := indexesToIDs(tagIDs, imageTags[i])
 
+	rating100 := getRating(i)
+	rating5 := rating100
+	if rating5.Valid {
+		rating5.Int64 = int64(models.Rating100To5(int(rating5.Int64)))
+	}
+
```
WithoutPants, line 1061: This isn't necessary.

pkg/sqlite/setup_test.go
```diff
@@ -1046,9 +1053,16 @@ func makeImage(i int) *models.Image {
...
 	return &models.Image{
 		Title:        title,
-		Rating:       getIntPtr(getRating(i)),
+		Rating100:    getIntPtr(rating100),
+		Rating:       getIntPtr(rating5),
```
WithoutPants, line 1065: This can be reverted.

pkg/sqlite/setup_test.go
```diff
@@ -1129,10 +1143,17 @@ func makeGallery(i int, includeScenes bool) *models.Gallery {
 	pids := indexesToIDs(performerIDs, galleryPerformers[i])
 	tids := indexesToIDs(tagIDs, galleryTags[i])
 
+	rating100 := getRating(i)
+	rating5 := rating100
+	if rating5.Valid {
+		rating5.Int64 = int64(models.Rating100To5(int(rating5.Int64)))
+	}
+
```
WithoutPants, line 1151: This isn't necessary.

pkg/sqlite/setup_test.go
```diff
@@ -1129,10 +1143,17 @@ func makeGallery(i int, includeScenes bool) *models.Gallery {
...
 	ret := &models.Gallery{
 		Title:        getGalleryStringValue(i, titleField),
 		URL:          getGalleryNullStringValue(i, urlField).String,
-		Rating:       getIntPtr(getRating(i)),
+		Rating100:    getIntPtr(rating100),
+		Rating:       getIntPtr(rating5),
```
WithoutPants, line 1156: This can be reverted.

WithoutPants (changes requested):
The rating conversion implementation is missing from Gallery, Performer, and Studio create methods. I think it's probably worth moving the conversion code into a separate function in `api` and using that in the create and update code where possible, rather than having repeated code everywhere.

skier233:
> The rating conversion implementation is missing from Gallery, Performer, and Studio create methods. I think it's probably worth moving the conversion code into a separate function in `api` and using that in the create and update code where possible, rather than having repeated code everywhere.

I thought you said it was only needed for the classes I have now because the rest were doing it in sqllite?

internal/api/resolver_mutation_gallery.go
```diff
@@ -188,6 +188,7 @@ func (r *mutationResolver) galleryUpdate(ctx context.Context, input models.Galle
 	updatedGallery.URL = translator.optionalString(input.URL, "url")
 	updatedGallery.Date = translator.optionalDate(input.Date, "date")
 	updatedGallery.Rating = translator.optionalInt(input.Rating, "rating")
+	updatedGallery.Rating100 = translator.optionalInt(input.Rating100, "rating100")
```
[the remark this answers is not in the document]
skier233: added

internal/api/resolver_mutation_gallery.go
```diff
@@ -263,7 +264,7 @@ func (r *mutationResolver) BulkGalleryUpdate(ctx context.Context, input BulkGall
 	updatedGallery.URL = translator.optionalString(input.URL, "url")
 	updatedGallery.Date = translator.optionalDate(input.Date, "date")
 	updatedGallery.Rating = translator.optionalInt(input.Rating, "rating")
-
+	updatedGallery.Rating100 = translator.optionalInt(input.Rating100, "rating100")
```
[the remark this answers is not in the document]
skier233: added

internal/api/resolver_mutation_image.go
```diff
@@ -104,6 +104,7 @@ func (r *mutationResolver) imageUpdate(ctx context.Context, input ImageUpdateInp
 	updatedImage := models.NewImagePartial()
 	updatedImage.Title = translator.optionalString(input.Title, "title")
 	updatedImage.Rating = translator.optionalInt(input.Rating, "rating")
+	updatedImage.Rating100 = translator.optionalInt(input.Rating100, "rating100")
```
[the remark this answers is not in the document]
skier233: added

internal/api/resolver_mutation_image.go
```diff
@@ -190,6 +191,7 @@ func (r *mutationResolver) BulkImageUpdate(ctx context.Context, input BulkImageU
 
 	updatedImage.Title = translator.optionalString(input.Title, "title")
 	updatedImage.Rating = translator.optionalInt(input.Rating, "rating")
+	updatedImage.Rating100 = translator.optionalInt(input.Rating100, "rating100")
```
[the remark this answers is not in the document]
skier233: added

internal/api/resolver_mutation_scene.go
```diff
@@ -116,6 +116,7 @@ func (r *mutationResolver) sceneUpdate(ctx context.Context, input models.SceneUp
 	updatedScene.URL = translator.optionalString(input.URL, "url")
 	updatedScene.Date = translator.optionalDate(input.Date, "date")
 	updatedScene.Rating = translator.optionalInt(input.Rating, "rating")
+	updatedScene.Rating100 = translator.optionalInt(input.Rating100, "rating100")
```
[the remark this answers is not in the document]
skier233: added

pkg/models/model_scene.go
```diff
@@ -198,12 +201,13 @@ func (s ScenePartial) UpdateInput(id int) SceneUpdateInput {
...
-		URL:          s.URL.Ptr(),
-		Date:         dateStr,
-		Rating:       s.Rating.Ptr(),
+		ID:      strconv.Itoa(id),
+		Title:   s.Title.Ptr(),
+		Details: s.Details.Ptr(),
+		URL:     s.URL.Ptr(),
+		Date:    dateStr,
+		Rating:  s.Rating.Ptr(),
+		// TODO Rating100
```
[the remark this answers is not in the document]
skier233: not sure whether you're saying this needs a rating100 here or not.

WithoutPants:
> > The rating conversion implementation is missing from Gallery, Performer, and Studio create methods. I think it's probably worth moving the conversion code into a separate function in `api` and using that in the create and update code where possible, rather than having repeated code everywhere.
> 
> I thought you said it was only needed for the classes I have now because the rest were doing it in sqllite?

Covered in Discord, but repeating here for documentation. Unfortunately I haven't been very clear and consistent in my thinking around this:

> The backend should operate completely in 1-100 range. The API layer should be responsible for converting between the legacy 1-5 values and the new canonical 1-100 values. So the sqlite layer should accept only the latter values.

ui/v2.5/src/index.scss
```diff
@@ -406,32 +406,96 @@ div.dropdown-menu {
...
 
 .rating-2 {
-  background: $sepia1;
+  background: #9e8974;
 }
 
 .rating-1 {
-  background: $dark-gray5;
+  background: #9b8c7d;
+}
```
WithoutPants, line 487: Can we change this so that the new rating systems use classes `rating100-x` and leave the 5-star system to use `rating-x`, and leave the `rating-x` styles unchanged from the existing styles? This way, people on the 5-star system will have the same styling, and they won't need to change their custom css unless they change systems.

WithoutPants (changes requested):
There's one final thing I want to raise.

I'd like to comment out the ten-star and ten-point options for the initial merge. It will be trivial to add them in when ready, but it's a lot harder to remove them once they are in. The reason I want to do this, is that I'll be reworking the filter UI very soon, and I'd prefer to work with 5-star controls initially. The thing I noticed when testing this is that the ten-star controls are very large and unwieldy, and I'd like to think about how to condense it before adding it in - and I don't want this UI sticking point to prevent/delay this PR any further. The presence of multiple significantly different rating systems will add extra required effort to the filter refactor work.

ui/v2.5/src/index.scss
skier233: updated

skier233:
> There's one final thing I want to raise.
> 
> I'd like to comment out the ten-star and ten-point options for the initial merge. It will be trivial to add them in when ready, but it's a lot harder to remove them once they are in. The reason I want to do this, is that I'll be reworking the filter UI very soon, and I'd prefer to work with 5-star controls initially. The thing I noticed when testing this is that the ten-star controls are very large and unwieldy, and I'd like to think about how to condense it before adding it in - and I don't want this UI sticking point to prevent/delay this PR any further. The presence of multiple significantly different rating systems will add extra required effort to the filter refactor work.

commented 10 star system

ui/v2.5/src/index.scss
```diff
@@ -406,6 +406,90 @@ div.dropdown-menu {
   margin: 0 10px;
 }
 
+.ratinghundred-20 {
+  background: #f00;
+}
```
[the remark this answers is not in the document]
skier233: This can't be changed to that. it has to be in a form of [a-zA-Z]+-[a-zA-Z0-9]*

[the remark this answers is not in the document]
WithoutPants: The pattern has to be `"^(\.*[A-Z]*[a-z]+)+(-[a-z0-9]+)*$"` which `rating-100-x` complies with. I've made the change myself.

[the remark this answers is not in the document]
skier233: did you test it? When I tried it it broke everything…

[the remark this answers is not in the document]
WithoutPants: Yes, I've verified that it works correctly.

JanJastrow:
Hi, I think this needs a bit of styling when set to decimal.
On a performers page it looks weird having an `input` element with 100% width.
Screenshot 2022-11-16 at 22 31 15

Issue #1028, opened 2020-12-30
[RFC] Stash built for a particular type of conventual studio-produced porn, not great at managing porn outside that


secondfolder:
## Summary
Stash is currently build primarily for porn produced by a studio. As such it has patchy support for any porn that falls outside of this. I would love to see what peoples thinking about this is and ideally for support to improve over time or at least to make that this implicit focus more transparent in Stash's marketing.

##  Detailed description
For the purposes of this discussion I'm going to focus mostly on video although this definitely applies to images too, maybe to a lesser extend. I use the word "video" but Stash doesn't really have the notion of a generic video (at least in terms of the interface presented to the user). Instead stash has "Scenes" which can make up "Movies" which are concepts that are applicable for a particular type of porn. By catering specifically for porn that fits within this paradigm you can make more targeted assumptions about what sort of metadata is applicable. "Scenes" have a scene number from the "Movie" that they are a part of. "Movies" are made by "Studios". Etc etc. 

This has a bunch of advantages. With more specific metadata which you can assume to be broadly applicable for all media that is stored you can present a richer UI to the user. For example since most porn that fits in this particular idea of a "Movie" has a cover image (previously necessary for displaying on the front of a DVD case) you can make the cover image for a "Movie" highly prominent because the majority of this type of porn will have an image made for this which is representative of the movie. This more specific set of Metadata is also generally objectively agreed upon. That makes projects like Stash Box somewhat simpler (at least relatively speaking). 

But the problem is that not all porn fits this paradigm. Arguably quite a lot of porn in fact, probably increasing amounts with the internet driving the proliferation of more amateur/non-studio porn distributed though things like OnlyFans or Hentai Foundry. Rule-34 images, seconds-long clips by released on twitter by indie porn stars, non-professional porn made by individuals and uploaded to the internet, compilation videos made by randoms made by editing a whole bunch of different videos together. Any porn that doesn't fit into the studio model of porn catered to by Stash breaks the implicit assumptions made by Stash. Sometime this difference is fairly minor and can be relatively and it is relatively straightforward to shoehorn into. An example of this might be hentai/cartoon porn image drawn by a particular artist. The artist could be added as the "Studio" that made the image and any characters represented within might be added as "Performers". This is pretty clumsy but at least sort of works.

[...]

PR #2837, opened 2022-08-18
Custom localization

HijackHornet:
# Scope
Allow users to make and use custom locales json. The same way we have custom css, we now allow users to rename/translate part of the ui to their heart's content. 
Demo : https://cdn.discordapp.com/attachments/559169859351740418/1009741099973226547/firefox_mvCNuuOINH.mp4
## Closes/Fixes Issues
Helps toward https://github.com/stashapp/stash/issues/1028

## Other testing QA Notes
Tested on firefox. It need to shift+refresh to apply sometimes because the language is cached.


HijackHornet:
For people that are willing to use this and other PRs before merge, i've made an experimental merged build here : https://github.com/HijackHornet/stashStar/releases/tag/FileRefactorAndPRS
Use with caution.

HijackHornet:
Thanks 👍

Issue #2824, opened 2022-08-13
[Bug Report] Slow scanning with huge ammounts of videos


suppaberg:
With huge amounts of videos, scanning local video content gets slower and slower.

General info:
Windows 10
Stash v0.16.1
stash-go.sqlite size: ~110MB
DB contents: ~25k scenes, ~200k images
stash folder location: ssd attached via smb

According to log, slowdown occurs in the "Creating new item..." stage for movies (36s in below example):

    time="2022-08-13 09:32:47" level=info msg="Calculating oshash for t:\\dl\\1\\something1.mp4 ..."
    time="2022-08-13 09:32:47" level=info msg="t:\\dl\\1\\something1.png doesn't exist. Creating new item..."
    time="2022-08-13 09:33:23" level=info msg="[generator] generating phash sprite for t:\\dl\\1\\something1.mp4"
    time="2022-08-13 09:33:27" level=info msg="[generator] generating sprite image for t:\\dl\\1\\something1.mp4"
    time="2022-08-13 09:33:46" level=info msg="[generator] generating sprite vtt for t:\\dl\\1\\something1.mp4"
    time="2022-08-13 09:33:46" level=info msg="Calculating oshash for t:\\dl\\1\\something2.mp4 ..."

Interestingly for images it works instantly.

    time="2022-08-13 09:32:07" level=info msg="Calculating checksum for t:\\dl\\1\\something1.png..."
    time="2022-08-13 09:32:07" level=info msg="t:\\dl\\1\\something1.png doesn't exist. Creating new item..."
    time="2022-08-13 09:32:07" level=info msg="Associating image t:\\dl\\1\\something1.png with folder gallery"

Just to be safe it is not my system, i created another Stash instance. No such problems there.

Unrelated to above bug, just a thought:
perhaps it would be better to have nested folders in generated\screenshots and generated\vtt so both folders do not have such huge amounts of files in single folder (~50k+ files right now). Listing such folders for whatever reason gets very slow.

scruffynerf:
Curious if the problem continues with the file refactor...

The generated directories getting huge could be solved by using a subdirectories solution usually that's done by using 1-2 of the first or last digits as a reference, like putting Axxxxx into an A subdirectories, Bxxxxx into a B subdirectory, etc.

suppaberg:
Yes, it seems that is already in use for generated\thumbnails (nesting folders two levels deep of 00 through ff) so could be reused here as well.

bnkai:
[...]

PR #2839, opened 2022-08-18
scene: generate thumbnail of screenshot

gitgiggety:
Currently the thumbnail is generated based on the video, but this is slow, especially when the video is on network storage. So instead only generate the screenshot based on the video, and resize that image for the thumbnail.

Refs: #2824

### DIFF
--- a/pkg/scene/scan.go
+++ b/pkg/scene/scan.go
@@ -358,17 +358,31 @@ func (scanner *Scanner) makeScreenshots(path string, probeResult *ffmpeg.VideoFi
 		logger.Infof("Regenerating images for %s", path)
 	}
 
-	if !thumbExists {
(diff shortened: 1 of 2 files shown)


gitgiggety:
Based on my testing this, kinda as expected, reduces the time to generate the screenshot(s) (in other words: the time `makeScreenshots` takes) in half. Results vary from 0.6 seconds before to 0.3s after to 1.7s before and 0.9s after.

Didn't check the video files I did tests on, let alone check the generation time per video, but I do believe some files are only a couple hunderds of MB (so 300-400MB) while others are in the 2GB. So that would also explain while some screenshots take a lot longer to generate.

WithoutPants:
This is not currently relevant. Scan no longer (currently) generates thumbnail files.

PR #2846, opened 2022-08-22
Fix unclosed div tag in login.html

v-helmholtz:


HijackHornet:
Seems to be a good edit :)

PR #2872, opened 2022-09-03
Add AirPlay and Chromecast support

### LINKED ISSUE #1978: [Bug Report] AirPlay/Casting option not present
**Describe the bug**

Somehow any option for casting either by AirPlay or Chromecast seems to be missing from the JWPlayer panel. Even when castin through control center, no video is sent to the receiver. My stash instance is setup with a valid public SSL certificate (ZeroSSL).

I was digging throught the code and atleast the necessary property seems to be set during the setup of the player object: [https://github.com/stashapp/stash/blob/60c4f7e99216ff068d0207fe7a29d54c9cc63856/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx#L311](https://github.com/stashapp/stash/blob/60c4f7e99216ff068d0207fe7a29d54c9cc63856/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx#L311)

**To Reproduce**
1. Go to any scene video
2. Start playing
3. No option to cast possible

**Expected behavior**
Option to cast should be present on the right side of the video controls like in this screenshot:
Bildschirmfoto vom 2021-11-09 10-36-17

**Screenshots**
Bildschirmfoto vom 2021-11-09 10-39-04
No option here

**Stash Version: 0.10.0 compiled for Windows

Desktop (please complete the following information):

OS: Window Server 2019
Browser: Chrome and Safari
Version: latest

Smartphone (please complete the following information):
Device: iPhone 7 Plus, iPad Air 3
OS: 14.3
Browser: Stock Safari

Teda1:
Fixes #1978

This pull request adds AirPlay and Chromecast support to the videojs player.


kermieisinthehouse (approved):

Teda1:
I was able to reproduce this. It appears the issue is related to login credentials and not SSL. However, this is not an issue with the plugin used here since the native iOS video player with built-in airplay functionality is also affected. A separate issue can be opened to investigate the cause further.

Before this issue is merged, I do want to add CSS for these plugins directly to Stash.

Teda1:
I was planning on porting the CSS and SVGs from the Chromecast and AirPlay plugins into our repo since Skier23 mentioned he had some issues building this branch with those CSS imports. But after looking at the code again, and given that this branch built just fine here and on both my and Kermie's dev environments, I'm no longer sure doing this makes sense.

JoeSmithStarkers:
I've done some testing on this PR back-ported to v0.16.1 (thus could be missing something)

However my tests showed that out of the box airplay casting works fine. However google casting didn't work. I believe this is due to https://www.gstatic.com/cv/js/sender/v1/cast_sender.js?loadCastFramework=1 not being loaded anywhere, thus the google casting api is missing (eg window.chrome.cast is undefined), this prevents the casting button from appearing.

I hacked in cast_sender.js in to the index.html (also adjusted the content-security) and the button appears.

Teda1:
I used to have an android device lying around, which I used for testing and development. It's been a while since I've seen it, though. I'll try and find it this weekend, so I can use it to test solutions. If it doesn't show up, you may have to post your fix, and other android users could help confirm that it works.

Teda1:
I was able to locate the android phone, but it's apparently old enough not to have built-in Chromecast support. Either way, I pushed an update to load cast_sender.js. I just need someone with a Chromecast-enabled android phone to confirm this fix works.

JanJastrow:
Hi, thank you for this feature - really looking forward to it.
But I think it should not be enabled by default if it involves a http query to Google on *every* page load for *every* user (yes, browser-cache is a thing).

Instead it should be optional to activate like DLNA (at least Chromecast support) and be opt-in - if there is no way to have a local hosted version of the ```cast_sender.js``` javascript.

Teda1:
I have looked into other ways to load this within the scene player page. Since I can’t test this locally, I want to confirm my understanding of the first solution works.

JoeSmithStarkers:
Hey, i just had a quick look at the change but i believe that won't work because of the content-security-policy set by the backend.

Look at internal/api/server.go where cspDirectives is set.
change: script-src 'self' https://cdn.jsdelivr.net 'unsafe-inline' 'unsafe-eval';
to: script-src 'self' https://cdn.jsdelivr.net https://www.gstatic.com 'unsafe-inline' 'unsafe-eval';

I will test this change once it's been added.

@JanJastrow agreed it not ideal that cast_sender.js needs to be loaded on every page. There is a way of dynamically loading it which could be added to the sceneplayer init code before the video player is loaded, but lets just test the easy way first.

Teda1:
@JoeSmithStarkers sorry, I'm late to this. It's been a busy week. I just pushed the change. Let me know how it goes.

JoeSmithStarkers:
@Teda1 I can confirm that those changes make chromecasting support work for me.

Sorry for taking so long to reply, kinda fell off the face of the earth.

Teda1:
No worries. I'm glad it worked. Later this week I can push the change to only load `cast_sender.js` on the scene player page.

Teda1:
I've pushed the update to dynamically load `cast_sender.js` when on the ScenePlayer page. I've confirmed that the script gets loaded on the page. I just need someone to confirm that the Chromecast option still appears and works on an android device.

internal/api/server.go
```diff
@@ -405,7 +405,7 @@ func SecurityHeadersMiddleware(next http.Handler) http.Handler {
 		}
 		connectableOrigins += "; "
 
-		cspDirectives := "default-src data: 'self' 'unsafe-inline';" + connectableOrigins + "img-src data: *; script-src 'self' https://cdn.jsdelivr.net 'unsafe-inline' 'unsafe-eval'; style-src 'self' https://cdn.jsdelivr.net 'unsafe-inline'; style-src-elem 'self' https://cdn.jsdelivr.net 'unsafe-inline'; media-src 'self' blob:; child-src 'none'; object-src 'none'; form-action 'self'"
+		cspDirectives := "default-src data: 'self' 'unsafe-inline';" + connectableOrigins + "img-src data: *; script-src 'self' https://cdn.jsdelivr.net https://www.gstatic.com 'unsafe-inline' 'unsafe-eval'; style-src 'self' https://cdn.jsdelivr.net 'unsafe-inline'; style-src-elem 'self' https://cdn.jsdelivr.net 'unsafe-inline'; media-src 'self' blob:; child-src 'none'; object-src 'none'; form-action 'self'"
```
kermieisinthehouse, line 408: This should be added above to connectableOrigins

kermieisinthehouse:
I can't test this because `make docker-build` fails with 
```
[2/4] Fetching packages...
error Couldn't find the binary git
```

If git is actually needed by the new yarn dependencies (not sure why it would be?) then can you add ` git` to the end of `docker/build/x86_64/Dockerfile` on line 5?

Teda1:
I'm not sure why it would be necessary either, but I've gone ahead and added `git` to the line. Let me know if it helps.

WithoutPants:
> I can't test this because `make docker-build` fails with
> 
> ```
> [2/4] Fetching packages...
> error Couldn't find the binary git
> ```
> 
> If git is actually needed by the new yarn dependencies (not sure why it would be?) then can you add ` git` to the end of `docker/build/x86_64/Dockerfile` on line 5?

Looks like it's due to this line in `yarn.lock`: https://github.com/stashapp/stash/blob/b62b5ed6b4f14596ed801bafa747d5829a92e1ed/ui/v2.5/yarn.lock#L1228

This is a dependency of `@silvermine/videojs-chromecast@^1.3.4`

WithoutPants:
> Hi, thank you for this feature - really looking forward to it. But I think it should not be enabled by default if it involves a http query to Google on _every_ page load for _every_ user (yes, browser-cache is a thing).
> 
> Instead it should be optional to activate like DLNA (at least Chromecast support) and be opt-in - if there is no way to have a local hosted version of the `cast_sender.js` javascript.

Agree, I don't like the UI hitting an external google URL for the scene player for all users. I think this should be a togglable setting.

internal/api/server.go
WithoutPants: Moving to `connectableOrigins` caused the script to not work. `connectableOrigins` is added to `default-src` only, which is only used if the more specific clause is not found. I think the original way was probably the correct one.

Teda1:
In its current state, Chromecast support in this pull request should be on par with the airplay. However, the big caveat with Chromecast is that the `__onGCastApiAvailable` flag in cast_sender.js, which loads the remaining necessary resources, is only triggered if the sender is hosted with HTTPS. Since Chromecast also suffers from the HTTPS issues we've seen with airplay, proper Chromecast support is blocked until that issue is resolved. I've noticed that the `__onGCastApiAvailable` flag does trigger when the URL is localhost, but a Chromecast receiver does not know where to find the video at that URL without some tweaking. Since that is not a typical use case, I didn't see the value in pushing the tweak I wrote to provide the proper IP. I attempted to set up an environment that could enable me to debug the Chromecast connection when sending via HTTPS to get more insight into the cause. However, I have yet to get the environment to work as intended. I may give it another go later.

JoeSmithStarkers:
Sorry, I've finally got around to testing this on my setup. Chromecast and airplay work well with this feature applied to v0.19.1 custom version that i am running.

My setup includes a reverse proxy, and letsencrypt cert. So i get around all the https issues.

john-miller-831985:
It looks like this branch has gone stale.  Personally I would really like this feature.  Is this something that would be merged if the branch is updated?

Teda1:
Sure, I can go ahead and update this branch so it can be merged as is. The HTTPS issue can be resolved later.

Teda1:
This branch is no longer stale if there were any plans to merge it as is.

john-miller-831985:
I had to update the CUDA dockerfile to get this to build, and even with valid SSL it did not seem to work.
```
--- a/docker/build/x86_64/Dockerfile-CUDA
+++ b/docker/build/x86_64/Dockerfile-CUDA
@@ -2,7 +2,7 @@
 
 # Build Frontend
 FROM node:alpine as frontend
-RUN apk add --no-cache make
+RUN apk add --no-cache make git
 ## cache node_modules separately
 COPY ./ui/v2.5/package.json ./ui/v2.5/yarn.lock /stash/ui/v2.5/
 WORKDIR /stash
```

Teda1:
Yeah, I mentioned issues with SSL in my earlier post. Both Chromecast and Airplay, even the out-of-the-box airplay support iPhone users have without these changes, don't function properly when using SSL. Airplay users can get around this though, since SSL isn't required for Airplay to work. Chromecast, however does. JoeSmithStarkers seems to have found a way to work around the SSL issues to get Chromecast working which users might consider until the SSL issues get worked out.

john-miller-831985:
Makes sense.  I misunderstood the error you were discussing above, and thought the issue only occurred on self signed ssl certs, but overall I see real promise in the code on this branch.  I just hope it gets to a point where it can be merged.

PR #2885, opened 2022-09-05
[Feature] [Help Needed] Image scrappers

HijackHornet:
# Scope
Allow image to use scrappers. Useful for implementing AI tagging, or getting date/url/performer from a social media post.

# Help needed
So far i've only been able to do so much on this, and i'm actually stuck on a few compilation errors. I just cant find what to do, so i'm seeking some external help here.

If you would to compile right now, make generate would yield :
```
D:\Projets\stashStar\internal\api\resolver_mutation_image.go:28:67: undeclared name: ImageUpdateInput
D:\Projets\stashStar\internal\api\resolver_mutation_image.go:46:71: undeclared name: ImageUpdateInput
D:\Projets\stashStar\internal\api\resolver_mutation_image.go:88:67: undeclared name: ImageUpdateInput
D:\Projets\stashStar\internal\api\resolver.go:57:9: cannot use &(queryResolver literal) (value of type *queryResolver) as QueryResolver value in return statement: *queryResolver does not implement QueryResolver (missing method AllGalleries)
D:\Projets\stashStar\internal\api\resolver.go:63:9: cannot use &(imageResolver literal) (value of type *imageResolver) as ImageResolver value in return statement: *imageResolver does not implement ImageResolver (missing method Date)
D:\Projets\stashStar\internal\api\server.go:99:65: cannot use resolver (variable of type *Resolver) as ResolverRoot value in struct literal: *Resolver does not implement ResolverRoot (missing method ImageUpdateInput)
```


HijackHornet:
Updated to latest file refactor

pkg/models/image.go
```diff
@@ -39,6 +39,21 @@ type ImageFilterType struct {
...
+	ClientMutationID *string  `json:"clientMutationId"`
+	ID               string   `json:"id"`
+	Title            *string  `json:"title"`
+	URL              *string  `json:"url"`
+	Date             *string  `json:"date"`
+	Rating           *int     `json:"rating"`
+	Organized        *bool    `json:"organized"`
+	StudioID         *string  `json:"studio_id"`
+	TagIds           []string `json:"tag_ids"`
+	PerformerIds     []string `json:"performer_ids"`
```
WithoutPants, line 56: You've defined `ImageUpdateInput` here, which means that references to `api.ImageUpdateInput` need to be updated to `models.ImageUpdateInput`.

This is why you're seeing the following errors:
```
D:\Projets\stashStar\internal\api\resolver_mutation_image.go:28:67: undeclared name: ImageUpdateInput
D:\Projets\stashStar\internal\api\resolver_mutation_image.go:46:71: undeclared name: ImageUpdateInput
D:\Projets\stashStar\internal\api\resolver_mutation_image.go:88:67: undeclared name: ImageUpdateInput
```

Also, you are missing the `GalleryIds` field. 

You can comment out this struct, rerun `make generate` and look for the `ImageUpdateInput` struct in `api/generated_models.go` for what it expects.

This explains this error:

```
D:\Projets\stashStar\internal\api\server.go:99:65: cannot use resolver (variable of type *Resolver) as ResolverRoot value in struct literal: *Resolver does not implement ResolverRoot (missing method ImageUpdateInput)
```

graphql/schema/schema.graphql
```diff
@@ -148,6 +155,8 @@ type Query {
   allStudios: [Studio!]!
   allMovies: [Movie!]!
   allTags: [Tag!]!
+  allGalleries: [Gallery!]!
+  allImages: [Image!]!
```
WithoutPants, line 159: I don't think these interfaces are a good idea, and I'm not sure why they are added.

In any case, you need to add implementations for these resolves to `queryResolvers`. See `AllTags` in `resolver_query_find_tag.go`. The typical workflow is to just implement them with a `panic("not implemented")` until you're ready to implement them properly.

This explains this error message:

```
D:\Projets\stashStar\internal\api\resolver.go:57:9: cannot use &(queryResolver literal) (value of type *queryResolver) as QueryResolver value in return statement: *queryResolver does not implement QueryResolver (missing method AllGalleries)
```
There should be one for `AllImages` as well. Not sure why there isn't.

workspace.code-workspace
```diff
@@ -0,0 +1,8 @@
+{
+	"folders": [
+		{
+			"path": "."
+		}
+	],
+	"settings": {}
+}
```
WithoutPants, line 8: This should not be included.

pkg/models/model_image.go
```diff
@@ -15,10 +15,12 @@ type Image struct {
 	ID int `json:"id"`
 
 	Title     string `json:"title"`
+	URL       string `json:"url"`
 	Rating    *int   `json:"rating"`
 	Organized bool   `json:"organized"`
 	OCounter  int    `json:"o_counter"`
 	StudioID  *int   `json:"studio_id"`
+	Date      *Date  `json:"date"`
```
WithoutPants, line 23: You need to add a `Date()` resolver method to `api.imageResolver` for this field. See `Date()` in `api/resolver_model_scene.go`.

This explains the following error:
```
D:\Projets\stashStar\internal\api\resolver.go:63:9: cannot use &(imageResolver literal) (value of type *imageResolver) as ImageResolver value in return statement: *imageResolver does not implement ImageResolver (missing method Date)
```

WithoutPants (changes requested):
I would suggest first splitting this PR up. You're adding new fields to the image object, which needs its own PR.

workspace.code-workspace
HijackHornet: Yes my bad. It keeps getting through the gitignore

graphql/schema/schema.graphql
HijackHornet: > I would suggest first splitting this PR up. You're adding new fields to the image object, which needs its own PR.

This is because I want to make a scrap by URL directly. If I do a second pr for this, I will then need a third one to meet the two end.
Same goes for the date. I want scrappers to be able to grab said date

WithoutPants: Right, but adding these two fields is a significant PR alone. It's easier for me to review and test two separate PRs, one for adding the fields, and another for scraping, than it is to review one giant PR that does everything.

The way I would do it would be to create a branch/PR for adding the fields first. When that is complete, submit that PR, and if you want to start work on scraping, create a new branch off the PR branch to do so. You can rebase off `develop` when the first PR is merged.

HijackHornet:
If i get time i'll split this into different PR.
I might just restart from zero adding in your advices.

HijackHornet:
> I would suggest first splitting this PR up. You're adding new fields to the image object, which needs its own PR.

Part 1 : https://github.com/stashapp/stash/pull/3015
Part 2 : TBD

I close this draft since i replaced it by two other PRs

Issue #2857, opened 2022-08-26
[Bug Report] Touch issue in lightbox


HijackHornet:
**Describe the bug**
Outside of the fact that you cant "swipe" between images, there is a huge number of issue with touch screens.

- If you zoom with two finger but remove a finger slightly before the other, it will center the image on the location of your second finger
- The "reset zoom on image changes" do not seems to work after zooming on mobile
- The fact that clicking the "semi transparent" area arround the image close it (even in fullscreen mode) so if you try and switch between images a bit too fast you will constantly close it. 
- Swipping will simply move the image on the side
- Going from portrait to landscape break the all thing

**Desktop (please complete the following information):**
 - OS: Android
 - Browser Firefox

**Smartphone (please complete the following information):**
 - Device: Motorola G PRO
 - OS: Android 11
 - Browser Firefox

**Additional context**
I would simply suggest a big rework of the lightbox alltogether. 
The way the lightbox works is really weird. It load full scale images and then rescale them. So on slow connection and devices, images keep being super zoomed for a couple of second each time you display a new one.
There are tones of carousel/gallery library that are touch friendly already available. It would be worth checking them out instead of trying to make one ourself

HijackHornet:
I am willing to put the work into switching to a new carrousel if needed

Issue #2538, opened 2022-04-28
[Bug Report] Mobile controls on gallery lightbox


sleetx:
On mobile devices, the gallery lightbox does not respond to standard mobile controls. It still uses desktop image controls despite being in mobile format.

To reproduce:
Open gallery lightbox on a mobile device. When attemping to swipe left or right, the image follows the user's finger moving around the screen - often moving most of the image off-screen depending how far you swipe. In addition, double tapping the image does not control zoom, it simply detects which half of the image you tapped and decides to go back/forward one image (like the desktop controls).

Expected functionality should match most mobile UI interfaces:
- Left/right swipe moves to the previous or next image
- Double tap to zoom in; double tap again to fit-to-screen

echo6ix:
Just a note, touch gestures should not be limited to mobile viewports. Swipe navigation is also very useful on laptops with touchpads or large viewport touch devices.

echo6ix:
From [Discord](https://discord.com/channels/559159668438728723/644934273459290145/976945053798920332)
> Not sure what lightbox is being used in stash, but there are a few open source alternatives that support mobile swipe and double-tap zoom controls https://github.com/andreknieriem/simplelightbox

HijackHornet:
I'll start working on an overall

HijackHornet:
Indeed with https://github.com/dromru/react-photoswipe-gallery it should be pretty trivial to do.
After experimenting a bit, the main change would be the current lightbox component is used inside of ImageList and GalleryList.
The new lightbox would wrap those instead.

callmenoodles:
I created a [plugin](https://github.com/callmenoodles/stash-better-lightbox) which aims to add these features later but for now I moved both arrows to the bottom left making it a lot easier to navigate between images on mobile.

PR #2890, opened 2022-09-06
[Feature] [Help Needed] Lightbox V2

HijackHornet:
# Scope
The goal is to replace the current lightbox by a well tested, broadly used js library.

# Issues
This has been discussed in the following issues : https://github.com/stashapp/stash/issues/2538, https://github.com/stashapp/stash/issues/2857

# Solution
I've done a few experimentations with the [Photoswipe](https://photoswipe.com/) library. Its a JS lightbox library that offer not only all the features that we have but also a very good touch support. This include : single touch, double tap action, swiping, but is also compatible with infinite scrolling. Note that there is an unofficial React library of photoswipe, but iit seems to me that it will be more of a challenge to use than the real thing. Simply because some methods looks harder to reach through this extra layer of abstraction.

# Proof of concept (POC)
I've spent a couple hours only replacing the current lightbox by a very simple one of the library
[Animated Demo](https://drive.google.com/uc?id=1B5BP1ZTStLGvmzphBq9v9OU4jZNDM3u3)
[POC Code](https://github.com/HijackHornet/stashStar/tree/LightBoxV2)

# What now ?
The POC is nice but its not as reusable as the current lightbox implementation. As a matter of fact, it requires to modify each UI component that want to use it. but hope is not lost. It is possible to have things in a single component and simply use hooks to use it.
The above code is what i've done so far on this subject. It will require more work to finish but its outside of my capabilities i'm affraid. 

I have removed unnecessary lightbox code (but not everything because we want to keep the option overlay  etc).
I also add the "fullscreen" button but cant find how to set the icon.


HijackHornet:
Closing because I don't think this will ever be done and is till don't have the necessary knowledge to make it myself. I'll keep using my POC which is enough for me.

PR #2891, opened 2022-09-06
[Fix] add space between "Stash" and "IDs" on studio panels

halorrr:
After https://github.com/stashapp/stash/pull/2810 merged I realized that I missed in my first pass that there should be a space her in both the part I added and where I copied the logic from to match how it is displayed on the scene page and be in a format that a user would actually expect to read it in.

How it looks fixed:
image

image

### DIFF
--- a/ui/v2.5/src/components/Studios/StudioDetails/StudioDetailsPanel.tsx
+++ b/ui/v2.5/src/components/Studios/StudioDetails/StudioDetailsPanel.tsx
@@ -59,7 +59,7 @@ export const StudioDetailsPanel: React.FC<IStudioDetailsPanel> = ({
     return (
       <>
         <dt>
-          <FormattedMessage id="StashIDs" />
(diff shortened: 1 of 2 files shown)


DogmaDragon:
Might be just me but I prefer it without space.

Also, Performers also don't have a space, but Scenes do.

AdultSun:
I agree, "StashID" makes more sense to me. Feels more like a proper noun that way and always looked intentional to me.

halorrr:
Moved to draft, I'll switch this PR to change the scene page to make things uniform as per https://discord.com/channels/559159668438728723/798641040029777980/1016868829470670910

kermieisinthehouse (approved):

WithoutPants:
Closing as stale.

discourse-stashapp:

https://discourse.stashapp.cc/t/poll-do-you-prefer-term-stashid-or-stash-id/4626/1

PR #2899, opened 2022-09-08
Fix URL encoding

### LINKED ISSUE #2263: [Feature] URL-encode JSON
I use NGINX reverse proxy along with Authelia to password protect endpoints. I'm aware that there is a password option in stash, but I prefer SSO.

The presence of curly braces in many stash URLs causes authelia to enter an endless redirect loop. It's something that could potentially be fixed on their end, but that just means the next reverse proxy that someone else uses will break too.

Square brackets and curly braces in URLs should be URL-encoded

As mentioned above, this could be fixed by authelia, but in this case stash is the one using non-standard practices.

**Additional context**
As per [RFC 1738](https://www.rfc-editor.org/rfc/rfc1738#page-3), certain special characters, including some used heavily in JSON, are considered unsafe to use in URLs and should always be encoded.

### LINKED ISSUE #2864: [Bug Report] Path filter with special characters does not persist on page reload
**Describe the bug**
When using a Path filter with a "#" character on the Scenes page, the filter does not persist on a page refresh from the browser.

**To Reproduce**
1. Go to "Scenes"
2. Add a filter for "Path" which includes "#"
3. Refresh the page
4. Filter is removed

**Expected behavior**
Path filter should remain when page is refreshed. If the path does not contain "#" then the filter persists correctly. This appears to be a regression from 0.16

**Screenshots**

0.16.1

 - OS: Windows 8/64
 - Browser: MS Edge
 - Version: Version 104.0.1293.70 (Official build) (64-bit)

**Additional context**

DingDongSoLong4:
This changes the way that filter query parameters are URL-encoded to be more compliant.

As per RFC 3986, 'reserved characters' only need to be encoded if they can have other special meaning, so the only characters that need to be encoded in a query string after using `encodeURI()` are `?#&;=+`.

It appears that the reason why stricter URL-encoding was not done was to avoid "long and ugly" URLs, so here the JSON characters {} are replaced with the explicitly unreserved () characters which helps somewhat.

Fixes #2864
Fixes #2263


DingDongSoLong4:
Added an optimization for nullable criterion - the filter value is irrelevant when the filter modifier is `IsNull` or `NotNull`, so the value does not need to be included in a query.

A-Random-Mf:
The bug seems to be partially fixed. Multi word filters now return results but not the correct ones. Scenes that should be excluded by the filter are not

PR #2943, opened 2022-09-24
Fix JSON.parse console error

DingDongSoLong4:
This just suppresses a JSON.parse console error when the custom locales from #2837 are disabled.

I've added the fix in the backend because the Content-Type is application/json so the response should always be valid JSON.


WithoutPants (approved):
Good catch, thanks.

PR #2949, opened 2022-09-26
Add default thumbnails for scenes and images

WithoutPants:
Adds default thumbnails for scenes and images for when the thumbnail fails to load:

image
image

Doing this for galleries is a more difficult process so is not included in this PR.


bnkai (approved):

internal/api/routes_image.go
```diff
@@ -55,19 +58,21 @@ func (rs imageRoutes) Thumbnail(w http.ResponseWriter, r *http.Request) {
 	if exists {
 		http.ServeFile(w, r, filepath)
 	} else {
+		const useDefault = true
+
 		// don't return anything if there is no file
```
bnkai, line 63: update the comment as well ? it now returns a default/placeholder image?

PR #2951, opened 2022-09-26
Handle file rescan

### LINKED ISSUE #2939: develop build creates empty galleries when encountering duplicates
found this on v0.16.1-60-gb74428cb

there's still a smallish bug in how the files-refactor creates galleries for duplicates. i had it before on updating performer names, but now on a fully clean scan, it's creating empty galleries for the 2nd gallery it encounters containing duplicates. 

I can reproduce this on a new empty db by doing this:

- create a gallery with some images, name it `galleryA`. 
- copy that folder and name the copy `galleryB`. 
- let stash scan these folders.

my expectation would be that it creates 2 galleries, both `galleryA` and `galleryB`, and has entries for the images in the images/files tables in the db. i'd expect both galleries to be fully "functional", so browse to `galleryA` it should show all images, same for `galleryB`. if you then click an image in one of these galleries, i'd expect the image to be linked to both galleries.

result on new scan: (so none of the content was known by stash before): 

- `galleryA1` gets created correctly, has the images and shows the cover.jpg i've made available in the folder. 
- it then comes to `galleryB`, creates a new folder entry (`time="2022-09-21 21:53:56" level=info msg="\\\\nas\\folder\\performers\\performer\\galleryB.XXX.IMAGESET doesn't exist. Creating new folder entry..."`). 
- it then creates the folder based gallery (`time="2022-09-22 08:35:45" level=info msg="Creating folder-based gallery for \\\\nas\\folder\\performers\\performer\\galleryB.XXX.IMAGESET"`) 
- and starts processing images. on the images it calculates fingerprints, 
- and connects them to the duplicate on `galleryA` as expected (`time="2022-09-22 08:35:45" level=info msg="Adding \\\\nas\\folder\\performers\\performer\\galleryB.XXX.IMAGESET\\image_50_0000.jpg to image \\\\nas\\folder\\performers\\performer\\galleryA.XXX.IMAGESET\\image_29_0000.jpg"`)
- it mentions in the log for each image, that it's connecting them to the new folder based gallery for galleryB (`time="2022-09-22 08:35:45" level=info msg="Adding \\\\nas\\folder\\performers\\performer\\galleryB.XXX.IMAGESET\\image_50_0000.jpg to folder-based gallery \\\\nas\\folder\\performers\\performer\\galleryB.XXX.IMAGESET"`)
- but when browsing to that gallery, it shows up as empty.

It happens the same way when updating existing content (for example when performer in path changes)

WithoutPants:
- fixes issue where scene/image/gallery would not be created if it there was none associated with a file that had been subsequently moved
- scene/image/gallery post-update hooks are now called when the underlying file was updated
- scene generated files are now migrated when the underlying file hash has changed
- creates galleries when necessary instead of upfront when detecting a zip file. This fixes the issue where galleries would be created for zip files containing no valid image files
- clean now deletes empty galleries again
- fixes issue in clean where zip file folders would fail to be cleaned when its path is no longer valid because the zip file was moved
- fixes issue where duplicate images would not be associated with galleries - fixes #2939

### DIFF
--- a/internal/manager/task_clean.go
+++ b/internal/manager/task_clean.go
@@ -78,6 +78,14 @@ func (j *cleanJob) cleanEmptyGalleries(ctx context.Context) {
 			found = len(emptyGalleries) > 0
 
 			for _, g := range emptyGalleries {
+				if g.Path == "" {
+					continue
+				}
+
+				if len(j.input.Paths) > 0 && !fsutil.IsPathInDirs(j.input.Paths, g.Path) {
+					continue
+				}
+
 				logger.Infof("Gallery has 0 images. Marking to clean: %s", g.DisplayName())
 				toClean = append(toClean, g.ID)
 			}
--- a/internal/manager/task_clean.go
+++ b/internal/manager/task_clean.go
@@ -16,6 +16,7 @@ import (
 	"github.com/stashapp/stash/pkg/models"
 	"github.com/stashapp/stash/pkg/plugin"
 	"github.com/stashapp/stash/pkg/scene"
+	"github.com/stashapp/stash/pkg/txn"
 )
 
 type cleaner interface {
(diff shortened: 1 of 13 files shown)


djbtwothousandtwenty:
- first finding: the cleaning of empty galleries is not restricted to the "clean selected" path, it also cleaned a bunch of empty galleries outside of the path i selected


- second: it cleaned 3 empty galleries it made when encountering duplicate files previously, but on re-scan, it does not create new galleries for these duplicates. not sure if the folder paths need to be changed for them to be picked up, but right now it appears it's happy with the fact that the folders are in my folders table in the db, even if there's no gallery for the folder where there should be. i think if i change the folder name, it probably will pick it up as then the entry in the folders table "does not exist" for that specific path, so i'll try that for one of the dupes that's missing

it does pick that, and identifies that the images have been "moved" from old foldername to new foldername,  but again does not create the additional gallery. i'm now removing the performer folder alltogether, running clean to remove all these files/galleries/scenes, and then let it discover the whole performer folder again like it's never seen it before

this works, so when stash has never seen the content before, and does not have any path info in the folders or files tables, it created everything as i'd expect. No empty galleries, and 6 galleries with the correct names, the images are correctly connected to both the original as well as the "duplicate" gallery, and they show up correctly

the logic for the existing situation where the folders have previously been scanned and added to the folders table, but with an empty gallery created, leads to a weird state where it knows about the folder, it knows about the images, but it does not create the expected gallery

so i think the logic probably needs to check when it encounters a folder with images that would trigger the "create folder based gallery", whether it has one for that folder or not. if it does have one, update where required, if it does not have one, create a gallery and add the images found to it as it does now

- third finding might not be a bug at all: i know i've got a total of 5144 images in that specific performer folder. everything's tagged correctly, but when going to the images view, it reports a total of 4499 images. so that query "excludes duplicates". if i go filter for all images with a file count higher than 1, i get 639 duplicate images. which brings the total up to almost the correct number (4499+639=5138) which is still 6 short. this is because some cover files are duplicated in more than 2 folders, one of them is in 5 galleries for example. so i know by checking manually, all images are in the db, but maybe it would be good to check the files table for this performer for image filecount, and ignore whether duplicates are in the total? that way the imagecount would match the filesystem count

in this case, the reason for the duplicates is that it's very old met-art content, they were a bit messy in 2003, so some galleries got re-issued with different names, on different dates and sometimes with exactly the same photo's, sometimes with just a small difference in photocount, and some times with new cover files, sometimes with old ones. i could go and delete these dupes of course, but for me as a collector: this is how they released it, so that's how i would like to keep it, as close to the original as possible

WithoutPants:
Investigating this: I added two folders: `20220928` and `20220928.2`. I added three images to both folders, and an extra three to the latter folder. Scanning appears to be behaving correctly with regard to the image/gallery association. Duplicate files have a single image with both files associated, and associated to both folder galleries.

It is possible to add/remove images to/from folder/zip-based galleries, so I will plug that hole. Otherwise, unless I manually modify the database, it shouldn't be normally possible to get into the situation you describe, as far as I can tell. For scenarios where this has occurred, the recourse would probably be to either delete the folder-based gallery and rescan, or touch the affected image files, which should trigger a rescan and re-association.

I don't want to add a check to ensure that all images are associated to their correct galleries because it will add unnecessary overhead to scanning, it shouldn't be normally possible to get into this situation, and it is possible to workaround it if someone does get into the situation.

PR #2953, opened 2022-09-26
Add backup directory path setting

7dJx1qP:
This PR adds a new system setting to set a directory for database backup files.

image

The default backup directory path is the config path so there's no change to current behavior if the setting isn't set or changed.

### DIFF
--- a/graphql/documents/data/config.graphql
+++ b/graphql/documents/data/config.graphql
@@ -5,6 +5,7 @@ fragment ConfigGeneralData on ConfigGeneralResult {
     excludeImage
   }
   databasePath
+  backupDirectoryPath
   generatedPath
   metadataPath
   scrapersPath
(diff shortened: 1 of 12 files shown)


WithoutPants (changes requested):

WithoutPants (approved):

JanJastrow:
Hi, I'm having issues with this:

The initial path was empty, so I set it to `/root/.stash/backups` and run backup.
Log:
`Backing up database into: /root/.stash/backups/root/.stash/stash-go.sqlite.35.20220930_095307`

It seems to merge the default path with my custom path?
Now I set it to `backups` (relative path then?) and run backup.

Log:
`Backing up database into: backups/root/.stash/stash-go.sqlite.35.20220930_095456`
So it added my path to the front of the default path

Next: I tried to clear the field:

Red Popup (and **not** in log):
`Error
mkdir : no such file or directory`

Tried backup again:

Red Popup (and **not** in log):
`Error
vacuum failed: unable to open database: backups/root/.stash/stash-go.sqlite.35.20220930_095708: no such file or directory`

So it does not clean my previous entry correctly (it shows nothing is set on the settings page)
Right now I can't run the backup at all.

PR #2970, opened 2022-10-02
TagDisplay on Wall mode + refacto

HijackHornet:
# Scope
Allow for tag/performer/organized icon display on hover in wall view.
Also refactored a bit to separate ImageWall into its own file.

Demo : 

## Other testing QA Notes
There is a couple issues left that i'm stuck on : 
- For mobile user, i want to show the bottom bar on single tap on the bottom of the image (focus). But it doesnt work so far (see code notes)
- I have a warning at runtime saying i need "keys" to my list. I dont understand this error.


ui/v2.5/src/components/Images/ImageWall.tsx
```diff
@@ -0,0 +1,191 @@
...
+          <div className="organized">
+            <Button className="minimal">
+              <Icon icon={faBox} />
+            </Button>
+          </div>
+        );
+      }
+    }
+
+    function renderEdit() {
```
HijackHornet, line 103: Adding a "+" icon/button here so that you can still see the bottom bar and access editing if no other element present in it

ui/v2.5/src/components/Images/ImageWall.tsx
```diff
@@ -0,0 +1,191 @@
...
+          <Button className="minimal">
+            <Icon icon={faPlus} />
+          </Button>
+        </div>
+      );
+    }
+    const ref = useRef(null);
+    const handleTouchEnd = (event) => {
+      if (ref.current === document.activeElement) {
+        alert("If i would to display this, it would be nice!");
```
HijackHornet, line 115: Here is where the trouble is.
I'm trying to use the ref.current that points to my element (i checked it does), and focus().
But it doesnt seems to happen. If i do
`ref.current.focus()` and then check `ref.current === document.activeElement` its always false. This means the element is never focused.

HijackHornet: Solved by @InfiniteTF

HijackHornet:
Rebased on v0.20.1 (in case this ever get merges)

HijackHornet:
Closing because this included some refacto, and now there has been new changes made to things that i had moved. So it breaks the PR and i dont plan on updating it

PR #2974, opened 2022-10-03
Make scenes page list view checkbox fill entire cell

7dJx1qP:
The current checkbox inputs on the scenes page listview are hard to click, so I wrapped them in a label that fills the table cell. Clicking any part of the table cell will now check the input.

The blue outlines show the clickable area that will check the inputs:

Before
image

After
image


WithoutPants:
Extending the checkbox to the edge of the cell looks pretty ugly to me.

My suggestion is to follow what's in this stackoverflow answer: https://stackoverflow.com/a/55157487

I'd also prefer doing this in css rather than using the bootstrap classes.

7dJx1qP:
Ok, I changed the bootstrap classes to css and added padding to the label.
image

PR #3005, opened 2022-10-13
Support filtering by StashID endpoint

### LINKED ISSUE #2999: [Feature] Expand StashID filter to filter by endpoint
With more endpoints becoming available users will likely want to filter items with StashIDs from a specific endpoint, currently this cannot be done.

Add the ability to filter by stashbox endpoint for items with StashIDs

manually tagging items that have StashIDs from a given endpoint, this would not be ideal going forward

stg-annon:
supports filtering by stashbox endpoint for items with StashIDs

note the excludes filter for performers may need more work

resolves #2999


stg-annon:
resolves #2999

WithoutPants:
I've rewritten this to apply endpoint filtering to the stash id criterion. The new criterion allows for equals, not equals, is null and not null, and uses the optional endpoint as a predicate. That is, if endpoint is provided for an is null modifier, then it will find objects having stash ids for that endpoint.

stg-annon:
> if endpoint is provided for an is null modifier, then it will find objects having stash ids for that endpoint.

the opposite filter does not appear to do what one might expect where a provided endpoint with a `not null` will not provide a list of objects with stash ids matching that endpoint

dsrtusr88:
Using this feature in `development`, i'm seeing entries that include stashids from two different endpoints (stashdb and tpdb).  the only way I can show only stashids with stashdb is the primary is to filter out the tpdb, but that by itself shows entries that have no stashids.  a simple workaround is to just sort by updated, but as implemented I don't think this gets to the request of the PR

MrX292:
> Using this feature in `development`, i'm seeing entries that include stashids from two different endpoints (stashdb and tpdb). the only way I can show only stashids with stashdb is the primary is to filter out the tpdb, but that by itself shows entries that have no stashids. a simple workaround is to just sort by updated, but as implemented I don't think this gets to the request of the PR

works for me with `Stash ID is not null https://metadataapi.net/graphql` what is your filter set to?

dsrtusr88:
> > Using this feature in `development`, i'm seeing entries that include stashids from two different endpoints (stashdb and tpdb). the only way I can show only stashids with stashdb is the primary is to filter out the tpdb, but that by itself shows entries that have no stashids. a simple workaround is to just sort by updated, but as implemented I don't think this gets to the request of the PR
> 
> works for me with `Stash ID is not null https://metadataapi.net/graphql` what is your filter set to?

that works, but it also shows files that have no stashid which is not desired

MrX292:
> 



> > > Using this feature in `development`, i'm seeing entries that include stashids from two different endpoints (stashdb and tpdb). the only way I can show only stashids with stashdb is the primary is to filter out the tpdb, but that by itself shows entries that have no stashids. a simple workaround is to just sort by updated, but as implemented I don't think this gets to the request of the PR
> > 
> > 
> > works for me with `Stash ID is not null https://metadataapi.net/graphql` what is your filter set to?
> 
> that works, but it also shows files that have no stashid which is not desired

are you using `Stash ID is null` or `Stash ID is not null`

> > > Using this feature in `development`, i'm seeing entries that include stashids from two different endpoints (stashdb and tpdb). the only way I can show only stashids with stashdb is the primary is to filter out the tpdb, but that by itself shows entries that have no stashids. a simple workaround is to just sort by updated, but as implemented I don't think this gets to the request of the PR
> > 
> > 
> > works for me with `Stash ID is not null https://metadataapi.net/graphql` what is your filter set to?
> 
> that works, but it also shows files that have no stashid which is not desired

are you using `Stash ID is null` or `Stash ID is not null`? 
if i use `Stash ID is null https://metadataapi.net/graphql` it shows all scenes without stashid and all that dont have the endpoint `https://metadataapi.net/graphql`
`Stash ID is not null https://metadataapi.net/graphql` shows me all scenes with the endpoint `https://metadataapi.net/graphql` with a stash id

dsrtusr88:
@MrX292 yep, but when I use that option, it shows me files that have both stashids in a lot of cases.  there's no way to have a pure filter to only show scenes that 1: have a stashid 2: the only stashid is the filtered endpoint

MrX292:
@dsrtusr88 
that both works for me on a8f4c2c2
1.`Stash ID is not null` only show scenes with stashid
2. `Stash ID is not null https://metadataapi.net/graphql`  only shows scenes with stashid with that endpoint

PR #3006, opened 2022-10-13
Support file-less scenes. Add scene split, merge and reassign file

### LINKED ISSUE #2380: [Feature] Transfer Metadata to duplicates
Im hopefully looking for an feature that let you transfer already scraped metadata from a duplicate scene to the other/others duplicate scene/s:
Unbenannt

WithoutPants:
Adds `New` button to the scenes page, to allow creation of new scenes without a related file.

Adds `Reassign` and `Split` buttons to secondary file info details panel. Reassign reassigns the file to another scene. Split opens the new scene page, assigning the file to the newly created scene.

Adds `Merge` menu item to the scenes overflow dropdown menu. 

image

Merge opens a merge dialog allowing the selection of the source and destination scenes. The button in the middle moves the first source scene to the destination and the destination scene to the source list.

Clicking `Next` displays a scrape-like dialog allowing tailoring of the merged data:

image

The `Source` data is populated based on merging of the source scenes. For single value fields, the first populated field is used. For Tags, Performers etc, it combines all values into one.

Closes https://github.com/stashapp/stash/issues/2380


St4ndardDev1anT:
Although it may already be implied, support for merging scene studio, date, details, stash ID, and o-counter would be very useful, in addition to the scene data fields shown in the dialog mock-up.

To merge o-counters, the question would be whether to overwrite or add to the destination file's counter.

I would also suggest `Merge` be incorporated into the `Scene Duplicate Checker`, perhaps as a button beside or just below each matching group of scenes.

_(Long time lurker, first-time poster. Apologies if these kinds of comments should be made elsewhere.)_

WithoutPants:
> Although it may already be implied, support for merging scene studio, date, details, stash ID, and o-counter would be very useful, in addition to the scene data fields shown in the dialog mock-up.

All these except for o-counter are supported, just not shown in the screenshot above.

> To merge o-counters, the question would be whether to overwrite or add to the destination file's counter.

There's no handling for o-counter at all in the current implementation - an oversight. I think it should add them together.

> I would also suggest `Merge` be incorporated into the `Scene Duplicate Checker`, perhaps as a button beside or just below each matching group of scenes.

This is a good idea, and I had planned to do so. I will add it to the todo list.

> _(Long time lurker, first-time poster. Apologies if these kinds of comments should be made elsewhere.)_

No, this is very helpful and useful here. Thank you.

WithoutPants:
Added merge button to scene duplicate checker. By default it merges other duplicates into the scene whose button was clicked. You can also select specific source scenes to merge from using the select checkboxes.

image

WithoutPants:
Added o-counter to merged fields. It uses the sum of all scene o-counters as the source value:

image

Teda1:
I noticed that after I re-assigned a file to a file-less scene, the file-less scenes remained unplayable:
Screenshot 2022-10-30 103255

It looks like the split function doesn't have the option to carry over the o-counter as the merge functionality does. Perhaps this was intentional?

I also noticed that that is no way to carry over the cover image between merges or splits.

Teda1:
After merging 2 files into one scene, I noticed that the scene counter still displays 2 despite the fact that just 1 scene is displayed:
Screenshot 2022-10-30 105020

Carrying on from that bug the front page displayed that 1 scene twice which seems to cause the row to display oddly:
Screenshot 2022-10-30 105020

bnkai:
>I noticed that after I re-assigned a file to a file-less scene, the previously file-less scene remained unplayable:
Same thing happens if you merge a `normal` scene (source) to a `file-less`(destination) scene. With a refresh of the page it is fixed though.

>After merging 2 files into one scene, I noticed that the scene counter still thinks 2 scenes are displayed despite the fact that just 1 scene is displayed:
Confirmed here as well. The `/stats`  page has the correct updated count.

>I also noticed that that is no way to carry over the cover image between merges or splits.
Merges seem to carry over the cover for me. In a few cases changing the primary file is needed to display the cover.
For splits you get to select the image for the `created` scene. Maybe instead carry over the cover image by default?

WithoutPants:
- fixed latent bug in the query counts
- fixed issue where primary file would not be assigned when merging into an empty scene
- fixed UI caching issue when merging and reassigning between scenes
- UI now sets the image from the source scene when splitting
- merging now includes the _first_ scene cover image to allow choosing between existing and source cover image

Teda1:
Thanks for the updates. I no longer see the issues I saw earlier. I am noticing another problem that may have been present earlier. When I merge 2 scenes into one, I see that the scrubber sprites are from the secondary file, not the primary one. You can see what I mean in the screenshot below(notice that the spite from 36 seconds in does not match what is being displayed in the video). It seems like this isn't getting synchronized somewhere:
wrong thumbnail

Foucoubou:
Hello, 
If may I have a question: Does the file-less scene offer some visual disambiguation (eg. grayed cover) to distinguish that?
I see very good example in XBVR which is similiar app for VR content, where there is whole web scraped and the missing / not paired scenes to files are grayed out :)

Many thanks

WithoutPants:
> When I merge 2 scenes into one, I see that the scrubber sprites are from the secondary file, not the primary one. 

I wasn't able to reproduce this exact issue, but I did find that reassigning the primary was incorrectly renaming generated files, so it may be related to that. Can you retry with the latest version?

> Does the file-less scene offer some visual disambiguation (eg. grayed cover) to distinguish that? I see very good example in XBVR which is similiar app for VR content, where there is whole web scraped and the missing / not paired scenes to files are grayed out :)

This is a good idea. I've added a `fileless` class to the scene cards without files, and styled it accordingly (file-less on left):

image

Teda1:
I see that merging 2 scenes together now also combines the performers, tags, and stash ids instead of overriding one for the other. However, there doesn't seem to be a way to prevent a stash id from getting merged(the `x` button is not present.
Screenshot 2022-11-07 174219

It turns out that the apply button can't even be hit because of the following error:
Screenshot 2022-11-07 174305

I'm not sure this new merge behavior makes sense. In the example below I'm trying to merge `Real Model` into `Keisha Grey`. At the top it says left is the Destination and right is the Source however due to the new merge behavior all performers, tags, and stash ids are merged into the source section. Also for whatever reason, the cover from the source gets auto-selected which threw me off after my initial merge.
Screenshot 2022-11-07 175744

> When I merge 2 scenes into one, I see that the scrubber sprites are from the secondary file, not the primary one.
I did confirm that this issue is no longer present.

bnkai:
I have been using the build from this PR for some time now. Imho it is safe enough to merge so that we can have more feedback/use cases and then fine tune what is needed.

Issue #1433, opened 2021-05-25
[Feature] Add PHOTOGRAPHER metadata field for Images/Gallery/Scrapers


echo6ix:
Images and Galleries could use a `photographer` metadata string value field. And if this field is added consider also adding it as a variable that is accessible to scrappers.

PR #3015, opened 2022-10-18
[Feature] Images new fields : URL & Date

HijackHornet:
# Scope
As part of my work toward Image scraping, i first added URL and Date as fields for Images.
image
image
image
Also include the fix to 👍
image


Teda1:
Looking at the images you posted, the URL field should be displayed in the File Info tab using the `Downloaded From:` message to be consistent with URL fields elsewhere.

HijackHornet:
Arf i missed that part. Ok i'll make the changes

HijackHornet:
Turns out i cant. The way the props is made differently on images, that requires changing the props type aand i dont manage to do it. Would you lend me an hand please :3

Teda1:
I wouldn't be able to hop on your branch until the weekend at the soonest. With the changes, you've made so far though it seems like `props.image.url` should work from `ImageFileInfoPane.tsx`.

HijackHornet:
So i spent over an hour understanding the code but i managed to do it :)

DogmaDragon:
> So i spent over an hour understanding the code but i managed to do it :) image

For consistency could you move the `Downloaded From:` to the bottom?  That's how Scenes list it.

HijackHornet:
Done

echo6ix:
Way late to the party but another ~17 month old image metadata field request, [#1433 ](https://github.com/stashapp/stash/issues/1433), was to add a `photographer` metadata field for the photographer's name. As a metadata field it's not uncommon, Armour Angels, Show Beauty, and the ATKingdom network use the photographer metadata to name a few.

pkg/scene/export.go
```diff
@@ -40,6 +40,7 @@ func ToBasicJSON(ctx context.Context, reader CoverGetter, scene *models.Scene) (
 	newSceneJSON := jsonschema.Scene{
 		Title:     scene.Title,
 		URL:       scene.URL,
+		Date:      scene.Date.Time.String(),
```
WithoutPants, line 43: This is covered in the if statement below. This will panic if date is `nil`.

pkg/image/export.go
```diff
@@ -15,6 +15,8 @@ import (
 func ToBasicJSON(image *models.Image) *jsonschema.Image {
 	newImageJSON := jsonschema.Image{
 		Title:     image.Title,
+		URL:       image.URL,
+		Date:      image.Date.Time.String(),
```
WithoutPants, line 19: This will panic if date is `nil`. See other `Date` export examples.

pkg/gallery/export.go
```diff
@@ -15,6 +15,7 @@ func ToBasicJSON(gallery *models.Gallery) (*jsonschema.Gallery, error) {
 	newGalleryJSON := jsonschema.Gallery{
 		Title:     gallery.Title,
 		URL:       gallery.URL,
+		Date:      gallery.Date.Time.String(),
```
WithoutPants, line 18: This is already covered in the if statement below. This will panic if `Date` is `nil`.

ui/v2.5/src/models/list-filter/criteria/is-missing.ts
```diff
@@ -49,7 +49,7 @@ export const ImageIsMissingCriterionOption = new IsMissingCriterionOptionClass(
   "isMissing",
   "imageIsMissing",
   "is_missing",
-  ["title", "galleries", "studio", "performers", "tags"]
+  ["title", "galleries", "studio", "performers", "tags", "url", "date"]
```
WithoutPants, line 52: URL is covered already in the url criterion.

WithoutPants (changes requested):

pkg/image/export.go
HijackHornet: Done

pkg/gallery/export.go
HijackHornet: Done

ui/v2.5/src/models/list-filter/criteria/is-missing.ts
HijackHornet: Done

HijackHornet:
I've fixed the issues you raised, rebased onto master and changed the migration version for the DB. I think its ready for merging

WithoutPants (approved):

PR #3020, opened 2022-10-19
Scene player improvements

### LINKED ISSUE #2476: [Bug Report] Issues with video.js player and Hover on touch devices.
**Describe the bug**
The video.js player has several issues related to hover on touch devices. The hover effect persists when elements are no longer being touched and can really mess up the UI.

**To Reproduce**
Here is a video demonstrating the issue:

https://user-images.githubusercontent.com/72030708/161653790-56ea53b4-3aee-4da7-80ff-480fd3becbee.MP4

**Expected behavior**
I'm not sure yet how best to handle this. If we disable hover then it seems like we would lose the ability to preview timestamps. There may be a way to fix the hover issue while keeping the ability to preview timestamps.

### LINKED ISSUE #2609: [Bug Report] Transcoding options vanish and does not work on ipad
**Describe the bug**
The transcoding options are not working on ipad.

**To Reproduce**
View a video on ipad and change the transcoding. Than the waiting circle appears and after a while the video plays in the old resolution without transcoding. Further more the selected options vanishes from the transcoding menu.

**Expected behavior**
The movie should be transcoded like on desktop

**Screenshots**

The bug was introduced with v0.15.0. On v0.14.0 everything worked as expected.

 - OS: Linux
 - Browser firefox
 - actual version

 - Device: ipad 6
 - OS: iPadOS 15.4.1
 - Browser safari, chrome, firefox, duck duck go browser
 - actual versions

**Additional context**

### LINKED ISSUE #2688: Transcoding options are not available on older video formats.
**Describe the bug**
When I try to watch an older video, it does not give me transcoding options. I have tried this on my iphone with safari, chrome and edge. The video is in WMV, but this works when I try and newer mp4 video. No issues when I use edge on my pc. Is this a known bug with the video player?
**To Reproduce**
1. Go to 'Scenes'
2. Click on 'a scene that is WMV or another older format'
3. Scroll down to 'the gear icon'
4. See error: None of the transcoding options seem to show up.

**Expected behavior**
Expected behavior is to see the various transcoding options.

**Screenshots**

0.15

**Tablet (please complete the following information):**
 - Device: iPad Pro (3rd Generation)
 - OS: iPadOS 15.5
 - Browser: Safari, Chrome & Edge

### LINKED ISSUE #2771: Update video.js from 7.17.0 to 7.20.1
The existing video.js version in Stash is 7.17.0 (https://github.com/videojs/video.js/releases/tag/v7.17.0).

The latest video.js version is 7.20.1 (https://github.com/videojs/video.js/releases/tag/v7.20.1), released in June 2022.

Is there any chance at all that we could update to the latest video.js version? (Not sure how hard or trivial this is).

(I think this would also update [http-streaming](https://github.com/videojs/http-streaming) from [2.12.0](https://github.com/videojs/http-streaming/releases/tag/v2.12.0) to [2.14.2](https://github.com/videojs/http-streaming/releases/tag/v2.14.2). Although I think http-streaming is also pinned in the [yarn.lock](https://github.com/stashapp/stash/blob/develop/ui/v2.5/yarn.lock) file - so I assume that will need to be updated?)

### LINKED ISSUE #2793: [Bug Report] Fix Mobile Thumbnail Scrubbing
**Describe the bug**
Thumbnails don't scrub when touching and moving.

**Additional context**
I've been able to fix this by editing the library that supports this function, but the library looks abandoned. Maybe it would be best to simply just fork/copy the library into the project. 

https://github.com/GilgusMaximus/videojs-vtt-thumbnails-freetube/blob/12788f5923a097766587de03ed77595e56dcd3d9/src/plugin.js#L219
```js
this.progressBar.addEventListener('touchmove', this.registeredEvents.progressBarMouseMove);
```

https://github.com/GilgusMaximus/videojs-vtt-thumbnails-freetube/blob/12788f5923a097766587de03ed77595e56dcd3d9/src/plugin.js#L225
```js
this.progressBar.addEventListener('touchmove', this.registeredEvents.progressBarMouseMove);
```

### LINKED ISSUE #2831: [Feature] Better support for WMV/MKV
WMV and MKV need forever to load, sometimes don’t work at all. Fast forwarding/skipping in the video is nearly unusable.
This changed since the use of video.js, before it worked much better (at least on iOS)

Bring back JW Player for MKV/WMV or make it possible to start an external app (mobile) like VLC from the stream URL.

Perhaps if the user could modify the link, it could work. For iOS you should be able to call VLC directly (not tested): vlc-x-callback://x-callback-url/ACTION?url=...&PARAMETER=... 
You can also start intents like that on Android. Not a nice solution, since configuration is needed for it

Copying the stream URL to VLC, which works, but is not great regarding the usability. Converting all videos would be a solution as well, but there are to many videos

### LINKED ISSUE #2856: [Feature] Make spacebar always pause/play video while on video page
On a video page, the spacebar event appears to trigger the most-recently-clicked/selected element, when it should (optionally) be treated as a pause/play button.  For example, when a video is made fullscreen, the expand/contract (fullscreen) button is the last clicked item, so when the spacebar is pressed afterwards, this triggers the the expand/contract (windowed) button instead of pausing the video.  Compare this to the UI behavior of YouTube, Pornhub, and many, many other video sites, which always treat the spacebar as a pause/play button regardless of the last-clicked/selected element.

If some users prefer the current spacebar treatment, this could be made a toggle in settings, but having some way to always treat the spacebar as a pause/play trigger is the requested feature.

As a workaround, the video can be clicked in the main video screen area after going fullscreen and this makes pause/play the last-clicked element, but this is an extra step that is easily forgotten (and not needed in many of web video UIs), causing frustration.

**Additional context**
I looked for an option for this in the settings but didn't find anything.  I also searched in the open and closed Issues for this but found nothing there as well.  Please let me know if I missed it somehow or if there's any other way to accomplish this (without extra clicks on every video).  Thanks.

### LINKED ISSUE #2877: When paused, scene player starts playing when you move through the timeline
**Describe the bug**
When viewing a video and the video is currently paused, using the timeline to scrub over the video (or skip buttons) causes the video to start playing but should stay paused. If you are looking for a scene to mark or a frame to capture it is impossible as the video plays once you move to the correct position and release the mouse. 

**To Reproduce**
1. Open a scene and start playing clip from any position.
2. Click on pause to pause the clip.
3. Using timeline drag to another point in the clip.
4. clip starts playing from that point.

**Expected behavior**
Once paused it should stay paused after moving through the timeline. If however the video is playing then it should continue to play after moving through the timeline. This is standard mode of operation on all other video players.

**Screenshots**
N/A

v0.16.1

**Desktop **
 - Windows 11
 - Opera 90.0.4480.80  and MS Edge 105.0.1343.25

**Smartphone **
N/A

### LINKED ISSUE #2901: [Bug Report] Incorrect mime types for transcoded scene streams
**Describe the bug**
Transcoded streams have incorrect mime type `video/mp4` instead of for instance `video/webm`

**To Reproduce**
Run query:
```
{
  findScene(id:523){
    sceneStreams{url, mime_type},
  }
}
```
Response:
```
{
    "data": {
      "findScene": {
        "sceneStreams": [
          {
            "url": "http://192.168.66.6:9999/scene/523/stream",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.mkv",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.webm?resolution=FOUR_K",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.webm?resolution=FULL_HD",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.webm?resolution=STANDARD_HD",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.webm?resolution=STANDARD",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.webm?resolution=LOW",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.mp4?resolution=FOUR_K",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.mp4?resolution=FULL_HD",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.mp4?resolution=STANDARD_HD",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.mp4?resolution=STANDARD",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.mp4?resolution=LOW",
            "mime_type": "video/mp4"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.webm",
            "mime_type": "video/webm"
          },
          {
            "url": "http://192.168.66.6:9999/scene/523/stream.m3u8",
            "mime_type": "application/vnd.apple.mpegurl"
          }
        ]
      }
    }
  }
```
Base webm stream has correct `mime_type` but the transcoded ones, e.g. `http://192.168.66.6:9999/scene/523/stream.webm?resolution=STANDARD_HD` has a different mime type.

**Expected behavior**
`mime_type` should probably be the same for all webm streams. Also mkv should have `video/x-matroska`

0.16.1

### LINKED ISSUE #2934: [Bug Report] Default Player Resolution for HEVC scenes
**Describe the bug**
Currently, when playing HEVC videos, the default selected resolution is incorrect. For a 1080p scene for example, I've noticed "webm low (240p)" being selected by default. This only seems to occur on HEVC videos from what I've noticed. When you play a AVC video, it selects "Direct stream" as default, which isn't available for HEVC which is what I believe to be the cause.

I cannot confirm that this is the same on other codecs such as VP9 or VP8.

**To Reproduce**
1. Open a video that uses the HEVC codec
2. Play the video
3. The video will be played in a lower resolution instead of its full quality (if not, perhaps try a different video)

**Expected behaviour**
The video should always play at its full resolution and not default to a lower quality.

**Stash Version:** 3fa7b470 (built from develop branch)

 - OS: Windows 10 Version 21H1 (Build 19043.1889)
 - Browser: Vivaldi (Chromium browser, however confirmed to exist in Chrome as well)
 - Version: 5.4.2753.51 (Stable channel) (64-bit)

DingDongSoLong4:
This is quite big, but all the changes are quite integrated at this point so I'm sending it all together here.

~This replaces the transcoding implementation to only use HTTP streaming formats (HLS for H264 and DASH for VP9), based off of #2322. The videojs player doesn't support VP9 with DASH natively, so videojs-contrib-dash (using dash.js) is needed.~

I've also refactored the videojs plugins and added type information for them so that most of the `any` casts in `ScenePlayer` are no longer needed. The markers plugin is also now a class and stores the markers itself, instead of using subtitles and cues stored in the player.

The scene player component itself has been overhauled. Autoplay and scene markers now work more consistently, and the spacebar always pauses/plays the video. Most of the multiple source logic in the player component has also been moved into the source-selector plugin.

The scrubber has also been overhauled and is now more consistent and responsive in general. I also moved the scrubber auto-hiding into the component instead of CSS, so that the scrubber isn't rendered at all when it isn't being displayed on mobile for example. The scrubber also now uses videojs-vtt.js (which is already a dependency of videojs itself) to parse the sprite VTT files instead of the 'gnarly' previous implementation.

I've also copied the videojs-vtt-thumbnails-freetube library into the project as suggested in #2793 and rewritten it in TypeScript, with a few tweaks and a fix for thumbnail scrubbing on mobile.

I've also updated a few dependencies, ~and fixed a circular dependency warning that Vite gives when building the UI~.

2bd64e7 removes the now unneeded chapters_vtt endpoint - I guess it's possible that some external plugin or application might need it for some reason but I think it's fine to remove.

Fixes #2877
Closes #2856
Fixes #2793
Closes #2771
Should fix #2476


WithoutPants:
I'm seeing pretty significant jerkiness and slow performance when viewing wmv files using the default dash implementation. It's a definite step down in quality from the current implementation.

DingDongSoLong4:
I haven't noticed any jerkiness, but my PC is relatively powerful so the buffer builds up quite quickly. I copied all the video transcode quality-related ffmpeg options straight through from before, so the actual quality and encoding performance should theoretically be similar.

Do you still get bad performance when using the stream.mpd link in something that can play DASH like `mpv`? (VLC doesn't seem to be able to from my testing) I hadn't tested WMV files very thoroughly, but playing one now I can see that there does seem to be some scrubbing issues (only for WMV files) at least when using the site player - `mpv` plays them perfectly. I'll see if I can figure out what's causing that.

It might be a good idea to make the HLS stream the first in the list anyway, as that is more widely supported and stable. VP9 encoding is still a lot more intensive than the H264 used for the HLS stream, at least at the current encoding settings. An option to change transcode quality is something I'll probably add as well, perhaps even something for hardware encoding which would speed everything up drastically.

WithoutPants:
Both PCs (host and client) are pretty decent machines, so it shouldn't be a performance issue. I noticed that it's particularly bad at the start of a video. I'll do some more thorough testing later. 

I think that it's worth splitting up this PR, because there are a number of very good improvements unrelated to the HLS/dash stuff. If you don't get to it first, I'll spend some time rebasing and opening up some separate PRs.

DingDongSoLong4:
Thanks! I initially didn't split the HLS/DASH stuff out because the `live.ts` hackery wasn't playing nicely with the updated `source-selector.ts`, but that was quite early on so the version of `source-selector` that ended up here might actually work fine. I had reworked `live.ts` with types, less prototype hacking and into a class, but I just got rid of it after adding HLS/DASH, which in hindsight probably wasn't a good idea. I'll see if I can either find that version or just rewrite it.

WithoutPants:
I tried to split and rebase, but I'm running into issues with the big button plugin. The buttons don't appear to be firing when clicked. I'll probably do the circular vite fix manually, but otherwise you might need to do the splitting for me.

DingDongSoLong4:
Yeah the circular vite fix is just a one liner, can't say I understand why it gives that warning in the first place but that gets rid of it so all's good I guess.
I should be able to get the splitting working, I'll check it out later today.

DingDongSoLong4:
Alright I've split out all the transcoding stuff from here into a separate PR - let me know if you'd like me to split it up even further.

Teda1:
I just jumped on this branch to do some testing, and I'm also noticing some jerkiness with the playback compared to the previous version. I also noticed that play-progress bar disappears when transcoding to very low resolutions. 
Screenshot 2022-10-29 181633

I tested subtitle support while transcoding, and it doesn't seem as accurate(the timing is off). My server is also currently committing a lot of CPU power to preview generation so that might also be playing a role. However, playback from on v0.17 release is still more stable under the same condition. I'll circle back once the server resource is under less load.

DingDongSoLong4:
Is this on the version with the dash support or not? Both HLS and DASH split the transcode into 2 second segments for streaming which means that if the encoding can't keep up then playback will pause every 2 seconds - is that what you're seeing? The encoding settings are basically identical to before but the files have to be split and written to the filesystem before being streamed so that is probably adding a bit more overhead.

For the subtitle issue, what browser are you using? Everywhere I've tested them they seem to be working for me. I'm also unable to reproduce the progress bar disappearing issue.

Teda1:
Playback is more stable now that the server isn't under as much load. The progress bar still appears to be less responsive when transcoding. It disappears when selecting the `webm` option(shown 19 seconds in), and in some cases, it does not return when changing to other transcode resolutions(I was unable to capture this in the recording below). Also, skipping to different time intervals on this branch temporarily reset the progress bar position, which I don't see happening on the release build. 

https://user-images.githubusercontent.com/72030708/198896207-6f8fea6a-46a2-46c5-9dc6-096406f62041.mp4

Regarding https://github.com/stashapp/stash/issues/2476, Your change does address some of the hover issues. The big one being the preview image remaining visible. I still see that the hover glow on the progress bar sometimes persists. The big buttons are not very responsive on my phone with these changes as WP mentioned earlier. I have to mash the play, rewind, or fast-forward button for a while before the input is registered.

DingDongSoLong4:
Fixed the big buttons and hover issue. The progress bar is part of videojs itself so I've just force-removed the hover glow to fix it.
The disappearing progress bar issue is only happening on Chromium browsers, which is why I wasn't able to reproduce it originally. I can now vaguely remember that I was facing some kind of similar issue when I reworked `live.ts` the first time, so I should be able to find a solution.

DingDongSoLong4:
Alright I think I've fixed those scrubbing issues now, thanks for the testing and feedback.

Teda1:
I can confirm that the big button and progress bar issues are gone. Thanks for pushing the fixes.

WithoutPants:
When scrubbing a live-transcoded video, it's showing the cover image instead of a black screen while it loads. Is it possible to maintain this behaviour? New behaviour is a little jarring.

WithoutPants:
This is looking close to complete. It seems to have broken Handy integration when clicking the seek buttons. I'll need some time to work out what the problem is.

DingDongSoLong4:
I don't have a Handy so you'll need to test it, but I took a look at what player events get fired when scrubbing a transcoded stream vs a direct stream, and I've added a force trigger for the `timeupdate` and `pause` events which weren't being fired - hopefully that fixes the problem.

WithoutPants:
I'll try to do some debugging, but my testing is with a direct streamed file. Clicking on the timeline or scrubbing any other way works fine. It's only when clicking on the seek back 10 seconds/seek forward 10 seconds buttons that it gets out of sync.

WithoutPants:
I think I was able to fix this. The `seeking` event handler was removed, which I think ended up callling `onplay` as part of the call to `player.play()`. I re-added `seeking` back with the following code:
```
    function seeking(this: VideoJsPlayer) {
      if (!this.paused()) {
        if (scene?.interactive && interactiveReady.current) {
          interactiveClient.play(this.currentTime());
        }
      }
    }
```

And added `player.on("seeking", seeking");` and the equivalent `off` call.

WithoutPants (approved):

Issue #1901, opened 2021-10-26
[Bug Report] Cant build on mac. "make ui" fails with ....


gedw99:
**Describe the bug**

On both develop and master branch.

On mac:
- yarn version v1.22.17
- Node.js v17.0.1.
- go 1.17

**make ui**  failing with 

````

make generate
go generate -mod=vendor
cd ui/v2.5 && yarn run gqlgen
yarn run v1.22.17
warning ../../../../../../../../../../../package.json: No license field
$ gql-gen --config codegen.yml
  ✔ Parse configuration
  ✔ Generate outputs
✨  Done in 4.06s.
 make ui
make ui
export REACT_APP_DATE="2021-10-26 17:44:44" ; \
        export REACT_APP_GITHASH=1fffc051 ; \
        export REACT_APP_STASH_VERSION=v0.10.0-39-g1fffc051 ; \
        cd ui/v2.5 && yarn build
yarn run v1.22.17
warning ../../../../../../../../../../../package.json: No license field
$ craco build
node:internal/errors:464
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/tokenize' is not defined by "exports" in /Users/apple/workspace/go/src/github.com/gedw99/notes-product/video-editor/stashapp__stash/stash/ui/v2.5/node_modules/postcss-safe-parser/node_modules/postcss/package.json
    at new NodeError (node:internal/errors:371:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:416:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:669:3)
    at resolveExports (node:internal/modules/cjs/loader:482:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/Users/apple/workspace/go/src/github.com/gedw99/notes-product/video-editor/stashapp__stash/stash/ui/v2.5/node_modules/postcss-safe-parser/lib/safe-parser.js:1:17) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v17.0.1
error Command failed with exit code 1.
[...]

PR #3030, opened 2022-10-21
add `pre-ui` to default make target and update documentation

aghoulcoder:
Without `pre-ui`, `make` fails. `pre-ui` is already used in the GitHub workflows but omitted from the Makefile and docs.

This is also directly related to the ancient issue #1901. But was that issue ever resolved? Or is this a regression?
I don't care enough to do an archaeological excavation through `git log`. The build process needs to be simplified more.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ export CGO_ENABLED = 1
 
 .PHONY: release pre-build
 
-release: generate ui build-release
+release: pre-ui generate ui build-release
 
 pre-build:
 ifndef BUILD_DATE
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -43,9 +43,10 @@ NOTE: The `make` command in Windows will be `mingw32-make` with MingW. For examp
 
 ## Building a release
 
-1. Run `make generate` to create generated files
-2. Run `make ui` to compile the frontend
-3. Run `make build` to build the executable for your current platform
+1. Run `make pre-ui` to install UI dependencies
+2. Run `make generate` to create generated files
+3. Run `make ui` to compile the frontend
+4. Run `make build` to build the executable for your current platform
 
 ## Cross compiling
 


stg-annon:
https://github.com/stashapp/stash/blob/develop/docs/DEVELOPMENT.md#commands
> make pre-ui - Installs the UI dependencies. Only needs to be run once before building the UI for the first time, or if the dependencies are updated

rarely will `pre-ui ` need to be run outside of these situations so it doesn't really make sense to incorporate it into every build GitHub workflows would need it as they spin up a new image every time one runs

aghoulcoder:
I'm new.
hi @stg-annon 👋

I want to contribute.
I followed the steps from `docs/DEVELOPMENT.md` to build the app.
The build failed:
```
$ make generate
go generate -mod=vendor ./cmd/stash
cd ui/v2.5 && yarn run gqlgen
yarn run v1.22.19
$ gql-gen --config codegen.yml
/bin/sh: line 1: gql-gen: command not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
make: *** [Makefile:159: generate-frontend] Error 127
```

This is a horrible experience for new contributors. Instead of complaining or asking basic questions, I tried to fix it. It's entirely possible my suggested fix isn't optimal. Can you please suggest an alternative?

aghoulcoder:
Also, this is the performance impact subsequent runs of `make pre-ui`:

```
$ make pre-ui
cd ui/v2.5 && yarn install --frozen-lockfile
yarn install v1.22.19
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.24s.
```
> 0.24s

Meanwhile, I probably spent an hour trying to figure out why I can't build the app. And clearly other people have also faced this issue.

It is not my intention to rant. I just want to emphasize that, in my view, there's a lot of room for improvement here.

scruffynerf:
Sorry, @stg-annon , but @aghoulcoder is correct, this is a perfect example of a landmine: blocks new users, and experienced ones just step around it.  Just fix the damn thing.

stg-annon:
> Done in 0.24s.

fair enough, I guess I'm just used to it at this point and every time I have need to run it it's been one of the longer tasks, I probably had a similar experience setting it up my first time as well

WithoutPants:
From memory, I think `pre-ui` was omitted from the "Building a release" steps because it implied that it needs to be run every time, when it only needs to be run when the UI dependencies have changed. I can see how that approach could confuse people, and the same argument could be made for the `generate` target as well.

PR #3041, opened 2022-10-24
Fix path filters

### LINKED ISSUE #3037: [Bug Report] Wrong search results when using a trailing slash
**Describe the bug**
Wrong search results when using a trailing slash, for example `/foo/ -/bar/` isn't including all `/foo/` scenes nor excluding `/bar/` correctly.

**To Reproduce**
1. Go to 'scene page'
2. Click on 'search input'

**Expected behavior**
Work as before

**Stash Version: (from Settings -> About):** v0.17.1

 - OS: docker
 - Browser: chrome
 - Version: latest

DingDongSoLong4:
Before files-refactor, the path filter simply treated a file's full path as a single string field. The path filter is now in most scenarios (mostly when `getPathSearchClause` is not used) taking the folder path and file basename separately, which is unintuitive and different to before.
This changes the behaviour back, by concatenating the folder path and file basename in SQL for both the path regex filters and the path component of the generic search query filter. It also changes `galleryPathCriterionHandler` to use `getPathSearchClauseMany` for consistency with `pathCriterionHandler`.

48543d7 also replaces `getPathSearchClause` to concatenate the folder path and file basename as well. This change breaks tests so I've separated it out. I would argue that the old behaviour of having a path 'is' filter matching the entire path only is more intuitive, but the way it is right now with differing behaviour depending on slashes, etc. might be more useful so I will defer judgement.

Fixes #3037

### DIFF
--- a/pkg/sqlite/file.go
+++ b/pkg/sqlite/file.go
@@ -786,7 +786,8 @@ func (qb *FileStore) Query(ctx context.Context, options models.FileQueryOptions)
 	distinctIDs(&query, fileTable)
 
 	if q := findFilter.Q; q != nil && *q != "" {
-		searchColumns := []string{"folders.path", "files.basename"}
+		filepathColumn := "folders.path || '" + string(filepath.Separator) + "' || files.basename"
+		searchColumns := []string{filepathColumn}
 		query.parseQueryString(searchColumns, *q)
 	}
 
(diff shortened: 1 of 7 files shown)


WithoutPants:
I'm fairly sure that I implemented it the way I did because of a misguided attempt to ensure that indexes were used. I'll take a look at the tests before merging. Thanks!

PR #3044, opened 2022-10-24
Fix URL encoding again

DingDongSoLong4:
Fixes an issue where if the () characters are used in query strings (ie regex filters) then they get replaced with {} if the page is refreshed or if the URL is copied into a new tab. This happens because the browser replaces the force URL-encoded () characters (ie %28 and %29) back into () - which is perfectly legal according to the spec - causing them to be changed into {} when the URL is decoded again.

This fixes that hopefully for the last time by only replacing {} with () for the actual JSON object delimiters, and leaving brackets inside strings untouched.


WithoutPants:
I'm not able to reproduce the original issue here. 

I tried filtering scenes on title with regex `(foobar)` and it appears to be being decoded correctly on the `develop` branch already. Can you provide an example that decodes incorrectly?

DingDongSoLong4:
Ah I've just checked, it seems to only happen on Firefox. I'm also pretty sure that it wasn't happening the last time I fixed URL encoding so an update probably changed something since then.

On Firefox and `develop`, if you copy in the URL http://localhost:9999/scenes?c=(%22type%22:%22title%22,%22value%22:%22%28foobar%29%22,%22modifier%22:%22MATCHES_REGEX%22) it gives a regex filter of `(foobar)` the first time, but if you then just select the URL and hit enter a second time it changes to `{foobar}`. The `%28` and `%29` escapes are sent the first time, but the second time they are converted to `()`.

WithoutPants:
Yep, reproduced it. I've refactored to reduce the repeated code a little bit. Can you do a quick sanity check to ensure it's working the you expect it to? Looked good on my end.

DingDongSoLong4:
Looks good, and still working as expected. Thanks!

PR #3054, opened 2022-10-26
Adding info about duplicate files in various places in the UI

puc9:
This will show information about duplicate files (not scenes) in these places in the UI:

- In the Scene Duplicate Checker:

1 performer
5 duplicate files:
image

- In the scene details panel:

image

- Show file modification time in the file info panel:

image

This should help identifying oldest/newest file to make primary if needed.


kermieisinthehouse (approved):
Code reviewed and tested. @WithoutPants, do you think we can squeeze this into the release?

WithoutPants:
I've made a number of changes to this PR. They are mostly around the idea that multiple files assigned to a scene don't _necessarily_ mean they are duplicates.

I removed the warning in the scene details panel and instead added a counter on the `File Info` tab (only where file count > 1):

image

I changed the hover text in the Scene Duplicate Checker to say `x files` instead of `x duplicate files`.

I also added the counter and file modification time to the image and gallery pages for consistency.

Thanks for the submission!

PR #3055, opened 2022-10-26
Track watch activity for scenes.

### LINKED ISSUE #633: [Feature] Play Counter

### LINKED ISSUE #2769: [Feature] Resume marker
For longer videos, its nice to be able to resume playback from where you left off.

I'd like stash to provide an automatic way to create a "resume" marker, so you can pick up where you left off.

I also considered creating a button to quickly create a marker where you are in the video, with a "Left off" or "Last watched" tag perhaps.

**Additional context**

Teda1:
This pull request adds activity-tracking support for videos. Stash will now remember where you left off in a video and will resume from that position when you return to the video. Scene cards will now include a progress bar to show users how far into each video they are. Users can also now sort scenes by recently played to see what video they last watched.

Screenshot 2022-10-26 085006

I added a config option to tell Stash to always start a video from the beginning for users who might not want resume functionality.

Screenshot 2022-10-26 085401

Fixes #2769
Fixes #633


Teda1:
My latest push contains changes to address feature #633. This pull request now tracks watchTime and viewCount and provides options to filter and sort by those new fields. A config option was also created to allow users to set an ignore interval to delay tracking until a set amount of time has passed.

Screenshot 2022-10-27 005432

There is still a bug with sorting by View Count that I need to address which I will look into tomorrow. Some debug code also needs to be cleaned out in typescript files.

yoshnopa:
If I interpret your Code correctly, this would make it relatively easy to add a filter for already started videos, as well as a sorting by 'last watched'. I think both would be a great addition to the project, so I'd be thankful if you consider adding those to the request (or create a separate one of course).

yoshnopa:
If I interpret your Code correctly, this would make it relatively easy to add a filter for already started videos, as well as a sorting by 'last watched'. I think both would be a great addition to the project, so I'd be thankful if you consider adding those to the request (or create a separate one of course).

yoshnopa:
If I interpret your Code correctly, this would make it relatively easy to add a filter for already started videos, as well as a sorting by 'last watched'. I think both would be a great addition to the project, so I'd be thankful if you consider adding those to the request (or create a separate one of course).

yoshnopa:
Awesome, thank you!

RandomName250:
Hey, this looks great! Sorry, I'm new to Stash but will it be possible with this addition of this recently played data for me to import the played data I have stored in a different organizing program? I'm still working on the necessary plugin but I would like to transfer data such as Scene X - played 21.34 Aug 12 2022, 17.59 Feb 14 2022, 18.43 Dec 25 2021 etc. Or at least could this pull be set up to enable such info (I guess it's multiple dates played attached to a scene) to be included and open to plugins in the future? Thanks much!

Teda1:
You should be able to update the activity data via graphql mutations so a script can be written to import your data. However, to keep things simple, this change does not track all the dates an event occurred on. It only keeps track of the date of the most recent activity and increments the play count and duration accordingly. Storing this info by the date of each occurrence is something I'm looking into, but I likely won't have time to get in with this change since it would touch code that I need more time to become more familiar with.

pkg/sqlite/scene.go
```diff
@@ -1416,6 +1437,89 @@ func (qb *SceneStore) imageRepository() *imageRepository {
...
+		"play_count": goqu.L("play_count + 1"),
+	}); err != nil {
+		return 0, err
+	}
+
+	if err := qb.tableMgr.updateByID(ctx, id, goqu.Record{
+		"last_played_at": time.Now(),
+	}); err != nil {
+		return 0, err
+	}
```
WithoutPants, line 1518: This should only call `updateByID` once, using a single combined `goqu.Record`.

pkg/sqlite/migrations/38_scene_activity.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE `scenes` ADD COLUMN `resume_time` float not null default 0;
+ALTER TABLE `scenes` ADD COLUMN `last_played_at` datetime not null;
```
WithoutPants, line 2: This causes an error during migration for existing systems as there is no default and is `not null`. It should accept and default to a null value.

pkg/sqlite/scene.go
```diff
@@ -52,17 +53,21 @@ ORDER BY files.size DESC
...
+	Details      zero.String            `db:"details"`
+	URL          zero.String            `db:"url"`
+	Date         models.SQLiteDate      `db:"date"`
+	Rating       null.Int               `db:"rating"`
+	Organized    bool                   `db:"organized"`
+	OCounter     int                    `db:"o_counter"`
+	StudioID     null.Int               `db:"studio_id,omitempty"`
+	CreatedAt    models.SQLiteTimestamp `db:"created_at"`
+	UpdatedAt    models.SQLiteTimestamp `db:"updated_at"`
+	LastPlayedAt models.SQLiteTimestamp `db:"last_played_at"`
```
WithoutPants, line 67: This should be `models.NullSQLiteTimestamp`.

pkg/models/model_scene.go
```diff
@@ -35,6 +35,11 @@ type Scene struct {
 	CreatedAt time.Time `json:"created_at"`
 	UpdatedAt time.Time `json:"updated_at"`
 
+	LastPlayedAt time.Time `json:"last_played_at"`
```
WithoutPants, line 38: This should be `*time.Time`, where `nil` indicates never played.

graphql/schema/schema.graphql
```diff
@@ -175,6 +175,9 @@ type Mutation {
   """Resets the o-counter for a scene to 0. Returns the new value"""
   sceneResetO(id: ID!): Int!
 
+  """Updates resume time"""
```
WithoutPants, line 178: This needs to be documented in detail what happens in this call. For example, it's not clear that the play count is incremented on this call. 

To that end, I suspect that we might want more finer control over the scene count.

graphql/documents/data/config.graphql
```diff
@@ -60,6 +60,9 @@ fragment ConfigInterfaceData on ConfigInterfaceResult {
   autostartVideo
   autostartVideoOnPlaySelected
   continuePlaylistDefault
+  alwaysStartFromBeginning
+  trackActivity
+  ignoreInterval
```
WithoutPants, line 65: Given that these are exclusively used in the UI, we should consider moving them into the UI properties, instead of adding more backend config handles.

WithoutPants (changes requested):
Stuff I found while testing:
- in the settings, disable the interval setting instead of hiding if activity tracking is disabled.
- set track activity setting to true by default
- we should provide ways to set/reset `resume_time`, `play_duration` and `play_count`, even if these aren't used in the UI
- there's a stale cache issue when clicking the scenes button from the player. I had to refresh before the scene displayed the correct resume time in the scenes list
- going to the next scene in the playlist plays from the resume time, which I've not sure should be the intended behaviour
- export/import needs to include the new fields

Teda1:
I'm just now seeing your feedback. Thanks for going through this. I'll start working my way through your comments soon.

graphql/schema/schema.graphql
Teda1: I meant to update that comment after a few refactors did to that method.

In that last sentence. I assume you mean finner control over the play count?

graphql/documents/data/config.graphql
Teda1: Where is the UI properties file you want this to be moved to?

Teda1:
> there's a stale cache issue when clicking the scenes button from the player. I had to refresh before the scene displayed the correct resume time in the scenes list

I don't think the caching issue you're seeing is new to this change. I've been noticing that Stash doesn't always properly refresh when navigating to a page that has been visited earlier. I'm not yet clear on how this is handled in React/Stash.

graphql/schema/schema.graphql
WithoutPants: Sorry yes, play count.

WithoutPants:
Can you please merge or rebase against the `develop` branch. I've tried to do it, but haven't been able to restore functionality.

Teda1:
I should be able to do this tomorrow. I’ve been busier than usual lately, but things should start to slow down tomorrow.

Teda1:
I've taken care of the merge.

graphql/schema/schema.graphql
```diff
@@ -175,6 +175,9 @@ type Mutation {
   """Resets the o-counter for a scene to 0. Returns the new value"""
   sceneResetO(id: ID!): Int!
 
+  """Update resume time as well as play count and play duration"""
+  sceneSaveActivity(id: ID!, resume_time: Float!, playDuration: Float!): Int!
```
WithoutPants, line 179: I think we should have the external API as follows:

```
"""Sets the resume time point (if provided) and adds the provided duration to the scene's play duration"""
sceneSaveActivity(id: ID!, resume_time: Float, playDurationToAdd: Float): Bool!
```
This allows a client to set the resume time, and/or add to the scene's play duration. It should _not_ increment the play count.

Propose adding a _separate_ interface for incrementing the watch count.

```
"""Increments the watch count for the scene. Returns the new watch count value."""
sceneIncrementWatchCount(id: ID!): Int!
```

Other operations for watch count (setting/resetting or decrementing) can be done via the `sceneUpdate` interface.

graphql/schema/types/scene.graphql
```diff
@@ -52,6 +52,10 @@ type Scene {
   created_at: Time!
   updated_at: Time!
   file_mod_time: Time
+  last_played_at: Time
+  resume_time: Float
+  play_duration: Float
+  play_count: Int
```
WithoutPants, line 58: Please document these fields.

WithoutPants (changes requested):
The current way that the client notifies the server of play activity is not very reliable. It does not get triggered if the user enters a different URL, or if they close the browser.

I'd propose something similar to what Jellyfin looks like it's doing:
- run an interval function over 10 seconds while the video is playing - update resume and play times. This function should also increment the play count if and only if it has not done already, and the minimum amount of play time as been reached.
- when video is paused - update resume and play time
- when video player is destroyed - update resume and play time

graphql/schema/schema.graphql
WithoutPants: See my other comments around this.

ui/v2.5/src/core/StashService.ts
```diff
@@ -830,6 +830,8 @@ export const useTagsDestroy = (input: GQL.TagsDestroyMutationVariables) =>
     update: deleteCache(tagMutationImpactedQueries),
   });
 
+export const useSceneSaveActivity = () => GQL.useSceneSaveActivityMutation();
```
WithoutPants, line 833:
```suggestion
export const useSceneSaveActivity = () => 
  GQL.useSceneSaveActivityMutation({
    update: deleteCache(sceneMutationImpactedQueries),
  });
```

To fix the stale cache issue.

Teda1: Ahh, thanks for this. I'll need to remember this going forwards.

graphql/schema/schema.graphql
Teda1: I just saw your comments after I had already pushed changes to improve the existing approach. I can rework the code work the way you described above.

graphql/schema/types/scene.graphql
Teda1: By document are you asking for comments about these fields?

graphql/schema/schema.graphql
```diff
@@ -177,6 +177,12 @@ type Mutation {
   """Resets the o-counter for a scene to 0. Returns the new value"""
   sceneResetO(id: ID!): Int!
 
+  """Sets the resume time point (if provided) and adds the provided duration to the scene's play duration"""
+  sceneSaveActivity(id: ID!, resume_time: Float, playDuration: Float): Boolean!
+
+  """Increments the watch count for the scene. Returns the new watch count value."""
+  sceneIncrementWatchCount(id: ID!): Int!
```
WithoutPants, line 184: We have an inconsistency in terms (partly my fault). We have `play_count`, `play_duration`, `last_played_at` in the `Scene` type, but `watch` is used here. I'm more partial to `watch` in this context, but I'm not married to the term. One or the other should be changed, though it should be what is the most correct, rather than what is easiest to change.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -242,6 +295,16 @@ export const ScenePlayer: React.FC<IScenePlayerProps> = ({
...
 
     const player = videojs(videoRef.current!, options);
+    var playDurationHandler = window.setInterval(() => {
+      if (trackTime.current) {
+        playDurationRef.current++;
+        if (playDurationRef.current % 10 == 0) {
+          setUpdatePlayDuration(true);
+        } else {
+          setUpdatePlayDuration(false);
+        }
```
WithoutPants, line 305: This is not a great way to be doing the play duration update. Why is it done this way?

WithoutPants (changes requested):
I cleaned up the backend code a bit and added unit tests.

I'm finding that the resume time gets reset when I navigate to a scene page without playing. I assume this is not expected behaviour?

Currently missing code to export the new fields.

I'm getting database locked errors when incrementing the play count. I'm going to investigate this issue separately, so it can stay as a known issue for the moment.

graphql/schema/schema.graphql
Teda1: I landed on `play` after digging around to see what variables other sites and apps used. I can update everything to be consistent with `play`.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
Teda1: I needed to separate the logic that handles set database updates from this counter function. Otherwise, the duration count might be off. My knowledge of React is still limited, so there might be a better way to do this.

Teda1:
> Currently missing code to export the new fields.
The export logic is there. I just ran the function now to be sure it works. Here is a snapshot of the output:
Screenshot 2022-11-15 230735

> I'm finding that the resume time gets reset when I navigate to a scene page without playing. I assume this is not expected behaviour?
This should be fixed now.

> I cleaned up the backend code a bit and added unit tests.
Thanks!

PR #3057, opened 2022-10-28
Performer refactor

WithoutPants:
In preparation for upcoming changes to performers, I've separated the logical `Performer` struct from the sqlite struct, in the same way as was previously done for scenes, images and galleries. The sqlite `PerformerStore` now uses `goqu` for most queries.


Teda1:
I get the following error when attempting to make an edit on an existing performer:
Screenshot 2022-10-30 111847

WithoutPants:
Thanks for the report. This should now be fixed.

Teda1:
Thanks. Everything else from my testing seems good.

PR #3060, opened 2022-10-31
Change performer height to be numeric

### LINKED ISSUE #2113: [Feature] Performer filter: compare option for Height
The current **Height** filter don't have a option to give you performer based of a desired amount. 
Tbh the current option for **Height** are not really useful... (Include ?, regex ?)

Having `greater than`, `less than`, `between` and `not between`.

**Additional context**
**Weight** filter already have it but not **Height** 😋

WithoutPants:
Resolves #2113 

Migrates the performer height values from string to int. *Any non-numeric performer height values will be lost during this migration*.

Adds `height_cm` field to the performer type and filter criteria. The existing `height` field/criterion is deprecated and will convert from string to integer - failing if the conversion fails.

UI is modified to use the number criterion for height.


internal/api/resolver_mutation_performer.go
```diff
@@ -317,7 +329,12 @@ func (r *mutationResolver) BulkPerformerUpdate(ctx context.Context, input BulkPe
 	updatedPerformer.Ethnicity = translator.optionalString(input.Ethnicity, "ethnicity")
 	updatedPerformer.Country = translator.optionalString(input.Country, "country")
 	updatedPerformer.EyeColor = translator.optionalString(input.EyeColor, "eye_color")
-	updatedPerformer.Height = translator.optionalString(input.Height, "height")
+	// prefer height_cm over height
+	updatedPerformer.Height, err = translator.optionalIntFromString(input.Height, "height")
+	if err != nil {
+		return nil, err
+	}
+	updatedPerformer.Height = translator.optionalInt(input.HeightCm, "height_cm")
```
bnkai, line 337: Haven't tested yet, this looks a little weird though.
Shouldnt lline 337 be first and the other code executed only if the height_cm translates to an empty value?
As it is now
- if the input has both fields and the `input.Height` is not valid we will error out and return even  if height_cm is ok and should be updated?
- Height gets overwritten every time so the first part (333-336) doesnt matter unless it errors out. Even if the `return` is removed it doesnt make sense ?

WithoutPants: Thanks, fixed.

Teda1:
I'm not seeing any obvious issues with the recent changes. The only thing I could think to mention is that inputs like `6+9` are allowed and then ignored once the save button is hit. The text field could just accept numbers. There really isn't a need to support `-` either since height and weight should never be negative.

MC1R-13:
> _Any non-numeric performer height values will be lost during this migration_.

That would be my entire database. I painstakingly change all heights to x ft, xx inches. And I've modified several scrapers to do the same.
Why is this necessary?

WithoutPants:
See #2113 for why it's necessary. Non-numeric heights cannot be compared numerically, and cannot be converted simply. I can alter the migration to attempt to convert feet/inch strings into metric. There would need to be a separate PR to _view_ the heights in imperial units however.

I figured that text-based heights were not in use because the UI doesn't handle text-based heights in the details panel - you'd only see the correct formatting in the edit panel.

MC1R-13:
You can also see it under performer list view.
Well, it won't really need to be converted because the format I use keeps the metric values:
140 (4 feet, 7 inches)
So that should make it easier.

WithoutPants:
If that is the format you have, then it will convert using the first value anyway, so you won't lose any information - it will just become `140`.

MC1R-13:
Oh, good to know. 
Thanks for clearing that up.

PR #3069, opened 2022-11-02
Upgrade to go 1.19 and update dependencies

WithoutPants:
Should fix crash when reading some zip files on windows. Requires an update to the compiler image.


kermieisinthehouse (approved):

WithoutPants:
I completely forgot that 1.19 was already released. This PR will be repurposed to migrate to 1.19.

PR #3093, opened 2022-11-07
Replace react-markdown with react-remark

DingDongSoLong4:
In the help menu, there is a noticeable lag every time a new tab is opened. This replaces the react-markdown plugin with react-remark, which is significantly faster and basically removes this lag entirely. I did try updating react-markdown to the latest version first (8.0.3, currently on 7.1.0) but the lag was still there.

This should also improve performance in all the other places where `MarkdownPage` is used.

Edit: okay well this was working when in the debug server but now it isn't in a release build, there's probably something weird going on with the dependencies.


DingDongSoLong4:
Alright, this is working now.
I've also fixed a bug where the manual wouldn't open to the correct page on first open, ie if you refresh on the Tasks tab in the Settings page and click the question mark next to 'Identify' for example, the manual won't open to the the Identify page the first time. Only after opening the manual once will it work correctly.

DogmaDragon:
> Alright, this is working now. I've also fixed a bug where the manual wouldn't open to the correct page on first open, ie if you refresh on the Tasks tab in the Settings page and click the question mark next to 'Identify' for example, the manual won't open to the the Identify page the first time. Only after opening the manual once will it work correctly.

Thanks. this was bothering me, but I never committed to making a bug report.

PR #3106, opened 2022-11-09
Refactor autotag to use individual transactions

### LINKED ISSUE #3091: [Bug Report] "too many SQL variables"
**Describe the bug**
`too many SQL variables` error occurs when auto-tagging a lot of files(over 225k files)

**To Reproduce**
1. Create a new tag that will be applied to a lot of files when auto-tag (in my case, over 225k files)
2. Click on 'Auto Tag'
3. Wait until the operation ends

**Expected behavior**
Files should be auto-tagged with the new tag

**Screenshots**
image
`error auto-tagging tag 'twitter': processing images: error querying images with regex ... too many SQL variables` error occurs
The new tag is not applied to any files.

v0.17.2

 - OS: Windows 10
 - Browser: Chrome
 - Version: 107.0.5304.88

WithoutPants:
Fixes #3091 

Refactors the autotag code (yet again) to query for scenes/images/galleries in batches, and to update them in individual transactions.

### DIFF
--- a/pkg/match/path.go
+++ b/pkg/match/path.go
@@ -315,10 +315,6 @@ func PathToScenesFn(ctx context.Context, name string, paths []string, sceneReade
 			return fmt.Errorf("error querying scenes with regex '%s': %s", regex, err.Error())
 		}
 
-		if len(scenes) == 0 {
-			break
-		}
-
 		// paths may have unicode characters
 		const useUnicode = true
 
@@ -329,6 +325,10 @@ func PathToScenesFn(ctx context.Context, name string, paths []string, sceneReade
 					return fmt.Errorf("processing scene %s: %w", p.GetTitle(), err)
 				}
 			}
+		}
+
+		if len(scenes) < pp {
+			break
 		}
 
 		lastID = scenes[len(scenes)-1].ID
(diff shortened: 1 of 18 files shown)


pkg/match/path.go
```diff
@@ -291,31 +291,53 @@ func PathToScenes(ctx context.Context, name string, paths []string, sceneReader
...
 
-	r := nameToRegexp(name, useUnicode)
-	for _, p := range scenes {
-		if regexpMatchesPath(r, p.Path) != -1 {
-			ret = append(ret, p)
+		r := nameToRegexp(name, useUnicode)
+		for _, p := range scenes {
+			if regexpMatchesPath(r, p.Path) != -1 {
+				if err := fn(ctx, p); err != nil {
+					return fmt.Errorf("processing scene %s: %w", p.GetTitle(), err)
```
bnkai, line 325: I am not sure i get this right. If an error happens during the processing of a single scene the tagging stops altogether due to the return?
Same goes for lines 383,443

Testing seems ok for tagging from a tag, studio, peformer.
Tested also with ~250k images and i can verify the original isssue #3091 is fixed

WithoutPants: That's correct. I don't see that as a huge issue. If we really don't want to stop processing then we can log the error in `fn` instead of returning the error from it.

PR #3111, opened 2022-11-10
Order performers by name in All

WithoutPants:
Fixes issue where performers weren't ordered by name in the select components.


JanJastrow:
Nice, I noticed two days ago that the performers in DLNA where also sorted by `Date created`.
This seems to have fixed this as well. :)

Issue #2344, opened 2022-02-24
[Bug Report] Performer checksum not updated on rename


DampToast:
Edit: You can probably not bother reading this first post, but I'll leave it here in case there's useful info in it.  Steps to reproduce are in the 2nd post and proposed fix in 3rd.

**Describe the bug**
I am trying to create a performer with the name "Simon" but I am getting an error "UNIQUE constraint failed: performers.checksum".  I looked for performers with the name "Simon" using the regex "(?i)^simon$" and nobody shows up.

I did some data base looking and it looks like the hash matches to a performer "Simon Kitty".  I looked in the app code to see what it was building the hash off of and it looked like an md5 hash off the full name.  I did note that I currently have an alias of 'Simon' among others for that performer.

I did some verification that changing the performers name via the UI does change their hash.  I changed "Simon Kitty" to something else, then change it back, and her name hash changed.  I also noticed this with another performer but just assigned that scene to their a duplicate I had of them under another name that they would have been merged to anyway.

**To Reproduce**
Not sure.  I might do some digging and see which performer names, when I rehash them, don't match the hash in the database.  Of note, the two performers that were clashing are tagged with StashDB and I did refresh them this morning.  So maybe the stashdb sync did something.  I'm almost positive these performers were matched to StashDB during scene tagging.

**Stash Version: (from Settings -> About):**
v0.12.0-68-g299d59b5

I'll likely investigate this tonight, just wanted it here in case I forget.

DampToast:
Okay I am 99% sure this is what is going on.

1.  Create a performer named "Aali Rousseau".  No other info needed.
2.  Go to tag view in performer, and search StashDB for her.  It should return "Aali Kali".  Select that as the match and save.
3. Limit your Performer view down to "Aali Rousseau".  Then "Batch Update" with "Current Page" and "Refresh Tagged Performers".  Execute that.
4. Her name is now changed to "Aali Kali" but her checksum did not get updated.  You cannot create a new performer with the name "Aali Rousseau" because of this but you can create a new performer with the name "Aali Kali".

DampToast:
I think the fix to prevent this is probably needed at:
https://github.com/stashapp/stash/blob/develop/pkg/manager/task_stash_box_tag.go#L135

Should be pretty simple putting the md5 of the name into checksum.  However there will also have to be something done when people update to the next version of Stash that redoes all their performer MD5s and deals with conflicts since they can have two performers with the same name via the route I described.  I'm not sure how to do that bit.

[...]

PR #3113, opened 2022-11-10
Performer disambiguation and aliases

### LINKED ISSUE #464: [Feature] Alias of ​​the artist
If you edit the clip and enter the artist name, it would be good if he also displayed the alias names.

### LINKED ISSUE #1162: [Feature] Better handling for aliases (especially in scene details search)
Many performers sometimes have a lot of nicknames, which makes clean maintenance very cumbersome in the long run. Especially if the "right" names don't want to come to mind when entering them in the scene details.

First of all, a clean database structure for aliases is necessary. I thought here of an input of the actor details like for tags:

Aliases1

Aliases2

When entering the names in the details for a scene, the aliases should now also be searched. If a matching alias is displayed for the current input, the regular name can be displayed together with the found alias (only the matching one, not all) and made available for selection:

Aliase in scene details

**Additional context**
I think it adds significant value along with a better performer search ([see here](https://github.com/stashapp/stash/issues/1114))....

### LINKED ISSUE #1724: [Feature] Selecting Aliases of performers with multiple names.
It is annoying when a performer goes by multiple names, which can cause multiple entries in the Database.

I see there is already an Aliases section. If it could just have this section be used in the same way that the usual name is. Perhaps even with a checkbox which means you choose which aliases can be used in this way (may prevent aliases getting too many false hits)

You could also have a merge button where if you have found a performer with more than one entry you can merge them both together.

Or, Much like Studios can have Parent Studios, there could be Parent Performers (Sounds a bit weird but hopefully makes sense). where you can have multiple sub entries connect to the Main Performer. (This sounds like the easiest to do but could cause alot of duplicate data)

### LINKED ISSUE #2507: [Feature] Performer alias(es) in Auto-Tag
When Auto-Tagging Performers in Scenes, the system currently only uses the Performer Name field, and doesn't take the aliases into account.

When Auto-Tagging, use the Performer name and aliases when checking the Scene filename. This could be optional, by adding a flag in the settings "Use aliases in Auto-Tag task"

Changing the Performer name to the alias used in the Scene filename(s), and using the Auto-Tag job / task works correctly. But if the Performer has multiple aliases which are all uses in Scene filenames, you would have to cycle through all aliases and put them in the name field.

And thanks for this great application, keep up the fantastic work!

WithoutPants:
Includes refactoring of the performer object model to use relationship fields in the same way as scenes etc.

Adds disambiguation field. This field is shown in parentheses next to the performer title. This field can be scraped.

image

Converts the Aliases field to be a list of strings, in the same way as tag aliases. Scraped aliases are assumed to be comma-delimited and are split accordingly. The migration splits aliases in the same way.

image

While performer name + disambiguation must be unique, aliases are not enforced as such.

Removes the checksum field.

Updates the auto tag logic to take aliases into account. Disambiguation is _not_ currently used for auto-tagging.

Aliases are included when filtering with the performer selector, but only show `(alias)` suffix if the alias matches.

image

Resolves #2507
Resolves #1724 (I think)
Resolves #1162
Resolves #464


echo6ix:
Two things.

I. **Sorting**. It's easier to parse aliases when they're case insensitive alphanumerically sorted. After each save of the performer profile automatically sort the list of aliases. This would also apply to the migration process when converting the comma delimited string of aliases into a list.

II. **Bulk adding**. The `+` button is fine for a handful of aliases, but it's not uncommon for some performers to have [more than a handful of aliases](https://stashdb.org/performers/b813f145-f273-44bf-bf73-a486acaf3693). Entering every alias for these type of performers one line at a time would be tedious.

Possible solutions:

A. Add another button adjacent to the `+` button that displays a bulk add input form for a comma delimited list.

image

B. Pressing `<enter>` or `<shift>+<enter>` (if a single carriage return would cause too many false positives) in the alias input form automatically displays the next alias input form and moves the cursor to this new input form.

bnkai:
There seems to be an issue when scraping a performer's alias (the existing aliases are not displayed)

Example 
Performer has aliases


When scraping no existing aliases are listed

pkg/sqlite/migrations/40_postmigrate.go
```diff
@@ -0,0 +1,130 @@
...
+	if _, err := m.db.Exec("ALTER TABLE `performers` DROP COLUMN `aliases`"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *schema40Migrator) migratePerformerAliases(id int, aliases string) error {
+	// split aliases by comma
+	aliasList := strings.Split(aliases, ",")
```
bnkai, line 107: In my case i have a few performers where the alias delimiter is `/` (those are from the babepedia scraper, most of the other scrapers use `,` )
Only for the migration i changed the above to 
```golang
       // split aliases by comma or slash
       aliasList := strings.FieldsFunc(aliases, func(r rune) bool { return strings.ContainsRune(",/", r) })
```
I am not sure how many users have the same issue or if its worth it to change this
Some stats from my db
```bash
sqlite> select count() from performers;
5491
sqlite> select count() from performers where aliases is not null and aliases is not "";
3036
sqlite> select count() from performers where aliases LIKE "%,%";
1386
sqlite> select count() from performers where aliases LIKE "%/%";
459
```
3036/5491 performers had aliases and 1386 where comma delimited one, 459 where slash delimited

WithoutPants: Adding a second delimiter to the migration is pretty easy to do. I don't have an issue with including `/`.

WithoutPants:
> I. **Sorting**. It's easier to parse aliases when they're case insensitive alphanumerically sorted. After each save of the performer profile automatically sort the list of aliases. This would also apply to the migration process when converting the comma delimited string of aliases into a list.
> 

Thanks for the feedback. Performer aliases are now sorted in the edit panel.

> II. **Bulk adding**. The `+` button is fine for a handful of aliases, but it's not uncommon for some performers to have [more than a handful of aliases](https://stashdb.org/performers/b813f145-f273-44bf-bf73-a486acaf3693). Entering every alias for these type of performers one line at a time would be tedious.
> 
> Possible solutions:
> 
> A. Add another button adjacent to the `+` button that displays a bulk add input form for a comma delimited list.
> B. Pressing `<enter>` or `<shift>+<enter>` (if a single carriage return would cause too many false positives) in the alias input form automatically displays the next alias input form and moves the cursor to this new input form.

Thanks for the suggestions. I've gone with an approach similar to solution B - please let me know what you think.

The string list input now always shows an empty text field at the end. When you start typing into the field, it automatically adds it to the list and creates a new empty field to tab into. The add button has been removed.

image
image

If you clear the last field, then it is removed from the list. Clearing a field that's not at the end will show an error feedback as per existing behaviour.

This should streamline the approach somewhat. Entering aliases involves entering a value, pressing tab twice, entering the next value and so on. It's not as streamlined as pasting in a comma-delimited list, but it is an improvement.

echo6ix:
@WithoutPants Definitely a big improvement 👍

WeedLordVegeta420:
This is a huge improvement overall. It makes the performer aliases actually useful.

One minor issue is that when using the tagger, the result doesn't automatically resolve names based on alias, you have to manually search. See example below - performer wasn't automatically resolved but can be found by manually searching.

image

kermieisinthehouse:
I'm seeing an error while migrating in this branch:
```
error performing migration: UNIQUE constraint failed: performers.name in line 0: CREATE TABLE `performer_aliases` (
  `performer_id` integer NOT NULL,
  `alias` varchar(255) NOT NULL,
  foreign key(`performer_id`) references `performers`(`id`) on delete CASCADE,
  PRIMARY KEY(`performer_id`, `alias`)
);

CREATE INDEX `performer_aliases_alias` on `performer_aliases` (`alias`);

-- `performers`.`aliases` will be dropped in the post-migration

DROP INDEX `performers_checksum_unique`;
ALTER TABLE `performers` DROP COLUMN `checksum`;
ALTER TABLE `performers` ADD COLUMN `disambiguation` varchar(255);

CREATE UNIQUE INDEX `performers_name_disambiguation_unique` on `performers` (`name`, `disambiguation`) WHERE `disambiguation` IS NOT NULL;
CREATE UNIQUE INDEX `performers_name_unique` on `performers` (`name`) WHERE `disambiguation` IS NULL;
```

kermieisinthehouse:
Above error is from having duplicate named performers, tied to https://github.com/stashapp/stash/issues/2344 . 

There is another issue: filtering performers using the new "aliases" field throws a sql error.

WeedLordVegeta420:
> There is another issue: filtering performers using the new "aliases" field throws a sql error.

I see the same behavior. Looks like it's trying to do a join on the `tags` table instead of the `performers` table.

```
error executing count query with SQL: SELECT COUNT(*) as count FROM (SELECT DISTINCT performers.id FROM performers LEFT JOIN performer_aliases ON performer_aliases.performer_id = tags.id WHERE ((performer_aliases.alias LIKE ? OR performer_aliases.alias LIKE ?))) as temp, args: [%Arisa% %Takarada%], error: error executing `SELECT COUNT(*) as count FROM (SELECT DISTINCT performers.id FROM performers LEFT JOIN performer_aliases ON performer_aliases.performer_id = tags.id WHERE ((performer_aliases.alias LIKE ? OR performer_aliases.alias LIKE ?))) as temp` [[%Arisa% %Takarada%]]: no such column: tags.id
```

kermieisinthehouse:
I think that the user should be able to mark aliases individually for "Do not autotag" - otherwise extremely common aliases like "Mary" can't be used.

WithoutPants:
> I think that the user should be able to mark aliases individually for "Do not autotag" - otherwise extremely common aliases like "Mary" can't be used.

This is a significant change and will mean bumping this from 0.18. The alternative is that we leave the aliases out of the auto-tag behaviour for now to address later.

kermieisinthehouse:
Yes, I don't want to hold up the release for this - I think maintaining the current autotag behavior is probably a good middle ground, if that's an easy change

WeedLordVegeta420:
I do think that leaving the aliases out of the auto-tag for now is probably the better move until there is an option to exclude them individually from auto-tag. Otherwise, people will inevitably upgrade to the latest release and run auto-tag without realizing the impact those one name aliases will have.

kermieisinthehouse:
I can confirm that the most recent commits fix the duplicate performer issue!

WithoutPants:
$150 bounty placed (contribution no. 624029)

PR #3116, opened 2022-11-10
Fix duplicate log messages

DingDongSoLong4:
When a scan is finished, duplicate log messages regarding that scan are created in the logging tab if it is open. The duplicates are only on the UI side and can be cleared by just navigating somewhere else and back, but this prevents them from appearing in the first place.

I've also increased the number of displayed log messages, as well as put a limit on the number of messages kept in memory, preventing infinite growth - something which can occur with the trace log messages of large scans.

### DIFF
--- a/ui/v2.5/src/components/Settings/SettingsLogsPanel.tsx
+++ b/ui/v2.5/src/components/Settings/SettingsLogsPanel.tsx
@@ -1,10 +1,11 @@
-import React, { useEffect, useReducer, useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useIntl } from "react-intl";
 import * as GQL from "src/core/generated-graphql";
-import { useLogs, useLoggingSubscribe } from "src/core/StashService";
+import { useLoggingSubscribe, queryLogs } from "src/core/StashService";
 import { SelectSetting } from "./Inputs";
 import { SettingSection } from "./SettingSection";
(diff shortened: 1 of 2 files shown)


WithoutPants:
This didn't fix the duplicate log message for me, and I can't quite work out how it is supposed to from looking at the code. Can you explain the change?

DingDongSoLong4:
The reason why the bug was happening is as follows:
When a scan completes, the entire Apollo client cache is invalidated (createClient.ts:171). This means that all active queries are refetched - which causes the SettingsLogsPanel component to rerender with a new value for `existingData`. This new value, since it was refetched, contains the log messages generated by that scan. At the same time, the logging subscription was also notified of the new log messages, so the `data` variable also contains the log messages generated by the scan. As these two variables are simply concatenated together, the resulting `filteredLogEntries` contains duplicate entries for every log message that was generated since the component was created.

What my fix does is to make sure to only fetch the initial logs from the server once (using an empty dependency array for the useEffect hook), with a direct query instead of a query hook, and then append them to the stored entry list. After that, only the subscription is used for new logs, so cache invalidations shouldn't cause any duplicates.

It works for me on both Firefox and Chrome - I have no idea why it isn't working for you.

WithoutPants (approved):
Thanks for the explanation. With some further testing I was able to verify the original problem and fix.

PR #3117, opened 2022-11-10
Add UI brotli compression

DingDongSoLong4:
This adds secondary brotli compression to the UI bundle, and adds the extra backend dependency required to serve it. Brotli has been supported in all major browsers since 2019, but the gzipped files are still included and will be served if required.

|   | Gzip | Brotli | Improvement |
| ------------- | ------------- | ------------- | ------------- |
| Total Bundle Size | 2.50MB  | 2.13MB | 14.7% |
| Scene.js | 169.97KB | 144.32KB | 15.1% |
| index.js  | 301.16KB | 248.80KB | 17.4% |
| Binary Size | 34.47MB  | 37.50MB | -8.8% |

The differences are not enormous, but they are there. Technically the gzipped files don't also need to be embedded as statigz can create them in memory at runtime, but that would negatively affect startup time and memory usage so I wouldn't say it's worth it.


WithoutPants:
If we're storing both, what's the benefit?

bnkai:
- Adding brotli should have more effect the more times you serve the static content. In our case wont it be loaded once and then get cached?
- From the used go library <https://github.com/vearutop/statigz/#brotli-support> -> Brotli wont work with plain http by design. I think most of stash's user base do not have https configured
- Brotli doesnt work with older browsers. While most desktop users will be ok there may be issues like we had with dlna and some smart tvs

While adding it as a supporting/secondary option mitigates the above issues, its benefit is also reduced alot and maybe IMHO not worth the extra library/maintenance cost?

DingDongSoLong4:
For plain HTTP, I had checked but not thoroughly enough - it seems that Brotli is being requested by both Chrome and Firefox over HTTP but only on localhost and not on other addresses. That does mostly negate its benefits for most users, and using any compression at all on localhost would be worse anyway. Caching will also reduce its effects so yeah there probably isn't that much benefit left.

The server would only use Brotli if allowed, so there shouldn't be any issues with devices that don't support it. Even if we only embed the Brotli files then statigz would be able to create gzipped versions at runtime.

I just saw the option and thought I might as well add it, and I'll probably keep using it myself, but it seems like it isn't worth it in general - feel free to close this.

PR #3120, opened 2022-11-11
Player mobile improvements

### LINKED ISSUE #2053: [Feature] Add more skipping steps
Missing the opportunity to skip larger steps forward and backwards.

VLC player got a great solution with CTRL and SHIFT plus arrow keys:
- Just the arrow key: 10 seconds
- Shift + arrow: 5 seconds
- Ctrl + arrow: 1 minute

DingDongSoLong4:
These are a few improvements to the player, mainly related to support on mobile/smaller screens.

- Replaces `videojs-landscape-fullscreen` with `videojs-mobile-ui`, which seems to actually be the project that `videojs-landscape-fullscreen` was forked off of initially. This adds double-tap to skip and its own "big play button", so `big-buttons` is no longer needed.
- The player now updates its markers and background if they are changed, without requiring a reload.
- Various different style tweaks for better player control sizing and alignment on both mobile and desktop.
- Implements the modifiers for seeking hotkeys as suggested in #2053, and extends them to `[` and `]` as well.

Fixes #2053


WithoutPants:
Mousing over the marker on the seek bar doesn't show details about the marker any more.

DingDongSoLong4:
Do you mean this? I changed the styles to make it look nicer but it's still there.
image
It doesn't show on mobile but that was the case before as well.

WithoutPants:
Was an existing regression that's fixed in #3141.

WithoutPants:
#2053 says that `Ctrl + left/right arrow` scrubs by 1 minute. The current implementation only scrubs by 1 second, which doesn't seem that useful to me.

WithoutPants:
Please also update the `KeyboardShortcuts.md` file with the updated keybinds.

DingDongSoLong4:
> #2053 says that `Ctrl + left/right arrow` scrubs by 1 minute. The current implementation only scrubs by 1 second, which doesn't seem that useful to me.

My mistake, I must have misread. I've fixed that and removed the modifiers from `[` and `]` to keep the code tidy - they're probably not that useful there anyway as those keys are for quickly scrubbing an entire scene.

I've also updated `KeyboardShortcuts.md` and rebased to develop.

WithoutPants:
I'm seeing a regression for the mobile player: previously it would rotate to landscape for landscape videos when clicking to full screen. This version does not do this. Clicking full screen leaves it in portrait mode.

DingDongSoLong4:
Fixed and rebased, the feature to force landscape is only on the beta version of `videojs-mobile-ui` so I've updated to that.

kermieisinthehouse:
Tested against Firefox Mobile, didn't see any issues.

dumdum7:
Just updated to v0.18.0 and have to say that this is awful from a UX point of view. To pause a video I now have to wait for 2 separate animations: first for the control bar to pop up and then for the pause button. And when I do pause it, the play button disappears, while the control bar stays visible. Why would the play/pause button not be "paired" to the control bar so that both are shown at the same time, like they were before the change? Popular mobile video players like YouTube and Chrome's native video player all do this.

PR #3123, opened 2022-11-13
Update dependencies

DingDongSoLong4:
After not understanding why #3093 was refusing to work in a release build, I decided to just go ahead and try updating all the dependencies - which indeed ended up fixing the problem. I'll update #3093 after this is merged.

This updates all the UI dependencies to the latest versions, with the exceptions being leaving react at v17, react-router at v5 and bootstrap at v4. I have managed to do the react 18 and react-router v6 migration, but that involves a bit more editing so I'll leave that off here. I also attempted to migrate to bootstrap v5 but so many things broke and I very quickly gave up.

The one possible issue is that the formatjs v3 -> v4 migration broke the formatting for the `setup.welcome.config_path_logic_explained` intl string as HTML/XML characters now need to be escaped with an apostrophe. I've applied the fix to all languages, if that's not something I should have done then it can easily be reversed.

Besides that, the only other visible change that I could find is that most of the icons have been tweaked, from the update to fontawesome - but these differences are very minor.

And then I've also replaced `react-slick` with `@ant-design/react-slick` which looks to be better maintained, and removed `query-string` in favour of the native `URLSearchParams`. The latter does make the query filter code a bit more verbose, but we get to drop the dependency and theoretically achieve (very slightly) better performance.


WithoutPants (approved):
In future, I'd prefer to break this up into smaller parts. There's a lot to review and test here and it's very easy for something to slip through the cracks.

DingDongSoLong4:
I probably should have done that, but the reason I didn't was because of the yarn lockfile. I tried to keep things in separate commits though. I had to delete the entire lockfile and rebuild it from scratch to fix the conflicting dependencies which were probably causing the problems with #3093 - I couldn't just do `yarn install`.

I am also now sad because I was almost about to update this again as there have been a bunch of package updates since I last did that. I don't suppose I could quickly do that and have it merged again :)

PR #3133, opened 2022-11-16
Use RatingSystem control in RatingFilter

WithoutPants:
Uses the `RatingSystem` component in `RatingFilter`. Makes it easier to filter by rating and fixes issue where ratings with decimal points could not be filtered.

image

Defaults to 5 stars for single value modifiers and 1 and 5 stars for range modifiers - this is 2 and 10 in decimal rating system. It's not perfect, but provides an initial value at least.

### DIFF
--- a/ui/v2.5/src/components/List/Filters/RatingFilter.tsx
+++ b/ui/v2.5/src/components/List/Filters/RatingFilter.tsx
@@ -1,135 +1,71 @@
-import React, { useRef } from "react";
-import { Form } from "react-bootstrap";
-import { useIntl } from "react-intl";
+import React from "react";
+import { FormattedMessage } from "react-intl";
 import { CriterionModifier } from "../../../core/generated-graphql";
 import { INumberValue } from "../../../models/list-filter/types";
 import { Criterion } from "../../../models/list-filter/criteria/criterion";
-import {
-  convertFromRatingFormat,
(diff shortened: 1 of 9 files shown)


skier233:
Tested and looks good to me

ui/v2.5/src/components/List/Filters/RatingFilter.tsx
```diff
@@ -1,135 +1,71 @@
...
-                : criterion.value?.value2,
-              ratingSystem
-            ) ?? ""
-          }
-          placeholder={intl.formatMessage({ id: "criterion.less_than" })}
-        />
-      </Form.Group>
+    // ensure values are set
+    if (!criterion.value.value || !criterion.value.value2) {
+      onValueChanged({ value: 20, value2: 100 });
```
Teda1, line 56: Is there a reason we aren't starting from 0-100? Decrementing from 0.1 resets to 2.0.

WithoutPants: Good pickup. I've removed the default value code. Instead, I've made it so that the `Add/Update Filter` button is disabled if the criterion is invalid - rating not set or between range is invalid.

WithoutPants:
Also added tweak to styling for decimal rating on the performer page to address the issue that @JanJastrow reported.

PR #3139, opened 2022-11-16
Gitignore update

JackDawson94:
Minor change, adding a few gitignore lines.
Mainly the stash Config file and generated data if you run stash.


WithoutPants:
It's really not a good idea to run stash from the source directory.

JackDawson94:
That was a quick comment :) 
Yeah I thought it wasn't the recommended way of running it ^^

Can cancel the PR if you think it doesn't make sense

WithoutPants:
Discord lets me know when new stuff is posted :)

I think that we shouldn't encourage running stash from the source directory, and that having stash artifacts in the gitignore is probably not a good idea.

PR #3146, opened 2022-11-17
Performer Active Years is submitted to Stashbox

JackDawson94:
When submitting a performer to Stashbox, this allows the Active Years to be submitted.

Only works if the CareerLength field is using the "standard" format set by `formatCareerLength` something like `2009 - 2020` (also works if only start year is set)

(submitting a PR to StashBox too, to allow the UI to receive that info)


JackDawson94:
Note: This won't work until StashBox also updates, but won't generate any errors in the meantime.

kermieisinthehouse (approved):
Reviewed but not tested

PR #3157, opened 2022-11-19
Sort duplicate scenes by path

alexandra-3:
# Problem

I do have many duplicates I pretty much have to live with (CGI porn often comes in slightly different variants that are almost identical except for a few details, or even just the audio).

# Solution

In the absence of any more sophisticated solution (filter/exclusion rule, etc.) I have implemented a simple solution that sorts by path instead of the current sorting, which is a bit meaningless in this context (scene ID).

### DIFF
--- a/pkg/sqlite/scene.go
+++ b/pkg/sqlite/scene.go
@@ -1627,11 +1627,11 @@ func sortByPath(scenes [][]*models.Scene) {
 		firstPathJ := getFirstSortedPath(scenes[j])
 		return firstPathI < firstPathJ
 	}
-	sort.Slice(scenes, lessFunc)
+	sort.SliceStable(scenes, lessFunc)
 }
 
 func getFirstSortedPath(scenes []*models.Scene) string {
-	paths := make([]string, len(scenes))
+	paths := make([]string, 0)
 	for _, scene := range scenes {
 		paths = append(paths, scene.Path)
 	}
--- a/pkg/sqlite/scene.go
+++ b/pkg/sqlite/scene.go
@@ -1623,19 +1623,19 @@ func (qb *SceneStore) FindDuplicates(ctx context.Context, distance int) ([][]*mo
 
 func sortByPath(scenes [][]*models.Scene) {
 	lessFunc := func(i int, j int) bool {
-		firstPathI := getFirstSortedPath(scenes[i])
-		firstPathJ := getFirstSortedPath(scenes[j])
+		firstPathI := getFirstPath(scenes[i])
+		firstPathJ := getFirstPath(scenes[j])
 		return firstPathI < firstPathJ
 	}
 	sort.SliceStable(scenes, lessFunc)
 }
 
-func getFirstSortedPath(scenes []*models.Scene) string {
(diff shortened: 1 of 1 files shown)


kermieisinthehouse (approved):
Reviewed but not tested

alexandra-3:
Thanks for the review @kermieisinthehouse.

Incidentally I did find a bug (now fixed). The slice was initialized to the target length, and then we appended some more, resulting in empty slice entries screwing up the ordering.

I'm now confident that it's working as intended.

pkg/sqlite/scene.go
```diff
@@ -1615,5 +1616,26 @@ func (qb *SceneStore) FindDuplicates(ctx context.Context, distance int) ([][]*mo
...
+}
+
+func getFirstSortedPath(scenes []*models.Scene) string {
+	paths := make([]string, 0)
+	for _, scene := range scenes {
+		paths = append(paths, scene.Path)
+	}
+	sort.Strings(paths)
+	firstPath := paths[0]
+	return firstPath
```
bnkai, line 1640: Since you only need the first string dont you only need to compare the path and keep the `smallest`? You dont need to sort the whole list.
eg something like (untested)
```golang
	var firstPath string
	for i, scene := range scenes {
		if i == 0 {
			firstPath = scene.Path
		} else {
			if scene.Path < firstPath {
				firstPath = scene.Path
			}
		}

	}
	return firstPath
```
Also in theory you could get an empty list for this function then you would have an issue with the `paths[0]`?

alexandra-3: Thanks for the review @bnkai.

While I don't think either point was a major issue (we're sorting a very small collection, and in this code path we know the list is not empty), I have nonetheless refactored the code along the lines you suggest and the code is indeed better now.

pkg/sqlite/scene.go
```diff
@@ -1615,5 +1616,26 @@ func (qb *SceneStore) FindDuplicates(ctx context.Context, distance int) ([][]*mo
...
+}
+
+func getFirstPath(scenes []*models.Scene) string {
+	var firstPath string
+	for _, scene := range scenes {
+		if firstPath == "" || scene.Path < firstPath {
+			firstPath = scene.Path
+		}
+	}
+	return firstPath
```
bnkai, line 1640: This is not exactly correct 
If some of the scenes contain an empty `""` path (file-less scenes ?) then the `firstPath` will be wrong
It might be ok for the current use case but if someone wants to re-use the function/s ( mostly sortByPath ) the prerequisites/exact use-case you mention will not be noticed.
For example with paths  "/home/a.mp4" "/home/b.mp4" "" "/home/c.mp4" "/home/d.mp4" you will  get "/home/c.mp4" as the first path

Changing the if condition in your code should be enough
```golang
	var firstPath string
	for i, scene := range scenes {
		if i == 0 || scene.Path < firstPath {
			firstPath = scene.Path
		}
	}
	return firstPath
```

alexandra-3: Fixed that.

Phew, I think this time it's finally correct :)

PR #3182, opened 2022-11-24
Handle bad funscript at values

### LINKED ISSUE #3181: [Bug Report] Huge numbers in the heatmap generator crash stashapp
**Describe the bug**
If the time index exceeds a certain limit, it causes an index out of range in the heatmap generator, and crashes stashapp

**To Reproduce**
1. Save this as dummy.funscript `{"actions":[{"at":8533811180601344000,"pos":100}],"inverted":false,"metadata":{"creator":"","description":"","duration":1434,"license":"","notes":"","performers":[],"script_url":"","tags":[],"title":"","type":"basic","video_url":""},"range":100,"version":"1.0"}`
2. Put some random video next to it named dummy.mp4
3. Generate heatmaps

**Expected behavior**
I expect this exception to be caught and the heatmap generation for this file to be skipped

**Stash Version: 0.17.2:**

**Additional context**
The crash exception:
[crash.log](https://github.com/stashapp/stash/files/10079574/crash.log)
I also had to use strace to debug this, i'd like if there was a "generating heatmap for: dummy.mp4" in the log instead.

WithoutPants:
Fixes #3181 

This PR throws out any `at` values that are outside the duration range of the scene primary file. This means that generated graphs will only show the ranges within the duration of the scene.

### DIFF
--- a/internal/manager/generator_interactive_heatmap_speed.go
+++ b/internal/manager/generator_interactive_heatmap_speed.go
@@ -11,16 +11,18 @@ import (
 	"sort"
 
 	"github.com/lucasb-eyer/go-colorful"
+	"github.com/stashapp/stash/pkg/logger"
 )
 
 type InteractiveHeatmapSpeedGenerator struct {
-	InteractiveSpeed int
-	Funscript        Script
-	FunscriptPath    string
-	HeatmapPath      string
-	Width            int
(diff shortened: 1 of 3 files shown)


bnkai:
Code looks ok.
Can't actually test it though.

PR #3183, opened 2022-11-24
Improve gallery performance

### LINKED ISSUE #3167: 100% CPU usage (on all cores) for several minutes just to show Gallery Page (/galleries)
Hi,

Over 30 inefficient/unindexed SQL queries are done each time you open Galleries pages (/galleries) or change the page. Many of them take several minutes to resolve. 

`IMAGES SIZE 196.4 GB`

`GALLERIES 2,644`

`IMAGES 228,920`

The trace log is attached here (truncated because of stashapp max log size).
[trace.log](https://github.com/stashapp/stash/files/10050810/trace.log)

WithoutPants:
Resolves #3167 (hopefully)

- improves the cover resolver so that it doesn't query all of the images in the gallery. It now queries specifically for a file with basename `cover.jpg`. If it does not find one, then it returns the first image in the gallery by path, again with a query with a limit of 1. This should improve things significantly where there are many images in galleries.
- deprecated and removed usage of the `Gallery.images` resolver. This resolver was hugely inefficient, and was used in the gallery lightbox and gallery page, causing large loading delays for galleries with many images.


pkg/image/query.go
```diff
@@ -96,3 +97,55 @@ func FindByGalleryID(ctx context.Context, r Queryer, galleryID int, sortBy strin
...
+	imageFilter := &models.ImageFilterType{
+		Galleries: &models.MultiCriterionInput{
+			Value:    []string{strconv.Itoa(galleryID)},
+			Modifier: models.CriterionModifierIncludes,
+		},
+	}
+
+	if useCoverJpg {
+		imageFilter.Path = &models.StringCriterionInput{
+			Value: string(filepath.Separator) + "cover.jpg",
```
bnkai, line 137: Will droping the file separator and thus looking for a plain `cover.jpg` in the path get us a lot of false positives?
The existing behavior was having a suffix of `cover.jpg` so we are now requiring  the renaming of the image file to cover.jpg or making a copy of the file with that name.
Whatever is decided the help section -> Galleries needs to be updated

bnkai:
Cant comment on performance as my galleries have a limited number of images.
I didn't encounter any issues though while accessing or viewing the galleries/images.

pkg/image/query.go
WithoutPants: Good pickup. Changed to be consistent with the old behaviour.

PR #3187, opened 2022-11-25
Add filesystem based blob storage

### LINKED ISSUE #2271: [Feature] Store all blob binaries (images) outside of Stash's SQLite database

### LINKED ISSUE #2303: [Feature] Unused scene thumbnails

WithoutPants:
This is a replacement PR for #2491.

Our existing blob storage system is a number of tables containing images - `scene_covers`, `performers_image`, etc. Additionally, scene covers are stored both in the database and in the filesystem, in the `generated/screenshots` directory.

This PR consolidates all of these blobs into one location, either a single table in the database, or in the filesystem. 

Adds a new table `blobs` which is a pair of a string `checksum` value and optionally the blob data. When a blob is added to the system, an MD5 checksum is generated, and this key is used to insert the data into the `blobs` table. If the `filesystem` storage system is used, then the blob data is instead stored in `<blobs>/xx/xx/checksum`, where `<blobs>` is configurable, and `xx` refers to the first four letters of the checksum, used to split the files up. The referring objects now have columns for each of the blobs and the old tables are dropped - eg `movies` now has `front_image_blob` and `back_image_blob` with foreign keys to the `checksum` column.

Migrates all image implementations to this system, including scene covers. Scene covers will be stored in this system only, and the `screenshot` files will not be used unless a cover can't be found in the database. They will not be regenerated.

The cover generation is now split out into a generate task. This means that covers may optionally _not_ generated during scanning, and performed instead in the generate task.

Incidentally, there is now an `Is Missing Cover` scene filter criterion.

The setup wizard is changed to accept a `blobs` location, or to give no location and use the database storage system. An option to change between systems is in the System settings. 

The schema migration _will not_ migrate covers in the `generated/screenshots` directory. This is provided as a post-migration step. This migration will copy the images in the `screenshots` directory into the blob system - either into the database or the blobs filesystem - ignoring those which already have a cover blob. It will optionally delete the `screenshot` files as they are migrated, to save on disk space.

A migration task is implemented to migrate between systems. This will be a mandatory task that occurs when changing between the database and filesystem blob storage systems. The migration will involve reading the blob files in the filesystem and setting them in the database, or vice versa.

Existing systems will migrate their database blobs automatically, and the blob storage system will be set to `database` by default if there are any image blobs in the system, otherwise `filesystem` is the default. `filesystem` is also the suggested default for new systems.

~~This is not quite finished as denoted by the TODOs above, but the core of it is ready for testing and feedback on the concept.~~ 

The only significant drawback to the current approach is that one can no longer preview the files in Windows since the stored files do not have any extension.

Resolves #2303
Resolves #2271


puc9:
Are there any plans to handle scenes that have the cover stored next to the scene file? In those cases the covers should be treated just like the caption files and not copied to the generated/blobs folder.
This is useful in cases where stashapp is used as a library management tool only and the streaming is done with something else that could make use of the covers being next to the scene file (or under a subfolder next to the scene file).

WithoutPants:
> Are there any plans to handle scenes that have the cover stored next to the scene file? In those cases the covers should be treated just like the caption files and not copied to the generated/blobs folder. This is useful in cases where stashapp is used as a library management tool only and the streaming is done with something else that could make use of the covers being next to the scene file (or under a subfolder next to the scene file).

This is a good idea, but should probably be raised as a separate issue.

puc9:
> > Are there any plans to handle scenes that have the cover stored next to the scene file? In those cases the covers should be treated just like the caption files and not copied to the generated/blobs folder. This is useful in cases where stashapp is used as a library management tool only and the streaming is done with something else that could make use of the covers being next to the scene file (or under a subfolder next to the scene file).
> 
> This is a good idea, but should probably be raised as a separate issue.

Totally fair, I can see this change is super large already, no point in pilling even more on it, Once this change makes it into develop I will probably raise this again or try to implement it myself.

internal/manager/manager.go
```diff
@@ -633,6 +632,22 @@ func (s *Manager) Setup(ctx context.Context, input SetupInput) error {
 		s.Config.Set(config.Generated, input.GeneratedLocation)
 	}
 
+	// if blobs path was provided then use filesystem based blob storage
+	if input.BlobsLocation != "" {
+		if !c.HasOverride(config.BlobsPath) {
+			if exists, _ := fsutil.DirExists(input.BlobsLocation); !exists {
+				if err := os.Mkdir(input.BlobsLocation, 0755); err != nil {
```
bnkai, line 639: possible enhancement would be to use os.MkdirAll to create all needed dirs in path and not just the final one? ( probably the same can be applied for the generated and configdir as well)

pkg/sqlite/blob.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (qb *BlobStore) isConstraintError(err error) bool {
+	var sqliteError sqlite3.Error
+	if errors.As(err, &sqliteError) {
+		return sqliteError.Code == sqlite3.ErrConstraint
```
bnkai, line 293: Should be `err.Code == sqlite3.ErrConstraint` instead ? Can not actually test atm

WithoutPants: `err` is an `error` interface and does not have the `Code` field. This checks if `err` is an instance of `sqlite3.Error` and if so, then checks if it's `Code` field is `sqlite3.ErrConstraint`.

ui/v2.5/src/components/Settings/Tasks/DataManagementTasks.tsx
```diff
@@ -507,6 +556,69 @@ export const DataManagementTasks: React.FC<IDataManagementTasks> = ({
...
+              id="migrateSceneScreenshots"
+              variant="danger"
+              onClick={() => onMigrateSceneScreenshots()}
+            >
+              <FormattedMessage id="actions.migrate_scene_screenshots" />
+            </Button>
+          </Setting>
+
+          <BooleanSetting
+            id="migrate-blobs-delete-files"
```
bnkai, line 599: is the id here correct? it is the same as the next Booleansetting and it might explain the delete screenshot files toggle not working correctly for me
Screenshot_2023-03-06-23-13-49-857-edit_com android chrome
( actual screenshot was before latest 2 commits but behaviour is the same after last commit)

bnkai:
From limited testing ( only via phone )

Most functionality seems ok
I encountered an issue when stretch testing the new migrations, I am not sure how to replicate that though or if I did something wrong

Tried the below
Do first migration the one needed for the upgrade
Set file system as preferred db storage and do the blob / scene covers migration ( remove option ticked, didn't use the overwrite as I wasn't sure about it )
Set db as preferred storage and redo the same migration tasks
After that I ended up with a lot of scenes without covers ( I think, i am not sure though, they are ones that had a default one generated during scan and not a scraped/edited )
Also the is missing scene cover filter seems to have issues (maybe due to the above issue) as it includes some scenes with covers.
Will try to reproduce and post more details

Phasetime:
From my tests I can say a few things: 

1.  I found that checking "delete screenshot files" does not actually delete them. Manually deleting them afterwards though does not impact the system so the migration should have been completed fine.
2. To that what @bnkai said: I tested the migrations in both directions and have the same amount of missing scene covers in the is missing cover filter as before the migrations (all from corrupted files). The is missing cover filter doesn't show any scenes that have a cover for me.
3. Going from database -> filesystem -> database increased the database size dramatically for me (from about 3,082 MB to 6,117 MB and I have a blobs table + image_blob column in the performers, movies, scenes and maybe more tables) even though I'm pretty sure I checked delete old data in the database -> filesystem migration.

WithoutPants:
> 1. I found that checking "delete screenshot files" does not actually delete them. Manually deleting them afterwards though does not impact the system so the migration should have been completed fine.

Can you retry and check the debug log? It logs when it removes the screenshot files. I verified that it's working as expected for me.

> 3. Going from database -> filesystem -> database increased the database size dramatically for me (from about 3,082 MB to 6,117 MB and I have a blobs table + image_blob column in the performers, movies, scenes and maybe more tables) even though I'm pretty sure I checked delete old data in the database -> filesystem migration.

I've added a `VACUUM` at the end of the blob migration that should address this.

Phasetime:
> 1. I found that checking "delete screenshot files" does not actually delete them. Manually deleting them afterwards though does not impact the system so the migration should have been completed fine.

I can confirm that it now also deletes *.thumb.jpg and logs that as well. 

> 3. I found that checking "delete screenshot files" does not actually delete them. Manually deleting them afterwards though does not impact the system so the migration should have been completed fine.

This is working now, the database went down to 150 MB 🤯.

I can also confirm that updating covers from the setSceneCoverFromFile plugin (using graphQL [Mutation].sceneUpdate) works.

MikeLund:
Is it to be expected that API responds with `paths.screenshot` even when cover images aren't generated? Same with /scene/{id}/screenshot endpoint; it returns HTTP 200 despite just being the placeholder SVG. This is maybe expected when cover generation was forced, but it feels wrong when covers weren't generated for any specific scene?

WithoutPants:
It's consistent with the performer interfaces, but not with the studio ones. It probably needs a rethink.

PR #3207, opened 2022-11-29
Use the popover property consistently across cards

Teda1:
For some strange reason, performer, studio, and tag cards don't use the popovers property to handle the popovers, while the gallery, image, movie, and scene cards do. This pull request fixes this inconsistency. With this fix, it is now possible to add simple CSS to fix the hovering popover issue than can sometimes be seen in cards.

With this fix, the `margin-top: auto` style on the `hr` element can now consistently position the popover at the bottom of the card. Lucie Novac's card shows the issue without this fix. Salee Lee's card was edited in the image below to simulate this fix.

Screenshot 2022-11-28 190946

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerCard.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerCard.tsx
@@ -217,9 +217,9 @@ export const PerformerCard: React.FC<IPerformerCardProps> = ({
           ) : (
             ""
           )}
-          {maybeRenderPopoverButtonGroup()}
         </>
       }
+      popovers={maybeRenderPopoverButtonGroup()}
       selected={selected}
       selecting={selecting}
       onSelectedChanged={onSelectedChanged}
--- a/ui/v2.5/src/components/Studios/StudioCard.tsx
+++ b/ui/v2.5/src/components/Studios/StudioCard.tsx
@@ -155,9 +155,9 @@ export const StudioCard: React.FC<IProps> = ({
           {maybeRenderParent(studio, hideParent)}
           {maybeRenderChildren(studio)}
           <RatingBanner rating={studio.rating100} />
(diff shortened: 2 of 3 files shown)


DogmaDragon:
How will popovers look when there are 100+ or 1000+ items attached? Like studios or popular tags?

Teda1:
It would behave the same as it did before:
Screenshot 2022-11-28 193613

Teda1:
Here are screenshots of tags from with the change.
Screenshot 2022-11-28 193947
(I did hack in the values to simliate overflow)

DogmaDragon:
> Here are screenshots of tags from with the change. Screenshot 2022-11-28 193947 (I did hack in the values to simliate overflow)
> 
> Also note that this change does not include the margin CSS I mentioned earlier. I intentionally excluded that CSS.

👍 sorry I was thinking about hover, not popover, been doing other things when this pinged and I completely misunderstood what was happening...

HijackHornet:
This broke my css snipset so hard it cant be repaired :/
I used to hide the bottom part when not in "hover" but now that its has a different parent, it cant be done anymore...
https://cdn.discordapp.com/attachments/1032326528857554964/1034946642580869200/SPOILER_firefox_cr49uJwcRI.mp4
I will need custom js to reposition it where it was before in the tree. Sad..

Teda1:
The card components were written with a popover attribute that, before the change, was only properly used by the movie, gallery, and scene cards. The inconsistency there made creating a clean recommendation/front page tricky when I initially did that work. At the time, I could only properly fix the spacing issue on the cards I mentioned earlier. This tweak is the difference between cards that look like:

Screenshot 2023-01-04 190417

Compared to:

Screenshot 2023-01-04 190433

I think it's hard to argue which of the 2 is better. If you want to share your CSS on discord, I could help you with a fix.

PR #3216, opened 2022-11-30
Allow configuration of ffmpeg args

### LINKED ISSUE #2844: [Feature] Implement hardware accelerated decoding
Generating and transcoding are CPU heavy operations, offloading the decoding of videos to the GPU can help performance/speed greatly.

A checkbox enabling hardware accelerated decoding for:
- Generation (of previews and scrubber sprites).
- Live transcoding.
- Task transcoding.

Hardware accelerated decoding can be easily enabled in ffmpeg by prepending `-hwaccel auto` before the input video file.

Another solution could be user customizeable ffmpeg commands.

**Additional context**
Its important to note that hardware accelerated decoding can reduce the quality of the videos.
It should be noted that im unaware of what happens if the selected decoder cant decode the selected video because of resolution problems or some other incompatibility.

WithoutPants:
Resolves #2844.

Allows for the configuration of `ffmpeg` arguments. Adds the following configuration handles:
- `ffmpeg.transcode.input_args`
- `ffmpeg.transcode.output_args`
- `ffmpeg.live_transcode.input_args`
- `ffmpeg.live_transcode.output_args`

Currently, these aren't exposed in graphql or the UI - these must be entered manually by editing the `config.yml` file.

For example:
```
ffmpeg:
  transcode:
    input_args:
    - -foo
    output_args:
    - -bar
  live_transcode:
    input_args:
    - -baz
    output_args:
    - -xyz
```

When live transcoding, this would add `-foo` before the input parameter, and `-bar` before the output parameter. During generate operations - sprite, preview, transcode - it adds `-baz` before the input parameter and `-xyz` before the output parameter.

This will allow configuring hardware decoding by adding the following to `config.yml`:
```
ffmpeg:
  transcode:
    input_args:
    - -hwaccel
    - auto
  live_transcode:
    input_args:
    - -hwaccel
    - auto
```

This will _not_ allow hardware accelerated _encoding_ as that requires changing the video codec parameters, which is a separate and more significant change.

I have not yet decided if this should be added to the UI.


bnkai:
Seems to work ok
Since it was added to the UI it might need some more explaining/examples/in-app help instructions
For example trying the below
ffmpeg
instead of splitting into two args (as mentioned in the example config) will make ffmpeg error out

- For future reference supporting hwaccel via ffmpeg <https://jellyfin.org/docs/general/administration/hardware-acceleration/> requires some effort from the users

floydcg:
Looks like even though you can pass arguments to ffmpeg, its still using libx264 to generate previews. not sure if thats by design or not but if I try and pass -hwaccel auto it gives an error. Passing the specific qsv (for my intel encoder) gives:
 ERRO[2023-01-27 07:23:33] command stderr: Unrecognized option 'hwaccel qsv'.

WithoutPants:
As mentioned in the description, it does not allow changing the encoder, which is a more complex change. 

I think you're seeing that error because you've entered `-hwaccel auto` as a single argument. It should be `-hwaccel` for the first argument, `auto` for the second.

floydcg:
Ahh got it. I'll give that a shot.

Craig Floyd
Classroom Technology Programmer
Electronic Classroom Support Services
513-556-5467
________________________________

MrX292:
@floydcg i think you just doxxed yourself

floydcg:
@MrX292 good call thanks

PR #3226, opened 2022-12-01
Add shortcuts for decimal rating

### LINKED ISSUE #3225: [Feature] Update Rating Shortcuts for Decimals
After the update of the rating system, keyboard shortcuts doesn't reflect the behaviour for the decimal system

Make that, i.e., `r 9` puts a 9,0 rate

apache202119:
Resolves #3225

Not sure what key to use for resetting the rating, as I set `0` for 10.0 already


skier233:
> 

Perhaps you could use ` to reset the rating since it is on the left of 1.

apache202119:
> > 
> 
> Perhaps you could use ` to reset the rating since it is on the left of 1.

That was my first thought but that key it's not on the left for international keyboards.

skier233:
> > > 
> > 
> > 
> > Perhaps you could use ` to reset the rating since it is on the left of 1.
> 
> That was my first thought but that key it's not on the left for international keyboards.

I don't think it's possible to have a key thats in the same spot for everyone. Just have to make a compromise.

apache202119:
> > > > 
> > > 
> > > 
> > > Perhaps you could use ` to reset the rating since it is on the left of 1.
> > 
> > 
> > That was my first thought but that key it's not on the left for international keyboards.
> 
> I don't think it's possible to have a key thats in the same spot for everyone. Just have to make a compromise.

Done

WithoutPants:
I made some changes to this branch. For decimal ratings, I believe accepting two key presses would be better - ie `r 8 5` sets rating to `8.5`. I generalised the repeated rating keybind code into a separate hook and applied this to the other pages. I also updated the manual page accordingly.

erebous:
Sometimes I need to press the key combination (e.g. r 1 0) a couple of times for this to work

PR #3244, opened 2022-12-06
Correct title ordering for objects without titles

### LINKED ISSUE #3229: [Bug Report] Ordering by title is inconsistent for objects without titles
**Describe the bug**
Sorting scenes by title is not case agnostic.

**To Reproduce**
Be sure you have scene titles in different cases, such as: `ABC` and `aa`. Create these scene titles if required.

1. Go to `Scenes` (grid view)
2. Click on `Title` and sort ascending
3. The scenes will be sorted with `ABC` first followed by `aa`.

**Expected behavior**
(I think) expected behavior should be `aa` first followed by `ABC`.

v0.18.0-4-g87cea80e

WithoutPants:
Fixes #3229 

When sorting by title, sorts by the first non-null value of `title` or file `basename` - or folder basename for galleries.

### DIFF
--- a/pkg/sqlite/database.go
+++ b/pkg/sqlite/database.go
@@ -428,6 +428,7 @@ func registerCustomDriver() {
 				funcs := map[string]interface{}{
 					"regexp":            regexFn,
 					"durationToTinyInt": durationToTinyIntFn,
+					"basename":          basenameFn,
 				}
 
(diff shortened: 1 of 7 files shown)


bnkai (approved):
Code looks and tests ok.

PR #3275, opened 2022-12-13
Add DASH streams for VP9 transcoding

DingDongSoLong4:
This is an extension of #3274 to add DASH support. This could replace/supersede the current VP9 livestreaming implementation, but for now this is just another transcoding option.


WithoutPants:
This doesn't look like it's working correctly for me. For some of my scenes, I'm getting a 404 from this URL:
`http://<host>:9999/scene/2/stream.mpd/init_v.webm?resolution=ORIGINAL`

I got a 404 from a `1_v.webm` request in another scene. I don't know why I'm getting 404s for some scenes and not others. They are all mp4 files that can be directly streamed.

WithoutPants (approved):

PR #3276, opened 2022-12-14
Fix and improve captions detection

puc9:
When attempting to match a captions file if an image file with the same name as a video file is present it will be found first:
- sample.en.vtt
- sample.jpg
- sample.mp4

simply because the extension is alphabetically lower than the one for the video file and stash will try to associate the captions file with the image instead of the video file.

This change will find all the files that match the captions filename in the same folder as the captions files just like before, filter out non video files and then assign the captions to all the remaining video files, in case the scene has multiple files with the same name but different extensions (or multiple scenes).

So the new behavior is:
- sample.en.vtt
- sample.jpg - ignored
- sample.mp4 - assign EN captions
- sample.webm - assign EN captions

### DIFF
--- a/pkg/file/video/caption.go
+++ b/pkg/file/video/caption.go
@@ -98,13 +98,22 @@ func AssociateCaptions(ctx context.Context, captionPath string, txnMgr txn.Manag
 	captionPrefix := getCaptionPrefix(captionPath)
 	if err := txn.WithTxn(ctx, txnMgr, func(ctx context.Context) error {
 		var err error
-		f, er := fqb.FindByPath(ctx, captionPrefix+"*")
+		files, er := fqb.FindAllByPath(ctx, captionPrefix+"*")
 
 		if er != nil {
 			return fmt.Errorf("searching for scene %s: %w", captionPrefix, er)
 		}
 
-		if f != nil { // found related Scene
+		for _, f := range files {
+			// found some files
+			// filter out non video files
+			switch f.(type) {
+			case *file.VideoFile:
+				break
+			default:
+				continue
+			}
+
 			fileID := f.Base().ID
 			path := f.Base().Path
 
--- a/pkg/sqlite/file.go
+++ b/pkg/sqlite/file.go
@@ -575,6 +575,23 @@ func (qb *FileStore) find(ctx context.Context, id file.ID) (file.File, error) {
 
 // FindByPath returns the first file that matches the given path. Wildcard characters are supported.
 func (qb *FileStore) FindByPath(ctx context.Context, p string) (file.File, error) {
+
+	ret, err := qb.FindAllByPath(ctx, p)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(ret) == 0 {
+		return nil, sql.ErrNoRows
+	}
+
+	return ret[0], nil
+}
+
+// FindAllByPath returns all the files that match the given path.
+// Wildcard characters are supported.
+func (qb *FileStore) FindAllByPath(ctx context.Context, p string) ([]file.File, error) {
 	// separate basename from path
 	basename := filepath.Base(p)
 	dirName := filepath.Dir(p)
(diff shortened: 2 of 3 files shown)


Teda1:
I can confirm the regression you're seeing. It looks like the code just grabs the first file that matches the prefix instead of specifically looking for video files that match the prefix like it used to.

File scanning seems to be broken with your change, though. I haven't had any luck scanning any videos, regardless of if captions were provided or not. The log message below is from a scan of a video that doesn't provide any captions.
```
22-12-14 11:07:57 Info    Scan finished (3.7701015s)
2022-12-14 11:07:57 Error   error processing "C:\\Users\\Admin\\Videos\\Sample\\Sample.mp4": checking for existing file "C:\\Users\\Admin\\Videos\\Sample\\Sample.mp4": sql: no rows in result set
2022-12-14 11:07:53 Info    Finished adding files to queue. 1 files queued
2022-12-14 11:07:53 Info    C:\Users\Admin\Videos\Sample doesn't exist. Creating new folder entry...
2022-12-14 11:07:53 Info    scanning 1 paths
```

Your change seemed straightforward, but it looks like additional debugging will be required.

puc9:
@Teda1 thanks for testing!

Yes, I know there are some issues still to be resolved, I'm working on them.

I must be blind because I can't find how to turn this PR into a draft one until I properly fix it.

DogmaDragon:
> I must be blind because I can't find how to turn this PR into a draft one until I properly fix it.

It's in the sidebar 
image

puc9:
@DogmaDragon  Thank you! :)

pkg/sqlite/file.go
```diff
@@ -575,6 +575,23 @@ func (qb *FileStore) find(ctx context.Context, id file.ID) (file.File, error) {
 
 // FindByPath returns the first file that matches the given path. Wildcard characters are supported.
 func (qb *FileStore) FindByPath(ctx context.Context, p string) (file.File, error) {
+
+	ret, err := qb.FindAllByPath(ctx, p)
+
+	if err != nil {
+		return nil, err
+	}
```
bnkai, line 583: the existing FindByPath ignored sql.ErrNoRows as well ?
```golang
if err != nil && !errors.Is(err, sql.ErrNoRows) {
               return nil, fmt.Errorf("getting file by path %s: %w", p, err)
        }
```

pkg/sqlite/file.go
```diff
@@ -601,7 +618,7 @@ func (qb *FileStore) FindByPath(ctx context.Context, p string) (file.File, error
 		)
 	}
 
-	ret, err := qb.get(ctx, q)
+	ret, err := qb.getMany(ctx, q)
 	if err != nil && !errors.Is(err, sql.ErrNoRows) {
```
bnkai, line 622: probably 
```golang
if err!=nil
``` 
is  enough now ?

pkg/sqlite/file.go
puc9: Yes, that was the bug, I've validated that it now works correctly.

puc9:
I've validated that the scan now works, and I've removed a log line that I thought was too noisy. Feel free to restore the log line in captions.go:120 if you still want it.

pkg/sqlite/file.go
puc9: I don't think it would be enough as errors.Is is looking through the whole error chain. In any case I wanted to minimize how much I'm changing the code since surprises can pop up from anywhere as this changing of a ErrNoRows into a nil has shown.

pkg/file/video/caption.go
```diff
@@ -98,17 +98,25 @@ func AssociateCaptions(ctx context.Context, captionPath string, txnMgr txn.Manag
...
+			case *file.VideoFile:
+				break
+			default:
+				continue
+			}
+
 			fileID := f.Base().ID
 			path := f.Base().Path
 
-			logger.Debugf("Matched captions to file %s", path)
```
Teda1, line 111: I'd leave this message in place. This message and the one above are here to assist in debugging. If I had remembered these messages were here, I might have been able to identify that this was the issue another user was having.

puc9: Done, removed the last commit

Teda1 (approved):
I've confirmed the file-scanning issue, as well as the regression with subtitles association, has been addressed,

PR #3284, opened 2022-12-16
Allow use of Proxy

### LINKED ISSUE #664: [RFC] Split Stash Traffic
## Question
Is it possible to "split" Stash traffic between "web hosting" traffic (users accessing the webpage, streaming videos, etc.) and "scraping" traffic (look-ups to sites to fill in details on scenes, movies, etc.)?

## Why?
The "web hosting" traffic can be encrypted with SSL, and it's only point-to-point traffic between the Stash app and one of my devices.  No issue there.  The "scraping" traffic, however, involves the Stash app reaching out to many porn sites.  This is traffic that I would prefer NOT be related to my public IP.  If the traffic could be split, the the "scraping" traffic could be routed through a VPN or TOR without impacting the "web hosting" traffic.  I would like to try running Stash on a dedicated Raspberry Pi, and it would be ideal if the "scraping" traffic could be routed through TOR.

## Workaround
For now, I'm running the Stash Docker container attached to an OpenVPN container.  The OpenVPN container allows traffic on my LAN, so I can get to the webpage normally.  All "scraping" traffic should be going through the VPN.  What are others doing to solve this problem?

### LINKED ISSUE #691: SOCKS/Proxy support
SOCKS/Proxy support

Enable the use of proxy servers for scraping of data

JackDawson94:
The allows the use of a HTTP proxy by defining the parameter in the config file.

### Usage
Example Proxy configurations, in config.yml
```
proxy: http://proxy.site:8080
proxy: https://user:password@proxy.site:8888
```

### Functionality
Currently covers:
- [x] XPath/JSON Scrapers
- [x] CDP Scrapers
- [x] Python Scrapers
- [x] StashBox calls
- [x] Plugins
- [x] Image downloading
- [x] FFMpeg download
- [x] Update checks

Everything works as expected according to my testing.
Note that for the Python scrapers to work the proxy conf must contain "http://" or "https://"

SOCKS5 proxies are officially **unsupported** 😊     In practice they work, but crash python & CDP scrapers

### Testing
HTTP proxy tested using tinyproxy running locally
set config.yml to point to tinyproxy and use Stash normally, you should see the requests pop up in the tinyproxy logs (or stdout)

To verify the scrapers are correctly using your Proxy, use a real proxy (not tinyproxy) and the following scraper:
```
name: IP
performerByName:
  action: scrapeXPath
  queryURL: https://v4.ident.me/{}/
  scraper: performerSearch

xPathScrapers:
  performerSearch:
    performer:
      Name:
        selector: //text()
      URL:
        fixed: https://v4.ident.me/

# Last Updated December 16, 2022
```
Use it to scrape a performer, it should return your proxy's IP

Closes #664
Resolves https://github.com/stashapp/stash/issues/691 (partially)


JackDawson94:
Minor change, no longer printing the proxy's config to logs. To avoid leaking credentials if they are included in the url.

bnkai:
- To be more accurate it should be use of proxy during scraping? Downloading ffmpeg, checking for updates, accessing the handy api and other `system` related connections would need a lot more checking/testing?
- Script/python scrapers can do whatever rrequest they want how are they covered? Supporting or not a proxy depends on their code
- The image downloading part after scraping uses a separate http client so  the getImage function https://github.com/stashapp/stash/blob/520967b9d09cff4556e45cf77f714ee473958afe/pkg/scraper/image.go#L83 so it may need to be updated as well (not sure if it would work automatically due to the envs)?
- Same for stashbox requests? https://github.com/JackDawson94/stash/blob/520967b9d09cff4556e45cf77f714ee473958afe/pkg/scraper/stashbox/stash_box.go#L77
- Using the proxy for CDP scraping should also be possible with a bit of extra code <https://github.com/chromedp/chromedp/issues/645#issuecomment-704764992> to take care of the auth

JackDawson94:
My focus for now was on scraping. You are correct we should include the other calls too 😊

**Scripts**
In my testing, the scripts use the env vars (at least on Linux), so if you use a HTTP proxy, python will work out of the box.
SOCKS proxies need an additional python lib & code changes, so they won't work out of the box.

**Stashbox**
In my testing, Stashbox queries went through the proxy, so I'm assuming the lib uses the env vars too.

Adding everything else you mentioned to the list.

JackDawson94:
After a few checks
- getImage goes through Proxy
- Stashbox goes through Proxy
- Handy connectivity is in the frontend, so it will use the browser's proxy settings

I pushed an update to cover:
- Update check connecting to Github
- ffmpeg download


Unless I am still missing something, only thing left is CDP scrapers

bnkai:
@JackDawson94 while testing i just found another corner case (CDP)
My testing setup
- stash runs locally (plain binary)
- CDP through a docker container using an address in same lan
- proxy is running remotely on a vps

The calls to retrieve the `full address` for the cdp calls are also being run through the proxy and thus error out ( the proxy will try to get something like https://192.168.1.x:9222/json/version)

Changing the code so that the `getRemoteCDPWSAddress` function never goes through the proxy seems to fix the case for me
```diff
diff --git a/pkg/scraper/url.go b/pkg/scraper/url.go
index 70f7ff93..c17f54f0 100644
--- a/pkg/scraper/url.go
+++ b/pkg/scraper/url.go
@@ -272,7 +272,9 @@ func getRemoteCDPWSAddress(ctx context.Context, url string) (string, error) {
                return "", err
        }
 
-       resp, err := http.DefaultClient.Do(req)
+       transport := &http.Transport{Proxy: nil}
+       client := &http.Client{Transport: transport}
+       resp, err := client.Do(req)
        if err != nil {
                return "", err
        }
```
Did you test CDP with a path instead of an address or was the CDP address accessible from your tinyproxy?

All the other cases seem to work fine for me through the proxy as you mentioned

JackDawson94:
@bnkai  thanks for detecting this, with this new commit, the Proxy should now be ignored when connecting to localhost and local LAN ips

pkg/scraper/url.go
```diff
@@ -260,3 +300,27 @@ func cdpHeaders(driverOptions scraperDriverOptions) map[string]interface{} {
 	}
 	return headers
 }
+
+func proxyUsesAuth(proxyUrl string) bool {
+	if proxyUrl == "" {
+		return false
+	}
+	reg := regexp.MustCompile(`^(https?:\/\/)(([\P{Cc}]+):([\P{Cc}]+)@)?(([a-zA-Z0-9][a-zA-Z0-9.-]*)(:[0-9]{1,5})?)`)
+	split := reg.FindAllStringSubmatch(proxyUrl, -1)[0]
```
bnkai, line 309: this will panic if no matches are found, first check and then assign to [0]

pkg/scraper/url.go
```diff
@@ -260,3 +300,27 @@ func cdpHeaders(driverOptions scraperDriverOptions) map[string]interface{} {
...
+
+	return len(split) == 0 || (len(split) > 5 && split[3] != "")
+}
+
+func splitProxyAuth(proxyUrl string) (string, string, string) {
+	if proxyUrl == "" {
+		return "", "", ""
+	}
+	reg := regexp.MustCompile(`^(https?:\/\/)(([\P{Cc}]+):([\P{Cc}]+)@)?(([a-zA-Z0-9][a-zA-Z0-9.-]*)(:[0-9]{1,5})?)`)
+	split := reg.FindAllStringSubmatch(proxyUrl, -1)[0]
```
bnkai, line 319: same as above

pkg/scraper/url.go
```diff
@@ -157,6 +159,11 @@ func urlFromCDP(ctx context.Context, urlCDP string, driverOptions scraperDriverO
 				chromedp.UserDataDir(dir),
 				chromedp.ExecPath(cdpPath),
 			)
+			if globalConfig.GetProxy() != "" {
+				url, _, _ := splitProxyAuth(globalConfig.GetProxy())
+				opts = append(opts, chromedp.ProxyServer(url))
+			}
+
```
bnkai, line 166: Doing this here means that the proxy is set only for local CDP instances (using paths) and not remote.
From a quick look it doesnt seem that there is a way to do it for remote allocators via cdp (which  seems fair considering it needs to be done when starting the chromium instance and remote instances are already running)
Imho this is good enough as this is a limitation of the API, we should probably add some more documentation mentioning the use cases / limitations though?

For example i run the headless-shell container like this
```bash
docker run -d -p 9222:9222 --rm --name headless-shell chromedp/headless-shellchromedp/headless-shell
```
and use a local proxy eg `http://192.168.1.2:1080`
With latest PR i can access CDP but as mentioned nothing goes through the proxy when using it
I had to force the use of a proxy when running the container
```bash
docker run -d -p 9222:9222 --rm --name headless-shell -e "HTTP_PROXY=http://192.168.1.2:1080" -e "HTTPS_PROXY=http://192.168.1.2:1080" chromedp/headless-shell
```
to make the traffic go through the proxy (this has the limitation of authenticated proxies not being supported, for those cases one would need to run a local proxy that handles auth internally)

JackDawson94: That is correct, CDP itself needs to be configured to use a Proxy independently. Just like your own browser.
In this context we can only configure the Proxy for stash-backend.
I guess this should probably be clarified in the CDP documentation ? Not sure where this lives.

JackDawson94:
Thanks @bnkai  for the code review, I've amended your findings.

pkg/scraper/url.go
bnkai: I was thinking in the in-app documentation <https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/>
Probably a few stuff (for plain proxy and CDP) here <https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/Configuration.md#scraping>  and then we can update the docs site with more details/examples?

JackDawson94: I added a bit of documentation in-app, it's lightweight, to fit the amount of documentation for other features.
Considering CDP fits in 2-3 lines, I haven't added anything to it, would add unnecessary complexity.

I'm not sure how documentation is going to work in Stash in the future, what goes in-app and what goes to Stash-Docs. For now I think we can stop here, and discuss more on discord #documentation for the future

bnkai (approved):

PR #3326, opened 2023-01-01
Tweak IS/IS NOT NULL date filter

bnkai:
Includes the golang zero value date when checking dates for IS / IS NOT  NULL criterion
I am not sure how but a few of those dates ( 0001-01-01 ) ended up in my database

```bash
sqlite> select count() from scenes where date = "0001-01-01";
42
sqlite> select count() from performers where birthdate = "0001-01-01";
107
sqlite> select count() from performers where death_date = "0001-01-01";
351
sqlite> 
sqlite> select count() from galleries where date = "0001-01-01";
3
sqlite>
```


kermieisinthehouse (approved):
Code reviewed.

PR #3329, opened 2023-01-02
Upgrade react-select version

JackDawson94:
Upgrade react-select to major version 5
https://react-select.com/upgrade#from-v4-to-v5

Issue #3200, opened 2022-11-27
[Feature] Tag colors


xx790:
**Is your feature request related to a problem? Please describe.**

I have a lot of tags. Some tags fall under common categories. I'd like to make tags of the same nature to be more like each other and more distinguishable from other tags visually.

**Describe the solution you'd like**

Typical solution on booru sites: tags of certain kinds are distinguished by color.
stash doesn't have namespaced tags but it has tags graph that can potentially be used. Albeit multiple parents can make it complicated.

There are multiple ways this can be implemented:

- Each tag has an extra field for background color. No inheritance, more manual work but I can live with it;
- Color can be inherited from parent tag(s). Not sure how to resolve conflicts;
- Add semantic markup so I could solve coloring issue in CSS. Specifically, add a data property to tag cards that would list tag parents. Like `data-parents="tag1 tag2"`. Very DIY, mentioning it here in case it is cheap to add.

Color range:
Hue slider might be preferable compared to full RGB range. It might be easier to ensure consistent looks and good contrast in both dark and light theme. (Relevant: [HSL color model](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl) in CSS)

echo6ix:
People are already doing this with userscripts, no?

I think a built-in feature like this should be considered with caution. To mitigate confusion and improve consistency throughout the UI I think the UI needs to adhere to a strict hierarchal color palette for elements (red for destructive elements, bright blue for primary elements, blue/grey for secondary elements, etc). Therefore, IMO a custom *tag colors* feature should be implemented in a way that avoids color collisions with the core UI color palette.

xx790:
I mean, I can approximate it with a userscript to some extent. I will have to put keywords in the tag description, since parent tags are hidden from HTML once there are more than one.
And I have to check whether tampermonkey is happy to work on localhost:port urls.

I just think it is not just me and it can be useful for many people.

When only Hue is specified, it can be toned down (less saturation and lightness closer to regular background) so it won't collide with UI highlights.

AdultSun:
I've already been running a userscript to do this in Stash. It uses code prefixes in the tag name to know which color to use and also hides the prefix in most views. I was already using those leading codes to enforce a custom sort order for my tags already so it was the easiest way to do it. Ping me in Discord (same username) and I can send you a copy. Would be nice to get a native solution though since it's a little hacky.

Linking the issue "Sort_Name for Tags" #3108 since it's related to the current solution for color-coding.

[...]

PR #3330, opened 2023-01-02
Add colors to tags

JackDawson94:
A new feature allowing Tags to be color-coded

ToDo:
- [x] Color can be set / changed
- [x] Pretty UI 😊
- [x] Display tag color in all scenes / movies / performers / ...
- [x] Requested: Configuration setting to enable/disable the feature

For next PR:
- Display tag color when adding tags (requires https://github.com/stashapp/stash/pull/3329 to be merged first)
- Optional: Color inheritance from parent for nested tags


xx790:
related to #3200

Teda1:
There should be a toggle to ignore color values if this toggle doesn't already exist.

JackDawson94:
> 

Not sure I understand your request ?
If you don't want colors, the default will be kept to the current color, so it won't impact anything

Teda1:
If I go through the process of adding color values for all my tags but decide the color doesn't work for certain themes, there should be a quick and easy way to toggle the colors off without removing the color values for every tag.

xx790:
As I mentioned in #3200, throwing some color theory at the problem can help to make it work with any color theme. For example, if a frontend theme can compute a whatever lightness, saturation with a chosen hue.

Teda1:
Something like that could help, but I wouldn't consider it a requirement for this feature. Even with an implementation like that, colors would still be a problem for some themes that desire limited accent colors. If the finished work is implemented in such a way that it is painless to override via CSS, then a toggle may not be necessary.

xx790:
CSS allows to do color computations, so my approach would be something like this:

```css
:root {
    --tag-42-hue: 50;
}

.tag42.badge {
    background: hsl(var(--tag-42-hue) 70% 20%);
}

.tag42.something-else {
    border-color: hsl(var(--tag-42-hue) 80% 50%);
}
```

xx790:
I see an added checkbox for color inheritance.
As I mentioned in #3200, it is what can actually have an arguable behavior and might need a toggle. It forces certain behavior for meta tags.

Teda1:
> CSS allows to do color computations, so my approach would be something like this:

I probably wouldn't include CSS like that with this change. Users providing their own color would expect to get those exact colors when they save the change. That background filter would make color selection unnecessarily difficult. Having that snippet out as custom CSS user could use would be better.

xx790:
Custom CSS may or may not be concerned about variables, this should work as well:
```css
/* custom with explicit color */
.tag42.badge {
    background: red;
}
```
but having access to the variable and semantic classes/other way accessible via CSS would be a plus for those willing to make the most out of it.

Teda1:
A variable approach that users could define in the tag details page, which the backend would append as classes to the HTML element, would be my preferred approach. This would be far more flexible.

Teda1:
Maybe things could be simplified by adding the tag name as an id and then its parents as a class to the HTML element. Users could then key off those values to add whatever flavor they desire, which would not be limited to colors.

xx790:
I'm not talking about *how exactly* the CSS variable gets injected into HTML/CSS by the backend.
There are multiple ways.
The easiest way is most likely to inject it via `style` attribute of the tag.

A unique identifier of the tag would be easiest to inject as an additional class.
By the way, I didn't try hard to understand this PR code, but there might be a problem already - if it inserts a text name as a class, even if properly escaped, it will be a problem for customization.
Numeric ID of the class should better be used.
Properly escaped name can be inserted as a `data-` attribute if necessary.

```html
<div class="tag-item tag-{{tag.id}} badge" style="--tag-hue: {{tag.colorHue}}" data-name="{{htmlEscape(tag.name)}}">
  {{tag.name}}
</div>
```

There is no valid use for HTML `id` attribute.
Adding parent tag identifiers might be interesting, although I haven't thought about uses for that. It would simply require deciding on a naming scheme for those additional classes. Like `tag-parent-{{parent.id}}` for example.

Teda1:
Injecting anything style via the style attribute should be avoided as it is very inflexible. The use for the id attribute, in this case, would be strictly for styling. If a user doesn't care for inheritance, they would focus on theming using the id value. If they cared about inheritance, they would use the class value.  Ex:
```
<span id="tag-ponytail" class="tag-item badge badge-secondary tag-hair-style">
	<div>
		<a href="/tags/780">Ponytail</a>
	</div>
</span>
```

It doesn't look like the current code is going in this direction, but perhaps it should, as it appears there are some big prerequisites for going the current direction

xx790:
I've shown a way to inject variables without style attributes. It is doable but will be harder (will have to generate and inject part of CSS elsewhere).
Using style attribute for injecting variables is a rare valid use case because people behind web standards couldn't agree on different ways of data attribute access from CSS last time I've checked.
No other CSS besides variables is needed in style attributes.
Any theme or custom CSS is free to use variables / use fallback values if variable is missing / totally ignore variables if not needed.

Current PR is what actually seems to be an example of style attribute abuse because it sets direct style properties.

In case this argument goes nowhere, I'll be happy with just tag ids in html and color left completely for custom CSS (minimal goal for #3200).

But do you know how HTML `id` tag works? (or should work if we ignore backwards compatibility with invalid HTML).
Your sample has a couple of bad assumptions:
- a tag is only ever present once in the whole HTML page;
- a tag name only consists of safe characters.

Teda1:
The tag names would need to be sanitized to consist of lowercase alphanumeric characters. The concern about unique IDs existing on the page is valid, though. The HTML itself wouldn't fail because of this. Duplicate IDs are bad practice because selectors expect to find just 1 entry for each ID. I didn't look closely enough at the snippet you posted. That use of the style attribute should actually be fine. The main point I'm trying to make is that I don't think the backend changes and DB bump is necessary for this work.

Teda1:
Using the tag id rather than the tag name would be better now that I think about international characters. I imagine users with hundreds of tags would rather use an existing value than go through their tags again to define an additional value.

xx790:
Sanitized names can still be an issue. For example, tags in non-Latin languages could just be reduced to the same empty string. Numeric tag ID will just work as a safe unique identifier, and it is exposed in URLs anyway.

There are two levels for the solution.

1. DIY, custom CSS - you are only considering this level;
2. UI/DB - PR focuses on this level.

This might be the time to invite other people responsible for stash management and development direction.
I can imagine there might be a compatibility issue and responsibility concerns when considered in connection with stash-box and/or stashdb - from other systems point of view this information should not be stored the same way. That would mean if user still needed his colors across devices in his own local network using his own stash instance - there would have to be a separate table with instance-specific data.
This does not look like a matter of a personal opinion anymore.

If it were me doing the PR, I wouldn't be able to complete it without knowing what acceptable, other than the commonly agreeable minimum.

----

PS. If it were up to me to plan the future development of stash, I'd think about introducing another level:

3. Tag namespaces. Besides other possibilities, it would allow to assign colors to namespaces instead of individual tags, like it works in boorus. Some people might prefer to still be able to override colors for individual tags, so it would still be up for debate.

JackDawson94:
I'll be honest here, custom CSS is not a "user friendly" solution, my intention is to build something any Stash user can use.
It has to have a pretty color-picker UI, buttons, etc...

I'm willing to make an effort with the way it's coded to allow customization for more advanced users who want to use custom CSS, but that is nice-to-have. A toggle in the settings could be part of this.
As far as I know right now in Stash there isn't even a notion of "Themes".

**Re: adding a css class/id for each tag**
Out of scope of this PR, I've chosen to use `style` for a reason: _Keep It Simple !_
If someone finds a simple, efficient and maintainable way to do this they are welcome to contribute to the PR though 😊

**Re: tag namespaces**
That would be nice in the future, and I have many ideas for it 😊 But it's another point to be discussed elsewhere.
If / When that is implemented, tag colors could be linked to the namespace.

xx790:
"Keep it simple" for whom? For the one who wrote the code once, or for those who will use it ever after?

https://github.com/stashapp/stash/wiki/Themes - these work via custom CSS.

CSS properties set directly via style attribute make a really bad service for those who will try to override them in custom CSS, personal or shared.

Teda1:
I assumed the plan was to use hard-coded values. A toggle would be necessary in this case. I think it would save you some trouble to derive the implementation from user feedback. I don't have much interest in the feature, so my feedback here is limited, as I'm mostly concerned with how it can be turned off. You don't want to invest a lot of time into this just to redo things later if your implementation isn't accepted.

xx790:
I need this feature, even though it's not the first in my list.
I'm unlikely to make PRs any time soon because the backend is in Go language.
I'm afraid when yet another feature is kind of there but poorly considered and kind of a pain to use. There are enough of those already in stash.

Teda1:
The solution you posted in the snippet earlier should be straightforward to write if you opt out of defining an additional tag field. This pr is still a work in progress though, so perhaps it shouldn't be dismissed without seeing the final product.

JackDawson94:
I've been thinking of alternatives to using `styles` for this, but for now I haven't found a satisfactory one:

1. having a separate css class for each tag (eg. _tag-135_):  with 200 tags that would mean 200 more css classes, seems like a messy solution...
2. having a class per color (eg. t_ag-20c044_): could work but I don't see the improvement over styles
3. using styled-components: not a fan of adding a new way of doing styling for just this feature, will create maintainability issues in the future
4. css modules: would generate a unique class id for each tag, even worse than solution (1)


So right now I'm not really seeing any better solutions than plain old `styles`.
Happy to get suggestions though 




------
@xx790  your solution below intrigues me, what makes it better than this?
```
<div class="tag-item badge" style="background-color: {{tag.colorHue}}">
{{tag.name}}
</div>
```
> ```
> <div class="tag-item tag-{{tag.id}} badge" style="--tag-hue: {{tag.colorHue}}" data-tag-name="{{attributeEscape(tag.name)}}">
>   {{tag.name}}
> </div>
> ```

xx790:
Ideas behind my snippet:

- only use style attribute to pass variables. This solves three problems:
    - don't have to generate classes for each tag on screen;
    - don't set CSS properties directly, so they can still be overridden in custom styles;
    - allows to reuse the same value on a different property;

- only allow to choose the Hue aspect of the color, not full color. Leave the saturation and lightness to themes (built-in or custom) by using `hsl(var(--tag-hue) 70% 20%);`. This allows to adapt it for any color theme. Can't break down a color value in CSS afaik.

xx790:
> Adding parent tag identifiers might be interesting, although I haven't thought about uses for that. It would simply require deciding on a naming scheme for those additional classes. Like `tag-parent-{{parent.id}}` for example.

I realized this is actually what I very much need. This will allow me to paint tags in custom CSS based on their meta-tag parent and achieve the desired look with the minimal number of CSS rules. I think I might give it a shot somewhere later this month.

JackDawson94:
Added a setting to enable/disable, and now uses css variable to set each tag's color.
@xx790  I hope this is sufficient to allow for customisation of the colors if you use a Theme ?


I'll take a look at inheritance next, might be a bit more complex to implement so not sure it will be in this first release. (Anyways I'll wait for https://github.com/stashapp/stash/pull/3329  to be able to add colors to the select fields when adding tags)

xx790:
Yes, logically, CSS properties are now decoupled from the style attribute.
Since you're making a PR, I expect you should be able to run the code to see your changes, so you should be able to test custom CSS yourself. For example, try to reset the background color and use the same variable for the border color instead.

Regarding the color inheritance:
I can almost guarantee that your implementation won't fit me.
I don't want meta tags to share color with their children. I'm just trying to repurpose whatever is available as a substitute for tag namespaces.
In my opinion, current model was poorly conceived, multiple inheritance system of tags creates more issues than it solves.
Before further depending your code on the current system - I would rather consult with those behind the wheel about further design directions.

The changes I have in mind for id classes - would enable me to do whatever nerdy shit I might come up with, I'm on my own to solve multiple inheritance and inventing coloring rules.

JackDawson94:
I will leave out the last two features:
- color inheritance (this will come in a future PR, let's be agile about this)
- color in the tag-selector when editing tags, this is waiting for https://github.com/stashapp/stash/pull/3123

WithoutPants:
Can you please rebase or merge against the `develop` branch. I tried, but ended up with either yarn issues or styling issues.

Static code review looks ok on a first pass.

PR #3338, opened 2023-01-04
Option to leave stats page blank

Teda1:
This pull request adds an option to leave the stats page blank, making it ideal for custom charts via plugins or userscripts to run without waiting and dealing with load times that would vary based on the size of a user's stash.

Screenshot of new interface setting:
Screenshot 2023-01-04 103932

Screenshot of blank stats page:
Screenshot 2023-01-04 103958

### DIFF
--- a/ui/v2.5/src/components/Settings/SettingsInterfacePanel/SettingsInterfacePanel.tsx
+++ b/ui/v2.5/src/components/Settings/SettingsInterfacePanel/SettingsInterfacePanel.tsx
@@ -203,7 +203,7 @@ export const SettingsInterfacePanel: React.FC = () => {
 
       <SettingSection headingID="config.ui.stat_panel.heading">
         <BooleanSetting
-          id="blank_stats-view"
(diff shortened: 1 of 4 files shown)


WithoutPants:
I think the stats page should be deprecated and its functionality added as an option to the front page. The presence of the explicit stats button/page was done to maintain existing functionality when the front page functionality was added, but I don't think it should maintain its current implementation.

Teda1:
I do recall you mentioning the idea of having a plugin for the content counter that users could place on the front page. With that in mind, it probably doesn't make sense to create the blank stats page option since it will inevitably have to be removed. I do think a dedicated dashboard page would be good to have, as I don't see some users wanting a dashboard on the same page as their content recommendation page. Further discussion about this could be had later though. Feel free to close this pull request in the meanwhile.

Teda1:
This discussion can continue in Discord https://discord.com/channels/559159668438728723/645283324071247911/1063286653990473820.

PR #3343, opened 2023-01-05
Add tenth place precision star rating

Teda1:
This pull request adds an option for tenth-place precision star ratings, as I feel the quarter and half-star options aren't precise enough to consider them over the decimal option. It's always jarring to see so many performers with a 5-star rating with quarter precision.

### DIFF
--- a/ui/v2.5/src/locales/en-GB.json
+++ b/ui/v2.5/src/locales/en-GB.json
@@ -488,7 +488,8 @@
             "options": {
               "full": "Full",
               "half": "Half",
-              "quarter": "Quarter"
+              "quarter": "Quarter",
+              "tenth": "Tenth"
             }
           }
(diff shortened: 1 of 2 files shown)


Teda1:
The intention behind this change is mainly to allow the ability to view the number rating with a tenth-place precision rather than view the stars themselves. The star precision affects the decimal rating in the cards. So the quarter star precision ratings are displayed as 5, 5.75, 5.50, ..etc. That degree of rounding isn’t always ideal.

Teda1:
My latest push includes the star rating value by the stars to make the exact value more readable, as seen in this image:
Screenshot 2023-01-06 225424

WithoutPants:
Looks ok. I think the `0` value should be hidden when rating is unset.

Issue #3298, opened 2022-12-22
[Feature] Heatmap in video player


esge:
It would be great if heatmap for funscripts could be shown in video player, maybe in the scrubber section

Issue #3031, opened 2022-10-21
[Improvement] Funscript related ideas/goals


scruffynerf:
Braindump (and cut and paste of discord posted ideas) to improve how Stash handles funscripts:

Funscript support is great for a single script, but as the tech is evolving, the use of multiple scripts (alternative, or multiaxis) is being more common.  It would be awesome if Stash got ahead of the curve (and thus would become more recommended/used), by supporting this.  Scripts in general and even multi-axis scripts are being more available, as the tech gets cheaper, VR has embraced it (with multiple sites/studios now offering official or unofficial free or paid scripts), etc.

Current support:  https://github.com/stashapp/stash/blob/c43c695f5d795e9a48d788ecf76c6ca8b81b9d4c/ui/v2.5/src/docs/en/Manual/Interactive.md
"For example, a funscript file for video.mp4 must be named video.funscript."

### **Multiple Script support:**

As documented here:
https://github.com/Yoooi0/MultiFunPlayer

```
L0     Up/Down     <video name>.funscript
                              <video name>.stroke.funscript
                              <video name>.L0.funscript
L1     Forward/Backward     <video name>.sway.funscript
                              <video name>.L1.funscript
L2     Left/Right     <video name>.surge.funscript
                              <video name>.L2.funscript
R0     Twist     <video name>.twist.funscript
                              <video name>.R0.funscript
R1     Roll     <video name>.roll.funscript
                              <video name>.R1.funscript
R2     Pitch     <video name>.pitch.funscript
                              <video name>.R2.funscript

TCode v0.2
Axis     Description     Valid file names
V0     Vibrate     <video name>.vib.funscript
                              <video name>.V0.funscript
V1     Pump     <video name>.lube.funscript
                              <video name>.pump.funscript
                              <video name>.V1.funscript
L3     Suction     <video name>.suck.funscript
                              <video name>.valve.funscript
                              <video name>.L3.funscript

TCode v0.3
Axis     Description     Valid file names
V0     Vibrate     <video name>.vib.funscript
                              <video name>.V0.funscript
A0     Valve     <video name>.valve.funscript
                              <video name>.A0.funscript
A1     Suction     <video name>.suck.funscript
                              <video name>.A1.funscript
``` 
In addition, alternative versions (easy vs hard, or fast/slow (double/half), etc for examples) could be supported easily (alt1, alt2, hard, fast, slow, double, modified, etc... )

Ideally, just making it support ```videoname.something.funscript``` 
[...]

PR #3373, opened 2023-01-24
Add ranges to funscript heatmaps

pornstasche:
Apologies in advance for my go code, haven't written many lines of go before this. 

This change improves the visualization of funscript heatmaps by also including the range of the motions. Mentioned in https://github.com/stashapp/stash/issues/3031, but maybe in other places as well. 

I had to increase the resolution quite drastically for the images to not look blocky or miss a lot of detail. Increasing the size also makes sense for https://github.com/stashapp/stash/issues/3298, which I'm also hoping to be able to provide PR for. The size difference in bytes isn't too bad, the average size of my heatmaps went from `~2kb` per heatmap image to `~7kb`.

I also noticed that the current loop that iterates over the actions sometimes creates gaps in the `segments` list, this isn't really noticeable with the current visualization. But since the range would be 0 with this PR it suddenly becomes very obvious, here is an example of a worst case that I found:
 
gap

**The result of the changes**

Here is a before image from the Scenes view:

before

After, with ranges:

after-same-res

After, with ranges and increased resolution:

after


Teda1:
I think with this change, we need to reevaluate if the heatmap still maps sense on the scene cards. This new waveform design is less visually consistent than the previous design. I'm thinking there should either be an option to allow users to continue using the previous heatmap or do what other sites do and simply include an icon to indicate that a scene is interactive since the heatmap will also be present in the scrubber now.

Screenshot 2023-01-29 170123

Here is a before and after image of the heatmap on the scene cards with this change:
Before
Screenshot 2023-01-29 170453

After
Screenshot 2023-01-29 164608

Issue #3311, opened 2022-12-27
[Feature] Set custom strings to be read as 'cover.jpg' in galleries


timelord44:
For people using Stash as a media organizer to generate metadata for other programs such as Kodi, a regex feature similar to exclusions would be useful. For example, in Kodi, gallery covers are automatically applied from "folder.xxx" within the gallery, so it would be neat to be able to tell Stash to use 'folder' instead of 'cover'.

Issue #2720, opened 2022-07-06
[Feature] Manually select gallery cover thumbnail


HybrisF:
Just like in the settings for scenes, performers or movies it would be useful to be able to select which thumbnail is shown for an individual image gallery rather than being forced to having selected a random one.

I think it could be possible to either choose one image from the gallery itself or have an upload button for the gallery thumbnail similar to the one mentioned above.

feederbox826:
duplicate but not closed

Issue #1856, opened 2021-10-15
[Feature Request] Make gallery cover image filename detection more flexible


n0bodyyouknow:
**Is your feature request related to a problem? Please describe.**
Currently the gallery cover image detection is extremely sensitive, and only recognizes a filename that ends in exactly "cover.jpg".  It does not, for example, recognize a filename that ends in "Cover.jpg" or "cover.jpeg"

Depending on where you download image sets from, it's fairly common to find slightly different capitalization of the cover image, and sometimes a "jpeg" file extension instead of "jpg".

**Describe the solution you'd like**
It would be nice if the cover image filename detection was a bit more flexible, and supported:
  - File extensions "jpg" and "jpeg"
  - Ignore capitalization of the word "cover" and file extension at the end of the filename.  (e.g. "Cover.jpg" or "COVER.JPG" or "cover.JPEG" etc..)

**Additional comments**
Stash has no problem supporting images with various derivations of "JPG" and/or "JPEG" file extensions _in general_, it's only the cover detection logic that is very sensitive.

n0bodyyouknow:
Actually, I think to make this more predictable, and consistent with the main Stash image support, it should support any of the image format file extensions from the configuration settings, as opposed to hard-coding it to only "jpg" or "jpeg".

stg-annon:
I think a good solution here that should be simple here is a configurable regex expression to match a gallery cover that defaults to `cover\.*`

n0bodyyouknow:
> I think a good solution here that should be simple here is a configurable regex expression to match a gallery cover that defaults to `cover\.*`

I had kept the scope of my initial report as small as possible, but if we want to make this truly flexible, it would support one or more regex expressions, similar to the current "Excluded Image/Gallery Patterns" setting.  There are different naming conventions used by different studios, it would be very nice to be able to support more than one.

Issue #1432, opened 2021-05-25
[Feature] Gallery custom thumbnail cover


echo6ix:
**Describe the solution you'd like**
Add option to add custom thumbnail covers for each gallery

**Additional context**
There could be two avenues for adding a gallery cover image:

- 1. Add custom thumbnail via gallery edit options
  - Add cover image 
    - Upload image
    - image from Url
    - Delete image
- 2. Add gallery cover via `more menu` button
  - Select image in gallery > More menu button > Set as gallery cover

FlashSpazzbo:
Ideally, this would allow for a cover image for which the filename is not cover.jpg

Or, allow for a cover.jpg that is a duplicate file of another image.

YurikaL:
Related issues: #863 #872

DogmaDragon:
@chenw191 select image in the gallery > ... > Set as Cover.

Maista6969:
Pick an image in the gallery view and then it's in the kebab menu
![image](https://github.com/user-attachments/assets/4264b0ae-0056-4c3e-a03a-082152f52094)

PR #3391, opened 2023-02-03
Improved gallery cover lookup

Ksrx01:
Small PR that changes how gallery covers are looked up. Which should solve (or alleviate) several open issues, including #3311, #1432, #1856, #2720 and #872 

Right now only files ending exactly in `cover.jpg` are used as covers, any gallery missing such a file uses its first image.
These changes essentially replace `CriterionModifierIncludes` with `CriterionModifierMatchesRegex` (performance drop should be negligible), and introduce `gallery_cover_regex` in `config.yml`, which defaults to a value of `(poster|cover|folder|board)(\.jpg|\.jpeg|\.png)$`.

This makes the cover selection function by default much more flexible, but also allows it to be adapted.

**Note:** _First time coding with Go, so I'm unsure if the way the configuration value is passed down to the function isn't total nonsense._


internal/manager/config/config.go
```diff
@@ -129,6 +129,10 @@ const (
 	// rather than use the embedded UI.
 	CustomUILocation = "custom_ui_location"
 
+	// Gallery Cover Regex
+	GalleryCoverRegex        = "gallery_cover_regex"
+	galleryCoverRegexDefault = `(poster|cover|folder|board)(\.jpg|\.jpeg|\.png)$`
```
DingDongSoLong4, line 134: The default regex should probably match any file extension instead of just those three - maybe something like `(poster|cover|folder|board)\.[^\.]+$`.

DogmaDragon: I'm not really a gallery user but aren't they only importing image extensions listed in your `Settings > Library > Media content extensions`? This pretty much ensures that all the extensions will be image related unless the user manually adds something not supported.

DingDongSoLong4 (commented):
Looks good to me, the only addition I would make would be to also add an option on the settings page to avoid having to directly edit the config file. The docs should probably also be updated for these changes as well.

internal/manager/config/config.go
```diff
@@ -635,6 +639,10 @@ func (i *Instance) GetVideoFileNamingAlgorithm() models.HashAlgorithm {
 	return models.HashAlgorithm(ret)
 }
 
+func (i *Instance) GetGalleryCoverRegex() string {
+	return i.getString(GalleryCoverRegex)
+}
```
bnkai, line 644: Maybe add an extra check in to make sure that the regex compiles ok and if not log it and fallback to the default one?

internal/manager/config/config.go
```diff
@@ -129,6 +129,10 @@ const (
 	// rather than use the embedded UI.
 	CustomUILocation = "custom_ui_location"
 
+	// Gallery Cover Regex
+	GalleryCoverRegex        = "gallery_cover_regex"
```
bnkai, line 133: since this will be manualy edited in  the config for now, add some info to the in-app manual in the relevant section? <https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/Configuration.md#advanced-configuration-options>

internal/manager/config/config.go
bnkai: Yeah that adds a bit of an overhead (compile once here and then once more via the database regex driver), i dont think that is too much though considering the processing is done once for every gallery.
We could move the check later on in the save part of the config ( if the option is added in the UI ) so it will not affect performance or skip that for now and just make sure users check their regexes ( eg mentioning <https://regex101.com/?flavor=golang> , <https://github.com/google/re2/wiki/Syntax> ) if it affects performance (havent actually tested it yet)

internal/manager/config/config.go
```diff
@@ -640,6 +641,14 @@ func (i *Instance) GetVideoFileNamingAlgorithm() models.HashAlgorithm {
 }
 
 func (i *Instance) GetGalleryCoverRegex() string {
+	var regexString = i.getString(GalleryCoverRegex)
+
+	_, err := regexp.Compile(regexString)
+	if err != nil {
+		log.Warn(fmt.Sprintf("Gallery cover regex '%v' invalid, reverting to default.", regexString))
```
DingDongSoLong4, line 648: I haven't testing anything myself but I just noticed that the log function used here is technically from the wrong place (plugins), and while I think those just delegate to the normal logging functions you should probably just be using `logger.Warnf(...)` directly.

And also while I'm here I can't help but do some very minor nitpicking: the value from `i.getString` is already in the `regexString` variable so that can be returned directly instead of retrieving the value again.

Regarding the validation, any overhead is probably very minimal unless the regex is really complicated, and as mentioned the check can be moved to the saving section after this is added to the UI anyway.

DingDongSoLong4: My apologies, I should probably have been more clear before - `logger.Warnf` does Printf formatting itself so the `fmt.Sprintf` is redundant - ie it can just be `logger.Warnf("Gallery cover regex '%v' invalid, reverting to default.", regexString)`.

bnkai (approved):
With latest commit looks and tests ok.

Couldn't notice any performance issues.
Gallery page loading times seem the same and the FindGalleries query doesnt seem to be impacted ( a few ms maybe but the timings arent consistent enough to compare )

DingDongSoLong4:
I did some testing and removing the regex check in `GetGalleryCoverRegex()` actually ends up being disastrous if an invalid regex is passed by directly editing the config file. The current version of stash has a bug (my fix in #3428) where an invalid regex causes all filters to be ignored - which means that the query for the gallery cover returns every single image in the entire database. For my database with >1M images, loading 40 covers on the galleries page literally maxed my CPU and took over 10 seconds.

I also measured the time that only the `regexp.Compile` call took and for me it was only around 50µs, so there is no real benefit to removing the check even after the bug is fixed.

This also inspired #3424, adding similar validation to all the other regex settings.

PR #3406, opened 2023-02-07
Set performer disambiguation for names with parentheses

### LINKED ISSUE #3223: [Feature] Import disambiguated names on update to 0.19
If users already have a makeshift disambiguation field in their name field, it might be helpful to detect it, purge it, and add it to the new disambiguation field.

- Detect disambiguated strings in the name field
  - Where we will assume a disambiguated string is any string in the name field between parentheses
  - If detected, purge this string from the name field
- Add the detected string to the new disambiguation field

WithoutPants:
This modifies the schema 42 post-migration code to detect names with parentheses and automatically separate the name into the base name and disambiguation.

For example, the performer name `Jane Smith (Foobar)` becomes `Jane Smith` with disambiguation set to `Foobar`.

Resolves #3223

### DIFF
--- a/pkg/sqlite/migrations/42_postmigrate.go
+++ b/pkg/sqlite/migrations/42_postmigrate.go
@@ -153,18 +153,23 @@ func (m *schema42Migrator) migratePerformersDisam(ctx context.Context) error {
 	logger.Info("Migrating performer disambiguation")
 
 	const (
-		limit    = 1000
+		limit    = 1
 		logEvery = 10000
 	)
 
 	count := 0
+	lastID := 0
 
 	for {
 		gotSome := false
 
 		if err := m.withTxn(ctx, func(tx *sqlx.Tx) error {
 			query := `
-SELECT id, name FROM performers WHERE performers.name like '%(%)%'`
+SELECT id, name FROM performers WHERE performers.name like '% (%)'`
+
+			if lastID != 0 {
+				query += fmt.Sprintf(" AND `id` > %d ", lastID)
+			}
 
 			query += fmt.Sprintf(" ORDER BY `id` LIMIT %d", limit)
 
@@ -186,6 +191,7 @@ SELECT id, name FROM performers WHERE performers.name like '%(%)%'`
 				}
 
 				gotSome = true
+				lastID = id
 				count++
 
 				if err := m.massagePerformerName(id, name); err != nil {
@@ -211,13 +217,13 @@ SELECT id, name FROM performers WHERE performers.name like '%(%)%'`
 }
 
 // extracts the performer name and disambiguation from the name field based on
-// the format "name (disambiguation)". The whitespace is optional.
-var performerDisRE = regexp.MustCompile(`([^(]+)\(([^)]+)\)(.*)`)
+// the format "name (disambiguation)".
+var performerDisRE = regexp.MustCompile(`^((?:[^(\s]+\s)+)\(([^)]+)\)$`)
 
 func (m *schema42Migrator) massagePerformerName(performerID int, name string) error {
 
 	r := performerDisRE.FindStringSubmatch(name)
-	if len(r) != 4 {
+	if len(r) != 3 {
 		// ignore corner case invalid names
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


Teda1:
I ran into the following error during my migration:
```
error performing migration: running post migrations for schema version 42: executing schema changes: executing statement CREATE UNIQUE INDEX `performers_name_disambiguation_unique` on `performers` (`name`, `disambiguation`) WHERE `disambiguation` IS NOT NULL: UNIQUE constraint failed: performers.name, performers.disambiguation
```

It failed on a performer with the name `Milan Sterling (m)`, likely because I have another performer named `Milan Sterling` who did not have a disambiguation. I imagine this case would not be too uncommon for users.

WithoutPants:
> I ran into the following error during my migration:
> 
> ```
> error performing migration: running post migrations for schema version 42: executing schema changes: executing statement CREATE UNIQUE INDEX `performers_name_disambiguation_unique` on `performers` (`name`, `disambiguation`) WHERE `disambiguation` IS NOT NULL: UNIQUE constraint failed: performers.name, performers.disambiguation
> ```
> 
> It failed on a performer with the name `Milan Sterling (m)`, likely because I have another performer named `Milan Sterling` who did not have a disambiguation. I imagine this case would not be too uncommon for users.

Can you verify what performers you have with `Milan Sterling`? Having two entries with the same name should be fine provided the disambiguation value is different.

Teda1:
These are the performers:
Screenshot 2023-02-07 232934

I think the issue is that the female `Milan Sterling` has the same name but doesn't have a disambiguation.

WithoutPants:
In my testing, `Performer` and `Performer (disambig)` migrated correctly, so I think there is something else going on here.

WithoutPants:
Can you look in the logs for entries relating to `Milan Sterling`?

Teda1:
Here is the full log:
[migration_log.txt](https://github.com/stashapp/stash/files/10682139/migration_log.txt)

The only hits for `Milan Sterling` was `Separating "Milan Sterling (m)" into "Milan Sterling" and disambiguation "m"` which was the last action performed before the database restoration.

WithoutPants:
It didn't actually fail on `Milan Sterling` (and I should've picked this up earlier). The migration adds a unique constraint at the end. `Milan Sterling` was the last performer to be migrated, but it was not the cause of this issue. There are a number of performers in your database that aren't being correctly migrated. Specifically, these performers are in the form of:
- `Performer (disambiguation)`
- `Performer (disambiguation) 2`

There's also a bunch in the form of `Perfomer (disambiguation) (disambiguation) 2` which are probably not being migrated correctly. These are corner cases, but the migration should be able to handle these cases correctly.

Teda1:
That makes sense. I've been meaning to update those disambiguations from a very early MetArt scrape. This coming update would seem like the best time to do so. I agree; it is a corner case that likely won't affect most users.

WithoutPants:
Latest version should address the issues reported. 

`Performer (foo) bar` is migrated to `Performer bar` with `foo` as disambiguation.
`Performer (foo) (foo) bar` is migrated to `Performer (foo) bar` with `foo` as disambiguation.

I also identified an issue where one performer was named like `Performer (disamb)` and another like `Performer  (disamb)` and this caused a uniqueness issue after migrating. Changed the behaviour so that whitespace is preserved in the name and disambiguation, only removing the single ` ` character between the performer name and the `(` if present.

Teda1:
Good catch. I actually wasn't aware of the spaces in the performer's name. I've noticed the UI cleans up extra whitespaces making things like this hard to catch. Fortunately, it appears that the issue is only present in 5 of my performers. I imagine that case is more common.

I wonder if it would be better to ignore these corner cases and allow the user to figure out what they want to do with them post-migration. Some users may not be aware of these issues in their names. The following regex should `^((?:[^(\s]+\s)+)\(([^)]+)\)$` only match names that don't contain excess whitespaces and only use 1 disambiguation without any trailing characters. So name like `John (Disamb)`, `John Doe (Disamb)`, etc. I'll be addressing these disambiguation inconsistencies in my database pre-migration, so this isn't a big deal for me personally.

WithoutPants:
I agree. I've changed the migration to handle the simple case only, others are left unchanged.

Teda1 (approved):
I can confirm the migration completed successfully with the latest changes 👍.

PR #3407, opened 2023-02-07
Remove is (not) null modifier for resume and play times

### LINKED ISSUE #3228: [Bug Report] Resume Time and Play Duration should not allow null/not null filters
**Describe the bug**
Resume Time and Play Duration allow for null/ not null filters even though they can never be null

**Expected behavior**
They should not have a null/not null filter option if they can't be null

WithoutPants:
Resolves #3228


Teda1 (approved):
Looks good. I didn't run into any issues during my testing.

bnkai (approved):

PR #3409, opened 2023-02-07
Fix save disabled when creating new object with name

### LINKED ISSUE #3270: [Bug Report] Cannot create performer from search without changing name
**Describe the bug**
When you search for a performer and then press the "new" button, stash auto-populates the name based on the search string. If this name, or any other field, is not altered, then the Save button is disabled. I ran into this bug today while creating placeholder performers today from a list of 200 videos I had just downloaded. I wanted to make sure that all performers existed before scanning them, so that their names could be scraped from the filenames and automatically added to each scene, so I wasn't concerned with filling in any of their information yet. If I searched for just part of their name and then added the rest on the "Create Performer" page, then everything worked fine. But if I searched for their full name and then only wanted to save that on the next page, I couldn't do so. I tried entering text into the name field and then removing it to trigger an event that would enable the button, but the button only became available as long as the contents of the field was different than how it had started. After removing the extra text, the button was disabled again.

**To Reproduce**
1. Go to "Performers" tab
2. Search for any string
3. Click "new" to go to the "Create Performer" form with the name filled in
4. Click "save" and nothing will happen

**Expected behavior**
Although this is an edge case, I expect the save button to be available as long as there is data to save, whether it was entered by the user or populated from a search.

v0.18.0
Build hash: 8a649f02

 - OS: Windows 10
 - Browser chrome
 - Version 108.0.5359.73

WithoutPants:
Fixes #3270

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryEditPanel.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryEditPanel.tsx
@@ -120,6 +120,11 @@ export const GalleryEditPanel: React.FC<
     onSubmit: (values) => onSave(getGalleryInput(values)),
   });
 
+  // always dirty if creating a new gallery with a title
+  if (isNew && gallery?.title) {
+    formik.dirty = true;
+  }
+
   function setRating(v: number) {
     formik.setFieldValue("rating100", v);
   }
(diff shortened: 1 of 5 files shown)


bnkai:
Minor glitch on  new `tag` -> save 

Works fine for the rest

bnkai (approved):

PR #3433, opened 2023-02-13
Fix gallery zip scan context

xWTF:
### Problem

The zip scan is designed to be atomic (uncancellable), but the implementation simply creates a background context, discarding everything from the parent context.

This design leads to the loss of `contextUser`, when the user doesn't use authentication it isn't a big problem.

However, when credentials are set:
- The `contextUser` is lost during the [creation of background ctx](https://github.com/stashapp/stash/blob/8bd5f91e58619ddbd4245c9ffb765978928f2932/pkg/file/scan.go#L577)
- When plugin hook is executed:
  - [MakePluginCookie](https://github.com/stashapp/stash/blob/8bd5f91e58619ddbd4245c9ffb765978928f2932/pkg/session/session.go#L176) function calls [GetCurrentUserID](https://github.com/stashapp/stash/blob/8bd5f91e58619ddbd4245c9ffb765978928f2932/pkg/session/session.go#L126) on the empty background ctx created above
  - Obviously, `ctx.Value(contextUser)` returns nil
  - The plugin cookie is created without `currentUser` value
- Then, the plugin tries to authenticate the graphql API with this cookie:
  - Authentication calls [GetSessionUserID](https://github.com/stashapp/stash/blob/8bd5f91e58619ddbd4245c9ffb765978928f2932/pkg/session/session.go#L96), which returns empty user id
  - The userID [here](https://github.com/stashapp/stash/blob/8bd5f91e58619ddbd4245c9ffb765978928f2932/internal/api/authentication.go#L42) is `nil`
  - However, the credentials are set, so [the authentication fails](https://github.com/stashapp/stash/blob/8bd5f91e58619ddbd4245c9ffb765978928f2932/internal/api/authentication.go#L76)
- And the plugin is unable to access the Graphql API when its hook is triggered by images inside zip, and the user sets up credentials to protect _his/her/their_ stash setup 😢 

### Fix

This is a extraordinarily simple one line change, please refer to the code.

I do hope the stash project merges simple PRs faster, so I don't have to create a branch for each change and have a dirty local worktree for building binary with all the changes ¯\\\_(ツ)\_/¯


WithoutPants:
We have `job.valueOnlyContext` which would be better suited for this purpose. It should probably made public and moved into its own package.

xWTF:
Done. It's not exported so this PR would export `job.ValueOnlyContext` and use it for `zipCtx`.

xWTF:
The `ValueOnlyContext` is moved to `pkg/utils` now, ready for review. @WithoutPants

PR #3474, opened 2023-02-22
Item list refactor and related bug fixes

### LINKED ISSUE #3420: [Bug Report] Performer filter added instead of changed
**Describe the bug**
Switching performers and going back can cause performer filter to break. This is a poor description since I was unsure how to word this, but the reproduction steps should make it clear.

**To Reproduce**
1. Go to a random performer and switch to the Scenes tab.
2. Below a scene, hover over the cast count and click on a _different_ performer.
3. Go to that performer's scenes tab, and reverse the sort order.
4. Go back in your browser.
5. You should be back to the original performer's scenes tab, but the filter will specifically show scenes with both performers (with the default sort order flipped).
6. Refreshing from this point will fix it.

**Expected behavior**
After going back in your browser, the filter should only show scenes with the original performer (until you manually add another)

**Screenshots**

## Step 1 - A random performer's scene page
Step 1

## Step 2 - Second performer's scene page
Step 2

## Step 3 - Reversed sort order
Step 3

## Step 4 - Back to the original performer
Step 4

**Stash Version: (from Settings -> About):** v0.5.0-985-gd80ec1d7

 - OS: Windows 10
 - Browser: Vivaldi
 - Version: 5.6.2867.62 (Stable channel) (64-bit)

DingDongSoLong4:
I've been sitting on these changes for a while, only got around to making a PR for them now:

- 1d451c0 replaces ListHook with ItemList - having the list as a React hook never really made sense in my opinion. The switch to a component allows for cleaner code and fewer and better scoped rerenders, leading to better performance.
- b9ba794 adds caching to the list pagination data, which prevents the pagination component disappearing every time data is loaded, like when 'Next' is clicked for example.
- dde625f7 fixes an issue I came across where clicking on the marker tab on a tag page changes the page title to "Markers | Stash", which then doesn't disappear until navigating somewhere with a different title.
- 6bf2d0d started as a fix for #3420, and ended up becoming a more thorough improvement for the behaviour of item lists across different paths and when multiple lists are on one page.
- f8a8d65 is a tweak to #3408, as that is now causing a redundant query to be sent to the server on first load of an ItemList, with the root cause being that `searchTerm` can be both `""` and `undefined`. It also fixes a small issue where the clear button was not being hidden when the search term was cleared in this way.

Fixes #3420


WithoutPants:
There's something broken when entering a string into the query field. If I type slow enough, it appears to lose the value.

WithoutPants:
I can no longer reproduce the above issue.

The only other thing I think is a bit off is the pagination. The pagination control should only remain if we're paging between results.

DingDongSoLong4:
Fixed and rebased to develop. The pagination is now hidden when loading, except for when paging through results.

WithoutPants:
Starting from an empty query field, if I enter in `foobar`, the debounce is only working on the second and subsequent keypresses, resulting in a brief displaying of `f` results, then for `foobar` (this is confirmed by checking the `graphql` requests). I'm guessing the debounce isn't working on the first keypress.

DingDongSoLong4:
From what I can tell this is also the behaviour on current `develop`, but I'll look into it.

WithoutPants:
> From what I can tell this is also the behaviour on current `develop`, but I'll look into it.

You are correct. I think it became more obvious with these changes, but I can verify that the same issue exists on `develop`.

PR #3476, opened 2023-02-22
add build artefacts to .gitignore

Tildaar:


.gitignore
```diff
@@ -60,6 +60,15 @@ node_modules
...
+
+####
+# stash
+####
+/stash
+ui/**/package-lock.json
+ui/**/yarn.lock
+config.yml
+icon.png
+stash-go.sqlite
```
WithoutPants, line 74: These are not build artefacts. These are files generated when running stash from the source directory. `yarn.lock` is intended to be in the source tree. I'm not sure what `package-lock.json` is.

WithoutPants:
Closing as invalid. Don't run stash with the cwd in the source directory.

PR #3480, opened 2023-02-24
Update go dependencies

WithoutPants:
Obsoletes #3477 

#3477 broke gqlgen generation. This PR includes that PR, and updates some other dependencies to ensure that generation still works.


bnkai:
not sure if it is best dealt here (x/text relevant)
from #3484 it seems that the minimum versions needed for x/net v0.7.0 are the below? 
```
       golang.org/x/net v0.7.0
       golang.org/x/sys v0.5.0
       golang.org/x/term v0.5.0
       golang.org/x/text v0.7.0
```

PR #3495, opened 2023-02-28
[Feature] Development quickstart guide and Makefile additions

mnh86:
Hi, I'm new to stash and its architecture. I was setting things up for the first time.
I documented some of the confusing parts I ran into that might help other new devs.

Also added a two new Makefile tasks to make the instructions easier to follow:
- `make server-start` runs the server via: `go run ./cmd/stash` (keeping with the `ui-start` convention)
- `make server-clean` cleans up local config files in the project folder (ie. config.yml, stash-go.sqlite, generated/)

Let me know if I can help with any changes to this, like if the instructions need corrections or could be expanded upon.


WithoutPants:
It's not intended or advised to run stash from within the source directory. If we want to add a `server-start` target, then I think the target should `cd` to some other gitignored directory. This eliminates the need for adding runtime artefacts to the gitignore file. To clean the runtime artefacts, the target should probably be `server-clean` to be consistent with `server-start`. You will probably need to adjust the `go run` to use `../cmd/stash` instead.

For me personally, I keep my development instances in a directory outside of the source directory.

Since this has come up a few times now, we should explicitly state in the `DEVELOPMENT.md` file that stash should _not_ be run from the root source directory.

mnh86:
Thanks @WithoutPants I can try making those changes and update.

For stating that stash shouldn't be run from the root source directory, is there a reason that could be given?

WithoutPants:
In the shortest possible terms: to avoid polluting the source directory with runtime artefacts generated by stash.

mnh86:
Updated with fixes and moved to run within a `.local/` folder (and updated .gitignore). Thanks for the feedback!

PR #3497, opened 2023-02-28
Improve release notes dialog

DingDongSoLong4:
This is a fix and and two improvements for the release notes dialog:

- When multiple release notes are to be shown, multiple popups are shown despite the first one having each note displayed on it already. This happens after doing a clean installation, for example. The bug is due to `lastNoteSeen` being set to the date of the first release note object which is not the latest one, and is easily fixed by reversing the order of `releaseNotes`.
- Currently multiple release notes are just displayed consecutively in a list. I have improved the display by adding the versions as titles and inserting an `<hr>` between each note.
- The last improvement is to no longer show the release notes dialog after the initial setup. I don't think it makes too much sense to show it to a new user, and besides if in future the list gets very long then a new user will get a massive popup with all the notes in it which is unwieldy.

### DIFF
--- a/ui/v2.5/src/App.tsx
+++ b/ui/v2.5/src/App.tsx
@@ -207,7 +207,7 @@ export const App: React.FC = () => {
 
     return (
       <ReleaseNotesDialog
-        notes={notes.map((n) => n.content)}
+        notes={notes}
         onClose={() => {
           saveUI({
             variables: {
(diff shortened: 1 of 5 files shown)


WithoutPants:
Kudos for addressing something that's been bugging me that I hadn't got around to sorting out.

Issue #2695, opened 2022-06-25
[Feature] Performer scraper result list with image and more


TgSeed:
**Is your feature request related to a problem? Please describe.**
Yes, Sometimes there are several results with aliases and more which I have to go through each one by one to see their Image.
If the images have been right there in the result list, then I was able to better and faster choose the correct one.

**Describe the solution you'd like**
The same scraper results view as scenes.

**Describe alternatives you've considered**
Nothing

Clasp8775:
If I'm not wrong, this issue has been closed by https://github.com/stashapp/stash/pull/3965, right?

- https://github.com/stashapp/stash/pull/3965

PR #2697, opened 2022-06-25
feat: UI - Performer scraper result list with image and basic data

TgSeed:
As the title says, but also added toast and error messages.

closes #2695

PR #3499, opened 2023-03-01
Updated - feat: UI - Performer scraper result list with image and basic data

StashPRs:
This is the same as https://github.com/stashapp/stash/pull/2697, updated for the codebase as it sits currently and improved upon slightly.  I'd definitely appreciate a code review on this, it's been a few years since I've messed around with React.

From the original PR:

> As the title says, but also added toast and error messages.
> closes https://github.com/stashapp/stash/issues/2695

### To the performer scraper query modal:
query modal

### To the performer scraper confirmation modal:
confirmation modal

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerDetails/PerformerScrapeDialog.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerDetails/PerformerScrapeDialog.tsx
@@ -329,13 +329,12 @@ export const PerformerScrapeDialog: React.FC<IPerformerScrapeDialogProps> = (
     )
   );
 
-  const [images, _] = useState<string[]>(
+  const images =
     props.scraped.images && props.scraped.images.length > 0
       ? props.scraped.images
       : props.scraped.image
       ? [props.scraped.image as string]
-      : []
-  );
+      : [];
 
   const allFields = [
     name,


ui/v2.5/src/components/Shared/ScrapeDialog.tsx
```diff
@@ -354,6 +356,176 @@ export const ScrapedImageRow: React.FC<IScrapedImageRowProps> = (props) => {
...
+    if (imageIndex === 0) {
+      return;
+    }
+    const ret = props.result.cloneWithValue(props.images[imageIndex - 1]);
+    props.onChange(ret as T);
+    setImageIndex(imageIndex - 1);
+  }
+
+  function setNext() {
+    if (imageIndex === props.images.length - 1) {
```
USA-RedDragon, line 428: Make this wrap around

ui/v2.5/src/components/Shared/ScrapeDialog.tsx
```diff
@@ -354,6 +356,176 @@ export const ScrapedImageRow: React.FC<IScrapedImageRowProps> = (props) => {
...
+  props: IScrapedImageDialogRowProps<T, V>
+) => {
+  const [imageIndex, setImageIndex] = useState(0);
+
+  function hasNewValues() {
+    return props.newValues && props.newValues.length > 0 && props.onCreateNew;
+  }
+
+  function setPrev() {
+    if (imageIndex === 0) {
```
USA-RedDragon, line 419: Make this wrap around

WithoutPants:
Getting all images and encoding them as base64 is the wrong approach in my opinion. The stash-box implementation returns the URLs to the images only. We should probably return the direct URLs, and they should be loaded only as needed by the UI.

I think that this PR should be split up so that we can solve these issues separately:
- one PR for the query dialog to show age, gender and image
- one PR to support multiple images in the performer scrape dialog - stash-box results already return multiple images, so this can be tested with stash-box
- finally, one PR to change the backend to support multiple images from non-stash-box scrapers. We may need to figure out a way to ensure that the referrer header isn't set when getting each image.

I won't close this PR yet, but I won't be merging it with the current design.

TgSeed:
> Getting all images and encoding them as base64 is the wrong approach in my opinion. The stash-box implementation returns the URLs to the images only. We should probably return the direct URLs, and they should be loaded only as needed by the UI.

I'm not sure but as far as i remember, i have once again talked or wrote regarding this somewhere, maybe in the code comments!

That's the right approach in my opinion!
Just how it requests data directly from the device/system that has Stash running, itself should also fetch everything that can relate to that.

Requesting data with one Device/IP and other things such as assets from another Device/IP can simply leak IP and/or other things, it may not even be what the user really want.
Well, It barely happen, but imagine working in an office with a monitored network and connecting to your home stash, they may not know what are you doing with that IP/Domain, but they will definitely see you connected likely to "cdn.brazzers.com" or such.
That's not my case, I'm just so careful about my data leakages and privacy.

WithoutPants:
> Requesting data with one Device/IP and other things such as assets from another Device/IP can simply leak IP and/or other things, it may not even be what the user really want. Well, It barely happen, but imagine working in an office with a monitored network and connecting to your home stash, they may not know what are you doing with that IP/Domain, but they will definitely see you connected likely to "cdn.brazzers.com" or such. That's not my case, I'm just so careful about my data leakages and privacy.

Good point. I kind of touched the surface of that with the referrer stuff. I think the solution is as follows:
- add a server route that accepts an encoded URL and returns the data, by requesting the URL from the backend. For example: `<server:port>/scrape_data/<encoded URL>`. The server handles this request by decoding the encoded URL, performing the http request using its normal scraper http client, and returns the data, basically like a pass through.
- if a scrape result returns a single image, it can get the data and return it as a base64 chunk, but if there's multiple URLs, then encode them into the above format and return those. From the client's point of view it won't make a difference what the URL.

PR #3500, opened 2023-03-01
Improve dynamic import error message

DingDongSoLong4:
This is an improvement to the `ErrorBoundary` component, specifically to the handling of dynamic import errors for lazy-loaded components. I've seen so many questions or bug reports related to this error after a stash upgrade, so this simply adds a message on the error page to prompt the user to reload or clear the browser cache.
I've also added a minor tweak to display the component trace more nicely on Firefox.

### DIFF
--- a/ui/v2.5/src/App.tsx
+++ b/ui/v2.5/src/App.tsx
@@ -31,32 +31,32 @@ import { ReleaseNotesDialog } from "./components/Dialogs/ReleaseNotesDialog";
 import { IUIConfig } from "./core/config";
 import { releaseNotes } from "./docs/en/ReleaseNotes";
 import { getPlatformURL, getBaseURL } from "./core/createClient";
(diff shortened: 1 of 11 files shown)


WithoutPants:
Is there a reason that the lazy component stuff is using pipe_case instead of camelCase like the rest of the code?

Is there an easy way to trigger the error message for testing?

DingDongSoLong4:
> Is there a reason that the lazy component stuff is using pipe_case instead of camelCase like the rest of the code?
No there isn't, I just forgot what language I was using for a second there.

> Is there an easy way to trigger the error message for testing?
The way I was doing it was to run the binary from this branch alongside the latest develop binary, alternating between the two. Run this branch, load a page, quit it, run the develop branch, navigate to a new page without reloading. That simulates an upgrade to stash without reloading the browser.

PR #3504, opened 2023-03-04
Setup improvements

DingDongSoLong4:
These are some improvements mainly related to the initial setup screen.

- Improve setup redirects to use `react-router` instead of `window.location`
- Prevent the release notes popup from quickly appearing on first setup before redirecting to the setup page
- Add warnings against running the database over a network
- Set the cache directory during setup, with a default of `<config file dir>/cache`

Regarding running the database over a network, this is from the SQLite website: https://www.sqlite.org/useovernet.html. I haven't added it in here but perhaps a warning on the release notes popup for those that have already set up their databases over a network could also be a good idea.

### DIFF
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -6,6 +6,8 @@ input SetupInput {
   databaseFile: String!
   """Empty to indicate default"""
   generatedLocation: String!
+  """Empty to indicate default"""
+  cacheLocation: String!
 }
 
 enum StreamingResolutionEnum {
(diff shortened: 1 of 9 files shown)


WithoutPants:
The setup wizard is now going directly from the confirm step to the front page after creating the system, instead of showing the "finished" page. I also saw the release notes dialog popup briefly during this transition.

DingDongSoLong4:
Huh I could have sworn I fixed that... well fixed it now anyway. The release note popup is due to the early redirect.

WithoutPants (approved):
I think a refactor of the paths page of the wizard should be considered in the future (especially with #3187), as it is getting very verbose.

PR #3507, opened 2023-03-05
Safari skip file transcodes

NodudeWasTaken:
Given that safari doesn't support direct file transcoding, it seems reasonable to disable it.
This commit adds a client sided check for if its a webm or mp4 transcoding stream, and safari, in which case it removes the stream.


kermieisinthehouse (approved):
Code reviewed

DingDongSoLong4:
Were you perhaps using Safari on an iPad to check for WEBM support? I found a bug (fix in #3676) where it appears as if it is playing the WEBM stream, but it is in fact just playing the direct stream. If there isn't an ffmpeg process running on the machine hosting stash, then it's just streaming directly.

I also tried playing a WEBM stream in Safari on macOS 13.3 and it wouldn't work either - Safari just doesn't seem to support it anywhere.

NodudeWasTaken:
> Were you perhaps using Safari on an iPad to check for WEBM support? I found a bug (fix in #3676) where it appears as if it is playing the WEBM stream, but it is in fact just playing the direct stream. If there isn't an ffmpeg process running on the machine hosting stash, then it's just streaming directly.
> 
> I also tried playing a WEBM stream in Safari on macOS 13.3 and it wouldn't work either - Safari just doesn't seem to support it anywhere.

I can play webm on macOS 13.3 (M1), but not on iOS which corresponds to https://caniuse.com/webm
I should check if its mobile safari instead

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -479,21 +480,31 @@ export const ScenePlayer: React.FC<IScenePlayerProps> = ({
...
-          duration,
-        };
-      })
+      scene.sceneStreams
+        .filter((stream) => {
+          const src = new URL(stream.url);
+          const isFileTranscode =
+            src.pathname.endsWith("/stream.mp4") ||
+            src.pathname.endsWith("/stream.webm");
+          const isSafari = UAParser().browser.name?.includes("Safari");
```
WithoutPants, line 489: There's no need to call this for every stream. This can be moved to line 481.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -479,21 +480,31 @@ export const ScenePlayer: React.FC<IScenePlayerProps> = ({
...
-          offset: !isDirect,
-          duration,
-        };
-      })
+      scene.sceneStreams
+        .filter((stream) => {
+          const src = new URL(stream.url);
+          const isFileTranscode =
+            src.pathname.endsWith("/stream.mp4") ||
+            src.pathname.endsWith("/stream.webm");
```
WithoutPants, line 488: Is this not just the inverse of the `isDirect` logic further below? If it is, we can create a local function `isDirect` before line 482 and call that instead to determine if the file is direct or not.

WithoutPants (changes requested):

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
NodudeWasTaken: Should be fixed

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
NodudeWasTaken: Done!

PR #3510, opened 2023-03-07
Studio Tagger

### LINKED ISSUE #3460: [Feature] Studio Tagger with Image Support
Currently in the scene tagger, if you create a new studio, it does not include the image from Stashbox, or the parent studio.
On the Studios page, there is no Tagger view like there is for Scenes and Performers.

I am currently working towards the following features:

**Scene Tagger - Create Studio Modal**
- [x] The image from Stashbox should be displayed and saved on studio create, as well as the parent studio information.
- [x] A checkbox that defaults to checked, allows the option of also creating the parent studio if the parent studio does not already exist
- [x] If the parent studio matches an existing studio by name, instead the option will be to add the Stash ID and update the image/data

**Studio Tagger**
This is a new view which has the exact same feature set as the current Performer Tagger.
- [x] Provide a tagger view with a search box populated with the studio name or Stash ID
- [x] On an exact match, pop up the tagger create studio modal from above
- [x] Batch Add Studio button with exact name matching required
   - [x] Add option to also create parent studio
- [x] Batch Update Studios button with the same functionality as the batch performer update
   - [x] Add option to also create/update parent studio

Flashy78:
Resolves #3460 

Adds all of the options from #3460 with the tagger code copied over from the Performer tagger.
Since the Performer backend was refactored recently, I ended up refactoring the Studio backend to match. I found it was easier to copy how Performer was done rather than try to copy the tagger frontend and squish it into the existing backend.

Tested identify, import/export, scene tagger, studio tagger, studio tagger batch functions.. all with or without accompanying parent studio.


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/disintegration/imaging v1.6.0
 	github.com/fvbommel/sortorder v1.0.2
 	github.com/go-chi/chi v4.0.2+incompatible
+	github.com/gofrs/uuid v4.4.0+incompatible
```
Flashy78, line 14: This allows for passing in a search query that can be a name or a stash id, and parse the text to find out which one.

internal/desktop/systray_nonlinux.go
```diff
@@ -54,8 +57,9 @@ func systrayInitialize(shutdownHandler ShutdownHandler, faviconProvider FaviconP
 	c := config.GetInstance()
 	if !c.IsNewSystem() {
 		menuItems = c.GetMenuItems()
+		caser := cases.Title(language.English)
```
Flashy78, line 60: This was a linter failing.

graphql/schema/types/scraper.graphql
```diff
@@ -208,16 +223,18 @@ type StashBoxFingerprint {
...
-  "Fields to exclude when executing the performer tagging"
+  "Fields to exclude when executing the tagging"
   exclude_fields: [String!]
-  "Refresh performers already tagged by StashBox if true. Only tag performers with no StashBox tagging if false"
+  "Refresh items already tagged by StashBox if true. Only tag items with no StashBox tagging if false"
   refresh: Boolean!
-  "If set, only tag these performer ids"
-  performer_ids: [ID!]
-  "If set, only tag these performer names"
-  performer_names: [String!]
```
WithoutPants, line 222: We should not remove fields from inputs as this will break existing clients. We should mark these fields as deprecated and for now treat them as `ids` and `names` inputs.

WithoutPants (changes requested):
Code review is still in progress.

graphql/schema/types/scraper.graphql
```diff
@@ -48,11 +48,21 @@ type ScrapedStudio {
...
 }
 
+type ScrapedParentStudio {
+    stored_id: ID
+    name: String!
+    url: String
+    image: String
+
+    remote_site_id: String
+}
```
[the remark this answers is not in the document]
Flashy78: This was my initial attempt before getting the parent value in ScrapedStudio to also be a ScrapedStudio.
Will remove.

graphql/schema/types/scraper.graphql
Flashy78: Understood, will put them back in.

internal/api/resolver_model_studio.go
```diff
@@ -2,30 +2,14 @@ package api
...
 	"github.com/stashapp/stash/pkg/image"
 	"github.com/stashapp/stash/pkg/models"
 	"github.com/stashapp/stash/pkg/performer"
 )
 
-func (r *studioResolver) Name(ctx context.Context, obj *models.Studio) (string, error) {
-	if obj.Name.Valid {
-		return obj.Name.String, nil
-	}
-	panic("null name") // TODO make name required
```
Flashy78, line 19: Name is required in the UI, but I don't think it's enforced anywhere. I will make a note to address that later.

internal/api/resolver_mutation_studio.go
```diff
@@ -28,190 +24,284 @@ func (r *mutationResolver) getStudio(ctx context.Context, id int) (ret *models.S
...
+	r.hookExecutor.ExecutePostHooks(ctx, *studioID, plugin.StudioCreatePost, input, nil)
+
+	return newStudio, nil
+}
+
+func studioFromStudioCreateInput(ctx context.Context, input StudioCreateInput) (*models.Studio, error) {
 	// Populate a new studio from the input
-	currentTime := time.Now()
 	newStudio := models.Studio{
-		Checksum:  checksum,
```
Flashy78, line 48: I moved checksum and the date fields into model_studio.go NewStudio and NewStudioPartial to avoid duplicating it everywhere, but looking now that doesn't stop anyone from just making a new object manually without using those methods and missing out on the fields. I will review this further.

internal/api/resolver_mutation_studio.go
```diff
@@ -28,190 +24,284 @@ func (r *mutationResolver) getStudio(ctx context.Context, id int) (ret *models.S
...
-	}
-
-	if input.Details != nil {
-		newStudio.Details = sql.NullString{String: *input.Details, Valid: true}
+		rating := models.Rating5To100(*input.Rating)
+		newStudio.Rating = &rating
 	}
 	if input.IgnoreAutoTag != nil {
 		newStudio.IgnoreAutoTag = *input.IgnoreAutoTag
 	}
```
Flashy78, line 109: Image is now downloaded, but the saving to the db has been moved to pkg/sqlite/studio.go to avoid that duplicated block of code.
Same with saving the alias and stashid code blocks.

internal/api/resolver_mutation_studio.go
```diff
@@ -28,190 +24,284 @@ func (r *mutationResolver) getStudio(ctx context.Context, id int) (ret *models.S
...
-			if err := qb.UpdateStashIDs(ctx, studioID, stashIDJoins); err != nil {
-				return err
-			}
+	return updatedStudio, nil
+}
+
+// This is slightly different to studioPartialFromStudioCreateInput in that Name is handled differently
+// and ImageIncluded is not hardcoded to true
+func studioPartialFromStudioUpdateInput(ctx context.Context, input StudioUpdateInput, id *string, translator changesetTranslator) (*models.StudioPartial, error) {
+	// Populate studio from the input
```
Flashy78, line 256: Probably a better way to avoid duplicating 98% of this code block.

internal/identify/studio.go
```diff
@@ -2,51 +2,128 @@ package identify
...
 	}
 
-	if endpoint != "" && studio.RemoteSiteID != nil {
-		if err := w.UpdateStashIDs(ctx, created.ID, []models.StashID{
+	return studioID, nil
+}
+
+// Duplicated in task_stash_box_tag.go
+func studioFromScrapedStudio(ctx context.Context, input *models.ScrapedStudio, endpoint string) (*models.Studio, error) {
+	// Populate a new studio from the input
```
Flashy78, line 50: I wasn't able to find a good way to make this method and the next reusable in both places, hence the duplication.

internal/identify/studio_test.go
```diff
@@ -87,18 +105,6 @@ func Test_createMissingStudio(t *testing.T) {
...
-			args{
-				invalidEndpoint,
-				&models.ScrapedStudio{
-					Name:         validName,
-					RemoteSiteID: &remoteSiteID,
-				},
-			},
-			nil,
-			true,
-		},
```
Flashy78, line 101: With the removal of the UpdateStashIDs method, this test isn't really applicable.

internal/manager/manager_tasks.go
```diff
@@ -321,28 +323,38 @@ func (s *Manager) StashBoxBatchPerformerTag(ctx context.Context, input StashBoxB
...
+							if (input.Refresh && len(performer.StashIDs.List()) > 0) ||
+								(!input.Refresh && len(performer.StashIDs.List()) == 0) {
+								tasks = append(tasks, StashBoxBatchTagTask{
+									performer:       performer,
+									refresh:         input.Refresh,
+									box:             box,
+									excluded_fields: input.ExcludeFields,
+									task_type:       Performer,
+								})
+							}
```
Flashy78, line 357: Small refactor to load the StashIDs here instead of in the individual tasks. This way instead of adding a task for every single item and then deciding in the task if the user wanted to refresh existing or search for untagged, we are filtering here and only creating tasks for those items.

internal/manager/manager_tasks.go
```diff
@@ -418,3 +440,132 @@ func (s *Manager) StashBoxBatchPerformerTag(ctx context.Context, input StashBoxB
 
 	return s.JobManager.Add(ctx, "Batch stash-box performer tag...", j)
 }
+
+func (s *Manager) StashBoxBatchStudioTag(ctx context.Context, input StashBoxBatchTagInput) int {
+	j := job.MakeJobExec(func(ctx context.Context, progress *job.Progress) {
```
Flashy78, line 445: Almost a straight up duplication of the above method. I would like to refactor them in a future update.

internal/manager/task_stash_box_tag.go
```diff
@@ -51,7 +78,14 @@ func (t *StashBoxPerformerTagTask) stashBoxPerformerTag(ctx context.Context) {
...
 		var performerID string
 		txnErr := txn.WithReadTxn(ctx, instance.Repository, func(ctx context.Context) error {
-			stashids, _ := instance.Repository.Performer.GetStashIDs(ctx, t.performer.ID)
+			if !t.performer.StashIDs.Loaded() {
+				err = t.performer.LoadStashIDs(ctx, instance.Repository.Performer)
+				if err != nil {
+					return err
+				}
+			}
+			stashids := t.performer.StashIDs.List()
```
Flashy78, line 87: The StashIDs are now passed in to the task, but check to load them just in case.

internal/manager/task_stash_box_tag.go
```diff
@@ -135,6 +169,7 @@ func (t *StashBoxPerformerTagTask) stashBoxPerformerTag(ctx context.Context) {
 			if performer.Measurements != nil && !excluded["measurements"] {
 				partial.Measurements = models.NewOptionalString(*performer.Measurements)
 			}
+			//TODO: This seems incorrect, is there a reason it's opposite?
 			if excluded["name"] && performer.Name != nil {
 				partial.Name = models.NewOptionalString(*performer.Name)
 			}
```
Flashy78, line 175: This seems deliberate to have opposite logic for the excluded check, but I couldn't figure out what.
I did not replicate it for studio.

internal/manager/task_stash_box_tag.go
```diff
@@ -258,6 +293,260 @@ func (t *StashBoxPerformerTagTask) stashBoxPerformerTag(ctx context.Context) {
...
+
+			//TODO: This wasn't previously part of batch performer updates, but it probably should be for both perfomer and studio?
+			/*
+				if runParentCreateHook {
+					r.hookExecutor.ExecutePostHooks(ctx, *updatedStudio.ParentID, plugin.StudioCreatePost, input, nil)
+				} else if runParentUpdateHook {
+					r.hookExecutor.ExecutePostHooks(ctx, *updatedStudio.ParentID, plugin.StudioUpdatePost, input, parentTranslator.getFields())
+				}
+				r.hookExecutor.ExecutePostHooks(ctx, updatedStudio.ID, plugin.StudioUpdatePost, input, translator.getFields())
+			*/
```
Flashy78, line 407: Wondering if the hooks should be run during batch operations.
If so, they will need to refactored to be available to this package somehow.

pkg/models/model_scraped_item.go
```diff
@@ -6,15 +6,28 @@ import (
...
+	URL          *string        `json:"url"`
+	Parent       *ScrapedStudio `json:"parent"`
+	Image        *string        `json:"image"`
+	Images       []string       `json:"images"`
+	RemoteSiteID *string        `json:"remote_site_id"`
 }
 
 func (ScrapedStudio) IsScrapedContent() {}
 
+type ScrapedParentStudio struct {
```
Flashy78, line 20: Will remove.

pkg/models/model_studio.go
```diff
@@ -56,3 +101,12 @@ func (s *Studios) Append(o interface{}) {
...
 }
+
+// TODO: Seems like a good candidate for generics
+type StudioDBInput struct {
+	StudioCreate *Studio
+	StudioUpdate *StudioPartial
+
+	ParentCreate *Studio
+	ParentUpdate *StudioPartial
+}
```
Flashy78, line 112: This is definitely a different architecture than currently exists, but for calls to Create I needed a StudioCreate, and maybe a ParentCreate or ParentUpdate. For calls to UpdatePartial, I needed a StudioUpdate and maybe a ParentCreate or ParentUpdate.
I tried a few different things like passing them in as params, or stuffing the parent objects into the studio objects, and passing flags for what to create/update... but this seemed to work the best.

pkg/models/studio.go
```diff
@@ -56,19 +57,17 @@ type StudioReader interface {
...
-	GetAliases(ctx context.Context, studioID int) ([]string, error)
 }
 
 type StudioWriter interface {
-	Create(ctx context.Context, newStudio Studio) (*Studio, error)
-	Update(ctx context.Context, updatedStudio StudioPartial) (*Studio, error)
-	UpdateFull(ctx context.Context, updatedStudio Studio) (*Studio, error)
+	Create(ctx context.Context, input StudioDBInput) (*int, error)
+	UpdatePartial(ctx context.Context, input StudioDBInput) (*Studio, error)
+	Update(ctx context.Context, updatedStudio *Studio) error
```
Flashy78, line 67: Updated methods to match Performer, except Create returns the created id instead of accepting the studio as a pointer and updating in place. I've been hurt too many times by that pattern of not realizing my object is getting updated inside a method without me knowing, so I try to avoid it.

pkg/models/timestamp_test.go
```diff
@@ -37,7 +37,7 @@ func TestTimestamp(t *testing.T) {
 		{"rfc3339", "2021-11-04T01:02:03Z", "2021-11-04T01:02:03Z"},
 		{"date", "2021-04-05", "2021-04-05T00:00:00Z"},
 		{"datetime", "2021-04-05 14:45:36", "2021-04-05T14:45:36Z"},
-		{"datetime-tz", "2021-04-05 14:45:36 PDT", "2021-04-05T14:45:36Z"},
+		{"datetime-tz", "2021-04-05 14:45:36 PDT", "2021-04-05T14:45:36-07:00"},
```
Flashy78, line 40: This test was strangely failing for me, but I assume must have worked for previous PRs. At least my change also seems to pass the Github build.

pkg/sqlite/repository.go
```diff
@@ -516,21 +516,6 @@ func (r *stashIDRepository) get(ctx context.Context, id int) ([]models.StashID,
...
-	query := fmt.Sprintf("INSERT INTO %s (%s, endpoint, stash_id) VALUES (?, ?, ?)", r.tableName, r.idColumn)
-	for _, stashID := range newIDs {
-		_, err := r.tx.Exec(ctx, query, id, stashID.Endpoint, stashID.StashID)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
```
Flashy78, line 533: I guess Studio must have been the only one using this method as the linter wanted me to remove it.

pkg/studio/export_test.go
```diff
@@ -181,26 +172,13 @@ func TestToJSON(t *testing.T) {
...
-	mockStudioReader.On("GetAliases", ctx, studioID).Return([]string{"alias"}, nil).Once()
-	mockStudioReader.On("GetAliases", ctx, noImageID).Return(nil, nil).Once()
-	mockStudioReader.On("GetAliases", ctx, errImageID).Return(nil, nil).Once()
-	mockStudioReader.On("GetAliases", ctx, missingParentStudioID).Return(nil, nil).Once()
-	mockStudioReader.On("GetAliases", ctx, errAliasID).Return(nil, aliasErr).Once()
-
-	mockStudioReader.On("GetStashIDs", ctx, studioID).Return(stashIDs, nil).Once()
-	mockStudioReader.On("GetStashIDs", ctx, noImageID).Return(nil, nil).Once()
-	mockStudioReader.On("GetStashIDs", ctx, missingParentStudioID).Return(stashIDs, nil).Once()
-
```
Flashy78, line 203: No need to test these as they were removed.

pkg/studio/import.go
```diff
@@ -113,16 +98,6 @@ func (i *Importer) PostImport(ctx context.Context, id int) error {
...
-	if len(i.Input.StashIDs) > 0 {
-		if err := i.ReaderWriter.UpdateStashIDs(ctx, id, i.Input.StashIDs); err != nil {
-			return fmt.Errorf("error setting stash id: %v", err)
-		}
-	}
-
-	if err := i.ReaderWriter.UpdateAliases(ctx, id, i.Input.Aliases); err != nil {
-		return fmt.Errorf("error setting tag aliases: %v", err)
-	}
-
```
Flashy78, line 125: These were moved to pkg/sqlite/studio.go, but creating the image was left here.

ui/v2.5/src/components/Dialogs/IdentifyDialog/constants.ts
```diff
@@ -31,6 +31,8 @@ export const multiValueSceneFields: SceneField[] = [
 export function sceneFieldMessageID(field: SceneField) {
   if (field === "code") {
     return "scene_code";
+  } else if (field === "studio") {
+    return "studio_and_parent";
```
Flashy78, line 35: Defaulted to also creating parents when the user has chosen to create studios.

ui/v2.5/src/components/Tagger/performers/PerformerTagger.tsx
```diff
@@ -418,7 +418,7 @@ const PerformerTaggerList: React.FC<IPerformerTaggerListProps> = ({
             type="radio"
             name="performer-query"
             label={<FormattedMessage id="performer_tagger.current_page" />}
-            defaultChecked
+            defaultChecked={!queryAll}
```
Flashy78, line 421: Fix so the modal remembers what has been checked.

ui/v2.5/src/components/Tagger/queries.ts
```diff
@@ -204,6 +208,54 @@ export const useUpdateStudioStashID = () => {
...
+    updateStudio({
+      variables: {
+        input,
+      },
+      update: (store, updatedStudio) => {
+        if (!updatedStudio.data?.studioUpdate) return;
+
+        if (updatedStudio.data?.studioUpdate?.parent_studio) {
+          const ac = getClient();
+          ac.refetchQueries({ include: studioMutationImpactedQueries });
```
Flashy78, line 227: The goal of this was that if a parent was created/updated, since the data for that studio was not being passed back to update the cache directly, the whole thing should be refreshed.
Perhaps there is a better way of doing this.

ui/v2.5/src/components/Tagger/scenes/StudioModal.tsx
```diff
@@ -1,80 +1,236 @@
...
+      }
+    }
+
     handleStudioCreate(studioData);
   }
 
   const renderField = (
     id: string,
     text: string | null | undefined,
-    truncate: boolean = true
```
Flashy78, line 59: I think I should put this back in.

ui/v2.5/src/components/Tagger/studios/StudioTagger.tsx
```diff
@@ -0,0 +1,750 @@
...
+
+    if (event.type !== GQL.JobStatusUpdateType.Remove) {
+      setBatchJob(event.job);
+    } else {
+      setBatchJob(undefined);
+      setBatchJobID(undefined);
+
+      // Once the studio batch is complete, refresh all local studio data
+      const ac = getClient();
+      ac.refetchQueries({ include: studioMutationImpactedQueries });
```
Flashy78, line 586: Again the goal is to refresh the local studio cache, not sure if this is the best way to do it.

pkg/models/timestamp_test.go
Flashy78: Ok, it fails on Github, will revert.

internal/api/resolver_mutation_studio.go
Flashy78: Confirmed CreatedAt/UpdatedAt are enforced at the sqlite level.

WithoutPants:
The sqlite refactoring is now merged, so this can be updated when you're ready.

pkg/sqlite/studio.go
```diff
@@ -183,13 +291,16 @@ func (qb *StudioStore) Destroy(ctx context.Context, id int) error {
 	return qb.destroyExisting(ctx, []int{id})
 }
 
-// returns nil, nil if not found
+// returns nil, sql.ErrNoRows if not found
 func (qb *StudioStore) Find(ctx context.Context, id int) (*models.Studio, error) {
```
DingDongSoLong4, line 295: The reason why I intentionally split the Find and find methods in my refactoring is to prevent sql.ErrNoRows errors from escaping out from the sqlite package. External code (such as resolvers) use the Find method, meaning that a graphql query for a missing studio returns null instead of an error. Internal code can use either method depending on whether the id is valid or not should be considered an error.

DingDongSoLong4 (commented):
Ultimately it's not up to me to decide how these things are structured, but the way I see the `sqlite` package is that it should be a simple interface around the database, generating SQL to insert/update/delete records as required by the rest of the system. In other words, logic for things like ensuring parent studios, setting updated at timestamps, etc is a better fit for `pkg/studio` rather than `pkg/sqlite`.

The motivation for my sqlite refactoring, and for some of the other refactorings I have lined up, is that I want to eventually try and add support for other database engines (specifically, postgres). Having all entity create/update/delete actions first go through validation in their respective `pkg/{scene,image,studio,etc}` packages (ie the `Service` structs for scenes, images and galleries) rather than using the methods in `pkg/sqlite` directly would make it much easier for different database backends to exist and would also ensure invalid data never actually gets stored.

However that is all just my opinion, so until a clear structure is worked out nothing really has to change here.

pkg/sqlite/studio.go
Flashy78: Understood, sorry I didn't catch that reasoning from the code. I'll add it back in.

Flashy78:
> Ultimately it's not up to me to decide how these things are structured, but the way I see the `sqlite` package is that it should be a simple interface around the database, generating SQL to insert/update/delete records as required by the rest of the system. In other words, logic for things like ensuring parent studios, setting updated at timestamps, etc is a better fit for `pkg/studio` rather than `pkg/sqlite`.
> 
> The motivation for my sqlite refactoring, and for some of the other refactorings I have lined up, is that I want to eventually try and add support for other database engines (specifically, postgres). Having all entity create/update/delete actions first go through validation in their respective `pkg/{scene,image,studio,etc}` packages (ie the `Service` structs for scenes, images and galleries) rather than using the methods in `pkg/sqlite` directly would make it much easier for different database backends to exist and would also ensure invalid data never actually gets stored.
> 
> However that is all just my opinion, so until a clear structure is worked out nothing really has to change here.

That does make sense, but as you say I think there needs to be a slight rearchitecture of how the sqlite package is used in order to support that design.

My reasoning is because there is no clear structure, timestamps were set in some places and not in others. If someone used the models.NewPartialObject() method they got timestamps, if they just instantiated a new object on their own then they didn't. So enforcing those at the lowest level ensured they were always set correctly and the developer did not have to always remember to do it. It's even easier if those fields are set directly in the database.

There are so many places were logic is just duplicated everywhere, like saving the image for an object. Where instead if the logic is at the lowest point in the chain, developers don't have to remember to copy and paste that block of code every time.

I think if there was a way to stop developers from using the sqlite package directly, then all that logic could go in the layer above it, but it doesn't seem like things are architected that way currently.

DingDongSoLong4:
Agreed. Like I said, changing anything now isn't that necessary, as long as everything works. The reorganising can come later, I'll get to it eventually.

internal/api/resolver_query_find_studio.go
```diff
@@ -8,6 +8,10 @@ import (
 )
 
 func (r *queryResolver) FindStudio(ctx context.Context, id string) (ret *models.Studio, err error) {
+	if id == "" {
+		return nil, nil
+	}
```
WithoutPants, line 13: This is not consistent with the behaviour of the other resolvers. An empty id should be an error condition.

graphql/schema/types/studio.graphql
```diff
@@ -29,6 +29,7 @@ input StudioCreateInput {
   name: String!
   url: String
   parent_id: ID
+  parent: StudioCreateInput
```
WithoutPants, line 32: Looking at the way this is processed in the resolver logic, I don't think it's a good design to include this. The behaviour isn't clear from the outside and the behaviour is based on whether an existing studio exists or not. The client should be responsible for creating/updating the parent studio. The only benefit I see in combining these operations is that you end up with a single transaction to roll back if needed, but I don't think it's worth it.

pkg/models/model_studio.go
WithoutPants: See my comment in the graphql changes. The operation needs to be split up. Even if we disregard my comments there, I don't think this is a good solution to what you're trying to do. The `sqlite` code shouldn't be responsible for updating related objects; that's the job of the calling code.

WithoutPants (changes requested):
This is not a complete review as there's a significant design issue that I think needs to be addressed before I continue - regarding the `parent` field of the studio create/update mutation. Happy to discuss this further as needed.

internal/manager/task_stash_box_tag.go
```diff
@@ -183,7 +210,261 @@ func (t *StashBoxPerformerTagTask) stashBoxPerformerTag(ctx context.Context) {
...
+	}
+
+	partial.StashIDs = &models.UpdateStashIDs{
+		StashIDs: []models.StashID{
+			{
+				Endpoint: endpoint,
+				StashID:  *input.RemoteSiteID,
+			},
+		},
+		Mode: models.RelationshipUpdateModeSet,
```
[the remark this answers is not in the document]
Flashy78: This looks a bit ugly now because I first have to load the existing studio loaded in order to load the existing ids.

It feels like there should be a way to just add new/updated ids and then the sqlite UpdatePartial method knows it's doing a partial update and not to replace all ids.

internal/api/resolver_model_studio.go
```diff
@@ -99,7 +99,14 @@ func (r *studioResolver) ParentStudio(ctx context.Context, obj *models.Studio) (
...
 
-	return loaders.From(ctx).StudioByID.Load(*obj.ParentID)
+	if err := r.withReadTxn(ctx, func(ctx context.Context) error {
+		ret, err = r.repository.Studio.Find(ctx, *obj.ParentID)
+		return err
+	}); err != nil {
+		return nil, err
+	}
+
+	return ret, nil
```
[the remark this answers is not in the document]
Flashy78: I don't remember, reverted.

pkg/models/studio.go
Flashy78: Reverted my changed for Create.
Should I revert UpdatePartial as well? What's the use case for passing in the id separate instead of as part of the partial model being updated?

internal/manager/task_stash_box_tag.go
```diff
@@ -183,7 +210,261 @@ func (t *StashBoxPerformerTagTask) stashBoxPerformerTag(ctx context.Context) {
...
+	if input.Parent != nil && !excluded["parent"] {
+		if input.Parent.StoredID != nil {
+			parentID, _ := strconv.Atoi(*input.Parent.StoredID)
+			if parentID > 0 {
+				// This is to be set directly as we know it has a value and the translator won't have the field
+				partial.ParentID = models.NewOptionalInt(parentID)
+			}
+		}
+	} else {
+		partial.ParentID = models.NewOptionalIntPtr(nil)
```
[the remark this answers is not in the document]
Flashy78: This code is now in `model_scraped_item.go`.
I think it's kind of just like the URL or image field. If you've said you want to batch update all of your studios with Stash-box data and haven't excluded that field, then it gets set to whatever is scraped.

graphql/schema/types/studio.graphql
Flashy78: I've moved all parent studio logic out of the `sqlite` package.
I think I was trying to treat it logically as a single operation (not necessarily a single transaction). I wanted to avoid duplicating all of the parent matching/creating/updating logic in all of the multiple places where studios are created (identify/import/batch update/tagger view).
All of that code duplication can lead to bugs in the future when there's a new way to create a studio but all of the parent logic is not copied exactly. Like when disambiguation was added, it was added to all of the various places you can create a performer, but missed in identify.

I think this comment means you're suggesting to also write this parent logic in the frontend, so that the frontend can make an api call to create/update a parent studio, then attach it to the main studio and make another api call to create/update studio?

WithoutPants:
I've made a bunch of changes to the backend code.

I've removed the `parent` field from the create and update inputs. The front end code needs to be updated to separately create/update parent studios. Currently, the tagger hangs when creating a studio with parent info, because the server is returning an error and the error isn't being handled correctly. I'll investigate this myself tomorrow if you don't get to it in the meantime.

I've taken out the image data from the studio object and partial structs. It was a vestigial field on the studio struct, only used during creation. It's also inconsistent with the other objects/partials. It does result in more code on the calling side to set the image, but this can be mitigated in future by combining the create/update object and image code into a service-level function. I could have kept the image in the partial, but this would be inconsistent with the create behaviour.

I reverted some undocumented design decisions such as moving the timestamp setting code into the sqlite package. In principle, the sqlite package should be concerned only with querying and modifying data, and not with the logic. We want fine control of these timestamps outside the data layer, and as such should not have been moved here. The alias uniqueness checking should probably not be in sqlite either, but for now it's fine to stay there. In future it should be moved into the business layer. In future we should unify the create/update functions into a single function in the studio package, which would include this validation.

I also reverted a change that prevented aliases from being used in studio autotag. This looked like it was copied from the performer code, but the reasons for disabling aliases in the performer code do not apply to studios.

In future, I'd prefer if you could separate large refactors from new features. It makes reviewing far more difficult. In future, please also highlight low-level design decisions like the items mentioned above. These have far-reaching impacts on the design and behaviour, and detecting them with static code reviews are difficult.

Flashy78:
> I've made a bunch of changes to the backend code.
> 
> In future, I'd prefer if you could separate large refactors from new features. It makes reviewing far more difficult. In future, please also highlight low-level design decisions like the items mentioned above. These have far-reaching impacts on the design and behaviour, and detecting them with static code reviews are difficult.

Thanks for taking the time to get this where it needs to be.

I will avoid new features and stick to bug fixes to avoid creating more work for you.

WithoutPants:
> I will avoid new features and stick to bug fixes to avoid creating more work for you.

This wasn't intended to discourage you from making new features! The overall change here is all good, there was just some issues around the refactoring that was done at the same time.

WithoutPants:
I've done some refactoring of the UI code to handle creating/updating parent studios. I'd appreciate if you could give it a test run to ensure I haven't broken anything. It's looking about complete on my end.

Flashy78:
I was too overconfident in my ability to pick up and understand go. I will lurk longer to learn more about how the codebase is structured/designed (ie sqlite should only read/write with zero changes to the objects). I would rather write features that you can bug check/suggest changes, rather than having to spend days rewriting. In that case it would be more efficient just for you to write the feature whenever it appears on your list.

If I tag a studio, where the parent studio exists but has no stashId, it tries to create the parent instead of updating.
```
{
    "errors": [
        {
            "message": "inserting into studios: executing `INSERT INTO `studios`....: UNIQUE constraint failed: studios.name",
            "path": [
                "studioCreate"
            ]
        }
    ],
    "data": {
        "studioCreate": null
    }
}
```

Some cache seems to be getting messed up, which may be the cause above. I tried to reproduce and it wanted to create the parent. Refreshed the page and it correctly matched the parent and assigned the stashid.

Deleted stashId from Studio, deleted parent studio, navigate back to tagger, studio modal thinks the parent still exists and then 
```
{
    "errors": [
        {
            "message": "studio with id 30 not found",
            "path": [
                "studioUpdate"
            ]
        }
    ],
    "data": {
        "studioUpdate": null
    }
}
```

PR #3513, opened 2023-03-07
Fix VTT thumbnails

DingDongSoLong4:
This is fix for a rare bug that I just happened to come across by chance - scenes with oshashes that consist of only numbers (do a scene filter with regex `^\d*$`) don't display sprites in the scrubber. This is because the server parses the checksum as a scene ID, and thus fails to find the scene, returning 404. 

I couldn't find anywhere where the scene ID is actually used instead of the hash to get the `_thumbs` or `_sprite` file, so I've just removed that logic which greatly simplifies things.

As an afterthought, I noticed that the API key was not a parameter of `NewSceneURLBuilder` but was still used by `GetStreamURL`, which makes it very easy to forget to set the APIKey field manually on the struct before calling `GetStreamURL`. I refactored the API key into a parameter of `GetStreamURL`, and was surprised to find that setting the APIKey field manually had indeed been forgotten in the `SceneStreams` resolver. So this also fixes the `sceneStreams` query results not including the API key.


DogmaDragon:
So that's what caused it. Encountered it once, but wasn't able to reproduce it so never made a report.

WithoutPants:
I think there is some utility in being able to get the thumbs/sprite files from the scene ID. I think it's worth adding a route for `/{sceneId}/thumbs.vtt` and `/{sceneId}/sprite.jpg`. It doesn't resolve the breaking of backward compatibility for clients, but at least there is an alternative that doesn't involve getting the correct path through another API call.

DingDongSoLong4:
Agreed. I've added them as `/vtt/thumbs` and `/vtt/sprite` to be consistent, and done a bit of cleanup as well.

PR #3514, opened 2023-03-08
add divider on more details pages

Teda1:
This pull request adds dividers to the performers, studio, and tags details pages making the left tab collapsable to get a fuller view of the content. This change includes CSS to ensure the left and right tabs don't share the same scroll wheel, so the details stay in view as you scroll through content.


WithoutPants:
This change has made the left side of the affected pages larger on display width > 1200px.

image

vs

image

I think styling the `col-x-x` classes is probably not a great idea.

Teda1:
Thanks for taking a look. The size of the left side should now auto-adjust the same way it did previously. 

You're right styling bootstrap classes probably isn't reliable. I've also updated the CSS so that styling the `col-x-x` classes are no longer necessary.

Issue #2692, opened 2022-06-24
[Feature] Filter with tags using include & exclude


JackDawson94:
**Is your feature request related to a problem? Please describe.**
Was discussed in Discord and seems many people find filtering with tags a bit frustrating right now. Tags are one of the most powerful features, but can only use one filter operator.
It's either **include** or **exclude** but can't combine the two. 

**Describe the solution you'd like**
I'd like to be able to filter tags (probably other types as well ? but mostly tags) with something like:
**include A & exclude B**

xx790:
This is super frustrating, compared to booru/hydrus search queries.

I see no reason why available filter options are removed from the list when one instance is applied. They should be stackable.

I can manually craft an URL like
```http://localhost:9999/scenes?c=("type":"tags","value":("items":[("id":"6","label":"foo")],"depth":0),"modifier":"INCLUDES")&c=("type":"tags","value":("items":[("id":"8","label":"bar")],"depth":0),"modifier":"EXCLUDES")&perPage=1000&sortby=random_96764349```
but it doesn't seem to work - only one of `"type":"tags"` entries is applied while there should be no issue to apply them sequentially.

WithoutPants:
This is due to the limitations of the existing graphql API. This can be done, but it requires refactoring of said API, which impacts both the backend code and the UI. It's not a single thing to fix.

xx790:
Looks like someone outsmarted themselves

PR #3515, opened 2023-03-08
UI filter builder

WithoutPants:
Replaces the add filter dialog with an edit filter dialog. This allows adding/editing filter criterion all at once:

image

- text fields have a confirm button to modify the value
- current filter is shown at the bottom, with an extra option to clear all criteria if there is more than two criteria
- criteria with current values have an x next to their entries which allow deleting the criterion
- criteria with options are presented as a list of checkboxes

image
image

Mobile view presents the criteria list only, then shifts to the criterion editor when a criterion is selected, with a back button to return to the criteria list.

image

Also adds the criteria count as a pill to the filter button, and adds the Clear All button to the main view as well.


Teda1:
I encountered this bug that wouldn't allow me to specify another filter. Anytime I used the drop down it would send me back to the performers filter.
https://user-images.githubusercontent.com/72030708/223875732-e8604af8-5a75-4d3a-80f8-137a830dbe34.mp4

I also noticed that the submit button is inconsistently placed on a few of the filter menus. It's not clear if this was intentional:
Screenshot 2023-03-08 170251
Screenshot 2023-03-08 170316

WithoutPants:
Criterion selection bug should be fixed. Thanks for the report.

Yeah, there's a style discrepancy there currently. I wanted to add the confirm button to the input control, but this breaks down when you have a between modifier and need two fields. 

I can move it out into a separate button to be consistent, but I want to confirm that this is the best approach from a UX point of view.

The alternatives are:
- update on keystroke - this is kind of a waste of rendering, especially when rapidly typing into a field. We could debounce it slightly I guess
- update on blur - the problem with this approach is that it's not intuitive to the user how to _confirm_ a value. They will type in a value, not see any confirm button, and wonder why the filter hasn't updated.
- reposition the confirm button somewhere else?

I chose this solution based on playing around with the ebay filter UI.

Teda1:
Sorry, I forgot to get back to this. That makes sense. Yeah, I was leaning toward having the button on the input field, but I see that challenge with that when dealing with the between modifiers. I think consistent placement would be best, so the user always knows where to find the button without much thought. So perhaps the button could just stay below the input fields.

I also confirm that your latest push address the issue I mentioned earlier.

Teda1:
I wonder if a design like this might make more sense. The current design leaves a lot of unused space:
Screenshot 2023-03-10 170902

WithoutPants:
I agree, it looks better imo, and is more consistent with the mobile view:

image

The applicable element is scrolled into view when clicking on a criterion from outside the filter dialog. This behaviour is not triggered when clicking on the pill from inside the dialog, since I couldn't get the scrolling function to consistently scroll to the correct position.

xx790:
Will this UI work well when #2692 gets addressed? (We need multiple filters on the same property like tags)

Teda1:
> I agree, it looks better imo, and is more consistent with the mobile view:
> 
> image
> 
> The applicable element is scrolled into view when clicking on a criterion from outside the filter dialog. This behaviour is not triggered when clicking on the pill from inside the dialog, since I couldn't get the scrolling function to consistently scroll to the correct position.

This looks a lot better and functions as I would have expected.

Teda1:
I do think all the selectors could continue to take the full width as they used to. For example, the modifier selector for play count in your screenshot took up a little under 1/3 of the width where it used to take up the full width.
Screenshot 2023-03-12 143426

WithoutPants:
> Will this UI work well when #2692 gets addressed? (We need multiple filters on the same property like tags)

I don't think it will preclude it from working. The UI design can be iterated on as part of the work on that anyway.

WithoutPants:
Thanks for the feedback.

I've removed the confirm buttons, the value is now updated as typed. Stash ID and hierarchical filters should be fixed, and I've fixed the placeholder text on the timestamp filter.

DingDongSoLong4:
Lovely, looks great. I'm still seeing the hierarchical issue for Studios and Tags though, the fix seems to have only affected Performers and Movies.

WithoutPants:
Thanks, the other criteria should be fixed now.

echo6ix:
I'm having an issue where the country results get cropped
- Custom CSS disabled
- Chrome v111

image

echo6ix:
First, I think this is a big improvement, and adding a badge counter to the filter button is a great idea. My only feedback regarding that badge counter is the following:
1. Badge color. Red is used as a color to denote destructive actions or cancellations throughout Stash. A variation of bright blue might be better for this badge. Off the top of my head, the `--info` color defined in the CSS would work.
2. Badge placement. Placement seems a little awkward to me here. I think the above (and below) example from Github would work a lot better 
image

PR #3527, opened 2023-03-11
Automatic database optimization and performance tweaks

DingDongSoLong4:
This is a fix for an issue that came up in Discord, where some SQL queries are very slow in specific scenarios. The cause is SQLite not using indexes, because the involved tables were very small/empty when the last migration was run (ie since the last `ANALYZE`) but have since greatly increased in size.

The fix is to run `PRAGMA optimize` before closing every database connection, as recommended in the SQLite docs (https://www.sqlite.org/lang_analyze.html). To achieve this with go-sqlite3, a custom database driver wrapper around `sqlite3.SQLiteDriver` was needed.

I've also tweaked the connection pooling configuration for (slightly) better performance, tested with CJ's behemoth of a database. Opening new database connections has a noticeable performance penalty, which is easily seen by removing the `MaxOpenConns` limit. It therefore ends up being more performant to decrease `MaxOpenConns` but make them all stick around as idle connections with `MaxIdleConns`. This also has the benefit of reducing overall memory usage.

### DIFF
--- a/pkg/sqlite/database.go
+++ b/pkg/sqlite/database.go
@@ -2,7 +2,6 @@ package sqlite
 
 import (
 	"context"
-	"database/sql"
 	"embed"
 	"errors"
 	"fmt"
@@ -10,12 +9,10 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/fvbommel/sortorder"
 	"github.com/golang-migrate/migrate/v4"
 	sqlite3mig "github.com/golang-migrate/migrate/v4/database/sqlite3"
 	"github.com/golang-migrate/migrate/v4/source/iofs"
 	"github.com/jmoiron/sqlx"
-	sqlite3 "github.com/mattn/go-sqlite3"
 
 	"github.com/stashapp/stash/pkg/fsutil"
 	"github.com/stashapp/stash/pkg/logger"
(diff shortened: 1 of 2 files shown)


pkg/sqlite/database.go
```diff
@@ -202,9 +192,9 @@ func (db *Database) open(disableForeignKeys bool) (*sqlx.DB, error) {
 	}
 
 	conn, err := sqlx.Open(sqlite3Driver, url)
-	conn.SetMaxOpenConns(25)
-	conn.SetMaxIdleConns(4)
-	conn.SetConnMaxLifetime(30 * time.Second)
+	conn.SetMaxOpenConns(10)
+	conn.SetMaxIdleConns(10)
+	conn.SetConnMaxIdleTime(30 * time.Second)
```
WithoutPants, line 197: Since there's now reasons for choosing these particular values, could you make them constants with an explanation for why each value was chosen?

DingDongSoLong4: Added both constants

Flashy78:
Since I merged this into my local 2 days ago, I've been getting database locked errors here and there, where I haven't seen them before.
My db is about 3gb, on the same ssd as the stash docker container.
I have no tasks running, just manually editing a scene, adding some fields, saving.. updating the image from a url, then getting database locked, although the image downloads because I can then hit save and it successfully saves with the new image.

PR #3536, opened 2023-03-13
Add show sprite button to scene tagger list

### LINKED ISSUE #3534: [Feature] View sprites in scene tagger
The scene tagger is also a very convenient place to add tags to videos without having to click through them one by one, go to edit and add tags.
There was recently a discussion on discord of a user who would hack the scene tagger by using it with the Auto Tag source, and just using it for quickly adding tags.

Quickly being able to see the sprite file generated for the video can help find lots of tags without having to watch the video, while providing more detail than just the preview video.

Add an icon to the scene tagger to show the sprite file in a lightbox, if the file has been generated.

sprites

Flashy78:
Resolves #3534 

Adds an image icon to each scene in the tagger view that displays the sprite file in a lightbox. The icon doesn't not show up if there is no file for the scene.

Suggestions on a better UI than shown in #3534 are welcome.

### DIFF
--- a/ui/v2.5/src/components/Tagger/scenes/TaggerScene.tsx
+++ b/ui/v2.5/src/components/Tagger/scenes/TaggerScene.tsx
@@ -12,7 +12,11 @@ import { TruncatedText } from "src/components/Shared/TruncatedText";
 import { parsePath, prepareQueryString } from "src/components/Tagger/utils";
 import { ScenePreview } from "src/components/Scenes/SceneCard";
 import { TaggerStateContext } from "../context";
-import { faChevronDown, faChevronUp } from "@fortawesome/free-solid-svg-icons";
+import {
+  faChevronDown,
+  faChevronUp,
+  faImage,
+} from "@fortawesome/free-solid-svg-icons";
 import { objectPath, objectTitle } from "src/core/files";
 
 interface ITaggerSceneDetails {
@@ -84,6 +88,7 @@ interface ITaggerScene {
   doSceneQuery?: (queryString: string) => void;
   scrapeSceneFragment?: (scene: GQL.SlimSceneDataFragment) => void;
   loading?: boolean;
+  showLightboxImage: (imagePath: string) => void;
 }
 
 export const TaggerScene: React.FC<PropsWithChildren<ITaggerScene>> = ({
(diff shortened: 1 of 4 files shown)


ui/v2.5/src/components/Tagger/scenes/TaggerScene.tsx
```diff
@@ -186,6 +192,27 @@ export const TaggerScene: React.FC<PropsWithChildren<ITaggerScene>> = ({
     }
   }
 
+  function maybeRenderSpriteIcon() {
+    if (scene.paths.sprite && scene.paths.sprite.length > 0) {
+      // Even if a scene doesn't have a sprite generated, the path will be
+      // http://localhost:9999/scene/_sprite.jpg so check for that case
+      const spriteParts = scene.paths.sprite.split("/");
+      const filename = spriteParts.pop();
+      if (filename?.includes("_sprite") && filename.split("_")[0].length > 0) {
```
WithoutPants, line 201: Can this not be changed to check for `scene.files.length`?

WithoutPants:
The alternative I think would be similar to the duplicate scene checker tool. I'd move the icon to be overlapping the bottom right corner of the thumbnail, and have the sprite image show on hover. Showing the sprite image in a lightbox seems like too much to me, but I'm not totally against it.

DogmaDragon:
> The alternative I think would be similar to the duplicate scene checker tool. I'd move the icon to be overlapping the bottom right corner of the thumbnail, and have the sprite image show on hover. 
Something like this would be my preference. Maybe it's just me not being used to lightbox or me being used to how the dupe checker displays it.

ui/v2.5/src/components/Tagger/scenes/TaggerScene.tsx
Flashy78: Originally was trying to not show it if the sprite wasn't yet generated, but couldn't really figure it out. `scene.files.length` is a better suggestion.

Flashy78:
I moved the icon to overlap the top left of the thumbnail, it was much easier than caclulating where it needs to be if the thumbnail is portrait or landscape, and also because there can often be a useful watermark in the bottom right that is nice not to cover when tagging.

I found being able to zoom into the image very useful sometimes, so the lightbox worked best for me.
I also didn't want to be instrusive to existing users by having something randomly popping up and taking over their screen as they moved their mouse around.

PR #3541, opened 2023-03-14
Fix /stream endpoint serving directory for fileless scenes

### LINKED ISSUE #3526: [Bug Report] Stream link for fileless scene serves home directory overview
**Describe the bug**
When clicking the stream link on a fileless scene you get an overview of all files in the user's home directory

**To Reproduce**
1. (Create new fileless scene)
2. Click on scene -> file info -> stream url
3. Overview of the directories and files in the user's home directory gets returned

**Expected behavior**
Getting an error like 404

**Stash Version: (from Settings -> About):** v0.19.1-56-g9aa7ec57

 - OS: Windows 10
 - Browser Firefox
 - Version 110.0.1

**Additional context**
Clicking on any of the returned directories or files returns a 404 error. This behavior also only affects fileless scenes, so one has to know the id of such a scene, therefore I consider this issue uncritical enough to post it publicly

WithoutPants:
Fixes #3526 

`/stream` endpoint now returns 404 if the scene has no files. The `File Info` tab is also now hidden when the scene has no files.

### DIFF
--- a/internal/api/routes_scene.go
+++ b/internal/api/routes_scene.go
@@ -88,7 +88,14 @@ func (rs sceneRoutes) Routes() chi.Router {
 // region Handlers
 
 func (rs sceneRoutes) StreamDirect(w http.ResponseWriter, r *http.Request) {
+
 	scene := r.Context().Value(sceneKey).(*models.Scene)
+	// #3526 - return 404 if the scene does not have any files
+	if scene.Path == "" {
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
 	sceneHash := scene.GetHash(config.GetInstance().GetVideoFileNamingAlgorithm())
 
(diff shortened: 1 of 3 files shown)


DingDongSoLong4:
As it is now, some of the info on the File Info tab isn't explicitly related to the files, like the scene URL (Downloaded From), Stash IDs and the activity tracking info. They will now be hidden when the scene has no files. Activity tracking doesn't really make sense without a file so that's fine, but the scene URL and Stash ID might be set on a fileless scene.

I don't think any of these should really be in a "File Info" tab anyway, so maybe the tab should be renamed to just "Info" or "Scene Info"? It would then make sense to move the Created At and Updated At lines to it as discussed in #2188.

WithoutPants:
I think that non-file information should be moved out of the current File Info tab, and into the details tab, and that we could potentially split the details tab. For now, I'll revert the file info tab change and hide the stream link instead.

PR #3548, opened 2023-03-16
Fix batch performer tagging with multiple endpoints

### LINKED ISSUE #3547: [Bug Report] Bulk Update of Performers with goal of having multiple stashbox stashids
**Describe the bug**
If you want to use 2 stashboxs, despite manual scraping against each stashbox working with a performer, bulk scraping doesn't work as expected.  Either it ignores the stashbox it doesn't have an entry for yet (if you have 2 configured stashbox endpoints, but performer only has 1 stashbox stashid), OR it replaces the existing stashbox id with the new stashbox IF you remove one of the endpoints from your config (so it has a stale stashid, no endpoint to match, and no stashid for the current valid endpoint)

**To Reproduce**
Performer tab, Tagger, bulk updates.

If you select "Untagged", if your performers have tags for Stashbox A, but not for Stashbox B, despite B being the current set endpoint in the config, it shows Zero (ie all performers have _a_ tag, from Stashbox A.)  Updates do not work, it won't add Stashbox B stashids to the performers.

if you remove A from your current stash config entirely:
If you select "Untagged", if your performers have tags for Stashbox A, but not for Stashbox B, despite B being the current set endpoint in the config, it shows Zero.  Updates _do_ work, but it removes the stashID from Stashbox A, and replaces it with StashBox B stashid.

**Expected behavior**
Ideally, Untagged should show a count of performers missing an id _from_ the current selected Endpoint (in the config on that page)
Selecting and running Untagged should add Stashbox B stashids (assuming that's selected), and leave Stashbox A ids intact and functional.

v0.19.1-62-ge2b52a4b

WithoutPants:
Resolves #3547 

Fixes the backend code to not overwrite unrelated stash ids. Also has a minor change to not fail the operation if the performer image can't be read from stash-box - displays a warning instead.

Refactored the batch performer dialogs and corrected the performer counts.

### DIFF
--- a/internal/manager/task_stash_box_tag.go
+++ b/internal/manager/task_stash_box_tag.go
@@ -50,17 +50,10 @@ func (t *StashBoxPerformerTagTask) stashBoxPerformerTag(ctx context.Context) {
 
 	if t.refresh {
 		var performerID string
-		txnErr := txn.WithReadTxn(ctx, instance.Repository, func(ctx context.Context) error {
-			stashids, _ := instance.Repository.Performer.GetStashIDs(ctx, t.performer.ID)
-			for _, id := range stashids {
-				if id.Endpoint == t.box.Endpoint {
-					performerID = id.StashID
-				}
+		for _, id := range t.performer.StashIDs.List() {
+			if id.Endpoint == t.box.Endpoint {
+				performerID = id.StashID
 			}
-			return nil
-		})
-		if txnErr != nil {
-			logger.Warnf("error while executing read transaction: %v", err)
 		}
 		if performerID != "" {
 			performer, err = client.FindStashBoxPerformerByID(ctx, performerID)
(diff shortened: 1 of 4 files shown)


internal/manager/task_stash_box_tag.go
```diff
@@ -258,6 +179,87 @@ func (t *StashBoxPerformerTagTask) stashBoxPerformerTag(ctx context.Context) {
...
+			partial.Weight = models.NewOptionalInt(w)
+		}
+	}
+	if performer.Instagram != nil && !excluded["instagram"] {
+		partial.Instagram = models.NewOptionalString(*performer.Instagram)
+	}
+	if performer.Measurements != nil && !excluded["measurements"] {
+		partial.Measurements = models.NewOptionalString(*performer.Measurements)
+	}
+	if excluded["name"] && performer.Name != nil {
```
Flashy78, line 231: Had a question on this in the Studio tagger PR.. why is this one opposite logic vs all the other excluded fields?

WithoutPants: Looks like the original line was `if excluded["name"]`, so the nil check was added afterwards (in #1630). Makes no functional difference.

scruffynerf:
Tested, works great.  Was able to add stashdb ids when config was set to stashdb, pmvstash ids when pmvstash was selected in endpoint config, and both appear now.  Counts also work as expected, counting the number using the current endpoint.

scruffynerf (approved):
works for me, solves the issue.

PR #3556, opened 2023-03-17
Filter edit fixes

WithoutPants:
From feedback from @echo6ix in #3515.

Fixes country selector not displaying correctly in the dialog.

Changes the badge colour to the `info` variant: 
image

Removes the search term from the filter count, since it was misleading.

I've left the positioning as is for now. I couldn't get a position based on the feedback that I was happy with. If you have a css snippet to apply, I'm happy to look at it.

### DIFF
--- a/ui/v2.5/src/components/List/CriterionEditor.tsx
+++ b/ui/v2.5/src/components/List/CriterionEditor.tsx
@@ -174,6 +174,7 @@ const GenericCriterionEditor: React.FC<IGenericCriterionEditor> = ({
         <CountrySelect
           value={criterion.value}
           onChange={(v) => onValueChanged(v)}
+          menuPortalTarget={document.body}
         />
       );
     }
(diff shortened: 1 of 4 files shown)


DogmaDragon:
image
I don't think it looks better to the side. Even if increased in size. 

It's too small without text to act as a buffer between the badge and the icon. 

It works well with text like in the example on https://github.com/stashapp/stash/pull/3515#issuecomment-1472949802 or GitHub, but I'm not sure we want to add text to the toolbar icons. 
image

echo6ix:
DogmaDragon isn't wrong in that it does look better with the text label.

Flashy78:
I set the color to secondary, and I like how it blends in the background and is minimalist.
image

echo6ix:
@Flashy78 That's an improvement actually, but it's also unconvential imo.

This is so so so subjective, but something about the badge placement on that button looks awkward to me. The more I look at it, the more it seems adding a text label balances it (see above Github badges image). Maybe it could only get a text label when the badge count > 0 ?

PR #3557, opened 2023-03-17
Add move files external interface

WithoutPants:
Adds a much needed interface to move a file in the library.

`moveFiles` accepts an input with the following:
- `destination_folder` - full path to destination parent folder. This folder must be within the stash library
- `destination_folder_id` - id of a folder in the stash database
- `destination_basename` - new basename for the file. Only valid if a single id is supplied. Defaults to existing basename. Validates to ensure the new basename is one of the supported extensions. (**TODO** _should I change this to restrict it to keeping the same extension?_)

One of `destination_folder` or `destination_folder_id` is required.

Moves the file in the filesystem and updates the database for the file. Creates any required folders and adds their entries to the database. If the transaction fails, the files are returned to their original location (and logs if it can't do so), and any folders created are removed (and logs if it can't do so).

Note that there is no UI support for this functionality included. I will defer that to after this release. This should simplify some plugins.

### DIFF
--- a/pkg/file/delete.go
+++ b/pkg/file/delete.go
@@ -18,7 +18,7 @@ type RenamerRemover interface {
 	Renamer
 	Remove(name string) error
 	RemoveAll(path string) error
-	Stat(name string) (fs.FileInfo, error)
+	Statter
 }
 
 type renamerRemoverImpl struct {
@@ -44,6 +44,15 @@ func (r renamerRemoverImpl) Stat(path string) (fs.FileInfo, error) {
 	return r.StatFn(path)
 }
 
+func newRenamerRemoverImpl() renamerRemoverImpl {
+	return renamerRemoverImpl{
+		RenameFn:    os.Rename,
+		RemoveFn:    os.Remove,
+		RemoveAllFn: os.RemoveAll,
+		StatFn:      os.Stat,
+	}
+}
+
 // Deleter is used to safely delete files and directories from the filesystem.
 // During a transaction, files and directories are marked for deletion using
 // the Files and Dirs methods. This will rename the files/directories to be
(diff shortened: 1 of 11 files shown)


DingDongSoLong4:
This is amazing ❤️

I would say forcing the same extension is an unnecessary limitation, perhaps a wav file was originally scanned as an mp4  by accident and you want to rename it to be correct.

pkg/file/delete.go
```diff
@@ -44,6 +44,15 @@ func (r renamerRemoverImpl) Stat(path string) (fs.FileInfo, error) {
 	return r.StatFn(path)
 }
 
+func newRenamerRemoverImpl() renamerRemoverImpl {
+	return renamerRemoverImpl{
+		RenameFn:    os.Rename,
```
bnkai, line 49: Depending on the scope of this PR/implementation `os.Rename` might not be enough.
There 2 issues with using it directly
- It doesnt rename files when the source and the destination path are in diferrent filesystems/disks.
Relevant links [original issue](https://github.com/golang/go/issues/13766), [error detection sample](https://github.com/Masterminds/glide/pull/587/files). From a quick look it seems that this behaviour was by design and the most common workaround is to detect the error and then copy and delete the file.
- Rename overwrites existing files

A quick and dirty [code sample](https://go.dev/play/p/Fyh2OcM0Nln) demonstrating the issues
The second issue is pretty easy to handle from a stash plugin so it may or may not be a problem provided we mention that somewhere?
The first is a little more complex and imho it would be better handled by stash.

bnkai:
SafeMove seems to have a small bug
<https://github.com/stashapp/stash/blob/44c58d6e3ce6b22672e32e0703efd96e58a9f74b/pkg/fsutil/file.go#L41>
If the moved file can not be removed (eg read only file system for the source) then we end up with 2 files the original and the copy.

Everything else i tested with latest commit seem to work fine.

bnkai:
>I would say forcing the same extension is an unnecessary limitation, perhaps a wav file was originally scanned as an mp4 by accident and you want to rename it to be correct.

Forcing the same extension is unnecessary, is forcing a valid extension per filetype needed though? Eg renaming a video file to an image could cause issues?

bnkai (approved):

PR #3562, opened 2023-03-19
Log errors returned from graphql

WithoutPants:
Adds an error presenter to the graphql interface that logs any encountered errors to the error log level. Also logs the request arguments at the debug log level when an error is encountered.

### DIFF
--- a/internal/api/error.go
+++ b/internal/api/error.go
@@ -1,0 +1,36 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/99designs/gqlgen/graphql"
+	"github.com/stashapp/stash/pkg/logger"
+	"github.com/vektah/gqlparser/v2/gqlerror"
+)
+
+func gqlErrorHandler(ctx context.Context, e error) *gqlerror.Error {
+	// log all errors - for now just log the error message
+	// we can potentially add more context later
+	logger.Errorf("%s: %v", graphql.GetFieldContext(ctx).Path(), e)
+
+	// log the args in debug level
+	logger.DebugFunc(func() (string, []interface{}) {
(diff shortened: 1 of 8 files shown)


WithoutPants:
Added logging for when the user logs in or out, when invalid credentials are entered, and when credentials are changed.

pkg/session/session.go
```diff
@@ -34,7 +34,15 @@ const (
 	passwordFormKey = "password"
 )
 
-var ErrInvalidCredentials = errors.New("invalid username or password")
+type InvalidCredentialsError struct {
+	Username string
```
bnkai, line 38: Is this actually used anywhere? I only see it set when initializng ?
Wouldnt an empty struct be enough?

internal/api/error.go
```diff
@@ -0,0 +1,36 @@
...
+		s, _ := json.Marshal(args)
+		if len(s) > 0 {
+			args = string(s)
+		}
+
+		return "%s: %v", []interface{}{
+			graphql.GetFieldContext(ctx).Path(),
+			args,
+		}
+	})
```
bnkai, line 31: An extra check is needed here `if graphql.GetFieldContext(ctx) != nil`
If the graphql query is mispelled/missing  then `GetFieldContext` returns nil and thus we get a panic
To repro try running any mispelled or non exitsing query/mutation

pkg/session/session.go
WithoutPants: I originally had this field included in the `Error` method, but I've taken it out for now (until we support multiple users). Having the field means that we can read the username from the calling code, so it's useful, just not currently in use.

bnkai (approved):

PR #3571, opened 2023-03-21
Add linux package lists

myalow:
Added dependency installs for Arch & Ubuntu. There may be a more straightforward way to install yarn on Ubuntu, but I'm not certain.


WithoutPants:
From memory I just used `apt-get` to install yarn as well.

myalow:
`apt-get install yarn` will try to install cmdtest, for whatever reason. See https://github.com/yarnpkg/yarn/issues/3189. [Yarnpkg recommends corepack on their documentation](https://yarnpkg.com/getting-started/install), so that's what I put. Admittedly, the cmdtest package might also work and have yarn bundled or something, I just wanted to take the safe route and recommend what the devs recommend.

PR #3576, opened 2023-03-22
Use 100% width on cards in mobile

WithoutPants:
It's bothered me that the cards don't use all the horizontal space on mobile devices. Changes all cards to use 100% width on smaller devices. Open to feedback on this change.


kermieisinthehouse (approved):
Up to you. 

It's a matter of taste but it could be gotten used to. Maybe 95% or something?

PR #3577, opened 2023-03-22
Embed default icons

WithoutPants:
Embeds the tag image instead of using a var. Embeds a new studio image that matches the other images:

image


kermieisinthehouse (approved):
Code reviewed but not tested

bnkai (approved):

PR #3578, opened 2023-03-22
Require source selector click to show

WithoutPants:
Changes the source selector to require a click to show the menu instead of showing it on hover. This irritated me during scrubbing. I wanted to also do this for the playback rates, but they don't bring up a menu on click, changing the playback rate instead. I can hide the menu in injected css instead.


Flashy78:
Nice, this has annoyed me several times as well.

kermieisinthehouse (approved):

bnkai (approved):

Issue #671, opened 2020-07-17
[Feature] Clips Section


saturdaythrowaway:
I believe adding a separate section for clips (from Reddit/Imgur/Gfytube) would be useful as they can add loads of noise to the main Scene section. Additionally the ability to loop a clip would be convenient.

An additional library may be required as it would be easier to just drop any clips into it vs having to tag existing library items.

WithoutPants:
I'm hesitant to add an entirely new concept (with the associated schema tables, graphql schema and UI pages) when this feels like a pretty big overlap to the existing scenes concept. 

There is already a setting that allows looping of scenes under a given duration. 

I think the issue of small clips adding noise could be alleviated with some better filtering options - for example, saved/default queries - for filtering out videos under a given duration, or tagging such scenes with `clip`.

I think if we're to consider something like this, then it should have a distinct set of features that can't be met by the scenes subsystem.

hyde231:
> 
> 
> I'm hesitant to add an entirely new concept (with the associated schema tables, graphql schema and UI pages) when this feels like a pretty big overlap to the existing scenes concept.
> 
> There is already a setting that allows looping of scenes under a given duration.
> 
> I think the issue of small clips adding noise could be alleviated with some better filtering options - for example, saved/default queries - for filtering out videos under a given duration, or tagging such scenes with `clip`.
> 
> I think if we're to consider something like this, then it should have a distinct set of features that can't be met by the scenes subsystem.

+1 for quick filters, as I also see the problem of noise. Make this a system of user defined and permanently stored filters easily available as a quick drop down box, and everyone is able to define his/her own use cases.

echo6ix:
I landed here to see if anyone had ideas about how to handle clips, and yes, a "quick filters" idea sounds cool.

cloudstrife11:
What about Stash statspage are these little clips also added to the scenes. Maybe its better to add a clips stats also

rg9400:
I think having the concept of libraries could help here. We add paths and can set whether or not images/scenes get scanned, but if we are able to then associate those paths with libraries (a 1 library to multiple path relationship is what other media servers tend to use), then users could segregate out say a clips library from their scene library. Though I guess if someone comingles scenes + clips in the same folder, that may cause issues.

[...]

PR #3583, opened 2023-03-22
Support image clips/gifs

### LINKED ISSUE #926: GIF/Short Loop Image Support
I am not sure if this is in your current design philosophy, but many image pages support the upload of small video snippets, commonly called gif, but rarely actually being in the format of .gif, but most likely mp4 or webm. For an entire page of NSFW examples, see [redgifs.com](https://www.redgifs.com/), but also all image boorus, pornhub, and many more. Those loops are short (most likely less than a minute, always less than two), and are made for looping by default. 
My request is adding support for such videos in the galleries and images view. This could be done by looking at a scan with image only and make all found videos to such a gif entry. As this may break some setups, an additional checkbox might be used, left unchecked by default, or a third option besides scenes and images. It would also mean some kind of viewing in the carousel if possible.

yoshnopa:
Adds support for short videoclips in the image section (commonly referred to by their original image format GIF).

Basically, if a video format is present in a library that has no video enabled or if the format is in the image ending section only, it will not be scanned as a scene, but as an image. This image file has then set the boolean clip to true, which is then used to do all the frontend magic.
Recognition as a video is done with ffprobe, as the current list of supported image formats is limited the reverse of this can simply be tested.
As a thumbnail, a vp9-webm is generated that is a smaller representation of the file and cut to a maximum of 30 seconds. 
This number is pretty debatable and can be adjusted accordingly. Audio is of course disabled.
Generation of Thumbnails is done analogous to how its done before, either by choosing it at the point of scanning or on the fly when the image page is opened. As a transcoded video is a high load in general while the result being less than 1 MB most of the time, I decided that these clip thumbnails will be saved regardless of the image thumbnail saving setting. 
If this seems wrong and misleading, I can take that back out.

Frontend presents the video like an image in a loop with no controls and with autoplay enabled, so that it represents a classic GIF. You can get a pretty vivid image wall with this setting. 
The lightbox works by playing and pausing the video if it gets to the forefront/leaves it again. The navitems there are treated in the same way as the wall/grid, autoplay and loop.

Open for feedback here!

Closes #926 and arguably #671


WithoutPants:
I have significant issues with the way this is designed - and I think the original idea has issues as well.

Having an `ImageFile` have a flag that essentially makes it a video file is not great interface design. If it's a video file, it should _be_ a video file. A short clip is closer to a video than an image. This appears to me to be designed from the point of view of the UI, and it needs to be the opposite. 

I realise that the current design which classifies all videos as scenes is kind of the root cause of this, and it needs to be more flexible. I think the way that this is designed and implemented will make things harder to maintain going forward.

yoshnopa:
@WithoutPants  Ok, I don't want this to bring the project and the backend into a bad state.

From what you say, it seems like it would be ok if the clips would be added as video_files that are associated with the image table the same way the image_files are? 
That could be done.

But before I start going through all of this again, I'd like to know what exactly has to be done for this to be acceptable to you. Can you for example tell me the issues you have with the original idea?

I hope you agree that this is at its core an important feature, one that basically any bigger online Image collection supports (boorus, rule34, pornhub, imgur, redgif etc). While true, this is kind of a weird way to organize this data, it is relatively established that way, especially for "scrollerpages", a la tumblr, reddit etc.
That's also why I would see them as being not in a separated section, not only is the distinction there to scenes even smaller, they are commonly presented from the same source.

xx790:
> I realise that the current design which classifies all videos as scenes is kind of the root cause of this, and it needs to be more flexible.

I think Scene should be just a special case of a tag namespace that allows to attach a video to a tag instance. (I mention the same idea from various angles here and there under various issues but haven't made a write-up yet.)

Users should be able to configure stash data model to meet their needs. The subject of this PR can probably be boiled down to representing a discriminated union of underlying types under a common root.
I'd probably kept individual tag namespaces constrained to only accept a single type of attached content and then allowed to define collections and UI pages with sets of namespaces allowed to be included.

yoshnopa:
@xx790 While an interesting suggestion, this is unfortunately far more difficult done than said.
Go is very restrictive on types, opening that up means basically rewriting the entire backend (or like 80% of it). 
Also theres the problem of representing this change in the UI. How would a gallery in the scene page be viewed? Would the lightbox open on click? It would get really messy real quick, except if you rewrite the entire structure there as well.
So I think you can start over if you would like to have that completely configurable, and while you're at it you should probably use something more lax on types to grant more flexibility, if thats the goal.

WithoutPants:
> But before I start going through all of this again, I'd like to know what exactly has to be done for this to be acceptable to you. Can you for example tell me the issues you have with the original idea?
> 

- conceptually, video files are and should be videos (currently "scenes"). Images are by and large assumed to be static by the system - we store only the dimensions of them. I'm not really keen on anything that muddies the distinction between these two entity types.
- I agree that videos and images should be able to co-exist in galleries. It's an area that needs attention.
- It would be better to refactor the lightbox to be able handle images _and_ videos, with options to hide controls and looping etc on top of that.
- improving the presentation of videos for these sort of shorter clips is not a simple task and I don't really have simple solution. The first step I would try would be a new scene view that is more suited to shorter clips.
- having some sort of unified view for videos and images is also a possibility, but I don't have a simple solution to that either.

The summary is that the original idea needs to be workshopped further. I don't really have the solution for you or guidelines of what to write; I don't have the mental bandwidth required to solve such a complex problem when I have other areas of the system that are more pressing. I'm sorry if this response is not what you wanted to hear - I've spent a great deal of time trying to figure out an easy way to reach the goal you're aiming for, but I just haven't been successful.

yoshnopa:
Thank you for your explanation.

To be honest, I find that to be a bit frustrating, for two reasons:

The first one is very straightforward, I ASKED about the implementation of this feature on the 28.12.22 on the discord programming section (different Username). 
You stated that this would be a viable option. I implemented it basically exactly as I described it there. The internal about video and imagefile are of course left out there, I was at that time working on the chapter stuff, but I already offered to change that.
What you are writing now however is much more restrictive and definitively at odds with what you said there. It was exactly the reason why I asked beforehand.

The second thing is that your proposals, although surely a good addition to the project, is far more complicated and out of scope of what I try to achieve here. I agree that these things may be interesting to add, but to put all of this into this commit is a bit much and it doesnt necessarily even solve what I wanted to achieve with this commit in the first place.
So while I can certainly rewrite everything to PREPARE for such a scenario, basically keeping the "bigger picture" in mind, this is not what this pull request in general is for. For example, the lightbox in its current for is basically ready to support a scene being played in there already.

This request solves a certain oddity, that exists even if we try to solve it by meddling in the structure of image and scenes: 
Images can be animated. Whether you use a generally supported mp4, or a webp, avif or a gif file, it is possible and widespread for this type of media. Right now, the capability of this project to provide this in a satisfying way is very limited, the preview is static or may even break right now, and the supported formats are the most inefficient at their job (size and in case of gif color palette). 
We can solve this by making scenes more flexible, but they would need to be treated like an image in many circumstances, and we would need to distinguish them from the rest. I can't see why this would be better than treating an image like a video in the same circumstances, especially in the frontend, as the distinguishing from a clip to an image is optional for the user, making it easier to use. In contrast, a video being a clip and therefore an image would mean that they potentially populate two spaces, making them at least unintuitive, no matter what design decisions we make here. There would also be the question of preview generation for clips being potentially different than for a scene: Right now, a preview is set together from pieces of the content, while a clip always is limited to the first 30 seconds. The first one is good for identifying a scene and its content visually quickly, the second one is more suited to the presentation on an image wall/being viewed side by side, maybe even without even looking at them in their entirety.
The backend, that was the theory I was looking on, could do this with a more flexible interface, supporting video- and imagefiles and presenting all data necessary for the distinction. I agree that this is a bit weird, but I don't see it as wrong per se, not any more than a gallery that returns video and image type as their content for example.

I absolutely understand that you don't want to put out a solution for that thing and you are doing many things here (thanks for that!), but what I'd like to have is the easiest way to implement this feature that you are ok with. I will gladly do anything to make this as maintainable as possible and design it to be open for certain future implementations.

yoshnopa:
Just for Context:
Bildschirmfoto vom 2023-03-26 11-26-51
image

xx790:
"bigger picture" needs to be conceived and written somewhere.
Otherwise stash is going to join the set of bad examples of open-source bazaar, as a hodgepodge of local ideas that is difficult to move forward as a whole.

This also relates to my above comment and reply:

Yes, major refactoring is a lot of work. I can't expect anyone to do it as a part of a particular feature. But that's what a bigger picture for - each PR can be considered in scope of how much it adds/substracts from the amount of work required to get there.

Not doing anything about underlying architecture has its own cost that might be no less actually. It's a delayed trap. Any local feature increases the code complexity faster.

Starting over is a lot of work, especially for a single developer. Until the right conditions happen, we are stuck with what we have. If stash keeps going the apparent current way - that will have to happen someday, but who knows when...

No, I don't think Go has to be an obstacle. With enough abstraction any turing-complete language might have a solution, and I think here it is mostly about database design. (note: I'm not very familiar with Go and might be wrong here).

yoshnopa:
@xx790 I didn't mean that go per se is an obstacle. I mean that the design philosophy of go is not very friendly towards a structure like you propose it. I just wanted to express how many obstacles are in the way to do this, and I think go itself (and typescript) would be one of them. 
And starting over is a lot of work, but I'm quite serious that this might be easier for the goal you have in mind. You can see how this request already is not without its controversies, its just not how it is designed at all on any level, so making it entirely universal is really a complicated goal unlikely to be set by this project. Very sorry to say that. You can of course try it anyways, this is just a warning about the scale.

xx790:
Is the "design philosophy" written somewhere?

If someone put on a system architect hat, look over all feature requests and think how to enable them in a best way, a solution akin to mine might come up. It would require a lot of refactoring but eventually enable much more usage scenarios per line of code compared to bodging individual features on top of existing architecture. And if it is written somewhere - it will give much more assurance for those working on new PRs.

I see this PR as trying to get an individual feature with the least changes to the current code, regardless of any consequences. I fail to see it as a step in any direction, since there is no apparent direction.

yoshnopa:
@xx790 The design philosophy I'm talking about is the idea of static types. Every function or design choice you do has to be exactly defined with whats going in or out. If anything changes in that regard, you'll have to create new types/interfaces. Typescript follows the same idea in many ways.
This means that a backend that can basically can associate data types in multiple ways with each other is by the core design choice of go wrong. You can circumvent that with generics (that is what I would have to do with the videofile, and why I say that this would be a first step into that direction), but they have to be written to be compatible with every scenario you come up with, and explicitely so. 
When you write such a thing, you basically already have made the idea of static types meaningless, as a single type represents all kinds of inputs and outputs. That means that it is, again, not really in the vein of go and more typical for languages like python or javascript/nodejs. This projects holds onto this principle tightly from what I've seen, which is why I said that its unlikely to be considered in a future request/refactor.
What I do is write a more restrictive generic, that still represents only a type that is to be associated with images and nothing else, and itself can only be a videofile or an imagefile. So one could argue that, although not optimal, this is the way to keep everything as static as possible under these circumstances. That way I try to adhere to this core philosophy of the projects code and go (whether I'm successful is definitely part of this discussion).

But you are correct, this feature is not meant to be universally applied. What I try to say is that it is the way it would be done if your idea would be a future goal, so it can be interpreted as a step into that direction for somebody looking to implement that.

WithoutPants:
> The first one is very straightforward, I ASKED about the implementation of this feature on the 19.12.22 on the discord programming section (different Username). You stated that this would be a viable option. I implemented it basically exactly as I described it there. The internal about video and imagefile are of course left out there, I was at that time working on the chapter stuff, but I already offered to change that. What you are writing now however is much more restrictive and definitely at odds with what you said there. It was exactly the reason why I asked beforehand.

My apologies for the inconsistency and resulting frustration. I have spent some time prototyping to try to come up with a design that might work.

I think that in the backend `ImageFiles` should be reserved only for static images files (with the exception of `gif`). We only store the format and dimensions of these files. Video files should always _be_ video files. However, we can relax the relationship between `Image` and `ImageFile` to instead be to `File`. 

I think there should be better control over when a video file should be handled as an image instead of a scene. Having video files treated as images only if the image extensions includes video extensions _and_ if videos are excluded in the applicable library seems too much like magic behaviour. I think that a more advanced configuration eventually needs to be considered (like what stg-annon suggested in the attached chat screenshot). For the purposes of this PR, I propose the following:
- it will not be required for video file extensions to be added to image file extensions
- add a new global configuration flag that turns on the ability to scan video files as images (in the conditions detailed in the next point)
- if a video file is scanned in a stash library that excludes video files _and_ said global configuration flag is enabled, it will be handled as an image

Currently, the graphql `Image` type has a `files` field which are of the `ImageFile`. In the graphql schema, we can add a union type `union VisualFile = VideoFile | ImageFile` and add a new `visual_files` fields to `Image`. The UI can determine if it should treat an image's file as a static image or video by checking the `__typename` field of the resulting objects. The existing `files` field should be deprecated and only return image files. I'm not totally happy with the way that this weakens the abstraction of the `Image` type, but I think it's the best short-term approach.

I've pushed a very rough prototype for the backend changes to the following branch: https://github.com/WithoutPants/stash/tree/experimental/video-image

Thumbnail generation is another area that I think needs further workshopping. I think on-the-fly thumbnail generation of video files is probably not a great idea for this - it should be part of the generate functionality. I think on-the-fly thumbnail generation should generate a static image. This is much less resource intensive, and will only be used if a video thumbnail hasn't been generated. This will mean adding generate (and scan) flags for generating videos for video-based images. I don't want to get into anything more detailed than that, but I hope that it's a starting point. 

I'm reasonably confident that we can do this without database schema changes.

I'm happy to workshop this further as needed, but I hope that this gives a better direction and that the idea is acceptable.

yoshnopa:
@WithoutPants Thank you so much for looking into it!
This is basically what I suggested as change, I will look into your branch and take it as a starting point. I also think that I can go on from here, and no Database changes will be needed at the end. 

Just as a little note, the current scanning is very much the way you described how it should be done right now, but without the settings button:

A videofile will be treated as an image file if:
The videofile ending is in the image section but not in the video section
OR
The videofile ending is in the video section but scanning for videos is disabled in the library where the file was found

I will add a settings button if this seems to be to obscure to you, no problem.

yoshnopa:
@WithoutPants I've done the requested steps:

The Clips are now scanned, saved and delivered as VideoFiles.

The Frontend generates only stillframes on the fly, while the preview of clips can be triggered by a generate button and on scan when enabling the according option. This also means that there are actually 3 Paths for images, the image itself, the thumbnail, and the preview (always empty for imagefiles and empty before generation for clips). 
An option allows for the scan to include Video File Endings as Clips if Video scanning on the library the files are found in is disabled. Right now, the option is enabled by default. When its disabled, you can still find Clips by adding the according File Endings to the Image File endings AND removing them from Video File Endings. 

I've also tested the Clean Task, the general removal of Images/Clips and Import/Export function with a quick look into the database. Everything works fine.

I've also fixed the Problem I requested in Discord. I did this by manually adding the union definition in the createClient.ts.
If this should be automatically extracted from make generate command and the graphql files, I would need assistance on that. Otherwise this can stay like that until another union is added, so maybe its not relevant enough for the generation process.

WithoutPants:
This is looking ready to go. I've committed a bug fix (clip generation wasn't work during scan), some refactors and done some clean up. Can you please perform a final check to ensure its working correctly? Once that's confirmed this should be ready to merge.

One final thing that is missing is documentation. I'm happy to merge without it, but we should add it.

yoshnopa:
Tested a few things, works (still) great!
I've added a documentation internally, I think @DogmaDragon merged this into the website last time I did something to it?
So from my side no problems left, ready to merge @WithoutPants !
Thanks again for looking over this really not trivial stuff...

PR #3590, opened 2023-03-24
Build frontend source maps

BawdyInkSlinger:
This builds frontend sourcemap files. It should make life easier for those that want to create user scripts and help those who need to troubleshoot front end errors.

See https://vitejs.dev/config/build-options.html#build-sourcemap for more info.

### DIFF
--- a/ui/v2.5/vite.config.js
+++ b/ui/v2.5/vite.config.js
@@ -9,6 +9,7 @@ export default defineConfig({
   base: "",
   build: {
     outDir: "build",
+    sourcemap: true,
     reportCompressedSize: false,
     rollupOptions: {
       output: {


DingDongSoLong4:
It's worth noting that this increases the total UI build size from ~7MB to ~43MB, and therefore the resulting stash binary by the same amount, from ~40MB to ~76MB. The sourcemaps aren't loaded by the browser until you open devtools, so page loading isn't affected.

kermieisinthehouse (approved):

Flashy78:
Wow double the size of the binary, and yet we can't add commonly used python modules to the docker image due to a couple of megabytes?

WithoutPants:
I don't think such a large increase in binary size is worth it for what is more development-oriented tasks.

BawdyInkSlinger:
> I don't think such a large increase in binary size is worth it for what is more development-oriented tasks.

Genuine question: 

1. Have you ever heard somebody say, "I'd use stashapp because the binary is small."?  
2. Have you ever heard anybody say, "I prefer this app over stashapp because it has a smaller binary."?

The people that use apps like these buy new hard drives because they don't want to delete that 4k version of a 7gb Japanese vid they plan on watching some day; what's another 30mb?

WithoutPants:
> > I don't think such a large increase in binary size is worth it for what is more development-oriented tasks.
> 
> Genuine question:
> 
> 1. Have you ever heard somebody say, "I'd use stashapp because the binary is small."?
> 2. Have you ever heard anybody say, "I prefer this app over stashapp because it has a smaller binary."?
> 
> The people that use apps like these buy new hard drives because they don't want to delete that 4k version of a 7gb Japanese vid they plan on watching some day; what's another 30mb?

It's not a question of storage, it's a question of downloading bandwidth and time. We're talking about almost doubling the size of the binary for something that 95% of users (conservatively) are not going to benefit from.

bnkai:
Considering the usage (maybe 1% of of the users ) and the fact that devs only need to add a single line to a file isn't it better to just add a note in the [development](https://github.com/stashapp/stash/blob/develop/docs/DEVELOPMENT.md) section of the documentation? Imho either that or an extra target in the Makefile (if possible) are more appropriate than turning the stash binary into a dev/debug one.

BawdyInkSlinger:
> Wow double the size of the binary, and yet we can't add commonly used python modules to the docker image due to a couple of megabytes?

At the moment, I think I'm more interested in this topic than the PR. And, I think understanding this would provide necessary context to understand the concern of this PR. Is it okay to discuss this here or should it be moved elsewhere? 

Assuming the former, what is the philosophy guiding the executable's file size? On discord I learned the current download time is already slow for some, so the developers don't want to increase the file size unless there is a strong justification for it. That's pragmatic, especially there are many people with slow connections. 

But I'm confused about such a demographic; specifically, why would those types have interest in using stashapp? Let me lay out my assumptions so they can be corrected:

1. This app is meant for people who want to organize their pictures and videos. 
2. The typical user gets these pictures and videos from the Internet. 
4. The typical user is accustomed to waiting long periods for their pics/vids to download.
5. ~100mb is 0-60 seconds of a video and considered short by most.
6. The executable is ~40mb and this PR would increase it to ~80mb.

Why would a user care about how long it takes to download stashapp _and_ have with enough content to justify using it in the first place? That's where I get confused.

WithoutPants:
Closing in favour of #3603

PR #3594, opened 2023-03-25
Improve caching, HTTP headers and URL handling

DingDongSoLong4:
This is a whole bunch of fixes and improvements related to URL handling and HTTP headers:

- Fixes for the 'View All' links on the front page and the logout link not respecting the proxy prefix
- Improvements to the login page relative URL handling (fix broken CSS on `/login/` with a trailing slash)
- `/logout` now redirects to `/login` instead of serving the login page at `/logout`.
- The `/customlocales` endpoint is no longer fetched twice during locale initialization.
- Removes the `window.STASH_BASE_URL` assignment, the `<base />` element is already there and has the same information.
- The embedded UI filesystems now have their base directories in the correct place, removing the need for prefixes all the time.
- The CSP header is now only sent for the actual pages instead of every response, and the exceptions for the CDN used in the graphql playground are now only added there and not for every page. The playground also now works with a proxy prefix.
- Probably the most important improvement, a Cache-Control header along with an ETag header or a Last-Modified header is now sent along with every response. The UI assets all contain hashes in the filenames, so they are marked as `immutable` to avoid a request entirely. Everything else that is worth caching uses `no-cache` to have the browser validate with the server whether its cache is still valid. This should finally prevent all the issues where users have to hard refresh or clear caches for changes to show up.
- Responses to CORS preflight OPTIONS requests are now sent even if not logged in - preflight requests should not be blocked by authentication.
- A Forbidden error is now returned for all non-webpage requests (ie those with some extension that isn't `.html`) instead of redirecting to the login page.

(I don't actually use a proxy prefix, I found all those issues while testing the other changes)


DingDongSoLong4:
This is the location I was using in nginx for testing:
```
location /stash/ {
    # trailing slash removes proxy prefix from requests
    proxy_pass http://localhost:9999/;

    client_max_body_size 100M;

    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    proxy_set_header Host $http_host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Port $server_port;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_set_header X-Forwarded-Prefix /stash;
}

```

PR #3599, opened 2023-03-25
Blob fixes

### LINKED ISSUE #3595: [Bug Report] File Storage blobs not being removed after new blob
Blobs are currently not being removed after new ones are created for a record.

For example, if you change a performer or tag image, the old blob still exists within the file system and the new one is also added.

To retroactively clean these for early adopters like myself, maybe the current clean task could look at existing blobs in the file system and remove them if they aren't found in the database.

WithoutPants:
- log errors when getting image blobs instead of erroring out in a variety of places
- adds debug logging when writing and deleting blobs from the file system
- correctly remove blobs from the database/filesystem when they are no longer referenced - fixes #3595


WithoutPants:
Note that this does not include any changes to the clean task to delete stale blob data. This will be considered in a later feature update.

QxxxGit:
Tested and the blobs are now being removed. Thanks!

PR #3603, opened 2023-03-26
Add opt-in UI sourcemaps

DingDongSoLong4:
Adds a Makefile target `ui-sourcemaps` which uses the `VITE_APP_SOURCEMAPS` environment variable to add sourcemaps to the UI build. Sourcemaps can also be enabled by setting `STASH_SOURCEMAPS`, when running plain `make` or `make release` for example. The line in the Makefile can also be uncommented if you don't want to add the environment variable every time. 

Also did some cleanup on the Makefile, which includes adding `.PHONY` everywhere and adding build flags to `server-start`.

Related to #3590

Edit: Added a similar `STASH_NOLEGACY` environment variable to disable legacy chunk generation, which decreases build size and time if you don't need to support legacy browsers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -246,6 +246,10 @@ ui: pre-build
 	$(SET) VITE_APP_STASH_VERSION=$(STASH_VERSION) $(SEPARATOR) \
 	cd ui/v2.5 && yarn build
 
+.PHONY: ui-sourcemaps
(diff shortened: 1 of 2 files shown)


kermieisinthehouse:
I like this solution a lot better.

PR #3613, opened 2023-03-29
Add default blob data location

ChilledSlim:
On a new installation, Stash will default to 'blobs' (which relative to the CWD of /, becomes '/blobs'). In our recommended docker installation, this is NOT mapped and as a result will NOT be persistent across the container being removed or updated.

So by default, we are populating blob data which will be deleted, likely leaving to some confused new users. **This should also be noted for existing installations to make this a persistent path.**

### DIFF
--- a/docker/production/docker-compose.yml
+++ b/docker/production/docker-compose.yml
@@ -36,5 +36,7 @@ services:
       - ./metadata:/metadata
       ## Any other cache content.
       - ./cache:/cache
+      ## Where to store binary blob data (scene covers, images)
+      - ./blobs:/blobs
       ## Where to store generated content (screenshots,previews,transcodes,sprites)
       - ./generated:/generated


ChilledSlim:
@WithoutPants Drawing attention to the bold part which really should be in the release notes as I'm sure there are many docker installations out there.

WithoutPants:
Thanks for the pick up on this. In terms of existing installations, I'm not sure what exactly I should be noting. Existing installs will need to explicitly set the blobs path if they choose to change to filesystem-based blob storage, and much like the generated path this should be persistent.

ChilledSlim:
> Thanks for the pick up on this. In terms of existing installations, I'm not sure what exactly I should be noting. Existing installs will need to explicitly set the blobs path if they choose to change to filesystem-based blob storage, and much like the generated path this should be persistent.

You're probably right.  Maybe something in the release note of "Ensure the path set is persistent (such as a persistent volume in containerized installations)".

TBH, I would have probably missed it, set it to /blobs, migrated all of the images, deleted the images and found out a month from now that all of my images are gone.  I think there's a high risk of that right now, even if it is user error.

WithoutPants:
Yeah, I could see myself doing something like that as well. It's definitely a high risk. I can add it to the release notes, but I'm not sure if it's going to be enough.

Issue #3517, opened 2023-03-09
[Feature] Make height and weight fields floats instead of ints


SpedNSFW:
**Is your feature request related to a problem? Please describe.**
Title should be self-explanatory. Give us access to more precise measurements. Most American performers measure themselves in feet and inches like the silly people they are. They don't convert to centimetres exactly 99% of the time. For example, 6"4' is 193.04cm but since we can only enter 193, this is displayed as 6"3'.

**Describe the solution you'd like**
Change the data types to floats, from integers.

**Additional context**
The height field on stash-box should also be changed to a float.

LeGrosFromage:
Or how about being able to enter feet and inches?

SpedNSFW:
> Or how about being able to enter feet and inches?

To do that, you'd have to have a toggle or make it a string. Both of which impede filtering. Also, why on Earth would you want less accurate measurements?

LeGrosFromage:
How is entering someone's height as 5 feet 10 inches, 5 feet 8 inches, 6 feet 5 inches, less accurate?
You have 2 number-only fields - one for feet, one for inches.  From there, it shouldn't be too hard to filter for a certain height or in a range of heights.

Also, some people (like me) probably don't really care for height measurements - kinda irrelevant for me - so precision isn't really required.

DogmaDragon:
> How is entering someone's height as 5 feet 10 inches, 5 feet 8 inches, 6 feet 5 inches, less accurate? You have 2 number-only fields - one for feet, one for inches. From there, it shouldn't be too hard to filter for a certain height or in a range of heights.

1 cm = 0.3937007874 in
1 in =  2.54 cm

So centimeter is a lot more granular than an inch hence more accurate. 

> They don't convert to centimetres exactly 99% of the time. For example, 6"4' is 193.04cm but since we can only enter 193, this is displayed as 6"3'.

I'm pretty sure it's rounding to the Nearest Integer. 4 is less than 5 and 0 is less than 5, hence 193 instead of 194. And 193.04 is closer to 193 than 194. 

Again centimeter is a lot more granular than an inch hence more accurate.

DingDongSoLong4:
Storing heights as centimeters just makes more sense - splitting them into two fields for feet/inches makes everything more complicated. Storing floats is of course possible, but in my opinion the extra precision is unnecessary.

I think the real bug here is that the conversion from cm to imperial is always rounding down, which should be quite easy to fix.

SpedNSFW:
[...]

Issue #3267, opened 2022-12-11
[Feature] Additional Performer Database field for Penis Type


KrisVardo:
As [mountainmantra](https://github.com/mountainmantra) pointed out here:
https://github.com/stashapp/stash/discussions/3265#discussion-4649812

"The current implementation of the performer table is gender specific, for females. It seems that the original intention of the Measurements field is for Breast Size, with the following field being Fake Tits."

Screenshot:
https://user-images.githubusercontent.com/11653895/206876677-261cdd54-0d33-40d6-89ab-33938322002b.png

Based on our discussion here:
https://github.com/stashapp/stash/discussions/3265#discussioncomment-4372650

**A new text field should be added, "Penis Type".**  
Penis Type can capture circumcised status, as well as be open-ended for other situations (prosthetic, etc.).

It should be a new field, rather than overload "Fake Tits". There are trans performers with both. Both "Penis Type" and "Fake Tits" should probably be available always (for any gender), and can just be "n/a" if they don't apply.

KrisVardo:
Screenshot of how this would look:

<img width="1084" alt="Screen_Shot_2022-12-11_at_2_54_43_PM" src="https://user-images.githubusercontent.com/114971575/206933763-270eab64-ea89-4cb3-afba-8f295817566a.png">

strubdog:
I love this idea.

echo6ix:
Basically a duplicate of https://github.com/stashapp/stash/issues/2735 (though it might be phrased awkwardly in the latter issue).

- Displaying fields with `N/A` isn't necessary, it just adds visual clutter. It's essentially a null value, and the omission of the field communicates the same meaning (that it is N/A). 
- Penis size and "penis type" should probably be split into different fields at least on the backend to facilitate filtering/queries:
  - `Penis size`: *integer*
  - `Foreskin`: *boolean*
    - `True`: Uncut
    - `False`: Cut
  - `Penis type`: *string*
    - ????????????????????????????

binarygeek119:
Should also add thickness

KrisVardo:
I don't think you're going to get that much detailed data about that many performers, enough to make it searchable or worth it.  Or that's consistent - especially on size.  I think that's making it too complicated, even though it would be nice/interesting.  Just having a field to store some words that are searchable is enough.  It's easy to search for "contains 'uncut'" in the "penis type" field and find them all.

DogmaDragon:
[...]

Issue #2735, opened 2022-07-12
[Feature] Add penis size filters for male performers


MidnightDesigns:
**Is your feature request related to a problem? Please describe.**
I noticed the women have a field for breast size, but the men don't have anything for penis size. I would like to be able to sort using this function too since it's essentially the gay man's version of having a breast size field for straight guys. 

**Describe the solution you'd like**
Categorizes added for penis length, girth, and circumcised or uncircumcised. The girth info is probably not that widely available but the rest is. 

**Describe alternatives you've considered**
I was using tags but it gets tedious, it would be nicer to have it be part of the model page information. 

**Additional context**
None.

Maista6969:
Can be closed since #3627 was merged

PR #3627, opened 2023-03-31
Add penis length and circumcision stats to performers.

### LINKED ISSUE #3761: [Bug Report] Editing one or more performers using the Bulk Editor will fail if none of the performers have their gender set.
**Describe the bug**
Editing one or more performers using the Bulk Editor will fail if none of the performers have their gender set.

**To Reproduce**
1. Create two or more performers using the "New" button, set only their Names and nothing else.
2. Select one or more of the performers, click on the Edit icon and edit anything other than Gender.
3. Hit save, and the error message "Response not successful: Received status code 422" pops up.
I checked the GQL using the browser's debugger. The error message is:
`{"errors":[{"message":" is not a valid GenderEnum","path":["variable","input","gender"],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}}],"data":null}`
I think this is due to `gender` being set to an empty string, when it should be `null`.  

**Expected behavior**
The edited performer should be saved. 

**Stash Version: (from Settings -> About):** v0.20.2

 - OS: Windows 10
 - Browser: Vivaldi
 - Version: 6.0.2979.22 (Stable channel) (64-bit) 

**Additional context**
The circumcision field in PR #3627 has the same bug as it was based on the gender implementation.

departure18:
I've added the ability to edit a performer to add their penis length and whether they are circumcised or not.
penis_stats
- They can be edited from the performer page, scrapers and graphql.
- Performers can be filter by penis length and circumcision too.

This is my first time programming in Go and Typescript, also my first pull request. So any feedback would be welcome. Hopefully i haven't done anything too foolish!

Fixes #3761


Flashy78:
Nice, it addresses two wanted features #2735 and #3267 

It looks like you're capturing penis length in centimeters (to match the height in cm and weight in kg), so could this be a regular int type to match?
Can also add _cm to the graphql type name to match height.

It feels like circumcised should just be a nullable boolean of yes or no, or are there other values that can go in that enum?

You can merge your database migrations into a single migration.

departure18:
> Nice, it addresses two wanted features #2735 and #3267
> 
> It looks like you're capturing penis length in centimeters (to match the height in cm and weight in kg), so could this be a regular int type to match? Can also add _cm to the graphql type name to match height.

I originally wrote a version using integers, but i changed to floats. This was partly because the numbers are small so conversions between inches and cm led to comparatively big changes when rounded (and no guy wants people thinking his penis is smaller due to rounding). Also at least one popular gay porn site quotes penis size to the quarter centimetre.

Isn't height_cm just because height was originally a string and changed to an int a couple of versions ago? Weight isn't weight_kg in the graphql?

> 
> It feels like circumcised should just be a nullable boolean of yes or no, or are there other values that can go in that enum?
> 

I thought about that, but i felt it needed to have three values for circumcised, uncircumcised and unknown (or not set). I couldn't see a way to to do that with a boolean (I might have missed something, my starting point when i was writing this was looking at the sqlite database and i couldn't see any booleans set to null). So i ended up taking my influence from the implementation of the gender setting.

> You can merge your database migrations into a single migration.

I wrote the penis length first and then the circumcision. Seemed cleaner to have them as two separate commits, though i suppose i could do a further commit to combine the database migration, do we need to conserve database schema numbers?

swapnilgarg7:
10/10 agree

Flashy78:
I say that due to https://github.com/stashapp/stash/issues/3517#issuecomment-1463066278 regarding percision. If you do need quarter cm's, then perhaps what WP suggested of a two digit decimal value is more appropriate than float.

Denoting the measurement type in the name makes it explicit for everyone using it to know what type of value they should be sending. Weight should probably have probably had _kg on it so folks know not to send in lbs when using graphql, without having to go through the code to figure out what it's storing. Height remains in addition to Height_cm in order to keep backwards compatability with the api.

If there's no need for other values, then a nullable boolean keeps things simplified without having to do conversions on enums. True, False or Null.

Your PR is squashed when merged, there's no concept of your individual commits. So instead of _the_ DB migration for this PR, you have multiple migrations of the same table. Migrations have traditionally not been the fastest operation, although that has changed in .20 if people migrate their blobs out of the database.

departure18:
> I say that due to [#3517 (comment)](https://github.com/stashapp/stash/issues/3517#issuecomment-1463066278) regarding percision. If you do need quarter cm's, then perhaps what WP suggested of a two digit decimal value is more appropriate than float.

My reading of that issue is that a user saw measurements being incorrectly converted from inches to cm due to the program always using floor instead of round when rounding numbers and asked for more precision to avoid that. So the sensible answer was to fix the bug rather add more precision.

I agree that for heights (which tend to be in the 140-200 cm range) using integers is precise enough. But penises lend to be 30cm at most, so 1 or 2 decimal point precision is justified. 

I did think of using millimetres so i wouldn't have to use floats, but i wanted to keep the units consistent between height and penis length and anyway most sites that use metric use cm and i didn't want every scraper to have to convert cm to mm.

I don't entirely understand your suggestion of using a decimal. There's no native decimal type in Go or Typescript (that i know of) and Sqlite only stores numbers as floats and ints. So to use a decimal type would have required me to pull in additional library and would have no doubt have been a bigger pain to do than what i ended up doing.
 
> Denoting the measurement type in the name makes it explicit for everyone using it to know what type of value they should be sending. Weight should probably have probably had _kg on it so folks know not to send in lbs when using graphql, without having to go through the code to figure out what it's storing. Height remains in addition to Height_cm in order to keep backwards compatibility with the api.

height_cm was added to the GraphQL API in November, due to the height previously being a string rather than a int. So adding the unit onto the end of the quantity name doesn't seem to be a cast iron precedent in this program, rather a work-around for a bad decision in the early implementation. The UI specifies the unit for most non-geeky users, for people writing scrapers and using GraphQL, i'd say the unit should be specified in the documentation (admittedly i've not done any documentation so that would be a fair criticism).

> If there's no need for other values, then a nullable boolean keeps things simplified without having to do conversions on enums. True, False or Null.

Maybe. As i said i did think about using a boolean at first, but after looking through the code, ripping off the implementation of the gender stat seemed nicer to me. It might not be as simple as using a boolean, but i don't think i've added anything to the code base that is so mystifying that no one else could ever maintain it (if you understand the Gender stuff then you understand this). 

> Your PR is squashed when merged, there's no concept of your individual commits. So instead of _the_ DB migration for this PR, you have multiple migrations of the same table. Migrations have traditionally not been the fastest operation, although that has changed in .20 if people migrate their blobs out of the database.

It took barely any time to migrate on my stash database. I tried combining the migration files and it took a similarly negligible length of time to migrate. Migration only has to be done once and apart from the last migration when massive blobs of data were moved into the filesystem, it's never been slow for me. So unless anyone can give me a solid demonstration that my approach is slower, i don't see this as a valid criticism.

DingDongSoLong4:
> > Your PR is squashed when merged, there's no concept of your individual commits. So instead of _the_ DB migration for this PR, you have multiple migrations of the same table. Migrations have traditionally not been the fastest operation, although that has changed in .20 if people migrate their blobs out of the database.
> 
> It took barely any time to migrate on my stash database. I tried combining the migration files and it took a similarly negligible length of time to migrate. Migration only has to be done once and apart from the last migration when massive blobs of data were moved into the filesystem, it's never been slow for me. So unless anyone can give me a solid demonstration that my approach is slower, i don't see this as a valid criticism.

I think what was being referred to here is that you have added two separate migrations, but since your commits are squashed when merged there is no reason to keep them separate - ie just combine them into one. It is not expected that a database used to test this branch will be compatible with a database running on develop.

departure18:
> > > Your PR is squashed when merged, there's no concept of your individual commits. So instead of _the_ DB migration for this PR, you have multiple migrations of the same table. Migrations have traditionally not been the fastest operation, although that has changed in .20 if people migrate their blobs out of the database.
> > 
> > 
> > It took barely any time to migrate on my stash database. I tried combining the migration files and it took a similarly negligible length of time to migrate. Migration only has to be done once and apart from the last migration when massive blobs of data were moved into the filesystem, it's never been slow for me. So unless anyone can give me a solid demonstration that my approach is slower, i don't see this as a valid criticism.
> 
> I think what was being referred to here is that you have added two separate migrations, but since your commits are squashed when merged there is no reason to keep them separate - ie just combine them into one. It is not expected that a database used to test this branch will be compatible with a database running on develop.

I'm by no means an experienced software engineer, but i was working under the assumption that when you're implementing two features keeping the relevant code in separate commits was good practise. That way if one of them was a broken mess and one wasn't you guys could cherry pick the commits and take one but not the other. Hence the two migration files.

I'm not disputing that if my changes get committed to Stash's main tree my three commits become one. But my understanding of Flashy's argument was that two migration files would be slower than one. I tested both approaches, one migration SQL file versus two. In both cases i loaded Stash in my browser, clicked the migrate button and the migration was over in less than 11 seconds. I honestly couldn't say if one was faster than the other (not a surprise since the only difference between the two versions is is that the schema version number changes twice rather than once). So i don't see how one approach is technically superior to the other. 

This seems to me like a housekeeping argument rather than a technical one, one file is tidier than two. But since this seems to be important to people i've made the changes.

Flashy78:
> My reading of that issue is that a user saw measurements being incorrectly converted from inches to cm due to the program always using floor instead of round when rounding numbers and asked for more precision to avoid that. So the sensible answer was to fix the bug rather add more precision.
> 
> I agree that for heights (which tend to be in the 140-200 cm range) using integers is precise enough. But penises lend to be 30cm at most, so 1 or 2 decimal point precision is justified.
>
> I did think of using millimetres so i wouldn't have to use floats, but i wanted to keep the units consistent between height and penis length and anyway most sites that use metric use cm and i didn't want every scraper to have to convert cm to mm.
> 
> I don't entirely understand your suggestion of using a decimal. There's no native decimal type in Go or Typescript (that i know of) and Sqlite only stores numbers as floats and ints. So to use a decimal type would have required me to pull in additional library and would have no doubt have been a bigger pain to do than what i ended up doing.
> 
To quote the comment I linked "Extra precision _is_ unnecessary."
I was trying to find a middle ground so you could store your 1 or 2 decimal places, without resorting to a float and all of the additional code you had to add that goes along with it. SQLite uses type affinity, so that would allow you to have your decimal values, without any additional libraries. Not to mention if you do your conversions, you won't end up with the miniscule rounding errors you get with a float. Nobody wants their penis to be listed .000002% too small.

> height_cm was added to the GraphQL API in November, due to the height previously being a string rather than a int. So adding the unit onto the end of the quantity name doesn't seem to be a cast iron precedent in this program, rather a work-around for a bad decision in the early implementation. The UI specifies the unit for most non-geeky users, for people writing scrapers and using GraphQL, i'd say the unit should be specified in the documentation (admittedly i've not done any documentation so that would be a fair criticism).
> 
It's a quality of life thing, you don't need to add documentation and have the user look for that documentation if you include it in the name of the parameter. Why is it worse if you rename it?

> Maybe. As i said i did think about using a boolean at first, but after looking through the code, ripping off the implementation of the gender stat seemed nicer to me. It might not be as simple as using a boolean, but i don't think i've added anything to the code base that is so mystifying that no one else could ever maintain it (if you understand the Gender stuff then you understand this).
> 
Sure, it works as is. My suggestion is to point out how something could be done more efficient and save you all of that enum conversion code you added.

> It took barely any time to migrate on my stash database. I tried combining the migration files and it took a similarly negligible length of time to migrate. Migration only has to be done once and apart from the last migration when massive blobs of data were moved into the filesystem, it's never been slow for me. So unless anyone can give me a solid demonstration that my approach is slower, i don't see this as a valid criticism.
>
Many people have 10+ gigabyte databases. Their migrations take multiple minutes to run. But like I said, for those who chose to migrate their blobs out of their database with .20, then it won't be such an issue for them.

>I'm by no means an experienced software engineer
My review was not meant to be an attack on what you've written, there are many ways to solve a problem, and rarely one best way.
Like I initially said, it's nice that you've taken the time to address these two wanted features. I'm not even an admin here, so my review means nothing because I have no say in your PR anyway. I won't antagonize you further and let someone else review the front end part.

graphql/documents/data/performer-slim.graphql
```diff
@@ -16,6 +16,7 @@ fragment SlimPerformerData on Performer {
   eye_color
   height_cm
   fake_tits
```
rllane17, line 18: tits_size

rllane17 (approved):

WithoutPants:
> To quote the comment I linked "Extra precision _is_ unnecessary." I was trying to find a middle ground so you could store your 1 or 2 decimal places, without resorting to a float and all of the additional code you had to add that goes along with it. SQLite uses type affinity, so that would allow you to have your decimal values, without any additional libraries. Not to mention if you do your conversions, you won't end up with the miniscule rounding errors you get with a float. Nobody wants their penis to be listed .000002% too small.

I'd prefer to use definite precision when storing these values, so that we can eliminate any chance of floating-point errors. I'll investigate and address this before merging, so nothing required from the author right now.

> It's a quality of life thing, you don't need to add documentation and have the user look for that documentation if you include it in the name of the parameter. Why is it worse if you rename it?

The existing name is good enough. The suffix was added in the other cases for backward compatibility cases. As long as the unit is documented clearly, it should be fine.

> Maybe. As i said i did think about using a boolean at first, but after looking through the code, ripping off the implementation of the gender stat seemed nicer to me. It might not be as simple as using a boolean, but i don't think i've added anything to the code base that is so mystifying that no one else could ever maintain it (if you understand the Gender stuff then you understand this).

I'm not sure about this one. I'd potentially add an N/A value. This has impacts on the stash-box performer concepts as well, so I'll confer with others before making a decision.

> It took barely any time to migrate on my stash database. I tried combining the migration files and it took a similarly negligible length of time to migrate. Migration only has to be done once and apart from the last migration when massive blobs of data were moved into the filesystem, it's never been slow for me. So unless anyone can give me a solid demonstration that my approach is slower, i don't see this as a valid criticism.

> Many people have 10+ gigabyte databases. Their migrations take multiple minutes to run. But like I said, for those who chose to migrate their blobs out of their database with .20, then it won't be such an issue for them.

Performance isn't a factor here. Separate migrations in a single squashed commit (ie both migrations run at once) has a neglible impact on performance. The major performance drain of migrations is the `VACUUM` that occurs once all migrations are completed.

I'd still prefer migrations to be combined into a single file where possible so that we're not unnecessarily bumping schema versions.

> This seems to me like a housekeeping argument rather than a technical one, one file is tidier than two. But since this seems to be important to people i've made the changes.

Basically, yes.

I'll do a proper review when I have the mental capability to do so. Where possible, I will make changes myself. I intend to include it in 0.21.0.

departure18:
> > To quote the comment I linked "Extra precision _is_ unnecessary." I was trying to find a middle ground so you could store your 1 or 2 decimal places, without resorting to a float and all of the additional code you had to add that goes along with it. SQLite uses type affinity, so that would allow you to have your decimal values, without any additional libraries. Not to mention if you do your conversions, you won't end up with the miniscule rounding errors you get with a float. Nobody wants their penis to be listed .000002% too small.
> 
> I'd prefer to use definite precision when storing these values, so that we can eliminate any chance of floating-point errors. I'll investigate and address this before merging, so nothing required from the author right now.

After Flashy's last comment I had a look at using a DECIMAL type in the SQL schema, but after reading SQLite's documentation and doing some tests i decided that due to SQLite's very relaxed type system it  wouldn't actually make a difference.

SQLite's documentation (https://www.sqlite.org/datatype3.html) says it translates DECIMAL to it's NUMERIC affinity. The NUMERIC affinity confusingly means that any type (including BLOB and TEXT types) can be stored in the column. I tested what SQLite would actually do using it's command line program. I tried storing a bunch of different values in a test database using a DECIMAL column and got SQLite to tell me what type it had actually stored the value as, anything with number after the decimal point (e.g. 17.5, 19.99999) got stored as a REAL, anything else (e.g. 16.0, 19) was stored as an integer. Curiously, specifying the column as an INTEGER gives the exact same result as DECIMAL, SQLite doesn't really care what type you specify, it will just store a value in whatever way it thinks best.

I can't say that i've ever 100% understood floating point precision (believe me i've tried a few times), but my sense is that given that it's stored in a 64 bit float and i'm only ever displaying the value to two decimal places any error will be tiny and hidden to the end user. (I hope...) 

> > Maybe. As i said i did think about using a boolean at first, but after looking through the code, ripping off the implementation of the gender stat seemed nicer to me. It might not be as simple as using a boolean, but i don't think i've added anything to the code base that is so mystifying that no one else could ever maintain it (if you understand the Gender stuff then you understand this).
> 
> I'm not sure about this one. I'd potentially add an N/A value. This has impacts on the stash-box performer concepts as well, so I'll confer with others before making a decision.

My intention was that having both stats unset would cover the cases where the performer either didn't have a penis or had a penis, but we didn't know the size or whether it was circumcised or not. If both the stats aren't set, then it won't be displayed, so i didn't see the need for an NA value. In addition a minor irritation of mine is that when i scrape a male performer from StashDB, "Fake Tits" is always set to "Na". It seems superfluous to display that "Na", so i always delete it before saving and i didn't want to inflict something similar on anyone else! 😀

> I'll do a proper review when I have the mental capability to do so. Where possible, I will make changes myself. I intend to include it in 0.21.0.

Excellent, thanks for looking at it!

AdultSun:
Originally pitched this on Discord ([linked here](https://discord.com/channels/559159668438728723/642050893549928449/1109022215132946432)), expanding on it here at WP's request.

My proposal would be to expand the "Circumcision" field to a more generally applicable "Penis Type" field. This mirrors an old suggestion I made expanding the "Fake Tits" field to a "Breast Type" field, originally from [this Discord discussion](https://discord.com/channels/559159668438728723/559399634318131210/849574312138375179).
_(I can repost that as a separate feature request if necessary, I don't believe we have one covering it already. In short, my suggestion is using Natural/Augmented/Reduced/Removed/None as available options and replacing "Fake Tits" as the field name with "Breast Type".)_

The enum options would/could include, mostly in order of very common to least common:
- `NULL`: Default selection, would only be read as "unknown" or "unspecified".
- `Cut` and `Uncut`: Self-explanatory, only applicable for natural penises.
- `Natural`: Ambiguous option if a user/contributor isn't sure about circumcision status, but knows the performer still has a natural penis.
- `None`: Cis female or pre-op trans male.
- `Removed`: Post-op trans female.
- `Augmented`: Surgically enhanced somehow? Not very common, but always a possibility. Technically this could be used in combination with `Cut` or `Uncut`, so maybe it's better handled as a performer tag instead since it isn't mutually exclusive from the other options.
- `Artificial`: For trans male, don't know how common these are in porn.
- `Intersex`: Formerly known as hermaphrodite. Currently included as a gender option but probably a better fit here since this feels more relevant as a genital description rather than gender identity. Often, intersex individuals have one set surgically removed. (Traditionally that's done while they're still a baby but that's less common now since their gender identity doesn't always line up with their parents' decision.) I'm not aware of any intersex porn performers that have had one set removed, but I'm not sure that situation would require anything different from the other options mentioned above. Meaning, it would still be covered by `None` or `Natural` regardless of gender identity. I'd be of the opinion that `Intersex` would only be relevant when both sets of genitals are still present, but I guess this is more of a discussion for the StashDB guidelines.

I had a discussion at one point regarding gender options in Stash or StashDB (I believe it was in DMs on Discord so I can't link to it here). One topic that came up was that in the context of porn, we're often more concerned about describing _body configurations_ (a new term for me) instead of gender identity. Having these expanded options (as well as "Breast Type" options) could help delineate gender identity from body configuration more clearly in the performer fields. Building these various combinations into Stash-Box as well may also make [this discussion for the StashDB guidelines](https://github.com/stashapp/StashDB-Docs/issues/31) a little clearer for us as well.

AdultSun:
I also mentioned on Discord, I have this framed as "Penis Type" even though it really describes "Genital Type" more broadly just because it simplifies the language for each option. In order to answer the question that "Genital Type" implies for me, I feel like the language would be muddied up as `Uncut Penis` and `Natural Vagina` rather than the more concise and direct `Uncut` and `None`, respectively. Also it feels balanced enough having a "Breast Type" field on one side and a "Penis Type" field on the other  with several shared options between them (`Natural`/`Augmented`/`Removed`/`None`) that it doesn't feel too weird of an approach to me, if that makes sense. But maybe that's just me, probably not that important either way.

departure18:
First a disclaimer, I approached this from the perspective of a gay cis-man, who mostly watches gay porn featuring (mostly) cis performers. I can't claim to have any expertise in trans matters, there's a chance might be talking nonsense here. I'm happy to be corrected :)

> My proposal would be to expand the "Circumcision" field to a more generally applicable "Penis Type" field. This mirrors an old suggestion I made expanding the "Fake Tits" field to a "Breast Type" field, originally from [this Discord discussion](https://discord.com/channels/559159668438728723/559399634318131210/849574312138375179). _(I can repost that as a separate feature request if necessary, I don't believe we have one covering it already. In short, my suggestion is using Natural/Augmented/Reduced/Removed/None as available options and replacing "Fake Tits" as the field name with "Breast Type".)_
> 
> The enum options would/could include, mostly in order of very common to least common:

[Edited the options for brevity]

That seems like a lot to describe for one enum and it doesn't seem like all the options are mutually exclusive. You've already mentioned that an "Augmented" penis could be either circumcised or not. I'd also imagine that if a trans-male was getting cosmetic surgery to give them a penis (am i right in saying that the same as "artificial" in your options?), they might have the choice for their penis to appear cut or uncut (though i refer you to my disclaimer above, i'm no expert on this!).

Wouldn't it be better to have fields for length, circumcision and type? If you want to be a completist, also girth (Girth was suggested in one of the feature requests i read, but very few sites have actual measurements of girth, if anything they just flatter their performers and just label all dicks as "Thick", so i decided not to bother with girth).

> I had a discussion at one point regarding gender options in Stash or StashDB (I believe it was in DMs on Discord so I can't link to it here). One topic that came up was that in the context of porn, we're often more concerned about describing _body configurations_ (a new term for me) instead of gender identity. 

That is a very good point! I like that concept.

> Having these expanded options (as well as "Breast Type" options) could help delineate gender identity from body configuration more clearly in the performer fields. Building these various combinations into Stash-Box as well may also make [this discussion for the StashDB guidelines](https://github.com/stashapp/StashDB-Docs/issues/31) a little clearer for us as well.

One of the reasons that i didn't include a N/A value in the circumcision enum or restrict the stats to any genders was that if a user wants to include a performer who might identify as female and also have a penis, then i don't see any reason to stop them. I realise that might not be feasible for StashDB (in a publicly editable database guidelines and restrictions are are the only way to stop it descending into chaos), but for someone cataloging their personal porn collection using Stash adding restrictions seems like a bad to me.

AdultSun:
> First a disclaimer, I approached this from the perspective of a gay cis-man, who mostly watches gay porn featuring (mostly) cis performers. I can't claim to have any expertise in trans matters, there's a chance might be talking nonsense here. I'm happy to be corrected :)

No worries, I'm certainly no expert either. I'm just trying to keep all possibilities in mind since Stash/StashDB/Stash-Box cover a very wide range of content for a large userbase that often likes to dig into these kind of specifics, so we should try to keep as many of their needs in mind as we can.

> You've already mentioned that an "Augmented" penis could be either circumcised or not. I'd also imagine that if a trans-male was getting cosmetic surgery to give them a penis (am i right in saying that the same as "artificial" in your options?), they might have the choice for their penis to appear cut or uncut (though i refer you to my disclaimer above, i'm no expert on this!).
Wouldn't it be better to have fields for length, circumcision and type?

Those last three options (`Augmented`/`Artificial`/`Intersex`) I expect will get the least amount of use so I wasn't as worried about them being mutually exclusive from `Cut`/`Uncut`. For them, it also felt like those three characteristics would likely be more important to users than the apparent circumcision status of those performers. Performer-level tags could always be used to clarify further if necessary, like `Augmented` enum + `Cut` custom performer tag. It's going to be a rare enough situation that I don't expect splitting them off into a separate enum field would be necessary, though it is an option. I expect the trade-off is just keeping the performer fields a little cleaner and a little less cluttered, probably not the biggest deal either way. We can also leave any of those three out if they don't seem useful or relevant enough, just trying to cover all possible bases with these suggestions.

> Wouldn't it be better to have fields for length, circumcision and type? If you want to be a completist, also girth (Girth was suggested in one of the feature requests i read, but very few sites have actual measurements of girth, if anything they just flatter their performers and just label all dicks as "Thick", so i decided not to bother with girth).

I only meant to point at that there are more options for penis type than just `Cut`, `Uncut`, and possibly `N/A` or `None`. I'll leave more specific design decisions to WP and the other devs. I agree that Penis Length should still be added as well and wouldn't mind Girth as an option too, but you're right that it would be difficult to source. I have seen some sub-reddits that try to make accurate estimates for porn performers' length and girth though, so some may find it useful.

> One of the reasons that i didn't include a N/A value in the circumcision enum or restrict the stats to any genders was that if a user wants to include a performer who might identify as female and also have a penis, then i don't see any reason to stop them. I realise that might not be feasible for StashDB (in a publicly editable database guidelines and restrictions are are the only way to stop it descending into chaos), but for someone cataloging their personal porn collection using Stash adding too many restrictions seems like a bad idea to me.

Sorry, I was jumping back and forth a lot between Stash and StashDB/Stash-Box since we'll want to mirror the same system on both. My proposal here is that if we're going to make these a preset list of options, we might as well make it as useful as possible for the wide range of users and performers on both Stash and StashDB. So we would still have `Cut` and `Uncut` along with anything else I thought may be relevant to our userbase. Obviously Stash users can use whichever options any way they like with no hardcoded restrictions — just like the current Gender options — and can ignore any extra options that aren't relevant to their needs. My explanations for each one are really just suggestions, any usage requirements would only be for StashDB contributions. But most of that StashDB guideline stuff is a separate discussion for another time and another place.

departure18:
> Those last three options (`Augmented`/`Artificial`/`Intersex`) I expect will get the least amount of use so I wasn't as worried about them being mutually exclusive from `Cut`/`Uncut`. For them, it also felt like those three characteristics would likely be more important to users than the apparent circumcision status of those performers. Performer-level tags could always be used to clarify further if necessary, for example `Augmented` enum + `Uncut Penis` custom performer tag. The reverse of course would also work, `Uncut` enum + `Augmented Penis` tag. It's going to be a rare enough situation that I don't expect splitting them off into a separate enum field would be necessary, though it is an option. I expect the trade-off is just keeping the performer fields a little cleaner and a little less cluttered, probably not the biggest deal either way. We can also leave any of those three out if they don't seem useful or relevant enough, just trying to cover all possible bases with these suggestions.

To me `Augmented`/`Artificial`/`Intersex` / `etc` is one aspect of a penis and `Cut`/`Uncut` is another. I feel having all those options in one field would be akin to combining all the Eye Colour and Hair Colour options into one field because all of them can be used when describing a person's head.

> I only meant to point at that there are more options for penis type than just `Cut`, `Uncut`, and possibly `N/A` or `None`. I'll leave more specific design decisions to WP and the other devs. I agree that Penis Length should still be added as well and wouldn't mind Girth as an option too, but you're right that it would be difficult to source. I have seen some sub-reddits that try to make accurate estimates for porn performers' length and girth though, so some may find it useful.

I agree that `Cut` & `Uncut` don't fully describe all the types of penises in the world. But i didn't implement a `penis_type` field, i just added `penis_length` and `circumcision` fields to cover the information available for most performers in gay porn. I never meant it to be 100% comprehensive, but think you're attempting to be comprehensive in only two fields and that's  doomed to failure. You need way more than two pieces of information to fully describe a penis. We've already mentioned `Augmented`/`Artificial`/`Intersex` / `etc` and girth, but also whether it curves up / down / left / right when hard, the shape of the glans, the veininess, the list goes on. If you want to see some prior art that attempted to comprehensively describe a penis then look at the Dick Code link in this [article](https://mashable.com/article/dick-code-sexting) (contains cartoon dicks only, not too graphic), there are a lot of options and it doesn't even include trans options. (At an early point in writing my code, i did think about adding something like the Dick Code to Stash, then i decided that might be a bit over-ambitious).

AdultSun:
I think what it comes down to is this: What should be covered by a **performer field** and what should be left to **performer tags** instead. Because something like those Dick Codes would definitely be custom performer tags and measurements like penis length would need to be a field, but everything else falls somewhere in the middle. Where to draw the line for each possible characteristic would always be up for discussion.

Devs can correct me if I'm wrong here, but I believe the value in making these preset enums rather than an open text field is to simplify Stash-Box (and maybe scraper?) integration. So the enums we choose in Stash are the ones that we would want to add to Stash-Box as well. And remember, Stash-Box doesn't have performer-level tags like Stash does, so these fields are the only data points we can attach to performers over there. (I imagine implementing performer tags in Stash-Box is a more difficult project than adding a new "Penis Type" field, so I don't expect them anytime soon.)

That's why I was trying to think ahead of how to expand your circumcision field to be as flexible as possible without going too overboard. It covers what users could find useful in Stash and Stash-Box, doesn't try to overstep with what is best left for tags (IMO), and those additional options can always be ignored if/when they aren't relevant.

I'm not totally against a separate "Circumcision" field for cut/uncut and a "Penis Type"/"Genital Type" field for everything else, but I don't think it's strictly necessary. I realize this is purely subjective, but having a field that is solely dedicated to "Circumcised? Yes or No" feels like a bit of a waste of real estate in the UI. The current "Fake Tits" field feels the same to me, which is why I have that separate proposal for expanding it as well. Many users will still use either of those fields as essentially `Yes`/`No`/`NA`, but at least there would be some additional options built-in with the potential for proper Stash-Box integration. It also doesn't pigeon-hole an entire field into, "This is just for a particular type of performer/user," and instead tries to provide more flexibility and versatility to a wider range of users and performers. Stash and Stash-Box are designed to be very open and customizable, so hard-coded fields that are only relevant to a comparatively small-slice of the userbase (only straight male users, only cis performers, only "professional" content, etc.) just feels... *weird* I guess. It feels to me like a mismatch in design philosophy, if that makes sense.

I don't know, I also feel like I've spent far too much time rambling in these comments already. I usually lean towards, "If it's worth doing, it's worth over-doing," but that also gets me in trouble sometimes when I overthink something that should be simple. So I'll leave it at that for now.

DogmaDragon:
> I think what it comes down to is this: What should be covered by a **performer field** and what should be left to **performer tags** instead. Because something like those Dick Codes would definitely be custom performer tags and measurements like penis length would need to be a field, but everything else falls somewhere in the middle. Where to draw the line for each possible characteristic would always be up for discussion.

Probably unpoplar opinion, but anything that is reasonably available on 3rd party databases/studio sites/modelling agencies should be a performer field. It also doesn't prevent people who prefer tags over fields to move that information to tags, but not having fields would have adverse effect. 
Tags are great for some, but doesn't work for everyone as they can conflict with existing tag schemes people use. 

In general I have nothing to offer as to specific enums, but I would be in favor of more than less.

departure18:
_[This is going to be confusing! I wrote a reply to AdultSun's second last [post](https://github.com/stashapp/stash/pull/3627#issuecomment-1555408070) last night, accidentally hit the comment button through tiredness, immediately deleted it and decided to sleep before posting it. I'm guessing AdultSun's last [post](https://github.com/stashapp/stash/pull/3627#issuecomment-1556057146) is a reply to my post that no longer exists. So imagine the following text is a couple of posts higher in this thread!]_

> Those last three options (`Augmented`/`Artificial`/`Intersex`) I expect will get the least amount of use so I wasn't as worried about them being mutually exclusive from `Cut`/`Uncut`. For them, it also felt like those three characteristics would likely be more important to users than the apparent circumcision status of those performers. Performer-level tags could always be used to clarify further if necessary, for example `Augmented` enum + `Uncut Penis` custom performer tag. The reverse of course would also work, `Uncut` enum + `Augmented Penis` tag. It's going to be a rare enough situation that I don't expect splitting them off into a separate enum field would be necessary, though it is an option. I expect the trade-off is just keeping the performer fields a little cleaner and a little less cluttered, probably not the biggest deal either way. We can also leave any of those three out if they don't seem useful or relevant enough, just trying to cover all possible bases with these suggestions.

To me `Augmented`/`Artificial`/`Intersex` / `etc` is one aspect of a penis and `Cut`/`Uncut` is another. I feel having all those options in one field would be akin to combining all the Eye Colour and Hair Colour options into one field because all of them can be used when describing a person's head.

> I only meant to point at that there are more options for penis type than just `Cut`, `Uncut`, and possibly `N/A` or `None`. I'll leave more specific design decisions to WP and the other devs. I agree that Penis Length should still be added as well and wouldn't mind Girth as an option too, but you're right that it would be difficult to source. I have seen some sub-reddits that try to make accurate estimates for porn performers' length and girth though, so some may find it useful.

I agree that `Cut` & `Uncut` don't fully describe all the types of penises in the world. But i didn't implement a `penis_type` field, i just added `penis_length` and `circumcision` fields to cover the information available for most performers in gay porn. I never meant those two fields to be 100% comprehensive, there's definitely scope for more information. But think attempting to be comprehensive in only two fields is doomed to failure. You need way more than two pieces of information to fully describe a penis. We've already mentioned `Augmented`/`Artificial`/`Intersex` / `etc` and girth, but also whether it curves up / down / left / right when hard, the shape of the glans, the veininess, the list goes on. If you want to see some prior art that attempted to comprehensively describe a penis then look at the Dick Code link in this [article](https://mashable.com/article/dick-code-sexting) (contains cartoon dicks only, not too graphic), there are a lot of options and it doesn't even include trans options. (At an early point in writing my code, i did think about adding something like the Dick Code to Stash, then i decided that might be a bit over-ambitious).

departure18:
> I think what it comes down to is this: What should be covered by a **performer field** and what should be left to **performer tags** instead. Because something like those Dick Codes would definitely be custom performer tags and measurements like penis length would need to be a field, but everything else falls somewhere in the middle. Where to draw the line for each possible characteristic would always be up for discussion.

As i said, i thought implementing Dick Codes in Stash would be overkill, so i implemented the two fields that are on practically every gay porn site, length and circumcision. I agree with @DogmaDragon's unpopular opinion that "anything that is reasonably available on 3rd party databases/studio sites/modelling agencies should be a performer field". So i made circumcision a field rather than using tags.  

> That's why I was trying to think ahead of how to expand your circumcision field to be as flexible as possible without going too overboard. It covers what users could find useful in Stash and Stash-Box, doesn't try to overstep with what is best left for tags (IMO), and those additional options can always be ignored if/when they aren't relevant.

This is the thing, i implemented a circumcision field, you can only ever be circumcised or uncircumcised. Since it's a bit rude to check someone's genitals without consent, it's good idea to also have an unknown value. Some people don't have dicks so setting a value is irrelevant, this could be an `N/A` value, but i chose not to do that and bundle `N/A` with `unknown` for two reasons: 
 - I didn't want to make everyone with a Stash filled with predominantly cis-female performers have to spend time setting the value manually to `N/A` on the vast majority of their performers or have "Penis: N/A" displayed on a performer's page for no good reason.
 - I really didn't want to make assumptions on which performers might have a dick based on gender or any other factors because that could cause offense.

I saw on Discord that @WithoutPants made the valid point that "it's not clear whether a null value is missing data or not applicable data". When i was testing it, if i was looking for missing data, i just used this filter:
filter
 It works for my collection because it's mostly comprised of cisgender males and a few cis-girls who have crept in because some of my favourite male performers have done gay and straight porn. I think combining the Gender filter and setting "Is Missing" for circumcised to find missing data should work for most people without the software making assumptions about which genders can own which genitals and is less effort than setting `N/A` on every cis-girl in people's collections. That said if someone has a collection with lots of pre-op trans performers, who identify as purely as male or female, finding missing data will be harder using that approach. 

I think an `N/A` value is more justified on StashDB/Stash-Box (because guidelines stop edit wars and so on), but when a user uploads a performer as a draft they can add it at that point. When a performer is scraped from StashDB, i think an `N/A` should be translated to `null`, so the "Penis:" field isn't displayed if it's not relevant. For example here's a cis-female performer from my collection:
female
No mention of the penis that she doesn't have. 
However if i scrape a guy from StashDB:
male
I really don't need to be reassured that he doesn't have fake tits, so i usually delete the `Na` before saving. I just think it's better just not to display irrelevant data.

> I'm not totally against a separate "Circumcision" field for cut/uncut and a "Penis Type"/"Genital Type" field for everything else, but I don't think it's strictly necessary. I realize this is purely subjective, but having a field that is solely dedicated to "Circumcised? Yes or No" feels like a bit of a waste of real estate in the UI. 

Ok. You might feel that it's a waste of real estate, but circumcision status is important enough that pretty much every gay porn site will have that data and most will let you filter performers based on that criteria. 

I've added two new fields on the edit page. In order to accommodate the wishes of other people to record whether a penis `exists` / `is augmented` / `is a result of gender affirmation surgery` / `etc`, i've suggested a third field. It doesn't seem excessive. In my own efforts to save real estate in the UI, if both of my two fields aren't set then i don't display the "Penis:" field on the performer information page at all (as you can see in the screenshots above).

I think Stash should be as inclusive as possible, which is why i added the ability to store some information that's important to consumers of gay porn. I think i can speak with some authority on gay porn, but i wouldn't claim to have more than a passing knowledge of trans issues so i wouldn't try to speak with authority on that. I think you need to speak to someone with a much better knowledge of porn featuring a wide variety of trans-performers before coming up with a full categorisation. Finding that person might take some time though. Nothing i've added would prevent someone who has a better knowledge of trans performers (or surgically augmented performers or whatever) from adding categorisations that are important to their porn, but it seems likely that circumcision status would also be applicable to those performers too, so should be a separate field. 

> I don't know, I also feel like I've spent far too much time rambling in these comments already. I usually lean towards, "If it's worth doing, it's worth over-doing," but that also gets me in trouble sometimes when I overthink something that should be simple. So I'll leave it at that for now.

You and me both, i've spent far more time defending the design of this code than i spent writing it. And i haven't even gotten to the point where everyone finds the bugs yet!

WithoutPants:
I've done a bit of refactoring of the code. Included was changing the filtering code to allow for selecting a modifier to be selected in the UI. This meant that the new fields don't need to be added to the is missing filter, so they were removed from there.

I've also changed the filtering code to allow multiple values to be chosen. This is a bit of a superfluous change since there's only two possible values, but this can be generalised to other enum filters in future. I also changed the existing enum filters which accept single values to use radio buttons instead of checkboxes.

departure18:
Played around with your changes this afternoon, it looks good to me. I had stored penis length and circumcision in the measurement field as an interim measure, so i wrote a quick script to copy that data in to the new fields and everything seems to work.

Found a bug in the gender implementation that i've copied over into the circumcision code. If you edit one or more performers using the bulk edit dialog on the main performer screen it will fail if all of the selected performers have either Gender or Circumcision unset. I attempted a fix, but i'm not sure if my Typescript/React skills are up to it, my best attempt looked appallingly hacky. So i've opened a bug report here #3761

WithoutPants:
I've added a fix to address the enum bug. This is good to merge.

PR #3628, opened 2023-04-01
Fix queries referencing non-existent tables

wethekinky:
`performer_tags` and `marker_tags` tables do not exist but `performers_tags` and `scene_markers_tags` do

While working on something in python to help write queries as code easier, graphql returned an error when using the `AND` operator on the `performer_tags` criterion for scenes. The following query  

```gql
{
  findScenes(
    filter: {per_page: 1}
    scene_filter: {AND: {performer_tags: {modifier: INCLUDES, value: ["1"]}}}
  ) {
    scenes {
      id
    }
  }
}
```

results in an error

```json
{
  "errors": [
    {
      "message": "error finding IDs: running query: SELECT DISTINCT scenes.id FROM scenes LEFT JOIN performer_tags ON performer_tags.scene_id = scenes.id WHERE ((performer_tags.root_tag_id IS NOT NULL)) LIMIT 1 OFFSET 0  [[]]: error executing `SELECT DISTINCT scenes.id FROM scenes LEFT JOIN performer_tags ON performer_tags.scene_id = scenes.id WHERE ((performer_tags.root_tag_id IS NOT NULL)) LIMIT 1 OFFSET 0 ` [[]]: no such table: performer_tags",
      "path": [
        "findScenes"
      ]
    }
  ],
  "data": null
}
```

while the equivalent query without the `AND` 

```gql
{
  findScenes(
    filter: {per_page: 1}
    scene_filter: {performer_tags: {modifier: INCLUDES, value: ["1"]}}
  ) {
    scenes {
      id
    }
  }
}
```
returns successfully

```json
{
  "data": {
    "findScenes": {
      "scenes": []
    }
  }
}
```

Admittedly I haven't dug deep enough into the source to know if there's something else going on here but it will definitely take me a while to fully understand the db code and get a development env setup so I figured I would offer this up for review in the meantime.

### DIFF
--- a/pkg/sqlite/scene.go
+++ b/pkg/sqlite/scene.go
@@ -1369,15 +1369,15 @@ func scenePerformerTagsCriterionHandler(qb *SceneStore, tags *models.Hierarchica
 
 			valuesClause := getHierarchicalValues(ctx, qb.tx, tags.Value, tagTable, "tags_relations", "", tags.Depth)
 
-			f.addWith(`performer_tags AS (
+			f.addWith(`performers_tags AS (
 SELECT ps.scene_id, t.column1 AS root_tag_id FROM performers_scenes ps
 INNER JOIN performers_tags pt ON pt.performer_id = ps.performer_id
 INNER JOIN (` + valuesClause + `) t ON t.column2 = pt.tag_id
 )`)
 
-			f.addLeftJoin("performer_tags", "", "performer_tags.scene_id = scenes.id")
+			f.addLeftJoin("performers_tags", "", "performers_tags.scene_id = scenes.id")
 
-			addHierarchicalConditionClauses(f, tags, "performer_tags", "root_tag_id")
+			addHierarchicalConditionClauses(f, tags, "performers_tags", "root_tag_id")
 		}
 	}
 }
--- a/pkg/sqlite/gallery.go
+++ b/pkg/sqlite/gallery.go
@@ -1003,15 +1003,15 @@ func galleryPerformerTagsCriterionHandler(qb *GalleryStore, tags *models.Hierarc
 
 			valuesClause := getHierarchicalValues(ctx, qb.tx, tags.Value, tagTable, "tags_relations", "", tags.Depth)
 
-			f.addWith(`performer_tags AS (
+			f.addWith(`performers_tags AS (
 SELECT pg.gallery_id, t.column1 AS root_tag_id FROM performers_galleries pg
 INNER JOIN performers_tags pt ON pt.performer_id = pg.performer_id
 INNER JOIN (` + valuesClause + `) t ON t.column2 = pt.tag_id
 )`)
 
-			f.addLeftJoin("performer_tags", "", "performer_tags.gallery_id = galleries.id")
+			f.addLeftJoin("performers_tags", "", "performers_tags.gallery_id = galleries.id")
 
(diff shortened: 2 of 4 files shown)


pkg/sqlite/scene.go
```diff
@@ -1369,15 +1369,15 @@ func scenePerformerTagsCriterionHandler(qb *SceneStore, tags *models.Hierarchica
 
 			valuesClause := getHierarchicalValues(ctx, qb.tx, tags.Value, tagTable, "tags_relations", "", tags.Depth)
 
-			f.addWith(`performer_tags AS (
+			f.addWith(`performers_tags AS (
```
hphpanon, line 1372: I'm by no means a SQL expert so I could be 1000% wrong on all of this but:

Maybe I'm misunderstanding the original intent in this query, but isn't `performer_tags` here an alias of the query in parens as opposed to a reference to the table? Note that the reference to the table on line 1374 is to the correct table name. Now once we get to the bottom two changes in this file I'd expect/assume the `addLeftJoin` change is correct (unless you can join on an aliased sub-query?) and the `addHierarchicalConditionClauses` I'd have to coin flip without immersing myself in this codebase.

dogwithakeyboard:
@hphpanon is correct; without the alias it leads to circular reference error.

```
{
  "errors": [
    {
      "message": "error finding IDs: running query: WITH performers_tags AS (\nSELECT ps.scene_id, t.column1 AS root_tag_id FROM performers_scenes ps\nINNER JOIN performers_tags pt ON pt.performer_id = ps.performer_id\nINNER JOIN (VALUES(56,56)) t ON t.column2 = pt.tag_id\n) SELECT DISTINCT scenes.id FROM scenes LEFT JOIN performers_tags ON performers_tags.scene_id = scenes.id LIMIT 25 OFFSET 0  [[]]: error executing `WITH performers_tags AS (\nSELECT ps.scene_id, t.column1 AS root_tag_id FROM performers_scenes ps\nINNER JOIN performers_tags pt ON pt.performer_id = ps.performer_id\nINNER JOIN (VALUES(56,56)) t ON t.column2 = pt.tag_id\n) SELECT DISTINCT scenes.id FROM scenes LEFT JOIN performers_tags ON performers_tags.scene_id = scenes.id LIMIT 25 OFFSET 0 ` [[]]: circular reference: performers_tags",
      "path": [
        "findScenes"
      ]
    }
  ],
  "data": null
}
```
There's something buggy with the operators for performer tags because this works:

```
{
  findScenes(
    filter: {per_page: -1}
    scene_filter: {
      performer_tags: {modifier:INCLUDES,value: ["1"]}
      AND: {performers: {modifier:INCLUDES,value:["1"]}}
    }
    							
  ) {count
    scenes {
      id
    }
  }
}
```
And this fails:

```
{
  findScenes(
    filter: {per_page: -1}
    scene_filter: {
      performers: {modifier: INCLUDES, value: ["1"]}
      AND: {performer_tags: {modifier:INCLUDES,value:["1"]}}
    }
    							
  ) {count
    scenes {
      id
    }
  }
}
```
An implicit AND is fine:

```
{
  findScenes(
    filter: {per_page: -1}
    scene_filter: {
      performers: {modifier:INCLUDES,value: ["1"]}
      performer_tags: {modifier:INCLUDES,value:["1"]}
    }
    							
  ) {count
    scenes {
      id
    }
  }
}
```

wethekinky:
Thanks guys, I'll close this out and open an issue instead

PR #3636, opened 2023-04-04
Adds videojs-vr support

Teda1:
This request is based on the `proto-vr` branch, which proposes a vr button users can use to toggle projections. Currently, only the 360 projection option will be selectable until the `videojs-vr` developers update their code to use updated Threejs APIs. This pull request also addresses the projection canvas disappearing issue whenever a tab is clicked.

I'm considering having a setting option that will allow users to select which of their tags should trigger the VR button to appear on the videojs player. Currently, the VR button is available when watching any video, even non-VR videos, which may be annoying for some users.


ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -264,6 +266,17 @@ export const ScenePlayer: React.FC<IScenePlayerProps> = ({
...
   useEffect(() => {
+    function isVrScene() {
+      if (!scene?.id || !vrTag) return false;
+
+      scene?.tags.map((tag) => {
+        if (vrTag == tag.name) {
+          return true;
+        }
+      });
+      return false;
```
WithoutPants, line 277: This currently returns false every time.

```suggestion
      return scene?.tags.some((tag) => {
        if (vrTag == tag.name) {
          return true;
        }
      });
```

WithoutPants (changes requested):
I can't push to your branch due to `permission denied` errors.

Since merging from the develop branch, we lost the `crossorigin` attribute. The following needs to be added in at `ScenePlayer.tsx:341`:
```
videoEl.setAttribute("crossorigin", "anonymous");
```

With this and the other suggested fix, it worked successfully for me.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
Teda1: Ah, I see. I've gone ahead and pushed your suggested changes.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -268,13 +268,16 @@ export const ScenePlayer: React.FC<IScenePlayerProps> = ({
...
       scene?.tags.map((tag) => {
         if (vrTag == tag.name) {
-          return true;
+          hasVrTag = true;
+          return;
         }
       });
-      return false;
+
+      return hasVrTag;
```
WithoutPants: If you replace this with `some` per my previous review, it doesn't have to iterate through all of the tags. `map` is used incorrectly here.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
Teda1: You know I completely missed the suggested change you had here. I must have been multitasking when I looked at this that day.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
Teda1: I'm just now seeing your comment. I'm unsure how I missed it when I checked GitHub the day after.

WithoutPants (approved):

PR #3645, opened 2023-04-06
Adds ability to configure sort order for DLNA videos

### LINKED ISSUE #2166: [Feature] Add DLNA "Random" and "Last Added (Updated at)" categories
Being able to sort scenes by random or last-added in the web-interface scenes page is a great feature that can save you a lot of time, however in DLNA interface which is what I mostly use the scenes are always sorted alphabetically in  the "all" category, and my player doesn't have the option to change the sorting method.
Is it posible to add that as a "random" and "last added" categories to DLNA, and maybe give user an ability disable unwanted categories for DLNA?

Teda1:
This pull request adds the ability to configure the default sort order videos I'm DLNA.

Fixes #2166.

### DIFF
--- a/graphql/documents/data/config.graphql
+++ b/graphql/documents/data/config.graphql
@@ -99,6 +99,7 @@ fragment ConfigDLNAData on ConfigDLNAResult {
   enabled
   whitelistedIPs
   interfaces
+  videoSortOrder
 }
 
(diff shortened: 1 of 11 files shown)


kermieisinthehouse (approved):
Code reviewed but not tested.

PR #3648, opened 2023-04-08
Fix latest version error

DingDongSoLong4:
This is a fix for 'invalid Github API response' errors when trying to retrieve the latest version on a development build.

The cause is that there are now over 30 tags in the repo, which is the default page size of the GitHub API tags endpoint used to find the latest release hash. The results for this endpoint seem to be sorted in reverse order, so the `latest_develop` tag will always be the last one in the list. I couldn't find a better way, so we have to search through the entire list.

I've set an upper limit of 500 tags to retrieve, so this will break if/when stash hits such a tag count, but I don't think that will happen any time soon.

### DIFF
--- a/internal/api/check_version.go
+++ b/internal/api/check_version.go
@@ -113,7 +113,6 @@ type LatestRelease struct {
 }
 
 func makeGithubRequest(ctx context.Context, url string, output interface{}) error {
-
 	transport := &http.Transport{Proxy: http.ProxyFromEnvironment}
 
 	client := &http.Client{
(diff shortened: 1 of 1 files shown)


WithoutPants:
I'm thinking that given there's no easy way to get the version hash, we should probably deprecate and drop getting the hashes at all. We'll need to refactor how we check for new versions.

Having to potentially go through five pages of data just to get the sha doesn't seem worth it to me.

MrX292:
could using https://api.github.com/repos/stashapp/stash/releases/tags/latest_develop for develop and https://api.github.com/repos/stashapp/stash/releases/latest for latest work

DingDongSoLong4:
Agreed, although it will still be on the first page until we hit 100 tags, the 5 pages was just to be extra sure.

The release hash could just be parsed directly from the release name, which is already being parsed to find the version name anyway. It wouldn't be the entire hash, but it should still work fine to detect a new version. 
That would mean that the prerelease name format can't be drastically changed in the future though, or it will break.

PR #3650, opened 2023-04-08
Use big-buttons instead of videojs-mobile-ui touch controls

### LINKED ISSUE #3649: [UI/UX] Player touch controls UI regressions
I mainly use Stash on mobile, specifically on an Android phone and an iPad, very rarely on desktop. There were several UI/UX regressions introduced in 0.18.0 with #3120.
The problems stem from badly implemented touch controls in the newly added [videojs-mobile-ui](https://github.com/mister-ben/videojs-mobile-ui) dependency:
* The play button is decoupled from the control bar. There were previously only two states for the control bar and buttons: both visible or both invisible. Now, there are three:
    * control bar visible, play button visible
    * control bar visible, play button invisible
    * control bar invisible, play button invisible
      This is is a weird and annoying design choice that none of the mobile video players use: Chrome for Android, Firefox for Android, Chrome for iOS, Safari for iOS, native and web YouTube player.
* The play button has its own animation you have to wait for, making it harder to quickly pause a video.
* The play button has a way too tall clickable area, making accidental plays/pauses very likely, especially when playing a vertical video on mobile.
* The control bar now pops up on touchstart events as well, which is annoying for devices with touch screens because of the increased probability of accidental touches.
* The double tap to seek feature gets triggered not just on click events, but also on touchstart events, making accidental seeks more likely.

I suggest disabling videojs-mobile-ui's touch controls and reverting to using big-buttons. I understand we lose the double tap to seek feature by doing this, however I'd argue that we don't lose much on the UI/UX front. The big seek buttons are just as convenient, if not more, because you don't have to tap so many times. The only advantage double tap to seek has over them is the ability to seek without covering up the video with the control bar and play buttons. However, videojs-mobile-ui's implementation doesn't even do that correctly as it causes the control bar to pop up anyway.

Contributing to videojs-mobile-ui to fix the issues there.

**Additional context**
play button too big clickable area

https://user-images.githubusercontent.com/95527094/230716699-e7791dd0-06de-4821-8e52-5ccd3e7f0c4a.mp4

https://user-images.githubusercontent.com/95527094/230716702-aca44d2c-8d6b-4a38-8d6a-860261850fda.mp4

dumdum7:
Reverts to using previously deleted big-buttons.ts instead of videojs-mobile-ui's touch controls.
Fixes #3649.


AnonumousMouse:
Is there any way to toggle this or undo it? Not having double tap to forward on mobile is kind of insane for a video player ui. There were ways to fix the ui issues (that I never even noticed) without removing double tap. At the very least allow us an option to have it always use the default html5 player which has double-tap to seek.

PR #3652, opened 2023-04-08
Player touch control improvements

DingDongSoLong4:
This is a bit of CSS trickery to fix some of the issues pointed out in #3649, while still keeping double tap to seek enabled. The big play/pause button is now coupled to the control bar below, and its clickable area is a whole lot smaller. I've also removed some of the unused styles left over from the big-buttons plugin.

I believe this fixes the first three issues in #3649, but not the last two, relating to the touchstart event. That would require either disabling videojs-mobile-ui touch controls and using big-buttons again (ie #3650) or adding a custom replacement plugin to do the same job.

I don't use mobile all that often, and I don't really have a preference for double tap to seek over the old big buttons, so if the touchstart issue is deemed big enough then feel free to merge #3650 and not this.

I do think the better long-term solution would be to add a custom plugin and ditch videojs-mobile-ui entirely though, so that we can have both big buttons and double tap to seek. I'll get around to doing that eventually.

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/styles.scss
+++ b/ui/v2.5/src/components/ScenePlayer/styles.scss
@@ -34,28 +34,6 @@ $sceneTabWidth: 450px;
     outline: none;
   }
 
-  .vjs-big-button-group {
-    display: none;
-    height: 80px;
-    justify-content: space-around;
-    opacity: 0;
-    position: absolute;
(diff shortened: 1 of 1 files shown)


DingDongSoLong4:
Closing in favour of #3650. Will revisit to add a custom plugin to replace videojs-mobile-ui entirely at some later stage.

PR #3653, opened 2023-04-09
Add hint for alias duplicates

### LINKED ISSUE #3639: [UI/UX] "Alias must be unique" notification --> Denote which aliases using color
When there is a duplicate alias, the app notifies the user that the "alias must be unique". However, it does not specify which aliases. When there are a long list of aliases (often very similar variations) it can be cumbersome to find which is the duplicate alias.

Highlight which aliases are the duplicates, either by making the alias font red, or the input background red - something obvious.

Automatically removing the duplicate aliases without user intervention

**Additional context**
Screenshot 2023-04-04 161335

Also a tiny nitpick modification in the above image. Added vertical padding between the red buttons and made the remove button the same vertical size as the alias input form.

DingDongSoLong4:
This is an implementation for the suggestion in #3639, using the preexisting `is-invalid` class to display a error icon.

I've also fixed the overflow issue with the remove buttons, decreased the spacing above the 'aliases must be unique' message, and increased the spacing between the alias entries.

And then I've also added two unrelated style improvements:
- A tweak to the country select dropdown to show a border on hover, whose removal was causing the contents to move to the left by 1px on hover. All the other dropdowns show a border on hover so I don't know why it was removed for the country field specifically.
- Improved alignment for the left and right arrows in the date picker. It bothered me that they were really far up and not at all aligned with the line showing the selected month and year. The exact alignment is dependent on the actual fonts used on the system, so it doesn't line up perfectly everywhere, but I think it's better now than before.

Closes #3639

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
@@ -104,14 +104,19 @@ export const PerformerEditPanel: React.FC<IPerformerDetails> = ({
       .test({
         name: "unique",
         test: (value, context) => {
-          if (!value) return true;
-          const aliases = new Set(value);
-          aliases.add(context.parent.name);
-          return value.length + 1 === aliases.size;
+          const aliases = [context.parent.name, ...value];
+          const dupes = aliases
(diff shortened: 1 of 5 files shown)


kermieisinthehouse:
Tested and code reviewed.

PR #3663, opened 2023-04-14
Limit duplicate matching to files that have ~ same duration

puc9:
Add the ability to look for PHash duplicates that have similar duration.
It works independent on the PHash duration.
No UI but you could add durationDiff parameter in the URL for sceneDuplicateChecker like this:
image
The durationDiff parameter represent seconds and it is a float.
Default max allowed difference is 1s,
To disable checking that the duration also match pass a negative value to durationDiff parameter.
But I am honestly curious, can someone identify a case where you would want to match by PHash even when the durations are different?


DingDongSoLong4:
Very useful addition, it's unlikely that someone would want to include duplicates with wildly different durations but this does break backward compatibility for the `findDuplicateScenes` query if you don't include the new parameter. I think it would be better to have the `durationDiff` default to -1 on the backend and instead have the UI duplicate finder send a non-negative default `durationDiff` if the query parameter is missing.

I think the UI default should then probably also be more than 1 second - I've come across many older scenes that have been re-released with a new intro/outro, meaning that the lengths can be like 10 seconds off but the phashes are still very similar.

I would also make a suggestion to add a percentage-based filter if possible (ie the durations are within x% of each other) in addition to the absolute time difference - that could help with preventing shorter scenes from showing too many duplicates if you want to set a large `durationDiff`.

puc9:
> Very useful addition, it's unlikely that someone would want to include duplicates with wildly different durations but this does break backward compatibility for the `findDuplicateScenes` query if you don't include the new parameter. I think it would be better to have the `durationDiff` default to -1 on the backend and instead have the UI duplicate finder send a non-negative default `durationDiff` if the query parameter is missing.

Before I make the change, which is trivial, just adding a `-` in code I wonder if it is worth keeping the backwards capability? I did consider making the default be backwards compatible but to me the fact that a PHash matches scenes with very different timestamps it's more of a bug than a feature. I do not care that much one way or another but I don't actually see why we should keep it backwards compatible. Again, not a strong opinion on this, and I can definitely make the change.

> I think the UI default should then probably also be more than 1 second - I've come across many older scenes that have been re-released with a new intro/outro, meaning that the lengths can be like 10 seconds off but the phashes are still very similar.

On this point I do in fact disagree. I do not think that dup finder, in its default configuration, should find scenes that are different, even just by intro/outro. The initial version of the code had the 1s hardcoded in the SQL query but I added the parameter specifically to be able to increase the number of potential duplicates found. It's the same reason I suspect that the default value for distance is 0.

> I would also make a suggestion to add a percentage-based filter if possible (ie the durations are within x% of each other) in addition to the absolute time difference - that could help with preventing shorter scenes from showing too many duplicates if you want to set a large `durationDiff`.

I think it is harder to intuitively make sense of a percentage value for non mathematically inclined. I have been using this code for about 1/2 year (I postponed adding the parameter and that's why I didn't create a PR) with the 1s difference and I found it to be extremely good at filtering out false positives. I use stash to sort non-adult content as well and that content includes clips that are 2s-10s long and the 1s has proven to be useful enough. Also, trying to keep it simple.

Ideally, having a UI for this parameter will solve most of the issues with the default value for `durationDiff` but I'm not good at UI and JS and it would have taken far more time to include that in this PR as well. It could be added later by me or someone more UI skilled.

In any case, thank you for your comment.

DingDongSoLong4:
I don't use the duplicate finder all that often, but when I do it is to find re-released scenes that are slightly different and for that I need to use the accuracy option (ie changing the `distance` parameter). That's why I mentioned scenes with differing lengths - the duplicate finder does let you find those duplicates, and I do think that is a valid use case for it.

If you only ever use the duplicate finder to find complete duplicates then forcing the scenes to have (identical) length makes sense, but if you want to find scenes that are "close enough" by decreasing the accuracy option then you don't want to force the lengths to be almost identical.

So no I don't think it's a bug that phashes can match duplicate scenes with different lengths, and if someone has written some external tool that uses this functionality I don't think it's worth breaking it, especially since the default 1 second is also specified in the UI anyway.

For now I think leaving the default difference at 1 second is fine, as finding complete duplicates are likely what most people are using the duplicate finder for. The solution would then be a UI, which can be added later.

>
> > I would also make a suggestion to add a percentage-based filter if possible (ie the durations are within x% of each other) in addition to the absolute time difference - that could help with preventing shorter scenes from showing too many duplicates if you want to set a large `durationDiff`.
> 
> I think it is harder to intuitively make sense of a percentage value for non mathematically inclined. I have been using this code for about 1/2 year (I postponed adding the parameter and that's why I didn't create a PR) with the 1s difference and I found it to be extremely good at filtering out false positives. I use stash to sort non-adult content as well and that content includes clips that are 2s-10s long and the 1s has proven to be useful enough. Also, trying to keep it simple.
> 

Understandable, filtering by percentage would be more complicated than an absolute time difference and likely less useful to most people. No need to add it here, that can be done separately if the need arises.

puc9:
I have ~140k clips across 2 different libraries for different kinds of content and my duplicate count was at 1 point in time ~20k. Yes, 20.000 😊
The duplicate finder and the https://github.com/stashapp/CommunityScripts/tree/47639382072a3a4c01e2e47eb9a37c35a1e8450d/plugins/phashDuplicateTagger plugin are the tools I use most often.
I will however revert the default in the graphql, will check it in soon, because, as you pointed out, the UI default stays and that is what most people would use.

PR #3674, opened 2023-04-18
Use chevron icons instead of < and >

trashcom:
This PR uses chevron icons instead of the less-than/greater-than symbols for both the scene scrubber and the collapsable scene pane.

Please see the before and after for a visual comparison.

chevrons-before-after


kermieisinthehouse (approved):

kermieisinthehouse:
Thank you for your contribution!

We'll be glad to include this in the next release.

DingDongSoLong4:
A small, but very welcome change indeed.

Would you perhaps also be able to make this change to the divider on the other pages with collapsible sidebars (galleries, performers, tags, studios)?

trashcom:
> Would you perhaps also be able to make this change to the divider on the other pages with collapsible sidebars (galleries, performers, tags, studios)?

Good point! I just pushed changes for the other collapsible sidebars.

PR #3675, opened 2023-04-20
Pinned Filters

### LINKED ISSUE #3549: [Feature] Custom Filter Order
I like the new UI Filter very much and think it's the right way to go.
However, one problem that isn't solved by this structure is the amount of possible filters that are still presented when you are trying to add a filter.
Although an alphabetical order is a good way for categorically searching a certain way to filter, it falls short in terms of sorting by the actual relevance of a certain filter. 
For example, filtering by a studio, performer or tag is very likely the most prevalent feature presented here, while filtering by checksum is at the very least a very rare occasion. Still, checksum is one of the first options present, while you have to search for the others for a bit.

There are some solutions for that:
- For example, you could customize the order in the settings.
- You could also favorite certain filters, which are then pinned at the top and maybe separated by a small divider.
- You could also hardcode a more convenient order, however I think that is very subjective so I would recommend one of the other two options to provide more flexibility.

I would personally lean towards the second solution. It's probably the easiest to implement, relatively intuitive without making the settings more bloated and still provides an easier way to access the more common filters.

yoshnopa:
As Proposed in #3549, this Pull Provides the possibility to pin filters to the top of the Filter list. 
That way, more commonly used filters are available more quickly without compromising on the wide range of possible filter options. 
In the backend, this works very similar to the savedfilter, just without the actual filter and the messageid of the filter as the "name". 
In the frontend, the only directly visible change is a pin on the right side of the filters:
image
Clicking on the pin will put the according option to the top, separated by a divider:
image
Of course, the pinned options are saved per media type, so you can have different pinned filters in the scene section than in the tag section.

Closes #3549


kermieisinthehouse:
Thank you for contributing to Stash! This is a great change.

Functionality tested; working.
Code reviewed.

My only note is to not lay it out with a `right: 15%`, but instead to make the whole row a flexbox, and make the caption `flex-grow` to take up the entire middle. That way, we don't have weird layout and the button position is determined by the padding of the parent row.

yoshnopa:
Thanks @DingDongSoLong4 for the Code Revision and additions!
Just for documentation, the following was changed: 

- Divider is now just empty space, as we don't use such a thing elsewhere
- On hover, the pin rotates, giving a visual cue for the user
- The x and pin swapped places, it looks fancier. I'm open for debate on that one
- Flex is used for positioning. Generally the Code has been polished from my "let's try some react!"-state

kermieisinthehouse (approved):
Code reviewed and tested.

PR #3682, opened 2023-04-25
Updated the English auto_tag_based_on_filenames message

BawdyInkSlinger:
- I included "filepaths" in the description. By leaving that out, it implied that only file names were used

### DIFF
--- a/ui/v2.5/src/locales/en-GB.json
+++ b/ui/v2.5/src/locales/en-GB.json
@@ -405,7 +405,7 @@
         "auto_tagging_all_paths": "Auto Tagging all paths",
         "auto_tagging_paths": "Auto Tagging the following paths"
       },
-      "auto_tag_based_on_filenames": "Auto-tag content based on filenames.",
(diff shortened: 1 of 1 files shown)


WithoutPants:
`Auto-tag content based on file paths` is sufficient.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -405,7 +405,7 @@
         "auto_tagging_all_paths": "Auto Tagging all paths",
         "auto_tagging_paths": "Auto Tagging the following paths"
       },
-      "auto_tag_based_on_filenames": "Auto-tag content based on filenames.",
+      "auto_tag_based_on_filenames": "Auto-tag content based on filepaths and filenames.",
```
kermieisinthehouse, line 408:
```suggestion
      "auto_tag_based_on_filenames": "Auto-tag content based on file paths and names.",
```

WithoutPants: I'm not quite sure why you didn't go with my suggestion; file paths and names is redundant.

PR #3684, opened 2023-04-26
More concise auto tagging docs

BawdyInkSlinger:
- I made up the examples on line 13; I'm not sure if the ones with `/` are true.

### DIFF
--- a/ui/v2.5/src/docs/en/Manual/AutoTagging.md
+++ b/ui/v2.5/src/docs/en/Manual/AutoTagging.md
@@ -1,10 +1,8 @@
 # Auto Tagging
 
-This task matches your Performers, Studios, and Tags against your media, based on names only. It finds Scenes, Images, and Galleries where the path or filename contains the Performer/Studio/Tag. 
+When media filepaths or filenames contain a Performer, Studio, or Tag name, it is assigned those Performers, Studios, and Tags. It will **only** tag based on Performer, Studio, and Tag names that already exist in your database.
 
-For each scene it finds that matches, it sets the applicable field. It will **only** tag based on performers, studios, and tags that already exist in your database. In order to completely identify and gather information about the scenes in your collection, you will need to use the Tagger view and/or Scraping tools.
-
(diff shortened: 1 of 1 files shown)


WithoutPants:
Either the `/` cases should be verified or removed.

DogmaDragon:
Tested all cases except for `/`. It's an illegal character on Windows. And not used in [`performer_test.go`](https://github.com/stashapp/stash/blob/develop/internal/autotag/performer_test.go#L34) which leads me to believe it's not a valid use case. If someone on Linux could verify that would be great.

ui/v2.5/src/docs/en/Manual/AutoTagging.md
```diff
@@ -1,17 +1,15 @@
...
+When the Performer/Studio/Tag name has multiple words, the search will include paths/filenames where the Performer/Studio/Tag name is separated with `.`, `-`, `_`, and whitespace characters.
 
 For example, auto tagging for performer `Jane Doe` will match the following filenames:
 * `Jane.Doe.1.mp4`
 * `Jane_Doe.2.mp4`
 * `Jane-Doe.3.mp4`
 * `Jane Doe.4.mp4`
 
-Matching is case insensitive, and should only match exact wording within word boundaries. For example, `Jane Doe` will not match `Maryjane-Doe`, but will match `Mary-Jane-Doe`.
+Matching is case insensitive, and should only match exact wording within word boundaries. For example, the tag `Jane Doe` will not match `Maryjane-Doe` or `Jane-Doen`, but will match `Mary-Jane-Doe`, `Jane-Doe_n`, `Mary/Jane Doe`, `Mary-Jane/Doe`, and `[OF]jane doe`.
```
DogmaDragon, line 13: Based on @kermieisinthehouse comments [it appreas that `/` is not used as a separator](https://github.com/stashapp/stash/blob/9180a68c4570c62a11ab6c4dc600a85a6dfc1135/pkg/match/path.go#L22) so the `Mary/Jane Doe`, `Mary-Jane/Doe` examples need to be removed.

DogmaDragon:
Closing in favor of https://github.com/stashapp/stash/pull/4007

PR #3697, opened 2023-04-27
Tagger: Add primary sort for exact phash matches

### LINKED ISSUE #3695: [Bug Report]
Tagger shows scenes without a PHASH exact match and with fewer duration matches above the correct match which has an exact PHASH match and all fingerprints match duration. 

**To Reproduce**
Need specific files 

**Expected behavior**
The scene with one exact PHASH match and 100% 10/10 exact duration matches should be the default match rather than the scene with 1 PHASH exact match, lots of close matches and 2/29 duration matches.

**Screenshots**

v0.20.2

 - OS: Android 13
 - Browser Firefox Android
 - Version 114

**Additional context**

Flashy78:
Resolves #3695 

Currently the scene tagger is sorting multiple results by the number of phash matches up to a distance of 8.

This adds an initial sort before that, which first considers only exact phash matches (using duration as a tiebreaker), before falling back to the existing logic.


Flashy78:
After thinking about it more, I'm not sure this change is even useful.
I'm ok with just closing this PR.

PR #3698, opened 2023-04-27
Improve UI logging

DingDongSoLong4:
This makes a few improvements to the logging UI tab and the GraphQL logging API:

- The GraphQL logging query and subscription both now have optional minLevel parameters. This means that the UI can now only request the log messages it will actually end up showing, instead of always receiving all of them and having to potentially filter through thousands of trace messages.
- The backend now stores a cache of recent log messages for each log level rather than just the selected one. This means that, for example, if you have your Stash system log level set to "Info", and then open a new settings log tab and change the level there to "Trace", you will see the most recent trace messages as stored in the backend cache. Previously doing this would show no trace messages, as only "Info" messages would have been stored in the server cache.
- And then I've tweaked some log-related settings locale strings to improve clarity.


WithoutPants:
> The backend now stores a cache of recent log messages for each log level rather than just the selected one. This means that, for example, if you have your Stash system log level set to "Info", and then open a new settings log tab and change the level there to "Trace", you will see the most recent trace messages as stored in the backend cache. Previously doing this would show no trace messages, as only "Info" messages would have been stored in the server cache.

The system log level chosen impacts performance in addition to log verbosity. This change means that trace and debug log processing is performed even if the user has chosen log level less verbose than these. Further, this breaks the intention of the `logger.XFunc` methods, which are intended to only perform potentially expensive log string calculations if the log level is met. I think we'd be better off limiting the options for displaying the log messages to the minimum of the system log level setting.

Gykes:

PR #3707, opened 2023-04-30
Identify: Options to skip multiple results and single name performers

### LINKED ISSUE #3506: [RFC] Default Identify to skip items with multiple results
The common refrain when people are told to use Identify is that it can have incorrect matches.
I suggest adding an option to Identify to skip items with multiple matches coming back, and defaulting this option to on.

It would log that multiple results were returned and so the item was not matched.
The user could also choose a tag to attach to these items so that they can be easily found later.

This wouldn't eliminate incorrect matches, but might stop many of them? Especially helpful with all of the short form content that doesn't work very well with phash and returns lots of matches. At the detriment that users may end up with many scenes not matched.

Flashy78:
Resolves #3506 

Adds two new options to the Identify task, an option to ignore matches that are returned with more than one result, and an option to ignore creating performers with a single name and no disambiguation.
If either option is chosen, the user can optionally choose to tag any scene that meets those scenarios so they can be manually dealt with later on.

Identify Multiple Options

Also fixes the following bugs:
- Performer fields for Disambiguation, URL and Details are now also included from the scraped source
- Performer images are assigned on create (the first one in the list if there are multiple)
- Studio images are assigned on create


internal/identify/identify.go
```diff
@@ -62,64 +67,103 @@ type scrapeResult struct {
...
+			}
 		}
 	}
 
 	return nil, nil
 }
 
+// Returns a MetadataOptions object with any default options overwritten by source specific options
+func (t *SceneIdentifier) getOptions(source *ScraperSource) MetadataOptions {
+	options := t.DefaultOptions
```
WithoutPants, line 106: This should dereference `t.DefaultOptions` so that the original isn't modified. It's passed as a pointer, so all the stuff below is modifying it.

internal/identify/identify.go
```diff
@@ -62,64 +67,103 @@ type scrapeResult struct {
...
+		performerCreator:         t.PerformerCreator,
+		tagCreator:               t.TagCreator,
+		scene:                    s,
+		result:                   result,
+		fieldOptions:             fieldOptions,
+		skipSingleNamePerformers: *options.SkipSingleNamePerformers,
 	}
 
-	ret.Partial = getScenePartial(s, scraped, fieldOptions, setOrganized)
+	ret.Partial = getScenePartial(s, scraped, fieldOptions, *options.SetOrganized)
```
WithoutPants, line 166: Isn't `options.SetOrganized` potentially `nil` here? `getOptions` overrides the `DefaultOptions` if they are not `nil`, but as far as I can tell, it could still be `nil` if neither the source options or the default options have this field set to a non-nil value.

internal/identify/identify.go
```diff
@@ -130,19 +174,22 @@ func (t *SceneIdentifier) getSceneUpdater(ctx context.Context, s *models.Scene,
...
-	ignoreMale := false
-	for _, o := range options {
-		if o.IncludeMalePerformers != nil {
-			ignoreMale = !*o.IncludeMalePerformers
-			break
-		}
-	}
-
-	performerIDs, err := rel.performers(ctx, ignoreMale)
+	performerIDs, err := rel.performers(ctx, !*options.IncludeMalePerformers)
```
WithoutPants, line 177: Same issue as above.

internal/identify/identify.go
```diff
@@ -171,15 +226,7 @@ func (t *SceneIdentifier) getSceneUpdater(ctx context.Context, s *models.Scene,
...
-	setCoverImage := false
-	for _, o := range options {
-		if o.SetCoverImage != nil {
-			setCoverImage = *o.SetCoverImage
-			break
-		}
-	}
-
-	if setCoverImage {
+	if *options.SetCoverImage {
```
WithoutPants, line 229: As above.

internal/identify/identify.go
```diff
@@ -241,6 +288,51 @@ func (t *SceneIdentifier) modifyScene(ctx context.Context, txnManager txn.Manage
...
+			Mode: models.RelationshipUpdateModeSet,
+		}
+
+		if _, err := ret.Update(ctx, t.SceneReaderUpdater); err != nil {
+			return fmt.Errorf("error updating scene: %w", err)
+		}
+
+		logger.Infof("Added tag %s to skipped scene %s", *tagToAdd, s.Path)
+
+		return nil
```
WithoutPants, line 329: This could probably be done in fewer LOC with `scene.AddTag`.

internal/identify/performer.go
```diff
@@ -24,6 +27,12 @@ func getPerformerID(ctx context.Context, endpoint string, w PerformerCreator, p
 
 		return &performerID, nil
 	} else if createMissing && p.Name != nil { // name is mandatory
+		// skip single name performers with no disambiguation
+		if skipSingleNamePerformers && !strings.Contains(*p.Name, " ") && (p.Disambiguation == nil || len(*p.Disambiguation) == 0) {
+			// hack to let the upstream code know a performer was skipped
+			id := int(-1)
```
WithoutPants, line 33: I think returning and checking for an `ErrSkipSingleNamePerformer` error might be the better approach here.

WithoutPants (changes requested):
Static review only.

internal/identify/identify.go
```diff
@@ -62,63 +65,109 @@ type scrapeResult struct {
...
+		options.SkipSingleNamePerformerTag = source.Options.SkipSingleNamePerformerTag
+	}
+	return options
+}
+
+type ErrSkipSingleNamePerformer struct{}
+
+func (m *ErrSkipSingleNamePerformer) Error() string {
+	return "A performer was skipped because they only had a single name and no disambiguation"
+}
```
WithoutPants, line 136: Use `var ErrSkipSingleNamePerformer = errors.New(<message>)` instead.

internal/identify/identify.go
```diff
@@ -130,17 +179,20 @@ func (t *SceneIdentifier) getSceneUpdater(ctx context.Context, s *models.Scene,
...
+		includeMalePerformers = *options.IncludeMalePerformers
 	}
 
-	performerIDs, err := rel.performers(ctx, ignoreMale)
+	addSkipSingleNamePerformerTag := false
+	performerIDs, err := rel.performers(ctx, !includeMalePerformers)
 	if err != nil {
-		return nil, err
+		var singleErr *ErrSkipSingleNamePerformer
+		if errors.As(err, &singleErr) {
```
WithoutPants, line 191: With the above change, this can be changed to just be `errors.Is(err, ErrSkipSingleNamePerformer)`.

internal/identify/identify.go
```diff
@@ -62,63 +65,109 @@ type scrapeResult struct {
...
+			options := t.getOptions(&source)
+			if len(results) > 1 && options.SkipMultipleMatches != nil && *options.SkipMultipleMatches {
+				if options.SkipMultipleMatchTag != nil && len(*options.SkipMultipleMatchTag) > 0 {
+					// Tag it with the multiple results tag and ignore
+					err := t.addTagToScene(ctx, txnManager, scene, options.SkipMultipleMatchTag)
+					if err != nil {
+						return nil, err
+					}
+					return nil, nil
+				}
```
WithoutPants, line 88: I think this violates the intention of the `scrapeScene` method, which should only be concerned with getting and returning the results of the scrape. Instead, if multiple results are found and the options indicate that multiple matches should not be returned, then it should return an error value instead. This should be checked in the calling code, which can either integrate the multiple match tag in the `modifyScene` code, or just do the `addTagToScene` in a separate if branch.

WithoutPants (changes requested):

PR #3714, opened 2023-05-02
Close input file so SafeMove can delete it

puc9:
This is happening on Windows and over the network but at the end of SafeMove it fails the move with an error that it can't remove the input because it is in use. It turns out it is in use by the SafeMove itself :)


pkg/fsutil/file.go
```diff
@@ -33,6 +33,11 @@ func SafeMove(src, dst string) error {
 			return err
 		}
 
+		err = in.Close()
+		if err != nil {
+			return err
+		}
+
```
WithoutPants, line 40: This is redundant with the `defer in.Close()` call above. I think a better approach would be to convert the lines above `os.Remove` to be an inline function:
```
func() {
	in, err := os.Open(src)
	...
}()

if err := os.Remove(src) {
...
}
```

This way the deferred operations will be executed before the `Remove` is called. Can you make this change and ensure it fixes your original issue?
```

WithoutPants (changes requested):

pkg/fsutil/file.go
puc9: I'll test that change but it may take a day or two, just FYI.

puc9:
I hope this is what you meant (my Go knowledge is ~ 0).
It still correctly moves the file by copy and then remove

puc9:
Actually, I realized that the copy part doesn't copy the timestamps and that leads to a rescan of the destination file to recompute the MD5. I do not think that this is a desired behavior. If one wants to rescan after the move they can always do so.
I tried to make the code look good but I'm not sure about the outcome, feel free to suggest changes.

WithoutPants (approved):

PR #3715, opened 2023-05-02
Folder Gallery creation on a per folder basis

### LINKED ISSUE #3708: [Feature] Exclude/Include Folders from becoming Galleries
Right now, the option to create a Gallery from a folder with images is set globally or not at all, which is a bit frustrating, as the data saved in different libraries might be different. 
To put it differently, you cant prevent any image from being in a gallery if the folder to gallery option is set, because even if the images are put at the root directory of a library, still a gallery is created, with the name of the folder the libraries starting point.

Basically, I want to be able to exclude certain folders to be created as galleries, or include certain folders to be even if the option is not set.

For that, I see two Options: 
Either set this scanning behavior on a per library basis. This would be different than any other option, and as I've seen the code for scanning, it would also be fairly difficult.
Or simply set the possibility to set exceptions to the rule on a per folder basis. The easiest way to implement that would probably be to define a filename that, if present in a folder, will either force the creation of a gallery or force the folder to be exempted from this process. Basically, the way .nomedia works for android, so the naming could be something like .nogallery and .forcegallery.
This looks to me like the easiest option, I've already pinned down that all it takes is to add a little more boolean logic to the pkg/image/scan.go file for that. 

Thats why I tend to the second option, and I would also implement that myself, however, before I do, I'd like a little feedback whether this seems like a useful Idea or maybe I'm simply overlooking a feature here.

yoshnopa:
As I've described in #3708, this commit allows to create galleries from folders by setting a .forcegallery file inside of a folder, even if the setting "Create galleries from folders containing images" in the Library section is not activated.
Also, by creating a .nogallery file inside a directory, a folder is not turned into a gallery even if the setting is activated.

The check is case insensitive and the content of the file is irrelevant (although an empty file is the typical usage for such a system), in theory even a folder with that name should work. For those not familiar, on Linux, Mac and Android, files starting with a dot are hidden, so in case hidden files/folders are not set visible in your file manager, after creation this file will not show up anymore. 

Closes #3708 

A little Question about the translation system, I've expanded the text of an existing value (to document the behavior for the user). Do I need to mark this in some way so the translators are informed that this value needs to be retranslated/expanded? I have no idea how this system works unfortunately.

### DIFF
--- a/pkg/image/scan.go
+++ b/pkg/image/scan.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/stashapp/stash/pkg/file"
@@ -308,25 +307,31 @@ func (h *ScanHandler) getOrCreateGallery(ctx context.Context, f file.File) (*mod
 		return h.getOrCreateZipBasedGallery(ctx, f.Base().ZipFile)
 	}
 
-	// Read files in Folder to find out if the Folder is marked to be handled differently as the setting
+	// Look for specific filenames in Folder to find out if the Folder is marked to be handled differently as the setting
 	folderPath := filepath.Dir(f.Base().Path)
-	folder, err := os.Open(folderPath)
-	if err != nil {
-		return nil, fmt.Errorf("Could not open Folder %s: %w", folderPath, err)
-	}
-	files, err := folder.Readdir(0)
-	if err != nil {
-		return nil, fmt.Errorf("Could not read Files in Folder %s: %w", folderPath, err)
-	}
+	forceGalleryFileNames := [...]string{".forcegallery", ".Forcegallery", ".ForceGallery", ".forceGallery", ".FORCEGALLERY"}
+	noGalleryFileNames := [...]string{".nogallery", ".Nogallery", ".NoGallery", ".noGallery", ".NOGALLERY"}
 
 	forceGallery := false
-	exemptGallery := false
-	for _, file := range files {
-		if strings.ToLower(file.Name()) == ".forcegallery" {
+	for _, filename := range forceGalleryFileNames {
+		if _, err := os.Stat(filepath.Join(folderPath, filename)); err == nil {
 			forceGallery = true
+			break
+		} else if errors.Is(err, os.ErrNotExist) {
+			continue
+		} else {
+			return nil, fmt.Errorf("Could not test Path %s: %w", folderPath, err)
 		}
-		if strings.ToLower(file.Name()) == ".nogallery" {
+	}
+	exemptGallery := false
+	for _, filename := range noGalleryFileNames {
+		if _, err := os.Stat(filepath.Join(folderPath, filename)); err == nil {
 			exemptGallery = true
+			break
+		} else if errors.Is(err, os.ErrNotExist) {
+			continue
+		} else {
+			return nil, fmt.Errorf("Could not test Path %s: %w", folderPath, err)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


yorks-dev:
yes this looks like the perfect feature for this app.

pkg/image/scan.go
```diff
@@ -307,7 +308,29 @@ func (h *ScanHandler) getOrCreateGallery(ctx context.Context, f file.File) (*mod
...
+		return nil, fmt.Errorf("Could not open Folder %s: %w", folderPath, err)
+	}
+	files, err := folder.Readdir(0)
+	if err != nil {
+		return nil, fmt.Errorf("Could not read Files in Folder %s: %w", folderPath, err)
+	}
+
+	forceGallery := false
+	exemptGallery := false
+	for _, file := range files {
```
Flashy78, line 324: Rather than loop through possibly thousands of files in a folder, just check if the files you are looking for exist.
https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go

ui/v2.5/src/locales/en-GB.json
```diff
@@ -281,7 +281,7 @@
       "check_for_insecure_certificates_desc": "Some sites use insecure ssl certificates. When unticked the scraper skips the insecure certificates check and allows scraping of those sites. If you get a certificate error when scraping untick this.",
       "chrome_cdp_path": "Chrome CDP path",
       "chrome_cdp_path_desc": "File path to the Chrome executable, or a remote address (starting with http:// or https://, for example http://localhost:9222/json/version) to a Chrome instance.",
-      "create_galleries_from_folders_desc": "If true, creates galleries from folders containing images.",
+      "create_galleries_from_folders_desc": "If true, creates galleries from folders containing images by default. Create a File called .forceGallery or .noGallery in a folder to enforce/prevent this.",
```
Flashy78, line 284: The in-app help section should also be updated with a paragraph explaining why someone would want to use this new feature.

yoshnopa: Valid Point. Added a paragraph explaining the functionality, based on the updated codebase.

pkg/image/scan.go
yoshnopa: I was hoping for a case insensitive structure, to minimize usage problems. However I see your point. 
I've compromised by checking for a list of filenames, including camelcase, capitalized and lowercase writing. Those should probably take into consideration all scenarios while still reducing the actual looping to a low static value.

Is that acceptable?

Flashy78: LGTM

WithoutPants: I think there's little value in checking various combinations of case. My preference would be to make it all lowercase and clearly document it.

yoshnopa: Well, can't say I'm agree obviously, but I won't argue over it, its really not that important. Changed accordingly documentation and code.

stg-annon:
This would be out of scope for this PR but just something I was thinking about.

If we were to populate the `.forcegallery` file with metadata that stash could pick up on scan as well this could be a nice way of managing gallery folders, my first thought for data like this would be a unique identifier within the file so that on rescan stash can detect of the folder has been moved instead of creating another gallery.

yoshnopa:
@stg-annon 
As the one adding chapters to galleries and waiting for approval to add a plugin that opens gallery metadata from the comicrack standard https://github.com/stashapp/CommunityScripts/pull/160 you won't see me disagree :D
Actually, I think this could be something to add to the former mentioned plugin. Just look up .forcegallery files as well as Comicinfo.xml, and dont limit it to cbz files, and treat them the same. Shouldn't be too difficult I'd say. 
You can ask over on that PR/open an issue over there if you have any questions.

HijackHornet:
Thank you. This fix many issues i had with non zipped gallery.

Issue #3666, opened 2023-04-17
Subtitles do not display when video is fullscreen on iOS


org0ne:
Previous version v0.19.1 would display subtitles when the video was playing fullscreen, but the latest v0.20.2 only displays subtitles when the video is windowed.

NodudeWasTaken:
This is caused by this line https://github.com/stashapp/stash/blob/0069c48e7e9298936dc5494865d0a4f674d6b29a/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx#L278
If we set it to true it works.
We can either detect safari or just permanently set it to true.

PR #3729, opened 2023-05-11
Fix iOS captions

NodudeWasTaken:
iOS doesn't support fullscreen html captions and needs native text tracks.
By default video.js uses native text tracks when supported, which i reverted to.
An additional bug was that the video could exceed the container size (mostly when captions were enabled on iOS), which was fixed by adding a max-height to the video-wrapper.

Unrelated i also set preload to auto, such that interactive users can start loading the video immediately, shouldn't be too offensive.

Fixes: https://github.com/stashapp/stash/issues/3666

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
+++ b/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
@@ -275,7 +275,6 @@ export const ScenePlayer: React.FC<IScenePlayerProps> = ({
         chaptersButton: false,
       },
       html5: {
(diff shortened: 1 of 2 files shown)


WithoutPants:
Preload is set to none because otherwise it results in live transcodes incorrectly starting where a video cannot be directly streamed.

NodudeWasTaken:
> Preload is set to none because otherwise it results in live transcodes incorrectly starting where a video cannot be directly streamed.

I set it back to none to remove my previous transgressions 😢

PR #3756, opened 2023-05-23
Allow serving of interactive CSVs directly to Handy

hontheinternet:
### Description

This PR allows the interactive videos to directly serve the funscript files (in the CSV format) directly to Handy. 
This is possible when Handy is on the same WiFi network as the stash server, or when the stash server is hosted on the internet. 
This decreases the reliance on the TheHandy funscript hosting server and makes for a faster initial setup of loading the interactive script. 

- Added the following endpoint to serve CSV of funscript directly to Handy
  - `/<id>/interactive_csv`
  - For now, I bypassed the authentication for this just this API endpoint. Maybe a better way is to use the API key as a query parameter. 
- Added a setting `use-stash-hosted-funscript` which toggles between 
  1. the original way of uploading the funscript file to TheHandy server to get a temporary CSV format URL 
  2. the new way of just using the newly added endpoint as the CSV format URL
- Another thing to note is that I convert the funscript file to CSV and save it in the same folder as the Funscript (since the Handy needs it in a CSV format), lazily (meaning the CSV file is generated on first request) and use the CSV file it if it exists. 
  - Maybe we could change it to save it in the cache folder, but it worked fine for me. 

I did run the formatting and linting commands before making the PR. 

Please ignore the previous PR, there was an issue with GitHub. Just created another PR. 
First time submitting a PR here, let me know if there any issues, and hope that it gets merged. :)


ui/v2.5/src/locales/en-GB.json
```diff
@@ -693,7 +693,11 @@
           }
         }
       },
-      "title": "User Interface"
+      "title": "User Interface",
+      "use_stash_hosted_funscript": {
+        "description": "When enabled, given that the stash instance is accessible from the Handy (like on the same network), the funscript will be transferred directly from Stash server to Handy, without going through TheHandy servers.",
```
kermieisinthehouse, line 698:
```suggestion
        "description": "When enabled, funscripts will be served directly from Stash to your Handy device without using the third party Handy server. Requires that Stash be accessible from your Handy device.",
```

kermieisinthehouse:
Thank you for contributing!

I've reviewed the code and it looks good, but as you mentioned, passing an API key is preferable to hole-punching the authentication. You can probably add something to scene.go in urlbuilders to accomplish this, see GetStreamURL in that file for an example of how another feature does this.

WithoutPants:
Is there any benefit to using the Handy server at all? I'm wondering if we shouldn't just shift to directly uploading only and not bother with a separate setting.

I don't think converting and saving in the source file's directory is a good idea. Assuming that the conversion process isn't too intensive, I think we can probably convert on the fly in the direct funscript URL handler. We can look at caching it as a separate change later on if needed.

hontheinternet:
> Is there any benefit to using the Handy server at all? I'm wondering if we shouldn't just shift to directly uploading only and not bother with a separate setting.

Currently, we provide the URL of the Stash server to Handy so that Handy can try to access the CSV file directly. 
Suppose, we open the Stash server using http://localhost:9999,  then the URL sent to Handy is localhost, and if we want to send the correct accessible URL of Stash, then Stash has to determine among the different interface IPs stash is hosted on which URL/domain to send. 

For a general user, the default option of uploading to Handy might be easier and seamless, whereas in an opt-in feature, stash can allow direct serving to the Handy given the Stash server is accessible from Handy. 


> I don't think converting and saving in the source file's directory is a good idea. Assuming that the conversion process isn't too intensive, I think we can probably convert on the fly in the direct funscript URL handler. We can look at caching it as a separate change later on if needed.

Yeah, I was also not every inclined to adding files to the source directories. I just did it because it seemed easy with simple caching. I can benchmark the latency of converting Funscript to CSV for various sizes and if it is less than a few hundred milliseconds even for a large funscript, then I think on the fly is a good enough.

hontheinternet:
Cleaned up the commits (previously there were 2 copies of the same commits) and rebased to the latest `develop` branch. 
Also, removed the hole in the authentication and now it uses the API key as query parameter to let the handy access the csv hosted by Stash in case of an authenticated server. 

I guess the main thing left to do, is to not create a CSV in the source folder but somewhere in the cache folder or maybe on the fly. I'll try to make it on-the-fly conversion for now.

hontheinternet:
Made the Funscript to CSV conversion on the fly, and on my PC (CPU: 5600X) it takes much less a second. 
Let me know if there are any more changes needed. @WithoutPants

WithoutPants (approved):

PR #3758, opened 2023-05-23
Fix Interactive Heatmaps to match the length of the video

hontheinternet:
### Description

Previously, the interactive heatmaps used to only be generated for length of the funscript, hence for the scripts which end before the end of the video, the scrubber on the interactive heatmaps doesn't match the play position.
Now, fixed the Heatmap generation to be generated until the length of the video.


kermieisinthehouse (approved):
Code reviewed but not tested.

Thank you for contributing!

Issue #3753, opened 2023-05-22
[Bug Report] Some clip preview generations in image gallery erroring out


rg9400:
**Describe the bug**

A random assortment of mp4 and gif files are not generating clip previews for me. I suspect it is due to very short length, but not sure. A lot of other files did generate previews, so it seems isolated to only a percentage of my total library. All my webm files did generate previews. Sharing the error output that I could find in the logs for each file format (even at lower levels of logging, this is all that was outputted into the logs). Also noted elsewhere, gif files do not play when clicking on an image like webm/mp4 files do, only through the wall view.

Of 966 Clip Previews that are erroring out, around 224 are mp4 files. I think they were downloaded from a gif hosting website, but it's been so long that I do not remember for each example. 

`time="2023-05-22 17:38:42" level=error msg="getting preview for image /mnt/media/ViciousCautiousBooy-mobile.mp4: %!w(*fmt.wrapError=&{error running ffmpeg command <-v error -y -xerror -i - -t 15.610315 -max_muxing_queue_size 1024 -c:v libvpx-vp9 -vf scale=400:400:force_original_aspect_ratio=decrease -pix_fmt yuv420p -preset slow -crf 25 -threads 4 -strict -2 -f webm -an ->: exit status 1 0xc004d5e2c0})"`
`time="2023-05-22 17:38:38" level=error msg="getting preview for image /mnt/media/bcbbgf.gif: %!w(*fmt.wrapError=&{error running ffmpeg command <-v error -y -xerror -i - -t 0.66 -max_muxing_queue_size 1024 -c:v libvpx-vp9 -vf scale=640:640:force_original_aspect_ratio=decrease -pix_fmt yuv420p -preset slow -crf 25 -threads 4 -strict -2 -f webm -an ->: exit status 1 0xc003b12000})"`

@yoshnopa as they worked on the original PR

**To Reproduce**
1. Create an Image Library with video disabled
2. Enable Video Clip previews
3. Add mp4, webm, and gif into video extensions. I added gif into this section instead of images to see if it made a difference...it did not
4. Add short gifs and mp4s? Not sure if there is something specific about my files.
5. Run Generate and see error output

**Expected behavior**
These files should generate clip previews like other files since Stash can play them fine. MP4 files play fine when clicking on the image, and gif files play fine on the Wall view. To note: gif files do not play when clicking on the image, only through the wall. I tried with gif in both the image and video extension section.

[...]

PR #3764, opened 2023-05-24
Clip preview fix

yoshnopa:
The Generation of clips was done with piping, due to my initial attempt to support generation on the fly. As this is not wished, it has been built back. It was responsible for h264 not always being functional, as it does not support streaming/piping by default, see https://video.stackexchange.com/questions/34087/ffmpeg-fails-on-pipe-to-pipe-video-decoding. Aimed at #3753, however, it needs to be tested by the issue creator to make sure that it actually fixes the error.


rg9400:
This fixed all my pending clip previews, thanks!

PR #3765, opened 2023-05-24
fix Clip Gif Support

### LINKED ISSUE #3762: [Feature] gif support for Image Library
On the latest develop, you can set Image Libraries to include clips, including Clip Preview generation. This works for file formats such as `webm` and `mp4`, however `gif` support is not ideal. gifs do not generate clip previews, and while they play perfectly on the Wall view of the image gallery, they cannot be played when clicking into the image. In fact, the clip preview may be unnecessary since the gif itself is animated on the wall view, but ideally it should be excluded from the Generate task in that case, and the clicked carousel or image page views should play it.

I tested with `gif` in both the image and video file extension fields

Gif support in image libraries

Converting the gifs into webm files. This actually worked perfectly, and all the errors and issues disappeared. Webm might be a better format overall as well, but people may not know how or even want to convert them.

On windows though, below line can be run from the cmd window when in the library to convert all the gifs to webm. Probably a similar command can be created for bash as well

`for /R %f in ("*.gif") do (ffmpeg.exe -i "%~f" -c vp9 -b:v 0 -crf 40 "%~pf%~nf.webm")`

yoshnopa:
Gif files were handled as videofiles, which meant that they were in a videocontainer, so unable to display in lightbox or details page. This fixes that.

Sidenote: webm is currently handled as imagefile due to ffmpegs limits on decoding it, avif works perfectly as a videofile in a videocontainer, so gif was actually the only format broken.

Closes #3762

### DIFF
--- a/ui/v2.5/src/components/Images/ImageDetails/Image.tsx
+++ b/ui/v2.5/src/components/Images/ImageDetails/Image.tsx
@@ -261,7 +261,10 @@ export const Image: React.FC = () => {
 
   const title = objectTitle(image);
   const ImageView =
-    image.visual_files[0].__typename == "VideoFile" ? "video" : "img";
+    image.visual_files[0].__typename == "VideoFile" &&
+    image.visual_files[0].video_codec != "gif"
+      ? "video"
+      : "img";
 
   return (
     <div className="row">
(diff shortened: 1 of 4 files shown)


rg9400:
This is fixed

therandomguy42:
@yoshnopa how did you create the avif file, that it works as a videofile?

When i try to use avif files Firefox complains about an unsupported file format and MIME-type and Chrome does show a video player with black screen.

I did not use an animated avif though.
I used https://squoosh.app/ to create the avif.

ffprobe detects it as av1 with 1 frame.

PR #3772, opened 2023-05-26
Filter tweaks

DingDongSoLong4:
These are some tweaks and fixes to bugs I came across regarding the new filtering options from #3619:

- The search results are now cached, and new results are only shown once the graphql request has actually completed. This prevents the list from quickly disappearing while the query is loading and then reappearing once completed.
- The "null" and "not null" modifiers have been added back, and a criterion with an "is" modifier but no value is no longer treated as a valid criterion.
- Old criterion (ie from saved filters) using the "excludes" modifier are now transparently converted on the frontend to equivalent "includes" criterion, based on what the backend would have done anyway. This prevents these old criterion from appearing to have no modifier selected (since excludes is no longer shown).
- Criteria that have only excludes items are now displayed in the pills as if they are excludes criteria. Looks a bit nicer in my opinion than "includes (excludes ...)"
- The depth indicator (+all or +1,2,etc) has been added back, now using locale strings to be able to translate the "all" rather than having it hardcoded.
- The search query is now reset after an item has been selected, to behave more like the other UI inputs for performers, tags, etc. This prevents having to clear the field if searching for multiple items with different queries, as I suspect would be the case most of the time. I wouldn't mind reverting this if needed though.
- And lastly I have added back the normal "excludes" modifier to the other tag criteria (scene tags, performer tags, parent/child tags) since the excluded option on the "includes" modifiers isn't actually working for those filters. This should hopefully be temporary until those are fixed, which I will leave for someone more familiar with what's going on SQL-wise.


DingDongSoLong4:
Also another issue I just found is that "is not" modifiers (i.e. NOT_EQUALS in graphql) don't seem to be working for studios, performers or any of the tag criteria. I don't know if they worked before, but there is a use for them (except for studios, see below). The "is" modifier also doesn't work for studios and for the other tag criteria (scene tags, performer tags, parent/child tags).

For studios, everything that you can do with the "is" and "is not" modifiers is possible using "includes" and "excludes", so "is" and "is not" don't need to be shown in the UI, but I think for the sake of the API and external programs making them work is worth it.

ui/v2.5/src/models/list-filter/criteria/tags.ts
```diff
@@ -1,51 +1,70 @@
...
+  messageID: "tags",
+  type: "tags",
+  parameterName: "tags",
+  modifierOptions: [
+    CriterionModifier.IncludesAll,
+    CriterionModifier.Includes,
+    CriterionModifier.Equals,
+    CriterionModifier.IsNull,
+    CriterionModifier.NotNull,
+  ],
```
WithoutPants, line 16: Since these are all the same, make a local list value and use that instead.

WithoutPants: Actually these aren't quite the same, but I'm not sure why.

DingDongSoLong4: I went through each one to check what modifiers actually worked, and used which worked for each one. Excludes is in all but the first one because the excluded option wasn't working for them, I'll quickly test #3775 and then they should be able to be removed. Equals is only present for the normal tags modifier because it also didn't seem to work for the others.

Ideally they should indeed all be the same, when equals/not equals works for all the criteria.

DingDongSoLong4:
I've updated the supported modifiers to match #3775.

The search query input is also now focused after an item has been selected/deselected, saving a click when searching for multiple items.

PR #3775, opened 2023-05-27
Fix joined hierarchical filtering

WithoutPants:
Related to #3772 (ping @DingDongSoLong4)

Fixes the joined hierarchical filtering (scene/image/gallery performer tags, marker scene tags, and parent/sub tags).

This should remove the necessity to restore the excludes pills in the UI.

### DIFF
--- a/pkg/models/filter.go
+++ b/pkg/models/filter.go
@@ -135,12 +135,15 @@ type HierarchicalMultiCriterionInput struct {
 	Excludes []string          `json:"excludes"`
 }
 
-func (i *HierarchicalMultiCriterionInput) CombineExcludes() {
-	if i.Modifier == CriterionModifierExcludes {
-		i.Modifier = CriterionModifierIncludesAll
-		i.Excludes = append(i.Excludes, i.Value...)
-		i.Value = nil
+func (i HierarchicalMultiCriterionInput) CombineExcludes() HierarchicalMultiCriterionInput {
+	ii := i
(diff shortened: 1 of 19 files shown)


DingDongSoLong4:
Okay I've tested this and the excluded option is now being used rather than ignored, but the behaviour of all criteria except for marker scene tags seem to have changed.

For example, say you have a scene with two performers, where performer 1 has the 'foo' tag and performer 2 has the 'bar' tag.
Previously, if you filter for "Performer Tags excludes bar" then the scene would not show up - it effectively collects all the tags from each performer together and then filters using the resulting big list. This I think is the most intuitive and useful behaviour.
Now, if you filter for the same "Performer Tags excludes bar" then the scene would show up, because it sees that performer 1 doesn't have the bar tag and thus shows the scene.

The reason why the marker scene tags criterion isn't an issue is simply because a marker can only be part of one scene.

Also, using the "is" modifier on performer tags is now erroring with a `no such column: performers_scenes.id` error (for scenes that is, galleries and images have a similar error). Previously the filter would have no effect and just return all items, so it doesn't really matter, but I'm mentioning it anyway. Using "is" on parent/sub tags is still giving the previous behaviour of returning everything, but it seems to be working correctly now for marker scene tags.

WithoutPants:
Thanks for testing this. The behaviour should now be consistent with the previous behaviour. I've also fixed the "is" modifier error.

Issue #3779, opened 2023-05-28
[Feature] Use the same Zoom Picker for Wall View


persistentSoul:
**Is your feature request related to a problem? Please describe.**
I feel like there should be Zoom Picker for Wall in Scenes tab as well. Currently on Scenes Grid View has it.

**Describe the solution you'd like**
Zoom Picker for Wall View

**Describe alternatives you've considered**
Zooming in doesn't help.

**Additional context**
Use the same Zoom Picker for Wall View

PR #3780, opened 2023-05-28
feat: initial change for zoom index for wall

persistentSoul:
Adding a Zoom Picker for Wall View. I have modified some code made it more scalable for future too. 
Would have loved to add SFW video or animation to show the UI changes. Let me know a tool, I have used https://github.com/nsfw-filter/nsfw-filter but it doesn't support videos at the moment.

Issue Link - https://github.com/stashapp/stash/issues/3779

### DIFF
--- a/ui/v2.5/src/components/List/ListViewOptions.tsx
+++ b/ui/v2.5/src/components/List/ListViewOptions.tsx
@@ -138,7 +138,10 @@ export const ListViewOptions: React.FC<IListViewOptionsProps> = ({
   }
 
   function maybeRenderZoom() {
-    if (onSetZoom && (displayMode === DisplayMode.Grid || displayMode === DisplayMode.Wall)) {
+    if (
+      onSetZoom &&
+      (displayMode === DisplayMode.Grid || displayMode === DisplayMode.Wall)
+    ) {
       return (
         <div className="ml-2 mb-2 d-none d-sm-inline-flex">
           <Form.Control
--- a/ui/v2.5/src/components/List/ListViewOptions.tsx
+++ b/ui/v2.5/src/components/List/ListViewOptions.tsx
@@ -33,7 +33,7 @@ export const ListViewOptions: React.FC<IListViewOptionsProps> = ({
   displayModeOptions,
 }) => {
   const minZoom = 0;
-  const maxZoom = 3;
+  const maxZoom = 4;
 
   const intl = useIntl();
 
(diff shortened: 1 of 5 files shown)


WithoutPants:
You've introduced another zoom level but haven't accounted for this in the styling for other views. I think the new, largest zoom level should disable the mouseover zoom functionality, since it just zoomed it to be beyond the boundaries of the viewport. The mouseover transition zoom also needs to account for vertically oriented videos. On zoom level 3 and above (the middle zoom level), it zoomed in beyond the vertical boundaries of the viewport when moused over.

There's an existing bug with the mouseover zoom functionality when there are less than a full row of scenes are on the last row. Try setting the custom page size to 36 and you should be able to see the issue. The first item on the last row always moves to the right when moused over, regardless of what position it started from. Since you're already tinkering with the logic for this behaviour, it'd be good to have that addressed at the same time.

persistentSoul:
@WithoutPants issues are sorted now. I checked them locally. 

1. Zoom Level 4 doesn't transform scale anymore
2. Zoom Level 3 doesn't have transform scale beyond viewport
3. Existing bug for last row is also sorted now.
4. I have added a maxheight property to account for vertical videos.

kermieisinthehouse (approved):
Thank you for your first contribution to Stash!

I've successfully tested for functionality and reviewed the code. I did not test for any mouseover issues, and assume they are still present.

persistentSoul:
@WithoutPants just curious why version 0.22.0 milestone ? Why not next release which would be 0.20.3. Also I might be missing this information but how can I view release charter, so I know which features are in plan. I can pick up some.

DingDongSoLong4:
The next release is v0.21.0 and as I understand it is now considered feature-complete, i.e. only bug-fixes will be merged until it is released. v0.22.0 will then be the next release, as v0.21.x releases will be patch releases for fixing major bugs only.

If you're looking for something to work on, the [backlog milestone](https://github.com/stashapp/stash/milestone/5) is probably a good place to start.

WithoutPants:
> You've introduced another zoom level but haven't accounted for this in the styling for other views. 

This is still not addressed. See https://github.com/stashapp/stash/blob/1c13c9e1b143bf916caf74f98672e46623c9127f/ui/v2.5/src/index.scss#L133 for the other zoom levels. Either the styling needs to be added for the new zoom level, or the new zoom level should be disabled outside the wall view.

This looks like it's completely broken the markers view for me. The zoom control has been added to the images and galleries wall views, but does not actually do anything.

WithoutPants:
Closing in favour of #6011.

PR #3793, opened 2023-06-01
Fix videojs-vr issues

DingDongSoLong4:
These are some improvements and fixes related to the `videojs-vr` support added in #3636.

The video player no longer reinitializes when the scene or a scene property is changed, meaning that modifying a scene while playing will not reset the player back to 0:00. I also found a way to change the VR mode (ie turning it on and off) without restarting playback, using what I guess you could call undocumented functions but I think it's fine.

I've also added type definitions for `videojs-vr`.


Teda1 (approved):
I can confirm this does address the playtime reset issue and the issue seen when toggling 360 on and off. I also didn't encounter any new bugs from my testing. This change looks good.

PR #3794, opened 2023-06-01
Implement the HereSphere API

NodudeWasTaken:
Since i like VR, i thought it would be nice to implement the HereSphere API.
This commit implements most of the HereSphere api's functionality, except the /scan endpoint.
The /scan endpoint proved too troublesome compared to just letting the user press the "scan library" button.

Not implemented:
- /scan endpoint (optional, heresphere will manually scan scenes)
- HSP storage

HSP files are HereSphere config files with can optionally be stored on the server.
I dont think its appropriate to store this in stash, given that i dont expect this to be use by the majority of users, so i left it unimplemented.


yoshnopa:
Awesome, wanted to work on this later this year! Maybe I'll expand on your code then. 

For now however, my thoughts on whats needed and what could be nice:

Authentication is absolutely a must! Stash has a system that checks if your instance is reachable via internet and mandates a login then. This would be useless if you could access the stuff via the api, especially as it has (when fully implemented) the capability to also delete scenes.

I think all projections types should be supported, if not by any other means tags can be used.

Also, I think a very helpful feature would be to create multiple libraries from the saved filters: Every saved filter scene should become a separate library, the "All"-Library should remain as well.

NodudeWasTaken:
> Awesome, wanted to work on this later this year! Maybe I'll expand on your code then.
> 
> For now however, my thoughts on whats needed and what could be nice:
> 
> Authentication is absolutely a must! Stash has a system that checks if your instance is reachable via internet and mandates a login then. This would be useless if you could access the stuff via the api, especially as it has (when fully implemented) the capability to also delete scenes.
Authentication is now implemented!
> I think all projections types should be supported, if not by any other means tags can be used.
I have implemented filebased projection detection and tag based.
It also searches the scenes stashdb tags for VR tags.
> Also, I think a very helpful feature would be to create multiple libraries from the saved filters: Every saved filter scene should become a separate library, the "All"-Library should remain as well.
I would love to do this, but as far as i can tell constructing filters is client-sided.
I would have to remake the filter construction in go, which seems excessive.
Any help would be welcome

WithoutPants:
I'm planning to refactor the saved filter stuff. The current implementation isn't very good.

yoshnopa:
In that case I will look into saved libraries later... It's really just a nice to have anyway.

Kruk2:
@NodudeWasTaken just wanted to say: great work!

Few things that I found:
1) "fisheye190" projection which is basically rf52:
https://github.com/xbapps/xbvr/blob/047d1450cb251857a91e5357b00132ea313dc169/pkg/api/heresphere.go#L657
2) hardcoded "JAVR" tag
3) xbvr and [stash-vr](https://github.com/o-fl0w/stash-vr/blob/master/internal/api/heresphere/scan.go) implemented the hereshpere api (more or less accurately) so it might be worth checking how some things were done.
Stash-vr is parsing user filters and then displays them as sections in heresphere - it's working very nicely.
Initial configuration is not user friendly as you have to get filter ids you want to have in heresphere and provide them to stash-vr.

graphql/schema/schema.graphql
```diff
@@ -434,6 +435,19 @@ type Mutation {
...
+  "Sets a tag as favorite in the HSP API."
+  setHSPFavoriteTag(input: FavoriteTagHSPInput!): Boolean!
+  "Enables writing favorites to the HSP API."
+  setHSPWriteFavorites(input: HSPFavoriteWrite!): Boolean!
+  "Enables writing ratings to the HSP API."
+  setHSPWriteRatings(input: HSPRatingWrite!): Boolean!
+  "Enables writing tags to the HSP API."
+  setHSPWriteTags(input: HSPTagWrite!): Boolean!
+  "Enables writing deletes to the HSP API."
+  setHSPWriteDeletes(input: HSPDeleteWrite!): Boolean!
```
WithoutPants, line 450: These look unimplemented and unused.

graphql/schema/types/heresphere.graphql
```diff
@@ -0,0 +1,29 @@
...
+
+input HSPTagWrite {
+  "If writing tags to HSP Api should be enabled."
+  enabled: Boolean
+}
+
+input HSPDeleteWrite {
+  "If writing deletes to HSP Api should be enabled."
+  enabled: Boolean
+}
```
WithoutPants, line 29: I think these are unused.

internal/api/resolver_mutation_hsp.go
```diff
@@ -0,0 +1,29 @@
...
+	return true, nil
+}
+
+func (r *mutationResolver) SetHSPWriteTags(ctx context.Context, input HSPTagWrite) (bool, error) {
+	return true, nil
+}
+
+func (r *mutationResolver) SetHSPWriteDeletes(ctx context.Context, input HSPDeleteWrite) (bool, error) {
+	return true, nil
+}
```
WithoutPants, line 29: These are vestigial.

internal/api/routes_heresphere.go
```diff
@@ -0,0 +1,1402 @@
...
+			if processedScene.Stereo == HeresphereStereoMono {
+				processedScene.Stereo = HeresphereStereoSbs
+			}
+		}
+	}
+
+	// Detect VR modes from filename
+	file := scene.Files.Primary()
+	if file != nil {
+		path := strings.ToUpper(file.Basename)
```
WithoutPants, line 1038: This whole block should go into a separate function.

internal/api/routes_heresphere.go
```diff
@@ -0,0 +1,1402 @@
...
+					if height == 0 {
+						processedEntry.Resolution = mediaFile.Height
+						processedEntry.Height = mediaFile.Height
+						processedEntry.Width = mediaFile.Width
+					}
+					processedEntry.Url = addApiKey(fmt.Sprintf("%s%s?resolution=%s", sourceUrl, trans, res.String()))
+
+					typeName := transNames[i]
+					mediaTypes[typeName] = append(mediaTypes[typeName], processedEntry)
+				}
```
WithoutPants, line 927: Make this a separate function.

internal/api/routes_heresphere.go
```diff
@@ -0,0 +1,1402 @@
...
+	processedTags := []HeresphereVideoTag{}
+
+	// Load all relationships
+	if err := txn.WithReadTxn(r.Context(), rs.txnManager, func(ctx context.Context) error {
+		return scene.LoadRelationships(ctx, rs.repository.Scene)
+	}); err != nil {
+		return processedTags
+	}
+
+	if mark_ids, err := rs.resolver.Scene().SceneMarkers(r.Context(), scene); err == nil {
```
WithoutPants, line 672: I think there's a better way to do these queries. I'll need to think about it.

internal/api/routes_heresphere.go
```diff
@@ -0,0 +1,1402 @@
...
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+/*
+ * This endpoint is for letting the user update scene data
+ */
+func (rs heresphereRoutes) HeresphereVideoDataUpdate(w http.ResponseWriter, r *http.Request) error {
```
WithoutPants, line 338: This function is massive and needs to be split up.

internal/api/routes_heresphere.go
```diff
@@ -0,0 +1,1402 @@
...
+			// Update scene
+			if err := txn.WithTxn(r.Context(), rs.txnManager, func(ctx context.Context) error {
+				_, err := ret.Update(ctx, rs.repository.Scene)
+				return err
+			}); err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+		}
+	}
```
WithoutPants, line 324: This should be a separate function.

internal/api/routes_heresphere.go
```diff
@@ -0,0 +1,1402 @@
+package api
```
WithoutPants, line 1: Can we move this into a `heresphere` package and split this up into separate files as needed.

WithoutPants (changes requested):
Static code review

internal/api/routes_heresphere.go
NodudeWasTaken: The main problem with this is that many heresphere functions rely on functions in the api package.
Seperating it into a heresphere package would create an import cycle because the server.go always needs a reference to the routes from heresphere, and heresphere always needs a reference to GetBaseURL, context_keys, HandleFloat64Value, SceneMarkerCreateInput, SceneFinder, CaptionFinder, ResolverRoot

WithoutPants: This indicates a smell in the design. I'd suggest taking a look at the DLNA code, since it is doing something somewhat similar. As far as I can determine, there's no reason the heresphere code _needs_ to use the graphql-layer interfaces - you should be able to use the lower level functions to do what you need.

graphql/schema/types/heresphere.graphql
NodudeWasTaken: Its been removed

internal/api/resolver_mutation_hsp.go
NodudeWasTaken: Its been removed

internal/api/routes_heresphere.go
NodudeWasTaken: Done!

internal/api/routes_heresphere.go
NodudeWasTaken: Done!

internal/api/routes_heresphere.go
NodudeWasTaken: Done!

internal/api/routes_heresphere.go
NodudeWasTaken: Done!

NodudeWasTaken:
I dont really know how this works, am i supposed to write that this is ready for review or something?

WithoutPants:
I have spent a great deal of time mulling this change. I ultimately believe that this functionality belongs in a third-party utility or plugin.

There is a design consideration that the core system should be kept as lean as possible, and superfluous features should be deferred to third-party plugins and utilities. I believe this is one such feature.

This change adds further maintenance overheads that I'm not keen on introducing. Once this feature is introduced, the expectation will be that it will be maintained, and will be difficult to remove. It adds more consumers to the internal API, which means that changing the internal API will require even more code changes and testing to ensure the functionality isn't broken.

I know that this decision should have been made and communicated earlier and I can only apologise for that.

I don't believe that the current plugin infrastucture supports what is required for this functionality, but this could be converted to a third-party utility that consumes the stash graphql interface. I would like to add support for this kind of plugin in future, but I do not yet have a coherent design for it, so converting this to a plugin won't be feasible in the short-term.

NodudeWasTaken:
> I have spent a great deal of time mulling this change. I ultimately believe that this functionality belongs in a third-party utility or plugin.
> 
> There is a design consideration that the core system should be kept as lean as possible, and superfluous features should be deferred to third-party plugins and utilities. I believe this is one such feature.
> 
> This change adds further maintenance overheads that I'm not keen on introducing. Once this feature is introduced, the expectation will be that it will be maintained, and will be difficult to remove. It adds more consumers to the internal API, which means that changing the internal API will require even more code changes and testing to ensure the functionality isn't broken.
> 
> I know that this decision should have been made and communicated earlier and I can only apologise for that.
> 
> I don't believe that the current plugin infrastucture supports what is required for this functionality, but this could be converted to a third-party utility that consumes the stash graphql interface. I would like to add support for this kind of plugin in future, but I do not yet have a coherent design for it, so converting this to a plugin won't be feasible in the short-term.

Fair enough, but you should consider removing the deovr reward then 😋

NodudeWasTaken:
While we wait for the plugin interface, i started converting this to typescript, anyone interested can try it at:
https://github.com/NodudeWasTaken/stash-hsp

PR #3798, opened 2023-06-05
Make displayed resolution information coherent for 2:1/VR files

vt-idiot:
**Update to reflect the current state of the changes the commit makes.**

| Display Name | `>=`Dimension | Note (if any) / Common VR Resolution (and up) |
|--------|--------|--------|
| `HUGE` | 6144 | Technically tops out as 9999.<br>"Wait and see" what the "next label" after 8K will be. |
| 8K | 3840 | Common VR res of 7680x3840 |
| 7K | 3584 | 7168 x 3584 |
| 6K | 3000 | 6000 x 3000 |
| 5K | 2560 | 5120 x 2560 | 
| 4K | 1920 | 3840 x 1920 |

If Height > Width (portrait video) then it already looks at the larger of the two by default, ie. a 2160x3840 video will also say "4K". If there's some bizarre portrait aspect smartphone camera resolution I've borked with this, **please let me know**.

This deprecates the old "VR_HD" label. It also doesn't do anything to fix how it handles older 1:1 / TB VR videos, since the way Stash (already) handles resolution relies on "looking at the larger of height/width" and making its decision from there. e.g. 2160x2160 VR _already_ erroneously displays "4K" - since it's just looking at "2160". 16:9 3840 x 2160 _is_ 4K, it's not wrong. It does unfortunately do away with e.g. 2144x2144 _correctly triggering_ **Vr_Hd** internally even though it still showed "1920p" in the UI for those; those become "4K" too.

Someone smarter than me might want to try and implement handling for 1:1 video.

Original comment below.

---

Displayed resolutions in Stash were confusing as hell when it came to VR files - which are typically 2:1. Now I understand why, it's assuming 16:9 files/looking at height only.

These changes will make it display
* 8K for 7680x3840 - it currently will not display 8K for any VR titles at all to my knowledge, since none are 8640x4320
* 6K for 6144x3072
* 5K for 5400x2700
* 4K for 3840x1920

If that doesn't work, it should be modified to look at width.


ui/v2.5/src/utils/text.ts
Emilo2, line 1: I think the width should be used to approximate anything that has a K label.
That should work for both regular 2D and VR files better than this solution.

Emilo2:
There's also plenty of 7K (7168 x 3584) scenes out there and at least one 12K (12288 x 6144)

ui/v2.5/src/utils/text.ts
vt-idiot: Wouldn't that introduce issues for portrait content? I was trying to stay within the constraints of how it was already set up.

Plus I think that would introduce issues because the lower ones are all `###p`

I can try and re-write it to only look at width in general

vt-idiot:
> There's also plenty of 7K (7168 x 3584) scenes out there and at least one 12K (12288 x 6144)

@Emilo2 Added a 7K and a **HUGE** display for both. IMO to be "just right" it'd need to handle 1:1 / `width = height` videos separately (and spit out **VR**) look at height only for SD video since it could be 4:3 or 16:9, and then start looking at width only (or rather - the larger of the two, instead of the smaller of the two as it does now) at all widths >1000. Not sure how much 4:3 720p (960x720) porn there is out there, but it'd be a casualty of that.

But I'm not sure how to do that without mucking things up.

Kruk2:
If it helps that's a distinct list of VR resolutions from my library:

| width | height |
| ------ | ------ |
|2880 | 1440|
|3840 | 1920|
|3840 | 2160|
|4096 | 1920|
|4096 | 2048|
|4320 | 2160|
|4600 | 2300|
|4800 | 2400|
|5120 | 2560|
|5300 | 2650|
|5400 | 2700|
|5700 | 2400|
|5760 | 2880|
|5800 | 2900|
|6000 | 3000|
|6016 | 3008|
|6144 | 3072|
|6400 | 3200|
|6912 | 3456|
|7168 | 3584|
|7200 | 3600|
|7260 | 3630|
|7680 | 3840|
|8000 | 4000|
|8192 | 4096|

vt-idiot:
Thank you @Kruk2
> 2880x1440
I have some of those too, but 1440p ain't so bad as a display in Stash. At least it's accurate.
> 6000x3000, 8000x4000
Genuinely wasn't expecting those two. I am worried that trying to stick within the constraints of how it currently works might be too cumbersome.

Kruk2:
Yeah, I think there would have to be some kind of flexible rules so we can assign 8K to a range of heights?
8000x4000 examples
https://www.sexlikereal.com/scenes/luna-fae-is-worth-it-29098?tab=downloads
https://www.sexlikereal.com/scenes/all-hers-27660?tab=downloads

6000x3000
https://www.sexlikereal.com/scenes/sweet-teen-in-the-bedroom-12500

graphql/schema/types/filters.graphql
```diff
@@ -21,6 +21,7 @@ enum ResolutionEnum {
   "720p", STANDARD_HD
   "1080p", FULL_HD
   "1440p", QUAD_HD
+  "1920p", VR_HD @deprecated(reason: "Use 4K instead")
```
vt-idiot, line 24: Good call...didn't consider that, thank you!

vt-idiot:
**Update to reflect the current state of the changes the commit makes.**

| Display Name | `>=`Dimension | Note (if any) / Common VR Resolution (and up) |
|--------|--------|--------|
| `HUGE` | 6144 | Technically tops out as 9999.<br>"Wait and see" what the "next label" after 8K will be. |
| 8K | 3840 | Common VR res of 7680x3840 |
| 7K | 3584 | 7168 x 3584 |
| 6K | 3000 | 6000 x 3000 |
| 5K | 2560 | 5120 x 2560 | 
| 4K | 1920 | 3840 x 1920 |

If Height > Width (portrait video) then it already looks at the larger of the two by default, ie. a 2160x3840 video will also say "4K". If there's some bizarre portrait aspect smartphone camera resolution I've borked with this, **please let me know**.

This (now gracefully, thank you @WithoutPants ) deprecates the old "VR_HD" label.

It also doesn't do anything to fix how it handles older 1:1 / TB VR videos, since the way Stash (already) handles resolution relies on "looking at the larger of height/width" and making its decision from there. e.g. 2160x2160 VR _already_ erroneously displays "4K" - since it's just looking at "2160". 16:9 3840 x 2160 _is_ 4K, it's not wrong.

It does unfortunately do away with e.g. 2144x2144 _correctly triggering_ **Vr_Hd** internally even though it still showed "1920p" in the UI for those; those become "4K" now too.

Someone smarter than me might want to try and implement handling for 1:1 video, but they're far and few from the dark ages of VR, at least in my Stash.

PR #3799, opened 2023-06-05
Update vrmode.ts

vt-idiot:
Enabling the most common VR projection `180_LR` and an older but no longer used one `360_TB` in the UI.


WithoutPants:
These modes don't appear to work. As mentioned in #3636, the other modes don't work due to upstream issues with videojs-vr and three.js.

vt-idiot:
Weird. XBVR has it working. I'm not sure if it's because they're using an older version of videojs-vr/three.js or what.

Having 360 support is (practically) useless...in XBVR's case, they effectively only have 180 LR support. Or at least I think? It might really be 360.  I'll update stash tonight and see what the current state of "VR in a browser" is.

NodudeWasTaken:
This might sound stupid, but why dont we just downgrade to videojs-vr version 1.8.0 to make this work?
I can confirm that "180 lr" works when i do that 😳

scruffynerf:
as @NodudeWasTaken posted in discord:

```
diff --git a/ui/v2.5/package.json b/ui/v2.5/package.json
index c6473174..6391009b 100644
--- a/ui/v2.5/package.json
+++ b/ui/v2.5/package.json
@@ -72,7 +72,7 @@
     "videojs-contrib-dash": "^5.1.1",
     "videojs-mobile-ui": "^0.8.0",
     "videojs-seek-buttons": "^3.0.1",
-    "videojs-vr": "^2.0.0",
+    "videojs-vr": "1.8.0",
     "videojs-vtt.js": "^0.15.4",
     "yup": "^1.0.0"
   },
```

one line change, because clearly, v1.8.0 works, and XBVR did the same thing.

DingDongSoLong4:
Looking at the lack of activity on the `videojs-vr` repo, I doubt the issue going to be fixed soon. I agree with @NodudeWasTaken - if `1.8.0` works, we should just downgrade.

PR #3804, opened 2023-06-07
Fix tagger query path on Windows

WithoutPants:
Fixes the `parsePath` method to correctly parse relative windows paths.

The new logic will replace `\` characters with `/`, even on *nix systems and drive letters and UNC paths are stripped from the path - `\\foo` becomes `foo`, `D:\bar` becomes `bar`. Also fixes the following:
* fixes issue where the last directory component was stripped from the path - ie `/foo/bar/baz` became `foo baz` if `Path` query mode was chosen
* strips any `.` and `..` directory components from the list
* replaces any multiple space characters with single spaces

Fixes issue [raised in Discord](https://discord.com/channels/559159668438728723/559159910550732809/1105829146296336404).

### DIFF
--- a/ui/v2.5/src/components/Tagger/utils.ts
+++ b/ui/v2.5/src/components/Tagger/utils.ts
@@ -108,7 +108,7 @@ export function prepareQueryString(
     s = s.replace(new RegExp(b, "gi"), " ");
   });
   s = parseDate(s);
-  return s.replace(/\./g, " ");
+  return s.replace(/\./g, " ").replace(/ +/g, " ");
 }
 
 export const parsePath = (filePath: string) => {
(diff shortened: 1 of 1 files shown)


MaienM:
(I'm not familiar with the codebase but just happened to see the PR linked in Discord.)

Most of the changes outlined in the description sound like they break *nix to fix Windows. Is that desirable?

- `\` is a valid character in *nix paths (and _not_ a path separator), so replacing these with `/` actually mangles the path on *nix platforms (though most people probably don't actually have any `\` in their paths).
- Stripping `.` components from a path seems fine since they're pretty much a no-op anyway, but stripping `..` seems questionable. This will change `foo/bar/../baz` to `foo/bar/baz` while it really should resolve to `foo/baz` (on *nix anyway, not sure how Windows handles this).
- I'm not sure what the issue with multiple spaces is on Windows, but this is perfectly valid on *nix, so removing these mangles the path on *nix platforms.

WithoutPants:
I didn't write the description very clearly. `parsePath` is used in the UI to calculate the displayed query strings in the scene tagger, eg `d:\foo\bar\baz.mp4` becomes `foo bar baz` in the query fields. It doesn't matter if `..` is stripped out because it's meaningless when you're querying a scene name from stash-box. The multiple spaces thing means that the query string `foo   bar` becomes `foo bar`. The code strips out symbol characters, which is what causes the multiple spaces to be there. The result of this method isn't used in anything to do with actual file paths.

PR #3812, opened 2023-06-11
Add additional stats to the Stats page

hontheinternet:
### Changelog

- Added the following stats to Stats page 
  - Total O-counter
  - Total Play Duration
  - Total Play Count
  - Total Unique Scene Play Count


graphql/documents/queries/misc.graphql
```diff
@@ -49,7 +49,11 @@ query Stats {
     performer_count,
     studio_count,
     movie_count,
-    tag_count
+    tag_count,
+    o_counter,
+    play_duration,
+    play_count,
+    unique_play_count
```
WithoutPants, line 56: The fields here (and elsewhere in this PR) are a bit ambiguous. 

I'm open to other alternatives, but my suggestions would be:
- `total_o_count`
- `total_play_duration`
- `total_play_count`
- `scenes_played`

WithoutPants (changes requested):

graphql/documents/queries/misc.graphql
hontheinternet: I renamed the fields like you suggested, I think they sound good to me!

WithoutPants (approved):

Issue #3781, opened 2023-05-28
[Feature] Performer studio stats only toggle option on studio performer tab


echo6ix:
**Is your feature request related to a problem? Please describe.**
When navigating to the performer's tab on a studio page the performer's total library counts are displayed for scenes, images, galleries, etc. Also when sorting, such as by scene count, it is done based on the library totals. What if there were an option, via a toggle, to display and sort based on studio totals only.

The use case for this will vary by user. For me I would find it useful when determing who the most popular performer's are for a particular studio if I need to sleuth performers when I'm resolving a scene's missing performers.

**Describe the solution you'd like**
A toggle option nested within the overflow menu button...

![image](https://github.com/stashapp/stash/assets/37937507/cddf9e9e-d9ec-44c0-be34-f59ee14de870)

![image](https://github.com/stashapp/stash/assets/37937507/b876b237-a2e2-4dcc-a2c2-1b7c933abbde)

Not quite sure about the exact phrasing, but you get the gist.

**Describe alternatives you've considered**
Rather than using the overflow button drop down menu, having a dedicated toggle button. While this would give a clear indicator that the "studio stats only" filter is activated, it would also add more clutter to the toolbar menu.

**Additional context**
To indicate that the page is filtering by "studio stats only" we could add an indicator

![image](https://github.com/stashapp/stash/assets/37937507/8bcfc145-96f6-47c3-80a2-6a1020bd9d93)

dogwithakeyboard:
How about including both the total and filtered number? I find the fraction itself an interesting piece of data and it's fewer things to click.

![Screen Shot](https://github.com/stashapp/stash/assets/128322708/f4e39dfa-1998-405b-ba8e-d85c91af71bd)

It would make the popovers incredibly cramped. I don't know if flowing them on to two rows would be acceptable.

![Screen Shot 1](https://github.com/stashapp/stash/assets/128322708/ee373dfc-bd7b-4ccb-bfe1-c7eef9f2f7cb)

echo6ix:
Someone could easily argue that is it convenient to see both, and I'd agree.

But if you filter to see studio totals, seeing library totals is unnecessary. Also as you have identified, it causes more bloat on the UI (desktop and mobile). The UI imo should be clear and focused, with minimal amounts of excessive data cramped into a small area. It starts to make the presentation confusing and lacks visual hierarchy. Multiply all those popover ratios by every other card on the page and its going to be chaotic when the point is just to see the studio totals.

[...]

Issue #1961, opened 2021-11-05
Performers sub-page and performer cards include objects from other studios instead of only current studio


kermieisinthehouse:
The "performers" subtab of a Studio page currently considers ALL media from a performer, even from other studios.

The Performer preview cards should display only media from the current studio, and filters should only consider media from that studio. (e.g., filtering by scene count should only count scenes from the studio)

Gykes:
From an initial look I believe this is due to the backed Graphql. The counts are calculated globally and I dont see a way to pass filter criteria. Would need to be backend Graphql changes I think.

PR #3813, opened 2023-06-11
Appears With filtered gallery, scene, image counts and sorts

dogwithakeyboard:
Modifies the "Appears With" performer list to return counts of scenes, movies, galleries and images where both performers are credited instead of the total numbers for the individual performer. Sorting by scene count, gallery count and image count will also use the counts where the performers are credited together.

This is done by giving the appears with list its own filter mode and query and reworking the sql query to return counts.

- ~~Doesn't change the behaviour of the filters - they will still use the performers' individual totals.~~
- If acceptable, the same method could be used for studio > performer counts #1961 #3781
- I've made a horrible mess in the performer resolver. I'm sure there must be a smarter way to pass the orginating performer id through but I couldn't figure one out.


DingDongSoLong4:
I don't think adding appears with counts to the performer type directly is a good idea. Not only does it require the horrible mess in the performer model resolver, but it means that the fields in the performer type are no longer solely determined by the performer ID, which is bad for caching as this is done by object ID.

I think a better way of doing it would be to add another field to the `FindPerformersResultType` object, perhaps called `filtered_counts`, which would return a parallel array of objects (say `FilteredCountType`) containing the scene, gallery and image counts for each performer in the result, filtered by the current filter.

So for example if you were to do a performer query with filter "Studios includes A, B" then the `filtered_counts` field would contain the scene/gallery/image counts for each performer where those scenes/galleries/images are from studios A or B. Or, in the case of the appears with tab, the query is a performer query with filter "Performers includes A" where then the `filtered_counts` field would contain the scene/gallery/image count for that performer where those scenes/galleries/images include performer A. The Studios and Performers criteria are the only ones that make sense to be taken into account here. Using both at once should also work - i.e. both "Performers includes A" and "Studios includes A,B" should return counts for scenes/galleries/images containing both performer A and from either those studios.

To make this actually work, changing the performer `Query` method to work more like the scene `Query` method would probably be best, so that calculating these additional counts can be disabled if not needed. For filtering and sorting, `filtered_scene_count`, `filtered_gallery_count` and `filtered_image_count` options would then make sense, filtering and sorting according to those values.

The SQL is probably going to get more complicated though, especially since studio hierarchies would have to be taken into account. Getting only performers to work at first would be great, as once the framework is there adding the studio filter would just be a (not so simple) matter of adding even more complicated SQL to the query.

I think doing it this way is more generic and powerful, and wouldn't need a separate "appears with" filter type. On the frontend, besides adding the new filter/sort options, the only change would need to be adding another prop to the performer card to pass the filtered counts through to override the displayed counts. The scene/gallery/image count sort options should probably also correspond to their filtered versions for performer filters.

I've wanted to fix those counts for a long time, but every time I've basically put it off as being too difficult. I'm definitely not an SQL wizard and I don't really have a good understanding on how the filter SQL queries are put together. You do seem to have somewhat of an idea though, so hopefully what I'm proposing here isn't too difficult to get working.

dogwithakeyboard:
Thanks for the writeup! I was never very happy with the performer resolver so I made the pr as a request for feedback as much as anything.

I think I'm going to go and have long hard stare at the scene query and see if I can figure out what makes it tick.

PR #3816, opened 2023-06-12
Include subsidiary studios/tags in tab badge counters

### LINKED ISSUE #3492: [Bug Report] Scene count incorrect when including child studios
When the Settings -> Interface -> Studio view -> Display sub-studios content option is enabled and you view a studio that has sub-studios, the scenes/performers of the sub-studios are correctly displayed, but the counters along the top do not include them.

### LINKED ISSUE #3814: [Feature] Dynamically update details tab badge counters to reflect changes by filters
I use the details tab sections of pages quite often, and while it serves its purpose, there is a particular aspect of the UI that is very awkward and requires improvement which will enhance overall user experience. The issue is the static counter badges do not accurately reflect the content associated with the tabs when there are filters applied (which change the amount of items associated to said tabs).

For example, if I make a default filter to show subsidiary studios on studio pages. In this scenario I click on a primary studio that has 0 scenes and 0 performers. However, this studio has subsidiary studios with a total of 10 scenes and 10 performers. The badge counters will still read 0 respectively although clicking either tab will show 10 scenes and 10 performers.

This is a disparity in the number of items that the UI is communicating to the user and the number of items that is actually contained within that tab. When a badge indicates a specific number, users naturally expect to find that corresponding number of items within the tab. Consequently, this means the badge counters can be misleading. While I understand that the counts may be intended to be associated directly with unfiltered content, it still creates confusion when a badge suggests one thing, but the actual content differs.

To enhance clarity and accuracy of the badges I would suggest the counter badges get updated each time a filter is applied. More over, I think it would be ideal if there was a visual cue to indicate the count badges have been affected by the filter. The most obvious way to do this is to change the badge text color from white to bright cyan (bright cyan is the color of the filter badge)

image

The number `2` on the image counter changed from the base reading. It's color is the same as the filter badge color to visually indicate the correlation.

DingDongSoLong4:
This is a fix/improvement to the tab badges in the studios and tags pages. The counts displayed on the tab badges will now include subsidiary studios/tags if those options are enabled, meaning that the counts now accurately reflect the number of items which will be displayed by default when you click on the tab.

Potentially closes #3814: this only changes the initial value displayed in the badge, they do not dynamically update. As per my comment there, I don't think dynamically updating the counter is necessary, and getting that to work would likely require larger changes. I also haven't hidden the counter when the tab is selected, but that shouldn't be too hard to make happen if required, although I also don't think that's really needed.

Closes #3492

### DIFF
--- a/graphql/documents/data/studio.graphql
+++ b/graphql/documents/data/studio.graphql
@@ -17,10 +17,15 @@ fragment StudioData on Studio {
   ignore_auto_tag
   image_path
   scene_count
+  scene_count_all
   image_count
+  image_count_all
   gallery_count
+  gallery_count_all
   performer_count
+  performer_count_all
   movie_count
+  movie_count_all
   stash_ids {
     stash_id
     endpoint
--- a/graphql/documents/data/studio.graphql
+++ b/graphql/documents/data/studio.graphql
@@ -17,15 +17,15 @@ fragment StudioData on Studio {
   ignore_auto_tag
   image_path
   scene_count
-  scene_count_all
+  scene_count_all: scene_count(depth: -1)
   image_count
-  image_count_all
+  image_count_all: image_count(depth: -1)
   gallery_count
-  gallery_count_all
+  gallery_count_all: gallery_count(depth: -1)
   performer_count
-  performer_count_all
+  performer_count_all: performer_count(depth: -1)
   movie_count
-  movie_count_all
+  movie_count_all: movie_count(depth: -1)
   stash_ids {
     stash_id
     endpoint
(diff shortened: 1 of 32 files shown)


WithoutPants:
Could the `_count_all` fields not be integrated into the existing fields by adding an optional parameter to the resolver?

ie 
```
type Tag {
...
  scene_count(include_subs: Boolean): Int
...
}
```

The client queries would need to be changed to accept and pass the parameter.

DingDongSoLong4:
I wasn't aware that was possible, but that certainly is a cleaner way of doing it. I've also gone one better and made it a integer depth parameter rather than just a boolean for including subsidiaries.

And then also rebased to develop.

PR #3817, opened 2023-06-13
Add folder move detection

### LINKED ISSUE #2721: [Feature] Path update of folder based galleries
If folder based galleries are moved/renamed, a duplicate gallery (without metadata) is created. I know that switching to zip files solves this issue, but also creates other problems (e.g. acessing your galleries outside stash becomes inconvenient).

The checksum of a folder based gallery changes, if the folder is moved/renamed, while this is not the case for a zip gallery. 

The topic has been discussed here already: #1928

Folder based galleries should update the path (similar to zip galleries) in case the folder is moved/renamed instead of creating a duplicate gallery.

An alternative might be a functionality to copy metadata from one gallery/scene to another. The functionality should be scriptable to be convenient for a large number of galleries.

WithoutPants:
Attempts to detect when a folder has been moved during a scan and updates the existing folder entry.

When a new folder is encountered during scan (ie a folder that is not registered in the database), the system will perform move detection with the following algorithm:
- scan each file under the new folder (only - does not recurse to subdirectories). For each file:
  - try to find a file in the database with the same basename, size and mod time. 
  - for each existing file found, check if the parent folder exists in the file system. If it does not, add the missing folder as a candidate. Keep a count of matching files for each missing folder
- a rename will be detected if any folder candidates have >= 50% of their files in the new location. If there are multiple candidates, the candidate with the most matching files will be selected.

A folder must have at least one file in order to be detected for a rename.

To use an example:

Existing file system:
```
/foo
  /bar
    /1.jpg
    /2.jpg
    /baz
      /3.jpg
```

If `/foo/bar/baz` is moved to `/foo/bar/xyz`, then the scan detect `xyz` as a new folder, triggering rename detection. This will find `3.jpg` under `xyz`, and find the matching file in the database under the old location `/foo/bar/baz`. Since all of the files under `baz` match in the new folder, `/foo/bar/baz` will be updated to `/foo/bar/xyz`.

If `/foo` is moved to `/oof`, then `/foo` will _not_ be detected as a rename, since it has no files. A new folder `/oof` will be created. However `/foo/bar` has been moved to `/oof/bar` and it should be detected as a rename since all of its files will match against the old location. Likewise `/foo/bar/baz` will be detected as a rename to `/oof/bar/baz`.

There are many corner cases for this, and I haven't yet exhaustively tested them.

One such corner case is splitting a directory into multiples:
```
/foo/bar
  /1.jpg
  /2.jpg

to

/foo/baz
  /1.jpg
/foo/xyz
  /2.jpg
```

In this scenario, one of the folders will be detected as a move (50% of files match) and the other will be a new folder. Which folder is currently undefined, but perhaps it should use the path as a tie breaker.

If there were three files in the original directory, and all three ended up in different directories, then no folder will be detected as a rename, since no folder will have >=50% matching files.

Resolves #2721.

### DIFF
--- a/pkg/file/file.go
+++ b/pkg/file/file.go
@@ -154,10 +154,12 @@ type Getter interface {
 	FindByFingerprint(ctx context.Context, fp Fingerprint) ([]File, error)
 	FindByZipFileID(ctx context.Context, zipFileID ID) ([]File, error)
 	FindAllInPaths(ctx context.Context, p []string, limit, offset int) ([]File, error)
+	FindByFileInfo(ctx context.Context, info fs.FileInfo, size int64) ([]File, error)
 }
 
 type Counter interface {
 	CountAllInPaths(ctx context.Context, p []string) (int, error)
+	CountByFolderID(ctx context.Context, folderID FolderID) (int, error)
 }
 
 // Creator provides methods to create Files.
--- a/pkg/file/folder_rename_detect.go
+++ b/pkg/file/folder_rename_detect.go
@@ -1,0 +1,195 @@
+package file
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+
+	"github.com/stashapp/stash/pkg/logger"
+)
+
+type folderRenameCandidate struct {
+	folder *Folder
+	found  int
+	files  int
+}
+
+type folderRenameDetector struct {
+	// candidates is a map of folder id to the number of files that match
+	candidates map[FolderID]folderRenameCandidate
+	// rejects is a set of folder ids which were found to still exist
+	rejects map[FolderID]struct{}
+}
+
+func (d *folderRenameDetector) isReject(id FolderID) bool {
+	_, ok := d.rejects[id]
+	return ok
+}
+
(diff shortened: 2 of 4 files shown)


SirCumAlot1988:
Many thanks for implementing this! It's highly appreciated :-)

I've tested it on my side: Works perfectly

Just a quick question:
Would it be complicated to implement, that the algorithm also detects changing from a zip based gallery to a folder based gallery with the same content?

The situation on my side is now, that I have a lot of zipped galleries (because it was more convenient previously) and would like to migrate to folder based galleries now. However, with the current implementation this is not detected. I performed the follwoing steps:

- Unzipped Gallery
- Deleted zip file
- Set "Create galleries from folders containing images" in Stash settings to True
- Run a Scan

Now a new gallery is created (i.e. the migration is not detected).

Vice versa (folder to zip) is not needed from my side, but of course would make sense for the sake of completeness (if it's just a minor source code change).

If it's a big implementation effort, I might see if I can write a script for migration.

WithoutPants:
Can you please raise a new issue for this use case?

SirCumAlot1988:
Done, see #3921

PR #3828, opened 2023-06-15
Prevent fullscreen break when loading scene

### LINKED ISSUE #3615: [Bug report] Scenes player don't stay in fullscreen
Browser: Firefox 111.0.0 (64bit)
Stash:      v0.20.0
___
Hello,
If you play a video in fullscreen and it end, the next video will not be in fullscreen

1. Go to Scenes
2. Select the first one
3. Go Fullscreen
4. Wait for it to finish OR use shortcut P+N
5. Next Video isn't in fullscreen
___
It seems like if the next video as already been started once,  fullscreen mode will stay.

1. Go to Scenes
2. Select the first one
4. Wait for it to finish OR use shortcut P+N
5. Go back(shortcut P+P)
6. Go Fullscreen
7. Wait for it to finish OR use shortcut P+N
8. Then the next Video is in fullscreen

___
Edit:
The console indicate the following warning:
`Exited fullscreen because fullscreen element was removed from document.   index-f68f9f4c.js:106:5592`
___
Edit 2:
Didn't seems to be a problem in v0.19.1

WithoutPants:
Fixes #3615 

Basically maintains the existing value of `scene` while loading the next scene. This means that the loading spinner won't be displayed while loading a new scene, but the video player will be maintained during the loading period.

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneDetails/Scene.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneDetails/Scene.tsx
@@ -543,6 +543,17 @@ const SceneLoader: React.FC = () => {
   const { configuration } = useContext(ConfigurationContext);
   const { data, loading, error } = useFindScene(id ?? "");
 
+  const [scene, setScene] = useState<GQL.SceneDataFragment | undefined>(
+    data?.findScene ?? undefined
+  );
+
+  // only update scene when loading is done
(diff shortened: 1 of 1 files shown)


Sassy0P:
Damn, was finally about to try to solve that issue.

Works perfectly on my side

Issue #3840, opened 2023-06-19
[Bug Report] Marker exclude filter broken following #3619


StashNewbie:
**Describe the bug**
In v0.21.0, in the marker section, I've tried creating a new filter with "Tag excludes xyz" and this still shows items that include that tag.  If I flip the exclude to include, then this filter works as expected, but exclude appears to be broken?

**To Reproduce**
1. Go to /scenes/markers
2. Remove all filters.
3. Add a Tag / Exclude filter.
4. Markers that include that filter still displayed.

**Expected behavior**
Markers that have the excluded tag should be excluded from the filter 

**Screenshots**
Trivial to reproduce - assume not necessary

**Stash Version: (from Settings -> About):** v0.21.0

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] MacOS
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] 114.0.5735.106

**Additional context**
NA

PR #3842, opened 2023-06-19
fix: User joinedHierarchicalMultiCiterionHandlerBuilder to filter Marker

Sassy0P:
PR for fixing #3840

Using the **joinedHierarchicalMultiCiterionHandlerBuilder** in the **sceneMarkerTagsCriterionHandler** to apply the tag filtering.

Reviews are welcome as it is the first time that I take a look at this project or use Go.

Notes:
There is actually no test for this part of the project.


WithoutPants:
The `joinedHierarchicalMultiCriterionHandlerBuilder` doesn't really work here because it also has to check the primary tag. With this change, if you filter on tag `foo`, it won't include markers with the primary tag set to `foo`.

WithoutPants:
Closing in favour of #3846

PR #3843, opened 2023-06-20
Add codec filters

### LINKED ISSUE #2120: [Feature] Add codec filter type for scenes
There is no method to filter or sort scenes by (video) codec

Add a filter for codecs

**Additional context**
- List all all available codecs in user's Stash library within drop down combo box?
- List audio codecs as well?

goq2E6j9y78JRZapoAiEEL

plato178:
Resolves #2120 

This pull request is a first attempt at adding filters for audio and video codecs, as described in issue #2120.

Audio codec filter:

Screenshot 2023-06-20 at 10 55 50

Video codec filter:

Screenshot 2023-06-20 at 10 55 59

At present, the filters are populated using a pre-defined hard-coded list of codecs (separate lists for both audio and video). Ideally this should query for all codecs currently used in all scenes in a user's database, but I thought this would be a good first attempt. 

Please note that this is my first open-source contribution for... well, anything, plus this is my first time using Go and sqlite (I'm a JS/TS developer primarily), so please let me know if there's anything that I'm doing wrong or forgotten to do 😅


WithoutPants:
I think maintaining a list of possible codecs is probably not the right way to go. It's probably simpler and easier to maintain to  just make it a free text field.

plato178:
That’s fair - I had been looking at querying for all codecs available in the database when the filter loads, but with my limited knowledge of Go and the code base as a whole, I couldn’t figure out how to do this. 

I will change it to work as a text field for now - maybe I can work out a way to use the approach above.

DogmaDragon:
> That’s fair - I had been looking at querying for all codecs available in the database when the filter loads, but with my limited knowledge of Go and the code base as a whole, I couldn’t figure out how to do this.
> 
`matches regex` and `not matches regex` would work for something like that. Or just `includes`/`excludes` though regex is more flexsible.

plato178:
I've reworked my submission to be a string filter instead, as suggested.

WithoutPants (approved):

Issue #3850, opened 2023-06-22
[Bug Report] Queue:Next add scenes to the start


Sassy0P:
**Describe the bug**
Clicking on next(down arrow) in the queue tab of scene, add the new scenes at the top of it.

**To Reproduce**
Required enough scenes to have to pages for the queue (> 40)

1. Play a scene.
2. Scroll to the bottom of the queue tab.
3. Click on the down arrow (next).
4. Scroll to the top of the queue tab.
5. See the added scenes in the wrong spot.

**Expected behavior**
Newly added scenes should be at the bottom.

**Screenshots**
If applicable, add screenshots to help explain your problem please ensure that your screenshots are SFW or at least appropriately censored.
Before queuing more
![Normal queue](https://github.com/stashapp/stash/assets/126810860/d1b0e628-fb34-4944-bb9d-ea9d9ba0d1cd)
After queuing more
![New scenes in the wrong spot](https://github.com/stashapp/stash/assets/126810860/0bcde262-de16-4181-9061-830a8cc876d4)

**Stash Version: (from Settings -> About):**
v0.21.0-8-gbcf9019c
But have been a bug since at least v0.19

**Desktop (please complete the following information):**
 - OS: Linux (Pop!_OS 22.04 LTS 64bit  VM)
 - Browser: Firefox 111.0.1 (64-bit)

Issue #1449, opened 2021-05-28
[Feature] Queue Improvement


Belleyy:
**Describe the bug**
When pressing the "show more" button (:arrow_down_small:) on queue tab, the new scenes go to the top of this list. User need to scroll up to see new scenes.

**To Reproduce**
1. Go to Queue tab (Scene)
2. Click on the "show more" button :arrow_down_small: (Bottom button of tab)
3. New scene come from the top

**Expected behavior**
New scene should come next to the last scene. 

Current behavior: Last scene has N°40, pressing button, the scene is now N°80.
Want: Last scene has N°40, pressing button, the scene stay N°40, new scenes is after (N°41->80)

**Stash Version: (from Settings -> About):**

Version: | v0.7.0-35-gc5fed1bb
-- | --
Build hash: | c5fed1bb
Build time: | 2021-05-27 23:04:39

**Desktop (please complete the following information):**
 - Windows 10
 - Firefox 88.0

**Additional thing:**

If you are on the last scene and the "show more" button exist (:arrow_down_small:), the next button (:arrow_forward:) should appear in the `queue-controls`:
![image](https://user-images.githubusercontent.com/23707269/119988288-3a735980-bfc6-11eb-981b-ec0d6e95efe6.png)

PR #3851, opened 2023-06-22
fix: onQueueMoreScenes was adding new scenes at the start of the queue

Sassy0P:
Changing the order new scenes merge with the current queue for them to be at the bottom.

Should fix: #3850


Sassy0P:
Should also mark #1449 has closed.

PR #3852, opened 2023-06-23
Multiple scene URLs

WithoutPants:
Part of #1868.

Replaces the existing scene URL field with URLs.

image
image

The scraping interface is updated to accept `url` or `urls` fields.

The stash-box interface is updated to get multiple URLs where applicable.

The submit draft functionality currently only submits the first URL, since the stash-box draft interface does not accept multiple URLs.


DingDongSoLong4:
Would now perhaps be a good time to change the "Downloaded From" label to a more generic "URLs"? Links aren't necessarily directly from studio sites - displaying links to sites like IAFD as "Downloaded From" doesn't really make sense.
"URLs" is also shorter - "Downloaded From" takes up quite a lot of space.

Issue #3834, opened 2023-06-16
Hitting "Play Next" Endlessly


WithoutPants:
### Discussed in https://github.com/stashapp/stash/discussions/3833

<div type='discussions-op-text'>

<sup>Originally posted by **hepheastus** June 17, 2023</sup>
I've been trying to figure out how to hit next endlessly.
Right now, I can only hit next 40 times before I have to either hit shuffle in the queue log or go back to scenes (which I have defaulted to random) and click the first scene.  I have even tried to hit the bottom down arrow button in the queue list or even making my default scene show 1000 files instead of 40 and it still defaults to 40.

I just wanna keep hitting next forever.</div>

Sassy0P:
As it's also a problem I encountered, I I tried to fix it. 

So I've local branch that solve this problem, but I need to get rid of some quirks. 

I'll try to make a draft later.

PR #3853, opened 2023-06-23
fix: "Endless" queue by auto-loading the scenes 

Sassy0P:
This pull request as the goal to automatically load more of the queue/next page to play the next scene

Would fix: #3834 

It should not be possible for _loadNextScene_ and the effect to cause an infinite loop as there are guards in on _onQueueMoreScenes_ and the effect check for specific conditions and runned hen the queues is updated.

____

Seems to be working if someone want to try.


Sureiya:
@Sassy0P Couldn't figure out why this wouldn't build. Would you mind pushing an empty amend so the pipeline will run again?

Sureiya:
Adding prevQueueScenes as a dependency and removing the dependencies all-together on the respective useEffects fixed the eslint errors on make validate for me.

```// TODO: Find a better way to compare the needs to load next scene after adding more scenes to the queue.
  // Having the entire previous queue saved feels bad.
  const prevQueueScenes = usePrevious(queueScenes);
  useEffect(() => {
    if (prevQueueScenes != undefined) return;

    if (sceneQueue.query) {
      getQueueFilterScenes(sceneQueue.query);
    } else if (sceneQueue.sceneIDs) {
      getQueueScenes(sceneQueue.sceneIDs);
    }
  }, [sceneQueue, prevQueueScenes]);

  useEffect(() => {
    if (
      prevQueueScenes?.length == currentQueueIndex + 1 &&
      queueScenes?.length > prevQueueScenes?.length
    ) {
      loadNextScene(continuePlaylist);
    }
  });

WithoutPants:
Closed in favour of #4325

PR #3880, opened 2023-07-03
Performer filtered counts

### LINKED ISSUE #1961: Performers sub-page and performer cards include objects from other studios instead of only current studio

### LINKED ISSUE #3781: [Feature] Performer studio stats only toggle option on studio performer tab

dogwithakeyboard:
Shows accurate numbers of scenes, images, galleries and movies on performer cards when filtering for studios and/or performers. 

The `FindPerformersResultType` now returns an optional, parallel set of counts when using the studios or performers `performerFilter`. The background query can handle a combination of studio and performer filters and returns counts where performers appear together or with a studio or both.

The performer and studio filters are now picked up in the `Performer List` and passed to the `Performer Card` so the extra props in the `Studios > Performers` and `Performers > Appears With` tabs are unnecessary and everything behaves as expected in the Performers page.

On the frontend there is a dedicated button in the performer filter to toggle between displaying filtered and unfiltered counts.  When filtered the scene etc buttons will use the current filter settings to display the scenes so (hopefully!) the number of scenes you see will match the number you clicked on.

Untitled

See also the comment from @DingDongSoLong4 here #3813 - I was following it to the letter and it's explained more eloquently than I can.

Resolves #1961, resolves #3781


WithoutPants:
I've posted in the original issue regarding the UI elements, so I'm not going to cover that here.

I see that this the second attempt at this, but I'm sorry to say it might require a third. I should have caught this at the design point; I didn't review the design that @DingDongSoLong4 posted that you are following.

For the graphql, remember that we're wanting to be able to change the behaviour of the sorting, in addition to getting the filtered counts. I don't think that it's good design (or intuitive) to require the inclusion of the filtered counts fields to activate this behaviour. We'd want to document this behaviour in the graphql interface level, and I think it would be difficult to explain it.

I think that returning filtered only counts for any combination of filter criteria is probably overkill for what we need. I think the following is hopefully more elegant way to do this:

- add a `related_object` (I'm not sold on this name, so open to an alternative) field to the `PerformerFilterType`. The type will need to be a new input type `RelatedObject` or something, as follows:
```
input RelatedObject {
  type: ObjectType! # enum probably - Performer, Studio, etc
  id: ID!
}
```
- if `related_object` is not null, then the query builder will add the applicable `includes <object>` criterion to the filter, and the sorting behaviour will be changed to be based on the related object filter.
- the `_count` resolvers will need to be changed to check if the parent resolver is a `findPerformer` query, and if so check the performer filter for the related object field. If it's not null, then the count query will need to include a filter. It will probably need to call `QueryCount` instead of `CountByPerformerID`.
- this means that the sub-pages can pass a `related_object` input based on the parent object for the page, it removes the need to tie up the applicable filter criterion, and will remove the badge that shows up in all of the sub-pages.

I notice in this version that the `_count` filters aren't handled. I think that they probably should be handled, if we're thinking about the context in which the filters would be used. ie from the studio-performers panel, filtering on scene count = 1, but getting a bunch of results with the filtered counts that don't match up. I think it will be easier to do this with the `related_object` field.

I suspect that this will greatly simplify the changes to the query code at the sql layer. I don't think we should be combining the studio and appears with filter handler, and I don't think it will be necessary. 

Overall it will probably be a more complex change, but I think it will be cleaner. Let me know if you're up to this. If you can't, then I'll look at tackling it myself when I can.

dogwithakeyboard:
> Let me know if you're up to this

Amost certainly not but I'd like to give it a go anyway! Thanks for writing this up, it makes sense to me and is closer to what I was originally intending for the feature.

WithoutPants:
I've had a further think about this, and I wonder if the following may be cleaner still:

Add a `performers` resolver to the `Studio` type. This would accept `PerformerFilterType` and `FindFilterType` inputs, as per the `findPerformers` query. The resolver would do the same thing as `findPerformers` but with using the related object behaviour using the parent performer id. This would eliminate the need for a `related_object` field (which is admittedly a bit clunky). The studio-performers query graphql would then look something like:
```
findStudio(id: x) {
 performers(<filters>) {
  ...
 }
}
```

There'd be a similar resolver in the `Performer` type - `appears_with`.

PR #3901, opened 2023-07-11
Improve Makefile

DingDongSoLong4:
These are some more Makefile and general build improvements.

- `-buildmode=pie` is no longer used when building for arm32. Hopefully this fixes the new bootlooping issues, but I'll need someone with the actual hardware to test it.
- `phasher` is now included in the `build*` and `cross-compile-*` targets. I've added separate `stash*` and `phasher*` targets to build them separately.

And then I've updated the `DEVELOPMENT.md` readme, adding a few more commands and improving clarity.

I haven't changed any CI configs, because I'm not sure what exactly is going on there and I don't want to break anything. They'll need to be updated if we are to distribute `phasher` binaries.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,7 +68,7 @@ jobs:
     - name: Validate UI
       # skip UI validation for pull requests if UI is unchanged
       if: ${{ github.event_name != 'pull_request' || steps.cache-ui.outputs.cache-hit != 'true' }}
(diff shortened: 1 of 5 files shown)


WithoutPants (approved):
I've changed the `frontend` targets to `ui` instead to be consistent with the other targets. The shorter name is more convenient as well.

I've verified that this fixes the crash issue on raspi.

PR #3907, opened 2023-07-12
Chromcast support

Teda1:
I've included the fix to address the Uncaught TypeError mentioned in https://github.com/stashapp/stash/pull/3898.


internal/api/server.go
```diff
@@ -490,7 +490,7 @@ func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request) {
 	// The graphql playground pulls its frontend from a cdn
 	if r.URL.Path == playgroundEndpoint {
 		connectSrc += " https://cdn.jsdelivr.net"
-		scriptSrc += " https://cdn.jsdelivr.net"
+		scriptSrc += " https://cdn.jsdelivr.net http://www.gstatic.com https://www.gstatic.com"
```
DingDongSoLong4, line 493: I haven't actually run this branch, but these lines are only run in the graphql playground. Is this correct?

docker/build/x86_64/Dockerfile
```diff
@@ -2,7 +2,7 @@
 
 # Build Frontend
 FROM node:alpine as frontend
-RUN apk add --no-cache make
+RUN apk add --no-cache make git
```
DingDongSoLong4, line 5: Might have missed copying this change over to the CUDA Dockerfile as well (`docker/build/x86_64/Dockerfile-CUDA`)

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
```diff
@@ -22,6 +23,14 @@ import "./big-buttons";
 import "./track-activity";
 import "./vrmode";
 import cx from "classnames";
+// @ts-ignore
+import airplay from "@silvermine/videojs-airplay";
+// @ts-ignore
+import chromecast from "@silvermine/videojs-chromecast";
+import "@silvermine/videojs-chromecast/dist/silvermine-videojs-chromecast.css";
```
DingDongSoLong4, line 30: I believe these two CSS imports should be able to go in `styles.scss`: both plugins have a bit in their readmes about SCSS ([here](https://github.com/silvermine/videojs-airplay#using-the-css-and-images) and [here](https://github.com/silvermine/videojs-chromecast#using-the-css-and-images)).

internal/api/server.go
Teda1: I see I didn't pay close enough attention to that if statement when merging conflicts. This change should actually be moved to line 482. This is the bit that enables the `cast_sender.js` dependency located at  https://www.gstatic.com/cv/js/sender/v1/cast_sender.js?loadCastFramework=1 to load.

ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
Teda1: I went ahead and moved the imports to the SCSS file.

PR #3908, opened 2023-07-13
Fix docker build

NodudeWasTaken:
Move `make pre-ui` to after moving the Makefile


DingDongSoLong4:
Ohh that's why it was using `yarn` directly - my bad :(

PR #3913, opened 2023-07-15
Add allow adding avif as images

therandomguy42:
I know its a bad special case condition.

https://github.com/stashapp/stash/pull/3583 & https://github.com/stashapp/stash/pull/3765 allowed whitelisted avif to get added to the database, but they introduced the problem that avif gets detected as the movie format av1 wich does not work in Chrome (causes blackscreen video player) or Firefox(prints a message about invalid video codec).

This solution is not perfect either , ~~because you need to activate "create_image_clip_from_videos"~~ .


yoshnopa:
Sorry for not reacting sooner, I've tested the avif format without problems.
Interesting, it worked before I believe, but I cannot replicate it anymore. 
I've used this loop I got from the net, and I'm fairly certain it worked before in both an img and the video container in chromium:
https://www.file-upload.net/en/download-15168165/example.avif.html
This is still the case. You can quickly test it by opening the file in chromium, modifying the source from img to video, show controls and click play. Should yield the same results.
The reason I didnt look into this further is the lacking support for avif in general, with firefox not really ready and ffmpeg not capable of converting it this was only done as a proof of concept, nothing more.

yoshnopa:
I can confirm that a single image does not work in the video format, using example avifs:
https://raw.githubusercontent.com/link-u/avif-sample-images/master/hato.profile2.12bpc.yuv422.avif
I didnt bother with those due to the aforementioned reasons, sorry about that. I was hoping that it would behave like a loop with a single image, as the format is identical in every other way...

This should very likely work without this setting, you will need to set avif in image extensions, as they should be treated as images. Please elaborate on your settings if this is already the case so this can be done properly, because that should work, I'll help you there if necessary.

therandomguy42:
@yoshnopa not quite sure where you get that ffmpeg and firefox can't handle avif ( non-animated ), but https://caniuse.com/?search=avif says that general browser support is almost there and `ffmpeg -i image.jpg -c:v libaom-av1 -still-picture 1 image.avif` works with latesd ffmpeg ( on my windows 10 with the official av1 extension installed ).

I currently do not care about the animated version, i just want to have non-animated avif added as images to the stash. ( does not even need to be processed or anything, just added and viewable. like it was in version 16. ).

---

I've reset my local repo, the db, the config and only added the folders ( images folder as images only, video as video only. /* not that i like this differentiation */ ) and avif to image format.
Config: https://gist.github.com/therandomguy42/478372c88693d62a3387950360275517

So, what i've found is that FFProbe detects the file as an av1 video file.
Which means that pkg\file\image\scan.go:54 ( the error case ) does not get triggered and it gets classified as an video which then prompts the frontend to use <video/> instead of .

Thats why i've added this special condition to overwrite the classification to image.

With the changes they get displayed as images.

yoshnopa:
I find the missing support for the animated section to be rather important, as avif is, beside other things, handled as the successor of the gif format. 

Yes, I understand the reason for this commit and I can confirm that it is necessary, my text was a bit convoluted in that regard as I figured that out between the first and second comment. 
My request about your settings was in regard to your last sentence:

> This solution is not perfect either, because you need to activate "create_image_clip_from_videos" .

I don't think that should be the case with the current code, in fact, your gist has it deactivated right now, and for how I understood your comment it works with your commit/PR.

I wanted to offer help in case that this setting actually influences the result while the avif is entered as an image extension. Otherwise, without actually building it, I approve this commit for what it's worth.

therandomguy42:
@yoshnopa  Sorry, after i tested my changes again i noticed that the `create_image_clip_from_videos` isn't nessesary.
It just seemed like it was, while i was trying to learn & understand go code 😅

PR #3915, opened 2023-07-15
Model refactor

DingDongSoLong4:
This is a refactor to move the partial `Finder`, `Updater`, `Creator`, etc interfaces defined in the `files`, `scene`, etc packages into the `models` package.

The reasoning is that currently, if a package wants to make use of a smaller interface rather than the full `Reader`, `Writer` or `ReaderWriter` interface, it needs to add a dependency on the `files`, `scene`, etc package, which opens up potential circular dependency issues. Having to depend on an entire package just to import an interface I think is a bit unnecessary.

I've also moved most of the exported basic structs from the `file` package into `models` (`File`, `BaseFile`, `Folder`, `FileID`, etc). All the actual logic that operates on files (scanning, cleaning, etc) is still in the `file` package. This removes the last "internal" dependency that models has (the `file` package), meaning that `models` is now "free to import" - importing it will not cause circular dependency issues. This change allows for the import loop hack in `internal/manager/import_file.go` to be fixed.


Flashy78:
Nice, I was never quite sure what the different interface structures were for.

WithoutPants:
> This could be seen as against the "Go way" of defining interfaces at the use-site, but if we were to strictly follow that then the interfaces in the models package shouldn't exist at all and there should be interfaces in _every single_ package where the database is accessed (ie most packages). I think this is unnecessary code duplication - I don't see a real advantage to having the same method signatures defined all over the place.

The advantage is that we don't have to mock massive interfaces to test clients that only use a small subset of the reader interfaces. If we have duplicated interfaces, they should be commonly defined in either the applicable logic package (ie `scene`), or in models. Removing the interfaces altogether in favour of the large interfaces isn't a good direction imo.

> And then minor changes:
> 
> * I've moved `SceneFilenameParser` into the `scene` package, and the graphql `Timestamp` marshalling into `internal/api`. I think they fit better there than where they were previously.
> * `gqlgen` provides its own `Int64` marshaller so I've removed our custom one in favour of that.

While they are minor changes, they change a whole heap files, and thus should be put into a separate PR.

DingDongSoLong4:
Alright I've updated this to keep the more specific interfaces around, moving them to `models` instead of getting rid of them entirely.

PR #3916, opened 2023-07-16
Add Icons to tags if they have parent/child tags

elkorol:
Added calls to query parent and child tag length for tag, to see if they have any. And if so render the tag link with a FontAwesome icon, to give visual aid, of tag hierarchy in tag page. To see if parent tag of current tag, also has any parent tags. And for sub tag of current tag, to see if sub tag has any other sub tags


HijackHornet:
Good idea. Today I did this in a CSS only snippet by doing a ::pre if there is a subtag link as child. See community snippets for that. But this will makes it easier.

PR #3928, opened 2023-07-19
Create movies from scene edit dropdown

### LINKED ISSUE #1331: [Feature] Create movie from Scene edit dialog
To introduce the ability to manually create a new movie by entering a title in the Scene > Edit > Movie dialog box. This would bring Movies into parity with the Performer, Studio and Tag dialogs on the same page. Otherwise you have to go out to the Movies page > New then come back to Scene > Edit.

dogwithakeyboard:
Allows the creation of movies from the movie dropdown in the scene edit panel. Adds the option to enable/disable in the settings so this works exactly like tags, studios and performers.

I don't often create a movie but when I do I don't always remember to create the movie first before trying to assign the scene...

Closes #1331


WithoutPants (approved):

Stashing-fr:
https://github.com/stashapp/stash/pull/3928

Issue #3656, opened 2023-04-12
[Bug Report] tag image does not display on hover on the tag page (parent tags and sub tags)


xiosensei:
**Describe the bug**
tag image does not display on hover on the tag specific page. those tags are found in the parent tags and sub tags.

**To Reproduce**
1. Go to Tags
2. Select and click on a tag that has a parent or sub tags
3. Hover over one of the parent tags or sub tags

**Expected behavior**
similar logic to what happens when you hover over a tag on the scene page

PR #3931, opened 2023-07-21
Add Icons to tags if they have parent/child tags

elkorol:
As guided by DingDongSoLong4, I closed the last pull request to do with this linked [here](https://github.com/stashapp/stash/pull/3916) and started afresh to allow Parent tags and child tags to have a visual indicator on the tags page, if any have any parent or child tags themselves. Giving a visual cue without having to navigate away from current tag. To see if there is any hierarchy in those parent/child tags


ui/v2.5/src/components/Tags/TagDetails/TagDetailsPanel.tsx
```diff
@@ -41,9 +43,28 @@ export const TagDetailsPanel: React.FC<ITagDetails> = ({ tag }) => {
           <FormattedMessage id="parent_tags" />
         </dt>
         <dd className="col-9 col-xl-10">
-          {tag.parents.map((p) => (
+          {tag.children.map((p) => (
```
WithoutPants, line 46: Why was this changed? Shouldn't this be `parents`?

ui/v2.5/src/components/Tags/TagDetails/TagDetailsPanel.tsx
```diff
@@ -41,9 +43,28 @@ export const TagDetailsPanel: React.FC<ITagDetails> = ({ tag }) => {
...
+                    <span className="icon-wrapper">
+                      <span className="vertical-line">|</span>
+                      <OverlayTrigger
+                        placement="top"
+                        overlay={
+                          <Tooltip id="tag-hierarchy-tooltip">
+                            Explore tag hierarchy
+                          </Tooltip>
+                        }
+                      >
```
WithoutPants, line 61: I don't understand why this overlay is here. Clicking on the icon doesn't do anything different than clicking elsewhere on the badge. The text needs to be internationalised as well.

WithoutPants (changes requested):
This causes a regression of https://github.com/stashapp/stash/issues/3656 - the tag popover no longer shows. I think this probably needs to be added to the `TagLink` code instead.

elkorol:
Made requested changes. I couldn't get Tooltip to work without using OverlayTrigger. Where am I going wrong with this?

xx790:
This PR seems tangential to a feature I would like to have (thought about making a PR but no idea when I might get to it myself - not familiar with the project and with React).

Would love to have ids of the tag and all its parents as data attributes in HTML so I can make custom CSS easier.
Something like this:
```html
<div data-tag-id="1", data-tag-parents="2 3 4" ...>Tag</div>
```

Looking into changes, it seems like tag id might become available at least in some places, am I right?
Do you think it would be a good idea to add parent tags as well?

WithoutPants:
> This PR seems tangential to a feature I would love to have (thought about making a PR but no idea when I might get to it myself - not familiar with the project and with React).
> 
> Would love to have ids of the tag and all its parents as data attributes in HTML so I can make custom CSS easier. Something like this:
> 
> ```
> <div data-tag-id="1" data-tag-parents="2 3 4" ...>Tag</div>
> ```
> 
> Looking into changes, it seems like tag id might become available at least in some places, am I right? Do you think it would be a good idea to add parent tags as well?

This is possible, but would require changes to the client graphql queries to get the id numbers. It's beyond the scope for this PR.

Issue #3703, opened 2023-04-28
[Feature] Performer Page Restructure


elkorol:
**Is your feature request related to a problem? Please describe.**
Currently on the performer page, there is a lot of wasted space. The performer image is to the left, the diver and hide performer image chevron and then to the right, the username, flag, social media buttons. Edit, Auto Tag etc buttons. The tabbed navigation buttons for details, scenes, galleries etc. And the rest of it to the actual gallery image/video container.

With the performer image not hidden, on desktop for me, it limits the contents of the gallery container to 3 columns. And below the performer image is a ton of wasted space. Imo it's better when the gallery container displays 5 columns, but I still like to see the performer image and I think clicking the hid performer image chevron to have more gallery columns is not needed and we could have both.

**Describe the solution you'd like**
A restructuring the container div elements. Currently we have the div class of performer-tabs, which contains everything from the edit button down. I don't think it makes sense to include everything from the edit button down to be lumped in with the gallery grid container. But rather everything from the top, the username up until the element that shows the number of pages and total size of files, should be in a container apart from the gallery grid container. That way everything up until and including the element that shows the number of pages and total size of files, can be restructured and pushed to the right a bit. Allowing the performer image to always be above the gallery grid container, inline with all the other elements that are above it, allowing the gallery grid container to always to default to 5 columns, whilst still showing the performer image. 

Sure it sacrifices the possible size of display for performer image but imho always having 5 gallery grid columns is better. Could possibly keep a chevron that upon clicking would move the performer image to where it is now with 3 gallery grid columns. So people can see the performer image in its full size.

Restructuring could also even add the possibility of getting rid of the details navbar element entirely and just always show the details of the performer, I've included 3 images, the first is just what we have now, the second is just a possible mock-up of a potential layout and the third is just another mock-up with the details nav tab removed and the details always showing. With the option to always show the performer details, there may be an issue with long details text that need truncated. Most details fields would be fine like this, the only one that would cause an issue would be the actual details field. Which can sometimes contain a lot of text. As this is primarily a request for a restructuring primarily for desktop usage. There are probably several methods to circumvent too much text if choosing to always show details. One would be truncating text and showing the rest as a hover able tooltip, which I don't really like. Another would be specifically for the details field, for its text to be in a container that is scrollable. Or show all details fields like, gender, birth date etc, except the one for text details usually description, and for that just have a read more link to that text. Which could popup perhaps in a div element. 

**Additional context**
Current Structure
[...]

Issue #3773, opened 2023-05-26
[Feature] Persistent settings/toggles via cookies 


echo6ix:
**Is your feature request related to a problem? Please describe.**
Particular user toggles/settings on scene/images/gallery and settings pages are not remembered with navigating away from a page/refreshing a page.

**Describe the solution you'd like**
Make the following settings persistent using cookies:
- Scenes page
  - Zoom level
  - View
- Images page
  - Zoom level
  - View
- Gallery page
  - Zoom level
  - View
- Settings (tasks) page
  - Expand/collapse card toggles

MrX292:
for Zoom level and View you can set default filters

echo6ix:
> for Zoom level and View you can set default filters

Sure, but each time you want to change the zoom levels persistently you're going to set the default filter? It's a workaround imo, and should just be saved in the cookies.

DogmaDragon:
I agree on zoom levels and toggles, but I diagree with changed the default display mode. Maybe making a setting people can toggle, but the display modes persisting by default would create a lot inconvenience in my opinion.

elkorol:
Also the chevron in the Performers, Studios and Tags page that hides the left pane isn't persistent and would benefit from this

PR #3946, opened 2023-07-25
Details page redesign

Teda1:
Addresses the UI side of #3703

A video walkthrough of this change can be seen here:
https://discordapp.com/channels/559159668438728723/644934273459290145/1133237833839153153


Teda1:
My latest push includes the implementation of the show/hide more button. View the video below for a demonstration.
https://github.com/stashapp/stash/assets/72030708/2ccbcd07-d525-48f0-a818-98409e2c69a9

DogmaDragon:
Is the show/hide action persistent? It would be a huge inconvenience for me if I had to to click the button every time just to see full performer details when it's already the primary reason for me opening that page.

Teda1:
By default, everything but tattoos, piercings, details, tags, and StashIDs are shown until the button is hit. The main reason for the decision is to keep the clutter down, especially on smaller screens such as mobile devices. I can update the code so the state persists.

DingDongSoLong4:
I don't know if persisting the state would be a good idea - if you only want to see the extra data sometimes then having everything displayed at first open because the panel was expanded last time might be annoying. I think a simple option for whether the extra details are shown by default or not would be the most useful.

Teda1:
That option crossed my mind as well. I can go that route instead.

DogmaDragon:
Option to change default behaviour would be sufficiant for me.

Teda1:
The UI config option has been added.

Teda1:
My latest commit updates the Studio and Tag pages to display aliases consistently as the performer and movie pages. Here are screenshots of this from a mobile view.
image0
image1

DingDongSoLong4:
This looks amazing! The old sidebar really took up way too much unnecessary space.

I have noticed a couple (well maybe several) small things while clicking through pages, so here be a list:
- If a movie has no studio, the "Studio" label is still shown. This seems to be different to everywhere else where missing "attributes" are not shown.
- If a studio has no image, the image is just entirely missing instead of showing the default image. I'm not sure if this is something wrong with my setup, because the image link is there it just seems to be loading a zero-width image for some reason.
- On all pages except the performer one, when editing, the top of the first input label I think is a bit too close to the title/rating stars - some extra padding would look better imo.
- In the mobile view, I think extra padding above the title would also look better - it's a bit too close to the image.
- The image rounded corners seem to not be applying on the studio and tag pages when using a square image, but only when not editing - while editing the rounded corners are there.
- On all pages except movies, the image is vertically centered when not editing, but aligned to the top when editing. I'm unsure whether I prefer having the image at the top or centered. The centered image looks nicer when not editing, but if you click edit the image jumps up to the top. Here the top aligned image is nicer because it doesn't move when you click edit. I think the best would be to animate the image moving to the top when you click edit, but I'm not sure if that's possible (well with pure CSS that is).
- I think the dropdown button on the performer page and the new link buttons on the movie and studio pages are a bit too close to the title - the "halo" when you click the button overlaps the last letter of the title which doesn't look right to me. (sidenote: is there a reason you're using `display: contents` instead of just making the `<div>` a `<span>` like the other performer buttons? I don't know too much about fancy CSS stuff but it seems to look identical to me when I replace it with a span.)
- When you click "Edit", the background fade goes transparent for a split-second and you can see the title of the sticky header through it at the top.
- URL inputs on the edit pages are offset a bit to the right.
- When hovering over non-selected tabs and for selected tabs, the text is shifted up by a pixel or two. This seems to be caused by `align-items: center`.
- On all pages except performers, when editing, the inputs extend through the entire width of the screen. This is a bit annoying especially with wide monitors - it should be like the performer page where there's a margin on the right when using a larger screen.
- `rating100` needs to be removed from the yup schema and from `initialValues` in both the `StudioEditPanel` and `MovieEditPanel`. This prevents the form from resetting due to a change in initial values (thereby losing unsaved changes) if you set the rating using the stars, since they update the rating immediately using a mutation and bypass the form.

These are mostly small minor tweaks, some of which were probably present before these changes as well. I'd be happy to figure out fixes after this merges myself if you don't get to them.

Teda1:
Thanks for going through this change. I've been browsing the UI in my spare time and addressing minor issues like these as I find them. I'll start working my way through your findings later today.

Teda1:
All comments except the one about centered studio and tag images not transitioning smoothly to the top when switched to the edit page have been addressed. Before your comment, I spent a good amount of time looking into a solution but never landed on one.

Let me know if I missed anything else.

Teda1:
It makes sense that SVGs were the source of the width issue. I tweaked your code suggestion to use height, and weight instead of naturalWidth, and naturalHeight since those were showing legitimate values on my end, which resulted in the issue persisting. I also address the other items from your message.

My latest commit also added support for background images which can be toggled on or off. Currently, the image used in the background is the same as the image used in the detail image box. When support for providing custom background images is added, those images can be used there instead.

dogwithakeyboard:
This looks so much nicer than before! A couple of things I spotted against my database:

- The carriage returns and other whitespace in my details and synopsis fields is not displaying.
- Can we have the show/hide button on the other pages? I've scraped in some multi paragraph movie synopses and I think they need to be either hidden or truncated if I choose to hide details.

Lastly, perhaps it's time to stop displaying back cover on the Movie details page and only show it in the lightbox on click? It will be visually consistent with the other pages and it's commonly done by retailers.

QxxxGit:
Just quickly gave this a test. Not sure if it's intentional but when I have the options enabled to turn on backgrounds, it looks like the actual image (performer, tag, etc.) is somewhat transparent so you see the background bleed through.

Teda1:
@dogwithakeyboard I've gone ahead and added the show/hide button to the movie, studio, and tag pages. On those pages, the details fields will only show at most 3 lines of text until the show all details button is hit. I've also made the update to show whitespaces in the value span.

Regarding the last point, I like the practicality of showing both the front and back cover when both are provided. If no back cover is provided, then only the front cover is displayed. I don't think showing both when provided hurts the visual consistency. Keeping both on screen is also necessary when adding or modifying covers on the edit page.

Teda1:
@QxxxGit, I misunderstood the image you were referring to. I didn't notice the background image was bleeding on top of the details container. I've gone ahead and pushed a fix.

HereIam80:
So I tried this out. First of all thanks for all your work.
While this is way more practical for browsing your library and I am really linking the details up on top, I am kinda sad, that there is no view with a bigger performer picture anymore. I know that I can click on the performer image to view it but that is not really the same. I would have liked having one or two more performer images so the image would change on mouseover but since the image is so small now, I don't think it would look good anymore. I use full body pictures btw ... it is probably fine with just head pictures. Maybe showing the full picture on mouseover would be doable? Or maybe still having a full view of all the details view as an option somewhere?

Obviously people will have different opinions about this since @DingDongSoLong4 for one calls the bigger image "unnecessary space". ;-) Maybe if you go for efficiency but for the aesthetics I really like seeing those images, which I carefully handpicked for each performer.  So I don't know if this has been discussed before or not but this is just my first impression.

Teda1:
@HereIam80 I mentioned this earlier in Discord, but the default size of the image can easily be changed via CSS for users who want a larger image. See the following CSS for example:
```
#performer-page .detail-header-image img {
    max-width: {some-value};
}
```

This will increase the default size of the performer image when details are expanded. You can view this in action with the video https://github.com/stashapp/stash/assets/72030708/2ed5ae70-2da4-401a-a012-b1ad2e2c5642 . The max width used in the video was 22rem. Of course, you can also use CSS to make the image that size regardless of if all details were expanded.

echo6ix:
> So I tried this out. First of all thanks for all your work. While this is way more practical for browsing your library and I am really linking the details up on top, I am kinda sad, that there is no view with a bigger performer picture anymore. I know that I can click on the performer image to view it but that is not really the same. I would have liked having one or two more performer images so the image would change on mouseover but since the image is so small now, I don't think it would look good anymore. I use full body pictures btw ... it is probably fine with just head pictures. Maybe showing the full picture on mouseover would be doable? Or maybe still having a full view of all the details view as an option somewhere?
> 
> Obviously people will have different opinions about this since @DingDongSoLong4 for one calls the bigger image "unnecessary space". ;-) Maybe if you go for efficiency but for the aesthetics I really like seeing those images, which I carefully handpicked for each performer. So I don't know if this has been discussed before or not but this is just my first impression.

I agree with @DingDongSoLong4 because there's already a dedicated image section per performer page that can be utilized for exactly this purpose. 

This is outside the scope of the topic, but if Stash implements the following things you could reproduce a more flexible effect as the incumbent design
- Create a global option for users to select the default tab when opening the performer dashboard page --> https://github.com/stashapp/stash/issues/3961
- Implement star ratings for images --> https://github.com/stashapp/stash/issues/3960
- Make zoom levels persistent for grid view --> #3773 

With these three features the user could then
- Set the performer page to always open the image tab first
- Set a default filter to sort images based on highest rated
- Set the default zoom levels to size they desire

Boom, you have your favorite large scale performer images back on the performer landing page

HereIam80:
@Teda1 
Thanks for the reply. I totally missed that conversation regarding the css. This will however waste much space next to the image as it is just empty. I will definitely use it regardless.

By the way when I scroll down my performer list and click on one of them, the performer page doesn't scroll up to the top. I don't think this is intended, is it?

@echo6ix 
While I know what you mean, that is not something I would do. It is not a matter of what I see first but what I see while browsing and the overall look and feel. Having a smaller picture on the detail page than on the performer selection page just seems wrong to me.

As I said there are probably many different opinions on this and don't get me wrong I still like most of the changes I just want to have the full body picture look good.

Teda1:
> Thanks for the reply. I totally missed that conversation regarding the css. This will however waste much space next to the image as it is just empty. I will definitely use it regardless.

No worries. This is precisely the trade-off and the argument for compact images. As @DingDongSoLong4 mentioned, the previous implementation took up a lot of space and provided no easy compact solution. This implementation provides a more flexible solution for both types of users. Pursuing larger images inherently means trading space efficiency. You see this trade-off being made on site who use larger images:
Screenshot 2023-07-29 195524
Screenshot 2023-07-29 195201

> By the way when I scroll down my performer list and click on one of them, the performer page doesn't scroll up to the top. I don't think this is intended, is it?

Can you provide a video of what you are seeing? I'm not able to reproduce the issue based on your description.

AdultSun:
I agree with @HereIam80 about missing the full-size portrait on the performer page, actually. It's also more difficult to parse the performer's details at a glance when they're compressed horizontally instead of stacked into a single-column list like the current layout. I'm sure we'll all get used to finding everything after using it for awhile, but I expect I'm not the only one who finds it easier to skim through something in only one direction (top to bottom) instead of two directions (adding left to right).

To be clear, I also really like the changes and styling implemented here so I'm not asking to scrap all of that in favor of the status quo or anything. And from what I've seen of the reaction on Discord, it seems to me the majority of users prefer this approach. So consider this a pitch for the next evolution of the performer page's design in the next PR rather than a request to change it even further within this PR.

I think what's bothering me about this design is that we have a view that focuses on the performer's content but not the performer themselves anymore. What really illustrates that fact — and I think also presents a solution — is that the "expand for more details" is barely different from the compressed summary view. It adds a few extra fields, widening the overall view just a little, but that's really it. It feels like a wasted opportunity to me. My preference would be for the "see more" button to transition into a much wider, full-screen display. I have no clue what the technical constraints would be for something like this, but I'm imagining the performer's image expanding to it's previous size and the detail fields expanding into a less cramped single-column view similar to the old format.

The above approach would give us:
1. The hybrid view as the default, with a quick summary of the most important details at the top and content below
2. The content view after scrolling down, with the details compressing even further into the narrow header at the top
3. A new detailed view after clicking the chevron, expanding to reveal all of the detail fields with a layout and space to more easily parse everything at a glance

Personally, I think that approach would give us all of the improvements of this PR without sacrificing anything for users like myself and Herelam80 who will miss the old layout. The new default view you've built here does feel like the closest to "the best of both worlds" possible, plus it already has a built-in view for (nearly) content only. I'm just hoping there will be a way to have a "performer-only" view alongside the other two moving forward. If nothing like that makes it into this PR (and I expect it won't since you've put so much work into it already and I suspect I'm asking for a lot here), I'd be willing to write this up again into a new feature request.

Teda1:
@AdultSun could you provide a mock up of what you had in mind for me to take a look at in the morning?

AdultSun:
Uh, not really? Sorry, I'm not good enough with image editors or any of the other UI concept design programs I've seen others use around here to make anything very helpful. I could give it a try if someone could point me in the right direction.

Basically I was just hoping the chevron button would resolve to a view similar to the current details/edit screen: Full-length image on the left, single-column of details on the right with more padding in-between each field to separate them. Use the entire space to let everything breather for greater readability instead of trying to eliminate as much empty space as possible. 

The exact details in the styling aren't important to me but I imagine you'd want it similar to what this PR establishes (meaning fonts, labels, etc.). But the order should be the same as the order of the edit page. I spend so much time on the edit page that it's always bothered me that the current details view in Stash (the old one not this PR) is in a different order than the edit page. It always mixes me up and makes it harder to find things on the details tab.

I also expect that having a view dedicated to extra space for the details would help with future-proofing. We wouldn't have to add new fields/features to the compact summary of this PR if it makes more sense to use it only in an expanded "details" view instead where there's more room.

Teda1:
I think I have the visual in my head, but I'm struggling to differentiate between what you're describing and the edit page. It sounds like the edit page already serves this purpose. Most of this is likely already achievable with CSS using this layout. I'll mock something up and post it in Discord.

The concept can likely be flushed out more after this pr once more people have gotten their hands on this and have spent more time with the change.

Teda1:
As discussed on Discord, I've pushed to enable the more table-like view when showAllDetails is enabled. Images will no longer center in the expanded view, and the detail titles will end with a colon. See my Discord post for the video https://discordapp.com/channels/559159668438728723/644934273459290145/1135278697906176110.

WithoutPants (approved):

PR #3955, opened 2023-07-27
Fix scene marker issues

DingDongSoLong4:
This mainly fixes two bugs that came up in Discord relating to scene markers.

- If the `SceneMarkerForm` component is re-rendered while the scene is being played (ie when `getPlayerPosition()` returns a value) and a new marker is being created, the `initialValues` would change causing the form to be reset.
- Updating a scene marker with a blank title would cause a NOT NULL constraint error. An empty title is now saved as an empty string rather attempting to store a `NULL`.
- I also found and fixed another NULL-related issue, where `NULL` names in the performers or tags tables give a NULL conversion error when they are scanned into the `performerRow` or `tagRow` structs.

As mentioned in the TODO's I've added, the scene marker and gallery chapter DB schemas should be updated to remove the `NOT NULL` constraints, and the performer and tag schemas should be updated to add a `NOT NULL` constraint. For both performers and tags the UI already enforces that names should be non-empty.

I haven't added a migration to make these changes here, as they are relatively minor. They should probably be added to the next migration.

### DIFF
--- a/pkg/sqlite/movies.go
+++ b/pkg/sqlite/movies.go
@@ -26,7 +26,7 @@ const (
 
 type movieRow struct {
 	ID       int         `db:"id" goqu:"skipinsert"`
-	Name     zero.String `db:"name"`
+	Name     string      `db:"name"`
 	Aliases  zero.String `db:"aliases"`
 	Duration null.Int    `db:"duration"`
 	Date     NullDate    `db:"date"`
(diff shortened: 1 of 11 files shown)


pkg/sqlite/movies.go
```diff
@@ -26,7 +26,7 @@ const (
 
 type movieRow struct {
 	ID       int         `db:"id" goqu:"skipinsert"`
-	Name     zero.String `db:"name"`
+	Name     string      `db:"name"`
```
WithoutPants, line 29: This introduces a regression where movies with blank names can be created via graphql. What is the purpose of this change? I don't see it documented in the description.

pkg/sqlite/studio.go
```diff
@@ -28,7 +28,7 @@ const (
 
 type studioRow struct {
 	ID        int         `db:"id" goqu:"skipinsert"`
-	Name      zero.String `db:"name"`
+	Name      string      `db:"name"`
```
WithoutPants, line 31: Same issue as for movies.

PR #3966, opened 2023-07-30
First-pass studio scraper support

StashPRs:
This is a first-pass draft attempt at scraping studios from StashBox (and community scrapers). This is mostly intended to allow users to easily scrape the URL and image for a studio in their stash instance. 

What I've got so far:
image

### DIFF
--- a/graphql/documents/data/scrapers.graphql
+++ b/graphql/documents/data/scrapers.graphql
@@ -65,7 +65,7 @@ fragment ScrapedScenePerformerData on ScrapedPerformer {
 fragment ScrapedMovieStudioData on ScrapedStudio {
   stored_id
   name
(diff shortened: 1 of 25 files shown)


MrX292:
does #3510 not kind of do that?

Flashy78:
It looks like this is building support for scrapers for studios instead of the tagger interface, so it is slightly different in that you could support scrapers other than a Stashbox. Although I'm not quite sure the use case for that.

timo95:
> Although I'm not quite sure the use case for that.

Self published content where studios are just some websites user profiles. I actually used API calls in a python scraper before to add studios, so I would love to have this feature.

Gykes:

PR #3996, opened 2023-08-03
fix separator being used around null fields

Teda1:
This pull request fixes the issue where separators on the sticky headers are shown around null fields.


Teda1:
My latest push updated the code in `apple.ts` to use UAParser

PR #4000, opened 2023-08-04
Multiple image URLs

WithoutPants:
Similar change to #3852 but for images. Most of the copy was a copy/paste across to the image type.


pkg/sqlite/migrations/49_image_urls.up.sql
```diff
@@ -0,0 +1,70 @@
+PRAGMA foreign_keys=OFF;
+
+CREATE TABLE `image_urls` (
+  `image_id` integer NOT NULL,
+  `position` integer NOT NULL,
+  `url` varchar(255) NOT NULL,
```
its-josh4, line 6: Why limit to 255 chars? (Not that it matters anyways, since [SQLite ignores them](https://www.sqlite.org/datatype3.html))

PR #4006, opened 2023-08-07
Scene Duplicate Checker UI & Feature Improvement

DrDaveUK:
Feature Updates:

 - Added Selection dropdown, to enable the user to select three different types of files; largest, youngest and oldest

- Added a checkbox to ensure that the codecs are the same within the group
  - If checked, only duplicate groups with the same codec for all files will be processed (otherwise the group is skipped). 
    - Default : checked 
- If the checkbox or selection dropdown value is then changed, the selection is reset and then applied to ensure that there is always a file not checked.

UI Updates:
 * A white line separates the groups of files to make it easier to see each group
 * When the file is selected for deletion the title is made bold, with a red strikethrough and the thumbnail also has a red border
 * Pagination now allows a selection up to 1500 per page, but will only show up to the largest page size which is smaller than the actual number of duplicates, 
   * e.g. 282 duplicates will show a page size of 250, and not include a page size of 500 to select


WithoutPants:
I cleaned up the code a little and changed the select box to be a button, since it's more of an action than a state. I also changed the available size options to include the first option > number of scenes, since otherwise there's no way to show all of the elements without pagination - it also made the select box empty if you had less than 10 results.

PR #4011, opened 2023-08-08
Friendlier Scraper Error Message

Flashy78:
A common help request in Discord is users coming with the error message `could not unmarshal json from script output: EOF` after they run a scraper and wondering what to do next.

This leaves that technical message writing to the log, but presents a more useful message to the user, telling them to go look at the logs to see what actual error happened with the scraper.
There's also a hack to linkify the `check the logs` text in the Toast notification to allow the user to go to the logs in case they aren't sure. This will break once server error messages become translatable.
scraper_error

It also defaults the error toasts to show for 10 seconds instead of 3. The assumption is that if an error did happen, you want a few seconds to be able to read what it was (or click the link to go to the logs).

I looked into copying the scraper `stderr` to display the actual error in the toast, but it didn't make sense since python can generate very large errors. I thought it better to make sure people are able to get to the logs page.

### DIFF
--- a/pkg/scraper/script.go
+++ b/pkg/scraper/script.go
@@ -108,7 +108,7 @@ func (s *scriptScraper) runScraperScript(ctx context.Context, inString string, o
 		if lenientErr != nil {
 			// The error is genuine, so return it
 			logger.Errorf("could not unmarshal json from script output: %v", lenientErr)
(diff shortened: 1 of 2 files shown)


Gykes:
Not sure if this one is still valid but might be good for me to pull in if I am going to be giving some love to the scraper system

PR #4016, opened 2023-08-09
Auto Tag Confirmation 

### LINKED ISSUE #3909: [Feature] Auto tag confirmation dialog and other warnings to mitigate misuse
A minority (or more?) amount of users do not understand what the *Auto Tag* and *Identify* features do yet activate these features anyway. Some also accidentally fat finger the auto-tag button. Both features can critically alter a library's tags if not used appropriately. We need a simple solution to mitigate this from happening.

- Add a confirmation dialog
  - Any time the user initializes an auto tag anywhere in the UI they are prompted with a confirmation dialog
  - The dialog explains the potential for false positives and how auto tag functions
  - The dialog contains a checkbox to hide the reminder in the future to avoid annoying users with excessive prompts 
  - image

- In _Settings > Tasks > Auto tag_ modify the following:
  - Have performer and studio auto tag toggles disabled by defult
  - Modify the current description under the Auto tag header
  - 
image
 
- _Auto tag_ and _Identify_ dialog warning and task description should be educational (aimed at the least informed users) but as concise as possible (as not to deter users from reading it). For example:
  - `Auto Tag analyzes your file path names and attempts to find similarities with your existing metadata (tags, studios, or performers). This process is irreversible and may produce undesirable false positives, particularly with non-unique tags and names.`

A user now has to be willfully negligent to proceed in spite of those warnings and not understand how auto tag works

- Hiding the auto tag feature or demoting it from it's incumbent position in the UI ❌
  - This doesn't resolve the issue. It is akin to security though obfuscation, and will only alienate users that use the feature. Users who find it would still be culpable of not understanding how the process works.
- Removing the feature altogether ❌
  - Will alienate users that use the feature. This was an original core feature and not all libraries are amenable to the tagger.

---

EDIT: **Changing subject to specify the confirmation dialog for both *auto tag* and *identify***

---

EDIT 2: **The confirmation dialog warning should be clear and concise.** For new users it must explain what *Auto Tag* does, why they should be cautious, and in simple terms with an emphasis on brevity.

Flashy78:
Resolves #3909 

Implements most of the suggestions from the ticket, except for a button to hide the dialogue in the future. I didn't see any similar functionality in Stash, and don't want to incorrectly build new features.

- Updated descriptions for Auto Tag and Identify with more explanation
- Auto Tag buttons should all have the confirmation dialog
- For new users, in the Settings their Performers and Studios will be disabled by default for Auto Tag
- For new users, removed Auto Tag as a default source from Identify

Other

- Refactored `DirectorySettingDialog.tsx` to allow customization and potentially hide the directory picker. This was copied from how the Clean task dialogs were doing their own directory picking, and the Clean task was refactored to use the updated `DirectorySettingDialog.tsx`. So now you can have a directory picking dialog that doesn't show you any directories, but all of the directory picking code is no longer duplicated for the Clean task.
- Added slight padding to the bottom of the directory listing to remove an inline scroll bar that was showing on my system

autotag2
autotag4
autotag3

### DIFF
--- a/ui/v2.5/src/components/Settings/Tasks/DataManagementTasks.tsx
+++ b/ui/v2.5/src/components/Settings/Tasks/DataManagementTasks.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react";
 import { FormattedMessage, useIntl } from "react-intl";
-import { Button, Col, Form, Row } from "react-bootstrap";
+import { Button, Form } from "react-bootstrap";
 import {
   mutateMigrateHashNaming,
   mutateMetadataExport,
@@ -21,110 +21,11 @@ import { SettingSection } from "../SettingSection";
 import { BooleanSetting, Setting } from "../Inputs";
 import { ManualLink } from "src/components/Help/context";
 import { Icon } from "src/components/Shared/Icon";
-import { ConfigurationContext } from "src/hooks/Config";
-import { FolderSelect } from "src/components/Shared/FolderSelect/FolderSelect";
 import {
-  faMinus,
-  faPlus,
   faQuestionCircle,
   faTrashAlt,
 } from "@fortawesome/free-solid-svg-icons";
-
-interface ICleanDialog {
-  pathSelection?: boolean;
-  dryRun: boolean;
-  onClose: (paths?: string[]) => void;
-}
-
-const CleanDialog: React.FC<ICleanDialog> = ({
-  pathSelection = false,
-  dryRun,
-  onClose,
-}) => {
-  const intl = useIntl();
-  const { configuration } = React.useContext(ConfigurationContext);
-
-  const libraryPaths = configuration?.general.stashes.map((s) => s.path);
-
-  const [paths, setPaths] = useState<string[]>([]);
-  const [currentDirectory, setCurrentDirectory] = useState<string>("");
-
-  function removePath(p: string) {
-    setPaths(paths.filter((path) => path !== p));
-  }
-
-  function addPath(p: string) {
-    if (p && !paths.includes(p)) {
-      setPaths(paths.concat(p));
-    }
-  }
-
-  let msg;
-  if (dryRun) {
-    msg = (
-      <p>{intl.formatMessage({ id: "actions.tasks.dry_mode_selected" })}</p>
-    );
-  } else {
-    msg = (
-      <p>{intl.formatMessage({ id: "actions.tasks.clean_confirm_message" })}</p>
-    );
-  }
-
-  return (
-    <ModalComponent
-      show
-      icon={faTrashAlt}
-      disabled={pathSelection && paths.length === 0}
-      accept={{
-        text: intl.formatMessage({ id: "actions.clean" }),
-        variant: "danger",
-        onClick: () => onClose(paths),
-      }}
-      cancel={{ onClick: () => onClose() }}
-    >
-      <div className="dialog-container">
-        <div className="mb-3">
-          {paths.map((p) => (
-            <Row className="align-items-center mb-1" key={p}>
-              <Form.Label column xs={10}>
-                {p}
-              </Form.Label>
-              <Col xs={2} className="d-flex justify-content-end">
-                <Button
-                  className="ml-auto"
-                  size="sm"
-                  variant="danger"
-                  title={intl.formatMessage({ id: "actions.delete" })}
-                  onClick={() => removePath(p)}
-                >
-                  <Icon icon={faMinus} />
-                </Button>
-              </Col>
-            </Row>
-          ))}
-
-          {pathSelection ? (
-            <FolderSelect
-              currentDirectory={currentDirectory}
-              setCurrentDirectory={(v) => setCurrentDirectory(v)}
-              defaultDirectories={libraryPaths}
-              appendButton={
-                <Button
-                  variant="secondary"
-                  onClick={() => addPath(currentDirectory)}
-                >
-                  <Icon icon={faPlus} />
-                </Button>
-              }
-            />
-          ) : undefined}
-        </div>
-
-        {msg}
-      </div>
-    </ModalComponent>
-  );
-};
+import { DirectorySelectionDialog } from "./DirectorySelectionDialog";
 
 interface ICleanOptions {
   options: GQL.CleanMetadataInput;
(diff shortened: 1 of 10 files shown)


DingDongSoLong4:
Very nice, this will definitely help prevent accidental presses or unaware users from activating these irreversible features.

I have just two comments:
- It would be nice if the Auto Tag dialogs on the studios, performers and tags pages showed different messages, explaining that it will Auto Tag for that specific studio/performer/tag only.
- Since there's no big warning icon, I think adding a simple "Are you sure you want to continue?" message to the Auto Tag confirmation dialogs would help to emphasize that the action is irreversible.

ui/v2.5/src/components/Settings/Tasks/DataManagementTasks.tsx
```diff
@@ -232,6 +133,45 @@ export const DataManagementTasks: React.FC<IDataManagementTasks> = ({
...
+            setDialogOpen({
+              clean: false,
+              cleanAlert: false,
+            });
+          }}
+        />
+      );
+    }
+    return;
+  }
```
WithoutPants, line 173: Why couldn't this have been put into the existing `CleanDialog` instead of removing the `CleanDialog` altogether?

ui/v2.5/src/components/Settings/Tasks/DirectorySelectionDialog.tsx
```diff
@@ -15,12 +16,29 @@ interface IDirectorySelectionDialogProps {
...
 }
 
 export const DirectorySelectionDialog: React.FC<
   IDirectorySelectionDialogProps
-> = ({ animation, allowEmpty = false, initialPaths = [], onClose }) => {
+> = ({
+  animation,
+  allowEmpty = false,
+  initialPaths = [],
+  allowPathSelection = true,
```
WithoutPants, line 34: This entire concept is a bit broken from a design perspective. The component is a `DirectorySelectionDialog`, but we have a flag to disable the entire purpose of the dialog? I think there's a better way to do this.

ui/v2.5/src/components/Settings/Tasks/LibraryTasks.tsx
```diff
@@ -74,17 +74,17 @@ export const LibraryTasks: React.FC = () => {
...
   });
 
   const [scanOptions, setScanOptions] = useState<GQL.ScanMetadataInput>({});
   const [autoTagOptions, setAutoTagOptions] =
     useState<GQL.AutoTagMetadataInput>({
-      performers: ["*"],
-      studios: ["*"],
+      performers: [],
+      studios: [],
       tags: ["*"],
```
WithoutPants, line 88: The reason for specifying tags only as the default should be documented here. Right now, it just looks arbitrary.

ui/v2.5/src/components/Settings/Tasks/LibraryTasks.tsx
```diff
@@ -210,19 +210,36 @@ export const LibraryTasks: React.FC = () => {
...
+              autoTag: false,
+              autoTagAlert: false,
+            });
+          }}
+        />
+      );
     }
-
-    setDialogOpen({ autoTag: false });
+    return;
```
WithoutPants, line 242: Redundant

WithoutPants (changes requested):

ui/v2.5/src/components/Settings/Tasks/DataManagementTasks.tsx
Flashy78: First I duplicated the `CleanDialog` code over into `LibraryTasks.tsx` as an `AutoTagDialog`.
That was almost identical code, and now a third instance of the `DirectorySelectionDialog` code. So I pulled `CleanDialog` and `AutoTagDialog` out into a new component file.
This new file was almost identical to `DirectorySelectionDialog`, except for the extra styling options, message and option to hide the directory selection.

I thought about adding the styling options to `DirectorySelectionDialog` and replacing the directory code in my new file with that component, but I would have to deconstruct `DirectorySelectionDialog` because it already has a `<ModalComponent>` that I couldn't wrap in my higher level `<ModelComponent>`.

I figured since I was already going to add all the styling options to it, adding message and the option to totally hide it was less complicated and less error prone than trying to refactor `DirectorySelectionDialog` and pull out the directory selection code into a sub component that could be imported into an `OptionalDirectorySelectionDialog` component.

I could do that if you think it would be a better approach?

Flashy78:
>     * It would be nice if the Auto Tag dialogs on the studios, performers and tags pages showed different messages, explaining that it will Auto Tag for that specific studio/performer/tag only.
> 
>     * Since there's no big warning icon, I think adding a simple "Are you sure you want to continue?" message to the Auto Tag confirmation dialogs would help to emphasize that the action is irreversible.

Added a line confirming which object you just clicked Auto Tag for.

Took your suggestion and reworked the text into a description and warning, so I could add a warning icon along with the warning text. I've updated the PR description with screenshots.

ui/v2.5/src/components/Settings/Tasks/DataManagementTasks.tsx
WithoutPants: > First I duplicated the CleanDialog code over into LibraryTasks.tsx as an AutoTagDialog.
> That was almost identical code, and now a third instance of the DirectorySelectionDialog code. So I pulled CleanDialog and AutoTagDialog out into a new component file.
> This new file was almost identical to DirectorySelectionDialog, except for the extra styling options, message and option to hide the directory selection.

Please do this instead for now. We can address the duplicate code later if needed.

echo6ix:
When I proposed this in https://github.com/stashapp/stash/issues/3909 I never considered a potentially better idea that's already in use.

Instead of a simple warning modal, take inspiration by the deletion modal and provide the user with a list of affected files as a verification layer.

- Autotag is clicked
- Autotag modal window opens
- Modal shows all files that would be affected by the autotag operation
- Apply button commits the autotag operation

```
┌────────────────────────────────────────────────────────┐
│ {$warn_icon}                       Autotag (4 items)   │
├────────────────────────────────────────────────────────┤
│                                                     |s │
│   • path/filename1.jpg                              |c │
│   • path/filename2.jpg                              |r │
│   • path/filename3.jpg                              |o │
│   • path/filename4.jpg                              |l │
│                                                     |l │
│                                                     |b │
│                                                     |a │
│                                                     |r │
│                                                        │
├────────────────────────────────────────────────────────┤
│                [ Apply ]        [ Cancel ]             │
├────────────────────────────────────────────────────────┤
```
This is actually useful now, to new and experienced users who might otherwise just consider a boilerplate modal annoying.

The warning string become redundant now, because it's not some mysterious process happening in the background with unknown consequences. The consequences are enumerated in a reviewable list.

Gykes:
I am currently implementing something for this but just keeping it simple. I will add in a warning modal that pops up when auto tag is used to further prevent accidental clicks. With how potentially destructive auto tag is having it just be a click and go bothers me.

WithoutPants:
Superceded by #6735

PR #4026, opened 2023-08-13
Scene filters save

elkorol:
- Added database migration to include a scene filters table
- Added GraphQL to query and mutate scene video filter values such as Brightness, Contrast Etc
- Moved logic concerning Scene.tsx set sliderRanges and functions that update video filters to separate utility file, so update scene filters, turns out I didn't need to seperate the logic but if maintainers are happy where I've moved it, cool. Otherwise I can move it back.
- Added logic to apply scene filters on page load and scene changes


elkorol:
I think ScenePlayer.tsx and Scene.tsx could be ignored the addictions there seem to be from when I git pulled and merged with my branch before the pr. I'm not sure why that code is showing as an addition if it's already part of the main code. But I can remove it if needed.

pkg/models/scene.go
```diff
@@ -52,6 +52,8 @@ type SceneFilterType struct {
 	// Filter by duration (in seconds)
 	Duration *IntCriterionInput `json:"duration"`
 	// Filter to only include scenes which have markers. `true` or `false`
```
DingDongSoLong4, line 54: typo

pkg/sqlite/migrations/49_postmigrate.go
DingDongSoLong4, line 1: This postmigration is redundant? Both SQL queries are already run in `49_scene_filters.up.sql` so this achieves nothing.

graphql/documents/mutations/scene-filter.graphql
```diff
@@ -0,0 +1,77 @@
+mutation SceneFilterCreate(
```
DingDongSoLong4, line 1: I know you've copied this from scene markers, but this would be much cleaner if it accepted a `($input: SceneFilterCreateInput!)`, like the scene mutations, instead of each field individually. The same goes for `SceneFilterUpdate`.

graphql/schema/types/filters.graphql
```diff
@@ -138,6 +138,19 @@ input PerformerFilterType {
...
 }
 
+input SceneFilterFilterType {
+  """Filter by creation time"""
+  created_at: TimestampCriterionInput
+  """Filter by last update time"""
+  updated_at: TimestampCriterionInput
+  """Filter by scene date"""
+  scene_date: DateCriterionInput
+  """Filter by cscene reation time"""
```
DingDongSoLong4, line 148: Some typos just in the comments

internal/api/resolver_mutation_scene.go
```diff
@@ -643,6 +643,176 @@ func (r *mutationResolver) SceneMerge(ctx context.Context, input SceneMergeInput
...
+	}); err != nil {
+		return false, err
+	}
+
+	r.hookExecutor.ExecutePostHooks(ctx, filterID, plugin.SceneFilterDestroyPost, id, nil)
+
+	return true, nil
+}
+
+func (r *mutationResolver) changeFilter(ctx context.Context, changeType int, changedFilter *models.SceneFilter) error {
```
DingDongSoLong4, line 774: This will need to be refactored to remove the `changeFilter` method, like I did to `changeMarker` in #3945 - the changeType "enum" no longer exists.

pkg/scene/delete.go
```diff
@@ -185,6 +189,13 @@ func (s *Service) deleteFiles(ctx context.Context, scene *models.Scene, fileDele
 	return nil
 }
 
+func DestroyFilter(ctx context.Context, scene *models.Scene, sceneFilter *models.SceneFilter, qb FilterDestroyer) error {
```
DingDongSoLong4, line 192: I'm not sure I see the point of this - it's a barely-there wrapper around `qb.Destroy`.

pkg/sqlite/scene_filter.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	if findFilter == nil {
+		findFilter = &models.FindFilterType{}
+	}
+
+	query := qb.newQuery()
+	distinctIDs(&query, sceneFilterTable)
+
+	if q := findFilter.Q; q != nil && *q != "" {
+		searchColumns := []string{"scene_filters.id", "scenes.id"}
```
DingDongSoLong4, line 252: This is searching on ids, which is a bit strange. If you're going to use the filter `Q` at all, it should probably be on the scene title and not ids.

ui/v2.5/src/components/Scenes/SceneDetails/SceneVideoFilterPanel.tsx
```diff
@@ -21,316 +55,366 @@ type SliderRange = {
...
-      );
-      feFuncA.setAttribute("type", "gamma");
-      feFuncA.setAttribute("amplitude", "1.0");
-      feFuncA.setAttribute("exponent", "1.0");
-      feFuncA.setAttribute("offset", "0.0");
-      feComponentTransfer.appendChild(feFuncA);
-
-      videoFilter.appendChild(feComponentTransfer);
+    // Diabling checks because only props.scene is needed to trigger hook
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
DingDongSoLong4, line 250: You really don't want to be disabling `exhaustive-deps`. If you want something to not trigger the hook, use `useRef`.

ui/v2.5/src/core/StashService.ts
DingDongSoLong4, line 1: This will also need to be reworked after #3912.

ui/v2.5/src/locales/da-DK.json
DingDongSoLong4, line 1: Not sure where you got these translations from, but unless you do actually speak all these languages, you should probably stick to changing the base `en-GB` strings.

ui/v2.5/src/models/list-filter/criteria/factory.ts
```diff
@@ -119,8 +120,10 @@ export function makeCriteria(
       );
     case "favorite":
       return new FavoriteCriterion();
-    case "hasMarkers":
+    case "hasFilters":
```
DingDongSoLong4, line 123: typo - "hasFilters" and "hasMarkers" are the wrong way around

pkg/models/scene.go
elkorol: thanks changed

yikesman420:
Any updates on this?

WithoutPants:
Closing in favour of #4520

Issue #3731, opened 2023-05-13
[Feature] Add option to mark scene as organized when identifying through Tagger view


grabolento:
I think it would be practical if there was an option, when tagging scenes through the Tagger view, to immediately mark them as organized the moment you click the "Save" button.

I'm not sure, however, what would be the best way to do this, in terms of UX. It could be a second button with the label "Save and mark as organized"; it could be an "Organized" checkbox, like in the Edit Scene window from Grid or List view; or it could be a general option in the Tagger view configuration.

My use case it that I set up the renamerOnUpdate plugin to only rename files when they are marked as organized. Nevertheless, I could see how other people might find this feature useful, because, if a scene is getting identified from a database or scraper, it probably has good enough data to be considered organized by most people.

I couldn't find a similar request, so I decided to open this issue. I hope this is not a duplicate.

PR #4031, opened 2023-08-14
Tagger config option to mark scene as organized 

plato178:
This pull request is a first attempt at adding a config option to the Scene Tagger view to immediately mark a scene or scenes as Organized on save, as described in issue #3731.

I've added this as an additional option in the Configuration pane, like so:

Screenshot 2023-08-14 at 21 52 10

This currently defaults to current behaviour, i.e. a scene will not be set to Organized on save - the new behaviour is opt-in.

Also, please let me know if the wording of the label and description of the new config option is acceptable - this can be changed if needed.


grabolento:
Thank you for this PR!

My skills only allow me to comment on the wording. I think there's a editing mistake on the feature description: "[...] when the click the Save button is clicked".

While we're at it, my suggestions:

#### Label
For conciseness, since there will be a description below:
> Mark as Organized on save

#### Description
I could think of two options:
> Immediately mark the scene as Organized after clicking the Save button.
> Immediately mark the scene as Organized after the Save button is clicked.

plato178:
@grabolento Thanks for the feedback - I've updated the wording 👍

DogmaDragon:
Fixed by https://github.com/stashapp/stash/pull/4213

PR #4051, opened 2023-08-22
[Bug Fix] Nil pointer dereference in scraper

Maista6969:
If a scraper returns an empty result we still try to apply tag exclusions, which will fail when dereferencing a nil scene

### DIFF
--- a/internal/api/resolver_query_scraper.go
+++ b/internal/api/resolver_query_scraper.go
@@ -129,7 +129,9 @@ func (r *queryResolver) ScrapeScene(ctx context.Context, scraperID string, scene
 		return nil, err
 	}
 
-	filterSceneTags([]*scraper.ScrapedScene{ret})
+	if ret != nil {
+		filterSceneTags([]*scraper.ScrapedScene{ret})
+	}
 
 	return ret, nil
 }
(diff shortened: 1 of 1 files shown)


WithoutPants:
Can you please provide instructions to reproduce the original issue? I'm not able to reproduce and I'm not sure this is how I would solve it.

Maista6969:
Sorry, I could have been more clear: to reproduce I have created `EmptyScraper.yml` which never returns any results
```yml
name: EmptyScraper
sceneByURL:
  - action: scrapeXPath
    url:
      - example.com
    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    scene:
      Title:
```

URL scraping a scene from example.com should work fine, essentially a no-op, but adding any tag exclusions will pop up a nil-pointer dereference error


This mostly happens with broken or outdated scrapers that return no result, so I've had some confusing error reports for scrapers that are just failing because of new site layouts 😅

WithoutPants:
Ok, I think the better solution is to add `if ret != nil {` around the `filterSceneTags` calls that use a constructed list containing ret - in `ScrapeSceneURL` and `ScrapeScene`. In general, I think the other lists shouldn't contain `nil` values, and it's reasonable to expect that the input to `filterSceneTags` should not contain any `nil` values.

Maista6969:
Done 👍

PR #4075, opened 2023-08-31
Fix URL and aliases update nil dereference

### LINKED ISSUE #4072: [Bug Report] Scene: input: Scene Internal system error.
Hi there,

Like #4030, after upgrading to v0.22.0, attempting a sceneUpdate Mutation from the API results in the following error in the stash logs:

```
23-08-30 17:38:08Error   Scene: input: Scene Internal system error. Error <runtime error: invalid memory address or nil pointer dereference>
2023-08-30 17:38:08Error   runtime error: invalid memory address or nil pointer dereference
```

As a result, the scene doesn't update its data (performers, studio, title, etc). However, the API responds a OK status code to the client.

I upgraded to v0.22.1 but the issue is still here. Note that I didn't change anything to my client sceneUpdate call code. I only updated Stash from v0.21.0 which was working.

DingDongSoLong4:
Simple fix for #4072.

This will ignore a `null` value in `aliases` or `url`, to avoid unintentional data loss. Both these fields are deprecated anyway, `alias_list` or `urls` should be used instead which gives more control.

Fixes #4072

### DIFF
--- a/internal/api/resolver_mutation_performer.go
+++ b/internal/api/resolver_mutation_performer.go
@@ -200,7 +200,7 @@ func (r *mutationResolver) PerformerUpdate(ctx context.Context, input PerformerU
 			Values: input.AliasList,
 			Mode:   models.RelationshipUpdateModeSet,
 		}
-	} else if translator.hasField("aliases") {
+	} else if input.Aliases != nil {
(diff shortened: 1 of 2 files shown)


DingDongSoLong4:
Made the change mentioned in #4072.

PR #4090, opened 2023-09-06
Filter criterion fixes

DingDongSoLong4:
This started as a fix for an issue reported on Discord, caused by the saved filter refactor in #4054, where the Duration filter no longer uses the `DurationFilter` component but rather the standard `NumberFilter`.

I ended up making a few other tweaks and refactors:
- Merged `DuplicatedCriterion` and `PhashDuplicateCriterion`
- Reworked the `DurationInput` component. The main improvement is that the increment/decrement buttons now work when the input box is empty.
- Tweaked the `DurationFilter` code to be a lot more readable
- And then a completely unrelated tiny change to the scene player, to register the abloop plugin outside of the init function. This prevents the plugin from being registered twice if the player has to reinitialize, which throws a console warning.

Edit:

Fixed a few more issues with incorrect components, for both the Rating and Country filters. I ended up refactoring a lot of the filter model code, everything is a lot more consistent now. I did a side-by-side comparison of all the filter dialogs between this and before the saved filter refactor, where I found and fixed a few other criterion issues.

I've also removed the previous `"none"` criterion - I can't see where it's being used anymore. If this is a problem it can easily be reverted.

And then a few minor typing changes, to replace `any` with `unknown`.


WithoutPants:
The `refactor criteria` commit appears to be a comprehensive reversion of my previous refactoring of the criterion options (see #1474), and I can't really work out why this was done. I can't work out why changing the options from instances into classes is a good or simplifying change.

DingDongSoLong4:
@WithoutPants 

The instance to class change was mostly for consistency (ie `criterionOptions` in `list-filter/scenes.ts`), and since I removed `makeCriterionFn`, separate classes are needed to override `makeCriterion`. There I just thought it was cleaner to have a direct function override rather than passing a function through to the constructor.

The best example of why this is better would be one of the issues I fixed, the Country criterion. Previously if you call `makeCriterion` on `countryCriterionOption`, it gives you an instance of `StringCriterion`, not `CountryCriterion`, because it is an instance of `StringCriterionOption`. This breaks the filter dialog and shows a string input instead of the country input box (ignoring the fact that `countryCriterionOption` was never actually used in the performer criterion option list). Now we could add a constructor parameter like for `BooleanCriterionOption` to set the `makeCriterionFn`, but idk I just find that to be more clunky.

Besides changing a few instances to classes, and removing a few unused `Criterion` subclasses, I'm not sure how it's a "comprehensive" reversion of #1474 - I definitely did not look at that PR and was just changing what I believed would be cleaner.

In the end it doesn't really matter, but that's just my reasoning. I'd be happy to make all the `CriterionOption` subclasses with no-argument constructors into instances.

WithoutPants:
I guess I'm just seeing the refactor as unnecessary for the problem that it's trying to solve. I think it needs to be separated from the rest of this. There certainly needs to be some discussion around the design of the criteria/filter stuff, because it's been a bit of a dog's breakfast from the beginning. 

> The best example of why this is better would be one of the issues I fixed, the Country criterion. Previously if you call makeCriterion on countryCriterionOption, it gives you an instance of StringCriterion, not CountryCriterion, because it is an instance of StringCriterionOption. This breaks the filter dialog and shows a string input instead of the country input box (ignoring the fact that countryCriterionOption was never actually used in the performer criterion option list). Now we could add a constructor parameter like for BooleanCriterionOption to set the makeCriterionFn, but idk I just find that to be more clunky.

This could have been more simply solved by changing the `countryCriterionOption` to construct a `CriterionOption`, providing the correct constructor function, in the same way that many other criterion option instances do. Likewise for the rating criterion, we can add a ratingCriterionOption instance to address the rating filter bug.

The other commits look ok, but [206c0a1](https://github.com/stashapp/stash/pull/4090/commits/206c0a1b41177878980477d29f50a4bd4af695ba) belongs in a separate PR and discussion.

PR #4091, opened 2023-09-07
Github action to publish a CUDA build alongside the alpine one

NodudeWasTaken:
For some time people complained that there was no official cuda build.
So i changed the build a little bit and now there could be.

Comments? Do we want to do this? Did i use too many newlines? Is anyone actually gonna use the NVidia arm64 build?
There are also some people who have a jellyfin-ffmpeg build which could be interesting, but for now i just made it similar to the old one.

### DIFF
--- a/docker/ci/x86_64/Dockerfile
+++ b/docker/ci/x86_64/Dockerfile
@@ -47,10 +47,10 @@ ENV NVIDIA_DRIVER_CAPABILITIES=video,utility
 ENV STASH_CONFIG_FILE=/root/.stash/config.yml
 
 # NVENC Patch
-RUN mkdir -p /usr/local/bin /patched-lib
-RUN wget https://raw.githubusercontent.com/keylase/nvidia-patch/master/patch.sh -O /usr/local/bin/patch.sh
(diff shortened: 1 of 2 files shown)


kermieisinthehouse:
Have you tested that the nvenc restriction is actually lifted? I would think that this would need to be done in the host's Nvidia driver, not from inside the container, but I could be wrong.

NodudeWasTaken:
> Have you tested that the nvenc restriction is actually lifted? I would think that this would need to be done in the host's Nvidia driver, not from inside the container, but I could be wrong.

You know i actually didnt, i dont usually watch more than 3 porno's at the same time 😋
But the author claims it does, and i can confirm that it also works in my docker.
image
`ps aux` confirms its using h264_nvenc, so it works fine.

feederbox826:
bit of a necro, but 
- the current dockerfile encapsulates both non-CUDA/non-hwaccel and the cuda/hwaccel versions, this greatly inflates all image sizes
- I'm hoping to address this once and for all in https://github.com/stashapp/stash/issues/4300 with flavored builds for jellyfin-ffmpeg and pre-compiled driverless ffmpeg builds that can still use hwaccel

as for the nvidia patch, one of the explicit goals is to patch on hosts without patched drivers (https://github.com/keylase/nvidia-patch/tree/master#docker-support)

PR #4092, opened 2023-09-07
Model refactor, part 2

DingDongSoLong4:
This is some more refactoring and minor improvements, following on from my earlier model refactoring. I'm trying to be a bit better with sending through smaller PRs, so there's more to come - expect a part 3 :)

The commit messages are for the most part fairly self-expanatory:
- 4bec8516b29748b8dc384a9f070c4dd10d27b209 adds a bunch of conversion methods to the `changesetTranslator`, which allows for a lot of duplicated `hasField` checking and type conversions to be moved out of the various resolver functions.
- f74252b517637036c783326d80f567bf04888ccd ensures that every error resulting from a conversion is wrapped in a suitable message - an invalid ID will no longer just give a `strconv.Atoi: ...` error, but rather `converting id: strconv.Atoi: ...`.
- 8b33f955b5a1b4792e83b63a8fa9f7d975a1c22f adds `New` and `NewPartial` functions and ensures they are used everywhere, to avoid forgetting to set/update `CreatedAt` and `UpdatedAt`. While making this change, I indeed found, and fixed, a few places where this had been forgotten.
- 27016aaebced911aa389db0683478c7c5477c35e replaces `getStashIDsForStudio` and `getStashIDsForPerformer` with calling the `GetStashIDs` method directly - this is cleaner and reduces overhead vs `LoadStashIDs`.
- e6d865dd1607bafd62378d2437277540e8b1f674 just fixes a few tests broken by 8b33f955b5a1b4792e83b63a8fa9f7d975a1c22f.
- c594f7da5df68c9daa4791b8737a7c92034c0293 removes `ImageCreateInput` in favour of just passing in the `fileIDs` directly to `Create`. This is consistent with how `Create` works for both scenes and galleries.


internal/api/changeset_translator.go
```diff
@@ -235,3 +237,151 @@ func (t changesetTranslator) optionalFloat64(value *float64, field string) model
 
 	return models.NewOptionalFloat64Ptr(value)
 }
+
+func (t changesetTranslator) fileIDPtrFromString(value *string, field string) (*models.FileID, error) {
```
WithoutPants, line 241: While there is an argument to be made for consistency, I don't think it's worth keeping `field` in functions that don't use them.

internal/api/resolver_mutation_performer.go
```diff
@@ -249,10 +224,9 @@ func (r *mutationResolver) PerformerUpdate(ctx context.Context, input PerformerU
...
 		}
 
-		if err := performer.ValidateDeathDate(existing, input.Birthdate, input.DeathDate); err != nil {
-			if err != nil {
-				return err
-			}
+		err = performer.ValidateDeathDate(existing, input.Birthdate, input.DeathDate)
+		if err != nil {
+			return err
 		}
```
WithoutPants, line 230: The original code wasn't right, but I think we should keep the `if err := ...; err != nil {` structure.

internal/autotag/gallery_test.go
```diff
@@ -46,12 +46,24 @@ func TestGalleryPerformers(t *testing.T) {
...
+				}
+
+				// updated at should be set and not null
+				if !got.UpdatedAt.Set || got.UpdatedAt.Null {
+					return false
+				}
+				// else ignore the exact value
+				got.UpdatedAt = models.OptionalTime{}
+
+				return assert.Equal(got, expected)
```
WithoutPants, line 64: I think this can be reused in a utility function - there's a bit of repeated code in these.

pkg/gallery/scan.go
```diff
@@ -76,34 +75,31 @@ func (h *ScanHandler) Handle(ctx context.Context, f models.File, oldFile models.
...
 		for _, i := range images {
-			if _, err := h.ImageFinderUpdater.UpdatePartial(ctx, i.ID, models.ImagePartial{
+			imagePartial := models.ImagePartial{
 				GalleryIDs: &models.UpdateIDs{
 					IDs:  []int{newGallery.ID},
 					Mode: models.RelationshipUpdateModeAdd,
 				},
-				UpdatedAt: models.NewOptionalTime(now),
-			}); err != nil {
+				UpdatedAt: models.NewOptionalTime(newGallery.UpdatedAt),
```
WithoutPants, line 95: Add a comment for why this is not using the usual `NewPartial` pattern.

pkg/image/scan.go
```diff
@@ -215,17 +208,18 @@ func (h *ScanHandler) associateExisting(ctx context.Context, existing []*models.
...
+			if _, err := h.CreatorUpdater.UpdatePartial(ctx, i.ID, imagePartial); err != nil {
 				return fmt.Errorf("updating image: %w", err)
 			}
 
 			if g != nil {
-				if _, err := h.GalleryFinder.UpdatePartial(ctx, g.ID, models.GalleryPartial{
-					UpdatedAt: models.NewOptionalTime(time.Now()),
-				}); err != nil {
+				galleryPartial := models.GalleryPartial{
+					UpdatedAt: imagePartial.UpdatedAt,
```
WithoutPants, line 220: As above, add comment for deviation from pattern.

WithoutPants (changes requested):
Just some minor stuff.

internal/api/changeset_translator.go
DingDongSoLong4: Yeah the reason I did that was for consistency, and because other functions there did just that - `string`, `datePtr`, `intPtrFromString`, etc.

WithoutPants:
I've rewritten the assert functions into internal matching functions for the moment. The existing functions were sometimes returning false, and sometimes asserting and returning false and an assert was incorrect for a matching function. We can move them into a public location if the need arises.

PR #4100, opened 2023-09-10
Allow closing modals with the ESC key

its-josh4:
It's a common pattern for users to press the ESC key to close modals. It is currently disabled, and that causes confusion (pressing ESC makes my browser go out of full screen instead) - and friction (for example after submitting to stash box)

### DIFF
--- a/ui/v2.5/src/components/Shared/Modal.tsx
+++ b/ui/v2.5/src/components/Shared/Modal.tsx
@@ -44,7 +44,7 @@ export const ModalComponent: React.FC<IModal> = ({
 }) => (
   <Modal
     className="ModalComponent"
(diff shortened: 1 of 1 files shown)


its-josh4:
This doesn't seem to be working, at least not for the modal "submit to stash box". Moving to draft, perhaps someone could help advise.

traveller-san:
The default `onHide` is a noop function, and modal visibility is based on the `show` prop, so even though it's capturing the ESC key it isn't doing anything in response. My guess is you can't do this centrally the way the UI is written currently. You would have to set an `onHide` at each call site to `<ModalComponent` in addition to setting `keyboard={true}`

PR #4102, opened 2023-09-10
Add pystashlib to Dockerfile

AnonTester:
Add pystashlib requirement for userscripts


DogmaDragon:
The module is also only used by a limited numbers plugins as far as I know. Most of which are broken or unmaintained as of right now.

PR #4107, opened 2023-09-11
Update a number of dependencies (incl. CVE fixes)

### LINKED ISSUE #3733: [Bug Report] Possibly outdated chromedp causes unmarshaling error when scraping with CDP (unknown PrivateNetworkRequestPolicy value)
**Describe the bug**
Can't use CDP with some scrapers on macOS

**To Reproduce**
1. Run Stash on macOS
2. Add a scene by URL with a scraper that uses CDP (such as Brazzers or Kink)
3. A red banner with an error appears briefly complaining about not being able to process OuterHTML
4. In the log, several instances of `ERROR: could not unmarshal event: unknown PrivateNetworkRequestPolicy value` are logged

**Expected behavior**
The scenes can be parsed normally and no extraneous messages are added to the log

**Stash Version: (from Settings -> About):** 0.20.2 (build hash e685f80e)

 - OS: macOS 12.5.1
 - Browser: Firefox Developer Edition
 - Version: 114.0b3

No smartphone involved

**Additional context**
Tried with CDP based on Brave, Chromium and Chrome (all newest versions)
After looking up the error message, it appears that is related to old chromedp not being able to handle a new value for the PrivateNetworkRequestPolicy, but these changes appear to be at least 8 months old. And not seeing this issue being reported here before makes me suspect it is the Mac build of Stash which is made with an outdated version of chromedp. It could be that it is cross compiled and this part is not properly tested by the people preparing the builds.

its-josh4:
Includes some dependencies that were upgraded in #4106 as well as a few more dependencies.

Some deps that have been upgraded had CVEs.

Notably, upgrades deprecated dependencies such as:
- `github.com/go-chi/chi` (replaced with `/v5`)
- `github.com/gofrs/uuid` (replaced with `/v5`)
- `github.com/hashicorp/golang-lru` (replaced with `/v2` which uses generics)

Fixes #3733


DingDongSoLong4:
I'm running into a few issues with the `yaml` upgrade.

If you clone the [CommunityScrapers repo](https://github.com/stashapp/CommunityScrapers) and set your scraper path to that, you'll get a "did not find expected key" error for both the `Babepedia.yml` and `Hotmovies.yml` scrapers. I've managed to figure out what was causing the message: the parser seems to be choking on a string input of "\n\n{other chars}" (not "\n\n", it deals with that fine). The problematic lines are [here](https://github.com/stashapp/CommunityScrapers/blob/0b5e6becbe0606fbfb65cefe9e2fd3c665dabb50/scrapers/Babepedia.yml#L144) in `Babepedia.yml` and [here](https://github.com/stashapp/CommunityScrapers/blob/0b5e6becbe0606fbfb65cefe9e2fd3c665dabb50/scrapers/Hotmovies.yml#L122-L124) in `Hotmovies.yml`.

There are a couple issues on the https://github.com/go-yaml/yaml repo relating to newline issues, and many of them are quite old. I have a feeling we're going to have to fix the bug ourselves, or find a non-broken version to use.

its-josh4:
I can downgrade the YAML library

DingDongSoLong4:
@its-josh4 I think that's the best option for now yes. v2.4.0 seems to be the latest version that doesn't have this issue, and it is the version we were on previously as well.

DingDongSoLong4:
@its-josh4 I believe the last line in `go.mod` (`replace git.apache.org/thrift.git ...`) can be removed - it doesn't appear to be doing anything anymore.

DingDongSoLong4:
@its-josh4 If you wouldn't mind also upgrading `vearutop/statigz` to `1.4.0` - I've had this upgrade on my local branch adding brotli compression for ages with no issues. It shouldn't require any code changes.

its-josh4:
> @its-josh4 If you wouldn't mind, can you also upgrade `vearutop/statigz` to `1.4.0`? I've had this upgrade on my local branch adding brotli compression for ages with no issues. It shouldn't require any code changes.

Done. Maybe this PR could be merged and other deps can be updated separately?

PR #4110, opened 2023-09-11
Remove vendored dependencies

its-josh4:
Per conversation on Discord, remove the vendor folder and add it to .gitignore

Tested that building Docker works too, locally, with `make docker-build` (on a linux/amd64 box)


.github/workflows/golangci-lint.yml
```diff
@@ -42,19 +42,26 @@ jobs:
...
+          args: --timeout=5m
 
           # Optional: show only new issues if it's a pull request. The default value is `false`.
           # only-new-issues: true
 
-          # Optional: if set to true then the action will use pre-installed Go.
-          # skip-go-installation: true
+          # Optional: if set to true, then all caching functionality will be completely disabled,
+          #           takes precedence over all other caching options.
+          # skip-cache: true
```
its-josh4, line 55: Using the cache here because we don't vendor deps anymore, so this should make checks on PRs a bit faster

Issue #2948, opened 2022-09-25
[Feature] Show file path on list view


ghost:
As I've increased the number of files I have, stash has been getting slower and slower. Right now it takes 34 seconds on average for it to open a scene. I would like to be able to copy the path from the list view and open the file in a file manager.

ALonelyJuicebox:
Oof, 34 seconds is pretty significant...

What kind of media file are you trying to open in stash? Is Stash transcoding the file?
What is the size of your Stash SQL db file? 
How many scenes do you have in your db?
What are the specs of the machine running Stash?

JediMaster25:
> What kind of media file are you trying to open in stash? Is Stash transcoding the file?

I use mp4 and mkv, it doesn't change the time it takes to open. But it plays mkv slower so I never use stash for those.

> What is the size of your Stash SQL db file?

5.7 GB

> How many scenes do you have in your db?

100k

> What are the specs of the machine running Stash?

AMD 5600G

MrX292:
> I use mp4 and mkv, it doesn't change the time it takes to open. But it plays mkv slower so I never use stash for those.
It is probably Transcoding those. 
When the video plays in stash click on the cog in the video player if there is no direct stream option selected and it is not in the list then it is transcoding.
You can try the Userscript `Stash Open Media Player` from https://github.com/7dJx1qP/stash-userscripts

bucketcat:
I have quite a few issues that would be solved by a hotlink with the file path(such as the gallery/image viewer not being functional at all, 90% of the time, the scene viewer just breaks randomly making it impossible browse galleries. No clue what the source is so wont open my own issue, probably firefox ETP related). So please consider this topic instead of trying to solve OP's specific problem exclusively. This would be a useful band-aid for a lot of different problems that do/can arise as a fall back, without having to fully revert to using your local desktop environment and file manager to navigate the smut.

PR #4112, opened 2023-09-12
Add a setting to show file path on scene list table view (#2948)

trj6:
This code adds a new setting in the Interface section to allow users to toggle on/off showing the File Path on the Scene list table view. The default remains False, meaning the file path does not show. Toggling on this setting will show the full file path under the scene title. This attempts to fix issue #2948.


WithoutPants:
With #4368 implemented, this can be added to the list of supported fields in the field selector instead of having a separate setting.

PR #4143, opened 2023-09-22
Plugin settings

WithoutPants:
Adds the ability to get and set plugin-specific settings.

Setting plugin properties is done with mutation `configurePlugin` which accepts a plugin ID and a map of settings. The entire map of settings is replaced for the provided plugin ID.

Getting plugin settings is done with the existing `configuration` query. A new `plugins` field on the returned type presents the settings grouped by plugin ID. 

For example:
```
query {
  configuration {
    plugins
  }
}
```
Returns:
```
{
  "data": {
    "configuration": {
      "plugins": {
        "foo": {
          "bar": true
        },
        "testReact": {
          "bar": 2,
          "foo": false,
          "xyz": "ffsfff"
        }
      }
    }
  }
}
```

The resolver accepts an `includes` field which specifies which plugins to return results for.
```
query {
  configuration {
    plugins(includes: ["testReact"])
  }
}
```
Returns:
```
{
  "data": {
    "configuration": {
      "plugins": {
        "testReact": {
          "bar": 2,
          "foo": false,
          "xyz": "ffsfff"
        }
      }
    }
  }
}
```

The plugin config yaml file now accepts a list of settings which will be displayed on the plugin settings page.

```
settings:
  foo:
    displayName: Foo
    type: BOOLEAN
  bar:
    description: baz
    type: NUMBER
  xyz:
    type: STRING
```

Where `displayName` is not set, then the internal key name will be used instead. If type is not specified, it defaults to `STRING`. The above configuration is shown in the UI as follows:

image


scruffynerf:
Default values?

stg-annon:
> Default values?

should be able to do this from the plugin via the mutation

PR #4148, opened 2023-09-23
PerformerSelect Tagger bugfixes

DingDongSoLong4:
This fixes two bugs which I believe were introduced by the PerformerScene refactoring in #4013.

- When using the Tagger with a non-stashbox scraper, and a scraped performer matches one already in the database, the selection box is no longer highlighted in blue. It also no longer has fixed width, so the control changes size based on how long the performer's name is. This is due to the `performer-select` and `performer-select-active` classes not being properly passed down to the underlying `react-select` component.
- When "Skip" is clicked, the performer select box is not properly cleared anymore - the name still remains. This was just due to a missing `setSelectedPerformer(undefined)`.

Before:
before
After:
after

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerSelect.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerSelect.tsx
@@ -5,6 +5,7 @@ import {
   MultiValueGenericProps,
   SingleValueProps,
 } from "react-select";
+import cx from "classnames";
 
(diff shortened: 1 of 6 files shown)


WithoutPants:
The bug still persists when clicking on skip a after re-setting the performer for a second time. It's not enough to hold back on this PR, but should probably be addressed in future.

Issue #4134, opened 2023-09-17
[Feature] Display selected item count in modal header when editing multiple files


echo6ix:
**Is your feature request related to a problem? Please describe.**
Uses can select one or multiple cards to edit in grid view using the modal edit window. The modal header pluralizes the noun in the header to signify multiple items are being edited. This subtle visual indication (`Edit Scene` vs `Edit Scenes`) could be missed by users unaware that they are editing multiple files.

**Describe the solution you'd like**
Add the number of items being edited in the modal, i.e. `Edit scenes (10 scenes)`

![image](https://github.com/stashapp/stash/assets/37937507/b5627bbb-b82e-4913-a709-261bf6496dfe)

elkorol:
This is in my pull request [https://github.com/stashapp/stash/pull/4150](url)

Gykes:
Solved with https://github.com/stashapp/stash/pull/6647

PR #4150, opened 2023-09-24
Edit Modal extra fields if only one scene

elkorol:
Added conditional rendering on edit modal, so that if selected scene length is 1, then render all additional fields to be edited. 
Allowing for quick editing of a singular scene from the scenes page without having to click into the scene page.


elkorol:
Resolves this also [https://github.com/stashapp/stash/issues/4134](url)

Gykes:
Closed as stale

feederbox826:
Superceded by https://github.com/stashapp/stash/pull/6647

Issue #2284, opened 2022-02-04
Stash.app cannot find config file on MacOS


shanty-chits:
**Describe the bug**
On Mac, opening the latest development Stash.app opens the new setup screen no matter what. Console logs show current working directory as `<private>` even though the app is located with the library on a removable volume. Granting Stash.app full disk permissions does not fix the issue. If the setup wizard is completed, it ends with `error creating config file: open config.yml: read-only file system`. If `-c` or `--config` flags are passed like `open Stash.app --args -c [path/to/config]`, the app does not open at all. If a symlink to the config file is placed in $HOME/.config/stash, the new setup screen still opens.

**To Reproduce**
Open Stash

**Stash Version: (from Settings -> About):**
v0.12.0-40-gc857e301

**Desktop (please complete the following information):**
 - OS: MacOS 12.2 Monterey

kermieisinthehouse:
Hi, thanks for the report. We'll get this fixed before this hits stable.

Out of curiosity, when you filled out the setup to create a new stash folder, which option did you choose? If you put it in your home folder, does it work?

Is this a fresh Monterrey install, or did you upgrade an existing system to it?

shanty-chits:
> Out of curiosity, when you filled out the setup to create a new stash folder, which option did you choose? If you put it in your home folder, does it work?

Selecting "current working directory" fails with the indicated error. Selecting the $HOME option hangs and the program has to be force-quit. Moving the app to the home folder doesn't seem to have any effect.
 
> Is this a fresh Monterrey install, or did you upgrade an existing system to it?

Upgrade from an existing system.

kermieisinthehouse:
Alright, I think I've got a handle on what's causing this. I need special handling on MacOS for "current directory".

kermieisinthehouse:
Using `open ./Downloads/Stash.app --args -c ~/my_custom_path/stash/config.yml` should be a workaround at the moment.

MC1R-13:
This must be Monterey-specific. The latest Big Sur (11.6.4) has no such issue.

newadventure079:
I'm on macOS 12.2.1 and am seeing this too.  

`open /Applications/stash/Stash.app/ --args -c /Applications/stash/config.yml` doesn't work.  The command completes and nothing happens

newadventure079:
[...]

PR #4153, opened 2023-09-25
Improved/fixed macOS support

DingDongSoLong4:
This improves macOS support:
- Desktop notifications appear to be broken, and fixing them required a patch to `gosx-notifier` (https://github.com/kermieisinthehouse/gosx-notifier/pull/1). This was because the `-sender` flag was not being used.
- Distribute a unified (both Intel + ARM) `stash-macos` binary, rather than separate binaries.
- Distribute a packaged `Stash.app`, allowing macOS users to completely avoid the Terminal.
- Also changed the Bundle ID to match `stashapp.cc`.


DingDongSoLong4:
I've just come across #2284, but for me on macOS Ventura 13.5.2 Stash is able to create and find the config file in `$HOME/.stash` perfectly fine - not sure if something changed on the macOS side or the Stash side.

kermieisinthehouse:
Fantastic that someone who can figure out MacOS is looking at this.

Did you figure out how to reduce the executable size with the unified app? When I last tried this, the frontend assets were included twice, so the Mac binary was nearly twice the size of the other platforms.

DingDongSoLong4:
> Did you figure out how to reduce the executable size with the unified app? When I last tried this, the frontend assets were included twice, so the Mac binary was nearly twice the size of the other platforms.

Ah I hadn't thought of size, but I have no idea how, or if it is even possible, to deduplicate data in a unified binary. It makes sense that it would be twice the size (and it indeed is), since it basically contains two separate binaries.

The only real thing that could be deduplicated is the UI assets, as you mention, which are currently 7.8MB. The entire macOS unified binary is 85MB, so that's a potential ~10% decrease, which isn't _that_ much imo.

The alternative is to do what we're doing now and keep the two macOS binaries separate, but I think the convenience of the unified binary is worth the size tradeoff to most users.

And PS I don't actually know that much about macOS apps/development, I just have a MacBook so I'm able to test things - all of the fixes here came from trial, error and Google :)

kermieisinthehouse:
85 MB is definitely not catastrophic and I wouldn't let that hinder this.

DingDongSoLong4:
Alright, I believe this is now ready.

When running the `.app`, the working dir seems to always be set to `/`. If this is detected during setup, choosing a working directory is now disabled:
image
I'm not 100% happy with the message, but it's the best I've been able to come up with while keeping it as short as possible to avoid a truly enormous button.

I've also made a few other improvements to the Setup process, now that the OS and working/home directories are exposed to the UI. The most useful of these is probably that the two initial buttons now show the actual paths they refer to, rather than "In the current working directory" for example. This should hopefully make it more clear where Stash is actually being set up, and should also help if the current directory is set to something unexpected.

DingDongSoLong4:
I've added some tweaks/improvements to the Makefile: 
- The use of the `flags-*` targets is now more explicitly "advertised", which means that the separate `stash-release`, `stash-release-static`, etc. targets are no longer necessary. Only the `build-release` target is kept for convenience. This means we won't have to add a bunch of extra targets when a `stashpkg` binary is added.
- The `cross-compile-*` targets have been renamed to `build-cc-*`. This is in an attempt to emphasize that they are cross-compile versions of the `build` target _only_, and `make ui`, `make generate`, etc. are still required when using them.
- `-applesilicon` is now `-arm`.
- The `ui-nolegacy` and `ui-sourcemaps` have been removed - they never actually worked, since `ifdef`s are evaluated at the initial parsing stage, not when actually executing a recipe.

There's also now a new `stash-macapp` target which lets you build `Stash.app` when running on a Mac, without having to cross-compile.

I've also removed `cross-compile.sh` and `upload-pull-request.sh`. `upload-pull-request.sh` hasn't been used in a long time, and `cross-compile.sh` is very outdated and not really necessary.

And then I've updated various documentation files, to reflect these changes and remove outdated information.

PR #4162, opened 2023-09-30
Filtered counts on studio performer subpage

dogwithakeyboard:
Makes the Performers' scene, image, movie and gallery counts in the Studio > Performers page return the number of scenes etc that the performer has with that studio rather than the total for the performer. On the subpage the counts filters and sorts will use the filtered number.

- Adds a `findPerformers` resolver to the `Studio` type which accepts a `PerformerFilter` and `FindFilter`, and `depth` to handle the studio hierarchy.

- Adds a `ParentObject` to `Performer` so the studio id and depth can be chained down to the count resolvers.

- Adds a new argument to the `makeQuery` function to override the normal sorting behaviour when queried though the studio resolver.

I got into some typescript trouble when `FindStudioPerformersQueryResult` couldn't extend `QueryResult` when the graphql schema had id as required (`id: ID!`). I've muddled my way through it but there's probably a better way.

I intend to make a similar change for the Performer > Appears With page but I wanted to keep this as as small as possible.


ui/v2.5/src/core/StashService.ts
```diff
@@ -305,6 +305,38 @@ export const useFindStudios = (filter?: ListFilterModel) =>
     },
   });
 
+export const useFindStudioPerformers = (
+  filter?: ListFilterModel,
+  id?: string,
+  depth?: number
+) =>
+  GQL.useFindStudioPerformersQuery({
+    fetchPolicy: "no-cache",
```
DingDongSoLong4, line 314: What is the reason for using `"no-cache"` here? `"no-cache"` means that data will not be invalidated and refetched after mutations, which was what caused the issue I fixed in #3973.

dogwithakeyboard: When loading a studio's performers list and then clicking a performer the performer page tab badges would use the cached _counts from the previous page and not the total counts like they should. A shotgun approach? Reading the docs again and it looks like `cache-and-network` would probably work too. Or change the performer page query to `network-only`?

DingDongSoLong4:
Apologies for what could be yet another iteration, but I do have an alternative which I hope won't be too difficult to implement.

I don't think it's a good idea to have the `scene_count`, `image_count`, `gallery_count` and `movie_count` fields change based on what the parent query is. An object's fields should be uniquely determined by its ID + typename: no matter where in the schema you receive a performer for example, if the IDs are the same, the fields should be the same. Avoiding this was one of the goals of the design I suggested on the original PR (https://github.com/stashapp/stash/pull/3813#issuecomment-1587978695).

I propose to add an argument to the `scene_count`, `gallery_count`, etc fields, which specify the related object to filter on - similar to #3816. When I made the original comment, I was not aware that doing this was possible. The query for performers on the studio page could then look something like this (exact names are of course subject to change):
```gql
findStudio(id: x) {
  performers(<filters>) {
    id
    ...
    scene_count(related_object: { type: "Studio", id: x, depth: -1 })
  }
}
```
(superficial, but I've changed `findPerformers` to `performers`, I just think it looks better :) )

This is a bit more clunky, but it means that the value of the `scene_count` field is no longer dependent on its parent query. We don't have to now, but this also then means it would be possible to include a entire scene filter as an argument, allowing for more complicated filtered counts.

ui/v2.5/src/core/StashService.ts
DingDongSoLong4: Okay I think I see why this was done, presumably to avoid the `*_count` fields from being cached which would then put incorrect data in the cache. This is one of the reasons why fields should be a function of only the ID + typename - it breaks normalized caching.

dogwithakeyboard:
I never imagined getting four numbers out of a database could be so much work : )

At one point I had an intermediate `studioPerformers` with its own _counts a bit like this:

```
findStudio(id: x) {
    studioPerformers(<filters>, depth) {
        scene_count
        performer {
            id
            ...
        }
    }
}
```
It's more work on the frontend but it would have its own type.

ui/v2.5/src/core/StashService.ts
DingDongSoLong4: > When loading a studio's performers list and then clicking a performer the performer page tab badges would use the cached _counts from the previous page and not the total counts like they should. A shotgun approach? Reading the docs again and it looks like `cache-and-network` would probably work too. Or change the performer page query to `network-only`?

Unfortunately with the current implementation I believe `no-cache` really is the only option - the query is essentially poisoning the cache with incorrect `_count`s so it must not be cached at all. We could use `cache-and-network` or `network-only` for all _other_ queries that involve `_count`s, but that isn't really feasible - at that point why have a cache at all.

DingDongSoLong4:
> At one point I had an intermediate `studioPerformers` with its own _counts a bit like this:
> 
> ```
> findStudio(id: x) {
>     studioPerformers(<filters>, depth) {
>         scene_count
>         performer {
>             id
>             ...
>         }
>     }
> }
> ```
> 
> It's more work on the frontend but it would have its own type.

If I'm interpreting that correctly, it looks like a similar idea as my original design with the parallel `filtered_counts` arrays, although you're using an intermediate type, which is definitely more elegant than my parallel arrays. In general the latest iteration is definitely cleaner and more elegant than what I initially came up with though.

I think it's also worth mentioning that in my original design, I hadn't intended for the inclusion of the `filtered_counts` field to trigger filtering/sorting to take the "related objects" into account. I meant there would be _additional_ `filtered_*_count` options to trigger the behaviour, independent of the `filtered_count` field. I should have picked up on this, so my bad again.

WithoutPants:
> I never imagined getting four numbers out of a database could be so much work : )
> 
> At one point I had an intermediate `studioPerformers` with its own _counts a bit like this:
> 
> ```
> findStudio(id: x) {
>     studioPerformers(<filters>, depth) {
>         scene_count
>         performer {
>             id
>             ...
>         }
>     }
> }
> ```
> 
> It's more work on the frontend but it would have its own type.

I also prefer this solution best - though I'd change `studioPerformers` to just be `performers` as noted. The `depth` field also needs to be clearly documented.

DingDongSoLong4:
Looks good to me, and seems to work correctly.

The only issue is that StudioPerformer has two "ID" fields that uniquely determine its fields: `id` and `studio_id`. The client cache by default identifies objects in the cache based on only the field called `id`, so it should be configured to [use both fields](https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids).

Having said that, since the `StudioPerformer` object can only come from one place (the `performers` resolver on a `Studio`), I think it makes sense to [disable normalization entirely](https://www.apollographql.com/docs/react/caching/cache-configuration/#disabling-normalization) for `StudioPerformer`, since that is additional overhead that provides no real benefit here.

I also looked at the `FindStudioPerformersQueryResult` issue, and I think I kind of understand why it's erroring, but a "proper" fix seems like it would be rather complicated and non-trivial. I think your solution is fine for now - you might just want to add a comment to explain why it's needed.

dogwithakeyboard:
> The only issue is that StudioPerformer has two "ID" fields that uniquely determine its fields: `id` and `studio_id`. The client cache by default identifies objects in the cache based on only the field called `id`, so it should be configured to [use both fields](https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids).
> 
> Having said that, since the `StudioPerformer` object can only come from one place (the `performers` resolver on a `Studio`), I think it makes sense to [disable normalization entirely](https://www.apollographql.com/docs/react/caching/cache-configuration/#disabling-normalization) for `StudioPerformer`, since that is additional overhead that provides no real benefit here.

If that's how the cache works I can rename `id` to `performer_id` it would have the same effect? As you say, `id` doesn't uniquely identify `StudioPerformer`. Then `keyFields: ["performer_id", "studio_id"]` can be added down the line if necessary.

> I also looked at the `FindStudioPerformersQueryResult` issue, and I think I kind of understand why it's erroring, but a "proper" fix seems like it would be rather complicated and non-trivial. I think your solution is fine for now - you might just want to add a comment to explain why it's needed.

As a beginner I found it utterly baffling!

DingDongSoLong4:
> If that's how the cache works I can rename `id` to `performer_id` it would have the same effect? As you say, `id` doesn't uniquely identify `StudioPerformer`. Then `keyFields: ["performer_id", "studio_id"]` can be added down the line if necessary.
> 
This should indeed have the same effect yes.

> > I also looked at the `FindStudioPerformersQueryResult` issue, and I think I kind of understand why it's erroring, but a "proper" fix seems like it would be rather complicated and non-trivial. I think your solution is fine for now - you might just want to add a comment to explain why it's needed.
> 
> As a beginner I found it utterly baffling!

Yeah the error message is not very helpful, and I'm beginning to doubt myself on whether I fully understand why it occurs. I haven't been able to find a non-hacky solution that fixes it.

PR #4176, opened 2023-10-06
Fix type issues

InfiniteStash:
Removes a bunch of lint exceptions for `no-explicit-any`.

Generally speaking only types are changed, so there will be no practical impact. The exception is the `usePageVisibility` which I've trimmed down. All modern browsers support the `visibilitychange` so all the boilerplate to read legacy browser states can be trimmed. https://caniuse.com/mdn-api_document_visibilitychange_event

### DIFF
--- a/ui/v2.5/src/hooks/PageVisibility.ts
+++ b/ui/v2.5/src/hooks/PageVisibility.ts
@@ -1,50 +1,14 @@
-import { useEffect, useRef } from "react";
+import { useEffect } from "react";
 
 const usePageVisibility = (
   visibilityChangeCallback: (hidden: boolean) => void
 ): void => {
-  const savedVisibilityChangedCallback = useRef<(hidden: boolean) => void>();
-
   useEffect(() => {
-    // resolve event names for different browsers
-    let hidden = "";
-    let visibilityChange = "";
-
-    if (typeof document.hidden !== "undefined") {
-      hidden = "hidden";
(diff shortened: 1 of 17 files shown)


ui/v2.5/src/utils/index.ts
DingDongSoLong4, line 1: I got rid of these re-exporting `index.ts` files a while back because they were causing strange circular dependency warnings - see #3442. It's possible that this could no longer be happening, but I think it's safer to just not have `index.ts` files at all.

ui/v2.5/src/hooks/PageVisibility.ts
DingDongSoLong4, line 1: This could _potentially_ cause a problem, in that smart TVs often use crazy old browser versions (Chrome, usually) - see #3490. I'm just mentioning it for awareness - I think we can just make the change and if any issues pop up then revert or figure out some other workaround.

ui/v2.5/src/utils/index.ts
DingDongSoLong4: I tried my very best to figure out where the circular dependencies _actually were_ last time, to no avail. Unless I missed something, there were no actual cyclic imports anywhere - the warnings were all false positives.

At some point all the UI dependencies do need to be upgraded again, so perhaps that fixes the underlying issue.

I do agree the extra + longer import statements are annoying, I guess I just want the consistency of not using index files at all. But it's not giving any warnings now, so it's definitely not an issue.

PR #4184, opened 2023-10-09
Postprocess dimension to metric

Maista6969:
This issue crops up from time to time when writing scrapers for sites that do not have properly normalized data for performer measurements, most recently [in this scraper for PornHex](https://github.com/stashapp/CommunityScrapers/pull/1341#issuecomment-1616287820): sometimes performer heights are listed in feet and inches, but sometimes they're in metric already.

The examples used in that case were [Adriana Chechik](https://pornhex.com/pornstars/adriana-chechik) and [Abella Danger](https://pornhex.com/pornstars/abella-danger): when run through `feetToCm` Adriana would get the correct height (5'3" ≈ 160cm) but Abella would grow to be 191cm tall because 1.63m is parsed as 1 foot and 63 inches.

Similar cases have come up with the addition of the penis length field for performers, so I made the choice to interpret single-digit values as being inches.

I'm not fully satisfied with the name `dimensionToMetric` but naming is hard.


Maista6969:
Dropping this in favor of #4200

PR #4185, opened 2023-10-09
Log more when resolving Python

Maista6969:
It's very common for users to struggle with configuring their Python environment for use with Stash and more verbose logging will help us when diagnosing the common issue of not knowing which (if any) Python executable they are using.

I also think the `Resolve` method could use the result of `LookPath` directly to use the absolute path for the Python executables it might find: this should make no difference to its usage, but allows us to log the absolute path.

### DIFF
--- a/pkg/scraper/script.go
+++ b/pkg/scraper/script.go
@@ -84,7 +84,7 @@ func (s *scriptScraper) runScraperScript(ctx context.Context, inString string, o
 
 	go handleScraperStderr(s.config.Name, stderr)
 
-	logger.Infof("Scraper script <%s> started", strings.Join(cmd.Args, " "))
+	logger.Debugf("Scraper script <%s> started", strings.Join(cmd.Args, " "))
 
 	// TODO - add a timeout here
 	// Make a copy of stdout here. This allows us to decode it twice.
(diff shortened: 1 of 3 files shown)


pkg/scraper/script.go
```diff
@@ -88,7 +84,7 @@ func (s *scriptScraper) runScraperScript(ctx context.Context, inString string, o
 
 	go handleScraperStderr(s.config.Name, stderr)
 
-	logger.Debugf("Scraper script <%s> started", strings.Join(cmd.Args, " "))
+	logger.Infof("Scraper script <%s> started", strings.Join(cmd.Args, " "))
```
WithoutPants, line 87: I don't think this is necessary.

pkg/scraper/script.go
```diff
@@ -116,7 +112,7 @@ func (s *scriptScraper) runScraperScript(ctx context.Context, inString string, o
 	}
 
 	err = cmd.Wait()
-	logger.Debugf("Scraper script finished")
+	logger.Infof("Scraper script finished")
```
WithoutPants, line 115: As above.

WithoutPants (changes requested):

pkg/scraper/script.go
Maista6969: Changing the level or logging that the scraper script is running? This was mostly for the support cases were we get "ran the scraper, nothing in the log" and I think it'd be helpful to at least see that Stash ran the scraper successfully even if it didn't produce correct output. We always ask users to turn their logging level to Debug and then there's usually some output 🙂

WithoutPants (approved):

PR #4216, opened 2023-10-17
Show stash-ids with their endpoint

WithoutPants:
Changes the display of stash-ids to show their endpoint. Currently shows as pill, but open to changing this based on feedback.

The component includes the endpoint name as the value for `data-endpoint`, so the colours can be customised per endpoint using custom css.

Examples:
image
image
image
image


IAmKontrast:
I prefer a cleaner and less conspicuous look.

image
(second trash bin is mouse-over)

**Add following to Custom CSS:**

Stash IDs
`.stash-id-pill * {
  background-color: transparent!important;
  font-weight: 400;
}`
For endpoint separating use `.stash-id-pill[data-endpoint="<endpoint name e. g. StashDB>"] *` instead

Delete buttons
`button[title="Delete Stash ID"] {
  padding: .375rem .375rem;
  background-color: transparent;
  border-color: transparent;
}`

PR #4224, opened 2023-10-17
Scene card primary file video and audio information icons

elkorol:
Added the optional ability, configurable in settings, to display an informational icon along with amount of files attached to scene, and in popover show both audio and video codecs used per scene and clickable filter links, to filter scenes list to only show files that has that particular codec


DogmaDragon:
Thank you for adding a toggle! 👍

elkorol:
> Thank you for adding a toggle! 👍

No worries, I've got one last commit to make. But I need to add ""@popperjs/core": "^2.11.6" as a dependency. It seems to be already a dependency of one of the modules. But I'm getting linting errors without adding it. I've tried adding it with "yarn add @popperjs/core@2.11.6", and it stops the linting errors. But upon restarting the server I get a blank page. So I'm unsure what it's breaking. I tried adding it manually to the package.json and then doing yarn install and got the same results. To get the dev instance running again. I had to delete the node_modules folder. And run yarn cache clean and then make pre-ui again to get it working again. Do you have any idea what I'm doing wrong? I'm assuming it's some sort of conflict. As the module is already a dependency of a dependency and making it dependency itself, I think is breaking the code for the module which it's a dependency for?

I need popperjs, to render the little triangle element of the popover to be center aligned with the information icon. And as an added bonus, I could offset the popover a little. So it wasn't exactly centered with the scene card. Which I thought looked kind of good, like windows stacking and cascading over one another. I thought it being dead on center of the parent element didn't quite look right.

elkorol:
Just finished the hopefully last commit. I had to use "/* eslint-disable @typescript-eslint/no-explicit-any */" a couple of times when importing functions as the datatype for some fields they use are of the type any in the GraphQL spec.

elkorol:
Also the button that copies the file path for the associated files can be removed, if isn't liked. But I thought it would be a friendly addition, as many people might be copying file path to paste into their file browser anyway. This way, it's a simple click to copy and no need to highlight the file path minus the filename > right click and copy,

WithoutPants:
I think this is better implemented as a card customisation plugin.

PR #4235, opened 2023-10-22
Fixed: "make test" fails if host isn't using UTC as time zone

its-josh4:
Per title. Unit tests expect TZ to be UTC

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -309,7 +309,7 @@ lint:
 # runs unit tests - excluding integration tests
 .PHONY: test
 test:
-	go test ./...
+	TZ=UTC go test ./...
 
 # runs all tests - including integration tests
 .PHONY: it
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,11 @@ GO_BUILD_TAGS += sqlite_stat4
 # set STASH_SOURCEMAPS environment variable or uncomment to enable UI sourcemaps
 # STASH_SOURCEMAPS := true
 
-export CGO_ENABLED := 1
+# Enable CGo for all builds and tests
+export CGO_ENABLED ?= 1
+
+# Set the time zone to UTC by default for all tests
+export TZ ?= Etc/UTC
 
 .PHONY: release
 release: pre-ui generate ui build-release
@@ -309,7 +313,7 @@ lint:
 # runs unit tests - excluding integration tests
 .PHONY: test
 test:
-	TZ=UTC go test ./...
+	go test ./...
 
 # runs all tests - including integration tests
 .PHONY: it


DingDongSoLong4:
@its-josh4 This breaks on Windows when using cmd or powershell as the shell.

Using make itself to set the environment variable instead of the shell fixes that:
```diff
diff --git a/Makefile b/Makefile
index f802fc99..0b275ebb 100644
--- a/Makefile
+++ b/Makefile
@@ -308,8 +308,9 @@ lint:

 # runs unit tests - excluding integration tests
 .PHONY: test
+test: export TZ := UTC
 test:
-   TZ=UTC go test ./...
+   go test ./...

 # runs all tests - including integration tests
 .PHONY: it

```

The same change should probably also be added to the `it` target just below in the makefile as well.

What are the failing tests though? I've set the `TZ` variable to a whole bunch of different things but I haven't been able to make anything fail yet.

its-josh4:
@DingDongSoLong4 Ok I've done a few more tests and interestingly, it only fails if the TZ is America/Los_Angeles (UTC-7), which is where I'm located. Any other TZ works:

```
TZ=America/Los_Angeles go test ./internal/api -count=1 -v
=== RUN   TestTagCreate
--- PASS: TestTagCreate (0.00s)
=== RUN   TestTimestampSymmetry
--- PASS: TestTimestampSymmetry (0.00s)
=== RUN   TestTimestamp
=== RUN   TestTimestamp/reflexivity
=== RUN   TestTimestamp/rfc3339
=== RUN   TestTimestamp/date
=== RUN   TestTimestamp/datetime
=== RUN   TestTimestamp/datetime-tz
    timestamp_test.go:58: got 2021-04-05T14:45:36-07:00; want 2021-04-05T14:45:36Z
--- FAIL: TestTimestamp (0.00s)
    --- PASS: TestTimestamp/reflexivity (0.00s)
    --- PASS: TestTimestamp/rfc3339 (0.00s)
    --- PASS: TestTimestamp/date (0.00s)
    --- PASS: TestTimestamp/datetime (0.00s)
    --- FAIL: TestTimestamp/datetime-tz (0.00s)
=== RUN   TestTimestampRelative
=== RUN   TestTimestampRelative/past
=== RUN   TestTimestampRelative/future
--- PASS: TestTimestampRelative (0.00s)
    --- PASS: TestTimestampRelative/past (0.00s)
    --- PASS: TestTimestampRelative/future (0.00s)
FAIL
FAIL    github.com/stashapp/stash/internal/api  0.148s
FAIL
```

Thanks for pointing out that the current approach was failing on Windows. I've changed how I do it (which also impacts ITs now).

DingDongSoLong4:
Alright I've done some investigation into why this is actually happening, and it's due to how `time.Parse()` handles time zones and the coincidence that you happen to be located in the time zone specified in one of the test cases (`2021-04-05 14:45:36 PDT`.

From the [documentation](https://pkg.go.dev/time#Parse) for `time.Parse()`:
> When parsing a time with a zone abbreviation like MST, if the zone abbreviation has a defined offset in the current location, then that offset is used. The zone abbreviation "UTC" is recognized as UTC regardless of location. If the zone abbreviation is unknown, Parse records the time as being in a fabricated location with the given zone abbreviation and a zero offset. This choice means that such a time can be parsed and reformatted with the same layout losslessly, but the exact instant used in the representation will differ by the actual zone offset. To avoid such problems, prefer time layouts that use a numeric zone offset, or use ParseInLocation.

So if your system's time zone is not PDT, `time.Parse` does not recognize the PDT abbreviation and then records a **zero offset** for the time zone in the resulting `time.Time`. This means that running `time.Format(time.RFC3339Nano)` returns that time as if it was GMT. This is case that the test is written for.

But if your time zone _is_ PDT, then `time.Parse` _does_ recognize the abbreviation, and records a **-0700 offset** for the time zone. Running `time.Format(time.RFC3339Nano)` now returns the time with the "correct" offset of -0700, but this is not what the test expects.

I think this test failure is doing its job and revealing a bug in `ParseDateStringAsTime`: parsing depends on your system time zone setting. I think we should make a proper fix rather than force-setting `TZ` in the Makefile, so I see two options:
- Either we use `time.ParseInLocation(format, string, nil)` which will always return times with offset 0, thus always treating input times as being UTC+0. This makes the behaviour when your time zone is PDT equivalent to everywhere else.
- Or we just remove support for time formats that include time zones in `ParseDateStringAsTime`. Having a time that includes an explicit time zone string (ie "PDT") being parsed as UTC+0 I'd say is unexpected.

@WithoutPants what's your opinion? I'd lean towards the second one, I think that behaviour makes the most sense, but it does potentially parsing somewhere.

WithoutPants:
Based on the usage of `utils.ParseDateStringAsTime` I think we can safely go with the second option - remove support for timezones in the input strings (just remove the `railsTimeLayout` parse). It is unclear to me what the correct interpretation of a timezoned date string would be, so it's better to take it out I think.

PR #4242, opened 2023-10-26
Scraper and plugin manager

### LINKED ISSUE #623: [Feature] CommunityScraper Management/Sync Within Stash UI
Keeping community scrapers up to date requires you knowing they are broken, and finding new ones that might be added requires knowing what is added. 

A UI section that can be used to download and update community scrapers. In my head I envision this panel as allowing you to pick and choose scrapers to sync or just sync all.

The alternatives are the good old manual way that there is now.

**Additional context**
N/A

WithoutPants:
image

Adds a package manager to the scraper and plugin settings pages.

These allow the configuration of multiple package sources. 
image

A package source URL can be a local path or URL, and the URL must return a yaml index file containing all of the contained packages in the source:
```
- id: <package id>
  name: <package name>
  version: <version>
  date: <date>
  path: <path to package zip>
  sha256: <sha256 of zip>
...
```

I have example sources deployed at the following URLs:
https://withoutpants.github.io/CommunityScrapers/develop/index.yml
https://withoutpants.github.io/CommunityScripts/develop/index.yml

The package manager unzips package zips into their own directory named for the package id in the applicable local directory (plugins or scrapers). It also writes a `manifest` file to track the package version and files:
```
id: DateParser
name: Date Parser
description: Find date in path or filename and add it
version: 0.2-32f6e33
date: 2023-10-26 14:20:23 +1100
source_repository: https://withoutpants.github.io/CommunityScripts/develop/index.yml
files:
- requirements.txt
- date_parser.py
- date_parser.yml
```

The main branches of https://github.com/WithoutPants/CommunityScripts and https://github.com/WithoutPants/CommunityScrapers have been modified to build the package sources from the source files. This should be considered a proof of concept and for testing purposes only.

Resolves #623


stg-annon:
small improvement may be to index the domains/urls in scrapers so that you can search by that when looking up a scraper

stg-annon:
For plugins can the path be separate from the Index URL?

say I wanted to add [performerBodyCalculator](https://github.com/stg-annon/performerBodyCalculator) to a hosted/existing index.yml with the following

```yml
- id: performerBodyCalculator
  name: Performer Body Calculator
  description: Tags performers based on existing metadata, with tags matching the performers body type
  version: 1.0-b455ac6
  date: 2023-10-26 17:26:15 +0000
  path: https://github.com/stg-annon/performerBodyCalculator/releases/download/v1.0/performerBodyCalculator-1.0.zip
  sha256: 57A899ACC459383C4E74A5E7118D675EFA97556A5B0F4E0E327A0CF6EA8FA32A
```
 should this be possible? it would make for easier development so we could manage a community index that could pull from various places after a review of the entry into the index

stg-annon:
encountered an issue, some scrapers have many YMLs to a single py script, Algolia_* is a prime example of this, how do we want to deal with things like this or `py_common`, direct the user to get the dependency from the repo?

WithoutPants:
> encountered an issue, some scrapers have many YMLs to a single py script, Algolia_* is a prime example of this, how do we want to deal with things like this or `py_common`, direct the user to get the dependency from the repo?

For Algolia, I would just put all of the related ymls into a single package. I haven't yet got a solution for dependencies, though the short term solution would be to bundle the dependencies into the package. Obviously has issues with redundancy, but that's a less major issue.

Phasetime:
I encountered a few small issues while testing:

1. Installing a scraper/plugin usually makes it show up instantly in the list of installed items, but if it's the first item you install and the folder on the filesystem does not exist yet, the installed items only show up after a page reload. 
2. For plugins the installed section isn't showing at all if none are installed but for scrapers it does.
3. Check for Updates does add a column for latest version but it stays empty.

I have to say, great work with this already!! Installing/updating scrapers has been a major hassle and I love to see this addressed. Along with the UI plugin API v24 shapes up to be an absolute gamechanger for versatility and usability.

scruffynerf:
> > encountered an issue, some scrapers have many YMLs to a single py script, Algolia_* is a prime example of this, how do we want to deal with things like this or `py_common`, direct the user to get the dependency from the repo?
> 
> For Algolia, I would just put all of the related ymls into a single package. I haven't yet got a solution for dependencies, though the short term solution would be to bundle the dependencies into the package. Obviously has issues with redundancy, but that's a less major issue.

But this doesn't solve the Cropper.js issue, as an example.
If we have three different crop plugins (Scene, Performer, Tags) all of which use Cropper.js, we don't want to include it 3 times.

It's far nicer if we have Cropper.js in it's own plugin with a depends.

stg-annon:
I think the solution here is a very limited dependency implementation where you can only depend on packages within the same Source and the dependency is the package ID within that source
```yml
requires:
 - py_common
```
going though each dependency and recursively installing each pluginID/scraperID in the `requires` field should leave you with all the deps and the selected package

this makes it more reliant on Source/Package maintainers organizing things in a way that works but *should* be relatively easy to implement on the stash side of things

this will  not do any cleanup of dependencies if a package is uninstalled it essentially just automatically does what the user would have to do when a package calls for a dependency i.e.
> You need to install the package 'py_common' from the community repo!

user then goes and check the box for `py_common` and hits install


## Implementation 

say at [manager.go#L156](https://github.com/WithoutPants/stash/blob/plugin-manager/pkg/pkg/manager.go#L156) we add a function call to `ListPackageRequirements` which if implemented would return the list of IDs in the requires field or nul

```go
reqs, err := m.ListPackageRequirements(remoteURL, id)
if err != nil {
	return fmt.Errorf("retrieving requirements: %w", err)
} 
if reqs != nul {
	for reqId := range reqs {
		m.Install(remoteURL, reqId)
	}
}

```

Apologies if I am off base here I don't know the nuances of go so I could be very wrong with how this would work

scruffynerf:
I'm confused why zip is being used?   Where are these zips coming from?

stg-annon:
@scruffynerf 
> I'm confused why zip is being used? Where are these zips coming from?

ZIPs are transportable and can be hashed as a single file, they are generated and hosted on GithubPages to not run afoul of GitHubs TOS, its more of an automated process with Actions not a manual process

This approach helps with version control as once the package is archived it will stay that way with that hash unless updated, so there is no concern of alterations after a package is approved and added by maintainers (Plugin v1.2.3 == sha256 hash)

the way WP has implanted it is to retroactively work with the current CommunityRepos, but as I describe in one of my comments we could separate out the index from the hosting of the packages themselves https://github.com/stashapp/stash/pull/4242#issuecomment-1781557931

WithoutPants:
Thanks for the feedback everyone. I've made the following changes:
- packages now have a free-format `metadata` map field, and description has been removed in favour of using this field. This can be used to populate scrapers URLs and such. I don't yet have an example of this in the wild. My CommunityScripts fork has been updated to move description into the `metadata` field.
- packages now have a `requires` field, which is a list of package ids that are required by the package. There is no server logic for this, but on the client side, when a package is selected, its required packages are also selected, and a note is added to the description field in the UI
image
- the package zip path can now be a URL or an absolute or relative path. Local package list paths are no longer supported in favour of `file://` URLs. `file://` based package paths are disallowed for non-`file://` based package lists.
- I've improved the loading and error states when loading a package list

Apart from documentation, one thing that I want to squeeze in is some ability to show only selected packages in the UI. Seeing what you have selected - especially wrt package dependencies - is currently onerous. I haven't been able to settle on a UI approach that I'm happy with, so I'll be seeking feedback on that. I also need to go through the bugs that have been reported.

stg-annon:
Thanks for the update, I think a basic way to improve the UX would be to stretch the interface down and wider for at least the package manager portion, possibly a whole new page for all the package management things.

Some basic testing with `requires` looks good but it doesn't seem to select all chained dependencies even though it shows the requirement

[Example Source](https://gist.githubusercontent.com/stg-annon/4c1a09733f930784ed584c49d4a9126d/raw/0caf7d6e9a8dc54aec8a542b9344293eee5df3a1/index.yml)

image

stg-annon:
Something else to note, the package manager seems to default to adding all packages to the root of their respective folders, I personally like to subdivide some of these folders for organizational purposes

for example my scrapers folder looks like:
```
scrapers/
├─ CommunityScrapers/
├─ MyScrapers/
```
where all `CommunityScrapers` is a selection of scrapers from the community repo and anything I use personally is in `MyScrapers` this way I can easily organize and find things I'm working on and continue to use things from the `CommunityScrapers` repo without mixing them, the same concept applies for plugins

My suggestion would me to have an a configurable value so that managed packages could be placed into their own folder so continuing with the above example
```
scrapers/
├─ ManagedPackages/
├─ CommunityScrapers/
├─ MyScrapers/
```

scruffynerf:
> My suggestion would me to have an a configurable value so that managed packages could be placed into their own folder so continuing with the above example

I think using an 'source' folder would be super handy.  It would at a glance let us decide if a given plugin was the stock "Community Repo" version, a customized version (STG repo), or even hand installed.

Adding one "source" value to the Index.yml would allow us to identify things.  
https://stg-annon.github.io/CommunityScrapers/develop/index.yml would just get "source: STG scrapers"
and if @stg-annon had a private version of that (with a changing hash as it was developing), it could even be "My scrapers"

stg-annon:
@scruffynerf 
> I think using an 'source' folder would be super handy. It would at a glance let us decide if a given plugin was the stock

the source name is already kind of defined when adding an index, we could use that as a friendly folder name for the index or define a value at the top of a given index that is defined by the index maintainers instead of in every package
image

we would want to avoid cross index dependencies I think that would just overcomplicate things if packages require anything that is on the index maintainer to provide within that index and since all dependencies are relative to their index this should cause no problems

scruffynerf:
To be clear, though... if the complexity of nested folders is an issue, forcing us to use a single folder per item, no nesting, I'm ok with that.  We're not going to check out a repo into the scraper directory anymore, etc.

WithoutPants:
> My suggestion would be to have an a configurable value so that managed packages could be placed into their own folder so continuing with the above exampl

The current assumption that packages are installed in their own folder with name as `packageID` makes the code for finding a given package very simple. Removing this assumption is going to take a bit of development time.

stg-annon:
the config would apply to all packages not a config for each package, similar to a how we configure a plugins folder or scrapers folder within stash this would be a "managed plugins folder"

It really should not change any assumptions about the packages besides where we start to look for them, the idea is that we shift the folder down one level under the "root" scraper/plugin folder allowing for the root to still be manually organized without cluttering that folder

I saw two options to do this

1. a static path, so we are always looking in the same place for any managed package, this would work the same as it currently works just the package manager is always looking at these folders instead of directly the root folder
	`<plugins_path>/ManagedPlugins/`
	`<scrapers_path>/ManagedScrapers/`

2. adding all managed managed packages into folder relative to their index, this would create the folder for each Source/Index but then be the same after that, the advantage here is that it could avoid package collisions as each Index is separated but it also means multiple indexes cant depend on each other
	```
	<plugins_path>/
	├─ <index_source_name>/
	│  ├─ <plugin_id>/ 
	```
	```
	<scrapers_path>/
	├─ <index_source_name>/
	│  ├─ <scraper_id>/ 
	```

WithoutPants:
Added the ability to set local paths for package sources, and changed the caching behaviour so that it stores package lists in the cache directory (if set) and only re-downloads if it is newer. Should be good for another round of testing.

stg-annon:
This sill appears to be an issue https://github.com/stashapp/stash/pull/4242#issuecomment-1788338410

WithoutPants:
> This sill appears to be an issue [#4242 (comment)](https://github.com/stashapp/stash/pull/4242#issuecomment-1788338410)

Should be addressed now.

stg-annon:
> > This sill appears to be an issue [#4242 (comment)](https://github.com/stashapp/stash/pull/4242#issuecomment-1788338410)
 
> Should be addressed now.

Yup looks good

stg-annon:
How do we want to deal with `requires` attribute is this something that is exclusive to the Source Repo? Defined in some requirements map that CI will pick up or will they live in the plugin/scraper .yml file?

everything works as intended with the values defined in the Source `index.yml` but if they are defined in the scraper/plugin file stash currently will error out when loading scrapers/plugins with this value defined

```
yaml: unmarshal errors:
  line 2: field requires not found in type scraper.config
```

WithoutPants:
> How do we want to deal with `requires` attribute is this something that is exclusive to the Source Repo? Defined in some requirements map that CI will pick up or will they live in the plugin/scraper .yml file?

The `requires` field is package-specific, so not something that should be in the plugin/scraper yml file. Populating the requirements is up to the source repos to sort out. Possible methods might be to add a `.requirements` file to the scraper/plugin directory, or a comment in the yml file. Either way, the index-generation script should get this information and use it to populate the index file.

olivechicago:
Wouldn't you want installed plugins to either be hidden or "greyed out"/marked as installed in the Available Plugins section of the manager?

PR #4245, opened 2023-10-26
Sub navbar render extra optional buttons for ui & interface config settings. 

elkorol:
In this pr, allows a settings configurable option  to render the following settings as subnav bar button elements

- soundOnPreview: for scene & marker wall to toggle off playing of audio while hovering over card.
- showChildStudioContent: while in a studio view toggle showing of sub studio items as well as main studio items
- showChildTagContent: while in a tag view toggle showing of sub tag items as well as main tag items
- showTagCardOnHover: ability to toggle on and off that card HoverPopover.

### **Notes**
Some pages such as Scene, Image, Marker, Studios & the Tags page, do not render even if the setting is true, any additional sub navbar elements. As these pages don't actually display such things associated with any of the settings. The icons of the buttons have different tooltips when enabled or disabled. And when enabled the FontAwesome Icon rendered Lime Green - Green I thought was too dull. I thought as these are view control elements and the fact that they render the Icons on enable green seemed suitable. But if isn't liked, I can easily remove it. . The soundOnPreview setting uses the FontAwesome icon: faVolumeXmark while disabled and faVolumeHigh while enabled. I thought about making others have dual icons. But couldn't seem to find suitable icons other than the ones for the audio setting.

### **Scenes page**
2a
2b
2c
2d

### **Images page**
3-images

### **Markers page**
4-markers

### **Galleries page**
5-galleries

### **Gallery page**
6-gallery

### **Performers page**
7-Performers

### **Performer page**
8-performer

### **Studio page**
9-studio-a
9-studio-b

### **Tag page**
10-tag-a
10-tag-b

### DIFF
--- a/ui/v2.5/src/components/List/ItemList.tsx
+++ b/ui/v2.5/src/components/List/ItemList.tsx
@@ -29,11 +29,16 @@ import { EditFilterDialog } from "src/components/List/EditFilterDialog";
 import { ListFilter } from "./ListFilter";
 import { FilterTags } from "./FilterTags";
 import { ListViewOptions } from "./ListViewOptions";
+import { ListToggleConfigButtons } from "./ListToggleConfigButtons";
 import { ListOperationButtons } from "./ListOperationButtons";
 import { LoadingIndicator } from "../Shared/LoadingIndicator";
+import { ConfigMode } from "src/models/list-filter/types";
 import { DisplayMode } from "src/models/list-filter/types";
 import { ButtonToolbar } from "react-bootstrap";
-
+import {
+  useConfiguration,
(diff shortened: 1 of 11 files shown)


scruffynerf:
I'd rather see each new button with a toggle.

WithoutPants:
I think the underlying requirement here would be better served with a plugin to add these buttons. I'm not aware of any particular demand for being able to toggle these options within the list views, and this adds more buttons to an already crowded toolbar, and a rather vague UI setting.

As part of #4453, we should endeavour to make it easy for plugin developers to add new buttons to the toolbar, and then this can be implemented as a plugin instead.

PR #4248, opened 2023-10-26
Fix performer validation

### LINKED ISSUE #4229: [Bug Report] Case Insensitivity Confuses Dupe Check
**Describe the bug**
It looks like the duplicate check for performer and studio names includes the current name. Since the dupe checker is case insensitive, this means changing the capitalization of the name tricks Stash into thinking the name is already taken and throws the error: "performer/studio with name 'X' already exists"

The only way around it is to use an intermediate name before changing the capitalization. For example, if I have a studio named `test`, I would need to name it `test2` before Stash will allow me to change it to `TEST`.

This bug does not appear to affect tags or movies, both of which require unique primary names. It also does not affect aliases for tags or studios which also must be unique.

**To Reproduce**
1. Create a new performer/studio named `test`
2. Try to rename the object `TEST`
3. See error

**Expected behavior**
Stash should be able to recognize that an object cannot be a duplicate of itself, allowing for changes in capitalization.

v0.23.1-18-g21baa23f

### LINKED ISSUE #4250: [Feature] Add case-insensitive validation for name and aliases fields
Right now, Stash allows adding identical aliases as the name or other aliases if they have different capitalization. Since search and scraping is already case-insensitive, I feel like this is redundant clutter.

Add case-insensitive validation for name and aliases fields in the frontend.

**Additional context**
This creates weird situations where draft submissions to the stash-box can have the same name added multiple times even though stash-box frontend prevents adding identical aliases. 

image
image

DingDongSoLong4:
This improves/fixes the validation on performer create/update, and packages everything into convenient `ValidateCreate` and `ValidateUpdate` functions that are then used in the resolvers.

The main issue fixed is that performers were able to conflict with themselves when updating, causing the issue in #4229.

I've then also added a check to error on an empty name, which is enforced in the UI but not in the backend.

Similar changes to the validation for the other object types (studios, tags, etc) will come in the future.

Fixes #4229
Fixes #4250


DingDongSoLong4:
Added case-insensitive duplicate alias validation. Fixes #4250

PR #4249, opened 2023-10-26
Fix Docker compiler: Upgrade to gcc-12-multilib

ikmckenz:
Building the Dockerfile for the compiler currently fails with `Unable to locate package gcc-10-multilib`. Looks like this is because the underlying image `golang:1.19` recently updated to Debian 12 "Bookworm". Debian 12 does not have gcc-10-multilib, but [does have gcc-12-multilib](https://packages.debian.org/bookworm/gcc-12-multilib). So this one-character change fixes the Dockerfile for newly built containers. 

Tested by running cross compiles and it seems to build successfully with the newly built container.

### DIFF
--- a/docker/compiler/Dockerfile
+++ b/docker/compiler/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-get update && \
     apt-get install -y automake autogen cmake \
     libtool libxml2-dev uuid-dev libssl-dev bash \
     patch make tar xz-utils bzip2 gzip zlib1g-dev sed cpio \
-	gcc-10-multilib gcc-mingw-w64 g++-mingw-w64 clang llvm-dev \
+	gcc-12-multilib gcc-mingw-w64 g++-mingw-w64 clang llvm-dev \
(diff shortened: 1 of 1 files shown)


DingDongSoLong4:
@ikmckenz I believe `gcc-12-multilib`/`gcc-10-multilib` aren't actually needed at all - have you tried building without them?

I've been trying to get FreeBSD cross-compilation to work, and I have succeeded, just haven't got around to setting up a VM to actually test it yet. That involved a whole bunch of changes to the compiler container, where I discovered that everything appeared to build perfectly even without `gcc-10-multilib`.

ikmckenz:
Have not tried building without gcc-1X-multilib. I'm hoping to add OpenBSD cross compilation and was just making this fix as the first step for that. I could wait for your FreeBSD changes to land before starting my OpenBSD work, that might save us from stepping on each others toes.

DingDongSoLong4:
My changes are over at https://github.com/DingDongSoLong4/stash/tree/fix-freebsd, if you're interested. I doubt anything will change - I just need to verify that the binary actually runs and then I'll make a PR.

ikmckenz:
Superseded by #4251

PR #4266, opened 2023-11-06
Fix erroneous filesize units

bunkmate5127:
Units are all calculated in the base 2 variants (as they should be), but were all named, and carry the units for, the base 10 variants.


SyZeeGee:
@bunkmate5127 I think there might be a regression here.  

image

Same for Detail View.  

image

Maybe some string comparison elsewhere in the codebase is causing this though as I don't see how this particular commit would cause the issue

PR #4287, opened 2023-11-15
Add fingerprint resolver

DingDongSoLong4:
Adds a `fingerprint` resolver to the various `*File` types - basically just finds the correct fingerprint in the `fingerprints` field for you.

Example:
```gql
query {
  findScene(id: 1) {
    files {
      id
      fingerprint(type: "oshash")
      phash: fingerprint(type: "phash")
    }
  }
}
```
returns
```json
{
  "findScene": {
    "files": [
      {
        "id": "101",
        "fingerprint": "7f2d03ab82d3de00",
        "phash": "8a893b9e840d0ce4"
      }
    ]
  }
}
```

### DIFF
--- a/gqlgen.yml
+++ b/gqlgen.yml
@@ -14,6 +14,7 @@ resolver:
 struct_tag: gqlgen
 
 autobind:
+  - github.com/stashapp/stash/internal/api
   - github.com/stashapp/stash/pkg/models
   - github.com/stashapp/stash/pkg/plugin
   - github.com/stashapp/stash/pkg/scraper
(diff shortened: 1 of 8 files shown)


scruffynerf:
excellent!
Sorely needed.  The change from separate fields to the combined fingerprints makes sense, and makes it extendable, but lost the ability to return a single type of hash as a result.
This returns the ability, and even makes it better.

stg-annon:
This change breaks existing queries that don't define a value for `type` as it is now a required field. Would we want to keep the backwards compatibility of making type an optional field and if undefined return all values?

PR #4291, opened 2023-11-19
OS Sandbox the Phasher Binary

ikmckenz:
An example of how operating system sandboxing can be used to harden the system without performance impact. This is accomplished with Landlock on Linux, and pledge/unveil on OpenBSD. This was tested on OpenBSD 7.4 and Ubuntu 22.04 LTS, and serves as an example for [my RFC](https://github.com/stashapp/stash/issues/4292). 
A [new dependency](https://pkg.go.dev/github.com/landlock-lsm/go-landlock/landlock) is introduced for Linux's Landlock, but no new dependencies are needed for OpenBSD's sandboxing, only the built-in sys/unix.

### DIFF
--- a/cmd/phasher/main.go
+++ b/cmd/phasher/main.go
@@ -4,6 +4,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/stashapp/stash/cmd/phasher/sandbox"
 	"os"
 
 	flag "github.com/spf13/pflag"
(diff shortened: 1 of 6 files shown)


scruffynerf:
As discussed in Discord, this opens a huge timesink security wise because once you go there, you can't stop, and doesn't address plugins/etc.

Anyone actually wanting a sandboxed Stash should run it in Docker.  And that takes no additional dev time to maintain.

PR #4298, opened 2023-11-19
Manager refactor, part 1

DingDongSoLong4:
The time has come for more refactoring.

This PR mostly deals with the initialization process. In general, the whole process is now a lot simpler and more straightforward.

The `main` function in `cmd/stash/main.go` is now a lot cleaner. A channel is now used to exit with an exit code, rather than just calling `os.Exit` in `Manager.Shutdown`. This means that everything can be properly shut down using `defer`s.

I've also moved a few things around - the manager initialization code is now in a separate `init.go` file for example. And some renaming: `config.Instance` to `config.Config` and `Manager.FFMPEG` to `Manager.FFMpeg`.

And then a small fix to the `Makefile` after #4153, to prevent "nothing to be done" warnings even when specifying multiple `flags-*` targets (e.g. `make flags-release flags-pie stash`).


WithoutPants:
Setup is broken with these changes. I'm getting a crash due to concurrent map read/write access in the viper code. If I then restart stash and retry the final setup step, it returns a fetch error. This may be an existing latent bug when trying to re-setup the system after it was created (since the config.yml was created prior to the crash).

DingDongSoLong4:
I wasn't able to reproduce the crash myself, but I was able to find a data race which I assume was the cause with the help of `-race`. I was also getting the same data race warning on `develop` - the bug appears to have been there for a while.

And then the setup error I believe was simply that the error state in the Setup component wasn't being reset after clicking "Back" - so if an error occurs the first time it sticks around and you can't continue, even if no more errors occur.

PR #4307, opened 2023-11-22
UI plugin dependencies

WithoutPants:
Adds a `requires` fields to the UI config in the plugin config. This is used for UI plugins only and is used to determine the load order for javascript and css files.

TODO: documentation


scruffynerf:
looks good.

PR #4322, opened 2023-11-27
Filtered counts on studio performer subpage

dogwithakeyboard:
Following on from #4162 making a new PR as I managed to make a complete hash of updating to latest develop.

Makes the Performers' scene, image, movie and gallery counts in the Studio > Performers page return the number of scenes etc that the performer has with that studio rather than the total for the performer. On the subpage the counts filters and sorts will use the filtered number.

- Adds a `findPerformers` resolver to the `Studio` type which accepts a `PerformerFilter` and `FindFilter`, and `depth` to handle the studio hierarchy.

- Adds a new `StudioPerformer` type with its own count resolvers.

- Adds a new argument to the `makeQuery` function to override the normal sorting behaviour when queried though the studio resolver.

### DIFF
--- a/graphql/documents/queries/studio.graphql
+++ b/graphql/documents/queries/studio.graphql
@@ -14,15 +14,15 @@ query FindStudio($id: ID!) {
 }
 
 query FindStudioPerformers(
-  $id: ID!,
-  $depth: Int,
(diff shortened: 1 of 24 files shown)


Gykes:
closing as stale

PR #4364, opened 2023-12-14
Default package sources

WithoutPants:
Configures default package sources where they are not already set. Sets the package sources to the stable branches of the CommunityScrapers and CommunityScripts repositories.

Adds a release note popup communicating the changes to the scrapers and plugins systems.

Adds documentation for the new package management system.

### DIFF
--- a/ui/v2.5/src/docs/en/Manual/Plugins.md
+++ b/ui/v2.5/src/docs/en/Manual/Plugins.md
@@ -10,7 +10,43 @@ Plugin tasks can be implemented using embedded Javascript, or by calling an exte
 
 > **⚠️ Note:** Plugin support is still experimental and is likely to change.
 
-# Adding plugins
+## Managing Plugins
+
+Plugins can be installed and managed from the `Settings > Plugins` page. 
+
+Scrapers are installed using the `Available Plugins` section. This section allows configuring sources from which to install plugins. The `Community (stable)` source is configured by default. This source contains plugins for the current _stable_ version of stash.
+
+These are the plugin sources maintained by the stashapp organisation:
+
+| Name | Index URL | Recommended Local Path | Notes |
+|------|-----------|------------------------|-------|
+| Community (stable) | `https://stashapp.github.io/CommunityScripts/stable/index.yml` | `stable` | For the current stable version of stash. |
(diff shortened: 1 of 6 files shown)


ui/v2.5/src/docs/en/Manual/Scraping.md
```diff
@@ -30,18 +30,51 @@ Stash provides the following built-in scrapers:
 | Freeones | `search` Performer scraper for freeones.xxx. |
 | Auto Tag | Scene `fragment` scraper that matches existing performers, studio and tags using the filename. |
```
DogmaDragon, line 31: Does the new system impact these in any way?

ui/v2.5/src/docs/en/Manual/Plugins.md
```diff
@@ -10,7 +10,43 @@ Plugin tasks can be implemented using embedded Javascript, or by calling an exte
...
-# Adding plugins
+## Managing Plugins
+
+Plugins can be installed and managed from the `Settings > Plugins` page. 
+
+Scrapers are installed using the `Available Plugins` section. This section allows configuring sources from which to install plugins. The `Community (stable)` source is configured by default. This source contains plugins for the current _stable_ version of stash.
+
+These are the plugin sources maintained by the stashapp organisation:
+
+| Name | Index URL | Recommended Local Path | Notes |
```
DingDongSoLong4, line 21: Just to be consistent, "Index URL" should probably rather be "Source URL".

DingDongSoLong4: Hmm okay well I've checked the actual UI, and it uses "Index URL" - so this is fine, but the other mentions of "Source URL" should be "Index URL" instead.

Although I think "Source URL" is a more user-friendly term than "Index URL", so maybe we should change the UI to "Source URL" or just "URL".

DogmaDragon: I also like `Source URL` more.

ui/v2.5/src/docs/en/Manual/Scraping.md
WithoutPants: Nope. These scrapers are always available and not managed by the new system.

PR #4371, opened 2023-12-17
Fix studio svg scaling

### LINKED ISSUE #1472: [Bug Report] svg logos scale really small
Noticed a svg logos that overlay scenes does scale really badly. 
It scales into a very small size, hard to really read it cleanly and clearly.
svg files are vectors so i would guess it would scale to any size the user wants.
maybe the scaling should take up a set % of width of the thumbnail or studios should have a setting you can set per studio.

image
image

source logo:
<https://www.eroticax.com/m/cmxui2ekv9wso044/EroticaX-GREY-2019Logo.svg>

WithoutPants:
Fixes #1472

### DIFF
--- a/ui/v2.5/src/components/Scenes/styles.scss
+++ b/ui/v2.5/src/components/Scenes/styles.scss
@@ -139,8 +139,8 @@ textarea.scene-description {
   z-index: 8;
 
   .image-thumbnail {
-    height: auto;
-    max-height: 50px;
+    height: 50px;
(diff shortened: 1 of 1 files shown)


WithoutPants:
Thanks to @DingDongSoLong4 for the more compatible fix.

JanJastrow:
On the studios page the logos render at different widths, depending of SVG or Pixel graphic.
Might also need an CSS adjustment

PNG
SVG

PR #4382, opened 2023-12-20
Package manager UI-related tweaks

DingDongSoLong4:
These are a bunch of improvements and tweaks mostly related to the package manager UI.

- 07b2e78 is almost identical to #4377, the only difference being that it also refreshes the package managers when the plugin or scraper paths change, in addition to the plugin/scraper caches.
- 9e124de adds cache invalidation to refetch the plugin/scraper queries when the paths are changed, and improves the `SetPluginsEnabled` mutation to modify the cache client-side and avoid any refetching. I've also done some minor refactoring (the package manager mutations now use `client.mutate` rather than hooks), and I've removed the `sources` state which is no longer necessary since the cache is now being properly invalidated.
- 75e3dfc does what it says on the tin. Currently, if retrieving the available package list from a source fails, then the error message sticks around even after collapsing the source. Now the error will be cleared if you collapse the source, and it will retry the query if you expand the source again.
- 36bccb7 makes two style tweaks to the package manager UI. The first is to add some padding to the toolbar, which prevents the highlight on hover from being cut off at the bottom. The second is to add a gap between the "Edit" and "Delete" buttons - them being directly next to each other looks off to me. And then a minor refactor to the way the toolbar's flexbox works, to use a `flex-grow` spacer rather than `space-between`.
- 0db0840 also is fairly self-explanatory. Currently, if the installed packages query fails, then it just shows "No packages found" - so I've just made it display a simple error message instead.
- 704af96 prevents a "No packages found" message from showing while the installed package list loads after clicking "Check for Updates". 
- cabd147 makes the list display `<unknown>` if the package has no version specified, rather than showing nothing.
- a5d03b2 makes the "Latest Version" column always display the latest version info, rather than the current behaviour of just displaying it if the version is newer than the installed version. The `upgrade` GraphQL field is now `source_package`, and will always return a result if the package is available from the source. This change means that an upgrade needs to be highlighted in some other way, which I've done by making the text bold and adding an icon.
- efc8277 fixes two issues I came across when you have packages from different sources sharing the same ID (in my case, `py_common`). The first is that the `installedPackages` query doesn't return a stable sort order, which happens because `sortedPackageSpecKeys` only sorts by ID - it will now fall back to the source URL. The second issue happens when deselecting packages - if you select both `py_common` packages, and then deselect one of them, both of them will be deselected. This is again simply fixed by including the source URL in comparisons.

I've tried my best to demonstrate most of the changes in these two screenshots:

Before:
before

After:
after

Closes #4377

### DIFF
--- a/graphql/documents/data/config.graphql
+++ b/graphql/documents/data/config.graphql
@@ -9,6 +9,7 @@ fragment ConfigGeneralData on ConfigGeneralResult {
   generatedPath
   metadataPath
   scrapersPath
+  pluginsPath
   cachePath
   blobsPath
   blobsStorage
--- a/graphql/documents/queries/plugins.graphql
+++ b/graphql/documents/queries/plugins.graphql
@@ -55,7 +55,7 @@ query InstalledPluginPackages {
 query InstalledPluginPackagesStatus {
   installedPackages(type: Plugin) {
     ...PackageData
-    upgrade {
+    source_package {
       ...PackageData
     }
   }
--- a/graphql/documents/queries/scrapers/scrapers.graphql
+++ b/graphql/documents/queries/scrapers/scrapers.graphql
@@ -129,7 +129,7 @@ query InstalledScraperPackages {
 query InstalledScraperPackagesStatus {
   installedPackages(type: Scraper) {
     ...PackageData
-    upgrade {
+    source_package {
       ...PackageData
     }
   }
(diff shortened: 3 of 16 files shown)


DingDongSoLong4:
Added two more tweaks:
- 14bd231 prevents the package id, version and date fields from wrapping to the next line. This change is easiest to see with the CommunityScripts source, where the long descriptions "squash" the version column.
- 4fb7d17 decreases the padding around the collapse button, allowing the rest of the table to take up more space.

Before:
before

After:
after

PR #4404, opened 2023-12-29
added support for image orientation filter

keenbed:


WithoutPants:
Thanks for this. I've addressed some issues I found, added support for multiple values and added the orientation filtering to scenes as well.

PR #4414, opened 2024-01-01
Add join to intCriterionHandler

### LINKED ISSUE #4413: [Bug Report] Filtering by Interactive speed returns an error
**Describe the bug**
Filtering using the Interactive speed option gives an error like below:
`error finding IDs: running query: SELECT DISTINCT scenes.id FROM scenes WHERE (video_files.interactive_speed = ?) ORDER BY scenes.date ASC, COALESCE(scenes.title, scenes.id) COLLATE NATURAL_CI ASC LIMIT 40 OFFSET 0 [[1]]: error executing ``SELECT DISTINCT scenes.id FROM scenes WHERE (video_files.interactive_speed = ?) ORDER BY scenes.date ASC, COALESCE(scenes.title, scenes.id) COLLATE NATURAL_CI ASC LIMIT 40 OFFSET 0 `` [[1]]: no such column: video_files.interactive_speed`

The video_files table is not in the query, so the column is not found.

**To Reproduce**
1. Filter Scenes by Interactive speed
2. Sort by anything which is not a property of the video file like Date or Title

**Expected behavior**
The filter should work and not give an error.

v0.24.1 bd7d4ac7

bayured:
Fixes #4413

Allow intCriterionHandler to add a join to the query if required like other handlers. Not sure if floatCriterionHandler need the same done, I don't think any filter that uses it is broken at the moment.

### DIFF
--- a/pkg/sqlite/filter.go
+++ b/pkg/sqlite/filter.go
@@ -542,6 +542,9 @@ func getPathSearchClauseMany(pathColumn, basenameColumn, p string, addWildcards,
 func intCriterionHandler(c *models.IntCriterionInput, column string, addJoinFn func(f *filterBuilder)) criterionHandlerFunc {
 	return func(ctx context.Context, f *filterBuilder) {
 		if c != nil {
+			if addJoinFn != nil {
+				addJoinFn(f)
+			}
 			clause, args := getIntCriterionWhereClause(column, *c)
 			f.addWhere(clause, args...)
 		}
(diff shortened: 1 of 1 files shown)


DingDongSoLong4:
I'd say`floatCriterionHandler` should also have the same change made, or the `addJoinFn` should be removed since it doesn't actually do anything.

Otherwise, can confirm this fixes the interactive speed filter issue 👍

DingDongSoLong4 (approved):

Issue #4111, opened 2023-09-12
[Feature] Cumulative options filter


Gouvernathor:
![image](https://github.com/stashapp/stash/assets/44340603/c8999d43-1a01-4ec8-bdb4-a22bb1d86bc4)

The buttons here are round, forcing me to choose only one option. I would like to be able to select several options and see the performers (in this case) which satisfy any of the provided options (in this case where it's mutually exclusive).

DingDongSoLong4:
I believe this is a duplicate of #2593.

Gouvernathor:
The example is, but my request is more general and not restricted to gender

DingDongSoLong4:
Alright, which filter criteria are you referring to then? Most of the criteria support choosing multiple options - the only one I can think of that doesn't is the "Is Missing" criteria, but that is slowly being phased out for more specific criteria.

Gouvernathor:
After examination three other cases come up :
- Is missing which you covered
- Resolution but it could also be done adding a "is between" and "is not between" as for Rating
- Captions

PR #3009, opened 2022-10-15
[Help Wanted] Extend Performer Gender Filter

stg-annon:
resolves #2394

Issue #2593, opened 2022-05-16
[Feature] Extend gender filter options on /performers


armbyants:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to filter `/performers` by gender using boolean operators other than exact match. Today, I have to select between `Male,` `Female`, `Transgender Male`, etc. (screenshot below). I'd like to be able to filter for `IS NULL`, `IS NOT male`. etc.

<img width="701" alt="image" src="https://user-images.githubusercontent.com/49548655/168700052-72645585-1bd9-42d2-bb33-cb62f8d74e2e.png">

A specific use case that I often run into is when new performers have been created with default `null` gender (such as being added during a scrape), and I want to find all performers without stashIDs that I haven't already organized. I'd like to be able to add `stashID IS NULL` and (`gender IS NULL` or `gender IS NOT MALE`) filter options to find them.

**Describe the solution you'd like**
Extend performer gender filter matching logic to match others such as `URL` or `checksum`

<img width="518" alt="image" src="https://user-images.githubusercontent.com/49548655/168700706-8af2b483-dcb7-4536-8240-d0b10a54c8ea.png">

### TL;DR 
I want the above filter logic options (or at least an option to match `NULL`) for `gender` in `/performers`.

Happy to provide more context if helpful. Cheers 🎉

stg-annon:
#2394

WithoutPants:
Duplicate of #2394

PR #4419, opened 2024-01-04
Update Performer gender filter

### LINKED ISSUE #2394: [Feature] Extend Performer Gender Filter
The current gender performer filter lacks things like `includes`, `excludes`, `is_null`, and `is_not_null` currently it is not possible to filter where the gender 

Extend the performer filter to allow for things like `is_null` without having to use the deprecated `is_missing` filter as well as the ability to do exclude filters on the gender attribute

ofpcode:
Updated the Performer gender filter to allow selection of multiple genders (`IS`) or performers with no gender specified (`IS NULL`).

- [x] I've run code linters and formatters to make sure that my code is readable.
- [x] I have read through formerly submitted pull requests and issues to make sure that this contribution is required and isn't a duplicate. Also, so that I can manage to close any issues needing closed relating to this feature submission.
- [x] I commented adequately on my code with the expectation in mind that anyone else should be able to look at this code I've submitted and know exactly what's happening and what the expectations are.

Closes or partially satisfies #2394, #2593, and #4111. Supersedes draft PR #3009.

Resolves #2394


graphql/schema/types/filters.graphql
```diff
@@ -544,7 +544,7 @@ input MultiCriterionInput {
 }
 
 input GenderCriterionInput {
-  value: GenderEnum
+  value: [GenderEnum!]
```
WithoutPants, line 547: This will break existing external systems, since they expect to send a single value. We generally add a new field and deprecate the old one, supporting both fields until the deprecated field is removed.

WithoutPants (changes requested):
Thanks for the submission. This has been a long-standing issue that is definitely in need of fixing.

WithoutPants (approved):

PR #4428, opened 2024-01-06
Scene queue autoplay

### LINKED ISSUE #4427: [Bug Report] Disabling "Auto-start video when playing selected" does not affect scene queue
**Describe the bug**
When using the scene queue, the `autoplay` parameter is always set to true, even when "Auto-start video when playing selected" is disabled.

**To Reproduce**
1. Disable `Auto-start video when playing selected` option in Settings, Interface
2. Enter a scene queue.
3. Click the next/prev button or use keyboard shorcuts: `p n`, `p r`
4. Url contains `autoplay=true` and video starts automatically 

**Expected behavior**
When `Auto-start video when playing selected` is disabled, entering a new scene via the queue should not auto-start the video.

v0.24.1-29-g7a29ef2d1_jellyfin-ffmpeg

 - Browser Chrome
 - Version 120.0.6099.71

DingDongSoLong4:
In #3020, I unintentionally sort-of reversed the change in #2896, making it so that navigating the scene queue would _never_ autoplay when navigating to a new scene, irrespective of the `Auto-start video when playing selected` setting. That PR went through many iterations so it might have been a rebase mistake, or just an oversight on my part. I have always had that setting disabled, which is why I didn't notice a difference in behaviour.

In #4325, the scene queue was changed to now _always_ autoplay new scenes, which was the behaviour before #2896. 

So this PR makes it so that the scene queue now once again respects the `Auto-start video when playing selected` setting. I've also, as usual, taken the opportunity to do some refactoring, and removed some unnecessary `undefined` checks in and around the `QueueViewer` component.

Fixes #4427

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneDetails/QueueViewer.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneDetails/QueueViewer.tsx
@@ -15,7 +15,7 @@ import { objectTitle } from "src/core/files";
 import { QueuedScene } from "src/models/sceneQueue";
 
 export interface IPlaylistViewer {
-  scenes?: QueuedScene[];
+  scenes: QueuedScene[];
   currentID?: string;
   start?: number;
   continue?: boolean;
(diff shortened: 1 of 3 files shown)


smegmarip:
This issue is back in v0.27.1
https://github.com/stashapp/stash/issues/4427#issuecomment-2412270807

PR #4443, opened 2024-01-09
Fix error when creating/updating performer with alias == name

### LINKED ISSUE #4437: [Bug Report] Adding Performer from Scene Tagger Fails
**Describe the bug**
When scraping a scene from stash-box in the scene tagger view, the user is prompted to add missing performers with their data pre-filled from stash-box. If a performer being added has an alias matching their primary name, adding them to the user's stash instance will fail.

**To Reproduce**
1. Go to the scene tagger view and select a stash-box instance as the source
2. Choose a scene which exists in stash-box and for which not all performers are in the local database (if necessary, delete a performer or their Stash ID for testing)
3. Click on 'Scrape by fragment'
4. For the missing performer, click 'Create' and then 'Save'
5. If the performer has an alias matching their primary name, an error message will be displayed

**Expected behavior**
The duplicate alias should be ignored and the performer added as normal. Alternatively, details should be editable in the "Create Performer" modal, which would allow the user to manually delete the bad data.

**Screenshots**
image

 - OS: Windows 10
 - Browser: Chrome
 - Version: 0.24.1

WithoutPants:
Fixes #4437 

Filters out aliases that match the performer name when returning performer data from stash-box. Also patches a hole in the stash-box performer add/update task where the performer data would not be validated before creating/updating.

### DIFF
--- a/internal/manager/task_stash_box_tag.go
+++ b/internal/manager/task_stash_box_tag.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/models"
+	"github.com/stashapp/stash/pkg/performer"
 	"github.com/stashapp/stash/pkg/scraper/stashbox"
 	"github.com/stashapp/stash/pkg/studio"
 )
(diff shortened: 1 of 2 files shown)


Maista6969:
Does this account for the aliases being differently cased? `Morgan LaFey` with the alias `Morgan Lafey` for example

WithoutPants:
Argh, no. I'll test this particular case.

Maista6969:
Tested this with a few hundred performers from the list with duplicate aliases and a few hundred I already had, no issues 👍

PR #4461, opened 2024-01-16
Import video rotation from side_data_list as well

### LINKED ISSUE #4233: [Bug Report] Rotation information is ignored for determining if a video is portrait
**Describe the bug**
A portrait clip recorded on an iPhone will be stored as landscape but with rotation information embedded.
The clip plays fine and everything work except for the scene cards that do not get the css class portrait.

Relevant ffprobe output:
```
Stream #0:0[0x1](und): Video: hevc (Main) (hvc1 / 0x31637668), yuv420p(tv, bt709), 3840x2160
...
Metadata:
  creation_time   : xxxxx
  handler_name    : Core Media Video
  vendor_id       : [0][0][0][0]
  encoder         : HEVC
Side data:
  displaymatrix: rotation of -90.00 degrees
```

**To Reproduce**
1. Record a clip vertically with an iPhone
2. Scan it
3. The thumbnail and the video preview in the scene card will only show the top part of the image.
4. The scrubber in the scene card will show the whole image as expected

Somewhat related to this:
1. For portrait mode clips the scrubber in the player page only shows the upper side of the image.
2. For portrait mode clips that have landscape thumbnails (think a short clip from yt downloaded with yt-dlp) the video preview is taller than the thumbnail. To fix it I had to disable making the portrait mode taller by adding the custom css (only for zoom-1, since that is what I use):
```
.zoom-1 .portrait {
  height: 180px;
}
```

**Expected behavior**
The way the scrubber looks on the scene card is the way that the portrait thumbnail and video preview should look like on the scene card.

Few days old custom build

puc9:
Imports rotation info for a video out of the first side_data_list structure if available and if rotate tag is not present.

Fixes #4233 but I'm not sure if the fix is robust enough. It works on all my affected files.

One issue left unsolved is that once a file has been added to the DB with the wrong rotation info then it is impossible to fix it short of directly editing the DB.

If I remove the scene without deleting the scene file then the file row in the DB is not removed so the rotation information is not removed either.
Ideally a rescan should reimport the rotation info but I couldn't figure out how to do that.

### DIFF
--- a/pkg/ffmpeg/ffprobe.go
+++ b/pkg/ffmpeg/ffprobe.go
@@ -176,7 +176,12 @@ func parse(filePath string, probeJSON *FFProbeJSON) (*VideoFile, error) {
 			framerate = 0
 		}
 		result.FrameRate = math.Round(framerate*100) / 100
-		if rotate, err := strconv.ParseInt(videoStream.Tags.Rotate, 10, 64); err == nil && rotate != 180 {
+		rotate, err := strconv.ParseInt(videoStream.Tags.Rotate, 10, 64)
+		if err != nil && len(videoStream.SideDataList) > 0 {
+			err = nil
+			rotate = videoStream.SideDataList[0].Rotation
+		}
+		if err == nil && rotate != 180 {
 			result.Width = videoStream.Height
 			result.Height = videoStream.Width
 		} else {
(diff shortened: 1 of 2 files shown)


Gykes:

PR #4497, opened 2024-01-25
Improve list table readability

cj12312021:
This is a simple change to make identifying non-actionable cells easier by dimming the text colors in those cells compared to ceilings that use anchors.
Screenshot 2024-01-25 113017


cj12312021:
My latest push includes a fix for a bug that allowed mandatory columns to be unselected if you click on the label. The latest push also addresses consistency issues with the column selector style.

cj12312021:
I've pushed a few more improvements to the table. My latest push makes it so that columns such as resolution, codec, frame rate, etc, do not use the float on hover styles to present the cell data when hovered. The primary reason for updating the styles on those columns was that they were much wider than they needed since the data they stored wasn't lengthy. They also don't have actionable elements.

My latest push also reorders the placement of the Tags, Performers, and Studio column (previously in this order) to Studio, Performers, Tags.

PR #3922, opened 2023-07-17
Recording Play/O Dates with Play Counter & History Panel

randemgame:
This pull request implements functionality that has been requested here https://github.com/stashapp/stash/issues/3392 and https://github.com/stashapp/stash/issues/3607 and https://github.com/stashapp/stash/issues/3680. I have been unable to commit fully to Stash due to this missing feature and wanting to transfer data I've recorded in older organisers. I feel this feature will be beneficial to all users wanting to keep a more accurate historical record of their Stash usage and is something that would ideally be implemented soon so that user's play/o date history can stop being lost to time.

Screenshot 2023-07-17 221856

The Play Counter is modeled on the O Counter and functions similarly. The icon is custom made myself for Stash and I think it fits nicely in the interface, but it's fine to replace it.

Incrementing/Decrementing or Resetting the Play Counter or the O Counter now writes or erases a date to two new tables in the SQL database called scenes_playdates and scenes_odates - consisting of an id, scene_id and play/odate. I have kept the odates and playdates in separate tables as per this comment from WithoutPants https://github.com/stashapp/stash/issues/3607#issuecomment-1487947899

The History Panel provides a view that consolidates date information currently spread across several panels, and further incorporates the new Play/O Date data to provide a chronological view; from when the scene was released to when the scene file was created to when it was added to Stash, updated in Stash and finally watched or O'd. While I have kept the Release Date in the main 'Details' window, I think the Created At/Updated At information is probably less relevant to be displayed there, compared to the scene's Director, Description, etc (i.e the difference between the scene's studio data and what is just specific to the file in stash). Likewise, I think the Play Count/Play Duration is out of place in the 'File Info' tab, whilst moving the 'File Modification Time' (i.e File Creation Time) along with all the other dates to the History Panel enables this interesting timeline, as visible in the attached screenshot. Each recorded Play or O Date then adds a new line to this display, but they can always be viewed at a glance and modified via the counters across all panels. There are a number of different implementations that presenting the Play/O Dates could have taken but I think I found a nice solution here in collating all the date data associated with a scene and presenting it in this linear timeline. 

[...]

PR #4532, opened 2024-02-08
Scene play and o-counter history view and editing

### LINKED ISSUE #3392: Improved Play Counter (i.e. modify the scene's 'Last_Played Date' variable to store multiple plays rather than only the most recent)
While I really appreciate the work towards the Play Counter https://github.com/stashapp/stash/issues/633 that was developed here https://github.com/stashapp/stash/pull/3055, it is still missing some essential functionality that is available in other, far less developed organisers, and it's an issue that has prevented me from switching to Stash for several years now. I had already mentioned these issues in comments lost to those closed/merged issues, and I could not find any other related discussion on the topic here, so I'm finally writing up this request as best I can, as three very interconnected issues. 

**1: The Play Counter should be recording the multiple DATE's a scene is viewed**

The current implementation of the Play Counter does not seem to keep track of anything more than the singular date that a scene was last played. Adjusting this to record the multiple views of a video would allow for a _more accurate record of the user's historical play data_. This sort of functionality is especially valuable as more and more personal view history is gathered over the years, providing a cool timeline of information for the user, e.g. to see how their interests have developed over time or just to reminisce over what video was played on Valentine's Day a decade ago. All this info currently seems to be lost when a scene is re-watched and then just jumps to being the last played entry, removing any historical date information. I can think of two options for implementing this:

a) Ideally this history would be displayed when browsing Scenes in the order of 'Last Played At'. However, I realise that this would then require the same videos to be displayed multiple times in the view, which might be a problem to implement?

b) A separate, dedicated 'view history' part of the interface. This is the approach taken by Porn-Vault, but is entirely separate from the rest of the Scenes interface.

I would hope that this enhanced view history could then be one of the options to see on the Stash front page, but really more than anything I just need this multiple date history to be recorded in the database.

**2: The DATE of Plays should be visible to the user alongside the functionality to add/remove/edit the dates**

This was something requested in the original issue that has not yet been implemented: "On the Edit tab, the functionality to decrease or reset the count (of plays)". The ability to remove dates would have a use case if the user accidentally plays a scene or otherwise has a change of mind on what level of activity they want to track. The real need I see for this function is to _manually add dates when viewing scenes outside of Stash_. While I appreciate Stash is accessible on almost every device, there are still many instances where users may be viewing videos from their library outside of the interface and still want to keep track of the play data. I currently change the system clock in order to accurately record what time and date these views happened outside of other organisers, so the option to _freely edit the dates_ would be a great quality of life improvement.

With the ability to add, remove or edit the dates, I am really advocating more _manual control of the play data_, which I prefer over Stash's automated implementation. I think the 'Minimum Play Percent' method of tracking plays is useful, and could be used in combination with manual tracking, but I would still like an option to disable it whilst still 'Tracking Activity'. I'm sure there are a number of other users who prefer to take control of things manually.

**3: This multiple DATE information should be accessible via a plugin or scraper so that play data can be imported from other organisers**

I would really need that with any implementation of this functionality, that the multiple DATE data attached to a scene would be accessible in a way that I could import my historical view data. This is the last important piece of data keeping me tied to the other organisers, and I'm sure there are others in a similar situation. I have already simplified the data I have in the PGV4 database (which recorded both the date a scene was watched, and the date their equivalent 'O-counter' was used) into a JSON, importing the view history alone into Porn-Vault (which does not record O's), which is enough data for my needs. The lack of any date info on the O-Counter in Stash had been preventing me using the program, but I think it would be fine to remain a simple integer as long as the view data was being additionally recorded somewhere. 

I can no longer ignore how advanced and impressive Stash is, especially since gaining access to Stash-DB, and would _really_ like to switch to the software and hopefully contribute to the community. Please help myself and others escape the trappings of proprietary or abandoned organisers, by implementing this missing functionality and providing a way to rescue my data one last time.

### LINKED ISSUE #3607: [Feature] O-History
Presently Stash implements O-Counter as a single number for each scene/image. I would like to be able to track the history of "O-Events", i.e., they should be stored as timestamped rows in a new table.

- Replace the "o_count" column of scenes/images tables with a new "o_history" table, which has columns "timestamp", "scene_id", "image_id"
- When displaying O-Count elsewhere in the UI, perform a simple query on the o_history table
- Pressing the "O" button adds a new row with the current timestamp
- Decrementing the O-Counter is replaced with removing the most recent event from the o_history table matching that scene/image (and resetting the count removes all) 
- Optionally, a UI that displays the O-History

None.

**Additional context**
The idea is to use Stash for O-Tracking.

### LINKED ISSUE #3680: [Feature] Manual Control Over Watch Activity
Currently, all watch activity metrics are tracked automatically. You can turn tracking on/off and adjust the watch percentage for the play counter, but everything else is handled automatically. There is no easy way to manually modify any of these metrics.

The current monitoring method is imperfect. I would like to manually input watch history that pre-dated the feature, but there is no way to do that in the UI. I would also like to track plays over DLNA, but those are not monitored currently and, again, I cannot manually track those plays within Stash either.

Users should be able to manually modify or reset the last viewed date, play counter, play duration, and progress tracker. These values should be exposed to the edit panel where they can be changed or reset as needed.

I will also create a separate issue for DLNA tracking.

I guess I could edit the database directly but that's asking too much of the average user.

WithoutPants:
Supercedes #3922 
Resolves #3392 
Resolves #3607 
Resolves #3680 

This branch reimplements the #3922 changes. Adds graphql interfaces to remove and clear the play history, and adds an optional `times` parameter to the increment/decrement view count and o-count interfaces. This allows adding and removing multiple dates to/from the view/o-count history.

Unlike #3922, this transitions to using two tables for view and o histories, and removes the `play_count`, `o_count`, and `last_played` fields from the `scenes` table. These values are now derived from the applicable tables.

From the other PR, I have removed the views button in favour of putting all of the info and manipulation into the `History` tab.
The `History` tab allows for adding to the view/o-counts, deleting specific entries, clearing the entire date history, and entering specific dates.

image

The UI should be considered the minimum viable for now. I have focused on functionality - further refinements can be done in future PRs. Minor changes will be considered, but I won't be spending much more time on the UI.

Like the previous PR, the migration uses the last played date value to populate the latest play time, with the rest populated with the created time of the scene. All o-count entries use the created time when migrated.

@randemgame: I'd appreciate if you could test this. Timely testing will dictate if this goes into 0.25 or the following release.

Affected areas will be:
- history tab (obvs)
- o-counter button
- play/o-count sorting and filtering
- performer o-count sorting and filtering (~~**known issue discovered while typing this**~~ fixed)
- scene merging (play/o history merging)
- import/export


jpnsfw:
@WithoutPants huge thank you for working on this one!!! I tried out your branch locally and just had two comments:

- The Stats page gives an error "querying `SELECT COALESCE(SUM(`play_count`), 0) FROM `scenes`` [[]]: error executing `SELECT COALESCE(SUM(`play_count`), 0) FROM `scenes`` [[]]: no such column: play_count"
- Looks like old plays and O's show up with date "December 12, 2022 11:18 PM" (at least on my end), would it make sense to instead set these as NULL and render text like "Unknown" in the UI?

WithoutPants:
> @WithoutPants huge thank you for working on this one!!! I tried out your branch locally and just had two comments:
> 
> * The Stats page gives an error:
> 
> ```
> querying `SELECT COALESCE(SUM(`play_count`), 0) FROM `scenes`` [[]]: error executing `SELECT COALESCE(SUM(`play_count`), 0) FROM `scenes`` [[]]: no such column: play_count"`
> ```
> 
> * Looks like old plays and O's show up with date "December 12, 2022 11:18 PM" (at least on my end), would it make sense to instead set these as NULL and render text like "Unknown" in the UI?

Thanks for testing and the feedback. Both these issues should now be addressed.

WithoutPants:
From Discord:
> I just noticed something - the red trash can for 'Remove date' did not work for either the o_date or the view_date (although Clear date data worked fine). I could probably have imported the data better, for now the dates all had the +00:00 ending, e.g. 2017-11-08T19:49:57+00:00. If I change the timezone to +01:00 or even some nonsense like +00:01 in the SQL database then they can be deleted again (I wouldn't care much about the data being a minute off perfect accuracy). Anyway not sure if this is an actual thing or not that would effect users in the +00:00 UTC or a quirk of my data transferring.

forsexonly93:
Thanks a lot for working on this @WithoutPants!! Could we add this to next version 0.25.0? Much needed actually to make sure we're not missing data till this is released...

randemgame:
Not a huge deal at all, but something I just remembered from this was that when viewing Stash on my mobile iPhone SE3 2022 (likely the smallest res phone available at 750 x 1334), it was no longer possible to select 'Edit' when viewing the scene page in a vertical orientation, the bar with 'Details', 'Queue' etc does not slide, and the new History option pushes Edit out the way. But then you can flip the phone horizontal and see Edit anyway, so who cares, and who is doing much editing at such a small res anyway. Thought you'd like to know regardless.
IMG_0535
IMG_0536

WithoutPants:
@randemgame I have changed the migration and storage to always use UTC dates. This should address the delete issue you encountered, provided that all dates are imported as UTC dates.

discourse-stashapp:

https://discourse.stashapp.cc/t/add-view-counts-for-images/4449/2

PR #4533, opened 2024-02-09
Fix UI config mutation not working

WithoutPants:
Fixes issue presumably introduced by #4511 where the `ConfigureUI` mutation errored out due to the server not knowing what `UIConfig` is. @DingDongSoLong4 you may have a cleaner fix for this, but this appears to fix the issue for the moment.


DingDongSoLong4:
The scalar overriding in #4511 is a tiny bit hacky, so I think this is just one of the limitations. I don't think there is a cleaner way.

PR #4538, opened 2024-02-10
Fix url not updated for default tab on Performer/Studio/Tag pages

### LINKED ISSUE #4209: [Bug Report] SQL Error upon deletion of all items in active tab, with sorting order which is non compliant to the order that the first populated content tab, after deletion can accept
**Describe the bug**
[PR: 4032](https://github.com/stashapp/stash/pull/4032) in the newest version of Stash has introduced a bug when on the Details pages, while on the active tab, upon deleting all items, if the current sort order is set to a setting that does not conform to what the sort settings of the first populated content tab that will be active upon deletion of all items, then Stash throws an sql error. Somehow the current sort order of the active pane isn't being reset to what the first populated content tab can handle, or the sort order is being passed to the new content tab after deletion. 

**To Reproduce**
1. Create a test performer.
2. Add some scenes, and a gallery with some images
3. Activate the scenes tab, change the sorting order to anything that Gallery tab sorting cannot accept. I.e. Duration Ascending.
4. Select all scenes & delete.
5. See SQL error on screen.

v0.23.1
Build hash:
479e7163

 - OS: Windows 11
 - Browser Chrome
 - Version 118.0.5993.70

WithoutPants:
I think it should be expected that the URL show the full URL on Performer/Studio/Tag (`/performers/x/scenes` vs `/performers/x` displaying the default scenes tab). I think also that some plugins are listening for URL changes, and the existing behaviour doesn't indicate the current view correctly for them.

Fixes #4209

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
@@ -48,7 +48,7 @@ import { ExternalLink } from "src/components/Shared/ExternalLink";
 
 interface IProps {
   performer: GQL.PerformerDataFragment;
-  tabKey: TabKey;
+  tabKey?: TabKey;
 }
 
 interface IPerformerParams {
@@ -65,8 +65,6 @@ const validTabs = [
   "appearswith",
 ] as const;
 type TabKey = (typeof validTabs)[number];
-
-const defaultTab: TabKey = "default";
 
 function isTabKey(tab: string): tab is TabKey {
   return validTabs.includes(tab as TabKey);
@@ -135,8 +133,7 @@ const PerformerPage: React.FC<IProps> = ({ performer, tabKey }) => {
 
   const setTabKey = useCallback(
     (newTabKey: string | null) => {
-      if (!newTabKey || newTabKey === defaultTab)
-        newTabKey = populatedDefaultTab;
+      if (!newTabKey) newTabKey = populatedDefaultTab;
       if (newTabKey === tabKey) return;
 
       if (isTabKey(newTabKey)) {
@@ -147,7 +144,7 @@ const PerformerPage: React.FC<IProps> = ({ performer, tabKey }) => {
   );
 
   useEffect(() => {
-    if (tabKey === defaultTab) {
+    if (!tabKey) {
       setTabKey(populatedDefaultTab);
     }
   }, [setTabKey, populatedDefaultTab, tabKey]);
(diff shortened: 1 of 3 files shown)


DingDongSoLong4:
Ironically I left this out of #3995 because I was told it could break userscripts (https://github.com/stashapp/stash/pull/3995#pullrequestreview-1560013826). I do agree that the previous behaviour was weird though (and now buggy as well).

Just a small potential refactoring I noticed, I think would be a bit cleaner to use `undefined` rather than a magic `"default"` value for the placeholder default tab key.

So something like this, for the performer page:
```diff
diff --git a/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx b/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
index 39095bd9..511e2c20 100644
--- a/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
@@ -48,7 +48,7 @@ import { ExternalLink } from "src/components/Shared/ExternalLink";
 
 interface IProps {
   performer: GQL.PerformerDataFragment;
-  tabKey: TabKey;
+  tabKey?: TabKey;
 }
 
 interface IPerformerParams {
@@ -66,8 +66,6 @@ const validTabs = [
 ] as const;
 type TabKey = (typeof validTabs)[number];
 
-const defaultTab: TabKey = "default";
-
 function isTabKey(tab: string): tab is TabKey {
   return validTabs.includes(tab as TabKey);
 }
@@ -135,8 +133,9 @@ const PerformerPage: React.FC<IProps> = ({ performer, tabKey }) => {
 
   const setTabKey = useCallback(
     (newTabKey: string | null) => {
-      if (!newTabKey || newTabKey === defaultTab)
+      if (!newTabKey) {
         newTabKey = populatedDefaultTab;
+      }
       if (newTabKey === tabKey) return;
 
       if (isTabKey(newTabKey)) {
@@ -147,7 +146,7 @@ const PerformerPage: React.FC<IProps> = ({ performer, tabKey }) => {
   );
 
   useEffect(() => {
-    if (tabKey === defaultTab) {
+    if (tabKey === undefined) {
       setTabKey(populatedDefaultTab);
     }
   }, [setTabKey, populatedDefaultTab, tabKey]);
@@ -626,7 +625,7 @@ const PerformerLoader: React.FC<RouteComponentProps<IPerformerParams>> = ({
     return <ErrorMessage error={`No performer found with id ${id}.`} />;
 
   if (!tab) {
-    return <PerformerPage performer={data.findPerformer} tabKey={defaultTab} />;
+    return <PerformerPage performer={data.findPerformer} />;
   }
 
   if (!isTabKey(tab)) {
```

PR #4571, opened 2024-02-16
Error if duplicate plugin id is loaded

### LINKED ISSUE #4484: [Bug Report] Plugin duplicates override .yml data
**Describe the bug**
plugin duplicates inherit the .yml of the lowest copy (when sorted by alpha)

**To Reproduce**
1. Add 3 otherwise idential plugins with different verisions - in my example I have a Nothing.js plugin that just prints to console
2. move plugins to seperate subfolders
3. view plugins in settings -> plugin view
4. The plugin with the lowest path sorted alphabetically has it's settings applied
[plugins.zip](https://github.com/stashapp/stash/files/14005707/plugins.zip)

**Expected behavior**
seperate plugins with the same name have seperate settings. The list of plugins installed from index.yml have the correct versions, but the plugin block with settings is incorrect.

**Screenshots**

tested from v23.0-v24.3 to develop v0.24.3-9-gdd8da7f3

N/A

N/A

**Additional context**
The plugins all behave as one and I have observed them enabling and disabling together, although I have not tested this behaviour across versions.

WithoutPants:
Resolves #4484

### DIFF
--- a/pkg/plugin/plugins.go
+++ b/pkg/plugin/plugins.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/stashapp/stash/pkg/fsutil"
 	"github.com/stashapp/stash/pkg/logger"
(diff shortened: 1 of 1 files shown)


WithoutPants:
TODO - duplicate check should be case insensitive

PR #4573, opened 2024-02-16
Add galleries to image edit panel

### LINKED ISSUE #1003: [Feature] Assign images to galleries while viewing them
I would like the option to assign an image to galleries while viewing the image.

The only way right now is to go to galleries, open the specific gallery then click on "add" and search for the image there (unless I am blind, which is also an option).

bob123491234:
Adds the galleries selector to the image edit panel. 

Resolves (partially) #1003


WithoutPants:
I'm leaving #1003 open for the bulk image assignment use case.

PR #4582, opened 2024-02-17
Update scene, image, and gallery detail panels

QxxxGit:
* Updates the display of metadata on the detail panels
* Removes studio logo from detail panels
* Uses `Show all details` setting to persist expanded details/description

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryDetails/Gallery.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryDetails/Gallery.tsx
@@ -347,17 +347,6 @@ export const GalleryPage: React.FC<IProps> = ({ gallery, add }) => {
       {maybeRenderDeleteDialog()}
       <div className={`gallery-tabs ${collapsed ? "collapsed" : ""}`}>
         <div className="d-none d-xl-block">
-          {gallery.studio && (
-            <h1 className="text-center">
-              <Link to={`/studios/${gallery.studio.id}`}>
-                <img
-                  src={gallery.studio.image_path ?? ""}
-                  alt={`${gallery.studio.name} logo`}
-                  className="studio-logo"
-                />
-              </Link>
-            </h1>
-          )}
           <h3 className="gallery-header">{title}</h3>
         </div>
         {renderTabs()}
--- a/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryDetailPanel.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryDetailPanel.tsx
@@ -1,4 +1,6 @@
-import React from "react";
+import React, { useState } from "react";
+import { Button } from "react-bootstrap";
+import { Icon } from "src/components/Shared/Icon";
 import { Link } from "react-router-dom";
 import { FormattedDate, FormattedMessage, useIntl } from "react-intl";
 import * as GQL from "src/core/generated-graphql";
@@ -9,6 +11,10 @@ import { PerformerCard } from "src/components/Performers/PerformerCard";
 import { RatingSystem } from "src/components/Shared/Rating/RatingSystem";
 import { sortPerformers } from "src/core/performers";
 import { galleryTitle } from "src/core/galleries";
+import {
+  faChevronDown,
+  faChevronUp
+} from "@fortawesome/free-solid-svg-icons";
 
 interface IGalleryDetailProps {
   gallery: GQL.GalleryDataFragment;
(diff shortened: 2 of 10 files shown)


WithoutPants:
Thanks for working on this. In future, I think changes like this would benefit from showing some screenshots so people can provide feedback without having to build the branch.

I think the handling of empty fields needs improving:
image

I think empty fields shouldn't be shown at all. For rating specifically, having the empty rating control displayed infers to the user that it can be interacted with.

I personally think losing the studio image from the detail pages is a step backwards.

The expand details button probably needs a rethink as well.

image

The button area is very large - it would probably benefit from being a minimal variant. As shown above, it's also shown when it doesn't need to be. A `Show more...`/`Show less...` button at the end of the details text, shown only when necessary is probably the better way of doing this imo. When expanding the details, I also noticed that we get some movement in the elements from the scrollbar popping in and out.

Performer cards are now aligned to the left instead of being centered, which I think looks wonky on smaller viewports.

echo6ix:
I've briefly tested this build. I suspect much of my feedback may be regarding incomplete work, but I thought I'd write this out anyway.

### 1. CSS and Layout

The layout displays detail items titles (aka the item heading) and detail item values using h5 and h6 tags respectively. This contributes to an inconsistent look with the other detail pages (performer, tag, studios) that we are trying to emulate.

Why not just re-use the CSS and the structure those pages are using:

```html
    <div class="detail-item {NAME}">
        <span class="detail-item-title {NAME}">...</span>
        <span class="detail-item-value {NAME}">...</span>
    </div>
```

Where `{NAME}` is a placeholder for the title field title of each detail-item. For example,

```html
<div class="detail-group">
    <div class="detail-item studio">
        <span class="detail-item-title studio">...</span>
        <span class="detail-item-value studio">...</span>
    </div>
    <div class="detail-item release-date">
        <span class="detail-item-title release-date">...</span>
        <span class="detail-item-value release-date">...</span>
    </div>
    <div class="detail-item studio-code">
        <span class="detail-item-title studio-code">...</span>
        <span class="detail-item-value studio-code">...</span>
    </div>
</div>
```

Aside from the consistency, reliability, and prospective ease of maintenance by reusing the same CSS and structure, I suspect this will make it a lot easier to add the [order property](https://www.w3schools.com/cssref/css3_pr_order.php) to above CSS. The order property is important because it allows the devs (and users through custom css) to easily change the order of detail items.

```css
detail-item.studio { order: 1; }
detail-item.release-date  { order: 2; }
detail-item.studio-code { order: 3; }
detail-item.tags { order: 4; }
detail-item.updated-date  { order: 5; }
detail-item.created-date  { order: 6; }
detail-item.details { order: 7; }
detail-item.director { order: 8; }
...
```

### 2. What does the date heading mean?
I believe we should take this opportunity to rename the date heading and its corresponding class name to `Release date` and `release-date` respectively. While this item certainly is a date, I think we should specify what the date actually refers to, a [scene] release date. Some people may use it for other purposes, but by and large, all the scrapers are populating a release date.

### 3. Detail items with null values should not be displayed
When detail items have null values they're still being displayed. Per consistency with the detail pages we're imitating, it should not do this.

### 4. Created at and Updated at dates need styling
Those values still don't have the same styling as the other detail items

### 5. Details expand/collapse button
I'm thinking it might be optimal to replace this button with `show more`/`show less` functionality appended at the end of the details text. [See example](https://rubyyagi.com/how-to-truncate-long-text-and-show-read-more-less-button/). Why? When the user reads the details, and when the details are especially long, I suspect the mouse pointer will be at the end of the details text. Rather than having to scroll back up to collapse the details, they can do so with the appended `show less` link.

WithoutPants:
> I believe we should take this opportunity to rename the date heading and its corresponding class name to Release date and release-date respectively. While this item certainly is a date, I think we should specify what the date actually refers to, a [scene] release date. Some people may use it for other purposes, but by and large, all the scrapers are populating a release date.

Hard disagree here. Release date is one interpretation of the date field, but I don't think it's one that should be blanket applied. Users may have studio release, amateur videos or personal content, or a mix of all three and `Release date` is not a good fit for these. In any case, it's out of scope for this PR.

echo6ix:
> Hard disagree here. Release date is one interpretation of the date field, but I don't think it's one that should be blanket applied. Users may have studio release, amateur videos or personal content, or a mix of all three and `Release date` is not a good fit for these. In any case, it's out of scope for this PR.

That's reasonable. I noticed after posting you also expressed concern for displaying a null field and the preference for a show more/show less functionality for details. Does this mean you're in agreement with the rest of my feedback, namely points __1__ and __4__.

WithoutPants:
Yep, though my preference would be to use `<dl>`, `<dt>` and `<dd>` tags per the `SceneFileInfoPanel` (using the existing classnames you mentioned), though that might be better done as a separate blanket change for consistency, and it's not really that important now.

cj12312021:
> I personally think losing the studio image from the detail pages is a step backwards.

I agree with WP about this. The logos are more recognizable than text, so we shouldn't lose that completely.

> 1. CSS and Layout

I agree with all of this.

> 4. Created at and Updated at dates need styling

I actually like Q's approach to those fields. Those fields are really not the same in that they are not editable by the users. I don't think their style should be the same as the editable fields.

I'm not sure I like the rating being presented the same way as the other fields. It currently looks awkward to me and is not handled this way on the other details pages. I haven't thought of a better solution yet.

Issue #4177, opened 2023-10-06
Merged file 'generated files' remain after file is merged and original is deleted


floydcg:
As discussed in bugs on discord, if you merge a file into another, then you delete one of the files from the merge, the generated files remain behind.

My stash instance shows 106,426 files, but the generated folder shows 107,477 generated previews in the folder.

floydcg:
Is there a task or anything to clean up the orphaned files from previous merges before the fix or will it need to be done by hand?

PR #4607, opened 2024-02-21
Add Clean generated files task

WithoutPants:
Adds a `Clean generated files` task, available in the `Maintenance` section of the tasks page.

image

image

Deletes (or logs for a dry run) generated files for which there is no corresponding database entry. In most cases, this task should be unnecessary, but there's certain circumstances that might require it (see #4177).


Maista6969:
Tested this against my main database (with a dry run first) and it worked flawlessly: cleaned out a bunch of files I hadn't even realized were left over from scene merges

PR #4614, opened 2024-02-24
Set UpdatedAt on SceneMarkers after generation tasks & include it in URLs

damontecres:
If any of the SceneMarker related generator tasks create something, update the UpdatedAt timestamp for that SceneMarker.

Additionally, include the UpdatedAt timestamp in the preview & screenshot URLs for the SceneMarker.

I modeled this change after how Scene screenshot generation works and how that URL includes the timestamp.

I didn't see an issue about this, but I'm happy to open one if that's the protocol.

### DIFF
--- a/internal/api/urlbuilders/scene_markers.go
+++ b/internal/api/urlbuilders/scene_markers.go
@@ -7,16 +7,18 @@ import (
 )
 
 type SceneMarkerURLBuilder struct {
-	BaseURL  string
-	SceneID  string
-	MarkerID string
+	BaseURL   string
+	SceneID   string
+	MarkerID  string
+	UpdatedAt string
 }
 
 func NewSceneMarkerURLBuilder(baseURL string, sceneMarker *models.SceneMarker) SceneMarkerURLBuilder {
 	return SceneMarkerURLBuilder{
-		BaseURL:  baseURL,
-		SceneID:  strconv.Itoa(sceneMarker.SceneID),
-		MarkerID: strconv.Itoa(sceneMarker.ID),
+		BaseURL:   baseURL,
+		SceneID:   strconv.Itoa(sceneMarker.SceneID),
+		MarkerID:  strconv.Itoa(sceneMarker.ID),
+		UpdatedAt: strconv.FormatInt(sceneMarker.UpdatedAt.Unix(), 10),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


WithoutPants:
Scene screenshot generation doesn't update the `UpdatedAt` field. The intention of the `UpdatedAt` field is that is it bumped when there is a change to the scene metadata, and not when generating it's files.

This looks like an attempt to fix a caching issue. Is that correct? Maybe if you describe what problem you're trying to solve, I can suggest an alternative approach.

damontecres:
> Scene screenshot generation doesn't update the `UpdatedAt` field

I think I'm misunderstanding this generate task: https://github.com/stashapp/stash/blob/develop/internal/manager/task_generate_screenshot.go#L84-L85

But I guess looking again, that task stores data in the DB versus an image file, so maybe that's why it updates the `UpdatedAt`?

> This looks like an attempt to fix a caching issue. Is that correct?

Yes, exactly. The built-in cache for [Glide](https://github.com/bumptech/glide) basically only uses the URL as the cache-key, so if it fetches a new marker's sceenshot (before generating), it will cache it and never update.

I noticed that all of the other image urls I've been using include a `t` param for the `UpdatedAt` and looked at that task above and thought maybe the markers should do that same.

So I guess, TL;DR: if the `UpdatedAt` is not supposed to change due to generated assets, then probably the right thing to do is switch things internally on my side so the cache uses the `Etag`/`If-None-Match` headers instead.

WithoutPants:
> > Scene screenshot generation doesn't update the `UpdatedAt` field
> 
> I think I'm misunderstanding this generate task: https://github.com/stashapp/stash/blob/develop/internal/manager/task_generate_screenshot.go#L84-L85
> 
> But I guess looking again, that task stores data in the DB versus an image file, so maybe that's why it updates the `UpdatedAt`?

So this is because the cover is considered part of the metadata of the scene, whereas the other generated assets aren't.

> Yes, exactly. The built-in cache for [Glide](https://github.com/bumptech/glide) basically only uses the URL as the cache-key, so if it fetches a new marker's sceenshot (before generating), it will cache it and never update.
> 
> I noticed that all of the other image urls I've been using include a `t` param for the `UpdatedAt` and looked at that task above and thought maybe the markers should do that same.
I'm not even sure if the `t` param is useful on the others, given that we're using `no-cache`, e-tags and the generated files modification times. Does Glide not use these things by default? We could add the `t` param to them, but without also changing the meaning of the `UpdatedAt` field, it wouldn't help you. You don't have any issues with any of the other generated files that use the `t` param?

> So I guess, TL;DR: if the `UpdatedAt` is not supposed to change due to generated assets, then probably the right thing to do is switch things internally on my side so the cache uses the `Etag`/`If-None-Match` headers instead.

If you can do that, yeah that would be the correct approach. I figure that's what most clients would do.

damontecres:
> So this is because the cover is considered part of the metadata of the scene, whereas the other generated assets aren't.

That makes sense. That's probably why I don't have this issue with performer or studio images for example.

> You don't have any issues with any of the other generated files that use the `t` param?

The only generated files I'm using are scene cover, scene preview, scene scrubber sprites, marker screenshot, & marker preview. The previews are always fetched since they are videos. And the scene cover will have the `t` param (and generating updates it).

So just generated scene scrubber sprites & marker screenshots have this cache problem for me I think.

> If you can do that [`etag`], yeah that would be the correct approach. I figure that's what most clients would do.

Yeah makes sense. Thanks so much for talking this through with me! I guess this PR can closed then?

WithoutPants:
No problem at all. Yeah I'll close this. Let me know on here or discord if you run into more caching issues.

Issue #4605, opened 2024-02-21
[Feature] Download image with original filename


vidyava:
**Is your feature request related to a problem? Please describe.**
Any time you go to download an image from Stash, it's presented as "image.jpg" (extension dependent on original filename). This becomes a hassle when you want to download more than one picture without overwriting/renaming every time.

**Describe the solution you'd like**
Either to present the image as the original filename so the user's browser defaults to saving with that name, or just some sort of "Download original" button that preserves the name.

**Describe alternatives you've considered**
In some instances this can be solved by clicking on the "Path" link if you have local access and the paths are the same, but if you're running Stash on a server (without exactly matching mount points) or docker container with non-matched paths, this would still fail.

**Additional context**

Thanks!

PR #4616, opened 2024-02-24
Serve original filename with image

NodudeWasTaken:
This commit addresses https://github.com/stashapp/stash/issues/4605
It simply sets the "Content-Disposition" header which tells the client what the filename is.
Tested in: Firefox, Chrome and Safari

The only real question is "Do we want this?"


vidyava:
I absolutely appreciate your work even if it turns out I'm the only one who wants it

PR #4644, opened 2024-03-02
fix lightbox display modes

### LINKED ISSUE #4643: [Bug Report] Broken lightbox display modes
**Describe the bug**
Since https://github.com/stashapp/stash/pull/4569 was merged, the Lightbox no longer scales images appropriately to fit the screen (see: https://discordapp.com/channels/559159668438728723/644934273459290145/1213285460290314370). The full-screen button, as well as the different display modes, are also broken as a result of the change. Adding the `max-height: 90vh;` property will at least force the image to fit the screen and force the fit-to-screen option. However, this does not work for the other display modes. I don't use image clips, so it's unclear how or if those modes work with clips after that change, but I find it hard to believe they are.

cj12312021:
Unfortunately, I can't test this because my dev environment has suddenly been hit with a bug that prevents the release notes modal from closing. I recall a user encountering a similar issue at some point, but I don't remember how they got around it. However, the idea behind the change is that the code confirms that a valid width and height are provided and calculates the values when necessary.

Fixes #4643 (will need a tester)

### DIFF
--- a/ui/v2.5/src/hooks/Lightbox/LightboxImage.tsx
+++ b/ui/v2.5/src/hooks/Lightbox/LightboxImage.tsx
@@ -98,11 +98,11 @@ export const LightboxImage: React.FC<IProps> = ({
   const [moving, setMoving] = useState(false);
   const [positionX, setPositionX] = useState(0);
   const [positionY, setPositionY] = useState(0);
-  const [imageWidth, setWidth] = useState(width);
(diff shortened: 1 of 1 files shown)


cj12312021:
Rebooting my system seems to have fixed the unclosable release note modal bug I saw earlier, so I was able to test that these changes resolve the display mode issue.

Issue #3681, opened 2023-04-25
[Feature] Monitor DLNA Watch Activity


AdultSun:
**Is your feature request related to a problem? Please describe.**
Currently, watch activity is only monitored within a web browser. Any plays over DLNA is not tracked at all.

**Describe the solution you'd like**
Plays over DLNA should be monitored the same way as plays in the web browser. Stash should update the last played date, play counter, play duration, and progress tracker accordingly.

Even better would be to allow DLNA and web browser tracking to be turned on/off separately. I only ever use the web browser to preview videos and do not want any of that activity to be tracked. I always use DLNA to watch scenes instead. I would want Stash to only monitor watch activity over DLNA and to ignore all watch activity from the browser. However, the current system is the exact opposite of that scenario.

**Describe alternatives you've considered**
A nice stopgap for me would to allow for manually adjusting watch history as needed instead of leaving everything to automatic tracking. I could leave monitoring switched off and just manually input anything I want saved. This would be helpful for other scenarios as well and I've submitted it as a separate feature request (#3680).

feederbox826:
DLNA doesn't have any reporting at all, the only info we could gleam would be http ranges for files which we would have to attempt to reverse in order to find the assumed start play time. Any such feature would also need to be implemented upstream in DMS

PR #4670, opened 2024-03-07
play monitoring improvement

cj12312021:
This pull request updates the play monitoring code to detect video buffering and not count that towards the total play duration.


cj12312021:
My latest push ensures that Track Activity is enabled by default.

AdultSun:
Any chance this PR could also add tracking activity through DLNA? I don't know how difficult that would be.

https://github.com/stashapp/stash/issues/3681

cj12312021:
DLNA doesn't use the front-end VideoJS player, so tracking activity there would likely not be done with the typescript code. So, the code would be completely separate from what we currently have. I have had it in mind to look into support on that side. I just need to spend some time understanding how playback works when using the DLNA service.

AdultSun:
Thanks, I don't mean to pressure you if it's outside of the scope of your PR so no worries if you don't get to it. I just wanted to bring it up in case it was an easy add while you're working in this area.

cj12312021:
No worries. I'm also interested in DLNA support, so you better believe I'd have a pull request out for the DLNA watch activity tracking once I've sorted out the details.

WithoutPants:
> My latest push ensures that Track Activity is enabled by default.

If this changes existing systems, this will require an accompanying release note.

cj12312021:
The update should only affect affect Stash configs. But I'll pull that change into a separate PR to avoid complicating the delivery of the play monitoring improvement.

Issue #2201, opened 2022-01-04
[Feature] Add a thumbnail size slider to the Performers page


CupOnMyTable:
The scenes page has a slider which can be used to increase or decrease the thumbnails size on the page.

It would be good to have this on the Performers page too so that the screen real estate can be easily changed as needed.

Thanks.

philpw99:
Use the custom css to control the size then:
```css
.performer-image-container{
  flex: 0 0 50%;
  max-width: 50%;
}

.col-md-8 {
  flex: 0 0 50%;
  max-width: 50%;
}
```

VonTittyslappen:
I agree that this should just be a normal slider like it is for the scenes section, it makes sense to have it in there

hyde231:
This could be applied for movies as well!

melonyoda:
Please apply this to Markers!

PR #4674, opened 2024-03-09
Add zoom slider to other grid views

cj12312021:
This pull request brings the zoom slider to the movie, performer, and studio pages. It makes sense to save the state of the zoom index for each page, but to minimize conflicts, it probably makes sense to wait until https://github.com/stashapp/stash/pull/4453 is merged.

Closes https://github.com/stashapp/stash/issues/2201


cj12312021:
Hmm, I was under the impression that the zoom level wasn't stored in saved filters, but that does not seem to be the case. I don't think anything else needs to be done here.

WithoutPants:
I can see merit in a zoom slider for movie cards, but is there really a need for zooming of performer and studio cards?

cj12312021:
I remember Echo specifically requested the zoom slider on the performer page so he could fit more of those cards on his screen https://discord.com/channels/559159668438728723/644934273459290145/1204623730895228989. Everything else was just for consistency.

cj12312021:
Of the new additions, I’d probably only use the slider to reduce the size of the movie cards. I’m not against removing the slider on the studio page, but with the sliders on all the other pages, it seems like we might as well include it.

Gykes:
Going to reference the FR here so that it's easier to track.

https://github.com/stashapp/stash/issues/2201

cj12312021:
This branch is up to date with the latest changes in the development branch.

cj12312021:
@WithoutPants, what are your thoughts on closing this PR out on time for the coming release? Sometime after our initial conversation above, you encouraged Gykes to pick up this work, indicating you supported it. Of course, Gykes held off on his change after seeing that this PR existed. However, your continued silence on this despite bringing the PR up on separate occasions leaves me guessing that you do not support this feature.

Without knowing where your thoughts are on this pull request, I'm going to go ahead and make the case for the work under the assumption you've changed your mind about considering this pull request. One of the main appeals of having a slider on any page to adjust card sizes rather than hard-coding a set size in the CSS is to support users who view their content on different monitors with varying sizes and resolutions. For this user, being able to quickly adjust the slider based on the monitor in use, rather than tweaking CSS each time, is invaluable. The current experience of having the slider on some pages and not others yields unproportional cards across pages. A user on a large 4K monitor who blew up the scene cards to the max size would likely have some frustrations that the movie and performer cards remain very small to look at. As long as we have the slider on one of the list pages, I can't see a good reason why we wouldn't include it on the other list pages.

WithoutPants:
I noticed that at certain window sizes (1280 pixels wide), some slider settings wouldn't change the card size. When I maximised I didn't get this issue. It's a minor issue that can be fixed in a separate PR.

PR #4710, opened 2024-03-20
Enable track activity by default

cj12312021:
This pull request sets the track activity setting to true by default on new Stash configurations. New users often ask about automatically updating the ply count without realizing this setting exists.


WithoutPants:
This sets the track activity field to true for existing systems as well, unless the user has enabled/disabled the track activity switch in the past. If we are going ahead with this change, then we _must_ have a release note indicating that the track activity switch will be set to true and the user needs to manually disable it.

WithoutPants:
I've added release notes for this change, and I've changed the setting name from `Track Activity` to `Enable Scene Play History`. I imagined we would get a lot of concerned users about what `Track Activity` means.

cj12312021:
Thanks! That sounds good to me.

Issue #4673, opened 2024-03-09
[Feature] Add data-values attributes to div.scene-specs-overlay span elements; add span.overlay-duration class; report framerate shorthand


echo6ix:
The reported video data within `div.scene-specs-overlay` could still use a bit of polishing, including the addition of displaying overlay framerate appended to the resolution to complete the resolution and framerate shorthand that's now commonly used.

1. Duration still requires its own span element and class, `<span class="overlay-duration">` so it can be properly styled
2. Each span element within `div.scene-specs-overlay` could get a data-value attribute. This will let devs/users manipulate the element based on its value
3. Add framerate shorthand in this section (only two additional characters), where resolution and framerate shorthand is just the framerate (always rounded up to the nearest integer) appended to the resolution with no spaces.

E.g. of resolution and framerate shorthand:

- 1080p60
- 1080p50
- 1080p30
- 1080p25
- 1080p24

Basic hypothetical CSS
```css
.overlay-framerate {
    display: none; /* hide the framerate by default */
}

.overlay-framerate[data-value="50"],
.overlay-framerate[data-value="60"],
.overlay-framerate[data-value="100"],
.overlay-framerate[data-value="120"] {
    display: contents; /* show only high framerates */
}

```

Basic structure of the HTML

```html

<span class="overlay-resolution" data-value="1080p">1080p</span>
<span class="overlay-framerate" data-value="60">60</span>
<span class="overlay-duration" data-value="1:30:22">1:30:22</span>
```

![image](https://github.com/stashapp/stash/assets/37937507/293ddb22-31e8-4ebb-a758-81039eadfcbd)

PR #4714, opened 2024-03-21
[RFC/WIP] Details operation toolbar

WithoutPants:
This is an offshoot of #4699. I'm taking my own advice and reducing the scope of changes so that more focused feedback can be provided and get this implemented sooner.

The summary of changes is as follows:
* date, frame rate and resolution are moved to a sub-header under the main title
* a toolbar has been added between the date/resolution sub-header and the tabs. This toolbar includes the rating, a new play count button, the o-count button, organised button and the overflow menu.
* rating has been removed from the edit panel. Rating can be set in any panel now. Keybinds have been shifted across as well
* o-count button has been restyled. The dropdown menu is removed - the other functions can be accessed in the history tab. The play count works the same way. 
* the numeric rating control is displayed as a read-only field by default, the editable text field is accessed by clicking the accompanying pencil button, which shows and focuses on the field. Clicking away from the field restores the read-only field.

image
image
image

I plan to add to this PR to roll this out to the image and gallery detail pages, once I've gotten some feedback.

Image will have the date and resolution in the sub-header, and all but the view count button in the toolbar. Gallery will have the date and potentially image count in the sub-header, and the same toolbar buttons as image.

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneDetails/Scene.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneDetails/Scene.tsx
@@ -7,7 +7,7 @@ import React, {
   useRef,
   useLayoutEffect,
 } from "react";
-import { FormattedMessage, useIntl } from "react-intl";
+import { FormattedDate, FormattedMessage, useIntl } from "react-intl";
 import { Link, RouteComponentProps } from "react-router-dom";
 import { Helmet } from "react-helmet";
 import * as GQL from "src/core/generated-graphql";
@@ -74,6 +74,8 @@ const SceneVideoFilterPanel = lazyComponent(
   () => import("./SceneVideoFilterPanel")
 );
 import { objectPath, objectTitle } from "src/core/files";
+import { RatingSystem } from "src/components/Shared/Rating/RatingSystem";
+import TextUtils from "src/utils/text";
 
 interface IProps {
   scene: GQL.SceneDataFragment;
@@ -420,27 +422,6 @@ const ScenePage: React.FC<IProps> = ({
               <FormattedMessage id="actions.edit" />
             </Nav.Link>
           </Nav.Item>
-          <ButtonGroup className="ml-auto">
-            <Nav.Item className="ml-auto">
-              <ExternalPlayerButton scene={scene} />
-            </Nav.Item>
-            <Nav.Item className="ml-auto">
-              <OCounterButton
-                value={scene.o_counter || 0}
-                onIncrement={onIncrementClick}
-                onDecrement={onDecrementClick}
-                onReset={onResetClick}
-              />
-            </Nav.Item>
-            <Nav.Item>
-              <OrganizedButton
-                loading={organizedLoading}
-                organized={scene.organized}
-                onClick={onOrganizedClick}
-              />
-            </Nav.Item>
-            <Nav.Item>{renderOperations()}</Nav.Item>
-          </ButtonGroup>
         </Nav>
       </div>
 
@@ -509,6 +490,11 @@ const ScenePage: React.FC<IProps> = ({
 
   const title = objectTitle(scene);
 
+  const file = useMemo(
+    () => (scene.files.length > 0 ? scene.files[0] : undefined),
+    [scene]
+  );
+
   return (
     <>
       <Helmet>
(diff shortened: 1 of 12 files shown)


DogmaDragon:
How about also adding an icon next to decimal rating?
image
Used https://fontawesome.com/icons/star?f=classic&s=solid in the example. 

---

> This is consistent with the current date display in the details pages. The short form is suitable for cards and input, but for details presentation, I think we should stick with the locale-specific longform format.

That does make sense. I'll just have to learn to live with it.

WithoutPants:
> How about also adding an icon next to decimal rating? 

Good idea. I went with gold colouring and added this to other components as well:
image

(image lightbox)
image

(gallery wall)
image

I also improved the handling of changing the numeric rating so that it doesn't set on every keypress - now updates on blur instead.

randemgame:
Looks good to me! I didn't realise decimal allowed for even more specific grading so have switched over from tenth stars. I was put off by the aesthetic before. In Firefox the input box is way too large, see:
Screenshot 2024-03-21 142250
but I noticed in the screenshot above that it isn't true for other browsers like Chrome
Screenshot 2024-03-21 142302. Not sure that you can do anything about that.

Could you fix it so the decimal input for the Performer page would have this same new look you've done with scenes, with the star and pen icon to input, rather than it permanently being in the input box mode? This was part of the off putting look before for me.
Screenshot 2024-03-21 142500

Also, just noticed in both browsers that the decimal input box is filling the whole width in movies, which looks pretty messed up! From Chrome:
Screenshot 2024-03-21 142823

WithoutPants:
> Looks good to me! I didn't realise decimal allowed for even more specific grading so have switched over from tenth stars. I was put off by the aesthetic before. In Firefox the input box is way too large... Not sure that you can do anything about that.
> 
> Could you fix it so the decimal input for the Performer page would have this same new look you've done with scenes, with the star and pen icon to input, rather than it permanently being in the input box mode?
>
> Also, just noticed in both browsers that the decimal input box is filling the whole width in movies, which looks pretty messed up!

Thanks for the feedback, this has now been fixed and made consistent across the other pages.

WithoutPants:
These changes have been added to the image and gallery details pages. 

I've made the numeric rating changes to the list table cells as well:
image
image

To reduce the amount of vertical space used up by the header, I've used `TruncatedText` for the object titles, restricting the rows to 2. For most titles found on stashdb, this accommodated for most titles. This can be tweaked later if needed.

This is now ready for review/testing.

echo6ix:
To me this is a regression from the incumbent layout. It doesn't reduce inconsistency and ad hoc design elements, it adds to it. 

#### Row with date, resolution, and fps
There's no reason the resolution and fps has to be justified right and cannot be grouped with the date using a pipe or other bullet separator. The pipe separator is already introduced btw fps and resolution; the fact that two of values are more alike than the date is a trivial difference that won't confuse anyone. Moreover, the resolution and fps are redundant here since they also appear in the file tab, therefore display them using the condensed shorthand format (ie. 1080p30, 2160p60, no decimals for the fps on shorthand). The fps and resolution seem to be acting like a summary report to the file tab info. This reduces the amount of characters and required bullet separators: `January 1, 1973 • 2160p60`

#### Row with ratings and operation bar
The rating stars don't need to be partitioned from the rest of the operation sbar. Rating stars, watch icon, and o-counter icon are all actionable user interactions (things that are alike), and you're already willing to put them on the same row. Therefore ratings should be grouped with the operations bar and left justified too. The footer on Github cards shows how to do this right where each button has a subtle border to visually partition it.

image
*Github user interaction buttons*

I would personally implement it like this for the different rating star types:

image

The `pencil` edit icon is superfluous like this because the borders around the buttons already indicate the button is actionable, and that should be enough space to edit a numerical value in place. It makes sense to have the `pencil` edit icon in a table's cell because that's a completely different layout altogether.

Each operation button is now the same. Consistent and grouped together for visual coherency.

image

JanJastrow:
I like the redesign, even tho it looks a bit crammed (just judging from the screenshots).
I'd like to throw Issue https://github.com/stashapp/stash/issues/1744 in here while you are working on this page.

- Save ratings without the additional need to click "save"
(or is this the case now?)

WithoutPants:
> There's no reason the resolution and fps has to be justified right and cannot be grouped with the date using a pipe or other bullet separator.

Apart from making it somewhat consistent with the scene cards, these positions provide a definite place for the user to find this information. Not all scenes/images/galleries will have dates, so the position of the resolution/frame rate will differ based on whether the date is set. 

image
image

I don't think this is an improvement over what's currently on this branch.

Condensing and combining the frame rate and resolution values to my eyes makes it more difficult to parse, and I just can't see a good reason to condense it this way. It's not like there's a dearth of space on that row.

> The rating stars don't need to be partitioned from the rest of the operation sbar. Rating stars, watch icon, and o-counter icon are all actionable user interactions (things that are alike), and you're already willing to put them on the same row. Therefore ratings should be grouped with the operations bar and left justified too.

This seems like a pretty arbitrary approach and I'm not sure how this improves usability. Partitioning the rating from the other buttons means that there's fewer buttons grouped together, making it easier for the user to find the button/information they are after. Apart from looking ugly and unbalanced to my eyes, I just don't see it improving usability either:

image

> The footer on Github cards shows how to do this right where each button has a subtle border to visually partition it.

This is potentially a good idea as it makes it clearer to users that these buttons are interactable, _but_ right now it's is inconsistent with the entirety of the rest of the UI. If the goal is to minimise inconsistency, this directly contradicts it. We should reconsider the `minimal` button variant design across the UI though.

@JanJastrow 

> I'd like to throw Issue https://github.com/stashapp/stash/issues/1744 in here while you are working on this page.

This does in fact resolve that issue.

randemgame:
I agree with the man without pants and have been running this branch for a good week now and would love to see it in the main dev build while discussion continues around the greater design of things. I thought I could try knock up a few different watch icon designs, I was thinking it would be cooler if looking towards the scene video to the right rather than left, or maybe straight forward is simpler to parse (though I'd be a little worried I was being watched) 👁️

echo6ix:
@WithoutPants whatever approach you decide, if the time investment isn't a lot, can you add `data-value` attributes to the html for things like the date, resolution, and fps. It helps a lot with the potential customization, see here where I include a hypothetical example https://github.com/stashapp/stash/issues/4673

JanJastrow:
I like it.
I need to adjust my muscle memory, because I was rating my large library at the moment. ¯\_(ツ)_/¯
The stars are a bit to small / hard do use.
On the phone as well as on desktop with a mouse.

echo6ix:
Someone on Discord was mentioning they can no longer style the counter elements.  Adding a `data-value` property to the o-counter and play counter respectively would resolve this. Probably would be good practice going forward to add that property to all elements with badges and counters etc

IAmKontrast:
Hide framerate and resolution from the new layout with css:

/* Hide Framerate and resolution */
.scene-subheader .frame-rate,
.scene-subheader .divider,
.scene-subheader .resolution {
  display: none;
}

PR #4733, opened 2024-03-29
Replace queue tab with discover tab which include other related content

cj12312021:
The pull request improves the queue tab (now named discover), which will allow users to view other scenes from the studio or performers within the player page. This work also added support for video previews when hovering over scenes within the tab. I uploaded a video of this work here https://discordapp.com/channels/559159668438728723/644934273459290145/1223103538893164708.


ghfsa:
Can this be put below the player? Then we can have more real estate & hence larger thumbnails & more related content in grid view?
Horizontally placed buttons/tabs like  'Same Studio', 'Same Performer' etc placed just above that grid for filtering would be awesome

cj12312021:
This pull request isn’t meant to do any significant restructuring of the player page. That task will happen in https://github.com/stashapp/stash/pull/4714#issuecomment-2011051571 so that feedback will be better served there. But I will say that moving this tab below the video container will interfere with the advantage we have of being able to browse these options while keeping the video on screen, which imo would not be worth it. Browsing on this page is meant to be a secondary task and shouldn't get in the way of the primary task of viewing the video.

cj12312021:
I can update the queue thumbnails to take up 40% of the available horizontal width. That should help improve the thumbnail size.

PR #4765, opened 2024-04-16
Full hardware transcoding

NodudeWasTaken:
This adds a "test" that checks if full hardware transcoding is possible for a given file.
The reason its done this way is because there are a gazillion different incompatibilities that i dont want to check for.

This is naturally optional and has a setting in the menu.
I've also added an option to add hardware decoding in any case (-hwaccel auto).

Also contains a fix for that one guy that had an AMD APU that needed atleast 256px.

My tests from 7k60 -> 4k60 using an NVIDIA 10 series (using h264_nvenc for all):
No accelerated decode: 0.7x
Accelerated decode: 0.17x (yes my cpu/ram is slow)
Full hardware transcode: 1.3x

Accelerated decode seems to be a footgun in my system, but i dont know that that's true for all systems.

This is explicitly only tested for NVidia and Intel, i dont know that vaapi works properly.
As a side note it also adds an ffmpeg version check, because i needed it for nvenc 10bit transcoding.


WithoutPants:
It's a bit unclear to me what the impacts of these new options are, and the naming is a bit unintuitive.

Is there a reason for the naming `alwaysAddHardwareDecoding`? Isn't `hardwareDecoding` sufficient?

> This is naturally optional and has a setting in the menu.

Why is this optional? When would a user use the various combinations? I'm assuming `transcodeHardwareAcceleration = false` implies that the `transcodeFullHardwareAcceleration` option is also false. What is the impact of both values being true, and why would a user _want_ to not use full hardware acceleration?

Your performance test results aren't really clear to me.

NodudeWasTaken:
> It's a bit unclear to me what the impacts of these new options are, and the naming is a bit unintuitive.
> 
> Is there a reason for the naming `alwaysAddHardwareDecoding`? Isn't `hardwareDecoding` sufficient?
> 
> > This is naturally optional and has a setting in the menu.
> 
> Why is this optional? When would a user use the various combinations? I'm assuming `transcodeHardwareAcceleration = false` implies that the `transcodeFullHardwareAcceleration` option is also false. What is the impact of both values being true, and why would a user _want_ to not use full hardware acceleration?
> 
> Your performance test results aren't really clear to me.
Alright i removed the options such that it always tries full hardware transcoding if hardware acceleration is enabled.
You're right that it was unnecessary, the only real downside is increased vram usage, but that should be negligible.
Always add decode is also gone, the only reason its there is because i saw many users using it, but it conflicts with full hardware transcoding.

feederbox826:
Just a footnote for future - This will cause problems for containerized users as `scale_cuda` on most ffmpeg builds will require the CUDA drivers and subsequently NVIDIA drivers on the host

PR #4766, opened 2024-04-17
chore: fix struct names

hidewrong:


HereIam80:
The main dev is taking a few days off at the moment.
I guess right now you can't do anything but wait.

Thanks for the contribution.

PR #4770, opened 2024-04-18
Select first result in Tagger when selectedResult is not defined

### LINKED ISSUE #4734: [Bug Report] Extra click necessary when using tagger
**Describe the bug**
When using tagger and matches are found, normally the first one / best match is "pre selected". With the last Update the behaviour changed. Now even the only match needs to be clicked to be shown and be able to save it.

**To Reproduce**
1. Use tagger
2. Scrape
3. Get results

**Expected behavior**
Should work like before. First/best match should be loaded directly without interaction.

**Screenshots**
grafik
grafik

v0.25.1-14-g034fd440

 - OS: Windows 11
 - Browser Chrome
 - Version 123.0.6312.60 

**Additional context**
Probably linked to https://github.com/stashapp/stash/pull/4715

Emilo2:
Fixes #4734


hrxcodes (approved):

Maista6969:
Tested this and it works as expected 👍

PR #4801, opened 2024-05-02
Clean duplicate aliases when creating from performer tagger

### LINKED ISSUE #4596: [Bug Report] Error when scraping performer with duplicate alias
**Describe the bug**
A user on Discord [reported an issue](https://discord.com/channels/559159668438728723/559159931505213440/1209234970896236654) when scraping a performer from PMVStash and getting an error message about duplicate aliases. After going through their performers and aliases to make sure they didn't have a local duplicate we narrowed it down to that particular performer having duplicate aliases on the stash-box itself: `MyCherryCrush` and `Mycherrycrush`

I do not have an account on PMVStash and I wasn't able to find a performer on StashDB with a duplicate alias, but over on FansDB [Brandi Braids](https://fansdb.cc/performers/4d02fc23-75ca-4c8a-804c-b492b2a8d808) currently has the same kind of duplicate alias: `BrandiBraids` and `brandibraids`

edit: Brandi Braids has had her duplicate alias removed

**To Reproduce**
I believe the original issue occurred when scraping a performer as part of a scene, but it also occurs when you try to add the performer directly:

1. Go to Performers page
2. Switch to Tagger view and configure it to use FansDB as the source
3. Click "Batch Add Performers"
4. Search for "Brandi Braids" and select the only performer that comes up

No performer is created and no error is displayed, but the logs output `Failed to create performer Brandi Braids: performer contains duplicate alias 'brandibraids'`

**Expected behavior**
Just like creating a performer who has an alias that's the same as their own name, I expect the duplicate alias to be dropped and the performer to be created. This was fixed in #4443 

v0.24.3-46-g6fb1c41a

WithoutPants:
Fixes #4596

### DIFF
--- a/ui/v2.5/src/components/Tagger/performers/StashSearchResult.tsx
+++ b/ui/v2.5/src/components/Tagger/performers/StashSearchResult.tsx
@@ -22,7 +22,10 @@ interface IStashSearchResultProps {
 function cleanAliases(currentName: string, aliases: string[]) {
   const ret: string[] = [];
   aliases.forEach((alias) => {
-    if (alias !== currentName && !ret.includes(alias)) {
+    if (
+      alias.toLowerCase() !== currentName.toLowerCase() &&
+      !ret.find((r) => r.toLowerCase() === alias.toLowerCase())
+    ) {
       ret.push(alias);
     }
   });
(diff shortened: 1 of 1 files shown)


Maista6969:
Could you make this case insenstive? Stash-boxes allow Performer A to have both "Alias A" and "alias A" but local Stash will not

Maista6969:
Actually since performers can be created through the Tagger view (original bug used Batch Add Performers as an example) this might need to be done on the back end

PR #4822, opened 2024-05-08
Update to Go 1.22

its-josh4:
Updates to Go 1.22 because 1.19 is un-supported and has some CVEs.

Also updates a small number of low-risk deps


its-josh4:
Looks like this requires a new builder image to be pushed to Docker Hub.

Darthagnon:
Would it still work with Go v1.21.4? Go v1.22 breaks Windows 7, 8, 8.1 compatibility, and I don't really trust it to keep compatibility with older releases of Win10.

Issue #4659, opened 2024-03-06
[Feature] Add last_played_at to filters


indigocalifornia:
**Is your feature request related to a problem? Please describe.**
Feature request

**Describe the solution you'd like**
To be able to filter scenes which are have been played last later/earlier/etc compared to value

**Describe alternatives you've considered**
Sorting by Last played at

**Additional context**
Adding last_played_at to filters would allow soring by a different metric, e.g. Play count

WithoutPants:
Closed by #4829

PR #4829, opened 2024-05-11
Add last_played_at filter

NotForMyCV:
Adds the ability to filter scenes by their most recent view, addresses https://github.com/stashapp/stash/issues/4659


WithoutPants (approved):
I'll merge this as-is, but I'm not certain that the null behaviour is correct.

In other cases, we would include explicit is null/not null modifiers, and `last played at < date` would exclude scenes that have not been played. I'm undecided on what mechanism is more intuitive/has more utility.

Issue #4790, opened 2024-04-27
[Feature] Add "disambiguating" select values to scene, gallery, and movie select dropdowns


echo6ix:
**Is your feature request related to a problem? Please describe.**
Nothing precludes duplicate titles for scene, gallery, or movie objects. Therefore, dropdown select values of each may contain duplicate string values that are impossible to differentiate from one another.

_Example of official duplicate gallery titles from ATK_
![image](https://github.com/stashapp/stash/assets/37937507/088563eb-b2c1-4b39-9c9e-79f8e0a4c2d9)

This is more prevalent among particular studios that heavily recycle titles, but regardless, given enough content, just as with performer names, duplicates emerge.

**Describe the solution you'd like**
Display complimentary metadata to disambiguate each select title value in the drop down.

- Date
- Studio code
- Cover image, just as implemented in #4227 

A tiny thumbnail of a cover image may seem irrelevant due to its size, but consider small distinctions in colors and shapes would be enough to disambiguate it from other cover images all other values being equal.

![image](https://github.com/stashapp/stash/assets/37937507/3fa1984b-f804-4b92-ba87-54ce54e40d60)

Styling should consider the following:
- Keep height of select consistent with the performer select (#4227)
- Title select value should take precedence (ie. white color)
- Complimentary select values are secondary, and can be conveniently presented below title value with a muted color

For example, something like:

```css
.select-options {

     /* Set default styling for all span elements */
     span {
          color: white;
          display: block;
     }

     span.select-option-title {
     }

     /* Mute text of complimentary metadata span elements */
     span.select-option-date,
     span.select-option-studio_code {
          opacity: 0.7;
     }

}
```

**Additional context**
I should have included this broad request in my initial request https://github.com/stashapp/stash/issues/3076 (_to add performer images to the performer select_) implemented here https://github.com/stashapp/stash/pull/4227, since it falls under the same conceptual umbrella.

Side note: if studio objects ever have the ability to either use non-unique names or can have favicons, this concept should be applied to them as well since duplicate studio names do exist (outside of Stash).

```html
<select-options>
  <div class="primary line">
        <img>$favicon</img> <span>$studio_name</span>
  </div>
  <div class="secondary line">
        <span>$parent_studios</span>
  </div>
</select-options>
```

echo6ix:
> Correct for scenes and galleries, but that's not true for movies... I get a "Unique constraint failed" SQL error when attempting to create a Movie with the same title as an existing one.

[...]

PR #4832, opened 2024-05-13
New scene select with additional fields

dogwithakeyboard:
Adds a new scene select similar to the other recently updated selectors - this is effectively a copy + paste of the new gallery select. The new matching works much better at returning the scene whereas the existing selector frequently finds nothing when the query returns too many items. 

Taking some inspiration from #4790 the results include the thumbnail, title, studio, date and code (up to three lines) to help distinguish between identically named scenes.

swars

The path matching remains and matched file paths are shown below.

ss1


WithoutPants (approved):
This is fantastic. Thanks for this contribution. A future iteration would be to show a popover for a scene when hovering over the current scene value.

PR #4835, opened 2024-05-14
update popover delay

cj12312021:
This pull request adjusts the card popover entry and exit delay. Previously, the popovers were set to delay exit by 400 milliseconds, making hovering over individual popovers feel clumsy as the delayed exit often gets in the way of attempting to hover over different popover items. Interestingly, when adjusting the values to remove the existing delay, I noticed the UI is more responsive with other transitions I have applied to the cards in my CSS.


cj12312021:
Further testing revealed that the delay was there to ensure the popover doesn't disappear when users attempt to navigate a popover item. So, my latest push adjusts the exit delay to match the entry delay to ensure that the 2 don't collide with each other.

PR #4837, opened 2024-05-14
Add all available bulk edit fields to edit dialogs

### LINKED ISSUE #1408: [Feature] Enhance bulk editing of Performers and Scenes to edit every allowed field

### LINKED ISSUE #4823: [Feature] Add full list of editable fields in the edit modal for images
Many fields are missing from the edit modal for images

Add the full list of editable fields to the edit modal for their respective objects:
- Image edit modal `*`
  - url
  - studio code
  - date
  - photographer

`*` When bulk editing, many images will share these above values, particularly when they belong to a gallery. It's currently not feasible to edit each image manually for this use case. This would be an easier way to facilitate that task.

bob123491234:
Adds a toggle switch to the Scenes, Galleries, Images, and Performers edit dialog boxes that, when enabled, shows all fields present in the respective bulk update mutation. A hotkey 'q' was added that opens the dialog box with all fields visible.
Fields shown in the default view of the performer edit dialog reduced to Rating, Favorite, Gender, Country, Tags, Ignore Auto Tag. 

Minor issues still present: 
- Background of URLs, Aliases text boxes doesn't match the other text boxes
- No check is performed on URLs, Aliases for uniqueness
- No check is performed on dates for correct format (error popups when attempting to save incorrect date)

Resolves #1408
Resolves #4823


Gykes:
Superseded by https://github.com/stashapp/stash/pull/6647

PR #4842, opened 2024-05-15
Return 401 code for `ErrUnauthorized` errors

damontecres:
A very tiny PR so that an `401 Unauthorized` is returned when the request is unauthorized instead of a `500 Internal Server Error`.

This was the behavior before an [error lint PR reversed the if condition](https://github.com/stashapp/stash/pull/1796/files#diff-2591b9ac79fca516f3e900bddd650c09d147dba12bac4c002d6eec976216621aR43) probably accidentally.


WithoutPants (approved):
Good catch, thanks.

PR #4845, opened 2024-05-16
Replace viper with koanf

### LINKED ISSUE #4575: [Bug Report] settings doesn't appear when plugin id is in upper camel case
Plugin IDs should be lower camel cased otherwise Stash will not display your plugin setting values on the plugins page.

To demonstrate the issue, you can try creating a plugin with an upper camelcased name `TestPlugin.yml` with the following config
```name: Test Plugin
description: My plugin does awesome things
version: 1.0
settings:
  test:
    displayName: Test
    type: STRING
```

If you load this plugin and give the setting a value, you'll see it:
image
The issue appears when you refresh the page. The test value disappears:
image
However, the value is still in the stash `config.yml`, so it still exists.

So if you rename the plugin id to `testPlugin.yml` and reload it, then the value comes back:
image

OP: https://github.com/Tetrax-10/stash-plugin-builder/pull/1

### LINKED ISSUE #4740: [Bug Report] configurePlugin mutation unloads plugin settings
**Describe the bug**
When using the configurePlugin mutation to update a plugin's settings config, all other plugins have their settings unloaded unless Stash is restarted.

**To Reproduce**
1. Create two plugins:
`test1.yml`
```
name: Test 1
description: Test 1
version: 0.0.0
settings:
  a:
    displayName: A
    type: STRING
```
`test2.yml`
```
name: Test 2
description: Test 2
version: 0.0.0
settings:
  b:
    displayName: B
    type: STRING
```
   
2. Load the plugins and give the plugin settings values.
image

3. Restart stash **The bug doesn't reproduce without a restart**
4. Execute the following GQL query in the playground:
```
mutation ConfigurePlugin($plugin_id: ID!, $input: Map!) {
                        configurePlugin(plugin_id: $plugin_id, input: $input)
                      }
```
variables:
```
{
  "plugin_id": "test1",
  "input": {
    "a": "x"
  }
}
```
5. Refresh the plugin settings page and observe that the test2 "B" setting no longer has a value
image
However, the value is still in the `config.yml`. It just seems like Stash won't reload the setting until it's restarted
image
6. Run the following GQL query:
```
query Configuration {
                        configuration {
                          plugins
                        }
                      }
```
7. See that the test2 plugin with the "B" setting is no longer returned:
```
{
  "data": {
    "configuration": {
      "plugins": {
        "test1": {
          "a": "x"
        }
      }
    }
  }
}
```
If you restart stash and rerun the query, the test2 plugin settings will be returned again:
```
{
  "data": {
    "configuration": {
      "plugins": {
        "test1": {
          "a": "x"
        },
        "test2": {
          "b": "b"
        }
      }
    }
  }
}
```

**Expected behavior**
configurePlugin shouldn't affect the settings of plugins that aren't the given plugin_id. All plugin settings values should appear in the UI and be returned when querying the Stash configuration for plugins.

v0.25.1
Build hash:
bf7cb78d
Build time:
2024-03-13 03:30:08

I first noticed the issue on the 0.24.3 build but have confirmed it happening on v0.25.1 as well
v0.24.3
Build hash:
aeb68a58
Build time:
2024-01-15 00:32:38

 - OS: Win10
 - Browser: Brave

**Additional context**
As noted in step 3 of the reproduction steps, a restart is required. If stash isn't restarted, then the other plugin settings are retained as expected and will get returned by the Configuration query

WithoutPants:
Viper has long been a pain to deal with, particularly around its handling of camelCase field keys. Investigation of #4740 also determined that viper was the root cause. 

This PR replaces `viper` with `koanf`. koanf handles camel case fields correctly. This means that all of the camel case to snake case conversion hacks can be removed.

Since the field keys are changed, a (non-schema) migration is required to convert existing config files. The migration backs up the existing config file in the same directory to `config.yml.57.<datetime>`. The migrated config file will not be compatible with older stash versions.

Fixes #4740 
Fixes #4575


internal/manager/config/init.go
```diff
@@ -20,7 +22,28 @@ type flagStruct struct {
 	nobrowser      bool
 }
 
-var flags flagStruct
+var (
+	flags flagStruct
+
+	defaultConfigLocations = []string{
+		"config.yml",
+		filepath.Join("$HOME", ".stash", "config.yml"),
```
Maista6969, line 30: This won't work on Windows

Maista6969 (changes requested):
Tested this on my Windows system and I could not find any issues after migrating my config: tried toggling some settings, generating new API keys, changing some filters around, etc. and everything works fine 👍

internal/manager/config/init.go
WithoutPants: Good catch. `$HOME` is handled by viper but it doesn't look like `koanf` does.

PR #4857, opened 2024-05-20
Create missing items by default in Identify

bob123491234:
Changes the default behavior so that studios/performers/tags that are missing are created when using Identify.


WithoutPants:
If we can avoid it, we don't want to change the behaviour of backend because this will cause unexpected behaviour for other clients/plugins/scripts that use the identify interface.

WithoutPants:
Closing in favour of #4873

PR #4865, opened 2024-05-22
Enforce whitelist for sort values (CVE-2024-32231)

WithoutPants:
Fixes vulnerability to SQL injection when entering specific values into the `sort` URL parameter. Changes the behaviour so that sort strings are validated against a white list of values and returns an error if it is not a valid value.

Fixes CVE-2024-32231


Cvrt19:
Would it be possible to just add all of the possible filtering options to the list of allowed sorts? I had been using "career_length" to unofficially sort Performers by debut (by putting sort=career_length manually in the address bar) and it was working great, but this fix breaks that method. There may also be people who want to sort by "weight" or maybe even by "country."

PR #4873, opened 2024-05-23
Default create missing to true in Identify

WithoutPants:
Defaults create missing studios/performers/tags to true in the UI.

Supercedes #4857

### DIFF
--- a/ui/v2.5/src/components/Dialogs/IdentifyDialog/IdentifyDialog.tsx
+++ b/ui/v2.5/src/components/Dialogs/IdentifyDialog/IdentifyDialog.tsx
@@ -46,6 +46,21 @@ export const IdentifyDialog: React.FC<IIdentifyDialogProps> = ({
           field: "title",
           strategy: GQL.IdentifyFieldStrategy.Overwrite,
         },
+        {
+          field: "studio",
+          strategy: GQL.IdentifyFieldStrategy.Merge,
+          createMissing: true,
+        },
+        {
+          field: "performers",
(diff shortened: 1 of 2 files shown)


DogmaDragon:
I would suggest adding a warning or suggesting making a backup somewhere on the identify interface modal, since the task is irreversible.

WithoutPants:
> I would suggest adding a warning or suggesting making a backup somewhere on the identify interface modal, since the task is irreversible.

Is this a blocker to merging this change? I don't have a simple solution for adding such a warning currently, so I'll defer this change to the next release if we think it's destructive without accompanying warnings/documentation.

Ronnie711:
I've been deep diving with Identify over the past few days & have nearly finished with it. As I said on Discord @DogmaDragon I'm going to write up a sort of 'How to' and explain all the options. I do agree it could do with a disclaimer but I wouldn't call it a blocker at this stage as this change seems to be adding basic functionality which doesn't break how it's currently working.

I had Identify running with the current defaults and it half identified because I'd forgotten to set these changes from the current default (which seems to be a common complaint on discord).

I'll try and put down some thoughts on what could be potentially improved over the next few days and open an issue for everyone to expand on :)

DogmaDragon:
> Is this a blocker to merging this change? I don't have a simple solution for adding such a warning currently, so I'll defer this change to the next release if we think it's destructive without accompanying warnings/documentation.

I wouldn't say it's a blocker. Some people tend to run the task not fully understanding what it does and then come to ask if they can reverse it somehow. It would just be a big quality of life for both users and support people.

PR #4923, opened 2024-06-04
Add docker labels to have update tools be able to pull changelog/release notes

halkeye:
For example https://docs.renovatebot.com/modules/datasource/docker/

but other tools will pull those same labels


halkeye:
Apparently it pulls the difference between "latest" and the latest tag? not 100% sure

```
$ regctl image config --format '{{ jsonPretty .Config.Labels }}' stashapp/stash
null
```

WithoutPants:
I believe this is incomplete. The `build` dockerfiles are only used when building manually. The `ci` dockerfile is used by GitHub actions to deploy to docker hub. The build github actions file will also need to be updated to set the appropriate environment variable.

halkeye:
oh wow I can't believe i missed the build in https://github.com/stashapp/stash/blob/develop/docker/ci/x86_64/docker_push.sh
I'll get that second dockerfile updated

halkeye:
is there any easy way for me to confirm this ahead of time? I kinda did it blind

feederbox826:
Looks like this broke CI

PR #4927, opened 2024-06-04
added support for hooking into detail page item dropdowns

keenbed:


WithoutPants:
I've retooled the Gallery page one. I can do the others when I have more time. It's a bit of an ask for a seven-month submission, but if you can do the others I can fit it into this release.

PR #4928, opened 2024-06-04
Added Blur images via keyboard shortcut support

keenbed:

Issue #4510, opened 2024-01-27
[Feature] UI Plugin API Improvements


QxxxGit:
Per WP's suggestion, this is an issue for plugin developers to request improvements to the UI plugin API. This single issue would be used to keep requests in one place.

My current requests:

* Date picker support
* Country dropdown select (display countries with flags)
* Tag dropdown select (includes tag previews and pill badges)

ralle947:
We need functionality to move and create elements to all parts of all pages so that those who work on making themes and scripts can fulfill their creative ideas.

7dJx1qP:
* A way for plugins to provide custom mutations and subscriptions that can be executed and subscribed to by the UI. I currently work around this using tasks but I have to hide the ones that require arguments from the tasks page and I also need to poll the logs to be able to receive any output and have the UI act on it.

* A "multiple" plugin config setting property in addition to the "description" and "type" setting properties. A multiple value of true would indicate that the setting is a list of values rather than a single value. I currently work around this by allowing for comma-separated strings for multiple values, but it would be nice if the UI showed a list input for these kinds of settings.

* A "visible" plugin config setting property to control whether a setting is shown in the UI. This can currently be worked around by hiding the element through the UI script.

* Additional plugin config setting types like date, country, tag, studio, performer, tag, etc. that would use those specialized input components OR allow config settings to be hooked so the UI plugin can replace the text inputs with the specialized input components.

QxxxGit:
Another request: Scene page as a patched component, with access to some of the current components on that page. I think the essentials I would need are the video player and... maybe the queue? I think if I had those I'd have a good amount to work with. Extra tooling like the tab/pane functionality could be useful, but lower priority.

Ceequester:
Text blocks as a type to allow for users to create more complex configurations for things like pathParser without needing to edit the plugin code manually which should hopefully be on the way out with a built-in installer.

QxxxGit:
Can I get a confirmation: when I register a route, that route can only be viewed when navigating from an existing one. I cannot view the page by visiting `localhost:9999/plugin/test-route` on its own. Doing that or refreshing the page throws a 404.

Valkyr-JS:
Thanks for adding the `PerformerDetailsPanel` components, they're working nicely. Could I request the `TagCard` component as well, if that's okay?

Thanks

Valkyr-JS:
[...]

PR #4935, opened 2024-06-06
Plugin api improvements

### LINKED ISSUE #4880: [Feature] Add PerformerDetailsPanel to PluginAPI.components
I've been working on a plugin and I believe I need a component to be added to the list of available components that can be patched via the plugin API. I've attached a screenshot of where I'm looking to add my own UI component. I believe it's in between `PerformerDetailsPanel` and `DetailsEditNavbar`.

I'm not looking to replace the existing details panel, rather add a new component directly beneath it. My understanding is that will happen with using `PluginApi.patch.after("PerformerDetailsPanel", ...)` once the `PerformerDetailsPanel` component has been added to the list. If so, can the `PerformerDetailsPanel` component be added?

Apologies if I've misunderstood how this should work!

placement

WithoutPants:
Related #4510 
Resolves #4880
Obsoletes #4928 with a plugin.

Makes the `App` container component patchable.
Makes the settings components patchable and usable in plugins.
Makes the `PerformerDetailsPanel`, `CompressedPerformerDetailsPanel` components patchable, and adds a new patchable component `PerformerDetailsPanel.DetailGroup`.

Adds a `PluginSettings` component that can be hooked with `patch.after` to replace with a custom set of settings:
```
(function() {
    const PluginApi = window.PluginApi;
    const React = PluginApi.React;
    const { useSettings } = PluginApi.hooks;

    function pluginSettingsHook (props, _, result) {
        const { StringListSetting } = PluginApi.components;
        const { pluginID } = props;
        const { plugins, savePluginSettings } = useSettings();
        const pluginSettings = plugins[pluginID] ?? {};

        if (props.pluginID === "settingTest") {
            return React.createElement("div", null, [
                React.createElement(StringListSetting, {
                    heading: "String List Example",
                    value: pluginSettings.testSetting ?? [],
                    onChange: (value) => savePluginSettings(pluginID, { ...pluginSettings, testSetting: value }),
                }),
            ]);
        }
        
        return result;
    }
    PluginApi.patch.after('PluginSettings', pluginSettingsHook);
})();
```

### DIFF
--- a/ui/v2.5/src/App.tsx
+++ b/ui/v2.5/src/App.tsx
@@ -50,6 +50,7 @@ import { PluginRoutes } from "./plugins";
 // import plugin_api to run code
 import "./pluginApi";
 import { ConnectionMonitor } from "./ConnectionMonitor";
+import { PatchFunction } from "./patch";
 
 const Performers = lazyComponent(
   () => import("./components/Performers/Performers")
@@ -143,6 +144,13 @@ function sortPlugins(plugins: PluginList) {
 
   return sorted;
 }
+
+const AppContainer: React.FC<React.PropsWithChildren<{}>> = PatchFunction(
+  "App",
+  (props: React.PropsWithChildren<{}>) => {
+    return <>{props.children}</>;
+  }
+) as React.FC;
 
 export const App: React.FC = () => {
   const config = useConfiguration();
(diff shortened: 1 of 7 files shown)


Valkyr-JS:
Hey, thanks for adding the `PerformerDetailsPanel` components, they're working nicely. Could I request the `TagCard` component as well, if that's okay?

Thanks

PR #4958, opened 2024-06-11
Performer urls

### LINKED ISSUE #703: [Feature] Allow multiple url's for a performer
**Is your feature request related to a problem?
The performers has a single url field as well as a twitter and instagram profile.
This should be replaced with a list allowing an unlimited amount of url's to be added.
Url's are a useful way of adding identifiers for sites so you can submit the profile url on each site for example.

Create a new table to store url's and update the user interface to allow you to enter more information.
When the user adds a url the user interface should create a blank url field  encouraging the user to enter more url's

**Additional context**
Having more url's allows you to use these as identifiers in other databases.
If we can have the profile url on each adult website it can help with disambiguation of performers with similar or the same name.
When scraping a site you can look for url's that match the profile url attached to the performer.

WithoutPants:
Replaces the `url`, `twitter` and `instagram` fields with a single `URLs` field. The old fields are deprecated - scrapers using the old field will have the values populated in the URLs field.

The link button in the Performer page will now show a dropdown of all URLs. Twitter and Instragram links are still shown as separate link buttons and are not included in the URL dropdown.

Resolves #703


bob123491234:
The old Instagram and Twitter fields allowed for users to only enter a username, this does not account for that in the migration.

QxxxGit:
Small thing: is the same styling of URL icons intentional? I noticed the Instagram icon is no longer pink but blue.

WithoutPants:
> The old Instagram and Twitter fields allowed for users to only enter a username, this does not account for that in the migration.

Good catch, thanks. The migration has been updated to translate strings that don't start with `http[s]://` to URLs.

> Small thing: is the same styling of URL icons intentional? I noticed the Instagram icon is no longer pink but blue.

Nope. This is now addressed as well. Thanks for the heads up.

PR #4962, opened 2024-06-12
Default view filters

### LINKED ISSUE #2404: Default view / saved filters for gallery view
In gallery view, I am unable to set a default view.  Saved filters don't work, since the gallery-dependance is always included as a filter.

image

**Expected behavior**
I am able to use filters and "Set as default" just like when viewing scenes. This would probably mean handling gallery view without applying the filter "Galleries include".

Thanks for your efforts, I really enjoy stash
Sorry if this is a duplicate. I know, this problem exists for a while now..

### LINKED ISSUE #3079: [Feature] Change default sort

### LINKED ISSUE #4010: [Feature] performer page default filters (optional ofc)

### LINKED ISSUE #4137: [Feature] Custom default sorting
I'm always frustrated when I open the video tab, and I need to change the page size and the sorting method to be the way I like.

I would like an option to set current page size / sorting / display / filter on each tabs (videos, galleries, movies, ...)

WithoutPants:
Supercedes #4036 
Resolves #2404
Resolves #4010 
Resolves #3079
Resolves #4137 

Adapts @yoshnopa's work from #4036. Deprecates the default filter functions from the graphql interface. Instead, all  default filter settings are stored in the UI settings. Each view in the UI has its own default filter that can be set.

The default filter is used when the view is loaded without any parameters.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	github.com/knadh/koanf v1.5.0
 	github.com/lucasb-eyer/go-colorful v1.2.0
 	github.com/mattn/go-sqlite3 v1.14.22
+	github.com/mitchellh/mapstructure v1.5.0
 	github.com/natefinch/pie v0.0.0-20170715172608-9a0d72014007
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
 	github.com/remeh/sizedwaitgroup v1.0.0
@@ -88,7 +89,6 @@ require (
 	github.com/mattn/go-isatty v0.0.19 // indirect
 	github.com/mitchellh/copystructure v1.2.0 // indirect
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
-	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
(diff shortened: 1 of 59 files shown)


yoshnopa:
Great, just tested it!
However, I have an issue left:
I added the possibility to have the default filter/view applied on links as well. This does not seem to apply to your commit. 
That one I still find to be the source of quite some inconsistency. For example, I'd assume that if I click on the gallery symbol of a tag in the front page that it will at the very least use my default view options (basically be the same as if I visited galleries regularly, just with the tag filter as an addition). 
Or is there a reason that was phased out?

WithoutPants:
I left it out to reduce the scope. I don't have immediate plans to implement it, but no injection to its inclusion. 

Probably needs to be discussed in a separate feature request, but my preference would be that the tag links don't include a sort parameter and that the URL resolution code uses the default filter to determine the sort and view if these aren't provided.

Croxi1:
Great work! 

I am eternally grateful for everything you are doing!

Any chances of a timeline of version 0.27.0? Changing default scene number on a Performer page is a critical update for me which I cannot wait for.

Again, thank you!

PR #4963, opened 2024-06-12
Rename Movie to Group in UI

WithoutPants:
This is a largely cosmetic change to the UI. It renames `Movie` to `Group` for most labels and UI URLs. The icon is unchanged.

Importantly, graphql and database schemas are unchanged - these will still be `movie` for the time being.

The `movies` menu item in the navbar menu items configuration is silently translated to Groups.

Part of #4905


xx790:
Likely a typo in the number, and this is a part of
- #4905

Issue #4945, opened 2024-06-07
[Feature] Hardware transcoding on macOS (VideoToolbox)


honeybeaver:
**Is your feature request related to a problem? Please describe.**

In connection to this feature in the 0.26.0: 
> When enabled, full hardware transcoding is used where possible. (https://github.com/stashapp/stash/pull/4765)

I tested this on a Mac with Apple Silicon. It appears ffmpeg HW transcoding does not use VideoToolbox on Mac yet, when streaming media. 

**Describe the solution you'd like**

When on macOS, test & use these flags in ffmpeg to enable HW transcoding using VideoToolbox (from [ffmpeg docs](https://trac.ffmpeg.org/wiki/HWAccelIntro#VideoToolbox)):

> To use H.264/HEVC hardware encoding in macOS, just use the encoder `-c:v h264_videotoolbox` or `-c:v hevc_videotoolbox` for H.264 or HEVC respectively. 

**Describe alternatives you've considered**

Without hardware acceleration, software transcoding works on the CPU. In most cases though, transcoding takes a lot CPU time, either making it not possible to transcode quickly enough (rare) or affecting other apps running on the machine (common). 

**Additional context**

This may also apply to decode and other times ffmpeg is converting media, such as scan and previews, and not just streaming. 

Intel Macs also use this method to access hardware encoding. 

Searching the code, it does not look like VideoToolbox appears anywhere yet, and I didn't see a full discussion of this case in other HW transcoding discussions.

Thank you!

feederbox826:
The referenced full hardware transcoding is for NVIDIA cards only - I don't belive stash has a check for VideoToolbox support at all

NodudeWasTaken:
Unfortunately scale_vt (videotoolbox) currently doesn't call ff_scale_adjust_dimensions, meaning it cannot scale with respect to the aspect ratio unless i insert some hacks or refactor how that's done.
> The referenced full hardware transcoding is for NVIDIA cards only - I don't belive stash has a check for VideoToolbox support at all

The referenced full hardware transcoding is enabled for NVIDIA, Intel and VAAPI.

m13v:
We use VideoToolbox extensively in Fazm for encoding screen recordings on macOS. Here are the flags that work well for us:

For HEVC hardware encoding:
- `-c:v hevc_videotoolbox -q:v 65 -tag:v hvc1` gives good quality at reasonable file sizes
[...]

PR #4986, opened 2024-06-18
Add apple encoder and fix extra_hw_frames bug

NodudeWasTaken:
A discord user complained that the decoder was bugging out, i suspected this was due to extra_hw_frames, jellyfin fixed this by switching from nv12 to yuv420p which disables passthrough.
Jellyfin also adds -threads 1 to cuda to save vram.

Additionally adds the apple encoder for hardware acceleration and brings the requisite fixes.
Like scale_vt lacking dimension adjustments, so we retrofitted the intel hack to calculate the dimensions and used it for apple too.
It also lacks a format switcher, so no 10bit fixes for it.
It also has a higher min resolution than anything we've seen yet, so upped the test resolution.

The only note is that given the apple encoders nature, it really sucks at startup, and given HLS's current tendency to start and stop, it makes the video look like crap.
It technically works, but i dont know that we should keep it.

Last time i fixed a bug that caused using no filter to error out because of the pixel format, so now we can add a format switcher if available.

Closes https://github.com/stashapp/stash/issues/4945

### DIFF
--- a/pkg/ffmpeg/codec_hardware.go
+++ b/pkg/ffmpeg/codec_hardware.go
@@ -204,26 +204,19 @@ func (f *FFMpeg) hwFilterInit(toCodec VideoCodec, fullhw bool) VideoFilter {
 	return videoFilter
 }
 
-var scaler_re = regexp.MustCompile(`scale=(?P<value>[-\d]+:[-\d]+)`)
+var scaler_re = regexp.MustCompile(`scale=(?P<value>([-\d]+):([-\d]+))`)
 
 func templateReplaceScale(input string, template string, match []int, vf *models.VideoFile, minusonehack bool) string {
 	result := []byte{}
 
 	if minusonehack {
-		matches := scaler_re.FindStringSubmatch(input)
-		split := strings.Split(matches[1], ":")
-		if len(split) != 2 {
-			logger.Error("split not found")
-			return input
-		}
-
 		// Parse width and height
-		w, err := strconv.Atoi(split[0])
+		w, err := strconv.Atoi(input[match[4]:match[5]])
 		if err != nil {
 			logger.Error("failed to parse width")
 			return input
 		}
-		h, err := strconv.Atoi(split[1])
+		h, err := strconv.Atoi(input[match[6]:match[7]])
 		if err != nil {
 			logger.Error("failed to parse height")
 			return input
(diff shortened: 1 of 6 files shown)


kermieisinthehouse:
On this branch, I receive this error while playing any avi file under cuda:

```
[transcode] ffmpeg error when running command </usr/bin/ffmpeg -hide_banner -v error -hwaccel_device 0 -threads 1 -hwaccel cuda -hwaccel_output_format cuda -i video.avi -c:v h264_nvenc -rc vbr -cq 15 -movflags frag_keyframe+empty_moov -ac 2 -f mp4 pipe:>: [mpeg4 @ 0x56289828cf00] No decoder surfaces left
Error while decoding stream #0:0: Cannot allocate memory
[mpeg4 @ 0x56289828cf00] No decoder surfaces left
Error while decoding stream #0:0: Cannot allocate memory
Impossible to convert between the formats supported by the filter 'Parsed_null_0' and the filter 'auto_scaler_0'
Error reinitializing filters!
Failed to inject frame into filter network: Function not implemented
Error while processing the decoded data for stream #0:0
```

NodudeWasTaken:
> On this branch, I receive this error while playing any avi file under cuda:
> 
> ```
> [transcode] ffmpeg error when running command </usr/bin/ffmpeg -hide_banner -v error -hwaccel_device 0 -threads 1 -hwaccel cuda -hwaccel_output_format cuda -i video.avi -c:v h264_nvenc -rc vbr -cq 15 -movflags frag_keyframe+empty_moov -ac 2 -f mp4 pipe:>: [mpeg4 @ 0x56289828cf00] No decoder surfaces left
> Error while decoding stream #0:0: Cannot allocate memory
> [mpeg4 @ 0x56289828cf00] No decoder surfaces left
> Error while decoding stream #0:0: Cannot allocate memory
> Impossible to convert between the formats supported by the filter 'Parsed_null_0' and the filter 'auto_scaler_0'
> Error reinitializing filters!
> Failed to inject frame into filter network: Function not implemented
> Error while processing the decoded data for stream #0:0
> ```

Please check if thats fixed now.
As a side note this seems to be a ffmpeg version 4 issue

kermieisinthehouse (approved):
Tested working but not code reviewed

PR #5004, opened 2024-06-21
Address resize loop

cj12312021:
This pull request updates the card resize code to only adjust the size when the container width changes by a significant enough margin. In this case, that margin is 15 pixels, which is the amount of horizontal space I see that the default scrollbar occupies. This value can increased if necessary.

### DIFF
--- a/ui/v2.5/src/components/Shared/GridCard/GridCard.tsx
+++ b/ui/v2.5/src/components/Shared/GridCard/GridCard.tsx
@@ -44,7 +44,7 @@ interface IDimension {
 
 export const useContainerDimensions = <
   T extends HTMLElement = HTMLDivElement
->(): [MutableRefObject<T | null>, IDimension] => {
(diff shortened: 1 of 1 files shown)


cj12312021:
I browsed around to all the tools I was aware of, which dynamically sized the content based on container size, and noticed they all were either unaware of the issue or opted not to solve it. Said sites and tools are all the sites under the gamma umbrella (Adulttime, Evil Angel, Dfxtra, etc.) and the Mosicac plugin we use for the image wall view. Gamma sites won’t see this issue very easily because their footer or sidebar often ensures overflow is always after on their pages. I mention this because the solution comes at the cost of sensitivity and responsiveness. The margin I specified here is small enough to not be much of a hindrance. But I wanted to ensure the tradeoffs made here is clear.

WithoutPants:
I increased the sensitivity threshold to 20. The existing value didn't address the original issue on my Windows 10 system. Apparently the default scrollbar width is around 17 pixels on Windows.

PR #5061, opened 2024-07-08
Copy apikey query parameter to DASH & HLS manifest

Lenny3D:
When an API key is provided to the DASH and HLS manifest endpoints, this it will now be copied to the URLs inside the manifest. This allows for clients that are only able to pass an URL to an (external) video player to function in case authentication is set up on stash.


WithoutPants:
Thanks! I made some minor changes. The apikey setting code should really be outside the `ffmpeg` package, ideally with some kind of callback/hook. However, that's a larger change than this deserves and I'd rather merge this as-is and do that at a later date.

PR #5076, opened 2024-07-15
Add birthdate to performer select and restyle

### LINKED ISSUE #5062: [Feature] Scene Tagger Performer Picker: move disambig to under performer name
Some studios have a lot of single name performers and their disambiguation becomes very important.  The search box does not search the name and disambiguation so you need to hand select the proper individual based on a rather small image.  Rather than cross a UI with a functionality request, I will stick to the UI elements here.

image

1) There is a lot of wasted vertical space in the current UI.  When we added the images this made the name rows significantly taller.  Using that space for two or even three rows of information would be a great option.  I'd love to see at a minimum breaking the display down such that the Disambiguation is on the line underneath.  Furhter if available we could consider birthdates/ages/nationality on a thrid row.

2) There is not a lot of horizontal space in the current UI.  Even performers with relatively short names can get cut off.  It would be good if we could widen this picker say 25-50px.  *Not reflected in the mockup*

image

I have tried searching using alternative means, and in some cases I have added synthetic aliases.
I briefly tried to address this with a userscript but I don't have the skills for this.

dogwithakeyboard:
Restyles the Performer select in a similar pattern to the Scene, Gallery and  Movie/Group selects.

As suggested in #5062, moves the disambiguation to a new line and adds the birthdate and calculated age to a third line.

Example in the scene editor:
after1

Example In the tagger interface:
after2

Resolves #5062

### DIFF
--- a/ui/v2.5/graphql/data/performer-slim.graphql
+++ b/ui/v2.5/graphql/data/performer-slim.graphql
@@ -39,4 +39,6 @@ fragment SelectPerformerData on Performer {
   disambiguation
   alias_list
   image_path
+  birthdate
+  death_date
 }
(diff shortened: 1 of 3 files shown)


Maista6969:
Should we calculate the age based on the date on the scene/gallery (if it has been filled in) rather than the current date?

WithoutPants:
> Should we calculate the age based on the date on the scene/gallery (if it has been filled in) rather than the current date?

Ideally yes, but I'm happy to defer this to a separate PR.

PR #5080, opened 2024-07-17
Refactor ItemList code and re-enable viewing sub-tag/studio content

### LINKED ISSUE #4938: [Bug report] Enable quick toggling of sub-tag/sub-studio content
When using studio/tag hierarchies there may be many levels of parent tags/studios. Previously it was relatively easy to open the pre-applied filter and toggle inclusion of sub-content, now a new filter must be applied to include content you are looking at then toggled

Add an option to toggle inclusion of sub-tag/sub-studio content when viewing any tag/studio which is a parent. Default setting is available as an option already but changing views can be desired

Returning previous behavior changed by 4705 as plugin/settings toggle

**Additional context**
Behavior changed due to https://github.com/stashapp/stash/pull/4705 with v26.0

WithoutPants:
This PR provides a much-needed refactor to the ItemList code. This changes the `makeItemList` hook to be two separate components: `ItemListContext` and `ItemList`. This simplifies the item list code significantly, with URL filter setting moved into a hook, and filter and result states moved into contexts that can be accessed from anywhere in the child component tree.

While working in this area, I also changed the tag and studio views to show the filter tags so that users can change the sub-tag/studio visibility per previous behaviour. I think in future this should be changed to be a checkbox or something, but this will do for now.

Resolves #4938

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryList.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryList.tsx
@@ -4,7 +4,7 @@ import cloneDeep from "lodash-es/cloneDeep";
 import { useHistory } from "react-router-dom";
 import Mousetrap from "mousetrap";
 import * as GQL from "src/core/generated-graphql";
-import { makeItemList, showWhenSelected } from "../List/ItemList";
+import { ItemList, ItemListContext, showWhenSelected } from "../List/ItemList";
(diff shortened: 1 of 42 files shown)


WithoutPants:
I was unhappy with showing the filter tags, so I have instead added a toggle button.

image

It's still not ideal, as the toggle should be in the filter toolbar row, but this will do for this iteration.

PR #5082, opened 2024-07-18
Rename movie tables to groups in database schema

WithoutPants:
Renames the `movies` tables to `groups` in the SQL schema.

Also rename the `synopsis` column to be `description`. Renaming in the graphql schema, GUI etc will be in a separate PR.

### DIFF
--- a/pkg/sqlite/migrations/65_movie_group_rename.up.sql
+++ b/pkg/sqlite/migrations/65_movie_group_rename.up.sql
@@ -1,4 +1,5 @@
 ALTER TABLE `movies` RENAME TO `groups`;
+ALTER TABLE `groups` RENAME COLUMN `synopsis` TO `description`;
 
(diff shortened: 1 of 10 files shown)


pkg/sqlite/migrations/65_movie_group_rename.up.sql
```diff
@@ -0,0 +1,23 @@
+ALTER TABLE `movies` RENAME TO `groups`;
+
+DROP INDEX `index_movies_on_name`;
+CREATE INDEX `index_movies_on_name` ON `groups`(`name`);
```
bob123491234, line 4: I assume this should be `index_groups_on_name`.

WithoutPants: Thanks!

PR #5092, opened 2024-07-23
Update freeones.go

L3GW0RK:
updated FreeOnes scraper to fix selector changes over the years and refined fields to sync better with stashboxes/stashdb


WithoutPants:
Thanks for your PR. I'm closing this one in favour of #5171 which updates from the CommunityScrapers version. For reference, `BreastType` doesn't work as a field in the scraper spec.

PR #5110, opened 2024-08-02
Performer select calculated ages

dogwithakeyboard:
Following on from #5076 this shows the age of the performer in the select based on the scene/image/gallery date, when one exists. 

The width of the performer select in the tagger view has also been slightly increased.

perfage3


DogmaDragon:
Censored the image for safety reasons.

echo6ix:
From what I can tell it looks like the $age value is being inserted into the CSS birthdate selector. The age value should get its own CSS selector.

`<span className="performer-select-age">{$age}</span>`

Helloworldexampledotcom:
https://crypto.link.com?ref=lb&source_amount=1500&source_currency=usd&destination_currency=btc&destination_network=bitcoin
    
  
  
  
  
  
  

  
  
>

Helloworldexampledotcom (approved):

Helloworldexampledotcom:
https://dashboard.stripe.com/test/connect/financial-accounts

echo6ix:
@kermieisinthehouse @dogwithakeyboard @WithoutPants 

image

This is very nitpicky, and I do not want to stifle this PR, but the proposed locale string `{age} on this date` comes directly after the the birthdate (yyyy-mm-dd) value. See above image.

To you and I, we know which date is being referenced, but on it's face as far as clear and unambiguous language is concerned, this is really awkward phrasing as someone who's not familiar with the app could misconstrue what "on this date" is referencing, especially being juxtaposed to the birthdate .

To remove ambiguity I would suggest `{age} {years old} on scene date` and `{age} {years old} on image date` respectively. It goes back to referencing the context, but in a way that does not imply they were in the content.

Does that make sense?

WithoutPants:
> To remove ambiguity I would suggest `{age} {years old} on scene date` and `{age} {years old} on image date` respectively. It goes back to referencing the context, but in a way that does not imply they were in the content.

In order to move this forward, I've changed the wording on this to be `x at production`, which is concise, clear and unambiguous to me. 

The wording in the performer card in a scene/image/gallery was also incorrect for image or gallery - `x years old in this scene`. I have changed this to `x years old at production`. These don't necessarily have to be the final wordings, but that can refined in a separate PR.

PR #5130, opened 2024-08-15
Update SceneDuplicateChecker table to show audio codec and file modification time.

notsafeforgit:
This should make identifying different versions of scenes easier without having to click through to the full scene page. The reason I want this is sometimes I have a copy of a scene that's somehow missing audio and while it flags as an exact duplicate, I'd like to keep the one with the audio intact.


notsafeforgit:
The lint checks I'm failing aren't related to the file I changed, so they can probably be ignored.

Issue #4917, opened 2024-06-03
Make scenes filter by file size.


misterscv:
It would be nice to have ability to filter scenes by file size. Eg. More than some size, less than some size. Thanks in advance

discourse-stashapp:
This issue has been mentioned on **Stash**. There might be relevant details there:

https://discourse.stashapp.cc/t/file-size-and-file-modification-time/5213/2

PR #5131, opened 2024-08-15
File Size Filtering

Gykes:
Adding in support for file size filtering for scenes base on https://github.com/stashapp/stash/issues/4917


WithoutPants:
I'd prefer to do this via the `FileFilterType` using the related files filtering feature in #6158. Note also that in order to close #4917, the UI will need to be done as well. That doesn't necessarily need to be done in this PR, but we can't close it until the UI changes are also in.

PR #5133, opened 2024-08-17
Gallery scrubber

WithoutPants:
Adds a gallery preview scrubber to gallery cards. Like the scene preview scrubber, allows scrubbing through all of the images in a gallery. Clicking on the scrubber navigates to the current image detail page.

image


DogmaDragon:
Tested with zip/non-zip galleries, found no issues.

OKEAMAH (commented):
Cool

ui/v2.5/src/components/Galleries/GalleryCard.tsx
```diff
@@ -167,14 +207,13 @@ export const GalleryCard: React.FC<IProps> = (props) => {
...
-              loading="lazy"
-              className="gallery-card-image"
-              alt={props.gallery.title ?? ""}
-              src={`${props.gallery.cover.paths.thumbnail}`}
-            />
-          ) : undefined}
+          <GalleryPreview
+            gallery={props.gallery}
+            onScrubberClick={(i) => {
+              console.log(i);
```
cj12312021, line 213: Did you mean to commit this to production code

ui/v2.5/src/components/Galleries/GalleryCard.tsx
```diff
@@ -14,6 +14,45 @@ import { faBox, faPlayCircle, faTag } from "@fortawesome/free-solid-svg-icons";
...
+  gallery,
+  isPortrait = false,
+  onScrubberClick,
+}) => {
+  const [imgSrc, setImgSrc] = useState<string | undefined>(
+    gallery.cover?.paths.thumbnail ?? undefined
+  );
+
+  return (
+    <div className={cx("gallery-card-cover", { portrait: isPortrait })}>
```
cj12312021, line 37: I noticed the height and width dimensions were removed from the cover property, and it currently looks like `isPortrait` is always set to false. How did you plan to identify when a gallery card has a portrait image?

WithoutPants: Looks like it was a vestigial artifact from the `ScenePreview` that it was copied from. As far as I can determine, there's no need for it in the Gallery preview/scrubber, and the styling needs to accommodate for both orientations anyway.

PR #5139, opened 2024-08-19
More PatchComponent Changes

Gykes:
Modified all .tsx within /settings/tasks

### DIFF
--- a/ui/v2.5/src/components/Settings/Tasks/CleanGeneratedDialog.tsx
+++ b/ui/v2.5/src/components/Settings/Tasks/CleanGeneratedDialog.tsx
@@ -7,64 +7,68 @@ import { faTrashAlt } from "@fortawesome/free-solid-svg-icons";
 import { SettingSection } from "../SettingSection";
 import { useSettings } from "../context";
 import { LoadingIndicator } from "src/components/Shared/LoadingIndicator";
+import { PatchComponent } from "src/patch";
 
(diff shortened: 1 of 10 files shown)


ui/v2.5/src/components/Settings/SettingsScrapingPanel.tsx
```diff
@@ -26,13 +26,14 @@ import {
...
 import { Counter } from "../Shared/Counter";
+import { PatchComponent } from "src/patch"; // Import PatchComponent
 
 const ScraperTable: React.FC<
   PropsWithChildren<{
     entityType: string;
     count?: number;
   }>
-> = ({ entityType, count, children }) => {
+> = PatchComponent("ScraperTable", ({ entityType, count, children }) => {
```
WithoutPants, line 36: As above, I think this should be SettingsScrapingPanel.ScraperTable. I won't mark all of these, but please do a pass and correct the non-exported ones.

WithoutPants (changes requested):
It might be worth looking over the non-exported components to see their value in being patchable/usable in plugins. It may not make sense to do them all.

DogmaDragon (changes requested):
Changes needs to be documented in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/UIPluginApi.md.

PR #5145, opened 2024-08-20
Scene player presentation improvements

WithoutPants:
Scene player controls are now shown before the video is loaded, so that the volume, sources etc can be interacted with without playing the video.

image

The error dialog now allows interacting with the controls.

image

The source selector will no longer remove failed sources from the list. Sources that failed to load are now shown with a muted styling and italics.

image

The source selector will no longer automatically play the next source due to a load failure if the source was manually selected by the user.


DogmaDragon:
Fileless scenes now have player controls shown as a result of this change.

WithoutPants:
> Fileless scenes now have player controls shown as a result of this change.

I'll address this in #5141.

Issue #1363, opened 2021-05-03
[Feature] Short Form Title (50 chars or less.)


silembir:
**Describe the solution you'd like**
A setting to show the studio image for each scene in the wall view.  The studio image shows in grid view, but it would be helpful to have that same image when looking at the wall view.
 
Let me know if there are any questions.  Thanks!

echo6ix:
If implemented please add a toggle in the settings to enable/disable this feature.

PR #5149, opened 2024-08-20
Adding Studio Logo to Scenes in Wall View

Gykes:
Not 100% sure about this one tbh. At what point does the Wall view just become Grid with slightly larger scene cards?

closes https://github.com/stashapp/stash/issues/1363

Here is a [before](https://gyazo.com/90ca7dfa0cf1708f7a380ec45c82712b) and [after](https://gyazo.com/c4633382b7e60bc383739a7ce0f848cc)


Gykes:
> Not a huge fan... If we're moving forward with this, I agree with the comment in #1363 that it should be togglable.

That's the general consensus that I'm getting as well and, quite frankly, I agree. Would be much better suited as a plugin.

WithoutPants:
Closing this in favour of pursuing a plugin-based solution.

Issue #3856, opened 2023-06-25
[Feature] include high FPS in quality marker


Gouvernathor:
When displaying a scene in a list, the quality is shown as "720p" or "1080p".
![image](https://github.com/stashapp/stash/assets/44340603/dc3e59ae-17ba-4d6a-ba12-c6bf83114836)
For videos in 59.x to 60 fps, it should be indicated as "1080p60", or "720p60" if that's a thing, or "1080p50"...

PR #5150, opened 2024-08-20
Adding Frame_rate to Scene Cards

Gykes:
Currently stash uses 1080P, note the capital P, I prefer 1080p but who cares about what I prefer. 

I kept the P capital, since that's the standard, and just tacked the frame_rate to the end.

[Before](https://gyazo.com/7616c90c460150fefd739382093f80a7) vs [after](https://gyazo.com/74f8da62ca7ba5ddaaa94fe9cd9a3b2e)

closes https://github.com/stashapp/stash/issues/3856


Gykes:
Closing because this is better as a plugin

PR #5153, opened 2024-08-21
Tweak to allow for auto-convert of metric measurements into US (Imperial)

Darklyter:
Simple tweak to add a user-configurable option (Settings->Interface->Editing) to allow for auto conversion of metric measurements (Bust-Hips-Waist) into US (Imperial) values.  It will auto-convert any set of measurements that are greater than 50-45-50 in all three fields, otherwise it will not affect the entry.  It will also convert letter cup sizes into US equivalents (For example "F" becomes "DD" if it performs a conversion.

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
@@ -190,6 +190,69 @@ export const PerformerEditPanel: React.FC<IPerformerDetails> = ({
     }
   }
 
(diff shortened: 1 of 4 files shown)


Gykes:
So, with https://github.com/stashapp/stash/pull/6643 I suggested that we split the field into 3 separate fields for easier filtering/sorting. Once that is sorted I think we could come up with a better way to handle conversions. The problem with conversions, according to @feederbox826 and https://discourse.stashapp.cc/t/bra-cup-sizes-and-conversions/1407, the conversion isn't super straight forward.

Going to close this for now and we can come back to it at a later date after the fields have been split.

Issue #5069, opened 2024-07-13
[Bug Report] Intel QuickSync not working


gitgiggety:
**Describe the bug**
For me Intel QuickSync support is not working on an Intel N5105 based system. QSV is tested for / used with a combination of options which is not compatible with at least Jasper Lake CPUs. This is due to `-global_quality` being used which leads to a ratecontrol as described here: https://ffmpeg.org/ffmpeg-codecs.html#Ratecontrol-Method This while Jasper Lake CPUs are limited in functionality and seemingly only supports some of those ratecontrol methods.

**To Reproduce**
1. Install Stash on a system with a Jasper Lake CPU like an N5105
2. Enable "FFmpeg hardware encoding"
3. Restart

**Expected behavior**
Startup log message shows that h264_qsv is supported. But instead is just:
> [InitHWSupport] Supported HW codecs [2]:
>        h264_vaapi
>        vp9_vaapi

While this could/should include h264_qsv and possibly vp9_qsv.

**Screenshots**
-

**Stash Version: (from Settings -> About):**
0.26.2

**Desktop (please complete the following information):**
-

**Smartphone (please complete the following information):**
-

**Additional context**
Trace logs of startup:
> [InitHWSupport] Codec h264_qsv not supported. Error output:[h264_qsv @ 0x7fecc212dc00] Error initializing the encoder: invalid video parameters (-15)
[vost#0:0/h264_qsv @ 0x7fecc2ac7540] Error while opening encoder - maybe incorrect parameters such as bit_rate, rate, width or height.
Error while filtering: Invalid argument
[out#0/null @ 0x7fecc2ae5280] Nothing was written into output file, because at least one of its streams received no packets.

Used ffmpeg flags to check QSV support:
```
-hide_banner -v warning -init_hw_device qsv=hw -filter_hw_device hw -f lavfi -i color=c=red:s=1280x720 -t 0.1 -c:v h264_qsv -global_quality 20 -preset faster -vf hwupload=extra_hw_frames=64,format=qsv,scale_qsv=-1:256 -f null -
```
Execting ffmpeg with the same flags, except dropping `global_quality 20` executes succesfully.

Note:
I haven't tested if there are any other issues yet, but transcoding a 4K video to 1280x720 with a basic command like:
```
ffmpeg -hwaccel qsv -c:v h264_qsv -i input.mp4 -vf scale_qsv=w=1280:h=720 -c:v h264_qsv output.mp4
```
[...]

PR #5154, opened 2024-08-21
Add video codec profiles

NodudeWasTaken:
This commits adds profiles instead of direct codec strings to the encoder.
This enables us to adapt to different hardware scenarios for each codec, like different generations of hardware supporting different flags for a specific encoder etc.

Specifically this adds a QSV codec that removes global_quality as some hardware doesn't support it.
It also adds a NVENC profile that enables alot of quality flags 

Fixes [#5069](https://github.com/stashapp/stash/issues/5069)

### DIFF
--- a/pkg/ffmpeg/codec.go
+++ b/pkg/ffmpeg/codec.go
@@ -19,14 +19,14 @@ func (c VideoCodec) Args() []string {
 
 var (
 	// Software codec's
-	VideoCodecLibX264 VideoCodec = makeVideoCodec("x264", "libx264")
-	VideoCodecLibWebP VideoCodec = makeVideoCodec("WebP", "libwebp")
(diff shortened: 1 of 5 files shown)


JanJastrow:
I'm using AV1 files for my newly converted files, can you also add support for this?
As far as I could see it's not on the codec list.
(Or maybe this should be a seperate issue)

NodudeWasTaken:
> I'm using AV1 files for my newly converted files, can you also add support for this? As far as I could see it's not on the codec list. (Or maybe this should be a seperate issue)

I could, but its best if you raise an issue first

Net005:
Want to test this new feature out with an Arc GPU and does it require any (ffmpeg) config setitings to work properly with QSV?
As before it was conflicting with certain hard coded ffmpeg settings from with Stash (issue #5069) and wondering if this just works automatically now.

Issue #4600, opened 2024-02-20
[Feature] [UI/Localization] Simplify label of DELETE dropdown items


echo6ix:
Videos that are classified as images are correctly displayed in the image section of Stash. When viewing this content on an image detail page, the overflow menu dropdown item option to delete reads `Delete Image`.

![image](https://github.com/stashapp/stash/assets/37937507/dd8a5d45-4515-4d2c-861e-a98d636f18f7)

While trivial, this is technically inaccurate when the image content is a video clip. I think it might be optimal to rename all dropdown item delete options like this the string `Delete` only, without mention of content type.

This would be applicable to:

localhost:9999/scenes/n
localhost:9999/images/n
localhost:9999/galleries/n

PR #5157, opened 2024-08-22
Remove {entityType} from delete_entity

Gykes:
Removed the Entity type from delete_entity. This was proven on scenes tab and galleries tab. I have no images to delete in Dev so can't test that.

[Before](https://gyazo.com/47f25feb7eab0cacc3a0eeca8248ef1d) and [after](https://gyazo.com/2633bb262a062a36022b419df61599e3)

closes https://github.com/stashapp/stash/issues/4600

### DIFF
--- a/ui/v2.5/src/locales/af-ZA.json
+++ b/ui/v2.5/src/locales/af-ZA.json
@@ -19,7 +19,7 @@
         "create_entity": "Skep {entityType}",
         "create_parent_studio": "Skep ouerateljee",
         "delete": "Verwyder",
-        "delete_entity": "Verwyder {entityType}",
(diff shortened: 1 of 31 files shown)


WithoutPants (changes requested):
This change makes `delete_entity` exactly the same as `delete`. In this case, the better approach is to remove the `delete_entity` key and replace its uses with `delete`.

Gykes:
> This change makes `delete_entity` exactly the same as `delete`. In this case, the better approach is to remove the `delete_entity` key and replace its uses with `delete`.

Copy that, will get it done today some time probably

Gykes:
Will reopen new PR and reference this...

Issue #3122, opened 2022-11-12
[Feature] Create All/New/Missing on Tagger page


scruffynerf:
A common request/concern from new users is:

"Wait, if I use Tagger, like you recommend, I have to click on EVERY new performer/studio/tag to add them? That will take forever, and I don't want to do that. Don't you guys have a setting to just auto add?"

Alternatives like using Identify (which can auto create, but fail to add images/etc), also fail to allow any sort of manual checking... it's basically "all or nothing".

If a "Create All" button/option is added to Tagger, most of this resistance will go away.  While there is a userscript that add 'save all' as an option, the stock app would retain the relative safety of still requiring someone to manually review, but allow them to not have to hit 'create' (and then confirm/save, so 2 clicks, plus load time) for each new performer/studio and each new tag.

Either a button (Create Missing? Create All?) or a set of options ("Create Missing Performers, Create Missing Studios, Create Missing Tags") would be good, but I'd favor the latter, as allowing partial automation... Someone might not want new tags, for example.

WithoutPants:
I spent quite a bit of time prototyping this functionality based on the original description, but was ultimately unsatisfied with the resulting user experience. 

A `Create All` button feels like a sledgehammer approach, and I'd prefer a more nuanced solution.

The goal of the following design is to allow the user to configure the default approach for the tagger, then allow them to save all matching scenes with a single click. It should also be possible for the user to select scenes to ignore during the save all operation.

I would also like to eliminate the multiple server trips involved when handling a single scene. Currently, if a user wants to create everything attached to a given scene, they must click create on the studio, click to confirm the dialog, and repeat for each performer. Each of these results in a trip to the server to create the objects. 

Instead, the create button to be a `mode` button, like `Skip`. Clicking the create button once will set for the studio/performer to be created when the scene is saved. Clicking the create button a second time would bring up the create dialog per existing behaviour.

Similarly, the new tags list will be changed so that clicking on a tag will toggle it to be created (or not) during save. 

A `Save All` button would be added to save all of the matched/selected scenes, based on the current options on each result. 

We want to be able to prevent an unwanted scene from being saved during the save all operation. Adding a deselect, or close/x button somewhere on the card would facilitate this.

[...]

PR #5158, opened 2024-08-22
[WIP] Adding In Tagger Buttons

Gykes:
Adding in Search all and Create all buttons to tagger view.

So far this will add a Search All button, and then if any scenes have been found, generate a create all button. Once you are comfortable that all scenes are matched correctly you can click "create all" and it will create all performers and studios for you.

This is currently a WIP, the core functionality is there, but the CSS is a smidge wonky. I will also be adding in a "Save ID" button which will click the save StashID button so that any performers and scenes can get an associated StashID. This takes some very well used userscripts and plugins to the core app.

closes https://github.com/stashapp/stash/issues/3122

### DIFF
--- a/ui/v2.5/src/components/Tagger/scenes/SceneTagger.tsx
+++ b/ui/v2.5/src/components/Tagger/scenes/SceneTagger.tsx
@@ -185,6 +185,69 @@ export const Tagger: React.FC<ITaggerProps> = ({ scenes, queue }) => {
     }
   }
 
+  function maybeRenderSearchAllButton() {
+    const searchButtons = document.querySelectorAll('.input-group-append .btn.btn-primary');
+  
+    if (searchButtons.length > 0) {
+      return (
+        <Button onClick={handleSearchAllClick}>
+          <FormattedMessage id="component_tagger.verb_search_all" />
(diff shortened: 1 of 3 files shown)


WithoutPants:
This plugin way of finding and clicking buttons via the DOM isn't an appropriate approach in the core React UI. The correct approach is likely going to involve lifting the selected result states out of `SceneSearchResults` and into `TaggerContext`, which will also be responsible for handling the save all and search all operations.

This won't be a trivial change unfortunately.

Gykes:
> This plugin way of finding and clicking buttons via the DOM isn't an appropriate approach in the core React UI. The correct approach is likely going to involve lifting the selected result states out of `SceneSearchResults` and into `TaggerContext`, which will also be responsible for handling the save all and search all operations.
> 
> This won't be a trivial change unfortunately.

I figured as much which is why I had you take a look before I went further. 

I have no idea how to implement it the way you are wanting but I'm all for giving it a shot.

WithoutPants:
I've made a start for prototyping purposes [here](https://github.com/WithoutPants/stash/tree/prototypes/save-all).

Issue #4609, opened 2024-02-22
[Bug Report] Grammar fix: Incorrect pluralization of "studios" string


echo6ix:
Located within the `studio-child-studios` div class, the string "studios" is always in the plural form. When the studio count is one, the string should read `studio`.

![image](https://github.com/stashapp/stash/assets/37937507/7ff16120-063d-4e09-8f20-cec90b5848f4)

![image](https://github.com/stashapp/stash/assets/37937507/503dadb5-2e69-40fa-929f-63eef6a53526)

PR #5161, opened 2024-08-23
[Bug Fix] Fix Studio Pluralization

Gykes:
Small bug fix so that if a studio only has 1 child studio then the correct pluralization is used.

[Before](https://gyazo.com/ec063fb3c3301e77a2bdd61027db8b79) vs [after](https://gyazo.com/ecd8034358835d8281a6ebd6d2830ce1)

closes https://github.com/stashapp/stash/issues/4609

### DIFF
--- a/ui/v2.5/src/components/Studios/StudioCard.tsx
+++ b/ui/v2.5/src/components/Studios/StudioCard.tsx
@@ -51,8 +51,6 @@ function maybeRenderParent(
 
 function maybeRenderChildren(studio: GQL.StudioDataFragment) {
   if (studio.child_studios.length > 0) {
-    const studioLabel =
-      studio.child_studios.length === 1 ? "studio" : "studios";
     return (
       <div className="studio-child-studios">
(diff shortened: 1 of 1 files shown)


ui/v2.5/src/components/Studios/StudioCard.tsx
```diff
@@ -51,14 +51,16 @@ function maybeRenderParent(
 
 function maybeRenderChildren(studio: GQL.StudioDataFragment) {
   if (studio.child_studios.length > 0) {
+    const studioLabel =
+      studio.child_studios.length === 1 ? "studio" : "studios";
```
WithoutPants, line 55: This needs to be internationalised. Take a look how it's done in [TagCard](https://github.com/stashapp/stash/blob/c74456c07e1cce1140cce8633882c31b521c2e9b/ui/v2.5/src/components/Tags/TagCard.tsx#L123)

WithoutPants (changes requested):

PR #5165, opened 2024-08-26
Persist tagger settings and change defaults

### LINKED ISSUE #3210: Scenes Tagger config - "Set Tags" Persistence
**Describe the bug**
The configuration of Scene Tagger, specifically tickbox "Set Tags" gets set to false after restart of stash not regarding to previous set value.

**Expected behavior**
Expecting the persistent setting which would last even for restart.
image

### LINKED ISSUE #4982: [Bug Report] Tagger Settings are per browser, hidden and with problematic defaults
**Describe the bug**
When tagging via tagger view, the defaults for showMales, and setTags are false, and not persisted on the server.

**Expected behavior**
Defaults should not change without warning / being obvious because you use a new browser. 

Stash Version: v0.26.1

**Additional context**
Because these are per browser (stored in IndexedDB) every time a new browser is used these are unset

The current defaults being false for both are thus a problem because it's not obvious that they have changed since you used another browser and they're hidden by default in a collapsed menu. 

If they are false, tagger may return no tags / males because 
a) they don't exist in the stash box entry
b) you haven't checked this option

If the default were set to true for both, this would prevent this issue, because, if the inverse settings were desired, the added tags and males being shown would make it obvious that you need to change the setting. This is probably the easiest fix, but others are suggested below.

The reason this is a problem is you might go and match a few hundred scenes, then have to go back and do them all again to get the data you thought you were getting the first time. 

Alternatives:
- Persist these settings server side
- Expand the tagger settings by default until manually collapsed by the user, the first time on each browser
- Make it more obvious that data is available (tags / males) that are not being added

### LINKED ISSUE #5065: Issue: 'Show Male Performers' Setting Not Retained Across Sessions
**Description:**

The 'Show Male Performers' setting in the scraping configuration should persist across application restarts, page reloads, and different browser sessions. Currently, this setting does not remain enabled when loading the application in a new session (e.g., in a different browser) or after a server/application reload.

**Steps to Reproduce:**

1. Enable the 'Show Male Performers' setting in the scraping configuration.
2. Reload the page, restart the application, or open the application in a different browser session.
3. Observe that the 'Show Male Performers' setting is not retained.

**Expected Behavior:**

The 'Show Male Performers' setting should remain enabled across sessions and after application restarts.

**Actual Behavior:**

The 'Show Male Performers' setting does not stay enabled when reloading the application or starting a new browser session.

**Screenshot:**

image

**Additional Notes:**

It can be frustrating to have to re-enable this setting repeatedly, especially when other settings retain the user's preferences. Initially, I wondered if this behavior was intentional, but I quickly realized it was likely an oversight. Given the dedication and quality of the development team, I'm confident this issue will be addressed in a future update.

Thank you for the fantastic app! As a queer user, my partner and I greatly appreciate your ongoing efforts and commitment to inclusivity.

WithoutPants:
Resolves #3210 
Resolves #4982 
Resolves #5065 

Moves the tagger settings from the local forage into the UI config. This makes the settings and fingerprint queue persistent between browsers and browser sessions. Existing local forage settings are _not_ moved.

Accordingly, `Show male performers` and `Set Tags` now default to true. 

A release note is included to indicate that settings have been reset and the new defaults.

### DIFF
--- a/ui/v2.5/src/components/Tagger/config.ts
+++ b/ui/v2.5/src/components/Tagger/config.ts
@@ -1,0 +1,20 @@
+import { useCallback, useContext } from "react";
+import { ConfigurationContext } from "src/hooks/Config";
+import { initialConfig, ITaggerConfig } from "./constants";
+import { useConfigureUISetting } from "src/core/StashService";
+
+export function useTaggerConfig() {
+  const { configuration: stashConfig } = useContext(ConfigurationContext);
+  const [saveUISetting] = useConfigureUISetting();
+
+  const config = stashConfig?.ui.taggerConfig ?? initialConfig;
+
+  const setConfig = useCallback(
+    (c: ITaggerConfig) => {
+      saveUISetting({ variables: { key: "taggerConfig", value: c } });
+    },
+    [saveUISetting]
+  );
+
+  return { config, setConfig };
+}
--- a/ui/v2.5/src/components/Tagger/constants.ts
+++ b/ui/v2.5/src/components/Tagger/constants.ts
@@ -11,7 +11,6 @@ export interface ITaggerSource {
   supportSceneFragment?: boolean;
 }
 
-export const LOCAL_FORAGE_KEY = "tagger";
 export const DEFAULT_BLACKLIST = [
   "\\sXXX\\s",
   "1080p",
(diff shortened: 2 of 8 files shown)


DogmaDragon:
Defaults are set correctly. 
Release note shown correctly.
Tagger settings persists across browsers (tested Vivaldi, Brave, Firefox) and clearing session cookies, including incognito mode.

PR #5166, opened 2024-08-26
Truncate scenes_o_dates and scenes_view_dates as part of anonymize

ikmckenz:
Seems like this sensitive information is not currently being anonymized. From discord "probably an oversight when O history and watch history was introduced". 

Separately, this file uses strings for table names, but most are actually defined in constants elsewhere. Can raise a separate PR to convert them to the defined constants for the whole file.


WithoutPants:
> Separately, this file uses strings for table names, but most are actually defined in constants elsewhere. Can raise a separate PR to convert them to the defined constants for the whole file.

Please do. :heart:

Issue #5035, opened 2024-07-02
[Bug Report] Disabling Scene Play History leaves history artifacts


ikmckenz:
**Describe the bug**
When disabling scene play history, artifacts from having it enabled are left behind and can not be updated. 

**To Reproduce**
1. Default behavior is with scene play history enabled
2. Watch a video
3. Disable scene play history
4. Any progress bars remain frozen in place, never to be updated, play history remains, etc. 

**Expected behavior**
I feel disabling scene play history should clean out these artifacts, to truly disable it, rather than just freeze them in place.

**Stash Version: (from Settings -> About):**
Current

randemgame:
The history of when you watched a scene is recorded as a play/o date, either automatically or manually or both. The wording being off is a by-product of how things developed - first there was no history (i.e dates) being recorded, then they were recorded in an automatic sense, then more recently users could add and fix these dates manually.

From the other thread I'd agree a rewording of 'Enable Scene Play History' could be something like 'Automatically Increment Play History in Stash'

Distinct from this history recording is the position in a scene where you stop watching, which like most other video applications is saved and restored so that you can continue watching from the same position next time. This can be toggled with 'Always start video from beginning'. I don't think there is any way of clearing the 'artifacts' this leaves in the app, you would need to change the resume_time to 0.0 for all the affected scenes in the database. 

I agree from the other thread that 'Enable Scene History' & 'Minimum Play Percent' would make more sense to group together in options, they are currently interrupted with setting the VR Tag in the settings. 'Always start video from beginning' could then be grouped with those two. (Hiding the Play/O buttons seems an entirely separate thing for user customisation though)

What is requested here I think is that toggling options we currently have in settings should clear out some of this data but I think it's a bit more involved and would need some dedicated extra settings. Which could either be-

A) Clear All History (All Individual Video Resume Times, All Individual Video Play Durations (minutes watched), All Play/O Dates (including their tally)

B) Distinguish the O/Play History and the Resume Times as separate things to clear. I don't think Play/O Date would make any sense to distinguish as they are often interconnected. The Play Duration by itself is not so distinct that it couldn't be grouped with the Play/O Date History. Here is a suggestion which would keep them as distinct things to clear-

SUGGESTION for SETTINGS:

=== History ====

Automatically Increment Play History in Stash..........Y/N (i.e Enable Scene History)
Minimum Play Percent for Play History........................X%  (added 'for Play History')
Always Start Video from Beginning................................Y/N (no change)
[...]

PR #5168, opened 2024-08-27
Create a section in the history panel to reset scene activity

ikmckenz:
This resolves the main concern behind #5035 in a way that fits with the existing workflow to manage o history and view history. Tested working locally.


pkg/sqlite/scene.go
```diff
@@ -1234,6 +1234,23 @@ func (qb *SceneStore) SaveActivity(ctx context.Context, id int, resumeTime *floa
...
+	}
+
+	record := goqu.Record{}
+
+	record["resume_time"] = 0.0
+	record["play_duration"] = 0.0
+
+	if err := qb.tableMgr.updateByID(ctx, id, record); err != nil {
+		return false, err
+	}
```
WithoutPants, line 1249: This can be simplified to:
```
if err := qb.tableMgr.updateByID(ctx, id, goqu.Record{
	"resume_time":   0.0,
	"play_duration": 0.0,
}); err != nil {
	return false, err
}```

Or alternatively it can be in the `updateByID` call.

WithoutPants (changes requested):
Thanks for this. For future reference, it's good to include screenshots of UI changes so that reviewers can assess without having to run a build themselves.

image

From what I can see, you've added a `Resume Time` heading with a trash icon button. This calls `sceneResetActivity` which sets both the `resume_time` _and_ `play_duration` fields to `0`.

I think this current design is a little confusing for the average user. The meaning of the `Resume Time` heading isn't clear, and there's no obvious indication that clicking the trash button will reset both the resume time _and_ the play duration. 

I think there's probably a use case where the user wants to reset the resume time _without_ resetting the play duration, so these functions should probably be split up. The existing `sceneResetActivity` interface can be kept, but I'd add boolean flags for resetting play duration and/or resume time.

Given that this roughly falls under the `Play History` category, I think these functions are better placed in the Play History dropdown button.

pkg/sqlite/scene.go
ikmckenz: After moving to boolean flags for resume/duration, the simplification is no longer possible.

ikmckenz:
Thank you for the review! 

As requested I have split the functionality into separate "reset resume" and "reset duration" actions, and moved them to the Play History dropdown button. On the backend I have kept the `sceneResetActivity` but added the booleans to separate the different activities as requested. The only thing is that I didn't simplify the `ResetActivity`  as originally requested, since with the booleans it works better copying the format of the function above it (`SaveActivity`). 

The proposed UI change now looks like this: 
Screenshot 2024-08-27 at 11 55 22 PM

WithoutPants (approved):
Thanks! Just made some minor changes:
- made the `reset_` arguments optional so that clients don't have to pass both
- added Toast popups after success or failure for the reset operations

ikmckenz:
Thank you!

PR #5170, opened 2024-08-27
Replace group image with more consistent svg

WithoutPants:
Replaces the existing default group image with a white svg icon more consistent with the other default images:
image

For reference, this is replacing the current default image below:
image


DogmaDragon:
I had an issue where one group that had attached scene retain it's old image even after clearing cache, but removing front/back cover manually properly replaced with it SVG. But I can't reproduce it. Tried 3 times with fresh configurations (migration from v0.26.2 into this build) and each time it worked properly now.

WithoutPants:
> I had an issue where one group that had attached scene retain it's old image even after clearing cache, but removing front/back cover manually properly replaced with it SVG. But I can't reproduce it. Tried 3 times with fresh configurations (migration from v0.26.2 into this build) and each time it worked properly now.

I believe this is probably a remnant of previous behaviour that would explicitly set the default image on movie instances.

PR #5172, opened 2024-08-27
Bump linux ffmpeg URL for latest version

WithoutPants:
Updates the linux ffmpeg URLs to 6.1.

### DIFF
--- a/pkg/ffmpeg/downloader.go
+++ b/pkg/ffmpeg/downloader.go
@@ -12,11 +12,11 @@ func GetFFmpegURL() []string {
 	case "linux":
 		switch runtime.GOARCH {
 		case "amd64":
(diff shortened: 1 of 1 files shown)


feederbox826:
7.0.2 is the latest version [ffmpeg.org](https://ffmpeg.org/download.html)

why not go straight to [JVS](https://johnvansickle.com/ffmpeg/) for the binaries if they're [downstream as well](https://github.com/ffbinaries/ffbinaries-prebuilt/releases/tag/v6.1)?

WithoutPants:
> 7.0.2 is the latest version [ffmpeg.org](https://ffmpeg.org/download.html)
> 
> why not go straight to [JVS](https://johnvansickle.com/ffmpeg/) for the binaries if they're [downstream as well](https://github.com/ffbinaries/ffbinaries-prebuilt/releases/tag/v6.1)?

I agree, but it will be a larger change than this. The downloader code assumes that the downloaded file is a zip file, and changing that code to handle xz and tar formats. That's something for a separate PR.

PR #5183, opened 2024-08-29
Make Scrubbers touchscreen capable

yoshnopa:
Pretty self explanatory, moving in the designated area of the touchscreen will now behave as if it is done by mouse. This also includes the clicking and leaving event. 
More technically speaking, I converted the according functions to work with touch as well as mouse events, and added the according eventlisteners.

### DIFF
--- a/ui/v2.5/src/components/Shared/HoverScrubber.tsx
+++ b/ui/v2.5/src/components/Shared/HoverScrubber.tsx
@@ -14,9 +14,21 @@ export const HoverScrubber: React.FC<IHoverScrubber> = ({
   setActiveIndex,
   onClick,
 }) => {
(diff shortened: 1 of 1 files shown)


yoshnopa:
I am open to discussion here regarding the opening of a selected scrubbed place via touch. Right now, it is relatively difficult, by tapping with a second finger while being on the correct position, but it could also be made default that leaving the scrubbing area is marked as "open on that position", which would make more false inputs, but may be more intuitive when purposfully applied.

WithoutPants (approved):
Really good change, thanks!

I'm happy to merge this as it is. I found dragging behaviour a little inconsistent on my device, but it's working well enough. I think as a future iteration we may want to increase the draggable area on smaller devices.

PR #5191, opened 2024-08-30
Gallery scrubber wall view

yoshnopa:
I like the new Gallery scrubber, so I wanted it to be present in the wall view as well.
The way I did that was to move the gallery info (Name, date) a bit upwards to make space for the scrubber, and put the linear gradient above both. I think it looks pretty good, and the functionality is as expected.
I also decided that in this case, opening the lightbox at the specific page instead of switching to the gallery is more in character of the wall view. 
This commit include #5183 , which is another small edit I did. You can revert that if it makes trouble, but I think doing these commits in order would be the better strategy.


WithoutPants (approved):
Nice work. I did some styling tweaks for when there is a mix of orientations in the gallery, and removed a `console.log` line that was copied from the source.

Issue #5197, opened 2024-09-02
[Bug Report] Disable "onwheel" parameter for performer page input forms using numerical values


echo6ix:
As discussed on [Discord](https://discord.com/channels/559159668438728723/559159910550732809/1279967738542166016_): Numerical input forms on the performer page can unintentionally get incremented or decremented when using a mouse wheel to scroll the pointer out of the input form. Unless a user specifically reviews the values they submitted after they save, any change in value from the scrolling will likely be unnoticed.

I was able to disable the behavior by adding `onwheel="return false;"` to the relevant input tags, though there may be a better approach

WithoutPants:
After further investigation and discussion, I have come to the conclusion that _some_ of the behaviour here is buggy, and this seems to be limited to Chrome-based browsers.

Scrolling to change the values in numeric input fields is (as far as I can determine) a well-established feature of browser form behaviour for numeric input fields. The expected behaviour however is that when scrolling the mouse wheel while a numeric field is focused _and_ hovered, it should mutate the field value and _not_ scroll the viewport. What I am seeing on Chrome however is that it's doing both. This presents a usability issue since, as echo6ix discovered, it's not obvious to the user that the value has been changed. On other browsers, the window is _not_ scrolled.

TLDR: using the mouse wheel to change numeric input values is expected when the field is focused and hovered, but it should not also be scrolling the window at the same time.

PR #5198, opened 2024-09-02
Disable onWheel for Text-input and TextArea

Gykes:
So, a weird one because I don't think think this is a bug but it's a valid thing to change due potential of data being wrong.

Adding it into TextArea as well just to keep it uniform, can remove if you don't think it's needed

fixes https://github.com/stashapp/stash/issues/5197


WithoutPants:
Closing in favour of #5199

Issue #789, opened 2020-09-07
[Bug Report] Input should call trimspace to get rid of any leading or trailing spaces from input


Tweeticoats:
**Describe the bug**
You can create a performer with a leading space ie " test".
Stash should run a function over all input to remove any leading or trailing spaces to clean the input ie call strings.TrimSpace().
This should happen with all input boxes and fields parsed by scrapers to clean these up before they make it to the database.

PR #5211, opened 2024-09-04
Removing White Space from InputFields.

Gykes:
I only added it in to `<InputFIeld>` Reason being is, if I'm not mistaken, the other fields aren't really sortable anyway so it shouldn't affect sorting.

People should still be able to add following and trailing spaces in things like description, Piercings, Tattoos, etc. Personally I see no reason to remove that unless there is something I am missing.

fixes https://github.com/stashapp/stash/issues/789

### DIFF
--- a/ui/v2.5/src/utils/form.tsx
+++ b/ui/v2.5/src/utils/form.tsx
@@ -113,6 +113,12 @@ export function formikUtils<V extends FormikValues>(
       value = "";
     }
 
+    // trimming leading and trailing spaces from inputs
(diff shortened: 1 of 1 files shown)


WithoutPants:
Trimming whitespace should be done in the backend, not in the UI.

echo6ix:
`Settings > Metadata Providers > Excluded tag patterns` is one area where leading and trailing whitespace is not being trimmed.

image

DogmaDragon:
Superseded by https://github.com/stashapp/stash/pull/6226

Issue #4997, opened 2024-06-20
[Bug Report] No white-space btw tag string and parenthetical tag string in tag select forms


echo6ix:
**Describe the bug**
Cosmetic bug identified in the tag drop down select forms on scene/image/performers/galleries pages. When typing in this form, there is no space between the tag string and the matched parenthetical aliases. This only affects parenthetical aliases that get matched to the query string.

TLDR:
- There is no space/padding between a tag and corresponding tag alias
- This cosmetic bug affects **_all_** tag input fields

**To Reproduce**
1. Go to any specific scene
2. Go to the tag drop down select form
3. Type a known tag alias
4. See error... the alias in parenthesis lacks white-space with the adjacent tag string 

**Expected behavior**
White-space between tag and parenthetical alias

**Screenshots**
![image](https://github.com/stashapp/stash/assets/37937507/d7fa2619-eaae-486d-919e-df1ed2b3166f)

**Stash Version: (from Settings -> About):**
 v0.26.1-22-ga7e5ccd0 - Official Build - 2024-06-19 20:13:58

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome 126.0.6478.62

PR #5218, opened 2024-09-04
Fix Tag and Alias odd spacing

Gykes:
As Echo6ix brough up the HTML engine doesn't generate whitespace at the beginning of a string. Modifying it to use `&nbsp;` so that the spacing will be correct.

fixes https://github.com/stashapp/stash/issues/4997


WithoutPants:
Can you please also fix this in the performer and studio select components?

Gykes:
Updated

Issue #4066, opened 2023-08-27
[Bug Report] Transcoding selector UI blocked by skip button on iPhone


anonstash:
**Describe the bug**
On an iPhone, the big buttons for playing/skipping block the transcoding selector UI. Only certain qualities can be selected as a result, and scrolling through the list is difficult.

**To Reproduce**
1. On an iPhone, open any scene in Stash in Safari/Firefox
2. Click on the gear to bring up the quality selection UI
3. Notice that the big skip button blocks selecting most transcoding qualities

**Expected behavior**
User should be able to select any quality available. This should be able to be fixed by having the quality selection pop-up box be in front of the big playback buttons.

**Screenshots**
![example](https://github.com/stashapp/stash/assets/143359807/1896230d-48ef-4468-a46c-319e82bb9da6)

**Stash Version: (from Settings -> About):**
v0.22.1

**Smartphone (please complete the following information):**
 - Device: iPhone 12 Pro
 - OS: iOS 16.6
 - Browser: Safari/Firefox
 - Version: N/A

PR #5229, opened 2024-09-06
Fix `source-selector` being blocked on mobile

Gykes:
Small CSS change to allow the `source-selector` to be brought to the front of the controls to allow people to select which source they would like. Added it into the section for mobile changes ( < 576px) as it's not really needed on anything bigger but can move it if told otherwise

At first, I was going more complex. I created functions that would disable the `big-buttons` if a menu was open and I got it mostly working but I felt like it was overkill when we can modify like 3-4 lines of CSS and get the same result.

fixes https://github.com/stashapp/stash/issues/4066

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/styles.scss
+++ b/ui/v2.5/src/components/ScenePlayer/styles.scss
@@ -360,6 +360,10 @@ $sceneTabWidth: 450px;
         height: 1.4em;
         width: 1.4em;
       }
+
+      .vjs-source-selector .vjs-menu {
+        z-index: 9999;
+      }
     }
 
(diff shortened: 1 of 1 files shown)


MinasukiHikimuna:
Could you change the title of the pull request? Small CSS changes doesn't tell anything about the actual change. Worst case is that the version history would be full of "Small CSS changes" commits and the reader would have to check every commit manually to find the correct one.

MinasukiHikimuna:
For example "Allow `source-selector` to be brought to the front" would be rather descriptive title. I took it from the commit description and shortened it a bit.

Gykes:
Thanks, DD :)

Issue #5233, opened 2024-09-07
[Bug Report] Usability issue with updated scrape with menu


cj12312021:
**Describe the bug**
This one isn't really a bug but a usability issue I was having. My muscle memory has me always reaching for the top item on the list when I want to use the StashDB scrapper (I imagine other users will run into the issue as well). However, with the recent change, a reload scrapper button, as well as a query filter, now exist with the reload button placed above the query filter. The placement makes my muscle memory think that the reload button is the first scrape option, leading me to always click that when attempting to use the StashDB scrapper.

Also see discussion on discord https://discordapp.com/channels/559159668438728723/559159910550732809/1282062156938084383

**To Reproduce**
Steps to reproduce the behavior ():

1. Ensure you have at least 5 scrappers installed to your Stash.
2. Enter the Scene Details page
3. Navigate to the `Edit` tab
4. Presh the `Scrape with...`

**Expected behavior**
A solution similar to the below implementation CSS (done with just CSS for ease of implementation outside of the source code) would be a better approach. It removes the reload button from the top and fixes it to the left of the text input field without the Reload Scrappers. The CSS also positions both those elements so they are stuck and stay on screen when scrolling through scrappers.

[https://github.com/user-attachments/assets/77a737b5-e5a3-49bb-b037-0262369b576a](https://github.com/user-attachments/assets/77a737b5-e5a3-49bb-b037-0262369b576a)

Since we'll have access to the DOM, some of the CSS tweaks can be bypassed by moving the reload button group where the reload button sits to the right of the text input element. This will ensure no conflicts with the clear text button when it appears. On the reload button, we would add a tooltip when hovering with the `Reload scrapers` text. This way, we can eliminate the label used directly next to the button. The refresh/reload icon here has a pretty universal meaning so there should be much confusion. As you can see in my CSS implementation, ideally, the search field would have its own color to make it more identifiable.

**Screenshots**
[If applicable, add screenshots to help explain your problem please ensure that your screenshots are SFW or at least appropriately censor

[https://github.com/user-attachments/assets/121d6200-9a26-4617-82e4-17c7896d2c0f](https://github.com/user-attachments/assets/121d6200-9a26-4617-82e4-17c7896d2c0f)

PR #5235, opened 2024-09-07
Updating Reload Scrapers formatting

Gykes:
Per convo with people on Discord. I have updated the Reload Scrapers UI. It now adds a button if the filter box appears and then the button extends and takes up the whole space if the filter box does not exist. Added in different `<div>`s as well so that it was easier to target for theming/plugins

The CSS is... janky at best. Not too great with it so opinions and potential fixes would be greatly appreciated. I figured using the default colors was best practice and people can theme them to be whatever they want.

With filter: 
filter

Without filter
no filter

closes https://github.com/stashapp/stash/issues/5233

### DIFF
--- a/ui/v2.5/src/components/Shared/styles.scss
+++ b/ui/v2.5/src/components/Shared/styles.scss
@@ -640,12 +640,9 @@ button.btn.favorite-button {
   display: inline-block;
 }
 
-.dropdown-menu:has(.scraper-filter-container) {
-  padding-top: 0;
-}
-
 .scraper-menu .dropdown-menu {
   min-width: 250px;
+  padding-top: 0;
 
   .dropdown-divider {
     border-top-color: $textfield-bg;
@@ -653,21 +650,21 @@ button.btn.favorite-button {
   }
 
   .scraper-filter-container {
-    background-color: $dark-gray5;
-    border-bottom: solid 1px rgba(16, 22, 26, 0.3);
+    background-color: $secondary;
+    border-bottom: solid 1px $textfield-bg;
     padding: 5px;
     position: sticky;
     top: 0;
     z-index: 1;
 
     .btn-group {
-      border: solid 1px rgba(16, 22, 26, 0.3);
+      border: solid 1px $textfield-bg;
       border-radius: 5px;
       width: 100%;
     }
 
     .clearable-text-field {
-      background-color: #26333d;
+      background-color: $textfield-bg;
     }
 
     .clearable-text-field-clear {
--- a/ui/v2.5/src/components/Shared/styles.scss
+++ b/ui/v2.5/src/components/Shared/styles.scss
@@ -154,6 +154,38 @@
   }
 }
 
+.filter-input {
+  border: 1px solid #ffffff20;
+  border-radius: 0.25rem;
+  color: #ffffff30;
+  max-width: 125px;
+}
+
+.refresh-button {
+  align-items: center;
+  background-color: #137cbd;
+  border-radius: 0.25rem;
+  color: #fff;
+  display: flex;
+  height: 37px;
+  justify-content: center;
+  min-width: 37px;
+  padding: 6px;
+}
+
+.dropdown-menu.show > .filter-container {
+  justify-content: center;
+  margin-left: 1px;
+  width: 180px;
+}
+
+.full-width-button {
+  align-items: center;
+  display: flex;
+  justify-content: center;
+  width: 100%;
+}
+
 .scrape-dialog {
   .modal-content .dialog-container {
     max-height: calc(100vh - 14rem);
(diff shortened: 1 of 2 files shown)


cj12312021 (commented):
I'll be pushing changes shortly.

cj12312021:
I've wrapped up my changes. I'm just waiting to push the assuming you are okay with me pushing directly to your branch. I did end up redoing much of what was there due to UI issues, which I couldn't pinpoint the source of. But the core concept we discussed yesterday is still intact. For your awareness, I left comments in the code to call your attention to changes that could have contributed to the issue I was seeing or changes that were not necessary. The hope is that this will help you avoid some of these pitfalls as you continue your work here.

The video below shows the main issues I was seeing. A decent portion of the drop-down menu fell outside of the edit tab, leading to those sections being cut off, and the button group at the top did not correctly fill the width of the menu. Then, there was the centering issue I mentioned with the Reload Scrapers label and the SVG it used.

https://github.com/user-attachments/assets/8ef522df-9f65-415a-b8ed-8b8397a23193

### The Changes
I've attached a video here showing the updates I made to the section. The abovementioned issues have been addressed. The button group now stays in view when scrolling through scrapers. There was a discussion yesterday about an appropriate style for the scrapper button to address the contrast issue "issue" I was seeing with that button and the Scrape/search buttons above it. Ultimately, I realized that a `Reload Scrapers` button already exists in Stash in the Settings > Metadata Providers > Scrapers section, so it was most appropriate to leave the button design the same in the source for consistency. However, I adjusted the search box style, which helped counter some of the visual issues I had with the combination.

https://github.com/user-attachments/assets/33605d3b-9b49-4683-bf76-b5c82d9e78e9

Screenshot 2024-09-08 131720

Gykes:
> I've wrapped up my changes. I'm just waiting to push the assuming you are okay with me pushing directly to your branch. I did end up redoing much of what was there due to UI issues, which I couldn't pinpoint the source of. But the core concept we discussed yesterday is still intact. For your awareness, I left comments in the code to call your attention to changes that could have contributed to the issue I was seeing or changes that were not necessary. The hope is that this will help you avoid some of these pitfalls as you continue your work here.
> 
> The video below shows the main issues I was seeing. A decent portion of the drop-down menu fell outside of the edit tab, leading to those sections being cut off, and the button group at the top did not correctly fill the width of the menu. Then, there was the centering issue I mentioned with the Reload Scrapers label and the SVG it used.
> 
>  Screen.Recording.2024-09-08.124456.mp4 
> ### The Changes
> I've attached a video here showing the updates I made to the section. The abovementioned issues have been addressed. The button group now stays in view when scrolling through scrapers. There was a discussion yesterday about an appropriate style for the scrapper button to address the contrast issue "issue" I was seeing with that button and the Scrape/search buttons above it. Ultimately, I realized that a `Reload Scrapers` button already exists in Stash in the Settings > Metadata Providers > Scrapers section, so it was most appropriate to leave the button design the same in the source for consistency. However, I adjusted the search box style, which helped counter some of the visual issues I had with the combination.
> 
>  Screen.Recording.2024-09-08.122954.mp4 
> Screenshot 2024-09-08 131720

Thanks for checking me out. I could have spent more time fully testing it but figured I would get it up and yall can tell me what's wrong... which is what happened. I'll give you write access here now once I figure it out. Thanks for the changes!

cj12312021:
No problem! Yeah, banging away at work yourself to learn can be insightful, but after enough time, it becomes more advantageous to get assistance if it is available. 

If you go to the settings page in your fork, at the top, there will be a collaborators option. Click that option and click the green Add people button in the center. You can then provide my GitHub user name in the text box that appears and just give with `Write access`

WithoutPants:
If we're using up the vertical space regardless, it probably doesn't make sense to hide the filter field.

cj12312021:
I'm in favor of that mostly because the full-width button isn't very appealing.

Gykes:
I agree, as well. I can modify it a little later. Have dinner plans tonight so it will be tomorrow that I can get to it. Feel free to do it though, CJ.

WithoutPants (approved):
Thanks for sorting this. 

I made some minor tweaks to the styling. The `:has` directive was no longer necessary if we always show the filter and that directive is a fairly recent addition so may not have full support.

I changed the colours to use the theme variables instead and changed the background colour of the filter input field to be `$textfield-bg` instead of a specific colour that I couldn't find elsewhere in the styling code. 

image

If that colour is actually necessary, then we should add it to the theme with an appropriate name - that can be a separate change though.

Everything else looks good.

ui/v2.5/src/components/Shared/styles.scss
```diff
@@ -640,34 +640,31 @@ button.btn.favorite-button {
...
 
     .btn-group {
-      border: solid 1px rgba(16, 22, 26, 0.3);
+      border: solid 1px $textfield-bg;
       border-radius: 5px;
       width: 100%;
     }
 
     .clearable-text-field {
-      background-color: #26333d;
```
cj12312021: Good catch. I picked up the color by using the color picker tool on one of the text fields on the page. At the time, it didn't cross my mind that the opacity applied to those fields would not be applied to the color code generated by the color picker.

PR #5253, opened 2024-09-11
Add portals for selects in dialogs

WithoutPants:
Fixes issue introduced with #5242 where showing select menus in dialogs would resize the dialog. Changed so that these mount to `document.body` instead.


echo6ix:
It's almost there, and useable but there is a glitch. Pay attention to when I access the tag dropdown selector repeatedly, and how by doing that it scrolls the content behind the modal.

screentogif

Issue #3553, opened 2023-03-16
[Feature] add updated_at to StashIDs


stg-annon:
Part of #2903

### **Is your feature request related to a problem? Please describe.**
As part of a system to support the ability of syncing stash-box data stash_ids would need an additional attribute to determine if the remote scene metadata has been updated since the entity was initially tagged or last updated from that endpoint

### **Describe the solution you'd like**
The addition of  the `updated_at` field to the stash id entities, have them populated, and add functionality to check, display and update scene data for tagged scenes

### **Describe alternatives you've considered**
I have used the local scenes `updated_at` field to attempt to do something similar with a plugin but this is limited in usefulness as any edit to the scene would be an update not just updates from the specific endpoint

WithoutPants:
$100 bounty placed (https://opencollective.com/stashapp/orders/636980)

ikmckenz:
I have this kinda half done in this branch: https://github.com/ikmckenz/stash/tree/updated-at

Trying to keep it only on the backend side and not the frontend, because adding the type to the StashID GraphQL type means a StashIDResolver interface gets generated which needs an UpdatedAt function, and I don't know if implementing all that is necessary for what should be a small change. 

Currently using OptionalTime as the type for the updated_at field instead of Time, so that it can be null. I figured that should keep things easy (?) and let the database column be null for the existing entries. 

Current state is that the updated_at is stored on identification, but then later after manual editing the updated_at disappears, and we get an error: 
```
sql: Scan error on column index 2, name "updated_at": unsupported Scan, storing driver.Value type <nil> into type *time.Time
```

WithoutPants:
@ikmckenz please submit an expense to the Open Collective (https://opencollective.com/stashapp/expenses/new) for $86.02 ($100 - fees), or indicate if you'd like to reassign the bounty elsewhere.

ikmckenz:
I would like to reassign to the general expenses fund you get paid out of, please.

Issue #2903, opened 2022-09-10
[RFC] Checking For Updates to Scene StashID Matches


halorrr:
### Problem
Once a scene has been matched, a user typically wouldn't be able to receive any updates if details have been updated (maybe males were added or titles were fixed), or if the scene was deleted as a duplicate. There is an update function for performers currently but I think it could be improved as well along with adding the scene behaviour, though this issue will focus on the scene side of things as the performer behaviour is likely similar if not the same and could later be extended.

I believe the solution for this lives on both Stash and Stash-Box sides.

### Stash Backend Side
Currently a `StashID` in Stash is made up of an `endpoint` and a `stash_id`. I believe another attribute should be added to it, to store the lasted `updated_at` time of the external resource. Not hard set on the name but this field could be something like `external_updated_at`.

### Stash-Box Backend Side
For the sake of making the process optimal Stash-Box should have a new query `findUpdatedScenes` that takes an array of `SceneUpdateInput` structured something like:
```JSON
{
  "scene_id": "stash_id",
  "last_recorded_update": "external_updated_at"
}
```
This call can return a `updated` Boolean, deleted Boolean, and `Scene` object for each input.

- If the `last_recorded_update` is NOT older than the `updated_at` on the `scene_id`, it will return `updated: false` , `deleted: false`, and no scene object
- If the `last_recorded_update` is older than the `updated_at` on the `scene_id` but the scene has been deleted, `updated: true`, `deleted: true`, and no scene object
- If the `last_recorded_update` is older than the `updated_at` on the `scene_id` and the scene is not deleted, `updated: true`, `deleted: false`, and the updated Scene object

### Stash Frontend Side
The scene tagger view will get a new `Batch Check for Updates` button. This will collect all the scenes the user wants to check (whether this is all scenes or just a sub-set) and call back to `Stash-Box` with `findUpdatedScenes`. 
- If the scene has no updates the scene item will show a message of `No new updates`
- If the scene has been deleted the scene item will show a message of `Scene has been deleted` with an external link to the ID so the user can investigate
- If the scene has been updated the scene item will show the returned scene object like it normally would on a scrape (it would be nice if we could highlight updates in a different colour, but that would be scope of another issue, since that would also generally improve any scraping, not just updates).

Buttons could also be added to `Show only updated` and `Show only deleted` for quick filtering of the list.

stg-annon:
[...]

PR #5259, opened 2024-09-14
Add updated_at field to stash_id's

ikmckenz:
Stored as a time (not OptionalTime), and so on migration set all existing StashIDs to have time zero, then update on scrape, tagger, etc.

First and biggest part of #3553 , only thing left would be displaying, and maybe a task to update things older than X.


ikmckenz:
Build broken for unrelated reason, fix is in #5260

ikmckenz:
Rebased on latest develop and bumped the schema version

graphql/schema/types/stash-box.graphql
```diff
@@ -13,11 +13,13 @@ input StashBoxInput {
 type StashID {
   endpoint: String!
   stash_id: String!
+  updated_at: Time!
 }
 
 input StashIDInput {
   endpoint: String!
   stash_id: String!
+  updated_at: Time!
```
WithoutPants, line 22: I think this needs to be optional to avoid a breaking change for clients. It should default to `now` if not provided.

internal/identify/scene.go
```diff
@@ -210,22 +211,19 @@ func (g sceneRelationships) stashIDs(ctx context.Context) ([]models.StashID, err
 
 	for i, stashID := range stashIDs {
 		if endpoint == stashID.Endpoint {
-			// if stashID is the same, then don't set
-			if stashID.StashID == *remoteSiteID {
-				return nil, nil
-			}
-
```
WithoutPants, line 217: I think this is incorrect. This will result in identified scenes with no changes being updated.

ikmckenz: There is maybe a use case for this, which is something like “re scrape scenes last updated x time ago”, which would be a feature I think people would want until the stash-box side of #2903 is implemented.

WithoutPants: I think that the logic should be that the `updated_at` field should _only_ be updated if there is other changes. The aim of the field is to be able to determine if the remote data has changed since we last grabbed it. Ideally, the date would be set to the updated date on the remote end, but failing that, it should only be refreshed if we actually detect a change.

PR #5281, opened 2024-09-21
New: Add "Select by Codec" to Duplicate Checker

y-e-r:
Added ability to select scenes in the duplicate checker based on preferred video codec. Added dropdown menu to select preferred video codec.

### DIFF
--- a/ui/v2.5/src/components/SceneDuplicateChecker/SceneDuplicateChecker.tsx
+++ b/ui/v2.5/src/components/SceneDuplicateChecker/SceneDuplicateChecker.tsx
@@ -48,6 +48,7 @@ import { objectTitle } from "src/core/files";
 const CLASSNAME = "duplicate-checker";
 
 const defaultDurationDiff = "1";
+const defaultSelectedCodec = "hevc";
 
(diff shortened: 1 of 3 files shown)


JanJastrow:
Maybe also `vp9`

y-e-r:
sure i can add more codecs, but i was also thinking about a more dynamic solution that would present a set of existing options from the search results.

Gykes:
So, I went ahead and fixed this up to have a modular system like you requested above. I wasn't too fond of having to go in and add new codecs. How this new system works is just displaying any codec for any file that is currently in the list. It displays the list alphabetically as I don't really feel like hardcoding HVEC or AV1 for every user was the best way to handle it. 

I also took out the all but smallest size option as I don't see that anywhere in your description. That leads me to believe it was either added by mistake or added and forgot. Either way I would prefer that one be broken out into it's own PR for tracking.

I will also say that the duplicate checker is old and, quite frankly, needs a redesign. This was a pretty small addition that I feel like could bring some good value for the amount of work was required

UI:
Screenshot 2026-09-02 at 02 24 15
Screenshot 2026-09-02 at 02 27 30

PR #5282, opened 2024-09-22
Fix migrations not using tx

WithoutPants:
Fixes `database is locked` errors during migrations.


magicswarmingswarm:
Still getting this error

Maista6969:
> Still getting this error

You're getting this error during a migration when upgrading? Which version are you upgrading to?

magicswarmingswarm:
Getting this once I have already upgraded to 0.27. Every third time or so I try get into 'scenes' it locks me out, showing that error in large text. Reloading does not fix. Only thing which fixes is moving to images, then back to scenes. After a while it will trip up and reset. No other messages in logs. It said successfully migrated back when I updated

Issue #1647, opened 2021-08-17
Navigating back in the browser doesn't remember scroll position


victorhooi:
ChromeOS Canary - 95.0.4609.3
OS - macOS 12.0 (Build 21A5304g) - although it also happens on Big Sur etc.

Steps to reproduce.
1. Navigate to a Stash page with enough results that you need to scroll.
2. Scroll to the bottom of the page.
3. Click on a video, to open it.
4. Navigate back in your browser.

What happens
The browser seems to automatically scroll to the top of the page every time - even if you were scrolled to the bottom prior to opening a video (or somewhere else in the page).

What should happen
The browser should remember your position, then when navigating back, go back to it.

halogod35:
I came here today to submit this bug. I'm leaving a comment here to confirm this bug on:

OS - iOS 14.7.1
Browser - Chrome v92.0.4515.90

OS - Windows 8.1
Browser - Chrome v92.0.4515.131

Also, +1'ing this issue.

WithoutPants:
This appears to be browser-specific. On Chrome I can reproduce this only on Performers, Movies, Galleries and Tags pages. On Edge however, I can reproduce this on all list pages.

DogmaDragon:
Can't reproduce on latest version. Create new issue if the bug is still reproducible on your end.

PR #5288, opened 2024-09-23
Remove unnecessary scroll to top on mount for top level query pages

WithoutPants:
Related to #1647 

The presence of `useScrollToTopOnMount` on all of the top-level pages meant that the page would scroll to the top when navigating back to these pages. I'm not sure what the purpose of the original inclusion was.

### DIFF
--- a/ui/v2.5/src/components/Galleries/Galleries.tsx
+++ b/ui/v2.5/src/components/Galleries/Galleries.tsx
@@ -5,7 +5,6 @@ import { useTitleProps } from "src/hooks/title";
 import Gallery from "./GalleryDetails/Gallery";
 import GalleryCreate from "./GalleryDetails/GalleryCreate";
 import { GalleryList } from "./GalleryList";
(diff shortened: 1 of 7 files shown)


secondfolder:
Is stash a Single Page App (SPA)? I'm assuming so and I'm assuming this was originally put in to deal with a limitation that SPAs often have/had. If you scroll 500px down page x and click a link to page y then page y will load but because it's just swapping out content on the DOM then you'll be 500px down page y. 

That's a very simplified case and in practice that behavior depends on a bunch of things such as page x having at least 500px that it's possible to scroll and that when loading page y the content gets immediately swapped out for content of that height, e.g. it doesn't go to a short loading screen then show all the content. Also these days browsers tend to be better at detecting that you're swapping between "pages" in a SPA and manually reset the scroll position as needed.

All this to say if the scroll position seems to work as expected in all major browsers without that hook then sounds like it is no longer needed but thought you might be interested in some of the history of why that's often used in SPAs (at least use to be).

PR #5311, opened 2024-09-28
Markers can have end time

### LINKED ISSUE #3147: [Feature] Optional end for markers to use them as clips
Allow the option to give an end to markers to use them as clips for example in [this case](https://github.com/stashapp/stash/issues/2511#issuecomment-1318682305).

MinasukiHikimuna:
Other metadata sources such as ThePornDB and timestamp.trade support end times for markers but Stash did not yet support saving those. This is a first step which only allows end time to be set either via API or via UI. Other aspects of Stash such as video player timeline are not yet updated to take end time into account.

- User can set end time when creating or editing markers in the UI or in the API.
- End time cannot be before start time. This is validated in the backend and for better UX also in the frontend.
- End time is shown in scene details view or markers wall view if present.
- Existing markers in the database will be updated to have -1 for end.
- GraphQL API does not require end_seconds. Omitted end_seconds will default to -1.

Closes #3147.


MinasukiHikimuna:
Added GraphQL documentation for seconds and end_seconds as recommended by stg-annon.

MinasukiHikimuna:
There were two linting errors that weren't for some reason detected when I run `make validate` or even more specifically `yarn run lint && yarn run check && yarn run format-check` in the `ui/v2.5`. The errors should be fixed now but the Github pipeline needs to be run again.

MinasukiHikimuna:
Forgot to run Prettier format after the lint fixes.

BonerFide:
Not strictly related, but ....

Is it possible to have these also be sub-second? Since they're already a float maybe for end and start we can just skip the  Math.round(getPlayerPosition()  ?

I often find a part of a scene etc starts/ends at a weird spot, or a marker will need that precision to avoid starting at an awkward point etc.

MinasukiHikimuna:
> Not strictly related, but ....
> 
> Is it possible to have these also be sub-second? Since they're already a float maybe for end and start we can just skip the Math.round(getPlayerPosition() ?
> 
> I often find a part of a scene etc starts/ends at a weird spot, or a marker will need that precision to avoid starting at an awkward point etc.

I agree that it would be useful but as you said, it is not strictly related. This is just a first step in improving markers so I would prefer to get this merged rather than including all kinds of related changes. For example the player could surely be changed somehow with the introduction of end times but that should be in a separate PR.

MinasukiHikimuna:
Update the pull request on top of current develop so the migration numbers match.

WithoutPants:
I've changed to use null values instead of `-1`. It's not great practice to use magic numbers like these to specify as null values if we can help it. I also changed the end label to `End Time` rather than `Time End` as it parses a bit better.

Otherwise, this is looking good. Thanks!

SyZeeGee:
Very neat feature!

Any chance that the scrubber can be easily updated to show the end time as well?  I can open as a new issue as well, just wanted to ask here first.

Quick mock (shows both scrubber and preview start / end range)
Marker_Start_End

PR #5315, opened 2024-09-29
Fix internal links not routed correctly in tagger view

### LINKED ISSUE #5006: [Bug Report] URL prefix missing in some links in tagger view
**Describe the bug**
Tagger view matched links (performers and studios) do not respect the X-Forwarded-Prefix header

**To Reproduce**
1. Run stash behind a reverse proxy with a URL Prefix (as documented [here](https://docs.stashapp.cc/networking/reverse-proxy/))
1. Match a scene in the tagger view and click on an matched performer or studio
1. links point to host.tld/performers/ID instead of host.tld/prefix/performers/ID

**Expected behavior**
You're redirected to host.tld/prefix/performers/ID

**Stash Version: (from Settings -> About):** v0.26.1

gujikabeepbop:
Fixes #5006


WithoutPants:
Thanks for your submisison. I've rewritten this in #5337.

PR #5327, opened 2024-10-01
fix: fix slice init length

huochexizhan:
The intention here should be to initialize a slice with a capacity of   len(u.Groups)  rather than initializing the length of this slice.

### DIFF
--- a/pkg/models/model_joins.go
+++ b/pkg/models/model_joins.go
@@ -33,7 +33,7 @@ func (u *UpdateGroupIDs) SceneMovieInputs() []SceneMovieInput {
 		return nil
 	}
 
-	ret := make([]SceneMovieInput, len(u.Groups))
+	ret := make([]SceneMovieInput, 0, len(u.Groups))
 	for _, id := range u.Groups {
 		ret = append(ret, id.SceneMovieInput())
 	}


Maista6969:
From the Golang docs on [make](https://pkg.go.dev/builtin#make):
> Slice: The size specifies the length. The capacity of the slice is
equal to its length. A second integer argument may be provided to
specify a different capacity; it must be no smaller than the
length. For example, make([]int, 0, 10) allocates an underlying array
of size 10 and returns a slice of length 0 and capacity 10 that is
backed by this underlying array.

This change is not necessary, and looking at your profile you're making this exact same type of PR to several other repositories which is kind of suspicious

huochexizhan:
> From the Golang docs on [make](https://pkg.go.dev/builtin#make):
> 
> > Slice: The size specifies the length. The capacity of the slice is
> > equal to its length. A second integer argument may be provided to
> > specify a different capacity; it must be no smaller than the
> > length. For example, make([]int, 0, 10) allocates an underlying array
> > of size 10 and returns a slice of length 0 and capacity 10 that is
> > backed by this underlying array.
> 
> This change is not necessary, and looking at your profile you're making this exact same type of PR to several other repositories which is kind of suspicious



> From the Golang docs on [make](https://pkg.go.dev/builtin#make):
> 
> > Slice: The size specifies the length. The capacity of the slice is
> > equal to its length. A second integer argument may be provided to
> > specify a different capacity; it must be no smaller than the
> > length. For example, make([]int, 0, 10) allocates an underlying array
> > of size 10 and returns a slice of length 0 and capacity 10 that is
> > backed by this underlying array.
> 
> This change is not necessary, and looking at your profile you're making this exact same type of PR to several other repositories which is kind of suspicious

I am collaborating to detect such issues in many warehouses, and our ultimate goal is to add relevant testing to Go vet.

These two methods are obviously different, you can refer to this example https://go.dev/play/p/vNUPNjdb2gJ

Maista6969:
I apologize for my hasty comment, your code example makes it very clear

huochexizhan:
> I apologize for my hasty comment, your code example makes it very clear

It's okay, thank you for the discussion.

PR #5335, opened 2024-10-03
Exclude null values from image studio id index

WithoutPants:
With the existing index, if no images have a studio id, then the index is not used when filtering by studio id. The assumption with this change is that most images don't have a studio id, so filtering by non-null studio id should be faster with this index. This is a tradeoff, as filtering by null studio id will be slower.

Fixes issue where studio performer counts were very slow when the database had a large number of images with no studios.

### DIFF
--- a/pkg/sqlite/database.go
+++ b/pkg/sqlite/database.go
@@ -34,7 +34,7 @@ const (
 	cacheSizeEnv = "STASH_SQLITE_CACHE_SIZE"
 )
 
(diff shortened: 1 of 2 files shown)


WithoutPants:
For reference, the performance issue was introduced in #4822 and appears to have been caused by the upgrade of `github.com/mattn/go-sqlite3` to v1.14.22. Testing with v1.14.17 without the index change results in much better performance.

PR #5355, opened 2024-10-08
Postgres support

NodudeWasTaken:
Hello, this is an attempt at implementing postgres support.
If you want to test this branch, you can add these changes to your config.yml:
```
database: postgresql://username:password@localhost/dbname
```

In the ui setup you must set the database path as the connection string to use this.
Also if you want anonymise to work a backup path has to be set

There are alot of changes, highlights:

- Seperate migrations for postgresql
- Abstractions around the database class
- Several SQL changes, particularly no lastinsertid and removal of the distinctids selector
lastInsertId is a SQLite specific feature, other databases need you to specify "returning id". Which our version of SQLite supports, but goqu doesnt mark it as supported, so we update to a dev version.
Call regexp as a function, as postgresql doesnt support named operators.
- Blob changes to make it cross-db compatible
- Postgres DLL extension to add missing functionality
Postgresql C/CGo extension to support missing functionality (basename, phash_distance, regexp):
https://github.com/NodudeWasTaken/stash_pge
- Migration script
https://github.com/NodudeWasTaken/stash_sqlite_to_pgsql

Some particularly egregious changes are:

- Lack of backups for postgresql
This will be the admins responsibility
- Lack of automated testing
We should probably add a postgres test
`PGSQL_TEST='postgresql://username:password@localhost/dbname_for_testing' go test -tags "sqlite_stat4 sqlite_math_functions integration" ./...`

Some particularly lackings are:
- automated testing on git
- that one bug that causes tests on postgres to sometimes fail
I added testmode which mitigated it a little, but it still happens sometimes.
It happens to file_test's for whatever reason.
- find the optimal number of concurrent read/writes
- it would be nice to warn if the dll isnt installed
- There is also a bug in the funscript speed generator that causes it to sometimes return a max value negative 64 bit integer, and i thought it was ridiculous to use a bigint for a value that's < 800, so i wrote a clamper into the migrator 
But i fear that it may cause insertion issues, maybe, since the thing that updates it doesnt clamp the value, and it unlike sqlite actually returns an error on overflow.
- Its slower than sqlite, resulting in many slow sql logs

To look at (help me):

- Number of concurrent reader/writers
- Warn if extension missing
- UI setup for postgresql
- Testing for pgsql
- Code cleanup, some of it is a little ugly

You can use dbeaver for testing the database, its nice for browsing various databases like sqlite and postgresql.
To support disable foreign keys we do (if anybody knows a better way, let me know):
"SET session_replication_role = replica;"
which requires extra postgres user permissions, just a note.

This is mainly a request for comments, or help, mainly help finishing this.

Fixes https://github.com/stashapp/stash/issues/3892


pkg/sqlite/performer.go
```diff
@@ -770,7 +770,7 @@ func (qb *PerformerStore) getPerformerSort(findFilter *models.FindFilterType) (s
 	}
 
 	// Whatever the sorting, always use name/id as a final sort
-	sortQuery += ", COALESCE(performers.name, performers.id) COLLATE NATURAL_CI ASC"
+	sortQuery += ", COALESCE(performers.name, cast(performers.id as text)) COLLATE NATURAL_CI ASC"
```
its-josh4, line 773: Nit, in Postgres it's more common to do type casting as `performers.id::text` rather than using the `cast` function

pkg/sqlite/repository.go
```diff
@@ -456,10 +461,10 @@ func idToIndexMap(ids []int) map[int]int {
 func (r *filesRepository) getMany(ctx context.Context, ids []int, primaryOnly bool) ([][]models.FileID, error) {
 	var primaryClause string
 	if primaryOnly {
-		primaryClause = " AND `primary` = 1"
+		primaryClause = fmt.Sprintf(" AND \"primary\" = %s", getDBBoolean(true))
```
its-josh4, line 464: Readability nit
```suggestion
		primaryClause = fmt.Sprintf(` AND "primary" = %s`, getDBBoolean(true))
```

pkg/sqlite/repository.go
```diff
@@ -456,10 +461,10 @@ func idToIndexMap(ids []int) map[int]int {
...
 	}
 
-	query := fmt.Sprintf("SELECT %s as id, file_id, `primary` from %s WHERE %[1]s IN %[3]s%s", r.idColumn, r.tableName, getInBinding(len(ids)), primaryClause)
+	query := fmt.Sprintf("SELECT %s as id, file_id, \"primary\" from %s WHERE %[1]s IN %[3]s%s", r.idColumn, r.tableName, getInBinding(len(ids)), primaryClause)
```
its-josh4, line 467: Readability nit

```suggestion
		primaryClause = fmt.Sprintf(` AND "primary" = %s`, getDBBoolean(true))
	}

	query := fmt.Sprintf(`SELECT %s as id, file_id, "primary" from %s WHERE %[1]s IN %[3]s%s`, r.idColumn, r.tableName, getInBinding(len(ids)), primaryClause)
```

pkg/sqlite/repository.go
```diff
@@ -500,7 +505,7 @@ func (r *filesRepository) getMany(ctx context.Context, ids []int, primaryOnly bo
 }
 
 func (r *filesRepository) get(ctx context.Context, id int) ([]models.FileID, error) {
-	query := fmt.Sprintf("SELECT file_id, `primary` from %s WHERE %s = ?", r.tableName, r.idColumn)
+	query := fmt.Sprintf("SELECT file_id, \"primary\" from %s WHERE %s = ?", r.tableName, r.idColumn)
```
its-josh4, line 508:
```suggestion
	query := fmt.Sprintf(`SELECT file_id, "primary" from %s WHERE %s = ?`, r.tableName, r.idColumn)
```

pkg/sqlite/repository.go
```diff
@@ -97,6 +98,10 @@ func (r *repository) runIdsQuery(ctx context.Context, query string, args []inter
 	for i, v := range result {
 		vsm[i] = v.Int
 	}
+
+	// We removed distinctIDs for postgresql, but now we have duplicates
+	vsm = sliceutil.AppendUniques(nil, vsm)
```
its-josh4, line 103: Using sliceutil.AppendUnique is quite inefficient from a time complexity (performs a full scan of the slice for each element to add). Since this method is in a hot path, consider using a map to make this a lot faster:

```go
vsm := make(map[int]struct{}, len(result))
for _, v := range result {
    vsm[v.Int] = struct{}
}

return maps.Keys(vsm)
```

pkg/sqlite/sql.go
```diff
@@ -16,11 +16,6 @@ func selectAll(tableName string) string {
 	return "SELECT " + idColumn + " FROM " + tableName + " "
 }
 
-func distinctIDs(qb *queryBuilder, tableName string) {
```
its-josh4, line 19: Why was this removed?

pkg/sqlite/scene_filter.go
```diff
@@ -455,12 +455,12 @@ func (qb *sceneFilterHandler) performerFavoriteCriterionHandler(performerfavorit
 			if *performerfavorite {
 				// contains at least one favorite
 				f.addLeftJoin("performers", "", "performers.id = performers_scenes.performer_id")
-				f.addWhere("performers.favorite = 1")
+				f.addWhere(fmt.Sprintf("performers.favorite = %s", getDBBoolean(true)))
```
its-josh4, line 458: Just as a nit, that using `fmt.Sprintf` for string concat is really bad for perf, it causes extra allocations and pressure on the GC
```suggestion
				f.addWhere("performers.favorite = " + getDBBoolean(true))
```

pkg/sqlite/database_postgres.go
```diff
@@ -0,0 +1,90 @@
...
+
+		if err != nil {
+			return nil, fmt.Errorf("conn.Exec(): %w", err)
+		}
+	}
+
+	return conn, nil
+}
+
+func (db *PostgresDB) Remove() error {
```
its-josh4, line 67: I would suggest that we implement a set of methods like `CanRemove() bool`, `CanReset() bool` so then the Stash UI can hide the buttons accordingly

pkg/sqlite/database_postgres.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type PostgresDB Database
+
+func NewPostgresDatabase(dbConnector string) *Database {
+	dialect = goqu.Dialect("postgres")
+
+	db := &PostgresDB{
+		storeRepository: newDatabase(),
+		lockChan:        make(chan struct{}, 1),
```
its-josh4, line 20: The reason we have `lockChan` is due to the fact that SQLite can't perform multiple queries in parallel. This is not a problem for Postgres and I'd strongly recommend not using locks in Stash to ensure a single query at the same time, or it would defeat most of the perf benefits of using Postgres.

In the implementation, this code: https://github.com/NodudeWasTaken/stash/blob/aaaad7916f73306fe6af4c49eea9d3c9a22f96a4/pkg/sqlite/database.go#L256-L270

Should just be a no-op when using Postgres (no locks acquired, nor released)

PS: I also think we should not use locks for SQLite too. Instead, we should open the SQLite connection with a `busy_timeout(5000)` parameter added to the query string, which causes SQLite to wait if the DB is busy. This will make the need for these locks redundant.

pkg/sqlite/database.go
```diff
@@ -81,19 +80,67 @@ type storeRepository struct {
...
 type Database struct {
 	*storeRepository
+	dbInterface
 
-	readDB  *sqlx.DB
-	writeDB *sqlx.DB
-	dbPath  string
+	readDB   *sqlx.DB
+	writeDB  *sqlx.DB
+	dbConfig interface{}
```
its-josh4, line 136: Looks like this is a `string` in both implementations, so could be reverted to type string?

pkg/sqlite/migrationsPostgres/1_initial.up.sql
```diff
@@ -0,0 +1,475 @@
...
+  containing_id integer not null,
+  sub_id integer not null,
+  order_index integer not null,
+  description varchar(255),
+  primary key (containing_id, sub_id),
+  foreign key (containing_id) references groups(id) on delete cascade,
+  foreign key (sub_id) references groups(id) on delete cascade,
+  check (containing_id != sub_id)
+);
+CREATE INDEX index_tags_on_name on tags (name);
```
its-josh4, line 420: I'd recommend putting the index creation in the same statement as CREATE TABLE (within the brackets), so it's a lot easier to track

pkg/sqlite/migrationsPostgres/1_initial.up.sql
```diff
@@ -0,0 +1,475 @@
...
+  updated_at timestamp not null, 
+  ignore_auto_tag boolean not null default FALSE, 
+  description text, 
+  image_blob varchar(255) 
+  REFERENCES blobs(checksum), 
+  favorite boolean not null default false
+);
+CREATE TABLE folders (
+  id serial not null primary key,
+  path varchar(255) NOT NULL,
```
its-josh4, line 19: IMPORTANT: In this migration file, please use `text` everywhere you're using `varchar(255)`. I see that you're using `varchar(255)` in many places including things where it's very much realistic to have longer strings (URLs, file paths).

This doesn't matter with SQLite because [SQLite doesn't enforce the length anyways](https://www.sqlite.org/faq.html):

 > 9) What is the maximum size of a VARCHAR in SQLite?

 >    SQLite does not enforce the length of a VARCHAR. You can declare a VARCHAR(10) and SQLite will be happy to>  store a 500-million character string there. And it will keep all 500-million characters intact. Your content is>  never truncated. SQLite understands the column type of "VARCHAR(N)" to be the same as "TEXT", regardless of the>  value of N. 

However, Postgres pretty much does.

`text` is a lot better here than varchar in all cases

pkg/sqlite/migrationsPostgres/1_initial.up.sql
```diff
@@ -0,0 +1,475 @@
+CREATE COLLATION IF NOT EXISTS NATURAL_CI (provider = icu, locale = 'en@colNumeric=yes');
```
its-josh4, line 1: One note that this requires Postgres to be compiled with ICU support, which I believe is optional. Could this be made optional maybe?

pkg/sqlite/migrationsPostgres/1_initial.up.sql
```diff
@@ -0,0 +1,475 @@
...
+CREATE TABLE IF NOT EXISTS images (
+  id serial not null primary key,
+  title varchar(255),
+  rating smallint,
+  studio_id integer,
+  o_counter smallint not null default 0,
+  organized boolean not null default FALSE,
+  created_at timestamp not null,
+  updated_at timestamp not null, 
+  date date, code text, photographer text, details text,
```
its-josh4, line 99: Nit: split to newlines

pkg/sqlite/migrationsPostgres/1_initial.up.sql
```diff
@@ -0,0 +1,475 @@
...
+);
+CREATE TABLE IF NOT EXISTS scenes_o_dates (
+  scene_id integer not null,
+  o_date timestamp not null,
+  foreign key(scene_id) references scenes(id) on delete CASCADE
+);
+CREATE TABLE performer_stash_ids (
+  performer_id integer,
+  endpoint varchar(255),
+  stash_id varchar(36),
```
its-josh4, line 203: Recommend using the `uuid` column type to store UUIDs, rather than strings

pkg/sqlite/migrationsPostgres/1_initial.up.sql
```diff
@@ -0,0 +1,475 @@
...
+  foreign key(performer_id) references performers(id) on delete CASCADE
+);
+CREATE TABLE studio_stash_ids (
+  studio_id integer,
+  endpoint varchar(255),
+  stash_id varchar(36),
```
its-josh4, line 209: `uuid` here too

pkg/sqlite/migrationsPostgres/1_initial.up.sql
```diff
@@ -0,0 +1,475 @@
...
+  image_id integer,
+  tag_id integer,
+  foreign key(image_id) references images(id) on delete CASCADE,
+  foreign key(tag_id) references tags(id) on delete CASCADE,
+  PRIMARY KEY(image_id, tag_id)
+);
+CREATE TABLE IF NOT EXISTS scene_stash_ids (
+  scene_id integer NOT NULL,
+  endpoint varchar(255) NOT NULL,
+  stash_id varchar(36) NOT NULL,
```
its-josh4, line 336: `uuid` here too

pkg/sqlite/migrationsPostgres/1_initial.up.sql
```diff
@@ -0,0 +1,475 @@
...
+    scene_id integer NOT NULL,
+    file_id integer NOT NULL,
+    "primary" boolean NOT NULL,
+    foreign key(scene_id) references scenes(id) on delete CASCADE,
+    foreign key(file_id) references files(id) on delete CASCADE,
+    PRIMARY KEY(scene_id, file_id)
+);
+CREATE TABLE IF NOT EXISTS performers_scenes (
+  performer_id integer,
+  scene_id integer,
```
its-josh4, line 281: Should these be non-nullable?

its-josh4 (commented):
Some initial review/feedback from me

internal/manager/config/config.go
```diff
@@ -50,7 +50,9 @@ const (
 
 	DefaultMaxSessionAge = 60 * 60 * 1 // 1 hours
 
-	Database = "database"
+	Database                 = "database"
+	DatabaseType             = "database_type"
```
its-josh4, line 54: Couldn't this be inferred from the database connection string?

If it starts with `postgres://` then we use postgres. Otherwise, if it starts with `sqlite://` or is just a local file, use SQLite

pkg/sqlite/sql.go
NodudeWasTaken: Using distinct in postgresql requires you to select any column you plan to use, like if you do "order by".
SQLite doesn't have this limitation, but that meant that i either had to rewrite alot of sql, or simply removed distinctIDs.

its-josh4: Ok, fair. We can add a TODO for the future if we want to optimize this

pkg/sqlite/database_postgres.go
```diff
@@ -0,0 +1,90 @@
...
+			return nil, fmt.Errorf("conn.Exec(): %w", err)
+		}
+	}
+	if !writable {
+		_, err = conn.Exec("SET default_transaction_read_only = ON;")
+
+		if err != nil {
+			return nil, fmt.Errorf("conn.Exec(): %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
NodudeWasTaken: I agree that this is ugly and should be removed.
But that would necessitate adding another connection string.
Im gonna have to think about it.

Or simply add "database_string_ro" 🤣

pkg/sqlite/database.go
NodudeWasTaken: The reason this was done was to hopefully allow for other db implementations, as i dont know what specific permissions they might want, so i changed it to interface.
Im not against changing it to string.

pkg/sqlite/database_postgres.go
NodudeWasTaken: Rewriting the sqlite locks is beyond the scope of this pr, but i have removed the locks for postgresql.

pkg/sqlite/repository.go
NodudeWasTaken: The key is that it must retain the order of elements here, so i dont think a map is sufficient.

pkg/sqlite/database_postgres.go
NodudeWasTaken: I have not yet determined the solution for how to backup in postgresql.
Suggestions are welcome, current suggestions are:
- Fetch everything and like anonymise does and dump it.
- Force the user to create a backup file called "backup_postgres_{date}.gz", to attempt to force some backup.

pkg/sqlite/migrationsPostgres/1_initial.up.sql
NodudeWasTaken: Grrrr sqlite, grrr 😡
Yeah i'll change this

pkg/sqlite/migrationsPostgres/1_initial.up.sql
NodudeWasTaken: You wouldn't have me... write MORE C LANG?! 😱😱😱
Yeah maybe i could, i'll have to look at it

pkg/sqlite/migrationsPostgres/1_initial.up.sql
NodudeWasTaken: This is how the original sqlite schema was written, and im not really in favor of diverting too much.

pkg/sqlite/migrationsPostgres/1_initial.up.sql
NodudeWasTaken: I dont want to divert from the original schema too much.

pkg/sqlite/migrationsPostgres/1_initial.up.sql
NodudeWasTaken: Ehh, they are written like that in the original schema, and they are foreign keys that reference other tables.
So i think its fine

internal/manager/init.go
```diff
@@ -38,13 +38,10 @@ func Initialize(cfg *config.Config, l *log.Logger) (*Manager, error) {
...
 	{
-		var dbType = sqlite.DatabaseType(strings.ToUpper(cfg.GetDatabaseType()))
+		var dbType = sqlite.SqliteBackend
 
-		switch dbType {
-		case sqlite.SqliteBackend, sqlite.PostgresBackend:
-			// Valid case
-		default:
-			dbType = sqlite.SqliteBackend
+		if strings.HasPrefix(strings.ToUpper(cfg.GetDatabaseConnectionString()), string(sqlite.PostgresBackend)) {
```
its-josh4, line 43: You may want to check for for prefix + `:` or this would match a file name of "postgres.sqlite" (unlikely, but possible)

```suggestion
		if strings.HasPrefix(strings.ToUpper(cfg.GetDatabaseConnectionString()), string(sqlite.PostgresBackend) + ":") {
```

pkg/sqlite/repository.go
its-josh4: Then how about this?

```go
found := make(map[int]struct{}, len(result))
vsm := make([]int, 0, len(result))
for _, v := range result {
    _, ok := found[v.Int]
    if !ok {
        found[v.Int] = struct{}
        vsm = append(vsm, v.Int)
    }
}

return vsm
```

internal/manager/init.go
NodudeWasTaken: GetDatabasePath is its own seperate function that returns the config entry "database" that has the database path (for sqlite). 
Where GetDatabaseConnectionString is "database_string" which has a connection string for pgsql.

pkg/sqlite/database_postgres.go
its-josh4: Well for now I think if the backend doesn't support it, just disabling the feature may make sense. That's why I was suggesting `CanRemove() bool`, `CanReset() bool`, etc... (or another option is to create an optional interface for `Remove() error` and then check if the object implements it at runtime)

As for backups, my .02 are that we shouldn't worry too much about backing up through Stash because we should let the DB admin do backups from the server.

The anonymization part is the most useful one. Maybe that could be done by creating a JSON dump?

pkg/sqlite/database.go
its-josh4: True, but in any case I think this could be a string. The underlying DB driver will decide if they want to parse it or not. It's hard to pass structured properties as config vars, and uncommon.

pkg/sqlite/migrationsPostgres/1_initial.up.sql
its-josh4: If this surprises you, wait till you also read this: https://www.sqlite.org/faq.html#q3 :D

pkg/sqlite/migrationsPostgres/1_initial.up.sql
its-josh4: LOL I would not suggest to put more C

I was just thinking if there was a way to make the natural collation optional. If it's not available in the server, then just fallback to a "regular" sorting (non-natural)?

pkg/sqlite/migrationsPostgres/1_initial.up.sql
its-josh4: It's mostly an optimization thing, since PG has a native uuid type while sqlite doesn't. Storing UUIDs in a varchar or text takes ~37 bytes (36 chars + 1 byte for size), while using uuid takes only 16 bytes. Not a biggie

internal/manager/init.go
its-josh4: maybe we could simplify this even further

- If the `database` config option begins with `postgres:`, then we use postgres and parse it as URL
- If it begins with `sqlite:` we use SQLite (and the path to file after, e.g. `sqlite:/path/to/file.db` or `sqlite:file.db`)
- If it doesn't have any prefix, then we assume it's a SQLite DB path (e.g. `file.db`) - this is for backwards-compatibility

Then we need a single config option only

its-josh4: https://github.com/NodudeWasTaken/stash/pull/1

pkg/sqlite/repository.go
its-josh4: I did some benchmarks and for small number of results, this seems to actually be more efficient than using a lookup map. Ignoring for now

pkg/sqlite/database_postgres.go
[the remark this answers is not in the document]
its-josh4: If you absolutely need to tell Stash to make this RO, you could add a query stirng parameter to the connection string.

Something like `?readonly=1` to the URL... and then we remove it

pkg/sqlite/sql.go
NodudeWasTaken: I added this back

pkg/sqlite/database_postgres.go
```diff
@@ -71,27 +71,51 @@ func (db *PostgresDB) open(disableForeignKeys bool, writable bool) (conn *sqlx.D
...
+	_, err = db.writeDB.Exec(getDBCloneCommand(backupPath, "stash"))
+	return err
 }
 
-func (db *PostgresDB) RestoreFromBackup(backupPath string) error {
-	logger.Warn("Postgres backend detected, ignoring RestoreFromBackup request")
-	return nil
+// RestoreFromBackup restores the database from a backup file at the given path.
+func (db *PostgresDB) RestoreFromBackup(backupPath string) (err error) {
+	sqlcmd := "DROP DATABASE stash;\n" + getDBCloneCommand("stash", backupPath)
```
its-josh4, line 112: Don't drop a database. Instead, drop all tables (and functions etc) within the DB. The reason is that dropping a DB, and creating new ones, requires certain privileges on a user. Most people will want to use a user with privileges for a single DB.

pkg/sqlite/database_postgres.go
```diff
@@ -71,27 +71,51 @@ func (db *PostgresDB) open(disableForeignKeys bool, writable bool) (conn *sqlx.D
...
+WHERE pg_stat_activity.datname = '%[2]s' AND pid <> pg_backend_pid();
+CREATE DATABASE %[1]s WITH TEMPLATE %[2]s;
+`, backupPath, dbname)
 }
 
+// Backup creates a backup of the database at the given path.
 func (db *PostgresDB) Backup(backupPath string) (err error) {
-	logger.Warn("Postgres backend detected, ignoring Backup request")
-	return nil
+	_, err = db.writeDB.Exec(getDBCloneCommand(backupPath, "stash"))
```
its-josh4, line 106: This assumes the database backup is on the local disk?

NodudeWasTaken: No, see DatabaseBackupPath

pkg/sqlite/database_postgres.go
NodudeWasTaken: I'll consider switching to remote scripting.

pkg/sqlite/migrationsPostgres/1_initial.up.sql
NodudeWasTaken: Im not gonna change this since it might complicate converting the migrations.

pkg/sqlite/database_postgres.go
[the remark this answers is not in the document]
NodudeWasTaken: Im gonna let a reviewer weigh-in on this

pkg/sqlite/migrationsPostgres/1_initial.up.sql
NodudeWasTaken: Postgresql 16 bundles ICU by default, so im gonna let it be a hard dependency

NodudeWasTaken:
Im cutting my losses on this and migrating my own setup back to sqlite.
If anyone wants this merged, feel free to revive it, but i no longer have time to develop it.

doneisenhower:
Hi,

Thanks for your significant effort on this, even though it hasn't been merged yet.
How much (roughly) additional work do you think would be required to get this production ready?

I'm not super experienced in Go, but I am sure either myself or someone else in the community would like to pick this up eventually.

NodudeWasTaken:
I did some work to clean up this branch into:
https://github.com/NodudeWasTaken/stash/tree/postgres-support-3
It now contains alot of duplicate code, because postgres is now seperate from the sqlite package.
I also introduced a "database" package that has alot of interfaces to try and keep an agreed upon interface for accessing the database.
I probably wont aim to merge this, but there is a chance i will keep it alive

Issue #4300, opened 2023-11-21
[Feature] Docker container overhaul with LSCR.io-like features


feederbox826:
**Is your feature request related to a problem? Please describe.**
- Running as root in the docker container can cause issues with permissions
  - also is a minor security concern #684 
  - while USER is an option, it breaks since stash looks under `$HOME/.stash` for config files
  - since it is not container-wide, ffmpeg might possibly be running as another user? #2797 
- re-installing new python libraries on every new update is very troublesome

**Describe the solution you'd like**
- A proper LSCR.io - like base for handling common environment variables like
  - TZ #1238
  - PUID/ PGID #684 
- Github Container Registry + DockerHub builds
- a proper DockerHub README
- officially supported CUDA/ QSV images (eventually ™️)
- an entrypoint script to install and update python dependencies **specified by the user on startup**
- a persistent volume for python dependencies
- a default `/config` directory instead of `/root/.stash`

**Describe alternatives you've considered**
N/A

**Additional context**
I want to work on this. Please assign me to this. I've done this for other projects and I am eager to finally do it for stash

mchangrh:
# container
- [ ] s6 init system
- [ ] runtime python dependencies
  - [ ] armhf support
- [ ] split /data /config directories
- [ ] migrations
  - [ ] migrations from hotio
  - [ ] migrations from stashapp/stash
- [ ] TZ ENV support
- [ ] rootless
- [ ] hardware acceleration
  - [ ] patched multi-stream nvenc 
  - [ ] custom ffmpeg implementations
    - [ ] jellyfin-ffmpeg 
    - [ ] lscr.io ffmpeg
    - [ ] rpi-ffmpeg
# CI/CD
- [ ] concurrent builds

Leopere:
The theory of why the changes need made is on par with what I’d want implemented. Privilege reduction proper installation of software requirements my biggest issue is that it needs to help users from before keep working ideally

feederbox826:
> it needs to help users from before keep working ideally

Yep, my plan right now is to test it in my own repository to iron out any bugs, and write mgirations for both hotio/stash and stashapp/stash so that users can migrate over seamleessly or as seamless as `mv && chown -R` can get

stg-annon:
another problem people have with docker installs is python packages, might be worth taking a look at setting up a venv so that plugins and scrapers that require other packages don't need to reinstall them on every reboot

alternately we may be able to get away with just setting a couple environment variables to a mount from the host
[...]

Issue #305, opened 2020-01-03
[Feature] hardware encoding


binarygeek119:
i would love for nvidia NVenc for tarnscoding and Generation 
this would also work with amd and intel encoders 
this could speed up the Generation  process

bnkai:
For the file transcoding we use x264 with the faster preset and thats probably the only place it might be quicker with nvenc (not sure about quality) BUT file transcoding imho is not that needed anymore since we now have live stream transcoding for unsupported files.( IMHO transcodes in the generated content section should be left unticked in 99% of the cases )

For live transcoding we have vp9 webm files that are not supported through hardware encoding except through intel vaapi i think and not sure about the stability / quality of that also.

Finally for the generated previews and markers we use x264 with preset veryslow to get the highest quality , since they are only generated once but viewed many times. If you wanted to make the generate faster thats where maybe we could opt to change the veryslow preset to medium or even fast and still get better quality/performance than hardware encoders. Thats ofcourse only for anyone that is willing to compromise the quality  for speed and only as an extra selection not as the default.

HASJ:
The only way live streaming would even remotely be viable here is by hardware acceleration. Software-bound encoding is a no-go.
VP9 is even worse. I am using a FX-6300. It was not optimized for these tasks, to put it kindly

CenterThrowaway:
I'd add that after Pascal on the Nvidia side, hardware encoding with their GPUs is leaps and bounds better, comparable to CPU based x264 up to Medium presets I believe, while being much faster.

praul:
I too would be very pleased about this feature. It does not have to be as user friendly as for example jellyfin with its hardware encoding support. It could be an advanced setting to add parameters for ffmpeg (playback/live transcoding or preview generation). If it causes problems, users could just set it back to default. But advanced users would be able to fiddle with it a litte more.

It's quite easy to pass vaapi support to docker containers, and hardware encoding would greatly benefit my high cpu loads.

r3538987:
Can someone share command-line which is being used when generating such video previews.
Only thing I see at the moment is this when generation fails on WMVs sometimes.

[...]

PR #5366, opened 2024-10-12
chore(docker): switch to ubuntu base image

kmou424:
1. install intel non-free driver for QSV support
2. use jellyfin-ffmpeg6 for best hardware transcoding
3. provide optional google-chrome installation

disadvantages: larger image size

### DIFF
--- a/docker/build/x86_64/Dockerfile
+++ b/docker/build/x86_64/Dockerfile
@@ -16,8 +16,8 @@ ARG STASH_VERSION
 RUN BUILD_DATE=$(date +"%Y-%m-%d %H:%M:%S") make ui
 
 # Build Backend
-FROM golang:1.22-alpine as backend
-RUN apk add --no-cache make alpine-sdk
+FROM golang:1.22-bullseye as backend
+RUN apt update && apt install -y build-essential golang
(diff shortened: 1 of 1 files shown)


kmou424:
#305

feederbox826:
Already has a [HWAccel image based on CUDA](https://github.com/stashapp/stash/blob/develop/docker/build/x86_64/Dockerfile-CUDA)

Adding google-chrome instead of headless-cdp adds an extreme amount of unnecessary bloat

https://github.com/stashapp/stash/issues/4300

`i965-va-driver-shaders` and `intel-opencl-icd` are required for ARC and tone mapping https://github.com/feederbox826/stash-s6/blob/main/stash/root/defaults/intel-drivers.sh

kmou424:
> https://github.com/nerethos/docker-stash and https://github.com/feederbox826/stash-s6 already both use jf-6, and I have a beta build with jf7

sure, your repo `stash-s6` provided a such better solution. Because of I have switched to your branch 2 weeks ago. But I'm confused about why these changes can't be merged to main branch. In addition, I think hardware acceleration should default supported by official image, rather than manually build it when user want to use HWAccel.

PR #5391, opened 2024-10-18
Upgrade gqlgenc and regenerate stash-box client

WithoutPants:
Fixes error when generating stash-box graphql client. Not tested.


WithoutPants:
Includes a necessary bump to the compiler since the new dependencies require a more recent go version.

Issue #3738, opened 2023-05-17
[Feature] Added Funscript changes "Updated At" date


nerfdaderp:
**Is your feature request related to a problem? Please describe.**
If you add a funscript file to a scene already in your library, the date fields for the record do not update, which on a larger library, can make it annoying to track down at times.

**Describe the solution you'd like**
When a funscript file is added to a scene (or even updated), or metadata is reported as changed in logging, have the Modified Date for the scene be adjusted accordingly. This way views built off of and sorted by this field make the updated content easy to find.

**Describe alternatives you've considered**
Remembering the names of the actresses in question, and filtering in Heresphere accordingly.

**Additional context**

randemgame:
Yes, I was about to request the same feature.
When I run the Generate task on a large library it knows to create the heatmap.png and connect it to the scene as well as updating File Info in the Scene details. This should trigger the 'Updated At' value to update. 
On Eroscripts I keep finding old classic scenes updated with free scripts from the community but they are very difficult to keep track of now (a time-consuming hack I'm using is to filter by Interactive, and Sort by Updated At values in reverse)

PR #5401, opened 2024-10-19
Update Scenes' 'Updated At' Date on Heatmap Gen & Expand Error Log

randemgame:
The "Updated At" field of a Scene will now update when an associated Heatmap is created via the Generate task..

I am not sure whether a Scene's 'Update At' would be better updated:
- when the 'File Info' is updated with some generated funscript data during the Scan task 
- when the heatmap is attached to a scene during the Generate task
at the moment it is neither. perhaps it could be both? but it should at least be one.

see this long-standing issue https://github.com/stashapp/stash/issues/3738 about why this functionality is needed, which this PR will close. 

If necessary I can work to move this functionality from the Generate to the Scan task. I attempted that at first, tracing where a Scene's 'Update At' property could be modified as a result of the scan.go message about 'Updating Metadata'. I thought that the best trigger for this may be when a Scene's 'Interactive' property is modified there, but I could not work it out.

I was very surprised by the simplicity of the solution in this PR but I am not a go developer so blind to any issues or offenses I could have included here. I am not sure if the error messaging in the code I added is necessary or not and I was told 'I settled on a simplified error messaging due to this function not being so critical', but I really have no idea what I'm doing.

---

Sorry to include a second thing, but I happened to notice that this file also included the same error message which has been frustrating me for a long time. In my logs after any Generate task I would keep seeing the following, filling up 40 or more lines of text:
"error generating heatmap: json: cannot unmarshal number into Go struct field Script.version of type string"
including the Scene Path in this error message has made it possible for me to identify those problem scripts lost in my library.

### DIFF
--- a/internal/manager/task_generate_interactive_heatmap_speed.go
+++ b/internal/manager/task_generate_interactive_heatmap_speed.go
@@ -3,7 +3,6 @@ package manager
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/stashapp/stash/pkg/file/video"
 	"github.com/stashapp/stash/pkg/fsutil"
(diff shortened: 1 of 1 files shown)


WithoutPants:
#3738 concerns updating the `UpdatedAt` field when a funscript is added to a file during scan. The solves part of the problem, but updating the scene when the file changes to add a funscript is probably not a simple or necessarily desirable change. It _should_ update the _file_ `UpdatedAt` field, and it should be easier to find updated files, but it's an entirely different change.

PR #5431, opened 2024-11-01
Sub second marker timestamp precision

### LINKED ISSUE #5382: [Feature] Add millisecond (or frame) precision to markers
It's difficult to get markers to start / end at the right time when they only have second precision, sometimes the half second before is too early, and the half second after too late. 

Markers should be able to be created using the interface from the exact current position of the player with ms / or frame precision. 

The loop plugin could also be used to set/create/be set by markers.

**Additional context**
The database already stores markers as floats, so the database should support this, it just needs the interface to allow that precision.
This pull request adds end time, which should also have the same interface / precision change.
https://github.com/stashapp/stash/pull/5311

WithoutPants:
Resolves #5382 

Allows specifying marker times by frame within seconds:

image
image

Frames are specified with a trailing `.ffff` which is the 0-based frame number. 

Times with a fractional component are formatted as `[hh:]mm:ss.ffff` - frame number 12 at time `3:45` is displayed as `3:45:0012`. The time field accepts the same value without the leading zeroes as well, ie `3:45.12`, but will be re-formatted to the other format.

Times at frame 0 are accepted and displayed without the `.ffff` suffix.

Times from the current frame of the video will include sub-second precision.

The length of the `ffff` field was an arbitrary choice and I'm happy to hear alternatives.

This PR will be rebased once #5311 is merged.

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
+++ b/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
@@ -297,9 +297,13 @@ export const ScenePlayer: React.FC<IScenePlayerProps> = ({
     sendSetTimestamp((value: number) => {
       const player = getPlayer();
       if (player && value >= 0) {
-        player.play()?.then(() => {
+        if (player.hasStarted() && player.paused()) {
           player.currentTime(value);
-        });
+        } else {
+          player.play()?.then(() => {
+            player.currentTime(value);
+          });
+        }
       }
     });
   }, [sendSetTimestamp, getPlayer]);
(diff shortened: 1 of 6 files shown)


feederbox826:
I personally don't prefer frames in SMPTE, I much prefer `HH:MM:SS.ms` since you

1. avoid frame -> ms conversion, also dealing with DF/NDF (and video.currentTime is .ms)
2. are able to get sub-frame precision easily
3. .ms remains consistent across framerates and conversions
4. makes semantic sense if finely adjusting it, since DF wraparound would be at 29.97 and jump to 00

If you do want to stick to SMPTE, then I suggest sticking with HH:MM:SS:FFFF to not confuse it with .ms

WithoutPants:
Thanks for the feedback. I've updated to use `hh:mm:ss.ms` instead. It simplified the code doing it that way anyway.

PR #5437, opened 2024-11-02
Setting marker end time with clock button uses full precision

MinasukiHikimuna:
This is similar to how start time setting already worked.


WithoutPants:
Whoops, yeah that should've been picked up when I rebased it. Thanks!

PR #5443, opened 2024-11-03
Scene Marker grid view

dogwithakeyboard:
Adds a Marker grid view and Marker grid card with the aim of improving the bulk management of Markers.

- Adds a bulk delete mutation to delete multiple markers at once from the UI.
- Uses the same preview as the scene cards with a static image then video on mouse over. The generation options have been changed to reflect this new use.
- Clicking the Scene title/filename in the card body builds a filter with all Markers for that Scene.
- Clicking the Performer pill in the popover builds a filter for all Markers with that Performer.
- Clicking the Tag pill in the popover builds a filter for all Markers with that Tag.

markers


WithoutPants:
Nice work. I'm happy with the presentation for this as an initial version. We can always refine further down the track. Only needed a minor refactor.

PR #5471, opened 2024-11-11
Line break titles

randemgame:
Added support for line breaks in Scene Titles.
Affects how titles can be now displayed in Scene Cards and Main Scene View. Should not affect any other view (added code to fix it affecting Performer Cards)
Allows for a greater level of customisation and flexibility with the presentation of text titles so that they can match the rest of your Stash aesthetic, rather than relying on how a title is automatically spread to a second row.

- I do not believe this was possible via a plug-in alone.
- As always, apologies for any problematic code - I have tested the implementation on my machine and it is functional and passed tests but sorry if the implementation is whack

https://discord.com/channels/559159668438728723/644934273459290145/1305598180477370479


cj12312021:
The front end already supports what you are trying to do here. If you perform a GraphQL scene update with a title like `One\nTwo\nThree`. The line breaks will render without our change. The issue you are likely running into is with inputting those new lines in the input box presented in the UI. Changing the form input to a text area would enable the entry of new lines, but I'm not sure that is an ideal change.

randemgame:
@cj12312021 ah, how embarrassing, i have hugely reduced the changes in this issue then to just allow for a line break when the enter key is pressed in the edit field, but I couldn't work out a good way to give visual feedback on this (i tried literally adding a /n or enter key emoji to represent it but couldn't do this without it reflecting in the literal title). If it is better to be a plugin then ok, but if the title field does allow for line breaks and the edit field does not allow for line breaks, perhaps it should be part of core functionality. I dunno! I'll let you all decide

Gykes:
Probably good as a plugin.

PR #5472, opened 2024-11-11
Scene Marker duration filter and sort

dogwithakeyboard:
Adds the ability to sort and filter Markers by their duration (the difference between `seconds` and `end_seconds`). Markers without an end point have zero duration so a 0:00 duration filter can be used to filter for them.


WithoutPants:
Thanks for this. I changed the filter type to be a nullable float filter, since imo a marker without an end point should have a null duration, rather than a duration of 0. The other duration filter fields should probably be changed to be floats as well, but that's a problem for a separate PR.

PR #5526, opened 2024-12-02
Filter scenes & images by filename/basename

damontecres:
Adds a new `filename` filter to images and scenes which filters using the basename of the file's path.

I'm unsure of two things though:
1. Is `filename` descriptive enough?
2. Is using `component_tagger.config.query_mode_filename` for the message ID okay for this?


Gykes:
So, there are some backend changes that will be coming at some point. I don't have an ETA but I think that this new solution would be a better way of handling it than this. 

Closing for now.

PR #5531, opened 2024-12-04
Add Sort Name to Tags

### LINKED ISSUE #3108: [Feature] Sort_Name for Tags
This issue is for Tags specifically, but this concept could be applied to other types of objects as well.

### **Is your feature request related to a problem? Please describe.**
sorting tags by name there is often an order that would be preferable but does not work with a standard alphanumeric sort without altering the name of the tag

### **Describe the solution you'd like**
the addition of a `sort_name` attribute to Tags that when present would be used over `name` when sorting tags 

example:
```sql
SELECT COALESCE(sort_name, name) as sname ORDER BY sname
```

### **Describe alternatives you've considered**
I often add a prefix that changes the sort order of the tags, while useful this makes the tag ugly where it is displayed in the UI this also changes the main name for the tag which can make scraping difficult

### **Implementation**
Ideally the sort_name would also be available in the DOM so that userscripts might take advantage of it.  Perhaps as an addental attribute in the `<a>` element like `<a href="link"  data-tag-sort-name="sort_name">Tag Name<a>`

### **Additional context**
As has been mentioned elsewhere this could also be used as an ad-hoc way of a user adding namespaces (#1253) to tags.

For example:
a sort name for a Performer tag  could be `Performer - Tall` and with relatively minimal effort, tags being added in a performer context could be filtered by `{"sort_name": {"modifier":"MATCHES_REGEX", "value": "^Performer"}}`, which could be defined in a config by the user to filter tags in such a context, allowing for namespaces in a way which preserves most of the current functionality of tags

stg-annon:
adds a sort name to tags 
Resolves #3108


DogmaDragon:
Lightly tested, no more issues found.

WithoutPants (approved):
Thanks! This looks good. I made the following changes:
- Changed the display name to be `Sort Name` instead of `Sorting Name`
- Added new field to anonymiser 
- Added new field to import/export

PR #5542, opened 2024-12-08
Use Marker endSeconds value when generating previews

stg-annon:
tested with a few different combinations of Markers, appears to be working properly


WeedLordVegeta420:
This seems like a good addition, but I wonder if this may necessitate the need for an optional max preview length setting.

WithoutPants:
> This seems like a good addition, but I wonder if this may necessitate the need for an optional max preview length setting. I can definitely see some people creating markers that span a large portion of a video and inadvertently end up generating a massive preview that isn't really going to be all that useful.

Agreed. I've made a change to limit the preview duration to a maximum of the old default length - 20 seconds. In future, we might consider making the marker preview use the same logic as the scene preview generation.

WithoutPants (approved):

PR #5582, opened 2024-12-30
update python and ffmpeg version in docker cuda

dongfengweixiao:
1. update python version to 3.12(the version in the official repository is 3.10, which encountered errors while executing some tasks)
2. update ffmpeg version to 6.0(the version in the official repository is 4. x, which does not meet the running requirements of stash. The downloaded ffmpeg does not support Nvidia's CUDA.)
3. preinstall cloudscraper bs4 stashapp tools lxml using pip

Other suggestions:

Adding the cuda version to cuda build suffix in the makefile is crucial.
cuda requires that the version in the container is not higher than the version in the host computer.

### DIFF
--- a/docker/build/x86_64/Dockerfile-CUDA
+++ b/docker/build/x86_64/Dockerfile-CUDA
@@ -1,7 +1,7 @@
 # This dockerfile should be built with `make docker-cuda-build` from the stash root.
 
 # Build Frontend
-FROM node:alpine as frontend
+FROM node:alpine AS frontend
 RUN apk add --no-cache make git
 ## cache node_modules separately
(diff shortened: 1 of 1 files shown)


SyZeeGee:
Does this reduce the need for transcoding in some cases?

feederbox826:
Python3.13 should be used, and you don't address the CUDA upgrade arg

you are also statically linking ffmpeg when the one that is downloaded by stash already supports nvenc/ CUDA https://johnvansickle.com/ffmpeg/

feederbox826:
> It appears unnecessary to upgrade to Python 3.13 since all plugins function well with Python 3.12. Could you provide any links that specifically require Python 3.13?

CommunityScrapers targets latest (3.13)

PR #5597, opened 2025-01-07
Stash Docker Installer

David-Maisonave:
CreateContainer.cmd is a DOS script that is used to create a Stash container on Docker.
- Installs Stash container with following options:
  - Mount paths Read-Only
  - Mount paths read-write
  - Install both image and container
  - DLNA option
  - Stash port number option
 
See readme for more details. (https://github.com/David-Maisonave/Axter-Stash/tree/main/Docker)

This batch script is for Windows, but if there's interest, I'll add a Linux script as well.


WithoutPants:
This would be better suited for [CommunityScripts](https://github.com/stashapp/CommunityScripts), under `scripts`.

PR #5608, opened 2025-01-14
Scene Gender Filter

Gykes:
Basic attempt at adding in Gender filtering on scenes.

Still a very rough draft.


bob123491234:
The ability to filter scenes by performer gender was already added in #4861, it just needs to be added to the UI.

PR #5624, opened 2025-01-25
Added f4v to default video extensions

RampantDespair:
Don't ask me how I know this, but Naughty America serves older scenes as `.f4v` files.


DogmaDragon:
Tested on Gecko and Blink engines, both are able to play it natively. Not tested on WebKit.

PR #5637, opened 2025-01-30
Update postprocessing.go to group 'x.com' with 'twitter.com' domains

L3GW0RK:
added the 'x.com' domain to the twitter handle recognition in the performer window. cleans things up a bit with urls. tested, working.

### DIFF
--- a/pkg/scraper/postprocessing.go
+++ b/pkg/scraper/postprocessing.go
@@ -2,6 +2,7 @@ package scraper
 
 import (
 	"context"
+	"strings"
 
 	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/match"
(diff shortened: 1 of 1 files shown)


WithoutPants:
It's not really clear to me what this is fixing.

The code you changed transforms twitter profile names; if the input string is `xxx`, then it currently becomes `https://twitter.com/xxx`. If the input string is already a URL, then it is left unchanged. 

From my static code inspection, I believe that the only different behaviour is if the input field _starts with_ `x.com`, in which case the call to `URLFromHandle` is changed to provide `https://x.com` as the base URL instead. The only way that this would be triggered would be if the input field is `x.com/xxx`, but the result would be `https://x.com/x.com/xxx`.

WithoutPants:
Closing as stale

Issue #5644, opened 2025-02-03
[Feature] Preserve JAV title formatting in Scene Tagger


javstash:
**Is your feature request related to a problem? Please describe.**
Currently, stash remove all hyphens (except those in date format) from the query in Tagger Mode.  When you have a file named 'XXX-1111.mp4' or if your scene title is 'XXX-1111', the default query string in the scene tagger will be 'xxx 1111' which might return search results that are not intended.

**Describe the solution you'd like**
Expand the `ParseDate()` function in `src/components/Tagger/utils.ts` to take a JAV code regex `JavCodeRegex = '.*?([a-zA-Z|tT28|tT38]+)-?(\d+)[zZ]?[eE]?(?:-pt)?(\d{1,2})?.*'` so that the hyphen in the JAV code is preserved.

**Additional context**
This will mirror the behavior of stash-box, which allow hyphens in the title when immediately followed by digit, to allow for JAV titles

https://github.com/stashapp/stash-box/blob/master/pkg/database/migrations/postgres/35_websearch.up.sql

PR #5645, opened 2025-02-04
Preserve JAV title in Tagger

javstash:
Check if string contains a JAV style code, if so preserve the dash in it.

Need to fix lint and maybe rename parseDate function

Related: https://github.com/stashapp/stash/issues/5644


WithoutPants:
Lint failure is unrelated and will be fixed by merging from the `develop` branch. Can you please provide some test scenarios so I can understand what this code change is trying to fix?

javstash:
This change will preserve the dash in jav codes. For example, when querying a file named `AAA-891.mp4` with query mode set to filename in tagger view, the query string will default to `aaa-891` instead of `aaa 891`.

PR #5661, opened 2025-02-14
Fix variables for help links in Stash Setup Wizard

DogmaDragon:
https://github.com/stashapp/stash/pull/5459 changed the variable from `githubLink`, `discordLink` to `GithubLink`, `DiscordLink`.

Not sure if reverting the variable name is preferred instead.


WithoutPants (approved):
Thanks. I've changed the values in the Setup.tsx file instead, otherwise we'd have to change it in all of the localisation files.

PR #5671, opened 2025-02-20
Fix markers ui bug

skier233:
Fixes the bug found here: https://github.com/stashapp/stash/pull/5633#issuecomment-2668662458

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
+++ b/ui/v2.5/src/components/ScenePlayer/ScenePlayer.tsx
@@ -697,59 +697,54 @@ export const ScenePlayer: React.FC<IScenePlayerProps> = ({
     const player = getPlayer();
     if (!player) return;
 
-    const loadMarkersAsync = async () => {
-      const markerData = scene.scene_markers.map((marker) => ({
-        title: getMarkerTitle(marker),
-        seconds: marker.seconds,
-        end_seconds: marker.end_seconds ?? null,
-        primaryTag: marker.primary_tag,
-      }));
+    const markerData = scene.scene_markers.map((marker) => ({
+      title: getMarkerTitle(marker),
+      seconds: marker.seconds,
+      end_seconds: marker.end_seconds ?? null,
+      primaryTag: marker.primary_tag,
+    }));
 
-      const markers = player!.markers();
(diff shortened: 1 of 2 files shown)


WithoutPants:
@skier233 I moved `loadMarkers` into a separate callback function, and removed the `async` keywords from `computeBaseHue` and all its client functions. It looks to have maintained the fix, and I couldn't find a reason for having the `async` keyword in that function. If you can take a quick look and verify, I'll get this merged.

SyZeeGee:
I just had a chance to test the new dev build and although it seems better (there is less phantom markers) clicking it fast still seems to induce the bug.  Figure since your mind space is still here, I'd let you know.  Here is a theory though possibly specific to my use case: is it possible the signal being listened to clear markers isn't being received if the file isn't accessible ("No such file or directory" due to being on an external drive not plugged in?)  I didn't have time to fully test this theory but it's possible.  

Note I appreciate any work at all on this really, so thanks really.  It's overall a really cool marker/video feature (for reasons past the main utility of this software.  Well done

On that note, related here: @WithoutPants : when an external hard drive is not present (and the file is not playable), the way the software works currently is that it repeatedly dumps out errors (The "No such file or directory" found one, until the page or video is left.)  It just doesn't seem to like the correct implementation (nor intended) here and wanted to let you know of it as well as it might be linked to the above issue. I just confirmed the CPU usage of the browser is going through the roof when it happens so it's not ideal for that reason alone.  Let me know if you'd like me to open up a dedicated issue or if its known already (I didn't know what to search)

WithoutPants:
> On that note, related here: @WithoutPants : when an external hard drive is not present (and the file is not playable), the way the software works currently is that it repeatedly dumps out errors (The "No such file or directory" found one, until the page or video is left.) It just doesn't seem to like the correct implementation (nor intended) here and wanted to let you know of it as well as it might be linked to the above issue. I just confirmed the CPU usage of the browser is going through the roof when it happens so it's not ideal for that reason alone. Let me know if you'd like me to open up a dedicated issue or if its known already (I didn't know what to search)

Please raise a separate issue for this. I'm not aware any existing issue that covers this.

SyZeeGee:
Thanks for opening those issues.  I just posted the one about the looping file reads  :+1:

PR #5677, opened 2025-02-24
Fix scraping multiple URLs 

### LINKED ISSUE #5294: [Bug Report] XPath scraper is missing string array support
**Describe the bug**
XPath scrapers can't return URLs in array format

**To Reproduce**
- Try to scrape with [POC fapello scrapers](https://gist.github.com/feederbox826/c6314601c401d1570f1271f50e371c7e)
- when using stringarr
```
scrapePerformerURL: input: scrapePerformerURL Internal system error. Error <reflect.Set: value of type string is not assignable to type []string>
reflect.Set: value of type string is not assignable to type []string
[3][URLs] = https://NSFW
[0][URLs] = https://NSFW
[2][URLs] = https://NSFW
[1][URLs] = https://NSFW
```
- when using objarr
```
key 'URLs': selector '': parse error: expr expression is nil
```

**Expected behavior**
xPath selector is taken as an un-merged array or nested support like with Tags is possible

v0.27.0

**Additional context**
Related to https://github.com/stashapp/stash/pull/4958
Blocking https://github.com/stashapp/CommunityScrapers/issues/2042

WithoutPants:
The mapped scraper functionality is in sore need of a rework. #5294 has highlighted this with the new `URLs` field. For now, I've hacked in a solution that should correctly convert URLs to a list. 

Notably, this does _not work_ where `URLs` is used in a sub-object scenario. For example, it will populate only a single URL in the `URLs` field of performers within a scene scrape. This is because there is currently no way to determine which performer a given URL value belongs to. Therefore, it follows existing convention and assigns one URL for each performer result. This means that it is not possible to have multiple URLs for any performer within a scene scrape.

Resolves #5294 

For testing, I used the following performer scraper for LinkTree:
```
name: LinkTree
performerByURL:
  - action: scrapeXPath
    url:
      - linktr.ee
    scraper: performerScraper

xPathScrapers:
  performerScraper:
    performer:
      Name: //div[@id='profile-title']/h1/text()
      URLs:
        selector: //div[@id='links-container']//a/@href
```

### DIFF
--- a/pkg/scraper/mapped.go
+++ b/pkg/scraper/mapped.go
@@ -1148,7 +1148,7 @@ func (s mappedScraper) scrapeImage(ctx context.Context, q mappedQuery) (*Scraped
 	imageStudioMap := imageScraperConfig.Studio
 
 	logger.Debug(`Processing image:`)
-	results := imageMap.process(ctx, q, s.Common)
+	results := imageMap.process(ctx, q, s.Common, urlsIsMulti)
 
 	// now apply the performers and tags
 	if imagePerformersMap != nil {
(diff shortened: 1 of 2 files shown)


DogmaDragon:
Tested with several existing community scrapers (IAFD, The Nude, Babepedia, MFC) and some custom scrapers. 
Tested across different actions (script, scrapeJson, scrapeXPath).

Everything looks to work in my testing so far. 

If `URLs` is used, `URL`, `Twitter` and `Instagram` are ignored while `URLs` is processed correctly. Which maintains compatibility for older scrapers.

PR #5696, opened 2025-03-03
add lumberjack log rotation

### LINKED ISSUE #5009: [Feature] Truncate & Archive Log Files
Turns out trying to open a GB size log file isn't always appreciated by the computer.
Would be great to be able to set a file size limit on Log File (eg 50MB), and once exceeding the limit, the log file is archived an a new one created. (Similar to AAR applications)

feederbox826:
Closes #5009

Tested and working, still in draft for UI config of rotation size


feederbox826:
not targeting 0.28 as iirc this is quite a niche feature and requires TRACE to generate anything substantial

edit: debug also does the job (came back to 10G logfile)

feederbox826:
To generate a 1MB log file,

- turn up log level to trace
- log http access

fill up the log
- go to a scene list
- go to tags
- repeat

WithoutPants:
This looks good, but I think it needs to be optional and turned off by default. Perhaps make the max log size value `0` by default, and when it is zero, the rolling log behaviour is zero. Having done a quick test, it zipped up my existing log file at startup, which is unexpected behaviour and probably not necessarily ideal for existing users. Some documentation in the manual might be worth doing as well to explain what it does (specifically what files it generates).

feederbox826:
No longer rotates by default, you can test this with this lovingly crafted 1GB zip bomb 
[stash.log.zip](https://github.com/user-attachments/files/23594570/stash.log.zip)

WithoutPants (approved):

PR #5697, opened 2025-03-03
Use existing formats for saved filters

### LINKED ISSUE #5690: [Bug Report] Default filters not being applied properly
**Describe the bug**
Default filters are not being applied properly. Type is read properly but value or the second value is nullified

**To Reproduce**
1. Set a default filter with a type and property (Organized: false, play_count between 0,2)
2. Set as default filter for scenes
3. Click on "Scenes" and watch as filter and URL are nullified

Seems to be similar result of #5108 but not manifesting in the same ways, more details below

develop v0.27.2-80-g23e36b12 hash 23e36b12

**Additional context**
Previous filters are stored and read back properly, but new filters seem to error out and inherit these properties, even when they are selected.

WithoutPants:
#5632 introduced some changes around the format of the saved filters that wasn't fully ready. This existing saved filter format is currently not correct. It's supposed to follow the format of the graphql filter inputs, but instead it's following the schema of the UI object model. Fixing that issue will require a migration to massage the saved filter formats. This PR rolls back the changes to the saved filter format.

Note that this will not fix corrupted saved filters created after the #5632 build.

Resolves #5690

### DIFF
--- a/ui/v2.5/src/models/list-filter/criteria/criterion.ts
+++ b/ui/v2.5/src/models/list-filter/criteria/criterion.ts
@@ -7,6 +7,7 @@ import {
   MultiCriterionInput,
   TimestampCriterionInput,
   ConfigDataFragment,
+  DateCriterionInput,
 } from "src/core/generated-graphql";
 import TextUtils from "src/utils/text";
 import {
@@ -1034,6 +1035,14 @@ export class DateCriterion extends ModifierCriterion<IDateValue> {
 
   protected encodeValue(): unknown {
     return encodeRangeValue(this.modifier, this.value);
+  }
+
+  public toCriterionInput(): DateCriterionInput {
+    return {
+      modifier: this.modifier,
+      value: this.value?.value ?? "",
+      value2: this.value?.value2,
+    };
   }
 
   protected getLabelValue() {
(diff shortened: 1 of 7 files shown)


ui/v2.5/src/models/list-filter/criteria/criterion.ts
```diff
@@ -940,19 +1023,17 @@ export class DateCriterion extends ModifierCriterion<IDateValue> {
     this.value = { ...this.value };
   }
```
feederbox826, line 1024: This might need to be removed? value and value2 are nested within a value object

feederbox826 (commented):
everything works **but** DateCriterionOptions. It clones it with 

```js
date:
  modifier: "BETWEEN",
  value: {
    value: "2025-01-01"
    value2: "2025-01-02"
  }
}
```
and similarly for single value

This occurs not on saving but when selecting the filter by itself. It is saved in this format as well. Hopefully this is as simple as removing the class overrides (looking at public cloneValues())

feederbox826:
:eyes: pulling, building and testing now

feederbox826 (approved):
LGTM. Fixes the filer format, including Date selector

PR #5713, opened 2025-03-11
Optimise card width calculation

WithoutPants:
Optimises and refactors the grid card width calculation code. Calculation will be performed once by the parent CardGrid component, instead of once per card. FYI @cj12312021


ui/v2.5/src/components/Shared/GridCard/GridCard.tsx
```diff
@@ -78,6 +79,31 @@ export const useContainerDimensions = <T extends HTMLElement = HTMLDivElement>(
   return [target, dimension];
 };
 
+export function useCardWidth(
+  containerWidth: number,
+  zoomIndex: number,
+  zoomWidths: number[]
+) {
+  return useMemo(() => {
```
cj12312021, line 87: Thanks for this update. I'll need to remember that `useMemo` can be used like this to minimize operations.

PR #5714, opened 2025-03-11
[WIP] Filter sidebar

WithoutPants:
This is a less ambitious attempt at adding a filter sidebar to the scenes page. 

Currently includes:
- saved filters - save/set as default functionality moved here
- performers
- studios
- tags

I intend to include rating and organised options in this PR.

Replaces the Saved Filter dropdown button with a sidebar toggle button.

image

This is currently WIP. PR is so that people can try out the build.


MinasukiHikimuna:
What do you think: could the expanded/collapsed state be saved? Now if I have specific tags and performers selected, open a scene and come back scenes list view, all the options are collapsed with no visual indication which of those have some options set. 

Proficient Stash users can determine the options from the standard filtering options though.

WithoutPants:
New updates from yesterday:

- sidebar now sits overlapping the content pane on smaller viewports. Adds a close button on these viewports and will automatically close the sidebar when clicking outside the sidebar.
- selected items now always show under the heading, instead of being hidden behind the collapse.

image

> What do you think: could the expanded/collapsed state be saved? Now if I have specific tags and performers selected, open a scene and come back scenes list view, all the options are collapsed with no visual indication which of those have some options set.

Saving sidebar state in the browser is at the top of my todo list.

echo6ix:
### Placement of sidebar toggle and sidebar specific buttons

@WithoutPants are you open to taking inspiration from ChatGPT's sidebar UI/UX? Specifically the placement of the toggle and other relevant primary buttons, which are placed at the top
- Toggle button is fixed (top right), regardless of sidebar being enabled or disabled
- Fixed toggle button allows for notification badge for sidebar specific items (as per animation). Stash may not have a use for that now, but the possibility is there for the future.
- Top of the sidebar (left justified) is a good potential placement for other fundamental buttons

screentogif

MinasukiHikimuna:
What is the plan for this feature? Will it be merged to develop now that 0.28 has been released?

I'm already used to rely on this so I cannot imagine using Stash without it. Great work on this!

WithoutPants:
Now that the prerequisite PR has been merged, I've rebased and done some tidying. This is now ready for testing.

Here is how it looks now.
image

- sidebar button is on the left per @echo6ix's suggestion. Close button is removed.
- filter input and edit filter buttons are now inside the sidebar. 
- current filters are studios, performers, tags, rating and organised. This is currently hard-coded, though the sidebar filter section is patchable to add more.
- rating shows stars regardless of rating system chosen, for quick access. Clicking on a star sets the filter to that star level and above
image
- when a filter option is set, the value is shown before the collapse part, so that they are always visible:
image

- studio and tag filters include a `(subsidiary studios)`/`(sub-tags)` option to include the subsidiary objects, instead of a check box
image

This is how it looks when the sidebar is hidden:
image

On mobile devices, the top of the screen now looks like this:
image

Clicking on the sidebar button brings the sidebar across the entire width of the screen with larger font sizes:
image

cj12312021:
I'm happy to see progress on this PR. I have to say though that lack of symmetry here really bothers me. All the toolbar items are not completely left aligned, while the pagination and the pagination index text remain centered directly below it. It's even more jarring when browsing the other list pages that still maintain a centered toolbar. I'd propose keeping the toolbar centered until we are closer to the design we workshopped last year:
Screenshot 2025-03-31 230212

The inefficient use of space at the top of the sidebar tool is also bothersome to look at (pardon my language; my background as an artist before becoming an engineer makes it hard not to be critical about things like this). For this, I'd suggest having the search bar and the sidebar buttons share the same line. The filter button can live at the bottom of the filters as per the initial design. I have a feeling this would be a bit harder to negotiate, so if one of my suggestions here were to be adopted, I'd say the suggestion in the first paragraph is the most important.

The concept of custom styles and plugins exists for users to adjust these things, but we really need to ensure the default design is appealing. This is the first thing new users will see. It needs to make a good first impression.

WithoutPants:
@cj12312021 first, thank you kindly for the forthright feedback. It's always appreciated. 

> I have to say though that lack of symmetry here really bothers me. All the toolbar items are not completely left aligned, while the pagination and the pagination index text remain centered directly below it. It's even more jarring when browsing the other list pages that still maintain a centered toolbar. I'd propose keeping the toolbar centered until we are closer to the design we workshopped last year: 

Fair call. I've made toolbar center aligned once again:

image

image

> The inefficient use of space at the top of the sidebar tool is also bothersome to look at (pardon my language; my background as an artist before becoming an engineer makes it hard not to be critical about things like this). For this, I'd suggest having the search bar and the sidebar buttons share the same line. The filter button can live at the bottom of the filters as per the initial design. 

This one is indeed a bit more contentious. I agree with you that the sidebar toolbar has a bit of wasted space. I think moving the edit filter button to the bottom makes the process of editing a filter from the hidden sidebar state a bit too onerous for my liking. You would have to click the sidebar button at the top, then move the mouse to the bottom (and the list can and will grow). You'd be swapping an aesthetic issue with a usability issue. Positioning it near where the mouse will be when the sidebar is opened would be ideal.

Having all three buttons on the one line makes the search input field prohibitively small at the current sidebar size:
image
image

I do like that the current positioning of the sidebar button is consistent when opening and closing the sidebar, so I'd be reluctant to change that. I'm happy to workshop this further, but to me it feels like a much more minor aesthetic issue. 

> The concept of custom styles and plugins exists for users to adjust these things, but we really need to ensure the default design is appealing. This is the first thing new users will see. It needs to make a good first impression.

Fully agree. We do have time to tighten this up before release though.

DogmaDragon (commented):
New `/` keyboard shortcut needs to be documented.

WithoutPants:
> These screenshots provided some therapy on what has personally been a stressful workday. This is much better.
:heart:
 
> I am cognizant of the fact that those buttons limit the size of the search bar. If it’s not too much hassle, one consideration may be having the search bar dynamically take up the full width when a user engages with it.

I'm not exactly sure how this would work. How would it look prior to interacting with the input field? How would it look with some input text entered when the user has focused elsewhere? How would the user access the sidebar/filter buttons while focused on the input field? At the moment also the search input is focused on when the sidebar is opened. We can of course change this behaviour to require an explicit action to focus it (currently `/` opens the sidebar and focuses on search input - an alternative would be first keypress opens the sidebar and second press focuses on the search input).

cj12312021:
> I'm not exactly sure how this would work

Take a look at the search buttons in the example here: https://codepen.io/k185/pen/PQajXE. The behavior seen in those search buttons is what I had in mind. It may appear that this adds an extra step to interacting with the search feature, but that is not actually the case. Even now, a user still needs to focus on the search box before they can type, whether that is by clicking on the search box or tabbing into it.

> I think moving the edit filter button to the bottom makes the process of editing a filter from the hidden sidebar state a bit too onerous for my liking.

Going back to this earlier point, we could simply move the edit button directly below the search box rather than at the very bottom of the sidebar.

cj12312021:
If we use the expanding search bar, the search text may need to appear as a pill below that menu once the search bar has collapsed.

QxxxGit:
I'm sure visual bugs will be addressed after the design is finalized but I'm logging it in case I forget.

There's a tiny visual bug when a tag/performer/studio label is long and you hover the exclude button. The label is breaks into a new line when it could probably make use of TruncatedText.

And while I'm on the subject of it, I'm not quite sure how everyone feels but it seems like bad UX to feel the need to include "exclude" since it's not clear what the button does. This may be worth reconsidering.

cj12312021:
Here is a screenshot of the option that suggests placing the filters button below the search box. The extra space provides us the ability to give context that users are not limited to the provided filters in the sidebar, as I'm sure this might not be clear to some users:

Screenshot 2025-04-02 232604

WithoutPants:
> And while I'm on the subject of it, I'm not quite sure how everyone feels but it seems like bad UX to feel the need to include "exclude" since it's not clear what the button does. This may be worth reconsidering, if WP feels it's worth touching with this update. It may be something to update in another PR since it's technically out of scope.

The idea was shamelessly stolen from the way that the bang website does it (though with a count number instead of a `-` icon. Happy to revisit this one but yeah, not something I really want to address in this PR.

echo6ix:
Search alternatives. ChatGPT's side bar search ui/ux

#### Desktop (large viewport) mode
- Search button opens a search modal
- Modal closes with close button, escape keypress, submission or carriage return

screentogif

#### Mobile (small viewport) mode

screentogifx

WithoutPants:
We'd need to include the query string in the filter tags if we were to do something like that, which again is probably out of scope for this change. Otherwise, we wouldn't be able to see the current filter string.

I like the idea of moving the edit filter button to the next line. I'm not convinced that the search input field looks great, but again for the purpose of this PR it's probably good enough.

cj12312021:
I think this design with the filter button below the search box looks more appealing than the one I provided earlier:
Screenshot 2025-04-04 202624

cj12312021:
For what it's worth, other sites I've with a search box in their sidebar also don't use the full width of the sidebar for text input.
Gamma:
Screenshot 2025-04-04 205359

Bang:
Screenshot_4-4-2025_21542_www bang com

echo6ix:
WithoutPants' initial concept hits the mark on a lot of sidebar search/filter examples I'm seeing promoted on uiux best practice articles. For example...

image

They're not cramping elements together for a sense of balance. They're emphasizing visual hierarchy and clarity.

```
|------------------------------------------|  <-- Sidebar here
|                                          |
| [Toggle btn]       [Other primary btns]  |  <-- Primary display block (primary control/visibility state buttons)
|                                          |
| [🔍 Search input for scenes            ] |  <-- Secondary display block (data fetching)
|                                          |
|------------------------------------------|
| [Filter stuff here]                      |  <-- Search/filter refinment controls

```

Primary top display block
- Sidebar toggle at the top, either left or right justified
- One or two other primary elements at the top justified opposite to the sidebar toggle (a logo in the above example. in the ChatGPTside bar example its elements that control states)

Secondary top display block
- search on its own visual display block + visual padding from other neighboring elements does the following: 
  - search stands out on its own w/o other unlike elements creating visual distraction
  - separates intent from other elements aka basic principle of grouping semantically similar elements together (the search input is different from primary buttons that the control the visibility of elements, such as sidebar toggle and filter button)
  - maximizes horizontal sidebar space constraints which important for search. Should display search icon, placeholder text of what of what's being searched, and also very useful to show keyboard shortcut that grabs search input focus


For the search input itself there's some good best practice suggestions here https://www.pencilandpaper.io/articles/search-ux namely:
- Include a search icon which shows users a very noticeable hover state when hovered upon (ensure the keyboard shortcut to initiate search is also thought through, especially in cases where search is a common action)
- Include the focus state when the input has been “clicked into”
- Include some preset suggestions in the placeholder text (in cases where it isn’t clear what you can search for in an app)
- Include what is being searched in the placeholder text. Ex. search entire website vs. search items in table below

cj12312021:
If we put appearances aside and focus on the UX of the initial design. I don't think the placement of additional filters above the search bar away from other filters was a thoughtful decision. It makes more intuitive sense to have it by the other filter with clarifying text if there is room for it. If the goal is to enable the sidebar to occupy the full width, a design like the one below would be preferable.
Screenshot 2025-04-04 224040

I just happen to think we can blend both. It's not as if the ability to increase the width of the sidebar is off the table if a user prefers more space. On a modern mobile device, where the sidebar occupies the full screen width, space should never be a concern.

cj12312021:
Here is a mock-up of how the badge pill could be used with the labeled filter button:
Screenshot 2025-04-04 232206

cj12312021:
As I continue to ponder this, the fact that the additional filter button is meant to be an addition on top of the filter options we've deemed are core to scenes strongly suggests to me that the option button should be at the button below the core filters. If our assessment of what is core is accurate, a user wouldn't need to access the additional filter button all that frequently, especially when they can save any filter combinations they applied with the extra filter options. 

This is especially true if the concept of allowing users to pin what they deem are core to the sidebar is something we eventually plan to bring over from your original concept last year. I am firmly in the camp now of giving users the decision to identify what filters are core to their use case and pinning it to the sidebar, as I can see a world where more filter options being added directly to the side would cause issues. 

Anyway, as I stated earlier, the symmetry issue was the core issue I felt most strongly about, and it has been addressed, so I'm happy. I do think the placement of the extra filter button should be reconsidered, but I shouldn't lose too much sleep over it if it isn't.

WithoutPants:
Good discussion - thanks for your feedback.

I agree that having the edit filter button above the search input is incorrect grouping. Moving it below the search input conveyed an importance that I felt was unwarranted. Having it next to the search input didn't make the search input smaller than the existing input on other pages, and with a slight bit of margin felt less crowded. To me, it seems right in this position:

image

I increased the margin between the sidebar button and the search input to separate them logically a bit more clearly. Likewise, I added a bit of space below the search field so it didn't feel as cramped.

I moved the sidebar section bottom border to encompass the whole section - you can see in the earlier screenshots that the border would appear above the currently selected items.

I tried to address the overflowing titles in the filter lists. 

image
image

As you can see, the positioning gets dinky once it does overflow. I'm deferring solving this to later. If you have any solutions, hit me up.

cj12312021:
I just pulled this change into my production system and have noticed a signifcant impact the query times. I'm currently attemptng to load my scenes page. My timer has appraoched 6 minutes and the page is still loading. You should be able to test this with the database I sent you earlier. My default filter is set to exclude the `Game Clips` tag.

cj12312021:
For reference, after reverting the merge, my load time with that default filter returns to 3 seconds.

cj12312021:
I should clarify that when I said, "I just pulled this change into my production system", I meant that I pulled the branch into my prod system for the first time. I wasn't strictly referring to the updates in your latest push.

cj12312021:
I also notice that using filter options such as studios and tags is very slow. The studio performance stood out to me the most. It doesn't matter if this was done within the side bar or via the Edit filter menu.

WithoutPants:
Thanks, this was an important catch. I did some restructuring so that the select filters won't load until they are opened for the first time, instead of loading all at once. I also fixed the select filters to use the select queries instead of the full object queries. I tested using your database and it's significantly improved.

bob123491234:
Some minor cosmetic things I noticed:
- The buttons in the Save filter dialog box don't have a margin added to them.
- Clicking the Open sidebar button on a performer/studio/tag page shits the button down a bit.
- Saved filters list isn't indented as much as the other lists.
- Stars in the rating section look off pushed all the way to the left; I think they would look better indented to where the text is.
- Highlighted border around the search and Edit Filter boxes when they are focused is cut off on the top.

On mobile:
- Font size in the sidebar seemed larger than it should be.
- The background is visible at the bottom of the filter panel.

WithoutPants:
Thanks @bob123491234. These should all be addressed now.

cj12312021:
The core of my issue here may be actually with the sidebar button remaining isolated at the top left of the sidebar. It feels even more off in this view, given the list is not at the top of the page. This may just be something I address locally. 

> the expand/collapse sidebar button is currently bugged in this view. The button's position shifts each time it is engaged.

The bug I mentioned in this view would still need to be addressed.

echo6ix:
It's difficult to adequately address the above feedback, if I understand correctly, because it's based on a subjective impression, which is 100% valid for you, don't get me wrong.

But I think we can mitigate subjective back-and-forths by identifying the goals of the design. I don't know what those are, only WP does, but I suspect one of his goals might be to prioritize an intuitive beginner-friendly sidebar design? There are objectively good and bad ways to achieve that.

Like ChatGPT's UI and other web apps using this exact sidebar design, it addresses that goal by
- using a recognizable sidebar button label of a literal sidebar icon
- using a commonly expected, logical, and convenient location. Sidebar button near the sidebar, and it remains static there regardless of toggle state
- using a commonly expected behavior (toggle button open, toggle button close)

It's very boring, predictable, but modern. It's not reinventing the wheel, which are all good things for addressing that design goal

cj12312021:
@echo6ix, the policing you are attempting to do here is unnecessary. The root of my earlier question was based on feedback we had received on previous revisions of this work. As far as I was aware, this sidebar work was only planned to be supported on the main list. Based on that feedback, the presence of the sidebar in the secondary list could have been considered a bug. The various issues I have seen with the side in this view would have supported the idea of this being a bug. In my second post, I concluded that its appearance in the secondary list may have been intentional due to my noticing that Bang also uses the sidebar in its secondary list. I assumed this because Bang has been stated as a source of inspiration here. With that being the case, I proceeded under the logic that the sidebar secondary list was likely intentional but pointed out that there are bugs in the view that would need to be addressed. The originally stated bug being the one I highlighted here:

> the expand/collapse sidebar button is currently bugged in this view. The button's position shifts each time it is engaged.

That, however, is not the only issue. There are several other aesthetic issues that still need to be addressed, many of which can be seen in this screenshot:
Screenshot 2025-04-10 140321

With that aside, let's not pretend that the suggestion made here won't come down to aesthetics. Your suggestion to have the sidebar button pinned to the top left was an aesthetic choice.

echo6ix:
I finally downloaded the binary with the scene filter side!

Some feedback a bit at a time.

image

1. **Search input placeholder text**. It's good practice for the search placeholder text to denote what is being searched. When on the scenes pages, the sidebar search should read `Search scenes`, if on the image pages then `Search images`, if there's a search input form for filters then `Search filters`, etc.
2. **Display hotkey to focus search**. The `/` key focuses the search form, but many users aren't away. There's enough space in the unfocused search input to denote that shortcut key. Some search input form just display the shortcut key, `/` as I have wrapped with a border, and some are more comprehensive but require more space like Github's 
image
3. **Filters**.  Having the filters in a dropdown, even on the sidebar and somewhat how the current implementation is, is actually not a bad idea. It's not that uncommon as a look at sidebar ui/ux concepts.
    -  The unfocused filter dropdown always displays the currently active filter name. That prevents it from having to display the filter area as a constantly expanded state to display the active filter
    - Each filter requires controls (save, delete, set as default), which can be merged into an overflow menu for each filter entry.
    - Filter search. Search filter input would also be contained within the dropdown

cj12312021:
@echo6ix I should clarify that the reason your message irked me, yielding the response I gave, was because my message was not addressed to you. Yet, you stated your ignorance in understanding my message and proceeded to "address" it on WP's behalf to minimize the feedback you didn't understand. I should also be clear that I have never stated I disliked your sidebar button. The feedback I've given about the button placement never suggested removing or relocating it. The feedback has only been around relocating the elements around that button.

WithoutPants:
Apologies, my communication around this feature has been split between here and the dev-diary posts on Discourse.

In [here](https://discourse.stashapp.cc/t/dev-diary-20-03-2025-0-28-release-and-0-29-plans/387), I discuss the goals for 0.29 and the sidebar stuff. In [https://discourse.stashapp.cc/t/dev-diary-26-03-2025-sidebar-progress-and-other-bits/404](here), I talk about the sidebar stuff in the secondary views.

The summary is that yes, I'm looking at having the sidebar support on secondary pages. I intend the edit filter dialog to be reworked to be more of an advanced filter editor, with the sidebar being the place for more immediate (and simpler) filtering.

For most of the page layouts I don't see a huge issue having the sidebar. The exception will be the gallery page, which currently has a left docked pane. The short-term plan will be for this page to not have a sidebar, but I do think that the gallery page should be redesigned to be more like the performer detail page.

All that said, I think it might be worth holding back on the secondary page changes and address them separately in another PR. It's only going to further delay getting the broad strokes of the sidebar functionality merged, and it has its own design issues that will need to solved.

cj12312021:
I appreciate the clarification. I've been out of touch with discourse these days. I agree the sidebar can work on the secondary list pages; we may just need to rethink some of the design choices that were made below the header area. I'll withhold remaining commentary on the secondary list pages until that work starts.

cj12312021:
I hope you will forgive my broken promise to disengage with commentary on the secondary list. Some of the fixes here are so straightforward that pushing them out to a second PR and leaving users on the dev build to bear with the issues in the meantime doesn't feel right. I'll make this as painless as possible. I've provided a video that highlights some of the issues. I also provided the CSS to address the issues and another video showing the improvements. 20 seconds into the last video, I captured an issue that was not addressed in my CSS. That will likely require additional consideration that may be out of the scope of what you want to address here, so I've left it alone.

Before Video: https://discord.com/channels/559159668438728723/644934273459290145/1360401979611807929

The CSS:
```
.detail-body .sidebar {
   margin-top: -15px;
}

.detail-body nav {
    border-bottom: 1px solid #394b59;
}

.detail-body .tab-content {
    padding-bottom: 0;
}
```

After Video: https://discord.com/channels/559159668438728723/644934273459290145/1360402000772075650

These are only suggestions.

ui/v2.5/src/components/Scenes/styles.scss
```diff
@@ -901,3 +901,18 @@ input[type="range"].blue-slider {
...
+}
+
+.scene-list.hide-sidebar .sidebar-toggle-button {
+  transition-delay: 0.1s;
+  transition-duration: 0;
+  transition-property: opacity;
+}
+
+.scene-list:not(.hide-sidebar) .sidebar-toggle-button {
+  opacity: 0;
```
cj12312021, line 917: I think I understand the reason you set the opacity here rather than setting `display: none`. However, the issue with just setting opacity to 0 is that the invisible button remains engageable if a user's mouse happens to stumble onto its location on the page. We can avoid this by also setting `pointer-events: none;`
```suggestion
  opacity: 0;
  pointer-events: none;
```

cj12312021:
Here are a couple more things to be aware of:

The studio filters seem to be the multi-select filter that doesn't use the green plus on the left side. I've noticed this for a while but was never sure if this was intended:
Screenshot 2025-04-12 203845

Currently, the alignment of the toolbar is slightly off due to the sidebar button on the left side, which occupies  37 pixels of horizontal space. There are a few ways this can be fixed, but the simplest way may be to add 37 pixel right padding to `filtered-list-toolbar`:
Untitled

cj12312021:
I want to document a solution for the `Include sub-studio content` toggle button placement issue with the sidebar I mentioned 2 days ago. The suggestion is to move the toggle to the same line as the nav tabs for the sub menus. I think this placement makes more sense as it better illustrates the impact the toggle would have.

Screenshot 2025-04-13 124151

Here is the CSS used to put this together, although it probably shouldn't be implemented with just CSS.

```
#studio-tabs-tabpane-scenes {
    position: relative;
}

#studio-tabs-tabpane-scenes .item-list-header {
    position: absolute;
    right: 15px;
    top: -43px;
}

.detail-body nav {
    border-bottom: solid 1px #394b59;
}
```

cj12312021:
I've been running this branch on my prod system for over a week now with some of the UI corrections I mentioned in my most recent posts. I have to say that I've been very happy with this work and plan to hold on to it until it eventually makes its way into the development branch.

Credit where it's due; placing the filter button to the right of the search input was a good call. When I suggested moving it to the button below the other filters, I made that suggestion, thinking that the window the button opens would only house the remaining filters. Which I later discovered was not the case. Since the `badge-pill` this button uses tracks all active filters (including the ones directly in the sidebar), it does make sense to me for it to appear before all the filters.

I think the implementation of persisting the sidebar state and autofocusing the search input when the sidebar is open works well in addressing some of the concerns people have had about moving the search input. There is still the issue that users currently have no way of knowing when a search filter has been applied to their list if the sidebar is closed. This problem should probably be addressed at some point, likely by adding a `filter tag` for the searches.

MinasukiHikimuna:
I've been enjoying these changes now for a few weeks and there is no going back for me so I'll be waiting for this to be merged until switching back to develop branch. Can't wait to get these changes included as well: https://discourse.stashapp.cc/t/dev-diary-10-04-2025-making-sidebar-filtering-even-better/579

WithoutPants:
I intend to merge this branch in the near future. The broad strokes of this feature are stable enough, and I think I've addressed the issues reported. I intend to revisit the sidebar button as part of redesigning the filter toolbar, which is what I'll move onto next. I'll also look at adding the search term to the filter tags.

DogmaDragon:
Minor visual bug
image

On focus filter button border is getting clipped.

cj12312021:
I just noticed the feedback [here](https://github.com/stashapp/stash/pull/5714#issuecomment-2800051670) was never addressed in the main repo. I really think it would be worth addressing this disjointness on the Studio page before release.

Current:
Screenshot 2025-07-04 102057

Expected:
Screenshot 2025-07-04 102539

PR #5718, opened 2025-03-12
Enhance blacklist handling: regex replacements and partial date removal

Gykes:
So this is still a draft as I am TERRIBLE with Regex and just as bad with coding. Please let me know if there is anything that needs fixed, I'm happy to do so. 

- Added support for regex-based replacements in blacklist processing
- Implemented `||` delimiter to allow custom replacements (e.g., "bwb||bigwetbutts")
- Improved regex handling for partial date removal:
  - Allows removing only the day (`-DD`) from `YYYY-MM-DD` while keeping `YYYY-MM`
  - Example: `"(\d{4}-\d{2})-\d{2}||$1"` removes `-DD` but preserves `YYYY-MM`
- Prevented regex errors using `try/catch`
- Ensured spaces are normalized after replacements

### DIFF
--- a/ui/v2.5/src/components/Tagger/utils.ts
+++ b/ui/v2.5/src/components/Tagger/utils.ts
@@ -3,6 +3,7 @@ import { ParseMode } from "./constants";
 import { queryFindStudio } from "src/core/StashService";
 import { mergeStashIDs } from "src/utils/stashbox";
 
+
 const months = [
   "jan",
   "feb",
@@ -92,6 +93,10 @@ const handleSpecialStrings = (input: string): string => {
   return output.replace(/-/g, " ");
 };
 
+function parseDate(input: string): string {
+  return input.replace(/\./g, " ").replace(/ +/g, " ").trim();
+}
+
 export function prepareQueryString(
   scene: Partial<GQL.SlimSceneDataFragment>,
   paths: string[],
@@ -99,17 +104,6 @@ export function prepareQueryString(
   mode: ParseMode,
   blacklist: string[]
 ) {
-  const regexs = blacklist
-    .map((b) => {
-      try {
-        return new RegExp(b, "gi");
-      } catch {
-        // ignore
-        return null;
-      }
-    })
-    .filter((r) => r !== null) as RegExp[];
-
   if ((mode === "auto" && scene.date && scene.studio) || mode === "metadata") {
     let str = [
       scene.date,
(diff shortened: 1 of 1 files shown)


ui/v2.5/src/components/Tagger/utils.ts
```diff
@@ -92,55 +93,68 @@ const handleSpecialStrings = (input: string): string => {
   return output.replace(/-/g, " ");
 };
 
+function parseDate(input: string): string {
+  return input.replace(/\./g, " ").replace(/ +/g, " ").trim();
```
feederbox826, line 97: `/ +/g` is replacing instances of 1 or more spaces, I think you want `/ {2,}/g` for 2 or more

ui/v2.5/src/components/Tagger/utils.ts
```diff
@@ -92,55 +93,68 @@ const handleSpecialStrings = (input: string): string => {
...
+  // Apply regex filtering and replacements
+  regexs.forEach((regex) => {
+    const replacement = regexReplacements.get(regex.source);
+    str = replacement ? str.replace(regex, replacement) : str.replace(regex, " ");
   });
-  s = handleSpecialStrings(s);
-  return s.replace(/\./g, " ").replace(/ +/g, " ");
+
+  str = parseDate(str);
+  return str.replace(/\s+/g, " ").trim(); // Normalize spaces
```
feederbox826, line 154: Spaces are already parsed out in parseDate, why not normalize them earlier?

ui/v2.5/src/components/Tagger/utils.ts
Gykes: Good call, will fix in next commit

ui/v2.5/src/components/Tagger/utils.ts
Gykes: Also good call, I can probably move it all into ParseDate and call it a day.

ui/v2.5/src/components/Tagger/utils.ts
```diff
@@ -92,55 +93,77 @@ const handleSpecialStrings = (input: string): string => {
...
+        if (replacement  !== undefined ){
+          regexReplacements.set(compiledRegex.source, replacement); // Store replacement
+        }
+
+        return compiledRegex;
+      } catch {
+        return null; // Ignore invalid regex patterns
+      }
+    })
+    .filter((r) => r !== null) as RegExp[];
```
WithoutPants, line 148: It's not ideal to use `map` and then mutate stuff outside it. Instead, I would return a tuple (`return [compiledRegex, replacement]`). Then, these pairs of values can be handled in the code below. I would also move this into a separate function.

ui/v2.5/src/components/Tagger/utils.ts
```diff
@@ -92,55 +93,77 @@ const handleSpecialStrings = (input: string): string => {
...
+  // Apply regex filtering and replacements
+  regexs.forEach((regex) => {
+    const replacement = regexReplacements.get(regex.source);
+    if (replacement) {
+      str = str.replace(regex, (match:string, ...groups: string[]) => {
+        return replacement.replace(/\\(\d+)/g, (_, groupIndex) => groups[parseInt(groupIndex, 10) - 1] || "");
+      });
+    } else {
+      str = str.replace(regex, "");
+    }
```
WithoutPants, line 159: This should also be moved into a separate function. The code between 154-156 needs some explanation.

WithoutPants (changes requested):
Run `make fmt-ui` to fix up the prettier errors.

Issue #5720, opened 2025-03-13
[Bug Report] Vacuum fails due to disk space


WithoutPants:
**Describe the bug**
While testing a 58GB database, the migration is failing with the following error in the post-migration optimisation stage:

`error performing migration: error optimising database: performing vacuum: database or disk is full`

This was occurring despite there being plenty of space on the filesystem containing database having plenty (>500GB) of space.

From the [sqlite docs](https://sqlite.org/tempfiles.html#tempdir):
> On unix-like systems, directories are searched in the following order:
> 1. The directory set by [PRAGMA temp_store_directory](https://sqlite.org/pragma.html#pragma_temp_store_directory) or by the [sqlite3_temp_directory](https://sqlite.org/c3ref/temp_directory.html) global variable
> 2. The SQLITE_TMPDIR environment variable
> 3. The TMPDIR environment variable
> 4. /var/tmp
> 5. /usr/tmp
> 6. /tmp
> 7. The current working directory (".")
>
> The first of the above that is found to exist and have the write and execute bits set is used. The final "." fallback is important for some applications that use SQLite inside of chroot jails that do not have the standard temporary file locations available.
> On Windows systems, folders are searched in the following order:
> 
> 1. The folder set by [PRAGMA temp_store_directory](https://sqlite.org/pragma.html#pragma_temp_store_directory) or by the [sqlite3_temp_directory](https://sqlite.org/c3ref/temp_directory.html) global variable
> 2. The folder returned by the GetTempPath() system interface.

For me, this was using `/var/tmp` which only had 35GB of space.

I suspect this will trip other users up, and the error message doesn't make it obvious as to what is going on.

My current thinking is that we should set `PRAGMA temp_store_directory` to the same directory as the database file.

The workaround is to set the `SQLITE_TMPDIR` or `TMPDIR` environment variables (or whatever the equivalent on windows is). 

**To Reproduce**
1. Ensure the temporary location per the spec above has less space than the size of the database file.
2. Migrate or run the Optimise task

**Expected behavior**
Migration/optimisation should complete, assuming there is enough space on the filesystem containing the database.

**Additional context**

WithoutPants:
Closing since migration no longer performs a vacuum, and the pragma to set the temp directory is deprecated. The workaround will be to set the applicable environment variable instead.

PR #5723, opened 2025-03-14
Fix various migration issues

WithoutPants:
Thanks to @cj12312021 for the massive database to test the migration process against. Resolves the following issues found while testing:
- UI now shows when the database is being backed up in the first step
- close the migration connection to the database before re-connecting the main database. This should hopefully address the `database locked` and database corruption issues that were encountered
- ~~when running a vacuum, the wal file is now flushed so that we're not left with a wal file the size of the whole database file~~
- removed the post-migrate vacuum. In many cases this isn't necessary, and it takes a very long time on larger databases. In future it might be worth adding an option to enable it, or to automatically do it after large schema restructures. Alternatively, we can just recommend it as a user-initiated post-migrate step.

### DIFF
--- a/pkg/sqlite/database.go
+++ b/pkg/sqlite/database.go
@@ -425,6 +425,15 @@ func (db *Database) Optimise(ctx context.Context) error {
 // Vacuum runs a VACUUM on the database, rebuilding the database file into a minimal amount of disk space.
 func (db *Database) Vacuum(ctx context.Context) error {
 	_, err := db.writeDB.ExecContext(ctx, "VACUUM")
+
+	// toggle journal_mode to flush the wal file
+	if _, err := db.writeDB.Exec("PRAGMA journal_mode=DELETE"); err != nil {
+		return err
+	}
+	if _, err := db.writeDB.Exec("PRAGMA journal_mode=WAL"); err != nil {
+		return err
+	}
+
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


pkg/sqlite/database.go
```diff
@@ -425,6 +425,15 @@ func (db *Database) Optimise(ctx context.Context) error {
...
 	_, err := db.writeDB.ExecContext(ctx, "VACUUM")
+
+	// toggle journal_mode to flush the wal file
+	if _, err := db.writeDB.Exec("PRAGMA journal_mode=DELETE"); err != nil {
+		return err
+	}
+	if _, err := db.writeDB.Exec("PRAGMA journal_mode=WAL"); err != nil {
+		return err
+	}
+
```
feederbox826, line 436: I'm not sure WAL disabling is what you want in this scenario. VACUUM, by principle writes the database into a temporary file, [WAL if possible](https://sqlite.org/lang_vacuum.html). If you keep the WAL set, it will write to a predictable and expected file, otherwise it will follow the directory chain as you outlined in #5720

WithoutPants: I can move this into a separate function for clarity, but all this is doing is disabling wal to flush and delete the wal file, then setting the journal mode back to wal.

feederbox826: I understand what it's trying to do, I just have anxiety from working with sqlite in plex DBs where anything will set it off, especially out-of-band WAL flushing and interupting long sequences like VACUUM

WithoutPants: I'm not quite sure how this interrupts the vacuum, but it isn't necessary for this PR in any case, so I can revert for now.

feederbox826 (approved):
feederbox approved

PR #5739, opened 2025-03-19
Display tag and performer image on hover. on the scene edit page

### LINKED ISSUE #4320: [Feature] Display Tag Image/Performer Image on hover, after they are selected on the Edit page
**Challenge**
On the **Edit** page of a Scene or Image, after selecting a Tag or Performer the UI currently does not display any information on hover. This feels a bit inconsistent considering you do see a nice visual of that tag in the dropdown.

selectedperformerandtag

**Proposed Solution**
In similar form to how the hover functionality works when hovering over an object in the dropdown for a Tag, display the same content when a user hovers over the object in the selected Tag/Performer box.

_(I reused an image I created for the initial "on hover" concept here, just assume I'm hovering over the object rather than the dropdown)_
StashIdeaForTagHover2

blaspheme-ship-it:
Closes #4320 

- adds component to display a performer pop over copied from the TagPopoverComponent and wraps the select label and options in it. 
- Also raps tag label in the currently existing TagPopoverComponent.

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerPopover.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerPopover.tsx
@@ -11,7 +11,9 @@ interface IPeromerPopoverCardProps {
   id: string;
 }
 
-export const PerformerPopoverCard: React.FC<IPeromerPopoverCardProps> = ({ id }) => {
+export const PerformerPopoverCard: React.FC<IPeromerPopoverCardProps> = ({
+  id,
+}) => {
   const { data, loading, error } = useFindPerformer(id);
 
   if (loading)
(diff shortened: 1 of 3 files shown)


WithoutPants:
I reverted the change to show the hover popover when mousing over options in the drop down. I personally find the hover popover behaviour for tags highly irritating. I tried making the popover only occur when mousing over the performer image, but it still had issues overflowing the window. To me it seems unnecessary: most relevant information is available in the option already, and you can navigate to the performer page by clicking on the performer image (which you can't currently do for tags). In any case, I don't want that to prevent merging in the rest of it, so feel free to submit that part as a separate PR for discussion if you wish.

It's not ideal that the popover components make a graphql request for the tag/performer object, when we have the objects available in the select control. I tried changing the popover components to accept an object, but it appears at least for tags, the returned data object is missing information, which caused a UI crash when hovering. I've left it as is for now.

Issue #4684, opened 2024-03-13
[Bug Report] New scenes display the scraped tags in non-alphabetical order


rastarr:
**Describe the bug**
It's very helpful to see the list of scraped tags in alphabetical order, for example, to see what's missing
With the v0.25 and v0.25.1 releases, these tags are in an unsorted order.

**To Reproduce**
1. Go to 'Scenes'
2. Click on 'Filter - Is missing is stash_id'
3. Click on 'Scrape All'
4. You will then see that the tags are unsorted

**Expected behavior**
To view scraped tags in alphabetical order

**Screenshots**
![2024-03-13_16-00](https://github.com/stashapp/stash/assets/2160922/f5442f9c-e6f6-46ec-83bb-c51ca64aa65a)

**Stash Version: (from Settings -> About):**
v0.25.1
Build hash:
bf7cb78d
Docker Build running on Synology NAS

**Desktop (please complete the following information):**
 - OS: nixOS Linux
 - Browser Chrome 
 - Version Version 122.0.6261.111 (Official Build) (64-bit)

**Additional context**

PR #5752, opened 2025-03-20
Sort tags by name while scraping or merging scenes

its-josh4:
When scraping or merging scenes, tags are displayed in a "random" order (technically, by their numeric ID). This makes it harder to compare tags from before/after, to see what's changed.

This PR sorts tags that are returned when scraping or merging a scene, so it's a lot easier to see differences.

Example while scraping a scene.

Before:

before

After:

after

Closes https://github.com/stashapp/stash/issues/4684


pkg/models/model_tag.go
```diff
@@ -46,6 +47,20 @@ func (s *Tag) LoadChildIDs(ctx context.Context, l TagRelationLoader) error {
...
+	if sortA == "" {
+		sortA = a.Name
+	}
+
+	sortB := b.SortName
+	if sortB == "" {
+		sortB = b.Name
+	}
+
+	return strings.Compare(strings.ToLower(sortA), strings.ToLower(sortB))
```
Maista6969, line 61: I would suggest that we do not lowercase these, at least not in all cases: if someone has set a SortName it's because they want full control over how their tags are sorted and this will override that

Maybe only lowercase the tagname but not the sortname?
```go
sortA := a.SortName
if sortA == "" {
  sortA = strings.ToLower(a.Name)
}
```

its-josh4: Makes sense, let me push that change

feederbox826:
closes https://github.com/stashapp/stash/issues/4684

internal/api/resolver_query_find_tag.go
```diff
@@ -46,6 +47,8 @@ func (r *queryResolver) FindTags(ctx context.Context, tagFilter *models.TagFilte
 			return err
 		}
 
+		slices.SortFunc(tags, models.TagSortFunction)
+
```
WithoutPants, line 51: Sorting is provided by the `filter` parameter. This will override that.

WithoutPants:
I resolved the merge sorting issue on the front end by sorting tags gotten with ids by their sort name/name.

Issue #5775, opened 2025-03-27
[Bug Report] Scene Export does not include the marker end time and can't be re-imported


BollywoodMaster:
**Describe the bug**
Importing a zip file with a scene fails with an error related to inserting into scene_markers. The error message indicates a FOREIGN KEY constraint failure, suggesting that the referenced scene_id or primary_tag_id does not exist in the database.

Additionally, when exporting a scene from Stash, the end time for scene markers is not included in the exported JSON. This may contribute to import issues or incomplete scene marker data.

**To Reproduce**
Import Failure:

1. Go to the import section of Stash.
2. Attempt to import a zip file with a scene.
3. Observe the import failure and error message.

Export Issue:

1. Export a scene from Stash.
2. Check the exported JSON file for scene markers.
3. Notice that end_seconds for markers is missing.

**Expected behavior**

- Importing a scene JSON should succeed without triggering a FOREIGN KEY constraint failure.
- Exported scene markers should include end time (end_seconds) to ensure proper re-importing.

**Screenshots**
![Image](https://github.com/user-attachments/assets/00fc3ea7-a6d1-4146-8ad7-034f59e5ee8a)
![Image](https://github.com/user-attachments/assets/fbc25ff9-7af9-4815-8fdb-cb6641f1ef7d)

**Stash Version: (from Settings -> About):**
v0.28.1

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Opera One
 - Version: 117.0.5408.53

**Additional context**

- The missing end_seconds in exported JSON may lead to incomplete scene markers when re-importing.
- The import failure suggests a possible orphaned reference issue, where scene_id or primary_tag_id does not exist.
- Possible fixes:
-- Ensure exports include end_seconds for scene markers.
-- Validate foreign key constraints before inserting scene markers.
-- Provide a fallback mechanism when importing markers without end_seconds.

PR #5777, opened 2025-03-27
Add marker end seconds import/export

bob123491234:
Addresses part of #5775

Error mentioned in #5775 occurs when attempting to import a new marker to an already existing scene with `Duplicate object handling` set to Ignore (the default value). That is not currently addressed in this PR.


bob123491234:
Added a check to skip attempting to add markers if the scene wasn't imported.
Resolves #5775

Issue #2761, opened 2022-07-20
[Bug Report] Home page links are wrong when using a reverse proxy


dominicschaff:
**Describe the bug**
When you have a reverse proxy enabled with a prefix enabled. Then the "View All" buttons on the home page do not use the prefix for the link.

So it should go to: `/stash/images?c=...` but instead goes to `/images?c=`

This issue also occurs on the logout button. It goes to: `/logout` instead of `/stash/logout`

**To Reproduce**
1. Have stash running behind a reverse proxy, and have `X-Forwarded-Prefix` set (to for example `/stash/`).
2. Go to the stash home page (with some customisation panels enabled).
3. Click on "View All"

**Expected behavior**
That the correct prefix is used in the URL.

**Stash Version: (from Settings -> About):**
v0.16.0

Johnnybegood90:
Did you set the ```
external_host
``` in your stash config file?

**Comment by dominicschaff** [comment]
Ah, I didn't I found the documentation for that now. I followed this section on the wiki: https://github.com/stashapp/stash/wiki/FAQ#how-do-i-serve-stash-in-a-subpath it didn't mention the other parameter. But awesome thanks I shall update that config now. I am going to assume that I can use the information from this wiki page: https://github.com/stashapp/stash/wiki/Reverse-proxy

**Comment by dominicschaff** [comment]
I closed it too quickly.

So if I set the `external_host`, to the new domain name (just the domain name) then it doesn't do anything, and I still get that error.

If I set the `external_host` to `external_host: http://zero.local/stash/` (with or without the trailing slash). Then no media loads and the "View all" links still stay the same. (Same with the logout button)

**Comment by zosky** [comment]
i have this setup in place too *& also no external_host* in my config.yml.

there seem to be a few links in the UI that do the same thing; (ignore `X-Forwarded-Prefix` header i am setting in apache?)

**left panel on a movie or scene > file info tab > studio & phash hyperlinks** 

meanwhile, the other tabs, (details > tags & performers) rendered w/ the correct proxyPath

**Comment by StashPowerUser** [comment]
I'm also encountering this bug in v0.19.0

**Comment by CarlNs92891** [comment]
I have the same issue. On top of this, HLS doesn't work with apache prefix either with error 
```
Scene-6486ccb1.js:10          
GET https://<domain>/scene/<scene_id>/stream.m3u8/0.ts?resolution=ORIGINAL 404 (Not Found)
```
which should be 
```
[...]

PR #5791, opened 2025-03-30
Partial fix for #2761 - Add reverse proxy prefix to HLS links

luigi611:
Adds prefix from X-Forwarded-Prefix to links generated in HLS file. Partial fix for issue #2761

### DIFF
--- a/pkg/ffmpeg/stream_segmented.go
+++ b/pkg/ffmpeg/stream_segmented.go
@@ -426,7 +426,7 @@ func serveHLSManifest(sm *StreamManager, w http.ResponseWriter, r *http.Request,
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/ffmpeg/stream_segmented.go
```diff
@@ -426,9 +426,11 @@ func serveHLSManifest(sm *StreamManager, w http.ResponseWriter, r *http.Request,
 		return
 	}
 
+	prefix := strings.TrimRight(r.Header.Get("X-Forwarded-Prefix"), "")
```
WithoutPants, line 429: `strings.TrimRight` with `""` is a no-op, so I'm not sure why this is here.

PR #5807, opened 2025-04-03
Add graphql playground link to tools panel

WithoutPants:
Adds a `GraphQL playground` link to the tools panel.


WithoutPants:
Didn't realise the only section is `Scene Tools`. I'll need to rework.

PR #5820, opened 2025-04-09
Add missing group scraper fields

### LINKED ISSUE #5793: [Bug Report] Group scrapers don't work
**Describe the bug**
Group scrapers don't work at all. This encompasses multiple bugs
- groupByURL index out of range
- `group` field not found

**To Reproduce**
Sample "Group" scraper

```yml
name: GroupScraper

groupByURL:
  - action: scrapeXPath
    url:
      - example.com/group
    scraper: groupScraper
# avoid load errors
# movieByURL:
#   - action: scrapeXPath
#     url:
#       - example.com/movie
#     scraper: movieScraper

xPathScrapers:
  groupScraper:
    movie:
# set to "movie" to avoid load errors
#   group:
      Name:
        fixed: Group Name
  movieScraper:
    movie:
      Name:
        fixed: Movie Name
```

## groupByURL Index out of range

When attemping to scrape with groupByURL, the following error appears
```
scrapeGroupURL: input: scrapeGroupURL Internal system error. Error <runtime error: index out of range [0] with length 0>
runtime error: index out of range [0] with length 0
```

replacing `groupByURL` with `movieByURL` magically fixes this issue and the scene scrapes perfectly fine. `https://http.cat/200` is the test URL, since example.com returns 404 for everything and scrape fails

<details>
<summary>My attempt at problem solving</summary>

scrapeGroupURL leads to two likely locations, graphql (which I don't suspect) and https://github.com/stashapp/stash/blob/develop/internal/api/resolver_query_scraper.go#L102

I couldn't spot anything that particularly stood out compared to ScrapeMovieURL, my best guess because of an index lookup is that when passing it to models.ScrapedGroup, the URLs field is not being looked up properly, leading me down to https://github.com/stashapp/stash/blob/develop/internal/api/scraped_content.go#L164-L171
or 
https://github.com/stashapp/stash/blob/develop/pkg/scraper/scraper.go#L173-L183
but I don't think it got that far, since it errors before it hits 404

</details>

## group field not found

Uncommenting `group` in xPathScrapers/* yields the following error

```
Error loading scraper /config/scrapers/community/dev/dev.yml: yaml: unmarshal errors:
  line 22: field group not found in type scraper.mappedScraper
```

The issue seems to be here: as group is not defined

https://github.com/stashapp/stash/blob/4bfc93b7ae34908b903c5112b2d99d5a163a2d1b/pkg/scraper/mapped.go#L848-L855

## Co-Existence edit:
`movieByURL disallowed if groupByURL is present` is overriden by errors such as, which incorrectly led me to believe that it was allowed on develop

```
Error loading scraper /config/scrapers/community/dev/dev.yml: yaml: unmarshal errors:
  line 8: field movieByURL already set in type scraper.config
  line 16: field group not found in type scraper.mappedScraper
```

WithoutPants:
Resolves #5793

### DIFF
--- a/internal/api/resolver_query_scraper.go
+++ b/internal/api/resolver_query_scraper.go
@@ -101,12 +101,12 @@ func (r *queryResolver) ScrapeMovieURL(ctx context.Context, url string) (*models
 }
 
 func (r *queryResolver) ScrapeGroupURL(ctx context.Context, url string) (*models.ScrapedGroup, error) {
-	content, err := r.scraperCache().ScrapeURL(ctx, url, scraper.ScrapeContentTypeMovie)
+	content, err := r.scraperCache().ScrapeURL(ctx, url, scraper.ScrapeContentTypeGroup)
 	if err != nil {
 		return nil, err
 	}
 
-	ret, err := marshalScrapedMovie(content)
+	ret, err := marshalScrapedGroup(content)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/api/scraped_content.go
+++ b/internal/api/scraped_content.go
@@ -113,7 +113,30 @@ func marshalScrapedMovies(content []scraper.ScrapedContent) ([]*models.ScrapedMo
 		case models.ScrapedMovie:
 			ret = append(ret, &m)
 		default:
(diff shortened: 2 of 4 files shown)


feederbox826 (approved):
Works on manual build

PR #5856, opened 2025-05-03
Adding the ability to support different Haptic Devices

xtc1337:
This PR refactors the `Interactive` class to allow support with different HapticDevice. The api has been worked to mimic the [`HapticDevice`](https://github.com/iveplay/ive-connect/blob/main/src/core/device-interface.ts#L59)  from the [ive-connect](https://github.com/iveplay/ive-connect/tree/main) project. This will allow to support other interfaces like buttplug.io. 

I tried to keep the core changes to a min while supporting the same api. I did remove some methods which may break plugins if they directly interact with `interactive`. This was needed as the api now works with promises so `await` is needed when updating settings (this is handled by the `InteractiveContext`)
This will be needed to continue to add support via the [SIT](https://github.com/blackx69/StashInteractiveTools) plugin.


kclay:
So I simplified the api a bit here. A plugin will now be able to completely replace the `InteractiveAPI` and provide their own setup. Example to follow
The way a developer would use this is as follows:
- In their main  plugin file use `PluginApi.InteractiveUtils.interactiveClientProvider` and supply your correct method.
- In cases where you may need  extra app context you can use the `PathFunction` to provide a hook and say setup your own context for your plugin to reference later down the chain.

```ts
export let setInteractiveClientValue: (val: IInteractiveClient) => void;
InteractiveUtils.interactiveClientProvider = (
  handyKey: string,
  serverOffset: number,
  defaultClientProvider?: IInteractiveClientProvider,
) => {
  if (!defaultClientProvider) {
    throw new Error('invalid setup');
  }

  const client = defaultClientProvider(handyKey, serverOffset);
  const connect = client.connect;
  client.connect = async () => {
    console.log('patching connect method');
    return connect.call(client);
  };
  setInteractiveClientValue(client);
  return client;
};
export function MyProvider({ children }: { children: React.ReactNode }) {
  const [value, setValue] = React.useState();

  // expose setter externally
  setInteractiveClientValue = setValue;

  return <MyContext.Provider interactiveClient={value}>{children}</MyContext.Provider>;
}
patch.before('App', function (props) {
  return [
    {
      children: <MyProvider>{props.children}</MyProvider>,
    },
  ];
});

```

blackx69:
So I simplified the api a bit here. A plugin will now be able to completely replace the `InteractiveAPI` and provide their own setup. Example to follow
The way a developer would use this is as follows:
- In their main  plugin file use `PluginApi.InteractiveUtils.interactiveClientProvider` and supply your correct method.
- In cases where you may need  extra app context you can use the `PathFunction` to provide a hook and say setup your own context for your plugin to reference later down the chain.

```ts
export let setInteractiveClientValue: (val: IInteractiveClient) => void;
InteractiveUtils.interactiveClientProvider = (
  handyKey: string,
  serverOffset: number,
  defaultClientProvider?: IInteractiveClientProvider,
) => {
  if (!defaultClientProvider) {
    throw new Error('invalid setup');
  }

  const client = defaultClientProvider(handyKey, serverOffset);
  const connect = client.connect;
  client.connect = async () => {
    console.log('patching connect method');
    return connect.call(client);
  };
  setInteractiveClientValue(client);
  return client;
};
export function MyProvider({ children }: { children: React.ReactNode }) {
  const [value, setValue] = React.useState();

  // expose setter externally
  setInteractiveClientValue = setValue;

  return <MyContext.Provider interactiveClient={value}>{children}</MyContext.Provider>;
}
patch.before('App', function (props) {
  return [
    {
      children: <MyProvider>{props.children}</MyProvider>,
    },
  ];
});

```

DogmaDragon (changes requested):
Changes needs to be documented in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/UIPluginApi.md.

xtc1337:
> Changes needs to be documented in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/UIPluginApi.md.

Updated, I wasn't too sure on the guidelines but added the types and an example.

DogmaDragon (commented):
Documentation looks good. For code review you will need to wait for the main developer.

WithoutPants (approved):

PR #5879, opened 2025-05-14
Use StashIDPill to show stash IDs in the tagger view

smith113-p:
This is visually nicer, but more importantly, lets you see easily which stash-boxes are already associated with this scene.


smith113-p:
Before:
image

After:
image

PR #5894, opened 2025-05-29
Submit all scene URLs to Stashbox

smith113-p:
Bug #1868
Closes https://github.com/stashapp/stash/issues/5899


smith113-p:
Unfortunately this will no longer automatically treat unmatched URLs as studio URLs:

Draft _Test_ _ StashDB - Google Chrome 2025-05-29 15_35_01

smith113-p:
This is a client-side change only.

Unless you are concerned about stashboxes who haven't been updated with https://github.com/stashapp/stash-box/pull/824?

PR #5900, opened 2025-06-01
feat: Auto-tag new files right after scanning

philMorel:
This PR adds a new auto-tagging option that kicks in for newly scanned files right after a scan finishes.

The main plus is that it's quick, as it just focuses on tagging these fresh files as they come in.

There's a new setting in the scan options so you can turn this feature on or off.

Quick heads-up: An AI helped get the initial code for this started, and it's working fine on my local server.


DogmaDragon (changes requested):
Please document the option in the manual. 

https://github.com/stashapp/stash/tree/develop/ui/v2.5/src/docs/en/Manual

DogmaDragon:
Documentation looks good. For code review you will need to wait for the main developer.

PR #5910, opened 2025-06-05
Performer merge

### LINKED ISSUE #1351: [Feature] Merging performers and scenes
It happens every so often that I apparently have duplicate performers in my list under different aliases. This usually happens because when adding one of them, they were not on Freeones yet and could not be scraped, or because over time their "primary" name has changed, which Stash then sees as a new user because the hash is based on the name.

When I realize this, I find the one that is "wrong", go through all the scenes they were in, remove them as a performer, and add the "right" performer instead and then delete the wrong performer, ending up with only the correct one with all the scenes that were linked to the wrong one merged into the new one. As a programmer I feel stupid doing this.

The same kind of goes for scenes, but with different qualities instead. If I have the same scene twice, but once in 720P, and once in 1080P, they are seen as 2 separate, unrelated scenes.

A way to merge performers. Just reassigning the scenes to the new performer and deleting the old one would be fine. Getting a comparison with "old vs new" for the performer values, just like when scraping, would be nice, but not as required.

As for the scenes: Something similar as with the performer merging would be fine, allow me to compare the 2, chose which fields to keep from which version, and which file to keep.

You could also borrow some experience from media servers and separate scenes and files, this way a scene can have multiple files, and you can have different qualities for 1 scene at the time allowing you to choose which one to play. But this would probably require a huge change in the infrastructure for very little benefit.

### LINKED ISSUE #3231: [Feature] Merge Into option on tags homepage
All of my content was automatically scraped and the tags were automatically added by the scraper. Because of this, I have a lot of tags I'd like to just consolidate under others, so I've been merging a lot of them. Right now it's cumbersome to find a tag to merge, go into it, then think of what kinds of tags to merge into it and type and select for every single one.

When on the Tags overview tab, I'd like to be able to select some or all of them, then in the ... menu select "merge into" and get the same merge into dialoge in the individual tag page, but have it applied to all selected tags. This would allow me to view a selection of tags based on search criteria, select them all, and merge them with a fraction of the clicks.

I've tried doing something through userscripts or plugins but my knowledge is too limited.

sezzim:
Fixes #1351. (The issue is about both performers and scenes, but scene merging was implemented some time ago.)

Also includes (as separate commits) some changes to make the UI of the different merge types consistent, although there's still a difference for scenes, where there's a convenient drop-down, so adding it there seemed preferable to adding a dedicated button.

No tests, and while it worked for me, I don't think that's complete enough coverage of all the fields to call it good. I could probably work on some tests if it generally looks good.


Gykes:
@WithoutPants 

I think this could potentially be a good addition. I know it is currently supported via plugins but since scenes can do it then might be good for performers. Haven't really looked at the code and there are conflicts to take care of.

internal/api/resolver_mutation_performer.go
```diff
@@ -225,16 +228,17 @@ func (r *mutationResolver) handleLegacyURLs(ctx context.Context, performerID int
 	return nil
 }
 
-func (r *mutationResolver) PerformerUpdate(ctx context.Context, input models.PerformerUpdateInput) (*models.Performer, error) {
-	performerID, err := strconv.Atoi(input.ID)
-	if err != nil {
-		return nil, fmt.Errorf("converting id: %w", err)
-	}
+type LegacyURLs struct {
```
WithoutPants: This can be private.

internal/api/resolver_mutation_performer.go
```diff
@@ -259,26 +263,30 @@ func (r *mutationResolver) PerformerUpdate(ctx context.Context, input models.Per
...
 	}
 
-	legacyURL := translator.optionalString(input.URL, "url")
-	legacyTwitter := translator.optionalString(input.Twitter, "twitter")
-	legacyInstagram := translator.optionalString(input.Instagram, "instagram")
+	var legacyURLs = LegacyURLs{
+		URL:       translator.optionalString(input.URL, "url"),
+		Twitter:   translator.optionalString(input.Twitter, "twitter"),
+		Instagram: translator.optionalString(input.Instagram, "instagram"),
+	}
```
WithoutPants: This can be moved outside instead of introducing another return value to this function.

internal/api/resolver_mutation_performer.go
```diff
@@ -381,16 +407,18 @@ func (r *mutationResolver) BulkPerformerUpdate(ctx context.Context, input BulkPe
...
 			return nil, err
 		}
 
 		updatedPerformer.URLs = translator.updateStringsBulk(input.Urls, "urls")
 	}
 
-	legacyURL := translator.optionalString(input.URL, "url")
-	legacyTwitter := translator.optionalString(input.Twitter, "twitter")
-	legacyInstagram := translator.optionalString(input.Instagram, "instagram")
+	var legacyURLs = LegacyURLs{
```
WithoutPants: No need for `var` here.

internal/api/resolver_mutation_performer.go
```diff
@@ -504,3 +532,93 @@ func (r *mutationResolver) PerformersDestroy(ctx context.Context, performerIDs [
...
+		qb := r.repository.Performer
+
+		dest, err = qb.Find(ctx, destID)
+		if err != nil {
+			return fmt.Errorf("finding destination scene ID %d: %w", destID, err)
+		}
+
+		sources, err := qb.FindMany(ctx, srcIDs)
+		if err != nil {
+			return fmt.Errorf("finding source scenes: %w", err)
```
WithoutPants: Typo

internal/api/resolver_mutation_performer.go
```diff
@@ -150,7 +153,7 @@ func (r *mutationResolver) validateNoLegacyURLs(translator changesetTranslator)
 	return nil
 }
 
-func (r *mutationResolver) handleLegacyURLs(ctx context.Context, performerID int, legacyURL, legacyTwitter, legacyInstagram models.OptionalString, updatedPerformer *models.PerformerPartial) error {
+func (r *mutationResolver) handleLegacyURLs(ctx context.Context, performerID int, legacyURLs *LegacyURLs, updatedPerformer *models.PerformerPartial) error {
```
WithoutPants: No need to accept pointer here.

internal/api/resolver_mutation_performer.go
```diff
@@ -504,3 +532,93 @@ func (r *mutationResolver) PerformersDestroy(ctx context.Context, performerIDs [
...
+	}
+
+	// ensure destination is not in source list
+	if slices.Contains(srcIDs, destID) {
+		return nil, errors.New("destination scene cannot be in source list")
+	}
+
+	var values *models.PerformerPartial
+	var imageData []byte
+	var legacyURLs *LegacyURLs
```
WithoutPants: We're only using this to validate that its not used, so this can be declared inside the `if` statement.

internal/api/resolver_mutation_performer.go
```diff
@@ -504,3 +532,93 @@ func (r *mutationResolver) PerformersDestroy(ctx context.Context, performerIDs [
...
+		values = &v
+	}
+
+	var dest *models.Performer
+	if err := r.withTxn(ctx, func(ctx context.Context) error {
+		qb := r.repository.Performer
+
+		dest, err = qb.Find(ctx, destID)
+		if err != nil {
+			return fmt.Errorf("finding destination scene ID %d: %w", destID, err)
```
WithoutPants: Typo

pkg/sqlite/performer.go
```diff
@@ -138,7 +138,7 @@ type performerRowRecord struct {
 	updateRecord
 }
 
-func (r *performerRowRecord) fromPartial(o models.PerformerPartial) {
+func (r *performerRowRecord) fromPartial(o *models.PerformerPartial) {
```
WithoutPants: I don't understand why this was changed to a pointer.

pkg/performer/validate.go
```diff
@@ -66,7 +66,7 @@ func ValidateCreate(ctx context.Context, performer models.Performer, qb models.P
 	return nil
 }
 
-func ValidateUpdate(ctx context.Context, id int, partial models.PerformerPartial, qb models.PerformerReader) error {
+func ValidateUpdate(ctx context.Context, id int, partial *models.PerformerPartial, qb models.PerformerReader) error {
```
WithoutPants: No reason for this to be a pointer.

pkg/sqlite/performer.go
```diff
@@ -864,3 +864,56 @@ func (qb *PerformerStore) FindByStashIDStatus(ctx context.Context, hasStashID bo
...
+	}
+
+	args = append(args, destination)
+	for table, idColumn := range performerTables {
+		_, err := dbWrapper.Exec(ctx, `UPDATE OR IGNORE `+table+`
+SET performer_id = ?
+WHERE performer_id IN `+inBinding+`
+AND NOT EXISTS(SELECT 1 FROM `+table+` o WHERE o.`+idColumn+` = `+table+`.`+idColumn+` AND o.performer_id = ?)`,
+			args...,
+		)
```
WithoutPants, line 931: Needs an explanation comment. This is likely copied from the tag merge code, so an explanation should be added there as well.

ui/v2.5/src/core/StashService.ts
```diff
@@ -1805,7 +1819,6 @@ export const usePerformerDestroy = () =>
       });
       evictQueries(cache, [
         ...performerMutationImpactedQueries,
-        GQL.FindPerformersDocument, // appears with
```
WithoutPants, line 1847: I'm not sure why this was removed.

WithoutPants (changes requested):
For expediency, I've addressed the identified issues in the backend code, and written some unit tests - this will be pushed shortly.

The biggest blocker to merging this is the lack of support for merging custom fields. For this reason, I'll defer this for 0.30. I'll sort out custom field support myself during 0.31 if it's not already addressed.

WithoutPants:
Just an additional testing note: merging from the performers page redirected to the scene list filtering on the destination performer, but showing the performer id rather than the name.

ui/v2.5/src/core/StashService.ts
sezzim: It is already included in `performerMutationImpactedQueries`, but maybe I am misunderstanding something.

sezzim:
> merging from the performers page redirected to the scene list filtering on the destination performer

This should be fixed.

ui/v2.5/graphql/data/tag.graphql
```diff
@@ -67,6 +67,11 @@ fragment TagListData on Tag {
   aliases
   ignore_auto_tag
   favorite
+  stash_ids {
+    endpoint
+    stash_id
+    updated_at
+  }
```
sezzim, line 74: This seems to be a merge conflict (of sorts) between #6398 and #6403.

ui/v2.5/src/core/StashService.ts
```diff
@@ -2150,16 +2160,14 @@ export const useTagsMerge = () =>
...
-
-      const obj = { __typename: "Tag", id: destination };
-      evictTypeFields(
-        cache,
-        tagMutationImpactedTypeFields,
-        cache.identify(obj) // don't evict destination tag
-      );
+      cache.evict({
+        id: cache.identify({ __typename: "Tag", id: destination }),
+      });
```
sezzim, line 473: For consistency, and because I don't think the old logic was correct (it didn't evict the "aliases" field), I changed this to evict the tag rather than update the cache manually.

The "refetch the performer" strategy didn't seem to work, and didn't seem necessary once I added eviction, so I've remove it from all three merge types.

graphql/schema/schema.graphql
```diff
@@ -373,6 +373,7 @@ type Mutation {
   performerDestroy(input: PerformerDestroyInput!): Boolean!
   performersDestroy(ids: [ID!]!): Boolean!
   bulkPerformerUpdate(input: BulkPerformerUpdateInput!): [Performer!]
+  performerMerge(input: PerformerMergeInput!): Performer
```
WithoutPants: This should be `Performer!`

internal/api/resolver_mutation_performer.go
```diff
@@ -506,3 +538,93 @@ func (r *mutationResolver) PerformersDestroy(ctx context.Context, performerIDs [
...
+		sources, err := qb.FindMany(ctx, srcIDs)
+		if err != nil {
+			return fmt.Errorf("finding source performers: %w", err)
+		}
+
+		for _, src := range sources {
+			if err := src.LoadRelationships(ctx, qb); err != nil {
+				return fmt.Errorf("loading performer relationships from %d: %w", src.ID, err)
+			}
+		}
```
WithoutPants: Why are we loading relationships here? We're not actually doing anything with the `sources`.

ui/v2.5/graphql/data/tag.graphql
WithoutPants: Yep, this is a merge error.

WithoutPants (changes requested):
More static review comments. I'll fix these while testing.

ui/v2.5/graphql/data/tag.graphql
WithoutPants: Nevermind, this is necessary for the change to tag merge

PR #5912, opened 2025-06-06
Rating system patched components

QxxxGit:
Patches RatingNumber, RatingStars, and RatingSystem components


DogmaDragon (commented):
Documentation included.

Issue #3875, opened 2023-07-02
[Feature] Display embedded subtitles


cmcglownjr:
**Is your feature request related to a problem? Please describe.**
Currently mkv files that have subtitles embedded will not display those subtitles.

**Describe the solution you'd like**
Stash can display external captions but it should also be able to display embedded captions. Most desktop and mobile media players can display embedded captions so if possible can embedded captions be displayed in the Stash video player.

7x5rg:
+1

LewdCollector:
+1

WithoutPants:
Videojs does not support displaying embedded subtitles.

There's a few possible ways that I can think of to display subtitles on the fly:
- have ffmpeg live transcode and embed the subtitles. This adds overhead that would be undesirable
- enable on-the-fly generation of subtitles. This is fast, but requires a bit of preparatory work before-hand and will need to be cached.

This improvement should be considered alongside the other caption bug-fixing/improvements (see #2619 and #3800).

In the meantime, this could probably be addressed with a plugin to generate srt files from files with embedded subs.

WithoutPants:
$50 bounty assigned (contribution [745825](https://opencollective.com/stashapp/contributions/745825))

kingmadwik:
+1 i need this as well, embedded subs adds to viewing experience. hope this gets included in coming updates

itsCARLsanity:
> In the meantime, this could probably be addressed with a plugin to generate srt files from files with embedded subs.

Anybody know if anyone already implemented this plugin?

kocy33:
I know this isn't a direct solution for displaying embedded subtitles, but I wanted to add my 'two cents' on how I handle the current lack of support for embedded subtitles in MKV files.

I run the following batch script on a Windows machine. It iterates through all subfolders and uses FFmpeg to extract the subtitles, saving them as .srt files named according to the Stash naming convention (as seen in the [Stash documentation](https://docs.stashapp.cc/in-app-manual/captions/)).

Please note that some .ass formatting can 'go awry' or look strange when converted to .srt, especially if the original file defines specific fonts or maps text to specific screen locations. A major benefit of this process, however, is that the original files remain completely unaltered.

I hope this helps someone! Perhaps this logic could even be incorporated into a plugin?

This is the batch file I use:

```
@echo off
:: Support for network paths (UNC)
pushd "%~dp0"

:: /r scans subfolders recursively
:: %%f is the full path to the file
[...]

PR #5918, opened 2025-06-10
Support embedded subtitle (#3875)

philMorel:
This PR adds functionality to extract embedded subtitles(SRT) from video files during scanning. The implementation was generated by AI and has been manually tested to work correctly.

- Store extracted subtitles in generated/subtitles directory
- Support multiple subtitle tracks with language detection

### DIFF
--- a/internal/api/routes_scene.go
+++ b/internal/api/routes_scene.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -422,12 +423,28 @@ func (rs sceneRoutes) Caption(w http.ResponseWriter, r *http.Request, lang strin
 		return
 	}
 
+	config := config.GetInstance()
+	generatedPath := config.GetGeneratedPath()
+
 	for _, caption := range captions {
 		if lang != caption.LanguageCode || ext != caption.CaptionType {
 			continue
 		}
 
-		sub, err := video.ReadSubs(caption.Path(s.Path))
+		var captionPath string
+
+		// Try path based on hash
+		sceneHash := s.GetHash(config.GetVideoFileNamingAlgorithm())
+		possiblePath := filepath.Join(generatedPath, "subtitles", sceneHash, caption.Filename)
+
+		exists, _ := fsutil.FileExists(possiblePath)
+		if exists {
+			captionPath = possiblePath
+		} else {
+			captionPath = caption.Path(s.Path)
+		}
+
+		sub, err := video.ReadSubs(captionPath)
 		if err != nil {
 			logger.Warnf("error while reading subs: %v", err)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
--- a/internal/manager/task_generate_embedded_subtitles.go
+++ b/internal/manager/task_generate_embedded_subtitles.go
@@ -1,0 +1,100 @@
+package manager
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/stashapp/stash/pkg/file/video"
+	"github.com/stashapp/stash/pkg/logger"
+	"github.com/stashapp/stash/pkg/models"
+)
+
+// ExtractEmbeddedSubtitlesTask extracts embedded subtitles from a video file.
+type ExtractEmbeddedSubtitlesTask struct {
+	repository          models.Repository
+	Scene               models.Scene
+	fileNamingAlgorithm models.HashAlgorithm
+	Overwrite           bool
+	VideoFile           *models.VideoFile
+}
+
+func (t *ExtractEmbeddedSubtitlesTask) GetDescription() string {
+	return fmt.Sprintf("Extracting embedded subtitles for %s", t.VideoFile.Path)
+}
+
+func (t *ExtractEmbeddedSubtitlesTask) Start(ctx context.Context) {
+	config := instance.Config
+
+	ffmpegPath := GetInstance().FFMpeg.Path()
+	ffprobePath := GetInstance().FFProbe.Path()
+	generatedPath := config.GetGeneratedPath()
+	videoFile := t.VideoFile
+
+	// Get the scene hash
+	sceneHash := t.Scene.GetHash(t.fileNamingAlgorithm)
+	logger.Debugf("Extracting embedded subtitles for scene with hash: %s", sceneHash)
+
+	logger.Infof("Extracting embedded subtitles from %s", videoFile.Path)
+
+	// Extract embedded subtitles (pass the scene hash)
+	extractedCaptions, err := video.ExtractEmbeddedSubtitles(ctx, videoFile, generatedPath, ffmpegPath, ffprobePath, sceneHash)
+	if err != nil {
+		logger.Errorf("Error extracting embedded subtitles from %s: %v", videoFile.Path, err)
+		return
+	}
+
+	if len(extractedCaptions) == 0 {
+		logger.Infof("No embedded subtitles found in %s", videoFile.Path)
+		return
+	}
+
+	// Associate the extracted captions with the video file
+	if err := t.repository.WithTxn(ctx, func(ctx context.Context) error {
+		fileID := videoFile.ID
+
+		// Get existing captions for the file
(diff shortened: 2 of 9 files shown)


pkg/file/video/caption.go
```diff
@@ -189,3 +192,119 @@ func CleanCaptions(ctx context.Context, f *models.VideoFile, txnMgr txn.Manager,
...
+
+		outputPath := filepath.Join(subDir, outputFilename)
+
+		// Build FFmpeg command to extract subtitle
+		args := []string{
+			"-nostdin",
+			"-y",
+			"-i", videoFile.Path,
+			"-map", fmt.Sprintf("0:%d", stream.Index),
+			"-c:s", outputFormat,
```
feederbox826, line 278: aforementioned srt conversion

pkg/file/video/caption.go
```diff
@@ -189,3 +192,119 @@ func CleanCaptions(ctx context.Context, f *models.VideoFile, txnMgr txn.Manager,
...
+
+	// Process subtitle streams directly without intermediate struct
+	for _, stream := range videoInfo.JSON.Streams {
+		if stream.CodecType != "subtitle" {
+			continue
+		}
+
+		// Skip non-SRT compatible subtitle formats
+		codecName := strings.ToLower(stream.CodecName)
+		if codecName != "subrip" && codecName != "srt" && codecName != "mov_text" && codecName != "text" {
```
feederbox826, line 231: why bother checking for non-srt formats when you convert them to srt with ffmpeg

internal/api/routes_scene.go
```diff
@@ -422,12 +423,28 @@ func (rs sceneRoutes) Caption(w http.ResponseWriter, r *http.Request, lang strin
...
 		if lang != caption.LanguageCode || ext != caption.CaptionType {
 			continue
 		}
 
-		sub, err := video.ReadSubs(caption.Path(s.Path))
+		var captionPath string
+
+		// Try path based on hash
+		sceneHash := s.GetHash(config.GetVideoFileNamingAlgorithm())
+		possiblePath := filepath.Join(generatedPath, "subtitles", sceneHash, caption.Filename)
```
feederbox826, line 438: Filename doesn't match up with your naming convention

internal/manager/task_generate_embedded_subtitles.go
```diff
@@ -0,0 +1,96 @@
...
+		// (avoiding duplicates based on language code and caption type)
+		allCaptions := make([]*models.VideoCaption, len(existingCaptions))
+		copy(allCaptions, existingCaptions)
+		for _, caption := range extractedCaptions {
+			duplicate := false
+			for _, existing := range existingCaptions {
+				if caption.LanguageCode == existing.LanguageCode && caption.CaptionType == existing.CaptionType {
+					// Skip if this language already exists
+					duplicate = true
+					break
```
feederbox826, line 70: duplicate check here but you also write duplicates based on stream ID later on

pkg/file/video/caption.go
```diff
@@ -189,3 +192,119 @@ func CleanCaptions(ctx context.Context, f *models.VideoFile, txnMgr txn.Manager,
...
+					langExists = true
+					break
+				}
+			}
+
+			// If language code already exists, append stream index to make it unique
+			if langExists {
+				lang = fmt.Sprintf("%s_%d", lang, stream.Index)
+			}
+			outputFilename = fmt.Sprintf("%s.%s.%s", fileHash, lang, outputFormat)
```
feederbox826: handling of the code is not any different, why not isolate lang setting

internal/manager/task_scan.go
```diff
@@ -237,6 +237,19 @@ func (f *handlerRequiredFilter) Accept(ctx context.Context, ff models.File) bool
...
 			}
+
+			// Extract embedded subtitles
+			scenes, err := f.SceneFinder.FindByPrimaryFileID(ctx, videoFile.Base().ID)
+			if err == nil && len(scenes) > 0 {
+				taskSubtitles := &ExtractEmbeddedSubtitlesTask{
+					repository:          GetInstance().Repository,
+					Scene:               *scenes[0],
+					fileNamingAlgorithm: f.videoFileNamingAlgorithm,
+					Overwrite:           true,
```
feederbox826: never taken into consideration or used

pkg/scene/delete.go
```diff
@@ -37,6 +37,14 @@ func (d *FileDeleter) MarkGeneratedFiles(scene *models.Scene) error {
 		}
 	}
 
+	subtitlesFolder := filepath.Join(d.Paths.Generated.Subtitles, sceneHash)
+	exists, _ = fsutil.DirExists(subtitlesFolder)
```
feederbox826, line 41: This deletes the entire subtitle folder on clean, I think you mean FileExists

feederbox826 (changes requested):
In my opinion, much better left for a plugin. Files are written to and read from different locations, Overwrite is present and does nothing.

pkg/file/video/caption.go
philMorel: I chose to check for SRT specifically to prevent issues with complex formats like ASS when converted.
While FFmpeg can convert them, advanced features in ASS don't always translate well to simple SRT. My concern was that this could lead to formatting loss or unexpected display in Stash.
Since Stash doesn't currently support these complex formats externally, I felt it was safer to ensure reliable, predictable subtitle display by sticking to SRT only.

pkg/scene/delete.go
philMorel: My intention was indeed to delete the entire subtitlesFolder (/generated/subtitles/{sceneHash}) when a scene is cleaned/deleted. This folder specifically contains all the generated embedded subtitle files for that particular scene.

The fsutil.DirExists check is therefore correct for my purpose, as I want to confirm the existence of this specific directory before attempting to delete it and its contents, ensuring a complete cleanup for that scene.

philMorel:
> In my opinion, much better left for a plugin. Files are written to and read from different locations, Overwrite is present and does nothing.

Thanks for your feedback on considering this for a plugin. I understand your perspective, and it's a valid point for certain functionalities.

However, my main motivation for integrating this directly into Stash, rather than as a plugin, stems from the user experience regarding internal subtitles and file organization.

If this were handled by a plugin, the extracted internal subtitles would typically need to be placed in the same directory as the video file for Stash to recognize and display them as external subtitles. While this works, a significant reason users want to extract internal subtitles (beyond simply not knowing how to do it manually) is to keep their video folders clean and avoid clutter. Adding separate .srt files directly into the video folders, even if extracted by a plugin, would counteract this goal.

By having Stash manage these generated internal subtitles in a designated, separate location (like /generated/subtitles/{sceneHash}), we can offer a cleaner solution that directly addresses the user desire for tidiness, without affecting the primary video folder.

internal/api/routes_scene.go
philMorel: The extractor saves subtitles as `<sceneHash>.<lang>.srt` (e.g. `8e2f1d…​.eng.srt`).  
`routes_scene.go` looks for the same pattern when serving (`generated/subtitles/<sceneHash>/<filename>`).  
So the filename is intentionally aligned with our current naming convention.

internal/manager/task_generate_embedded_subtitles.go
philMorel: Each subtitle stream is given a unique key before saving:

1. If the same language appears twice, we append the stream index (e.g. `eng_3`).  
2. The later duplicate check compares `(languageCode, captionType)` pairs, which now include that index.  

So the second write never collides with an existing entry—the index-suffix guarantees uniqueness.

pkg/file/video/caption.go
philMorel: I’ll refactor the block so the language-selection logic runs once and the filename is built afterward, eliminating the duplicated code.

internal/manager/task_scan.go
philMorel: I’ll remove the unused Overwrite flag.

feederbox826 (commented):
I do not enjoy the back and forth with LLMs and value my time too much to actually review and criticize the code and responses from LLMs, and your time going between myself and the coding agent.

I will withold my approval and/or further criticisms.

PR #5919, opened 2025-06-10
fix(navigation): Enhance entity create, edit, and delete flows

philMorel:
This PR introduces significant improvements to the navigation flow when creating, editing, and deleting entities.

---

### 1. Consistent "Back" Navigation from Edit Pages

* **Problem:** Pressing the browser's back button on an entity's edit page would incorrectly take the user back to the list page, not the detail page they came from.
* **Solution:** This is now fixed by using a URL hash (`#edit`) to toggle the edit mode on the detail page itself. This keeps the browser history clean, ensuring the back button reliably returns the user to the detail view.

### 2. State-Preserving Navigation After Deletion

* **Problem:** After deleting an item, the user would be sent back to a default list view, losing their previous search filters, sort order, and page number.
* **Solution:** We now pass a `returnTo` URL parameter when navigating from a list to a detail page. Upon deletion, the application uses this parameter to navigate back to the list page with its exact previous state preserved.
* ***Note:*** *A simple `history.goBack()` was considered but discarded, as it fails for detail pages opened directly in a new tab, which would have no history to return to and would result in showing a page for the now-deleted item.*

### 3. Improved "Back" Behavior After Creation

* **Problem:** After creating a new entity, pressing the back button would confusingly take the user back to the new/create form.
* **Solution:** This navigation flow has been corrected for a more logical experience post-creation.

---

The initial code for these enhancements was drafted with AI assistance.
The functionality has been tested on my local server and confirmed to be working as expected.

### DIFF
--- a/ui/v2.5/src/components/List/util.ts
+++ b/ui/v2.5/src/components/List/util.ts
@@ -1,4 +1,10 @@
-import { useCallback, useContext, useEffect, useMemo, useState } from "react";
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
 import Mousetrap from "mousetrap";
 import { ListFilterModel } from "src/models/list-filter/filter";
 import { useHistory, useLocation } from "react-router-dom";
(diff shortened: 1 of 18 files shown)


ui/v2.5/src/components/Tags/TagList.tsx
```diff
@@ -231,7 +232,7 @@ export const TagList: React.FC<ITagList> = ({ filterHook, alterQuery }) => {
                 </Button>
                 <Button variant="secondary" className="tag-list-button">
                   <Link
-                    to={NavUtils.makeTagScenesUrl(tag)}
+                    to={{ pathname: `/scenes`, search: `${NavUtils.makeTagScenesUrl(tag).split('?')[1] ?? ''}&returnTo=${encodeURIComponent(location.pathname + location.search)}` }}
```
feederbox826: ?

ui/v2.5/src/components/Tags/TagList.tsx
```diff
@@ -216,10 +216,11 @@ export const TagList: React.FC<ITagList> = ({ filterHook, alterQuery }) => {
           </ModalComponent>
         );
 
+        const location = useLocation();
         const tagElements = result.data.findTags.tags.map((tag) => {
           return (
             <div key={tag.id} className="tag-list-row row">
-              <Link to={`/tags/${tag.id}`}>{tag.name}</Link>
+              <Link to={{ pathname: `/tags/${tag.id}`, search: `?returnTo=${encodeURIComponent(location.pathname + location.search)}` }}>{tag.name}</Link>
```
feederbox826: ???

ui/v2.5/src/components/Studios/StudioCard.tsx
```diff
@@ -203,7 +204,7 @@ export const StudioCard: React.FC<IProps> = ({
   return (
     <GridCard
       className={`studio-card zoom-${zoomIndex}`}
-      url={`/studios/${studio.id}`}
+      url={{ pathname: `/studios/${studio.id}`, search: `?returnTo=${encodeURIComponent(location.pathname + location.search)}` }}
```
feederbox826: ?

ui/v2.5/src/components/Performers/PerformerCard.tsx
```diff
@@ -305,11 +305,12 @@ export const PerformerCard: React.FC<IPerformerCardProps> = PatchComponent(
       onSelectedChanged,
       zoomIndex,
     } = props;
+    const location = useLocation();
 
     return (
       <GridCard
         className={`performer-card zoom-${zoomIndex}`}
-        url={`/performers/${performer.id}`}
+        url={{ pathname: `/performers/${performer.id}`, search: `?returnTo=${encodeURIComponent(location.pathname + location.search)}` }}
```
feederbox826: ?

ui/v2.5/src/components/List/util.ts
```diff
@@ -25,6 +25,25 @@ export function useFilterURL(
...
+    for (const param of nonFilterParams) {
+      const value = currentSearchParams.get(param);
+      if (value !== null) {
+        newSearchParams.set(param, value);
+      }
+    }
+    
+    return newSearchParams.toString();
+  }, [location.search]);
+
```
feederbox826, line 42: an entire iteration and string to object loop just for a single parameter container in an array?

feederbox826 (changes requested):
half of the commits aren't linked directly, the searchParams are handled in so many different ways and inconsistently 

Also unable to replicate `#edit` hash because it's never set.

philMorel:
> half of the commits aren't linked directly, the searchParams are handled in so many different ways and inconsistently
> 
> Also unable to replicate `#edit` hash because it's never set.

Thanks for the feedback on the commit history. You're right, it looks a bit messy.

During the development of these navigation enhancements, I encountered several unexpected bugs that required iterative fixes and refactoring, which unfortunately led to a somewhat cluttered commit history. My apologies for that.

Regarding the searchParams being handled inconsistently, I appreciate you bringing that to my attention. My approach during development was somewhat fragmented, and I agree that a more unified and consistent way of managing searchParams is needed for better maintainability. I will work on refactoring these parts to use a centralized utility or approach.

Finally, regarding #edit hash not being replicable: The #edit hash should appear in the URL when you click the 'Edit' button on the detail pages for group, performer, studio, and tag entities.

feederbox826 (commented):
I do not enjoy the back and forth with LLMs and value my time too much to actually review and criticize the code and responses from LLMs, and your time going between myself and the coding agent.

I will withold my approval and/or further criticisms.

PR #5923, opened 2025-06-12
Display mode options dropdown

WithoutPants:
Consolidates the display mode options and zoom slider into a single drop down menu:

image

Zoom slider has been restyled and only shows if the display mode is grid.


cj12312021:
This functions fine from my testing.

PR #5924, opened 2025-06-13
Move pagination to a sticky bottom toolbar on scenes page

WithoutPants:
Part of the query page redesign. This change is currently limited to the scenes page only. This removes the top pagination component, and the bottom results summary. The bottom pagination is now shown in a sticky toolbar at the bottom of the screen.

stash - scenes page2-pagination

On mobile, the pagination is placed above the navbar:

image


cj12312021:
I am a bit concerned about the potential overuse of stickies that will eventually be introduced. The consolation is that the sticky can be disabled using custom CSS. 

I went ahead and tested this change with my monitor's resolution bumped down to 1080p, and so far, things don't feel too cramped yet.

JanJastrow:
The bar seems to be floating 1px above the bottom edge - so the content shines through.

Screenshot 2025-06-20 at 08 54 09

WithoutPants:
On what browser and platform is this occurring?

JanJastrow:
Sorry, forget the details.
This is in Vivaldi on macOS & Win10.

Custom CSS is disabled. Cache cleaned.

Firefox, Safari and Thorium seems to be fine.

v0.28.1-60-gc5fe6748 - c5fe6748

PR #5935, opened 2025-06-17
Show gallery cover on the edit panel

sezzim:
This displays the gallery cover on the edit panel, similar to where the scene cover is shown for scenes. It's not editable from there, which I guess could be confusing. I wasn't sure of a good way to add some hint about how to set the cover. But I do add a note about how to set the cover to the in-app manual.

### DIFF
--- a/internal/api/urlbuilders/gallery.go
+++ b/internal/api/urlbuilders/gallery.go
@@ -9,12 +9,14 @@ import (
 type GalleryURLBuilder struct {
 	BaseURL   string
 	GalleryID string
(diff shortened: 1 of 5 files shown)


WithoutPants (approved):
Looks fine. Ideally there'd be some mechanism to select an image as cover, but that isn't necessary for this PR. 

I'd be thinking something like clicking a `Set cover...` button which would allow the user to click on the image in the list and the image would be set as the cover. Doing so inside an edit operation would potentially be tricky though.

Issue #5824, opened 2025-04-12
[Bug Report] Lightbox of galleries on performers detail page is broken


vancrize:
**Describe the bug**
When on a performers page and having the layout set to image tiles, clicking on a image instead of opening the lightbox, like on the regular gallery overview, the page empties itself and errors are triggered.

When on /galleries, the handling and lightbox is working as intended.

I reproduced it in 2 Stash instances, even one that never handled images before, thus has no interfering plugins, settings or custom code running.

**To Reproduce**
1. Go to a performer page that has galleries
2. Click on the tile layout icon (the last square one)
3. Click on a image (not the title)
4. See the empty page. #root node in DOM is empty

**Expected behavior**
On click of one of the tiles, usually a lightbox opens, like on /galleries

**Screenshots**
Layout:
![Image](https://github.com/user-attachments/assets/ca11715a-8358-4b6c-a222-93ed8cea45cd)
Empty node:
![Image](https://github.com/user-attachments/assets/7dc1d99b-c23a-4476-8ff0-00414151f949)
Errors:
![Image](https://github.com/user-attachments/assets/d2b65e73-ee5e-46fc-9bd0-26f0127e23c1)

**Stash Version: (from Settings -> About):**
Tested on 0.28.0 upwards, including v0.28.1-26-gf9fb33e8

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome, Firefox
 - Version: Latest at 2025-04-12

**Smartphone (please complete the following information):**
 - Device: Android Nothing Phone (1)
 - OS: Android 15
 - Browser: Chrome for Android, Firefox
 - Version: Latest at 2025-04-12
**Additional context**

vancrize:
In the meantime i helped myself with a little custom JS, that builds a query for the /galleries page, so that its one click to get to a working gallery overview of a performer. It extracts the performer name and id and starts a prefilled search for galleries with that particular performer.

It adds a new button "Find Galleries" in the tab bar, which leads you to a /galleries result page

![Image](https://github.com/user-attachments/assets/4509dd6a-b7fa-4e15-a9ec-180afdb52765)

`(function () {
    'use strict';

    let currentUrl = location.href;

    function handlePerformerPage() {
        if (!location.pathname.includes("/performers/")) return;
        waitForNavTabsAndInsertLink();
    }

    function waitForNavTabsAndInsertLink() {
        const checkAndInsert = () => {
[...]

PR #5936, opened 2025-06-17
Add useLightbox and useGalleryLightbox in plugin api

QxxxGit:
Adds useLightbox and useGalleryLightbox hooks to the plugin api. Also added to docs.

Tested with my gallery card plugin.

### DIFF
--- a/ui/v2.5/src/docs/en/Manual/UIPluginApi.md
+++ b/ui/v2.5/src/docs/en/Manual/UIPluginApi.md
@@ -83,6 +83,8 @@ Returns a `Promise<void>` that resolves when all of the components have been loa
 ### `hooks`
 
 This namespace provides access to the following core utility hooks:
+- `useGalleryLightbox`
+- `useLightbox`
 - `useSpriteInfo`
 - `useToast`
 
--- a/ui/v2.5/src/pluginApi.d.ts
+++ b/ui/v2.5/src/pluginApi.d.ts
@@ -1046,6 +1046,16 @@ declare namespace PluginApi {
       uploadScript: (funscriptPath: string) => Promise<void>;
       sync: () => Promise<void>;
     };
+
+    function useLightbox(): {
+      state: any;
+      chapters: any;
+    };
+
+    function useGalleryLightbox(): {
+      id: string;
+      chapters: any;
+    };
   }
   namespace patch {
     function before(target: PatchableComponentNames, fn: Function): void;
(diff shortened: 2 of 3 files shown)


QxxxGit:
Through further testing, I'm noticing a bug with my plugin and I'm not certain if it's related to this PR but I wanted to mention it just in case I've missed something.

When I navigate to the gallery from the card, then hit back, then try to use the gallery lightbox, I get a blank page with some errors in the console.

Starts with

```
TypeError: r is undefined
    to http://localhost:9996/assets/Lightbox-5bdc230c.js:1
```

then a couple errors of

```
Error: [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.
```

My plugin code:
```js
import { components, libraries, React, hooks } from "../globals";
import IGalleryCardProps from "../types/IGalleryCardProps";

const GalleryCardImage: React.FC<{
	props: IGalleryCardProps;
}> = ({ props }) => {
	const { Icon } = components;
	const { faMagnifyingGlass } = libraries.FontAwesomeSolid;
	const [isHovered, setIsHovered] = React.useState(false);
	const { useGalleryLightbox } = hooks;
	const showLightbox = useGalleryLightbox(props.gallery.id, []);

	const openLightbox = async (e: React.MouseEvent) => {
		if (props.gallery.image_count === 0) return;
		e.stopPropagation();
		e.preventDefault();

		showLightbox(0);
	};

	return (
		<div
			className="gallery-card-cover"
			onMouseEnter={() => setIsHovered(true)}
			onMouseLeave={() => setIsHovered(false)}>
			<img
				src={props.gallery.paths?.cover}
				className="gallery-card-image"
			/>
			<div className={`overlay ${isHovered ? "show" : ""}`}>
				<div className="button" onClick={openLightbox}>
					<Icon icon={faMagnifyingGlass} className="icon" />
				</div>
			</div>
		</div>
	);
};

export default GalleryCardImage;
```

WithoutPants:
@QxxxGit I think I'll need more of the plugin files to test this. I haven't been able to create a working plugin using the code fragment above.

WithoutPants:
I managed to get it working to the point where I could see the magnifying glass button and could bring up the lightbox. Running a quick test as described, I wasn't able to reproduce the issue. It does sound very similar to the issue reported [here](https://discourse.stashapp.cc/t/galleries-in-wall-mode-under-performers-stopped-working/2185/4), and I'm wondering if it's related to another plugin. Have you disabled all other plugins?

WithoutPants (approved):

WithoutPants:
The issue you describe is already reported in #5824 so should be unrelated to this change.

Issue #5949, opened 2025-06-21
[Bug Report] Deleting a scene leaves artifacts in DB


ikmckenz:
**Describe the bug**
When deleting a scene, even when selecting "Delete generated supporting files", there are still entries in the `files` and `files_fingerprints` tables in the db. This means on a new scan these files go down the `onExistingFile` function (which itself could also then go down `onUnchangedFile`)  rather than through the `onNewFile` function. This is mostly fine for the behavior as-is and avoids recalculating oshash, but it has implications for scan filters, such as in PR #5950. 

**To Reproduce**
1. Scan files
2. Remove scenes and select "Delete generated supporting files"
3. Run scan again 

**Stash Version: (from Settings -> About):**
v0.28.1-63 (building off most recent commit)

WithoutPants:
`Delete generated supporting files` relates to the generated files such as previews, thumbnails etc and is not related the internal file entries in the database.

PR #5950, opened 2025-06-21
Add support to filter out short video clips in scan

### LINKED ISSUE #3160: Allow filtering/ignoring short duration video clips during scanning

Many shared video torrents online may include short clips - either samples, or various types of advertisements (common with JAV).

This means for every actual video/movie, you might have 5-10 various 30-second clips with garbage in them.

It would be good to allow filtering these out during a scan - e.g. filter videos < 30 seconds, or < 200 MB. (Duration probably makes more sense here).

One option might be to filter by name - however, there's no consistency to naming for the sample files, or advertisement files, among the various sharing groups.

**Additional context**

ikmckenz:
Resolves #3160. Tested working locally, there is an associated UI change that looks like so:

Screenshot 2025-06-20 at 6 49 33 PM

The implementation is slightly complicated by #5949 , all three scan paths for a file (`onExistingFile`, `onUnchangedFile`, `onNewFile`) need to support the filtering.

### DIFF
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -517,6 +517,7 @@ type ConfigScrapingResult {
 
 type ConfigDefaultSettingsResult {
   scan: ScanMetadataOptions
+  filter: ScanMetaDataFilterOptions
   identify: IdentifyMetadataTaskOptions
   autoTag: AutoTagMetadataOptions
   generate: GenerateMetadataOptions
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -70,6 +70,15 @@ type GeneratePreviewOptions {
 input ScanMetaDataFilterInput {
   "If set, files with a modification time before this time point are ignored by the scan"
   minModTime: Timestamp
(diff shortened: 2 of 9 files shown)


DogmaDragon:
Does this affect **Scan Video Extensions as Image Clip** functionality as I know some people are using it for their short videos?

ikmckenz:
Looks like as written, yes, these are filtered out since they are still processed as videos internally and so have a duration. Guessing I'll have to figure out how to not do the filtering on image-only libraries when `Scan Video Extensions as Image Clip` is enabled?

DogmaDragon:
I would probably wait for @WithoutPants feedback in general, I just wanted to bring attention to the use case since it can get easily overlooked, but is very important to some people.

WithoutPants:
This approach isn't something I'd be considering. The scan process should be ingesting video files regardless of duration. The intended solution is to provide better control over the creation of scenes from video files. This is still WIP pending features to manipulate assigned files to scenes and images.

ikmckenz:
What are those pending features out of curiosity, any help needed?

Gykes:
Closing as stale.

PR #5957, opened 2025-06-23
Support patching Pagination and PaginationIndex

WithoutPants:


DogmaDragon:
@WithoutPants should be documented in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/UIPluginApi.md.

WithoutPants:
Done https://github.com/stashapp/stash/commit/4b5424dd51ebd3b0c0d216dec37c3dcfd6ea90d4

PR #5969, opened 2025-06-26
Show filter tags in scene list toolbar

WithoutPants:
Adds the current filter tags to the scene list toolbar. If the tags don't fit in the available space, then the list is truncated, with the overflow accessible in a toolbar by hovering over (or clicking on touch devices) the `+x more` tooltip:

image
image

The tags in the tooltip can be interacted with to edit or remove them.

When scrolled to the top of the page, the filter tags are shown as usual and the tags in the toolbar are hidden, only appearing once the tags at the top of the page are out of view:

https://github.com/user-attachments/assets/173a6bbe-484d-4dd4-a0a6-076b4ebce7c7

### DIFF
--- a/ui/v2.5/package.json
+++ b/ui/v2.5/package.json
@@ -60,6 +60,7 @@
     "react-datepicker": "^4.10.0",
     "react-dom": "^17.0.2",
     "react-helmet": "^6.1.0",
+    "react-intersection-observer": "^9.16.0",
     "react-intl": "^6.2.8",
     "react-photo-gallery": "^8.0.0",
     "react-remark": "^2.1.0",
--- a/ui/v2.5/src/components/List/FilterTags.tsx
+++ b/ui/v2.5/src/components/List/FilterTags.tsx
@@ -2,6 +2,7 @@ import React, {
   PropsWithChildren,
   useEffect,
   useLayoutEffect,
+  useReducer,
   useRef,
 } from "react";
 import { Badge, BadgeProps, Button, Overlay, Tooltip } from "react-bootstrap";
@@ -11,7 +12,7 @@ import { Icon } from "../Shared/Icon";
 import { faTimes } from "@fortawesome/free-solid-svg-icons";
 import { BsPrefixProps, ReplaceProps } from "react-bootstrap/esm/helpers";
 import { CustomFieldsCriterion } from "src/models/list-filter/criteria/custom-fields";
-import { useCompare } from "src/hooks/state";
+import { useDebounce } from "src/hooks/debounce";
 
 type TagItemProps = PropsWithChildren<
   ReplaceProps<"span", BsPrefixProps<"span"> & BadgeProps>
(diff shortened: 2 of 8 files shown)


DogmaDragon:
Missing background-color that is used in .popover class for other tooltips.

cj12312021:
Forwarding my Discourse feedback here for visibility: https://discourse.stashapp.cc/t/query-page-redesign/2064/45?u=cj13

WithoutPants:
Per feedback, I've removed the secondary set of filter tags at the top of the page. The filter tags in the toolbar are now shown at all times.

cj12312021:
It seems I reached a consecutive reply limit on the Discourse forum so I'll forward my here:

Here is CSS to make that floating pagination at the bottom look like it's properly floating.
```
.pagination-footer {
	width: fit-content;
	margin: auto;

	.pagination.btn-group {
		box-shadow: 0px 8px 10px 2px rgb(0 0 0 / 30%);

	}
}
```
Before:
Screenshot 2025-06-26 205534

After:
Screenshot 2025-06-26 205852

WithoutPants:
Thanks. Added PR #5973

cj12312021:
I still can't post on the forum due to the same consecutive post limit, so I'm forwarding this here again.

In response to your [question](https://discourse.stashapp.cc/t/query-page-redesign/2064/44?u=cj13), I think something as simple as adding the magnifying glass SVG would suffice here (See image below). Also, the clear text has been feeling a bit out of place with its current styling. In the image below, I attempted to fix this by replacing it with a proper button that should convey the intended point. The button can display the "Clear" text on hover.

Screenshot 2025-06-27 102416

Edit: Regarding the clear text, actually, I realize that text would make sense given the overflow implementation styling. Feel free to disregard that part of this comment.

cj12312021:
I've been playing around with this a bit more. I think we could remove the large edit button from the sidebar and replace it with a slimmer button in the filter bar, as shown in the image below. I recall a user expressing interest in the fact that your original proposal included an edit button outside of the sidebar since they preferred the sidebar closed. However, this is no longer available in your current work. Additionally, I feel we need to add something to that bar to fill the space when no filters are provided. I also think this is a bit more intuitive.
Screenshot 2025-06-28 141223

cj12312021:
In this image, I added a very subtle border around the filter-bar area to emphasize the coexistence of the things in that area. This should make it even clearer to users that the plus button is used to add more filters. Also, note that I removed the round borders on the top right and bottom right areas of the filter button to better blend it with that area. This change, of course, also serves to fill white space.
Screenshot 2025-06-28 144759

With sticky active:
Screenshot 2025-06-28 150457

WithoutPants:
Border looks good. Added that in.

I'm not quite sold on the plus button. The action for the current `Edit filter` button doesn't really match the action inferred by a `+` symbol. I'll continue mulling over the design, as I feel like the filter icon isn't quite correct for it's function either.

WithoutPants:
I think this is probably stable enough to be merged with further refinements done in later PRs.

cj12312021:
It's basically an add filter button. The placement of the "Edit Filter" button, along with the name assigned to it, gives a wrong impression. 

> The action for the current Edit filter button doesn't really match the action inferred by a + symbol.

The "Edit Filter" button really added filters to the list as well as to the new filter bar. Users also know they can click on the filter pill to edit that existing filter. Which makes the purpose of the "Edit Filter" button more confusing.

Users pulling this change for the first time will see the familiar filter button and press it, attempting to open up the filter window. Instead, they will see the sidebar, thinking this is the replacement for the filter window. Then they'll notice the sidebar is very limited compared to the filter window they're used to and wonder how to access the missing filter options. They will see the "Edit Filter" button and think it might be a way to expand the filter options on the sidebar, only to find the old filter menu instead. I've gone through this cycle a few times myself despite my involvement in the redesign. 

The inclusion of the add filter button on the filter toolbar should eliminate this multi-step cycle. It should be clear that the button is used to add filters so users can quickly find what they are looking for. It's also right next to the button users are used to clicking, making it easier for their muscle memory to adjust. Users will then wonder what the old filter button does and find out it reveals a pinnable filter menu with additional smarts not offered in the original sidebar window.

This addition should alleviate the need for the walkthrough guide, which I mentioned might be required for users pulling this change ([here](https://discourse.stashapp.cc/t/query-page-redesign/2064/49)). I think this addition would save a bit of trouble for users consuming this work for the first time, having to adjust their muscle memory. It will also be invaluable to users who prefer to have the sidebar closed when possible, as they would no longer have to open and close it to add filters.

cj12312021:
I saw you posted on Discourse 4 days ago, enabling my ability to post there again. I can forward my recent feedback there.

cj12312021:
I've moved this feedback to Discourse. I don't want to hold up this work if you're still on the fence about this change.

PR #5980, opened 2025-06-27
Update HTML/CSS to accommodate mobile devices with notches

### LINKED ISSUE #5979: Top navigation bar cut off on mobile devices with notch/floating island
## Description
When viewing Stash on a mobile device that has a notch or floating island, the layout of the page becomes constrained at the sides. This has the affect of reducing the available content area available for elements on the page to render, but most notable is the abrupt cutoff of the top navigation bar as its width hits the invisible margins imposed by the browser.

## To reproduce
1. Turn your phone to landscape mode
2. Notice the layout abnormalities

## Expected behavior
For pure aesthetic reasons, it'd be nice to have the top bar extend all the way to edge of the screen.

## Screenshots

❌
Image

✅
Image

## Additional context
The solution involves adding `viewport-fit=cover` to the viewport meta tag in the HTML. Applying this will remove the extra margins added by the browser and result in the page extending all the way to the edges like normal.

Unfortunately adding this tag by itself is not enough, as content at the edges will become obscured when scrolling past the notch, this can be avoided by using the `safe-area-inset-*` variables in CSS to make room for any affected elements. This could be used to calculate an amount of padding to be applied to the main content container such that that gallery covers and scene previews aren't clipped when they scroll past the notch.

- [In-depth article from CSS Tricks on the problem + fix](https://css-tricks.com/the-notch-and-css/)
- [Original comment on Discord raising the issue](https://discord.com/channels/559159668438728723/644934273459290145/1372710163517276343)

angrytoenail:
Resolves #5979 

viewportfix


ui/v2.5/src/index.scss
```diff
@@ -61,6 +61,9 @@ body {
 }
 
 .main {
+  padding-left: env(safe-area-inset-left);
+  padding-right: env(safe-area-inset-right);
```
WithoutPants, line 69: This breaks styling on other viewports, since the `safe-area-inset-x` values are 0, when the bootstrap `container-fluid` class uses `15px` as the left and right padding values. It also appeared to provide too much padding on my Pixel device when in landscape mode and added a horizontal scrollbar. If I removed these lines (leaving the `viewport-fit` change), then the page rendered correctly.

I tried the following which worked on my desktop PC, and it eliminated the scrolling on the Pixel, but still rendered with too much left padding in landscape mode:
```
padding-left: calc(max(env(safe-area-inset-left), 15px));
padding-right: calc(max(env(safe-area-inset-right), 15px));
```

WithoutPants (changes requested):
Sorry for the delay in testing this.

Gykes:
Closing as stale.

angrytoenail:
Opps, I hadn't enabled notifications on this alt account and never saw your comments @WithoutPants !

I'm happy to look into the padding issues you mentioned, I'm sure there'll be a way of achieving the right balance of spacing that works across all devices and screen sizes (admittedly I assumed non-iPhone devices would be unaffected).

Would you mind re-opening this? @Gykes

Gykes:
Closing as stale.

PR #5985, opened 2025-06-30
Upgrade to koanf 2.2.1

stashcoder42:
I would like to upgrade mockery to v3, so that I may submit a pull request that has tests that mocks stuff out more than what's possible with v2.  But in doing so I found that it needs koanf to be upgraded first.

After upgrading I built this and started it, did the wizard and then started it again and it seemed to have loaded the config file just fine.


feederbox826:
https://github.com/knadh/koanf/releases/tag/v2.0.0 no functional changes for v2
https://github.com/knadh/koanf/releases/tag/v2.0.2 is directly linked to #5986 / #5981 
https://github.com/knadh/koanf/releases/tag/v2.2.0 is technically breaking as it bumps to golang 1.23 (already up to 1.24 in #5901)

PR #5988, opened 2025-07-01
Fix make generate-dataloaders and the extra time import

stashcoder42:
When you run `make generate-dataloaders` you'll see an extra import to `time`.  I found a workaround for this bug here: https://github.com/vektah/dataloaden/issues/54


Gykes:
Closing as stale.

PR #5989, opened 2025-07-01
Add extra SQLite UDFs

angrytoenail:
Included in this PR are a few additional user-defined functions for SQLite that provide similar functionality to functions in other database systems, but aren't included in SQLite by default. There's precedent for these functions given that Stash already makes use of a few custom functions (`regexp`, `durationToTinyInt`, `phash_distance`, `basename`).

The new functions are:
- `initcap(text)`
  - Converts the first character of each word in a string to uppercase.
- `trim(text)`
  - Removes leading and trailing whitespace from a string.
- `uuid4()`
  - Generates a version 4 UUID.
- `regexp_substr(string text, pattern text)`
  - Extracts a substring from a string using a regular expression.
- `regexp_replace(string text, pattern text, replacement text)`
  - Replaces substrings in a string using a regular expression.
- `regexp_capture(string text, pattern text, index integer)`
  - Extracts a captured group from a string using a regular expression.

The motivation for making these available via Stash (instead of just loading them separately via an independent database connection on a user's local machine) is that they'll be available for use via GraphQL's `querySQL` and `execSQL` mutations.

The addition of these functions would make it possible to achieve much of the same functionality currently provided by plugins. For example, the `regex_*` functions would allow parsing and identifying dates from a filename, `initcap` / `trim` would allow titles to be formatted appropriately and `uuid4` would facilitate creating identifiers consistent with those used throughout Stash.

### DIFF
--- a/pkg/sqlite/functions_test.go
+++ b/pkg/sqlite/functions_test.go
@@ -1,0 +1,72 @@
+package sqlite
+
+import (
+	"database/sql"
+	"regexp"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCustomFunctions(t *testing.T) {
+	db, err := sql.Open(sqlite3Driver, ":memory:")
+	if err != nil {
+		t.Fatalf("Failed to open database: %v", err)
+	}
+	defer db.Close()
+
+	// Test regexp functions
+	t.Run("regexp", func(t *testing.T) {
+		var result bool
+		err := db.QueryRow(`SELECT regexp('foo.*', 'seafood')`).Scan(&result)
+		assert.NoError(t, err)
+		assert.True(t, result)
+	})
+
+	t.Run("regexp_substr", func(t *testing.T) {
+		var result string
+		err := db.QueryRow(`SELECT regexp_substr('seafood', 'foo.')`).Scan(&result)
+		assert.NoError(t, err)
+		assert.Equal(t, "food", result)
+	})
+
+	t.Run("regexp_capture", func(t *testing.T) {
+		var result string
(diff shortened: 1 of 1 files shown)


feederbox826:
I still don't quite see the value of these over just handling them clientside, doing everything in SQL seems a lot more dangerous and unnecessary

aside from that, I dislike and disagree with some of the nomenclature for the functions

- `initcap`, if you wanted it to mirror databases then I think you should keep the consistency of the capitalization as `INITCAP`
- regexp_substr - why not just use rexec_capture and pass in index 1? unless you're returning an array in which ase reverse it and let regex_capture return the index of the array returned

Your test cases also worry me quite a bit as they seem to be very basic and are missing a lot of escapes, quantifiers and special characters

SyZeeGee:
> I still don't quite see the value of these over just handling them clientside, doing everything in SQL seems a lot more dangerous and unnecessary

If some bulk database function was required (operating on a long list or the entire DB,) then the SQL approach would be better / useful.  If that won't ever be the case for someone, then I agree (from that person's perspective)

Given the PR:
> GraphQL's querySQL and execSQL mutations

This is already possible and probably used by some, so I can see the value (necessary, I think?)

feederbox826:
For bulk DB operations, IMO validation should be handled in application side, SQL data should be as pure as possible, transformations and data manipulation should be handled by APIs and higher translation layers

feederbox826:
I don't think most of these are helpful or relevant

- uuid4 - stash-box is on [uuid7](https://github.com/stashapp/stash-box/issues/991) and stashapp uses integers
- regexp - golang regexp is [unique](https://en.wikipedia.org/wiki/Regular_expression#Implementations_and_running_times) as it's guaranteed to run in linear time, dropping backtracking and a handful of other niche features

yes you can run a plugin entirely via gql SQL, but what are you running GQL on?

PR #4781, opened 2024-04-23
Extend orientation / rotation support for files on load

OtterBotSociety:
Potentially fixes #4555, #4738, #3479 by adding support to parse out `rotation` frame side_data from FFMPEG / ffprobe

I manually tested this with a small set of reproducing JPEG images, but have not yet had the chance to try out other image types.

Thanks to @onlyhooops for their repro images, and to @zhyu for calling out the ffprobe tooling.

PR #6006, opened 2025-07-12
Treat images with no exif metadata as well-oriented

OtterBotSociety:
Some images fail to be loaded by `goexif`, but the error raised is not `EOF` -- it's a [hard-coded "failed to find exif intro marker"](https://github.com/rwcarlsen/goexif/blob/9e8deecbddbd4989a3e8d003684b783412b41e7a/exif/exif.go#L645-L653). In turn, the images are falsely assigned width and height of 0, and this causes the `ImageWall` view to fail the [`getCommonHeight` helper](https://github.com/neptunian/react-photo-gallery/blob/0bb8e4c4a027c021f8a5a06de71e89026596fd95/src/layouts/justified.js#L9-L13) returning NaN. In turn, this leaves all images on any `ImageWall` including at least one bad-EXIF image with width and height of `NaN`.

### DIFF
--- a/pkg/file/image/orientation.go
+++ b/pkg/file/image/orientation.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/rwcarlsen/goexif/exif"
 	"github.com/stashapp/stash/pkg/logger"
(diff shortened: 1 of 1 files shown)


OtterBotSociety:
Related to the same problem with ImageWall that led me to dig in to #4781

PR #6023, opened 2025-07-24
exclude empty regex exclude

### LINKED ISSUE #6022: [Bug Report] Blank exclusion regexes match every file
**Describe the bug**
When the library exclusion list includes blank regex entries, the Clean action will match every file in the library.

**To Reproduce**
1. Edit the exclusions list of the library
1. Add two entries (any pattern)
1. Remove the first regex textbox's value (do not click the Remove button)
1. Close the modal
1. Perform a Clean task

Image

**Expected behavior**
Blank regexes should be ignored during Clean and probably removed from the exclusion list upon Modal close.

feederbox826:
closes #6022

### DIFF
--- a/internal/manager/exclude_files.go
+++ b/internal/manager/exclude_files.go
@@ -60,6 +60,10 @@ func generateRegexps(patterns []string) []*regexp.Regexp {
 	var fileRegexps []*regexp.Regexp
 
 	for _, pattern := range patterns {
+		if pattern == " " {
+			logger.Warnf("Skipping empty exclude pattern")
+			continue
+		}
 		if !strings.HasPrefix(pattern, "(?i)") {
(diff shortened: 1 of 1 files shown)


internal/manager/exclude_files.go
```diff
@@ -60,6 +60,10 @@ func generateRegexps(patterns []string) []*regexp.Regexp {
 	var fileRegexps []*regexp.Regexp
 
 	for _, pattern := range patterns {
+		if pattern == " " {
```
embramere: Please see my update to https://github.com/stashapp/stash/issues/6022#issuecomment-3114024519

I think you'll need to change this to exclude `" "`, `""`, and `NULL` (depending on what the db allows in that column

PR #6031, opened 2025-07-26
Add media hardware key support

### LINKED ISSUE #5553: [Feature] Using hardware media keys for next/previous video
I wold like to change the video to next and previous by remote control.

The solution may not be so obvious since I'm using Linux (kubuntu) and Firefox.
Firefox has MPRIS support enabled, and the media keys (xf86AudioPlay, xf86AudioNext, xf86AudioPrevious, ect.) are working fine at YouTube.
In Stash works only "Play/Pause" and "Stop" command, when I invoke "Next" or "Previous" I get "No player could handle this command" (I'm using playerctl to check it)

I can't find also any other key shortcuts for next and previous to bind to my remote. The player buttons works only by clicking on them.

feederbox826:
closes #5553


feederbox826:
stopgap in https://github.com/feederbox826/plugins/commit/65b93bc9bb4e99a5647fd303a4e8bb4a5d723cdc since handling is easier

PR #6072, opened 2025-09-06
Moved getFileSize from scan.go into standalone util

gregpetersonanon:
Small change here. Moved `getFileSize` from `scan.go` into the fsutil package since it looks like it's also being called externally. Also, since `scan.go` is currently at 1200+ lines I thought it might help simplify things a bit.


feederbox826:
Not really any point in adjusting the ~20 LOC, maybe for a full refactor for scan.go but not over 20SLOC

PR #6077, opened 2025-09-09
Add separate sidebar toggle button

WithoutPants:
Related Discourse post: https://discourse.stashapp.cc/t/query-page-redesign/2064/79?u=withoutpants

- changes the behaviour of the filter icon button on the scene query page to show the filter dialog
- adds a sidebar toggle button to the right sidebar of the scene query page toolbar

image

### DIFF
--- a/ui/v2.5/src/components/List/FilterTags.tsx
+++ b/ui/v2.5/src/components/List/FilterTags.tsx
@@ -167,7 +167,7 @@ export const FilterTags: React.FC<IFilterTagsProps> = ({
         return (child as HTMLElement).classList.contains("more-tags");
       });
 
-      if (moreTags && !!cutoff) {
+      if (moreTags && cutoff !== undefined) {
         return;
       }
 
@@ -273,7 +273,7 @@ export const FilterTags: React.FC<IFilterTagsProps> = ({
 
   const filterTags = criteria.map((c) => getFilterTags(c)).flat();
 
-  if (cutoff && filterTags.length > cutoff) {
+  if (cutoff !== undefined && filterTags.length > cutoff) {
     const visibleCriteria = filterTags.slice(0, cutoff);
     const hiddenCriteria = filterTags.slice(cutoff);
 
--- a/ui/v2.5/src/components/Scenes/SceneList.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneList.tsx
@@ -368,7 +368,11 @@ const ListToolbarContent: React.FC<{
             onRemoveAll={onRemoveAllCriterion}
             truncateOnOverflow
           />
-          <Button className="minimal sidebar-toggle-button" variant="secondary" onClick={() => onToggleSidebar()}>
+          <Button
+            className="minimal sidebar-toggle-button"
+            variant="secondary"
+            onClick={() => onToggleSidebar()}
+          >
             <Icon icon={faSliders} />
           </Button>
         </div>
@@ -386,6 +390,13 @@ const ListToolbarContent: React.FC<{
           <span>{selectedIds.size} selected</span>
           <Button variant="link" onClick={() => onSelectAll()}>
             <FormattedMessage id="actions.select_all" />
+          </Button>
+          <Button
+            className="minimal sidebar-toggle-button"
+            variant="secondary"
+            onClick={() => onToggleSidebar()}
+          >
+            <Icon icon={faSliders} />
           </Button>
         </div>
       )}
--- a/ui/v2.5/src/components/Scenes/SceneList.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneList.tsx
@@ -357,6 +357,7 @@ const ListToolbarContent: React.FC<{
       className="minimal sidebar-toggle-button ignore-sidebar-outside-click"
       variant="secondary"
       onClick={() => onToggleSidebar()}
+      title={intl.formatMessage({ id: "actions.sidebar.toggle" })}
     >
       <Icon icon={faSliders} />
     </Button>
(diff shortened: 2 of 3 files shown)


cj12312021:
Welcome back. I hope the time off went well.

I’ve tested this branch on my development machine and haven’t noticed any bugs. I’ll ingest this into my production machine and get a better feel for how workable the placement of the sidebar toggle button is during regular use.

cj12312021:
I’ve been playing around with this for a bit. I think swapping the position of the sidebar toggle button and the filter button would be more practical. See image below:

Screenshot 2025-09-12 221044

Additionally, it wasn’t initially clear whether you intended to keep the Edit Filter button in the sidebar, as it also exists on the toolbar.

WithoutPants:
> I’ve been playing around with this for a bit. I think swapping the position of the sidebar toggle button and the filter button would be more practical.

In isolation this sounds ok, but with #6079 adding the search input field back in, to me it makes more sense to group the search field with the filter button. 

> Additionally, it wasn’t initially clear whether you intended to keep the Edit Filter button in the sidebar, as it also exists on the toolbar.

I think I was planning to do this in a separate PR, likely after merging #6079. I intend to hide the button where the filter icon button is visible while the sidebar is visible. This means on smaller viewports where the sidebar pops over the content, the Edit Filter button would still be visible.

If we were to move the filter button to the right, and we remove the Edit Filter button from the sidebar, we end up with a bunch of filter functionality on the left and the user having to mouse over to the right for that button. If we swapped the toggle and filter button positions, then I'd be more inclined to keep the Edit Filter button, even if it means having some redundancy.

cj12312021:
I’m not sure how restoring the search input to the toolbar solves the practicality issue of placing the button to access the sidebar on the opposite side of where the sidebar would be open. The location creates unnecessary friction for users, which worsens as the Stash window becomes wider. The UX gives the impression that it doesn’t want users to use the new sidebar, calling into question why effort was invested in implementing it.

I was not aware that the sidebar style is set to hover over list elements at certain viewport sizes. I think, even regardless of the sidebar button placement discussion, that behavior is unideal. Either way, that behaviour would hide a valuable button. If the filter button were kept on the left side, users would no longer be able to access all the filters while the sidebar is open, further incentivizing users not to bother with the sidebar.

Examining how this is handled on other sites, it appears that Bang removes its left sidebar once it has reached this sweet spot and places the sidebar toggle on the right side of the screen, where the button then reveals the sidebar on the right side of the screen. Gamma opts to keep the sidebar in place until the window reaches mobile width, where the sidebar would take up the full screen real estate, with a button placed on the sidebar to toggle it. 

Of the two behaviors, in my opinion, Gamma’s would be the best option, and was actually the behavior I imagined was already in place. Another option could be to place the sidebar toggle and the filter button on the sidebar when this sweetspot is hit, but that might be a lot of work for a window size that is likely niche. Although it sounds like you already had plans to do this for the edit filter button, so including the sidebar toggle there may not be a stretch.

WithoutPants:
> I’m not sure how restoring the search input to the toolbar solves the practicality issue of placing the button to access the sidebar on the opposite side of where the sidebar would be open.

Restoring the search input to the toolbar isn't intended to solve the issue of the sidebar toggle location. That's solving a separate issue. 

> The location creates unnecessary friction for users, which worsens as the Stash window becomes wider.

I could make the same argument for moving the filter button on the right side, especially given that the search input is returning to the toolbar. I'm imagining that the sidebar toggle button is going to be less frequently used than the filter button will be, so I'd rather minimise the friction of using the filter button over the sidebar toggle button.

> I was not aware that the sidebar style is set to hover over list elements at certain viewport sizes. I think, even regardless of the sidebar button placement discussion, that behavior is unideal.

On smaller viewports, the sidebar uses far too much horizontal screen real estate to not pop over it. I don't see a huge issue in not being able to access some functions while the sidebar is visible on these viewports since it's not expected that the sidebar will be permanently visible in these instances.

> The UX gives the impression that it doesn’t want users to use the new sidebar, calling into question why effort was invested in implementing it.

I suspect this might be the crux of our disagreements. Based on user feedback and my own usage, when making decisions around the design of this, I don't want to assume that the sidebar will be visible all or even most of the time. As such, the intention of the sidebar shifting more to a supplementary role rather than a primary one. The user needs to be able to perform most functions without the sidebar visible, with the sidebar offering shortcuts to functionality (like filtering on tags/performers for example). 

You may have missed the post describing the design iteration, so I'll repost it here: https://discourse.stashapp.cc/t/query-page-redesign/2064/79?u=withoutpants

cj12312021:
> I could make the same argument for moving the filter button on the right side, especially given that the search input is returning to the toolbar. 

This point may serve as an argument to consolidate the search and edit filter button back to the same location on the screen.

We may just have to agree to disagree. The practice of opening a menu at the opposite location to where a user clicks is, in my opinion, fundamentally flawed design. I would be interested to see examples of other sites that do this.

The crux of our disagreement does come down to whether the sidebar should serve a primary role. With the convenience it provides around no longer having to jump in and out of a pop-up window to add filters and view results, as well as the added ability it provides to narrow query filters based on the content in the current list, I don’t see why we wouldn’t move in the direction of trying to make this a more suitable primary function.

> You may have missed the post describing the design iteration, so I'll repost it here: https://discourse.stashapp.cc/t/query-page-redesign/2064/79?u=withoutpants

This post did slip past me. I can have a read through it in a bit.

WithoutPants:
The mouse travel is particularly noticeable on larger screens. I've moved the toggle to the left side on `xl` viewports:

image

I've also adjusted the visibility of the Edit Filter and filter icon buttons based on viewports and whether to sidebar is shown. When the filter icon button is visible, the Edit Filter button is hidden in the sidebar:

image

On smaller screens where the sidebar is visible, the filter icon button is hidden and the Edit Filter button is shown instead to increase the available space for the filter tags:

image

The filter icon button is always visible when the sidebar is hidden.

WithoutPants:
I'm merging this as-is for now, with the view that we can iterate further as necessary. Further discussion can be had in the Discourse thread.

PR #6092, opened 2025-09-17
Edit filter load save

WithoutPants:
Related Discourse post: https://discourse.stashapp.cc/t/query-page-redesign/2064/79?u=withoutpants

Adds `Load...` and `Save...` buttons to the Edit Filter dialog:

image

This provides a definitive and context-appropriate location for these functions. The new scene list design puts these functions into the sidebar, which may not necessarily be visible in all viewports.

Clicking the save button opens the same dialog as clicking Save Filter on the sidebar:

image

Clicking the load button opens a similar dialog to load a filter:

image

### DIFF
--- a/ui/v2.5/src/components/List/EditFilterDialog.tsx
+++ b/ui/v2.5/src/components/List/EditFilterDialog.tsx
@@ -564,10 +564,18 @@ export const EditFilterDialog: React.FC<IEditFilterProps> = ({
         </Modal.Body>
         <Modal.Footer>
           <div>
-            <Button variant="secondary" onClick={() => setShowLoadDialog(true)}>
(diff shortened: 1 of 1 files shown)


WithoutPants:
Thanks for testing :)

> Functionality, everything appears to be working just fine. Visually, it's odd that the “Load filter” and “Save filter” text will always overflow, only rendering as “Load...” and “Save...”. If we are okay with this, it would be worth adding the title property to those buttons so the full text is readable on hover.

`Load...` and `Save...` use ellipses to indicate that another dialog will load, rather than performing an action. I chose not to include `filter` for brevity. Adding a title for clarity should be fine though.

PR #6119, opened 2025-10-03
add content-disposition filename header to streams

### LINKED ISSUE #627: [Feature] Add download option to scene player
Via Discord:
>M4DNZToday at 7:41 PM
hey guys, nice work on the latest version, I have one small request, could we get a "download" option on the scene player?

Add a link to download the file using stash. The `stream.mp4` link will transcode. Want a direct download link.

fancydancers:
Add Content-Disposition filename header to streams so they can be saved with their original filename

Closes #627

### DIFF
--- a/internal/manager/running_streams.go
+++ b/internal/manager/running_streams.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"mime"
 	"net/http"
-	"path"
+	"path/filepath"
 
 	"github.com/stashapp/stash/internal/manager/config"
 	"github.com/stashapp/stash/internal/static"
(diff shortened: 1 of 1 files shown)


smith113-p:
I assume you didn't mean to include the Dockerfile change in this PR?

fancydancers:
> I assume you didn't mean to include the Dockerfile change in this PR?

Correct, I made the changes on the wrong branch by mistake.

feederbox826 (changes requested):
Don't forget quotes and to declare encoding, unless you're doing sanitization as well

fancydancers:
I'm not trying to declare content-type but the name of the file via Content-Disposition
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#as_a_response_header_for_the_main_body
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#filename

I'm just using FormatMediaType since it conveniently handles and encodes non ascii characters in the expected format even though the main purpose is for other parts of the header it works well for this. It would also be fairly simple to expand this to optionally use the scene title or perhaps even template the filename but for now I just set it to use the filename on disk.

here's a few examples
```go
package main

import (
	"fmt"
	"mime"
	"path"
)

var pathnames = []string{
	"/parent/studio.2025.01.01.performers.title.mp4",
	"/parent/studio.2025.01.01.performers.⭐🍆.title.mp4",
	"/parent/[Studio] 2025-01-01 Performer Name - Title.mp4",
}

func main() {
	for _, pathname := range pathnames {
		_, filename := path.Split(pathname)
		contentDisposition := mime.FormatMediaType("inline", map[string]string{"filename": filename})
		fmt.Printf("Content-Disposition: %s\n", contentDisposition)
	}
}
```
output
```
Content-Disposition: inline; filename=studio.2025.01.01.performers.title.mp4
Content-Disposition: inline; filename*=utf-8''studio.2025.01.01.performers.%E2%AD%90%F0%9F%8D%86.title.mp4
Content-Disposition: inline; filename="[Studio] 2025-01-01 Performer Name - Title.mp4"
```
Below you can see a SFW example, basically it will let you save with original filename, instead of stream.mp4 and if you're streaming the file to another player it can also pull the filename from the header.
q

PR #6123, opened 2025-10-06
Prevent scanner from failing when reading file info

gregpetersonanon:
Follow up to #6073. I was still encountering errors but this seems to have fixed it.


pkg/file/video/scan.go
```diff
@@ -29,7 +30,8 @@ func (d *Decorator) Decorate(ctx context.Context, fs models.FS, f models.File) (
 	probe := d.FFProbe
 	videoFile, err := probe.NewVideoFile(base.Path)
 	if err != nil {
-		return f, fmt.Errorf("running ffprobe on %q: %w", base.Path, err)
+		logger.Errorf("running ffprobe on %q: %v", base.Path, err)
+		return f, nil
```
WithoutPants: This is a significant change that could have far reached unexpected impacts. If a video file is corrupt then it shouldn't be processed. This should not halt scanning in the way that the other changes fixed.

pkg/scene/scan.go
```diff
@@ -74,7 +74,7 @@ func (h *ScanHandler) Handle(ctx context.Context, f models.File, oldFile models.
 
 	videoFile, ok := f.(*models.VideoFile)
 	if !ok {
-		return ErrNotVideoFile
+		return nil
```
WithoutPants: This is incorrect.

pkg/file/scan.go
```diff
@@ -239,7 +239,8 @@ func (s *scanJob) queueFileFunc(ctx context.Context, f models.FS, zipFile *scanF
 
 		info, err := d.Info()
 		if err != nil {
-			return fmt.Errorf("reading info for %q: %w", path, err)
+			logger.Errorf("reading info for %q: %v", path, err)
+			return nil
```
WithoutPants, line 243: This is the only change that would impact a scan process being aborted early. I think this is a valid change.

WithoutPants (changes requested):

PR #6133, opened 2025-10-12
Sort duplicate scene groups by total filesize descending

underprovisioned:
This is a new feature I would find personally helpful - I did not see any existing issues or PRs for it, but it looked simple so I just implemented it myself.

This adds a default sort to the Scene Duplicate Checker screen, sorting (in descending order) by the total filesize of all scenes in that duplicate group. Given that there did not appear to be a default sort already, it seemed safe to add one.

I believe that this sort behavior makes the most sense as this allows the end user to easily identify the duplicate groups that would potentially free the most disk space if addressed, which seems like it would be one of the primary use cases (certainly the one I find myself using it for).


WithoutPants:
The default sort returned by the backend is by file path. Sorting by total file size makes more sense to me. Ideally this would be properly done in the backend, but a bandaid fix is good enough for now.

PR #6141, opened 2025-10-15
Update marker background color logic to use primaryTag name instead of title

theqwertyqwert:
Currently marker colors are generated based on unique primary tags in `ScenePlayer.tsx`, but then the colors are fetched based on the title. this changes `markers.ts` to get the color by primary tag


feederbox826:
Your PR contains commits from two users. If merged, both users will be added to contributors and there is no way to remove them. You might want to consider [amending your commit](https://github.com/feederbox826/notes/blob/main/seperate-identity.md#amending-the-last-commit)

WithoutPants (approved):

PR #6152, opened 2025-10-17
Use the merged stash IDs by default

### LINKED ISSUE #6149: Scene merging does not merge stash IDs by default
### Describe the bug

I was mergine some scenes from the duplicate checker the other day. In the merge confirmation dialog, "destination" had a single stash id, "source" had multiple stash IDs, but "destination" was selected by default. That is, a stash ID was lost.

### Steps to reproduce

1. Have a scene with a stash ID from one stash box
2. Have another scene with stash IDs from multiple stash boxes
3. Merge them

### Expected behaviour

Merging scenes should default to merging stash IDs

Image

This code looks suspicious to me:
https://github.com/stashapp/stash/blob/de5a9129b3d01bc758a75d2f69e4279e67a23915/ui/v2.5/src/components/Scenes/SceneMergeDialog.tsx#L305

It will default to destination/originalValue as long as it has at least one stash ID. But newValue contains the merged result, shouldn't the code *always* use newValue?

### Stash version

develop version

### Device details

Chrome stable on Android

### Relevant log output

```shell

```

smith113-p:
Fixes #6149

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneMergeDialog.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneMergeDialog.tsx
@@ -311,8 +311,7 @@ const SceneMergeDetails: React.FC<ISceneMergeDetailsProps> = ({
           .filter((s, index, a) => {
             // remove entries with duplicate endpoints
             return index === a.findIndex((ss) => ss.endpoint === s.endpoint);
-          }),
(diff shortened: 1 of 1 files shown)


smith113-p:
Resulting dialog
Screenshot 2025-10-16 213532

PR #6169, opened 2025-10-22
Fix scene list table styling issues

WithoutPants:
- fixes overflow menu showing under list header
- fixes height of the scene list table not using available space


smith113-p (commented):
Considering that the header is a fixed-ish height, whereas dvh is a percentage, would it better to use something like `calc(100dvh - 50px)`?

PR #6171, opened 2025-10-22
Add O-Count to performer page

### LINKED ISSUE #6126: Add O-Count to Performer Page Header

(Apologies in advance if this is a duplicate, I did a lot of searching but wasn't able to find anything...)

I would love to see a performer's O-Count listed in their performer page header.

Here is a mock-up that shows 2 ideas for where the O-Count could be added. (I'm personally a fan of the one with the icon because: 1. It's more consistent with the display of O-Counts elsewhere. 2. It would not use additional vertical real estate on either desktop or mobile. 3. Most of the other text-based attributes in the header are directly editable by the user, which could give the user the wrong impression about the nature of how O-Count is used.)

Image

As an avid user of the O-Count feature and a frequent visitor of performers' pages, this would give me faster visibility to a performer's O-Count without leaving their page. As far as I'm aware, the fastest way to find a performer's O-Count when starting from their page is to open one of their scenes or images, and then find their performer card in the performer list. Having the O-Count in the performer page header would take away this unnecessary step to find their O-Count.

Additionally, this would make the performer page a better central hub of all a performer's information. All other pieces of data that can appear on a performer's card are available on the performer's page (such as scene count, image count, tag count, group count, etc.). The only exception is O-Count, so it would make the relationship between data displayed on performers' cards and their pages more cohesive.

This could probably be achieved by creating a plug-in, but I still think this would be a valuable addition to the base application.

### Additional context

_No response_

melon-scientist:
Displays a performer's O-Count on their performer page. The indicator uses the same design of the scene/image O-Count button and only appears when the performer has at least one recorded O.

### Screenshots
Star rating with full precision
image

Star rating with half precision
image

Decimal rating
image

Performer with zero O's
image

### Boring details
- The star rating number now has a `min-width`, which prevents the O-Count from bouncing around while hovering over stars.
- The value of `min-width` changes based on which rating precision type is being used, in order to reduce the potential for excess space between the O-Count and the rating system. For example, the `min-width` for full precision is much smaller than for quarter precision, because full only needs to show a single digit, but quarter needs to show up to 3 digits and a decimal point.
- Previously, the element containing the rating number would not exist if the rating was null. In order for the `min-width` to always be applied - even when the rating is null - the rating number element now always exists (it just has no content when null). I checked everywhere that the rating system is used, and wasn't able to find any problems or noticeable layout differences caused by this update, but if this is an unacceptable change, please let me know.
- A new container class has been added called `quality-group`, which is just for making sure the text lines-up between the rating system and the O-Count.

(This is my first time working with TypeScript or React, so I'm not familiar with best practices. I mainly just followed the lead from the surrounding code. If you'd like me to change anything, please let me know!)

---

[Original feature request](https://github.com/stashapp/stash/issues/6126)
[UI discussion](https://discourse.stashapp.cc/t/adding-o-count-to-performer-page/3998/6)

Closes #6126

### DIFF
--- a/ui/v2.5/src/components/Performers/styles.scss
+++ b/ui/v2.5/src/components/Performers/styles.scss
@@ -49,7 +49,7 @@
     }
 
     .rating-stars-precision-quarter .star-rating-number {
-      min-width: 2.0rem;
+      min-width: 2rem;
     }
   }
 
--- a/ui/v2.5/src/components/Shared/Rating/RatingStars.tsx
+++ b/ui/v2.5/src/components/Shared/Rating/RatingStars.tsx
@@ -247,9 +247,7 @@ export const RatingStars = PatchComponent(
         {Array.from(Array(max)).map((value, index) =>
           renderRatingButton(index + 1)
         )}
-        <span className="star-rating-number">
-          {maybeGetStarRatingNumber()}
-        </span>
+        <span className="star-rating-number">{maybeGetStarRatingNumber()}</span>
       </div>
     );
   }


DogmaDragon:
When using decimal rating system, the text of o-counter is not aligned with the text of the decimal rating. 

Preview 
image

Preview with
```
margin-top: auto;
margin-bottom: auto;
```
image

PR #6179, opened 2025-10-22
get sidebar to use all available vertical space

spaceyuck:
The sidebar uses the full vertical space now if needed. The current solution is rather simple, long selectable option lists like tags or performers are show in full, and the whole sidebar is just scrollable (no more sub-scrolling). the section header and active inclussions and exclussions are sticky, so in the list currently scrolled out of, it's always visible whats currently being scrolled and what's already selected.

Screenshot_20260531_144331

The search input is really hard to get sticky, and would probably only be possible through Javascript nightmares.

I've tried keeping to individual lists scrollable, but that comes at various prices like lists getting tiny when multiple lists are open on small screens, or elements with only a few entries getting stretched when multiple are open (because they will always size equally through flexbox).

### DIFF
--- a/ui/v2.5/src/components/List/styles.scss
+++ b/ui/v2.5/src/components/List/styles.scss
@@ -214,6 +214,10 @@ input[type="range"].zoom-slider {
   }
 }
 
+.sidebar-saved-filter-list-container .saved-filter-list {
+  max-height: unset;
+}
+
 .sidebar-saved-filter-list-container .toolbar {
   align-items: center;
   display: flex;
@@ -618,8 +622,6 @@ ul.selectable-list {
 .folder-filter ul {
   list-style-type: none;
   margin-bottom: 0.25rem;
-  max-height: 300px;
-  overflow-y: auto;
   // to prevent unnecessary vertical scrollbar
   padding-inline-start: 0;
   padding-bottom: 0.15rem;
--- a/ui/v2.5/src/components/Shared/CollapseButton.tsx
+++ b/ui/v2.5/src/components/Shared/CollapseButton.tsx
@@ -44,8 +44,8 @@ export const CollapseButton: React.FC<React.PropsWithChildren<IProps>> = (
           <Icon icon={open ? faChevronDown : faChevronRight} fixedWidth />
           <span>{props.text}</span>
         </Button>
+        {props.outsideCollapse}
       </div>
-      {props.outsideCollapse}
       <Collapse in={open} {...props.collapseProps}>
         <div>{props.children}</div>
       </Collapse>
--- a/ui/v2.5/src/components/Shared/styles.scss
+++ b/ui/v2.5/src/components/Shared/styles.scss
@@ -1006,9 +1006,11 @@ button.btn.favorite-button {
   border-bottom: 1px solid $secondary;
 
   .collapse-header {
-    // background-color: $secondary;
-
+    background-color: $body-bg;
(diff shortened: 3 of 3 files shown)


WithoutPants:
This simplest way forward would be to remove (or override to `unset`) the `max-height` in `.sidebar-list-filter ul` and `.sidebar-saved-filter-list-container .saved-filter-list`. This would eliminate the internal scrollbars which admittedly are pretty gross.

If we were to take this approach, I'd want to do some sticky positioning of the section headers and ideally the search input for the section as well. This allows the user to easily filter on whatever criterion they are scrolling through, or close the section to peruse other sections.

WithoutPants:
Here's a quick and dirty prototype: https://github.com/WithoutPants/stash/tree/prototyping/sidebar-no-extra-scrollbars

The headers are stickied, but the search filter inputs are not. The behaviour when closing a scrolled section is a bit weird as well, but probably not enough to be a problem.

spaceyuck:
I played around a bit with that approach. I like scrolling the whole sidebar more than scrolling small individual lists, so definitely an improvement in my book.

Stacking multiple sticky elements is only possible with fixed offsets, unless you want to dive into JS solutions with dynamically calculating offsets and IntersectionObserver fun and the likes (possible, but painful). So essentially, stuff can only be sticky that's within the same element. 

I pulled the included / excluded item list into the collapse header, and now they sticky like the header itself, which seems very helpful when hunting for more items to filter with.
Screenshot_20251030_214825

The search box seems more problematic, since it's a couple of elements and components down the stack. I don't think it's possible to sticky the search box without restructuring things more (especially since the selected item lists make the already stick content dynamic in height), or going the JS route for dynamically calculating offsets always seems. But honestly, that level of effort seems overkill for the search box.

Gykes:
@spaceyuck Are you still around to fix this?

spaceyuck:
Yeah, U had lost sight of this a bit, but hopefully going to spend a bit more time on this soon, at the very least need to check how it's doing with the current state, and look at the whole responsive / mobile perspective too.

Already rebased it onto latest develop as a start.

spaceyuck:
I've tried one more time to get the flexbox magic to work sanely, but it doesn't look like it's possible, so I'm giving up on that approach.

Custom CSS if you just want to try it on your instance:
```
.sidebar {
  display: flex;
  flex-direction: column;
}
.sidebar .sidebar-search-container {
  display: flex;
  margin-bottom: 0.5rem;
}
.sidebar .sidebar-section {
  display: flex;
  flex-direction: column;
}
.sidebar .sidebar-section:has(.collapse.show) {
  flex-grow: 1;
  flex-shrink: 1;
}
.sidebar .sidebar-section > .collapse.show {
  display: flex;
  flex-direction: column;
  flex-grow: 1;
  flex-shrink: 1;
}
.sidebar .sidebar-section > .collapse.show > .queryable-candidate-list,
.sidebar .sidebar-section > .collapse.show > .sidebar-saved-filter-list-container {
  display: flex;
  flex-direction: column;
  flex-grow: 1;
  flex-shrink: 1;
}
.sidebar .sidebar-section > .collapse.show > .queryable-candidate-list > ul,
.sidebar .sidebar-section > .collapse.show > .sidebar-saved-filter-list-container > ul {
  flex-grow: 1;
  flex-shrink: 1;
  flex-basis: 0;
  max-height: max-content;
}
```

PR #6186, opened 2025-10-23
[packaging] switch to pnpm

feederbox826:
- `yarn run` -> `npm run` for portability
- bump node to node-24
- update dependencies to exclude yarn and use corepack instead

code changes
- add react-router-dom types

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -338,7 +338,7 @@ server-clean:
 
 # installs UI dependencies. Run when first cloning repository, or if UI
 # dependencies have changed
-# If DOCKER_BUILD is set, configures pnpm to use a local store to avoid
+# If CI is set, configures pnpm to use a local store to avoid
 # putting .pnpm-store in /stash
 # NOTE: to run in the docker build container, using the existing
 # node_modules folder, rename the .modules.yaml to .modules.yaml.bak
@@ -346,7 +346,7 @@ server-clean:
 # building.
 .PHONY: pre-ui
 pre-ui:
-ifdef DOCKER_BUILD
+ifdef CI
 	cd ui/v2.5 && pnpm config set store-dir ~/.pnpm-store && pnpm install --frozen-lockfile
 else
 	cd ui/v2.5 && pnpm install --frozen-lockfile
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,7 @@ jobs:
         docker run -d --name build --mount type=bind,source="$(pwd)",target=/stash,consistency=delegated --mount type=bind,source="$(pwd)/.go-cache",target=/root/.cache/go-build,consistency=delegated --env OFFICIAL_BUILD=${{ env.official-build }} -w /stash $COMPILER_IMAGE tail -f /dev/null
 
     - name: Pre-install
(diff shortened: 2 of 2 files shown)


ui/v2.5/src/core/StashService.ts
```diff
@@ -1614,7 +1614,7 @@ export const mutateAddGalleryImages = (input: GQL.GalleryAddInput) =>
   });
 
 function evictCover(cache: ApolloCache<GQL.Gallery>, gallery_id: string) {
-  const fields: Pick<Modifiers<GQL.Gallery>, "paths" | "cover"> = {};
+  const fields: Partial<Pick<Modifiers<GQL.Gallery>, "paths" | "cover">> = {};
```
feederbox826, line 1617: Was throwing ts validation error so made it a partial

Makefile
```diff
@@ -340,7 +340,7 @@ server-clean:
 # dependencies have changed
 .PHONY: pre-ui
 pre-ui:
-	cd ui/v2.5 && yarn install --frozen-lockfile
+	cd ui/v2.5 && corepack enable && corepack install && pnpm install --frozen-lockfile
```
feederbox826: excessive since compiler image not updated

WithoutPants: Can you elaborate on this?

mlabrum: FYI corepack is no longer being distributed with nodejs https://nodejs.org/dist/latest/docs/api/corepack.html since v25.

I've been ripping corepack out of our applications and just using pnpm to manage the pnpm version and nodejs version

https://pnpm.io/settings#managepackagemanagerversions
https://pnpm.io/settings#usenodeversion

Though we usually disable the useNodeVersion setting in the docker files

feederbox826: > FYI corepack is no longer being distributed with nodejs https://nodejs.org/dist/latest/docs/api/corepack.html since v25.

I hate it here

feederbox826: > Can you elaborate on this?

Since compiler is mentioned by remote address instead of local build, it's impossible to test a build that affects the compiler image, so make pre-ui was amended to include pnpm.

mlabrum: 100% agree the corepack removal from nodejs core sucks :(

I would suggest keeping the packageManager field in package.json then pnpm will manage its pnpm version, otherwise you'll use latest 
https://pnpm.io/settings#managepackagemanagerversions

feederbox826:
CI fails because https://github.com/stashapp/stash/blob/869cbd496b7f3d0eef08bb020bf91d4dd9301da9/.github/workflows/build.yml#L65C259-L65C274

pulls stashapp/compiler which wasn't updated from the PR and is missing pnpm

Makefile
```diff
@@ -338,9 +338,19 @@ server-clean:
...
 # dependencies have changed
+# If DOCKER_BUILD is set, configures pnpm to use a local store to avoid
+# putting .pnpm-store in /stash
+# NOTE: to run in the docker build container, using the existing
+# node_modules folder, rename the .modules.yaml to .modules.yaml.bak
+# and a new one will be generated. This will need to be reversed after
+# building.
 .PHONY: pre-ui
 pre-ui:
+ifdef DOCKER_BUILD
```
feederbox826: if you wanted, maybe a CI=1 check? that's a bit more standard

Issue #2858, opened 2022-08-27
[Feature] Saved filters: save default scraper


YurikaL:
**Is your feature request related to a problem? Please describe.**
Saved filters are quite useful for scraping and organizing new content, since one can save a view (e.g. tagger view) besides filters. But scraper source isn't saved, and one have to manually select matching scraper every time (unless the desired scraper is not accidentally first in alphabetically sorted scraper list).

**Describe the solution you'd like**
Option to save scraper in saved filters, if tagger view is selected in this filter.

**Describe alternatives you've considered**
Some kind of option to sort scraper list manually, but that would only be useful with only one saved filter for scraping (I have few of them for different types and origins of content).

**Additional context**
<details><summary>Scraper source is just not saved in filter.</summary>

![Saved filter with unmatched scraper source](https://user-images.githubusercontent.com/43180936/187021233-b2007b88-3168-4bc9-8371-6523da939678.png)

</details>

PR #6192, opened 2025-10-24
Remember the selected stash box in scene tagger

### LINKED ISSUE #6191: Scene tagger should remember selected source

Currently, the performer and studio tagger remember the selected stash box (stored in config), but the scene tagger does not (neither in the config nor in the URL). 

I frequently click through to the scene itself to check some details (occasionally on accident), and when I go back, I have to scroll to the top to re-select the stash box/source I was using.

No

### Additional context

There was a little bit of discussion around this in #dev-chat

I played with saving the selected stash box in the URL in https://github.com/smith113-p/stash/commit/3429a72794f219fb8f5bfdb17a518f6f5b3e56a2, but that change needs work (because it conflicts with the filter in the URL), and it is inconsistent with performer/studio tagger.

Changing the scene tagger to store this in the config needs a bit of thought because the studio/performer tagger only supports stash boxes, so the config stores an endpoint URL. But scene tagging also supports fragment scrapers. Maybe those could be stored as a special URL "scraper://123" or something... Or the format could be changed and store "stashBox:https://..." vs "scraper:...".

Some relevant code:
https://github.com/stashapp/stash/blob/869cbd496b7f3d0eef08bb020bf91d4dd9301da9/ui/v2.5/src/components/Tagger/studios/StudioTagger.tsx#L705

https://github.com/stashapp/stash/blob/869cbd496b7f3d0eef08bb020bf91d4dd9301da9/ui/v2.5/src/components/Tagger/studios/Config.tsx#L21

smith113-p:
This stores the selected stash box in the config, the same way that studio and performer tagger do (it uses the same setting).

If a non-stashbox source (scraper) is selected, this is not remembered.

I have removed the `currentSource` dependency because there is no need to
re-run this code when the selected source changes and it caused issues
during testing at one point.

Resolves #6191
Related https://github.com/stashapp/stash/issues/2858

### DIFF
--- a/ui/v2.5/src/components/Tagger/context.tsx
+++ b/ui/v2.5/src/components/Tagger/context.tsx
@@ -174,13 +174,30 @@ export const TaggerContext: React.FC = ({ children }) => {
   }, [Scrapers.data, stashConfig]);
 
   useEffect(() => {
-    if (sources.length && !currentSource) {
+    if (!sources.length) {
+      return;
+    }
+    if (config.selectedEndpoint) {
+      let source = sources.find((s) => s.sourceInput.stash_box_endpoint == config.selectedEndpoint);
+      if (source) {
+        setCurrentSource(source);
+        return;
+      }
+    }
+    if (!currentSource) {
       setCurrentSource(sources[0]);
     }
-  }, [sources, currentSource]);
+  }, [sources, config]);
 
   useEffect(() => {
     setSearchResults({});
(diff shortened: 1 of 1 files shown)


smith113-p:
argh, if I re-add currentSource like lint wants me to, then if I change the stash box in the dropdown, it keeps toggling between the previous and the new stashbox in an infinite loop :(

smith113-p:
OK, figured out how to fix lint without causing that issue

PR #6197, opened 2025-10-26
Add swipe navigation to lightbox

sezzim:
Adds swipe navigation to the lightbox on touch devices. Changes the panning behavior to interrupt panning when the edge of the image is reached. In the implementation, an X/Y position of zero now means centered, and a non-zero position is an offset from that.

Relevant to #2538, but that asks for more than just swipe.

Separate from this PR, I have changes to auto-hide the lightbox controls, and to control wake lock, if that is of interest.

### DIFF
--- a/ui/v2.5/src/hooks/Lightbox/LightboxImage.tsx
+++ b/ui/v2.5/src/hooks/Lightbox/LightboxImage.tsx
@@ -1,4 +1,11 @@
-import React, { useEffect, useRef, useState, useCallback } from "react";
+import React, {
+  MutableRefObject,
+  useEffect,
+  useRef,
+  useState,
+  useCallback,
+} from "react";
+import useResizeObserver from "@react-hook/resize-observer";
 import * as GQL from "src/core/generated-graphql";
 
 const ZOOM_STEP = 1.1;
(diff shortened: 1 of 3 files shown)


WithoutPants:
Thank you for your submission and apologies for the lateness in reviewing this.

I'm running into some issues with this on Chrome on my Android device. Swiping left and right seemed to sometimes not work.  It would work for a run of images then it swiping would just stop responding. I also noticed that its no longer possible to navigate left and right by tapping the image.

> Separate from this PR, I have changes to auto-hide the lightbox controls, and to control wake lock, if that is of interest.

This is indeed of interest.

sezzim:
I think the onTouch handlers I added are interacting badly with the existing onPointer handlers. I will work on fixing it.

sezzim:
Updated. I think this is better, but the consolidation also increases the risk of breakage.

WithoutPants:
Doing a retest, this is still very finicky. Doing quick swipes from would sometimes navigate in the wrong direction - ie swiping quickly from right to left navigates to the previous image, rather than the next.

sezzim:
I pushed a change that excludes touch events from the logic that navigates left/right based on clicking on the left/right half of the screen, do you mind testing if that helps? I can't reproduce this latest issue.

WithoutPants:
> I pushed a change that excludes touch events from the logic that navigates left/right based on clicking on the left/right half of the screen, do you mind testing if that helps? I can't reproduce this latest issue.

This fixed the issue I described, but now I can no longer tap to advance to the next/previous image. Mouse clicking for the next/previous image still works.

ui/v2.5/src/hooks/Lightbox/LightboxImage.tsx
```diff
@@ -422,91 +455,82 @@ export const LightboxImage: React.FC<IProps> = ({
...
-  }
-
-  function onTouchStart(ev: React.TouchEvent) {
-    ev.preventDefault();
-    if (ev.touches.length === 1) {
-      startPoints.current = [ev.touches[0].pageX, ev.touches[0].pageY];
-      setMoving(true);
+    if (ev.pointerType === "touch" && startPoint.current !== null) {
+      // Swipe navigation
+      releaseCarousel(ev, ev.timeStamp - startTime.current, false);
```
WithoutPants: I suspect the fix to the swipe weirdness is to change `releaseCarousel` to return a boolean and return true if a previous/next operation was triggered. If it was, then it should return from this point so that the tap to advance code below isn't executed.

PR #6234, opened 2025-11-05
Add (hidden) pagination to list results header

WithoutPants:
Adds the pagination control back into the top results header on the scenes page.

I am not quite convinced that it should be shown for all users, so for the moment I've made it hidden with `display: none`. It can be displayed with the following custom css:
```
.list-results-header .pagination-index-container div.pagination {
  display: flex;
}
```

When enabled, pagination is shown like below:
image


DogmaDragon:
Custom CSS listed on https://discourse.stashapp.cc/t/custom-css-snippets/4043#p-8143-show-pagination-to-list-results-header-43

PR #6247, opened 2025-11-07
Separate search clear effect from config saving

### LINKED ISSUE #6245: Scrape results cache eviction in tagger view
### Describe the bug

With the latest pre-release, tagger view is broken.

Let's say you click "scrape all" button, if you save any result afterwards, scrape cache is evicted and all scraped results are lost.

EDIT: After some testing, I'm able to "restore" pending results if I trigger some data updates (e.g. creating a missing performer from scrape result while save is in progress, performer update makes results show back up)

EDIT2: After some more testing, cache is only evicted when using stashbox scrapers. Scraper plugins keep all results listed as expected.

### Steps to reproduce

1. Go to 'Tagger view' with at least 2 known scenes in StashDB
2. Click "Scrape all button" or scrape manually at least 2 scenes
3. Save any result
4. All results disappear after saving instead of just the one saved

### Expected behaviour

All cached scrape result are kept intact.

_No response_

### Stash version

v0.29.3-10-g96a7e087

### Device details

Google Chrome Version 141.0.7390.107 (Arch Linux)

### Relevant log output

```shell
N/A
```

WithoutPants:
Fixes Scrape results cache eviction in tagger view 
Fixes #6245

### DIFF
--- a/ui/v2.5/src/components/Tagger/context.tsx
+++ b/ui/v2.5/src/components/Tagger/context.tsx
@@ -173,6 +173,7 @@ export const TaggerContext: React.FC = ({ children }) => {
     setSources(stashboxSources.concat(scraperSources));
   }, [Scrapers.data, stashConfig]);
 
+  // set the current source on load
   useEffect(() => {
     if (!sources.length || currentSource) {
       return;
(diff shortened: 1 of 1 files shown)


smith113-p:
I *think* the reason this fixes it is that saving a scene saves the fingerprint queue to the config, which triggers this effect, and we don't want the result clearing for that, is that right?

Thanks for fixing my bug!

WithoutPants:
Yep, that's right :)

Larsluph:
Can confirm fix, thanks for the quick response!

Issue #5837, opened 2025-04-19
[Feature] More studio scrape fields


feederbox826:
**Is your feature request related to a problem? Please describe.**
Studio scrapers only support Name and URL, it would be nice to have access to more fields like `Details` and `Aliases` via scraper interfaces

**Describe the solution you'd like**
Add Tags, Details, Aliases to URL scraper
add URL(s) array to Studio

**Describe alternatives you've considered**
graphql bypass

**Additional context**

GalacticCheat:
Maista sucks cock
![Image](https://github.com/user-attachments/assets/ed12369e-cb07-434a-b49f-fbc470272035)

Gykes:
@feederbox826 Are you able to give me an example scraper that returns these fields? I have the work done (I think) just want to test it.

feederbox826:
I can whip something up real quick

PR #6249, opened 2025-11-08
Add Details, Tags, and Aliases Scraper Fields for Studios

Gykes:
This adds Details, Tags, and Aliases to studio scrapers. This does not update stashboxs. 

Currently untested. Per the FR I have not added anything with the Studio URL array as that should be taken care of in https://github.com/stashapp/stash/pull/6223

Related: https://github.com/stashapp/stash/issues/5837

### DIFF
--- a/graphql/stash-box/query.graphql
+++ b/graphql/stash-box/query.graphql
@@ -13,6 +13,7 @@ fragment ImageFragment on Image {
 fragment StudioFragment on Studio {
   name
   id
+  aliases
   urls {
     ...URLFragment
   }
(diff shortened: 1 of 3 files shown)


DogmaDragon:
Needs to be documented in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/ScraperDevelopment.md

Gykes:
@DogmaDragon I updated various sections of the docs. Let me know if I missed something. I more than likely did lol

WithoutPants:
While this adds the fields, it doesn't use any of it in the UI. This requires a bit more work which I'll split off into a separate PR. I've added aliases to the returned value when creating a studio from stash-box.

Gykes:
Yep, I only added in the backend support for now. I saw your various other PRs doing backend then front end support seperate.

Issue #4087, opened 2023-09-05
[Bug Report] yml scrapers aren't self-aware of url


scruffynerf:
a yml scraper, despite being called against a url, doesn't have a way to directly return that url as a value,
depending entirely on the resulting page having some form of url content to return.
rl
For most bigger sites where a url is provided, this isn't an issue, but the lesser sites where you can't easily get a url of what you are scraping, this is a problem, and often the result is the scraper doesn't return a url, which can be a problem when doing tasks that expect a url value to be returned.  I've seen this result in scraping and then emptying the url value, which means the scene can't easily be rescraped, for example.

What should happen:
some way to make the yml scraper return the called url AS the URL result, if not otherwise set.

GalacticCheat:
Maista sucks cock
![Image](https://github.com/user-attachments/assets/ed12369e-cb07-434a-b49f-fbc470272035)
https://i.ibb.co/ds49xcXt/image-2-copie.jpg

PR #6250, opened 2025-11-08
Feature: Support inputURL and inputHostname in scrapers

Gykes:
This PR adds a `{inputURL}` placeholder that can be used in both fixed values and selector expressions within YAML scraper configurations. The placeholder is automatically replaced with the actual URL that was used to perform the scrape.

  Implementation Details

  - Added url field to both xpathQuery and jsonQuery structs to store the input URL
  - Added getURL() method to the mappedQuery interface
  - Updated mappedConfig.process() to replace {inputURL} placeholder in both fixed values and selectors
  - Updated all query creation calls to pass the URL parameter through the scraping pipeline
  
  I havent officially tested this as I had so many GH issues. Will do it tomorrow or someone else can test.
  
  Fixes: https://github.com/stashapp/stash/issues/4087

### DIFF
--- a/.idea/go.iml
+++ b/.idea/go.iml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="WEB_MODULE" version="4">
+  <component name="Go" enabled="true" />
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$">
       <excludeFolder url="file://$MODULE_DIR$/certs" />
--- a/pkg/scraper/mapped.go
+++ b/pkg/scraper/mapped.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"net/url"
 	"reflect"
 	"regexp"
 	"strconv"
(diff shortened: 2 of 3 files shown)


DogmaDragon:
Needs to be documented in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/ScraperDevelopment.md.

feederbox826:
Was this less work than the scraping system? probably infinitely. Was it fun making the most obscure hacky workarounds for missing URL? yes.

Gykes:
If there would be a better way to do this as far as usability goes please let me know. I don't use scrapers so not sure if this is a painful or painless implementation.

DogmaDragon:
> Was this less work than the scraping system? probably infinitely. Was it fun making the most obscure hacky workarounds for missing URL? yes.

Ah yes, I remember first seeing @Maista6969 `__SEPARATOR__` trick. My mind was blown. 😆

feederbox826:
> If there would be a better way to do this as far as usability goes please let me know. I don't use scrapers so not sure if this is a painful or painless implementation.

Adding hostname as well as URL would be great, we wouldn't need to parse it out manually

Gykes:
> > If there would be a better way to do this as far as usability goes please let me know. I don't use scrapers so not sure if this is a painful or painless implementation.
> 
> Adding hostname as well as URL would be great, we wouldn't need to parse it out manually

Shouldn't be an issue. I can add it in here or if @WithoutPants wants it in a separate PR I can do that.

WithoutPants:
Adding to this PR is fine.

WithoutPants (approved):

Issue #2643, opened 2022-06-02
[Feature] Add StashID's to Tags


stg-annon:
**Is your feature request related to a problem? Please describe.**
When drafting to a Stash-Box instance, Tags are matched from only the name of tag from the local stash instance, there could be a local alias or remote alias match but this kind of matching should not be done on each draft, instead stash users could opt to manually add a StashID to their local tag to match against a given Stash-Box, or in the future use a tagger to try and manage their tags in bulk

**Describe the solution you'd like**
Add a StashID attribute to Tags in the same way they are used elsewhere, and allow a user to set this ID from either a search against a Stash-Box or manually

**Describe alternatives you've considered**
manually correcting tags every time a draft is submitted, this creates a lot of overhead and is not sustainable

WithoutPants:
$52.34 bounty added (3-Jun-2022)

echo6ix:
Just making a note here, that it might be a good idea to pump the breaks on implementing this before it possibly compounds the issue with StashDB tags.

Reference to the [Discord discussion](https://discord.com/channels/559159668438728723/935611121371611157/1152046020742615161).

Flashy78:
Stashbox PR to allow tags to send StashIDs when submitting a draft from Stash https://github.com/stashapp/stash-box/pull/717

Maista6969:
Flashy78's PR has been merged and associating stash-box IDs with tags would be even more valuable now 👍

WithoutPants:
$50 bounty added (txn [#743436](https://opencollective.com/stashapp/contributions/743436)) - total $102.34.

WithoutPants:
@Gykes you're eligible for this bounty. Please choose one of the following:
- submit an expense to the Open Collective (https://opencollective.com/stashapp/expenses/new) for $87.29 ($102.34 - fees)
- indicate another issue you'd like to reallocate the bounty to, or defer to general funds

Gykes:
Submitted. 

First time doing one so please advise if there are issues or mistakes that need to be fixed going forward.

PR #6255, opened 2025-11-10
Feature: Tag StashID support

Gykes:
This adds the ability to add and store the associated StashIds with tags.

Currently StashIDs are added automatically when you add a tag via tagger or via plugin and GraphQL. 

The only way to add StashId's to currently existing tags is to: 
 - Run Identify and let it rematch
 - Use a plugin to add them
 - Retag it all.....
 
 
 My testing: 
 Scrape a scene via tagger against StashDB. Add tags in the Tagger. Profit
 
 If there is a better way to handle this other than building a dedicated tag tagger similar to studio tagger then let me know.

Fixes: https://github.com/stashapp/stash/issues/2643
UI:
Screenshot 2025-11-09 at 23 27 51

### DIFF
--- a/internal/identify/scene.go
+++ b/internal/identify/scene.go
@@ -163,31 +163,6 @@ func (g sceneRelationships) tags(ctx context.Context) ([]int, error) {
 				return nil, fmt.Errorf("error converting tag ID %s: %w", *t.StoredID, err)
 			}
 
-			// If the scraped tag has a RemoteSiteID and endpoint, update the existing tag to add the StashID
-			if t.RemoteSiteID != nil && endpoint != "" {
-				// Get the tag updater interface
-				if tagUpdater, ok := g.tagCreator.(models.TagUpdater); ok {
-					// Create a partial update to add the StashID
-					partial := models.NewTagPartial()
-					partial.StashIDs = &models.UpdateStashIDs{
-						StashIDs: []models.StashID{
-							{
-								Endpoint:  endpoint,
-								StashID:   *t.RemoteSiteID,
-								UpdatedAt: time.Now(),
-							},
-						},
-						Mode: models.RelationshipUpdateModeAdd,
-					}
-
-					_, err = tagUpdater.UpdatePartial(ctx, int(tagID), partial)
-					if err != nil {
-						// Log the error but don't fail the whole operation
-						logger.Warnf("Failed to update StashID for tag %d: %v", tagID, err)
-					}
-				}
-			}
-
 			tagIDs = sliceutil.AppendUnique(tagIDs, int(tagID))
 		} else if createMissing {
 			newTag := t.ToTag(endpoint, nil)
--- a/pkg/models/jsonschema/tag.go
+++ b/pkg/models/jsonschema/tag.go
@@ -6,20 +6,22 @@ import (
 
 	jsoniter "github.com/json-iterator/go"
 	"github.com/stashapp/stash/pkg/fsutil"
+	"github.com/stashapp/stash/pkg/models"
 	"github.com/stashapp/stash/pkg/models/json"
 )
 
 type Tag struct {
-	Name          string        `json:"name,omitempty"`
-	SortName      string        `json:"sort_name,omitempty"`
-	Description   string        `json:"description,omitempty"`
-	Favorite      bool          `json:"favorite,omitempty"`
-	Aliases       []string      `json:"aliases,omitempty"`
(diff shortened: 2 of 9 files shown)


feederbox826:
Will test it later, havent had a chance to look at the code yet. Would it be possible to update them via GQL? [I have a plugin](https://github.com/feederbox826/plugins/tree/main/plugins/tag-link) that does auto-search and match either via my tag DB or by searching StashDB. Wouldn't recommend it for prod but would be great as a plugin

Gykes:
> Will test it later, havent had a chance to look at the code yet. Would it be possible to update them via GQL? [I have a plugin](https://github.com/feederbox826/plugins/tree/main/plugins/tag-link) that does auto-search and match either via my tag DB or by searching StashDB. Wouldn't recommend it for prod but would be great as a plugin

So, currently you can already do this with this:

New Tag:

``` 
mutation {
    tagCreate(input: {
      name: "Example Tag"
      stash_ids: [
        {
          endpoint: "https://stashdb.org/graphql"
          stash_id: "abc123-def456"
        }
      ]
    }) {
      id
      name
      stash_ids {
        endpoint
        stash_id
      }
    }
  }
```

Update Existing:

```
 mutation {
    tagUpdate(input: {
      id: "123"
      stash_ids: [
        {
          endpoint: "https://stashdb.org/graphql"
          stash_id: "abc123-def456"
        }
      ]
    }) {
      id
      name
      stash_ids {
        endpoint
        stash_id
      }
    }
  }
```

DogmaDragon:
Not user friendly as is.

Also, since the StashID can only be applied via scene tagger (via UI), there is effectively not way have multiple StashIDs linked to the same tag. Be it from the same stash-box instance or multiple ones. 

I think at least manual input field is needed. Though a proper search interface be it tagger or something else would be ideal.

Gykes:
> Not user friendly as is.
> 
> Also, since the StashID can only be applied via scene tagger (via UI), there is effectively not way have multiple StashIDs linked to the same tag. Be it from the same stash-box instance or multiple ones.
> 
> I think at least manual input field is needed. Though a proper search interface be it tagger or something else would be ideal.

I agree with you overall. Currently I kept it the same implementation as the rest of the StashID infrastructure. This could be step 1 of a larger project of eventually adding in a tagger for Tags. I could modify it to allow users to add in StashID's in the tag edit modal as a stopgap till that eventuality. I think it would need to be a conversation with WP for the larger scope of the project.

Edit: Since you can apply it via GraphQL you could also apply them via plugin

internal/identify/scene.go
```diff
@@ -161,12 +163,36 @@ func (g sceneRelationships) tags(ctx context.Context) ([]int, error) {
...
+					}
+
+					_, err = tagUpdater.UpdatePartial(ctx, int(tagID), partial)
+					if err != nil {
+						// Log the error but don't fail the whole operation
+						logger.Warnf("Failed to update StashID for tag %d: %v", tagID, err)
+					}
+				}
+			}
+
```
WithoutPants: I don't think this belongs here. None of the other matching functions add a stash id and I think it's undocumented and unexpected behaviour.

WithoutPants:
> I think at least manual input field is needed.

Agree with Gykes here - let's push manually linking into a separate PR targeting 0.30 release. I'd like to get something in for manually linking stash-ids to all supporting objects (performers, studios, scenes, etc).

internal/identify/scene.go
Gykes: Fair enough. 

The intent was so that users could run rerun identify and tags would automatically get the appropriate stash ID. 

Removed

Gykes:
> Static code review only. Not yet tested.
> 
> Needs the following missing bits:
> 
>     * cover tag stash id table in `anonymise.go`
> 
>     * add `StashIDs` to `jsonschema.Tag` and import/export functionality
> 
>     * add unit test coverage for the new fields, see the performer tests for reference

Fixed, took me a minute to figure out the unit testing but I think it's good?

internal/identify/scene.go
Gykes: Fixed

Gykes:
> > I think at least manual input field is needed.
> 
> Agree with Gykes here - let's push manually linking into a separate PR targeting 0.30 release. I'd like to get something in for manually linking stash-ids to all supporting objects (performers, studios, scenes, etc).

Once you are content with the major implementation I can go ahead and do the UI ability to manually add StashIDs.

PR #6261, opened 2025-11-12
Feature: Add Tagger Shortcut

Gykes:
This allows users to select v + t from the scene page to take them to the tagger similar to the other available shortcuts


WithoutPants:
TODO - good candidate to be backported to 0.29.4.

PR #6262, opened 2025-11-12
Add SFW content mode option

### LINKED ISSUE #2022: [Feature] R18 toggle/SFW toggle
I have both R18 content and SFW content in my Stash instance and would like only SFW content to be shown unless a toggle is enabled.

The navbar could have a toggle to switch between R18 and a "Safe" mode. When safe mode is enabled, the each page (eg. actors, scenes, images) will either only show SFW content or blur NSFW content.
Being SFW is strictly opt-in, meaning that unorganized content will not show up when the safe mode is enabled.

Add an allow/block list for tags to the settings. If content has a tag that is in a blocklist, do not show it. If the allow list contains a tag, only show content that has at least one of the tags in this list.

This solution is more generic and useful in more cases (for example hiding kink content), but adding/removing tags to/from a list is more of a chore. Additionally, having a natively supported "SFW" option is cleaner than adding a tag.

**Additional context**
none

---

I understand that Stash is primarily meant for porn, but I have some content stored in it that is adult-oriented in nature but not inherently explicit. Being able to hide explicit content is would be very useful to me.
A safe mode may also be useful for video showcases and screenshots.

WithoutPants:
Adds a `SFW content mode` option, found in `Settings -> Interface`:

image

Also adds an option to set the mode in the setup wizard:
image

When this option is enabled, the following changes are made:
- the following porn-specific performer fields are hidden:
  - ~~death date~~
  - ethnicity
  - ~~height~~
  - weight
  - hair color
  - eye color
  - measurements
  - penis length
  - fake tits
  - circumcised
- `o-count` is renamed to `likes` and uses the thumb up icon
- default performer image is replaced with a generic user icon - the same used for the performers navbar icon

Also adjusts the wording in the README to reflect that stash supports non-porn-related content.

Screenshots:

image

image

image

Closes https://github.com/stashapp/stash/issues/2022


smith113-p:
That list of performer attributes seems too big to me. For example, IMDb has birthday, death date, height for this definitely not porn performer:
https://www.imdb.com/name/nm0000026/bio/

softyoda:
> That list of performer attributes seems too big to me. For example, IMDb has birthday, death date, height for this definitely not porn performer: https://www.imdb.com/name/nm0000026/bio/

Can you use imdb as scrapper for sfw mode? I'm not using this soft for any nsfw. Thanks.

Issue #6272, opened 2025-11-13
Empty aliases created and populated for studios


feederbox826:
### Describe the bug

I suspect it's from #6249 @Gykes 

When creating a new scene from tagger view, the studio is populated with an empty alias which is not automatically removed

### Steps to reproduce

1. Create a new studio from "Studios" tab that has a parent studio in StashDB
2. Scrape the child studio in tagger view
3. Try accepting scrape
4. Alias " is in use by [Parent]

### Expected behaviour

Empty aliases are removed and/or ignored

### Stash version

v0.29.3-29-gc99825a4

### Relevant log output

```shell
2025-11-13 01:35:18
Error   
studioUpdate: input: studioUpdate name '' is used as alias for 'Nookies Network'
2025-11-13 01:35:00
Error   
studioUpdate: input: studioUpdate name '' is used as alias for 'A Girl Knows'
2025-11-13 01:35:00
Error   
studioCreate: input: studioCreate name '' is used as alias for 'A Girl Knows'
2025-11-13 01:34:39
Error   
studioUpdate: input: studioUpdate name '' is used as alias for 'A Girl Knows'
2025-11-13 01:34:39
Error   
studioCreate: input: studioCreate name '' is used as alias for 'A Girl Knows'
```

Gykes:
*sigh*

I'll look at it when I get home.

Gykes:
So, I think it's `StudioModal.tsx:239`. Currently it's splitting the studio by `,` and trimming them but it's not filtering out empty strings causing it to create empty aliases.

Edit: Now looking deeper I think it would be better to make this change in the backend so it applies to all areas where studio aliases are used.

Gykes:
Managed to test this really quick. Followed the above steps and received no errors.

Created blank studio called 1000 Facials
Scraped it in Tagger view and Selected the appropriate studio
Had tagger automatically create the parent studio (Blowpass) and child studio
Both studios were created and meta data populated.

WithoutPants:
This was my bad. 

The expected behaviour in the backend should be to fail on empty alias values, rather than silently ignoring them.

PR #6273, opened 2025-11-13
Bugfix: Fix empty Aliases Being Created for Studios

Gykes:
Fix empty aliases being created for studios. Filter out empty strings when processing studio aliases to prevent empty aliases from being saved. Fixes issue where empty aliases block studio creation/updates with 'Alias '' is in use' error.

This is untested. Will test in a few hours.

Fixes https://github.com/stashapp/stash/issues/6272


Gykes:
Managed to test this really quick. Followed the steps in the original issue and received no errors.


My Steps:
Created blank studio called 1000 Facials
Scraped it in Tagger view and Selected the appropriate studio
Had tagger automatically create the parent studio (Blowpass) and child studio
Both studios were created and meta data populated.
Scraped a scene in scene tagger for 1000Facials and also received no error at save

feederbox826 (approved):
fixes the bug, doesnt past lint :eyes:

PR #6278, opened 2025-11-14
Fix spanish translation

donlothario:


feederbox826:
Please submit changes instead via https://translate.codeberg.org/projects/stash/stash/

otherwise it can't be merged

Issue #6271, opened 2025-11-13
Feature Request: Add StashIDs Manually in UI for all Objects


Gykes:
### Describe the feature you'd like

With the addition of StashID support for tags it looks like it may be time to finally add the ability to manually add or search for StashIDs to the various objects that support them (Scenes, Performers, Studios, Tags).

### Describe the benefits this would bring to existing users

Better matching?

### Is there an existing way to achieve this goal?

Via plugins but you have to manually find the ID on your respective Stashbox and add it manually.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover the core request of my proposal

### Additional context

From WithoutPants: 
"I figure an add button dropdown to select the stash-box instance, then opens a dialog to search for the object via name or stash id"

I pretty much agree with that implantation. Something much simpler like how the current StashID Input plugin works would be good as well for just straight up manual entry but it's already a plugin so might not want to spend the time moving that to core.

Gykes:
So my thought on this are as such:

  1. User clicks Add/IDs or some variation button with a dropdown showing available stash-box instances
  2. Selects which stash-box instance to search
  3. Opens a search dialog where they can search by:
    - Name (e.g., "Brazzers" for a studio)
    - Stash ID (direct lookup)
  4. Shows search results from that stash-box
  5. User selects the correct match
  6. Adds that stash ID to the local object

I feel like this would essentially just be tagger but one Studio/Scene/Performer/Tag at a time. 

If we dont want to get that complex we can essentially move this plugin to core. <https://docs.stashapp.cc/plugins/list/#stash-stashid-input>

This would allow users to add tags manually but it would not give them the search option.

WithoutPants:
I'd prefer not to allow users to easily add potentially invalid stash-ids in the core UI. Using a search modal is a little more complex, but I think it's the nicer way to go.

Implementation considerations:
- reuse the search modal components as much as possible

Gykes:
Understood, i'll look into implementing it.

Gykes:
So, had a short think about this and I think the overall look should match how the plugin works but instead of just adding the stash ID it would search.  I added a very cheap and nasty paint product of what I think it could look like. 

Left box: Dropdown for Stashbox selection
Middle box: Enter Stash ID or name
Right box: Search.

The right box will open the modal and do an initial search for whatever is in the box. The user can then select the correct result or retry with another search.

UI Sketch: 

[...]

PR #6284, opened 2025-11-17
Manually Search Stash ID - Edit Page - Performers

Gykes:
This allows users to manually search for Stash IDs inside of the performer modal via name or stash ID. I built the Search Modal out more than needed to support the other sections later on but it still needs work to include tags. Once we get performers locked down then I can go ahead and move it over for the rest of the required edit pages

Screenshots:  https://github.com/stashapp/stash/issues/6271#issuecomment-3539891934

### DIFF
--- a/ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
+++ b/ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
@@ -15,15 +15,15 @@ import { stashBoxPerformerQuery } from "src/core/StashService";
 import { useToast } from "src/hooks/Toast";
 import { stringToGender } from "src/utils/gender";
 
-const CLASSNAME = "StashBoxIDSearchModal";
-const CLASSNAME_LIST = `${CLASSNAME}-list`;
-const CLASSNAME_LIST_CONTAINER = `${CLASSNAME_LIST}-container`;
-
 interface IProps {
   stashBoxes: GQL.StashBox[];
   excludedStashBoxEndpoints?: string[];
   onSelectItem: (item?: GQL.StashIdInput) => void;
 }
+
+const CLASSNAME = "StashBoxIDSearchModal";
+const CLASSNAME_LIST = `${CLASSNAME}-list`;
+const CLASSNAME_LIST_CONTAINER = `${CLASSNAME_LIST}-container`;
 
 interface IHasRemoteSiteID {
   remote_site_id?: string | null;
@@ -146,9 +146,9 @@ export const StashBoxIDSearchModal: React.FC<IProps> = ({
     null
   );
   const [query, setQuery] = useState<string>("");
-  const [results, setResults] = useState<GQL.ScrapedPerformerDataFragment[]>(
-    []
-  );
+  const [results, setResults] = useState<
+    GQL.ScrapedPerformerDataFragment[] | undefined
+  >(undefined);
   const [loading, setLoading] = useState(false);
 
   useEffect(() => {
@@ -196,7 +196,7 @@ export const StashBoxIDSearchModal: React.FC<IProps> = ({
   }
 
   function renderResults() {
-    if (results.length === 0) {
+    if (!results || results.length === 0) {
       return null;
     }
 
@@ -297,12 +297,12 @@ export const StashBoxIDSearchModal: React.FC<IProps> = ({
           <div className="m-4 text-center">
             <LoadingIndicator inline />
           </div>
-        ) : results.length > 0 ? (
+        ) : results && results.length > 0 ? (
           renderResults()
         ) : (
-          query !== "" &&
-          !loading && (
-            <h5 className="text-center">
+          results !== undefined &&
+          results.length === 0 && (
+            <h5 className="text-center mt-4">
               <FormattedMessage id="stashbox_search.no_results" />
             </h5>
           )


WithoutPants:
I think `Search StashBox for StashID` is way too wordy. I would replace this with a simple `+` icon button, or alternatively `Add...` with the ellipsis to show that it will open a dialog.

In the modal dialog, I think a simple select would be better than a `DropDown`.

Text field in the modal dialog needs to be made consistent with other text fields.

I wasn't able to actually set a value with this. I could get search results, but clicking on it didn't do anything.

Gykes:
I put that verbage due to the button being there but it's not super obvious what the button is for. We can make it to where the "StashID:" always appears then it makes a little more sense than a random button that just says "Add" or something along those lines. 

For the simple select how do you mean? I know I have roughly 4 Stashboxes and just listing them out would take up quite a lot of space. 

I will look at the setting the value. I was able to se it last night just fine we maybe I broke something in my latest commit.

ui/v2.5/src/components/Shared/StashBoxSearchModal.tsx
```diff
@@ -0,0 +1,470 @@
...
+import { Icon } from "src/components/Shared/Icon";
+import {
+  stashBoxPerformerQuery,
+  stashBoxStudioQuery,
+  stashBoxSceneQuery,
+} from "src/core/StashService";
+import { useToast } from "src/hooks/Toast";
+import { stringToGender } from "src/utils/gender";
+
+const CLASSNAME = "StashBoxSearchModal";
```
WithoutPants: Unless we want to generalise this as a general selection modal, I think `StashBoxIDSearchModal` is more informative. The filename and component should be renamed too.

ui/v2.5/src/components/Shared/StashBoxSearchModal.tsx
```diff
@@ -0,0 +1,470 @@
...
+interface IStashBox extends GQL.StashBox {
+  index: number;
+}
+
+interface IProps {
+  entityType: EntityType;
+  stashBoxes: GQL.StashBox[];
+  onHide: () => void;
+  onSelectItem: (
+    item: GQL.ScrapedPerformer | GQL.ScrapedStudio | GQL.ScrapedScene,
```
WithoutPants: Assuming we're only using this to select stash ids, this should return a stash id string.

ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
```diff
@@ -318,7 +321,6 @@ export const PerformerEditPanel: React.FC<IPerformerDetails> = ({
       newIDs?.push({
         endpoint: (scraper as IStashBox).endpoint,
         stash_id: remoteSiteID,
-        updated_at: new Date().toISOString(),
```
WithoutPants, line 321: Why was this removed?

ui/v2.5/src/locales/en-GB.json
```diff
@@ -1433,6 +1434,13 @@
   "stash_id": "Stash ID",
   "stash_id_endpoint": "Stash ID Endpoint",
   "stash_ids": "Stash IDs",
+  "stashbox_instance": "Select Stash-Box:",
```
WithoutPants: This is duplicated below in `stashbox_search.select_stashbox`, and I think that is the better handle. I'd remove this one and update the id in `StashBoxIDSearchModal.tsx`

ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
```diff
@@ -701,7 +739,20 @@ export const PerformerEditPanel: React.FC<IPerformerDetails> = ({
...
+          "performers",
+          undefined,
+          undefined,
+          <Button
+            variant="success"
+            className="mr-2 py-0"
+            onClick={() => setIsStashIDSearchOpen(true)}
+            disabled={!stashConfig?.general.stashBoxes?.length}
+          >
+            <Icon icon={faPlus} id="actions.add" />
```
WithoutPants: I think you've left in the formatted message id here.

ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
```diff
@@ -701,7 +739,20 @@ export const PerformerEditPanel: React.FC<IPerformerDetails> = ({
         {renderInputField("details", "textarea")}
         {renderTagsField()}
 
-        {renderStashIDsField("stash_ids", "performers")}
+        {renderStashIDsField(
+          "stash_ids",
+          "performers",
+          undefined,
+          undefined,
+          <Button
```
WithoutPants, line 745: Add a title text to this.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -113,6 +113,7 @@
     "scrape_scene_fragment": "Scrape by fragment",
     "scrape_with": "Scrape with…",
     "search": "Search",
+    "add": "Add...",
```
WithoutPants: This is out of order and duplicating an existing handle. I think your best bet is to add a `add_stash_id` entry and use that as the title of the add stash id button.

ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
```diff
@@ -0,0 +1,445 @@
...
+    </div>
+  );
+};
+
+// Studio Result Component
+interface IStudioResultProps {
+  studio: GQL.ScrapedStudioDataFragment;
+}
+
+const StudioSearchResultDetails: React.FC<IStudioResultProps> = ({
```
WithoutPants: I'm undecided on keeping the unrelated stuff in here, since we can't test it. Might be better to stash and remove it from this PR.

ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
```diff
@@ -0,0 +1,445 @@
...
+    } finally {
+      setLoading(false);
+    }
+  }, [query, selectedStashBox, entityType, Toast]);
+
+  function handleItemClick(
+    item:
+      | GQL.ScrapedPerformerDataFragment
+      | GQL.ScrapedStudioDataFragment
+      | GQL.ScrapedSceneDataFragment
```
WithoutPants: Optional: this could be an interface that has the `remote_site_id` and `endpoint` fields.

ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
```diff
@@ -0,0 +1,445 @@
...
+  function renderResults() {
+    if (results.length === 0) {
+      return null;
+    }
+
+    return (
+      <div className={CLASSNAME_LIST_CONTAINER}>
+        <div className="mt-1 mb-2">
+          <FormattedMessage
+            id={`dialogs.${entityType}s_found`}
```
WithoutPants: Having an `s` suffix here isn't going to work for all locales. Look at the `countables` entries in `en-GB.json` for guidance.

ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
```diff
@@ -0,0 +1,445 @@
...
+import { useToast } from "src/hooks/Toast";
+import { stringToGender } from "src/utils/gender";
+
+const CLASSNAME = "StashBoxIDSearchModal";
+const CLASSNAME_LIST = `${CLASSNAME}-list`;
+const CLASSNAME_LIST_CONTAINER = `${CLASSNAME_LIST}-container`;
+
+export type EntityType = "performer" | "studio" | "scene";
+
+interface IProps {
```
WithoutPants, line 18: Move this to be above the declaration that uses it.

ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
```diff
@@ -0,0 +1,445 @@
...
+  entityType: EntityType;
+  stashBoxes: GQL.StashBox[];
+  onHide: () => void;
+  onSelectItem: (stashId: string, endpoint: string) => void;
```
WithoutPants: The typical pattern is for `onSelect` to return an optional object (in this case probably a `GQL.StashIdInput` object), with a return type of `undefined` indicating that the operation was canceled.

WithoutPants (changes requested):
Result list styling needs fixing:
Image

`cursor: pointer` should also be added to the seach result items to indicate that they are clickable.

There is a peculiar usability issue that probably needs addressing as well. Adding an id where the performer already has one from the same stash-box instance replaces it: which is expected, but is a weird UX experience; there's no obvious indication that the existing stash id was replaced unless you're looking at the ids. I think the solution here is a warning that should appear when a stash-box instance is selected for which there is already one, OR a dialog should pop up after the id is selected to confirm the replace operation.

ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
Gykes: I assume you just mean to name the undefined parameter?

ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
Gykes: Removed this type

ui/v2.5/src/locales/en-GB.json
Gykes: I changed it over to `stashbox.source` the `...` in the other one was throwing me off.

Gykes:
Lots of changes that you request. 

One thing I did was remove the style on the list so that there were no more bullets.

Issue #4678, opened 2024-03-11
[Bug Report] Wrong Update file downloaded within Stash


Archigos:
**Describe the bug**
When using the macOS Stash.app version tested with v0.24.3 [aeb68a58] it says the new version v0.25.0 is available, but from within Stash if you click Download it will download the `stash-macos` CLI version instead of the `Stash.app` version. It should detect and download same 'style' version and not default to CLI version.

**To Reproduce**
1. Go to Settings -> About
2. Click on 'Check for new version'
3. Click 'Download'

**Expected behavior**
When using the CLI version (`stash-macos`) downloading the new CLI is expected, but when using Stash.app, the new version should be Stash.app as well.

**Screenshots**
Not needed for this.

**Stash Version: (from Settings -> About):**
v0.24.3 [aeb68a58] but it's happened in the past as well, I have to manually go to GitHub to download the proper version each time.

**Desktop (please complete the following information):**
 - OS: macOS Ventura 13.6.4
 - Browser Chrome (Latest)
 - Version v0.24.3 [aeb68a58]

**Additional context**
I mentioned this on Discord [here](https://discord.com/channels/559159668438728723/559399634318131210/1216889615420493955) and was advised by WithoutPants to report this as a bug. :)

PR #6289, opened 2025-11-18
Bug Fix: Update Macos Version Check

Gykes:
This adds in a condition for checking if mac users are using the CLI or the .app version and downloads the appropriate version.

I tested this on my macbook and the .app version was downloaded correctly from the .app version of Stash.

Fixes: #4678

### DIFF
--- a/internal/api/check_version.go
+++ b/internal/api/check_version.go
@@ -41,7 +41,7 @@ var stashReleases = func() map[string]string {
 // isMacOSBundle checks if the application is running from within a macOS .app bundle
 func isMacOSBundle() bool {
 	exec, err := os.Executable()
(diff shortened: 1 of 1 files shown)


internal/api/check_version.go
```diff
@@ -36,6 +38,24 @@ var stashReleases = func() map[string]string {
 	}
 }
 
+// isMacOSBundle checks if the application is running from within a macOS .app bundle
+func isMacOSBundle() bool {
+	exec, err := os.Executable()
+	return err == nil && strings.Contains(exec, ".app/")
```
WithoutPants: Could this not be `Stash.app/`?

Gykes: It could be if that's how you want it 👍

PR #6290, opened 2025-11-18
[markers] ignore generating markers past end

### LINKED ISSUE #4920: [Feature] Ignore generating markers past end
When generating markers, the duration of the file is not considered. This leads to errors generating these markers with the following ffmpeg error
```
[generator] failed to generate marker screenshot: ffmpeg command produced no output
```
This is the correct response, as these markers are impossible to generate given the file

When generating markers, start times past the duration of the file are ignored

Removing these markers is possible via `[Skip Sync]` and the corresponding `TMDB skip sync tags`

**Additional context**
This change could also be made on the marker sync side but I don't believe duration of the file is associated with the markers. In addition, scrapers can also generate invalid marker timestamps

feederbox826:
closes #4920


feederbox826:
Addl note that this **does not** nullify, delete or ignore the marker for future runs, it just skips generating & throws a warning in case the primary file changes

PR #6299, opened 2025-11-19
Use notify to watch filesystem for changes

### LINKED ISSUE #191: Filesystem monitoring
Feature enhancement suggestion to implement filesystem monitoring.

The different platforms have APIs for monitoring the filesystem for changes. This allows you to watch a filesystem tree for renames, movement within the tree, deletions, additions, or other modifications. The nice thing is that it requires no polling to keep the file paths up to date. If a file is renamed, the app immediately gets and event and can update the database with minimal processing requirements.

There are, unfortunately, quite a few shortcomings, such as incompatible APIs between platforms, limits on the number of watches, no way of monitoring network shares, etc. That said, I still think it's worth checking out since it can make working with a collection much smoother.

There is a cross-platform library for go which supports linux/mac/windows. I haven't looked at the limitations of the mac/windows implementations, but hopefully it's not too bad: https://github.com/fsnotify/fsnotify

More info about the APIs:
https://en.wikipedia.org/wiki/Inotify
https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-readdirectorychangesw

NodudeWasTaken:
I was tired of triggering scans manually.

### DIFF
--- a/internal/manager/watcher.go
+++ b/internal/manager/watcher.go
@@ -1,0 +1,261 @@
+package manager
+
+import (
+	"context"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/stashapp/stash/pkg/file"
+	file_image "github.com/stashapp/stash/pkg/file/image"
+	"github.com/stashapp/stash/pkg/file/video"
+	"github.com/stashapp/stash/pkg/fsutil"
+	"github.com/stashapp/stash/pkg/job"
+	"github.com/stashapp/stash/pkg/logger"
+)
+
+const (
+	// How long to wait for write events to finish before scanning file
+	defaultWriteDebounce = 30
+	// The maximum number of tasks on a single scan
+	taskQueueSize = 200
+)
+
+// findAssociatedVideo attempts to map an event path an associated video.
+func findAssociatedVideo(eventPath string, s *Manager) *string {
+	dir := filepath.Dir(eventPath)
+
+	findWithExts := func(prefix string) *string {
+		for _, ve := range s.Config.GetVideoExtensions() {
+			cand := filepath.Join(dir, prefix+"."+ve)
+			if _, err := os.Stat(cand); err == nil {
+				return &cand
+			}
+		}
+		return nil
+	}
+
+	// funscript
+	if fsutil.MatchExtension(eventPath, []string{"funscript"}) {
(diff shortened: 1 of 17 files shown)


DogmaDragon:
Related
- https://github.com/stashapp/stash/discussions/1965

internal/manager/watcher.go
```diff
@@ -0,0 +1,282 @@
...
+				}
+
+				// add watches for newly created directories
+				if ev.Op&fsnotify.Create == fsnotify.Create {
+					// Try to add if it's a dir (best-effort)
+					_ = watcher.Add(ev.Name)
+				}
+
+				// interested in Write/Create/Rename
+				if ev.Op&(fsnotify.Write|fsnotify.Create|fsnotify.Rename) == 0 {
```
feederbox826: Write/Create will trigger false positives when transferring over the network and depends heavily on use case and how it's being transferred. I would personally look for close_write,move and attrib (Rename is fine)

NodudeWasTaken:
```
	// File opened for writing was closed.
	//
	// Only works on Linux and FreeBSD.
	//
	// The advantage of using this over Write is that it's more reliable than
	// waiting for Write events to stop. It's also faster (if you're not
	// listening to Write events): copying a file of a few GB can easily
	// generate tens of thousands of Write events in a short span of time.
	xUnportableCloseWrite
```
This would make it not function on Windows and macOS

feederbox826: >	// listening to Write events): copying a file of a few GB can easily
	// generate tens of thousands of Write events in a short span of time.
	
While your debounce will aggregate the writes, it doesn't look like it's listening for close or for it to stop. This would generate corrupt files propogating into the stash db as they're being read before completion and would either error on read or couldn't be traced back since oshash would be garbled

NodudeWasTaken: Show me a system that takes more than 2 seconds between writes and I’ll show you a bad system.
I agree that using CloseWrite would be preferable, but it’s not even exported for us to use:
https://github.com/fsnotify/fsnotify/issues/698
At best, I can increase the debounce timeout.
Also, we specifically need to scan newly created folders to add them to the database, which means we’ll inevitably encounter files that haven’t finished writing yet.

feederbox826: > Show me a system that takes more than 2 seconds between writes and I’ll show you a bad system.

P2P

Gykes:
So, after some talks with the other devs I think keeping this as a plugin would be the best bet in the meantime. Currently there's really no good way to handle this on the stash side. Every way has a pretty large downside that can effect various users in a lot of ways. 

For now, plugin is the way to go.

PR #6303, opened 2025-11-20
Feature Request: Add organized flag to studios

Gykes:
This PR allows users to "Organize" their studios similar to scenes. The UI works in the same way as scenes... images below. Updated the autotag logic to respect if a studio is "Organized". Would love if our friendly neighborhood tester did some testing....

Studio Organized:

Screenshot 2025-11-20 at 13 57 29

Studio Organized (Studio Page):

Screenshot 2025-11-20 at 13 56 11


WithoutPants:
A scene, image or gallery being marked as organised essentially means that it's metadata is correct and should not be modified via auto-tag, identify and possibly other mechanisms.

I'm not entirely sure what the conceptual meaning of an organised studio is and if it really belongs in the system. We have a flag already for ignoring a studio during auto-tag, so as far as I can tell this only affects the batch studio update task. Why is this necessary, and do we need to add similar flags for performers and tags?

Gykes:
Currently I use organized as less of a "avoid tagger" and more of a piece of information that I already organized something. This may be a personal thing but I normally go though each studio one by one, get all the scenes, import to Stash, tag, then move on. Now I would like the organized flag on studios so I know that I have organized it accordingly. 

Is this "needed", probably not. Since we already have a way to avoid tagger for studios then I feel like it should just be made universal. If people want studios to avoid auto-tag then they can also set organized similar to other components that already support this. This means that it avoids tag and scratches that organizational itch. Seeing all the studios with that little flag means I did something right lol.

discourse-stashapp:

https://discourse.stashapp.cc/t/stash-0-30-release-development-thread/4484/1

WithoutPants:
Closing this as I'm not convinced that there's enough of a case to warrant adding this, and I feel that it's a bit conceptually ambiguous in what an organised studio means. Indicating that a studio's collection is complete can be done via tags, and omitting it from the auto-tag feature can be done with the flag specifically intended for that. If there's sufficient community sentiment to adding something like this, I'm happy to consider it.

Gykes:
reopening. Will remove the ignore autotag checkbox and utilize the organized feature.

Gykes:
Instead of depreciating I just went ahead and removed the old field. Since the DB migration converts the old to the new I figured it would be good to remove but can readd it back to graphql and depreciate instead if that is preferred.

WithoutPants:
We can't remove fields from the graphql interface because it will break existing clients. We deprecate them instead.

> Will remove the ignore autotag checkbox and utilize the organized feature.

I don't believe having `Organised` set to true should have the same functionality as `ignore_auto_tag` set to true, and the former should not replace the latter. Setting a scene/image/gallery as organised prevents auto-tag from touching it because that object is considered "complete". The same does not apply for performers/studios/tags for which the meaning of organised is less well-defined, and I believe its not obvious that an organised studio should be prevented from being assigned to media.

Gykes:
So, I always had confusion as to why you were so against combining them. In my brain organized feature in Scenes already did ignore_auto_tag. As someone who never uses autotag I did a little more digging to see if I could figure out your aversion. If I am reading it correctly, if organized or ignore_auto_tag is set, then that studio will no longer be a candidate when a scene is autotagged.

I think we have 3 options: 

1) Run with what I have, merge them, and then make actual documentation on what "organized" is for. 

2) undo what I did to get ignore auto tag into and then run them separate

3) Move away from "Organized". Create something new like "Collected" if it is going to be a purely front end "I have completed this collect" toggle

WithoutPants:
> In my brain organized feature in Scenes already did ignore_auto_tag. 

Correct, and this is covered in the manual:

`Scenes, images, and galleries that have the Organized flag added to them will not be modified by Auto tag.`

Similarly the ignore auto tag flag is also covered in the manual:

`Performers or Tags that have Ignore Auto tag flag added to them will be skipped by the Auto tag task.`

But these are two different sides of the same coin, scenes, galleries and images are _subject_ to auto tag, whereas performers, studios and tags are _applied_. The distinction is important, particularly when we add an organised flag to performers/studios/tags. The expected behaviour of an "organized" studio is _not necessarily_ that it should be excluded from auto-tag.

Were we to go with option 1, we'd also have to consider the behaviour for identify. I don't think it makes sense that organised studios should be excluded from being applied during either operation.

Option 2 should be what we go with, but we should also include documentation for what an "organised" studio is (or rather, document that there is no current specific meaning beyond what is inferred by the end user).

Gykes:
Yep, once I looked and saw that scenes cannot be applied to a organized studio I agreed with you. I thought that it only applied to the studio metadata being modified not that it would block new scenes from being applied to that studio.

I will go with option 2

WithoutPants (changes requested):
Static review looks ok. Please add to the documentation to indicate the intention and impact of the organised field on studios.

WithoutPants (approved):

PR #6306, opened 2025-11-22
Document "# requires" in the plugin config

DogmaDragon:
`# requires` field was somehow left undocumented.
`requires` is similarly used for the UI configuration, but is not the same thing.

Since `#` is used in YAML for comment, it's a mess of an explanation. Please offer suggestions.

### DIFF
--- a/ui/v2.5/src/docs/en/Manual/Plugins.md
+++ b/ui/v2.5/src/docs/en/Manual/Plugins.md
@@ -67,7 +67,7 @@ The basic structure of a plugin configuration file is as follows:
 
 ```yaml
 name: <plugin name> 
-# optional list of dependancies to be included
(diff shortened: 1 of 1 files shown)


feederbox826:
```yml
name: <plugiin name>
# optional dependnecies to be included, one per line
# requires: <plugin ID> # the '#' is manditory as the requires has to be a comment
```

You can use an inline comment for clarification and differenciation?

DogmaDragon:
> ```yaml
> name: <plugiin name>
> # optional dependnecies to be included, one per line
> # requires: <plugin ID> # the '#' is manditory as the requires has to be a comment
> ```
> 
> You can use an inline comment for clarification and differenciation?

That was my first approach too, but it didn't match other comments. It still might be better approach.

Issue #6038, opened 2025-07-28
[Bug Report] Creating new performers with same name unmatches other performers


Darklyter:
**Describe the bug**
When in Tagger view, if you create a new performer that has the same name as other performers (ignoring disambiguation) then it will "unmatch" the other performers from their mappings between local Stash and StashDB/TPDB.  Refreshing the page or rescraping the affected scenes will show the mapped performers correctly.

**To Reproduce**
1) Take an unmapped performer, for example "Trinity" and load scenes into Tagger.
2) Match scene 1 to "Trinity (2001-2007)" (create performer if necessary)
3) For scene 2, create a new performer named something like "Trinity (b1979, Hungary)" (creation will auto map performer)
4) When you save that new performer, scene 1 performer will now show as unmapped with the newly created "Trinity (b1979, Hungary)" as the recommendation.
5) Refresh page or rescrape scene 1, and "Trinity (2001-2007)" will again show as correctly mapped and matched.

**Expected behavior**
Would expect to not unlink other performers on new performer creation

**Screenshots**
You can get more explanation and a quick example video here: https://discord.com/channels/559159668438728723/559159910550732809/1399505594926370836

**Stash Version: (from Settings -> About):**
v0.28.1

feederbox826:
What endpoint are you scraping from? IIRC the tagger uses shared storage to reduce duplicated work across entries and maps them to Name without any metadata.

Possibly a feature turned bug, You can reproduce this if you scrape multiple scenes for an uncreated performer and watch it propogate to the other entries as you create it in one scene

Gykes:
I think the bug is in the `createNewPerformer` function which was matching performers by name only `(p.name === performer.name)`, causing all performers with the same base name to incorrectly receive the newly created performer's ID.

I can change it to match by `remote_site_id` instead, which uniquely identifies each performer from StashDB/TPDB regardless of name. This prevents different performers with similar names from interfering with each other's mappings.

PR #6308, opened 2025-11-22
Bugfix: Tagger Ignoing Disambiguation When Linking Performer

Gykes:
Changed the performer matching logic in `createNewPerformer` to use `remote_site_id` instead of name when updating search results. This ensures only the specific performer being created gets updated with the new `stored_id`, preventing the previous performer from being unmatched between performers with matching names

Fixes: https://github.com/stashapp/stash/issues/6038

### DIFF
--- a/ui/v2.5/src/components/Tagger/context.tsx
+++ b/ui/v2.5/src/components/Tagger/context.tsx
@@ -670,12 +670,7 @@ export const TaggerContext: React.FC = ({ children }) => {
           return {
             ...r,
             performers: r.performers.map((p) => {
-              // Match by remote_site_id if available, otherwise fall back to name
-              const matches = performer.remote_site_id
-                ? p.remote_site_id === performer.remote_site_id
-                : p.name === performer.name;
-
-              if (matches) {
+              if (p.remote_site_id === performer.remote_site_id) {
                 return {
                   ...p,
                   stored_id: performerID,


DogmaDragon:
Reminder for myself: test non-stash-box source.

Gykes:
So, I actually thought about this last night as I went to bed. I don't think non stashbox sources will return a remote ID so I will probably change it up so that if one exists use that if not revert to name

ui/v2.5/src/components/Tagger/context.tsx
```diff
@@ -665,7 +670,12 @@ export const TaggerContext: React.FC = ({ children }) => {
           return {
             ...r,
             performers: r.performers.map((p) => {
-              if (p.remote_site_id === performer.remote_site_id) {
+              // Match by remote_site_id if available, otherwise fall back to name
+              const matches = performer.remote_site_id
+                ? p.remote_site_id === performer.remote_site_id
+                : p.name === performer.name;
+
+              if (matches) {
```
WithoutPants: `performer.remote_site_id` is known to be defined at the top of this function. This change won't be necessary.

Gykes: Removed this from performer link. I haven't been able to do much testing on this as it's hard for me to reproduce.

PR #6310, opened 2025-11-24
Feature Request: Bulk Add by StashID and Name

Gykes:
This PR allows users to uses StashID and/or name when bulk adding studios or performers. 

I was able to get this to work on performers with no errors
Studios had errors that was related to the empty string aliases [bug](https://discourse.stashapp.cc/t/unable-to-bulk-scrap-studios-from-stashdb/4475) and I don't think it is due to this PR.

[This](https://github.com/stashapp/stash/pull/6273) PR was used to fix this issue with normal tagging and I believe it needs to be moved over to the batch create options too.

Testing:
1 performer by StashdID added (V4 and V7)
2 performers comma split by Stash ID added (V4 and V7)
3 performers comma split, 2 by stash ID and 1 by name (V4 and V7)

### DIFF
--- a/internal/manager/manager_tasks.go
+++ b/internal/manager/manager_tasks.go
@@ -375,19 +375,18 @@ const (
 
 // getBatchTagType determines the batch tag mode based on the input
 func (input StashBoxBatchTagInput) getBatchTagType(hasPerformerFields bool) batchTagType {
-	if len(input.Ids) > 0 {
+	switch {
+	case len(input.Ids) > 0:
 		return batchTagByIds
-	}
-	if hasPerformerFields && len(input.PerformerIds) > 0 {
+	case hasPerformerFields && len(input.PerformerIds) > 0:
 		return batchTagByIds
-	}
-	if len(input.StashIDs) > 0 || len(input.Names) > 0 {
+	case len(input.StashIDs) > 0 || len(input.Names) > 0:
 		return batchTagByNamesOrStashIds
-	}
-	if hasPerformerFields && len(input.PerformerNames) > 0 {
+	case hasPerformerFields && len(input.PerformerNames) > 0:
 		return batchTagByNamesOrStashIds
+	default:
+		return batchTagAll
 	}
-	return batchTagAll
 }
 
 // Accepts either ids, or a combination of names and stash_ids.
@@ -500,11 +499,7 @@ func (s *Manager) batchTagAllPerformers(ctx context.Context, input StashBoxBatch
 		var performers []*models.Performer
 		var err error
 
-		if input.Refresh {
-			performers, err = performerQuery.FindByStashIDStatus(ctx, true, box.Endpoint)
-		} else {
-			performers, err = performerQuery.FindByStashIDStatus(ctx, false, box.Endpoint)
-		}
+		performers, err = performerQuery.FindByStashIDStatus(ctx, input.Refresh, box.Endpoint)
 
 		if err != nil {
 			return fmt.Errorf("error querying performers: %v", err)
@@ -641,11 +636,7 @@ func (s *Manager) batchTagAllStudios(ctx context.Context, input StashBoxBatchTag
 		var studios []*models.Studio
 		var err error
 
-		if input.Refresh {
-			studios, err = studioQuery.FindByStashIDStatus(ctx, true, box.Endpoint)
-		} else {
-			studios, err = studioQuery.FindByStashIDStatus(ctx, false, box.Endpoint)
-		}
+		studios, err = studioQuery.FindByStashIDStatus(ctx, input.Refresh, box.Endpoint)
 
 		if err != nil {
 			return fmt.Errorf("error querying studios: %v", err)
--- a/internal/manager/task_stash_box_tag.go
+++ b/internal/manager/task_stash_box_tag.go
@@ -72,6 +72,8 @@ func (t *stashBoxBatchPerformerTagTask) findStashBoxPerformer(ctx context.Contex
 	client := stashbox.NewClient(*t.box, stashbox.ExcludeTagPatterns(instance.Config.GetScraperExcludeTagPatterns()))
 
 	switch {
+	case t.name != nil:
+		performer, err = client.FindPerformerByName(ctx, *t.name)
 	case t.stashID != nil:
 		performer, err = client.FindPerformerByID(ctx, *t.stashID)
 
@@ -122,8 +124,6 @@ func (t *stashBoxBatchPerformerTagTask) findStashBoxPerformer(ctx context.Contex
 				}
 			}
 		}
-	case t.name != nil:
-		performer, err = client.FindPerformerByName(ctx, *t.name)
 	}
 
 	if performer != nil {
@@ -303,6 +303,8 @@ func (t *stashBoxBatchStudioTagTask) findStashBoxStudio(ctx context.Context) (*m
 	client := stashbox.NewClient(*t.box, stashbox.ExcludeTagPatterns(instance.Config.GetScraperExcludeTagPatterns()))
 
 	switch {
+	case t.name != nil:
+		studio, err = client.FindStudio(ctx, *t.name)
 	case t.stashID != nil:
 		studio, err = client.FindStudio(ctx, *t.stashID)
 	case t.studio != nil:
@@ -327,8 +329,6 @@ func (t *stashBoxBatchStudioTagTask) findStashBoxStudio(ctx context.Context) (*m
 		if remoteID != "" {
 			studio, err = client.FindStudio(ctx, remoteID)
 		}
-	case t.name != nil:
-		studio, err = client.FindStudio(ctx, *t.name)
 	}
 
 	if err := r.WithReadTxn(ctx, func(ctx context.Context) error {


internal/manager/manager_tasks.go
```diff
@@ -439,6 +441,20 @@ func (s *Manager) StashBoxBatchPerformerTag(ctx context.Context, box *models.Sta
...
+				if len(stashID) > 0 {
+					tasks = append(tasks, StashBoxBatchTagTask{
+						stashID:        &stashID,
+						refresh:        true, // use refresh mode to query by ID
+						box:            box,
+						excludedFields: input.ExcludeFields,
+						taskType:       Performer,
+					})
+				}
+			}
```
WithoutPants: This looks like it won't handle if you have both stash ids and names in the one operation, due to the `if/else` structure. It will handle the stash ids only and not the names.

Gykes: I forgot to push my latest commit that fixes this.

feederbox826 (approved):
LGTM, still running into "studio alias must not be an empty string" bug

Gykes:
Yea, will need to rebase when WP makes his changes to see that bug fixed

DogmaDragon:
That fix was merged. 
- https://github.com/stashapp/stash/pull/6313

internal/manager/task_stash_box_tag.go
```diff
@@ -44,24 +45,32 @@ func (t *StashBoxBatchTagTask) Start(ctx context.Context) {
 }
```
WithoutPants: I think this task needs to be split up and refactored. I couldn't work out why you'd set `refresh` to `true` in `manager_tasks.go` until I looked through the code here. In the existing code `refresh` is used to determine whether to use the existing performer's applicable stash id or its name to search for the stash-box performer.

Your use of that parameter further muddies it's purpose because now it seems to be being used to indicate if it should search via a passed stash id. It's just a bit of a mess to follow.

Having a single task with two separate execution paths and inputs seems needlessly complicated as well.

My suggestion would be to split this into two private tasks:
```
type stashBoxBatchPerformerTagTask struct {
    // stash-box instance to search
	box            *models.StashBox
    // if set, searches by performer name
	name           *string
    // if set, searches by stash-id
	stashID        *string
    // if set, updates performer with matched stash-box performer. If unset, creates new performer
	performer      *models.Performer
}

type stashBoxBatchStudioTagTask struct {
    // stash-box instance to search
	box            *models.StashBox
    // if set, searches by performer name
	name           *string
    // if set, searches by stash-id
	stashID        *string
    // if set, updates performer with matched stash-box performer. If unset, creates new performer
	studio *models.Studio
}
```

I think this gives a clearer indication of the expected behaviour of each task, and simplifies the execution path a bit.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -1289,6 +1289,7 @@
     "number_of_performers_will_be_processed": "{performer_count} performers will be processed",
     "performer_already_tagged": "Performer already tagged",
     "performer_names_separated_by_comma": "Performer names separated by comma",
```
WithoutPants: This entry is now unused and should be removed.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -1498,6 +1499,7 @@
     "status_tagging_studios": "Status: Tagging studios",
     "studio_already_tagged": "Studio already tagged",
     "studio_names_separated_by_comma": "Studio names separated by comma",
```
WithoutPants: This entry is now unused and should be removed.

discourse-stashapp:

https://discourse.stashapp.cc/t/stash-v0-30-release-development-thread/4484/1

internal/manager/task_stash_box_tag.go
```diff
@@ -76,29 +34,59 @@ func (t *StashBoxBatchTagTask) stashBoxPerformerTag(ctx context.Context) {
...
+func (t *stashBoxBatchPerformerTagTask) GetDescription() string {
+	var name string
+	switch {
+	case t.name != nil:
+		name = *t.name
+	case t.stashID != nil:
+		name = *t.stashID
+	case t.performer != nil:
+		name = t.performer.Name
+	}
```
WithoutPants: Let's move this repeated code into a separate private method.

internal/manager/task_stash_box_tag.go
```diff
@@ -275,29 +261,47 @@ func (t *StashBoxBatchTagTask) stashBoxStudioTag(ctx context.Context) {
...
+func (t *stashBoxBatchStudioTagTask) GetDescription() string {
+	var name string
+	switch {
+	case t.name != nil:
+		name = *t.name
+	case t.stashID != nil:
+		name = *t.stashID
+	case t.studio != nil:
+		name = t.studio.Name
+	}
```
WithoutPants: Make a private utility method out of this as above.

ui/v2.5/src/components/Tagger/studios/StudioTagger.tsx
```diff
@@ -715,14 +715,30 @@ export const StudioTagger: React.FC<ITaggerProps> = ({ studios }) => {
...
+      const names: string[] = [];
+      const stashIds: string[] = [];
+
+      inputs.forEach((input) => {
+        if (uuidPattern.test(input)) {
+          stashIds.push(input);
+        } else {
+          names.push(input);
+        }
+      });
```
WithoutPants: Repeated code - let's put this into a utility function to get a list of names/uuidss.

graphql/schema/types/scraper.graphql
WithoutPants: These docs need to be updated.

internal/manager/manager_tasks.go
```diff
@@ -379,6 +379,8 @@ type StashBoxBatchTagInput struct {
 	Ids []string `json:"ids"`
 	// If set, only tag these names
 	Names []string `json:"names"`
+	// If set, only tag these stash ids from stash-box
+	StashIDs []string `json:"stash_ids"`
 	// If set, only tag these performer ids
```
WithoutPants: These docs need to be updated.

WithoutPants (changes requested):

internal/manager/manager_tasks.go
Gykes: I'm not sure if you meant like actual docs for this. I updated the `scraper.graphql` to output slightly better/more accurate verbage and updated the comments on this one.

graphql/schema/types/scraper.graphql
```diff
@@ -281,7 +281,7 @@ type StashBoxFingerprint {
   duration: Int!
 }
 
-"If neither ids nor names are set, tag all items"
+"If neither ids, names, nor stash_ids are set, tag all items"
```
WithoutPants: This is still potentially confusing. Here's my proposal:

```
"""
Accepts either ids, or a combination of names and stash_ids.
If none are set, then all existing items will be tagged.
"""
input StashBoxBatchTagInput {
...
```

```
"""
IDs in stash of the items to update.
If set, names and stash_ids fields will be ignored
"""
ids: [ID!]

"Names of the items in the stash-box instance to search for and create"
names: [String!]

"Stash IDs of the items in the stash-box instance to search for and create"
stash_ids: [String!]
```

And similar comments for the deprecated fields. This input is a bit of a nightmare to document. Make sure to update the go input struct as well.

internal/manager/manager_tasks.go
```diff
@@ -393,7 +395,7 @@ func (s *Manager) StashBoxBatchPerformerTag(ctx context.Context, box *models.Sta
 	j := job.MakeJobExec(func(ctx context.Context, progress *job.Progress) error {
 		logger.Infof("Initiating stash-box batch performer tag")
 
-		var tasks []StashBoxBatchTagTask
+		var tasks []Task
 
 		// The gocritic linter wants to turn this ifElseChain into a switch.
 		// however, such a switch would contain quite large blocks for each section
```
WithoutPants: Since we're modifying this code already, I think it's worth resolving this golint issue.

Use a `switch` block, and extract each case into a private method that will return the tasks.

internal/manager/manager_tasks.go
```diff
@@ -525,7 +532,7 @@ func (s *Manager) StashBoxBatchStudioTag(ctx context.Context, box *models.StashB
 	j := job.MakeJobExec(func(ctx context.Context, progress *job.Progress) error {
 		logger.Infof("Initiating stash-box batch studio tag")
 
-		var tasks []StashBoxBatchTagTask
+		var tasks []Task
 
 		// The gocritic linter wants to turn this ifElseChain into a switch.
 		// however, such a switch would contain quite large blocks for each section
```
WithoutPants: As above, use `switch` and split into private methods.

internal/manager/task_stash_box_tag.go
```diff
@@ -192,9 +168,6 @@ func (t *StashBoxBatchTagTask) processMatchedPerformer(ctx context.Context, p *m
 
 			partial := p.ToPartial(t.box.Endpoint, excluded, existingStashIDs)
 
-			// if we're setting the performer's aliases, and not the name, then filter out the name
-			// from the aliases to avoid duplicates
-			// add the name to the aliases if it's not already there
```
WithoutPants, line 197: Why was this comment removed? It's not clear enough in the code why we're doing this.

internal/manager/task_stash_box_tag.go
```diff
@@ -226,8 +199,7 @@ func (t *StashBoxBatchTagTask) processMatchedPerformer(ctx context.Context, p *m
 		} else {
 			logger.Infof("Updated performer %s", *p.Name)
 		}
-	} else if t.name != nil && p.Name != nil {
-		// Creating a new performer
+	} else if (t.name != nil || t.stashID != nil) && p.Name != nil {
```
WithoutPants: If I understand the code correctly, an `else` should be sufficient here in that an existing performer hasn't been provided, so we must be creating a new one. It should have a comment stating when this code branch will be hit though.

internal/manager/task_stash_box_tag.go
```diff
@@ -14,57 +14,28 @@ import (
...
 	box            *models.StashBox
 	name           *string
+	stashID        *string
 	performer      *models.Performer
-	studio         *models.Studio
-	refresh        bool
-	createParent   bool
 	excludedFields []string
-	taskType       StashBoxTagTaskType
 }
```
WithoutPants, line 28: Some documentation is needed to state how to use the task. Specifically that there are two modes of use: one with `performer` set, and one with `name` or `stashID` set.

internal/manager/task_stash_box_tag.go
```diff
@@ -263,7 +235,29 @@ func (t *StashBoxBatchTagTask) processMatchedPerformer(ctx context.Context, p *m
 	}
 }
 
-func (t *StashBoxBatchTagTask) stashBoxStudioTag(ctx context.Context) {
+type stashBoxBatchStudioTagTask struct {
```
WithoutPants, line 252: Document as above.

internal/manager/task_stash_box_tag.go
```diff
@@ -394,8 +381,7 @@ func (t *StashBoxBatchTagTask) processMatchedStudio(ctx context.Context, s *mode
 		} else {
 			logger.Infof("Updated studio %s", s.Name)
 		}
-	} else if t.name != nil && s.Name != "" {
-		// Creating a new studio
+	} else if (t.name != nil || t.stashID != nil) && s.Name != "" {
```
WithoutPants: As above, an `else` should be sufficient.

WithoutPants (changes requested):

internal/manager/task_stash_box_tag.go
Gykes: I think you are right. If `t.performer` != nil then we're updating a performer if it is then we're creating.

Gykes:
Lots of refactoring in this one. My eyes started to glaze over near the end but I was still able to test functionality. If documentation is still not to your liking then please feel free to update it yourself since that will probably be quicker lol

internal/manager/task_stash_box_tag.go
Gykes: Was a fat finger

internal/api/resolver_mutation_stash_box.go
```diff
@@ -39,7 +39,7 @@ func (r *mutationResolver) SubmitStashBoxFingerprints(ctx context.Context, input
 }
 
 func (r *mutationResolver) StashBoxBatchPerformerTag(ctx context.Context, input manager.StashBoxBatchTagInput) (string, error) {
-	b, err := resolveStashBoxBatchTagInput(input.Endpoint, input.StashBoxEndpoint)
+	b, err := resolveStashBoxBatchTagInput(input.Endpoint, input.StashBoxEndpoint) //nolint:staticcheck
```
WithoutPants: When using `nolint`, also document why. In this instance `//nolint:staticcheck // still need to handle deprecated field`

However, this should be unnecessary if we remove the deprecated flag from the field, see below.

internal/manager/manager_tasks.go
```diff
@@ -364,9 +364,38 @@ func (s *Manager) MigrateHash(ctx context.Context) int {
...
+	}
+	if hasPerformerFields && len(input.PerformerIds) > 0 {
+		return batchTagByIds
+	}
+	if len(input.StashIDs) > 0 || len(input.Names) > 0 {
+		return batchTagByNamesOrStashIds
+	}
+	if hasPerformerFields && len(input.PerformerNames) > 0 {
+		return batchTagByNamesOrStashIds
+	}
```
WithoutPants: Optional: this can probably be a switch?

internal/manager/manager_tasks.go
```diff
@@ -375,128 +404,147 @@ type StashBoxBatchTagInput struct {
...
+	err := s.Repository.WithTxn(ctx, func(ctx context.Context) error {
+		performerQuery := s.Repository.Performer
+		var performers []*models.Performer
+		var err error
+
+		if input.Refresh {
+			performers, err = performerQuery.FindByStashIDStatus(ctx, true, box.Endpoint)
+		} else {
+			performers, err = performerQuery.FindByStashIDStatus(ctx, false, box.Endpoint)
+		}
```
WithoutPants: This can be simplified to `...FindByStashIDStatus(ctx, input.Refresh, box.Endpoint)`

internal/manager/manager_tasks.go
```diff
@@ -521,104 +569,121 @@ func (s *Manager) StashBoxBatchPerformerTag(ctx context.Context, box *models.Sta
...
+	err := s.Repository.WithTxn(ctx, func(ctx context.Context) error {
+		studioQuery := s.Repository.Studio
+		var studios []*models.Studio
+		var err error
+
+		if input.Refresh {
+			studios, err = studioQuery.FindByStashIDStatus(ctx, true, box.Endpoint)
+		} else {
+			studios, err = studioQuery.FindByStashIDStatus(ctx, false, box.Endpoint)
+		}
```
WithoutPants: Simplify

internal/manager/task_stash_box_tag.go
```diff
@@ -133,14 +122,8 @@ func (t *StashBoxBatchTagTask) findStashBoxPerformer(ctx context.Context) (*mode
...
-	} else {
-		var name string
-		if t.name != nil {
-			name = *t.name
-		} else {
-			name = t.performer.Name
-		}
-		performer, err = client.FindPerformerByName(ctx, name)
+	case t.name != nil:
+		performer, err = client.FindPerformerByName(ctx, *t.name)
```
WithoutPants: Optional: I think this reads better if the name and stash id cases are located next to each other.

internal/manager/task_stash_box_tag.go
```diff
@@ -315,17 +323,12 @@ func (t *StashBoxBatchTagTask) findStashBoxStudio(ctx context.Context) (*models.
...
-	} else {
-		var name string
-		if t.name != nil {
-			name = *t.name
-		} else {
-			name = t.studio.Name
-		}
-		studio, err = client.FindStudio(ctx, name)
+	case t.name != nil:
+		studio, err = client.FindStudio(ctx, *t.name)
```
WithoutPants: Optional: I think this reads better if the name and stash id cases are located next to each other.

PR #6323, opened 2025-11-27
Bugfix: Update Markers to % Base Calc

### LINKED ISSUE #6286: Wrong Marker position on full screen
### Describe the bug

Scenes with markers with end time are on the wrong position and wrong size on full screen.
Maybe because of the ultra wide resolution? (3840x1200)

### Steps to reproduce

1. Go to scene with markers which have end time
2. Markers without end time are displayed correctly
3. Compare marker dash(es) on the time table with "normal" view and in full screen mode.

### Expected behaviour

Same relative position on the timeline.

Normal view:
Image

Full screen:
Image

### Stash version

v0.29.3

### Device details

Google Chrome v142.0.7444.163

### Relevant log output

```shell

```

Gykes:
Fixes range markers (markers with end times) displaying at incorrect positions in fullscreen mode. The issue was that range markers used fixed pixel positioning calculated at render time, which didn't scale when the player resized. Changed to percentage-based calc() positioning to match how dot markers work. Added in a bunch of comments so I could track my math lol.

Tested on my Mac with Waterfox:
Created a marker at 1sec-10minutes. Pre fix there was an offset due to the bug. After the fix the marker was in the expected location when going fullscreen.

Fixes #6286

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/markers.ts
+++ b/ui/v2.5/src/components/ScenePlayer/markers.ts
@@ -158,10 +158,7 @@ class MarkersPlugin extends videojs.getPlugin("plugin") {
       startPercent * 0.3
     }px)`;
 
-    // minimum width of 8px, using calc for percentage-based width
-    rangeDiv.style.width = `max(calc(${widthPercent}% - ${
-      widthPercent * 0.3
-    }px), 8px)`;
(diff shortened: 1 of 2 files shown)


smith113-p:
Why not use min-width for the 8px min width, to slightly simplify the calc expression?

Gykes:
Because I am a complicated person who hates CSS lol. You are not wrong tho, can change it over.

PR #6328, opened 2025-11-27
[hwaccel] increase timeout, fix formatting

feederbox826:
Thanks to @Gykes for helping find the formatting error

when printed in %d it returns the nanosecond equivalent, when cast to string, formatting is done properly

https://go.dev/play/p/jtEeJs6K2mG


feederbox826:
BTW 10s increase since seemed to be not enough for most nvidia users (reported on discord + [unraid forums](https://forums.unraid.net/topic/177317-support-feederbox826stash-s6-alternative-stash-container/page/2/#findComment-1592741))

Issue #3171, opened 2022-11-22
Synology NAS and folders table


DampToast:
**Describe the bug**
With the introduction of the 'folders' table in 0.17 (I believe), all folders under the root are added.  Synology makes hidden folders called '@eaDir' in any folder that contains one or more videos and inside that folder in a folder with the name of every video.  So if you have a folder with 100 videos in it, you get one '@eaDir' folder that has 100 folders in it, each named with the name of a video.  There is nothing in here that Stash needs or can use and no way for the user to turn this off on Synology's end.  They are only hidden in that browsing in the Synology UI they will not be displayed but they are displayed through most other protocols.

**To Reproduce**
1. Buy the most expensive Synology NAS and max out the hard drives. (optional)
2. Run Stash via Docker on a Synology machine.
3. Scan your files and browse the 'folders' table.

**Expected behavior**
It doesn't technically hurt anything, but there will be one useless entry per video a user has on Synology systems.  This could impact performance long term, although I have no clue by how much.  Could be negligible.

**Stash Version: (from Settings -> About):**
v0.17.2-31-gce17230c (should be any 0.17 version though)

**Additional context**
I did add '.*@eaDir.*' to both the video and gallery ignore regex and ran a clean which removed all these entries.

If you have two videos:

/content/A.mp4
/content/B.mp4

You also get the files:
/content/@eaDir/A.mp4/SYNOINDEX_MEDIA_INFO
/content/@eaDir/B.mp4/SYNOINDEX_MEDIA_INFO

which adds the following to the folders table:
/content/@eaDir/A.mp4/
/content/@eaDir/B.mp4/

VonTittyslappen:
I too run Stash on Synology NAS so if you need someone to test this issue out with let me know on Discord (have just returned after a long long absence!)

WithoutPants:
Stash should probably not be storing folder information for folders containing no relevant files.

2t9z16yo:
even if i add a regex to ignore these folders ".*@eaDir.*" in Library, they still throw these error messages during a clean generated files operation (because these eaDIRs exist in the /generated subdir as well. any chance we can add a global ignore option? thanks!

PR #6332, opened 2025-11-27
Bugfix: Add defaultExcludes List 

Gykes:
This add a defaultExcludes list. Currently it only contains a way to block out the Synology hidden folder but it can be expanded to include other patters if there are any.

Added to video and image scanner

I don't have a way to test this on Synology systems so this is 50% "that looks good" and 50% hopes and prayers. Hoping that some Synology users are willing to test this for me

Fixes: #3171


WithoutPants:
This is an invalid solution to the original issue. It's not a great practice to have these permanently in the excludes list. These items show in the exclusion list in the UI, and while you can remove them in the UI, they return on refresh, on account of them always returning.

I'd rather not add built-in exclusions for various vendor built in folders. The solution to the original issue is for the user to add the ignored folder names themselves. The original issue was left open to address the scanning system creating folder database entries for folders that have no relevant content.

PR #6336, opened 2025-11-28
Resume after scrubbing, hide player UI faster

hckrman101:
This PR includes 2 improvements to the scene player:
1. After scrubbing, it resumes the video playback if it was playing before the scrub
2. Reduces the timeout for hiding the video UI from 2000ms to 700ms


feederbox826:
700ms seems awfully arbitrary, I would much prefer a configuration

hckrman101:
@feederbox826 700ms is what "felt" the most right (to me). I agree that a configuration would be useful for this.

Issue #6277, opened 2025-11-14
Thumbnail generation fails in some AVIF images


z9gccs4zkw-ship-it:
### Describe the bug

Hi,

I've been recently getting the following error in my logs:

```
Error   
image encoder error when running command </usr/bin/vips thumbnail_source [descriptor=0] .jpg[Q=70,strip] 640 --size down>: 
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 0
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 1
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 2
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 3
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 4
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 5
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 6
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 7
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 8
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 9
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 0
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 10
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 20
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 30
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 0
(vips:580): VIPS-WARNING **: 14:47:03.419: error in tile 0 x 10
source: bad seek to 250
```

This used to work in the previous version and only happens to some images, after some investigation it seems the VIPS-libheif library deprecated a "partial read fallback" so is now unable to process an avif/heic file if it requires a seek() to another part of the file.

Some .avif files require to seek other parts of the file to correctly decode the image, for example:

- images with EXIF blocks at the end
- tiled HEIF images
- grid images
- HEVC chunks that reference offsets backward in the stream

It seems the problem arises because the command uses [descriptor=0]  which means it's reading linearly from STDIN and cannot randomly seek the file and it fails on all the avif files that require to seek to another part of the file.

[...]

PR #6288, opened 2025-11-17
Feature: AVIF support

Gykes:
From what I can tell ffmpeg does most of the heavy lifting for AVIF so this is a pretty small addition. I updated DLNA as well to be able to access this addition as well.

fixes: https://github.com/stashapp/stash/issues/1729

PR #6337, opened 2025-11-28
Feature Request: Vips  AVIF Support

Gykes:
Extends https://github.com/stashapp/stash/pull/6288 to include vips.

Fixes: https://github.com/stashapp/stash/issues/6277


pkg/image/thumbnail.go
WithoutPants: Let's update this as well:
- if vips is not nil
  - if on windows and not in zip
    -  use vipsPath (return)
  - not on windows
    - use vips stdin (return)
 - fallback to ffmpeg stdin

PR #6340, opened 2025-11-28
Manually Search Stash ID - Edit Page - Scenes, Studios

Gykes:
A followup to https://github.com/stashapp/stash/pull/6284

This adds in the same support to the other modals that support StashIDs: Scenes, Studios, and Tags.

Closes: https://github.com/stashapp/stash/issues/6271

### DIFF
--- a/ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
+++ b/ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
@@ -232,15 +232,15 @@ export const StudioSearchResult: React.FC<IStudioResultProps> = ({
   );
 };
 
-// Helper to get entity type display name for i18n
-function getEntityTypeDisplayName(entityType: StashBoxEntityType): string {
+// Helper to get entity type message id for i18n
+function getEntityTypeMessageId(entityType: StashBoxEntityType): string {
   switch (entityType) {
     case "performer":
-      return "Performer";
+      return "performer";
     case "scene":
-      return "Scene";
+      return "scene";
     case "studio":
-      return "Studio";
+      return "studio";
   }
 }
 
@@ -384,7 +384,9 @@ export const StashBoxIDSearchModal: React.FC<IProps> = ({
     );
   }
 
-  const entityTypeDisplayName = getEntityTypeDisplayName(entityType);
+  const entityTypeDisplayName = intl.formatMessage({
+    id: getEntityTypeMessageId(entityType),
+  });
 
   return (
     <ModalComponent


Gykes:
I modified some of the old code in the performers section to run off of the new utility. All 4 sections were doing the same thing so I figured it was best to break it out.

internal/api/resolver_query_scraper.go
```diff
@@ -353,6 +353,45 @@ func (r *queryResolver) ScrapeSingleStudio(ctx context.Context, source scraper.S
...
+			}); err != nil {
+				return nil, err
+			}
+			return ret, nil
+		}
+
+		return nil, nil
+	}
+
+	return nil, errors.New("stash_box_index must be set")
```
WithoutPants: This should refer to the non-deprecated field `stash_box_endpoint` instead.

pkg/stashbox/tag.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if tag.FindTag == nil {
+		return nil, nil
+	}
+
+	return &models.ScrapedTag{
+		Name:         tag.FindTag.Name,
+		RemoteSiteID: &tag.FindTag.ID,
+	}, nil
+}
```
WithoutPants: I think this should be returning a slice of tags rather than a single one. In the UI, we're potentially querying for a tag for which we may not have the exact name, similar to performers, scenes etc. I think if we're not searching by UUID, we should use [`queryTags`](https://github.com/stashapp/stash-box/blob/883a1500f05c298fc0b79a9181acc13ccfc83926/graphql/schema/schema.graphql#L24) instead of `findTag`.

internal/api/resolver_query_scraper.go
Gykes: Fixed

pkg/stashbox/tag.go
Gykes: Moving over to separate PR as requested to fix.

ui/v2.5/src/components/Shared/StashBoxIDSearchModal.tsx
```diff
@@ -132,8 +144,121 @@ export const PerformerSearchResult: React.FC<IPerformerResultProps> = ({
...
+function getEntityTypeDisplayName(entityType: StashBoxEntityType): string {
+  switch (entityType) {
+    case "performer":
+      return "Performer";
+    case "scene":
+      return "Scene";
+    case "studio":
+      return "Studio";
+  }
+}
```
WithoutPants: This needs to be internationalised as well. This should return a message id, which would be translated using `intl` in the calling code.

Gykes: Had to google what that even meant but I think I was able to copy it over from how another part handled it.

Gykes:
I branched the tags stuff off of this one a little earlier. Once this one looks good I will rebase with any changes we make and then PR it.

WithoutPants (approved):

PR #6342, opened 2025-11-29
Add fallback for 0-dimension images (webp animations)

OtterBotSociety:
I'm still running into an issue with animated WebP files. After verifying the output of `ffprobe` directly (on various versions from 5.0 up through 8.0), I've determined there isn't an FFMpeg-side fix. Certain WebP animations are still being detected with a width/height of 0, which ([just like in my earlier PR](https://github.com/stashapp/stash/pull/6006)) causes the gallery view to fail. 
 
Please adjust anything needed to align better with the project's coding conventions. Specifically, I'm not fully certain whether this is the best approach for delegating to the fallback decoder or if it would be cleaner to directly `return decorateFallback(...)`, among other possibilities.

### DIFF
--- a/pkg/file/image/scan.go
+++ b/pkg/file/image/scan.go
@@ -74,6 +74,25 @@ func (d *Decorator) Decorate(ctx context.Context, fs models.FS, f models.File) (
 		Height:   probe.Height,
 	}
 
+	// FFprobe has a known bug where it returns 0x0 dimensions for some animated WebP files
+	// Fall back to image.DecodeConfig in this case.
+	// See: https://trac.ffmpeg.org/ticket/4907
+	if ret.Width == 0 || ret.Height == 0 {
+		logger.Warnf("FFprobe returned invalid dimensions (%dx%d) for %q, trying fallback decoder", ret.Width, ret.Height, base.Path)
+		c, format, err := decodeConfig(fs, base.Path)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


OtterBotSociety:
Here's an example 3-frame animated webp image that ffprobe fails to parse (returning 0x0) but DecodeConfig parses successfully (returning 200x150)
<details>
  <summary>Epilepsy warning - example webp animation (flashing colors!)</summary>

test_animated_webp_example

</details>

You can recreate this image with imagemagick and libwebp-dev:

```bash
convert -size 200x150 -background red -fill white -gravity center -pointsize 40 label:"1" frame1.png
convert -size 200x150 -background green -fill white -gravity center -pointsize 40 label:"2" frame2.png
convert -size 200x150 -background blue -fill white -gravity center -pointsize 40 label:"3" frame3.png
img2webp -lossy -d 100 -loop 0 frame*.png -o test_animated.webp
```

feederbox826:
not sure if this will catch the actual cause of the edge cases - since webm/p is based on mkv, the dimensions can actually be dynamic [1](https://rebane2001.com/weirdm/) [2](https://www.reddit.com/r/discordapp/comments/y7mq59/i_need_this_video_of_the_music_all_i_want_for/) [3](https://cdn.discordapp.com/attachments/877217596129304602/1029801437946974289/gura2.webm?size=4096)

WithoutPants (approved):
I'm satisfied that this will at least iteratively improve scanning for such files.

PR #6343, opened 2025-11-29
Feature: Add "Troubleshooting Mode"

### LINKED ISSUE #4508: [Feature] Master toggle to disable all plugins
When troubleshooting it is often preferable to disable all plugins (and css/js injection). As plugins gain popularity the amount a user has to individually disable will increase. This is going to be a laborious process between the plugins, custom CSS, and custom JS.

Implement a master toggle in the settings to disable (and enable) all plugins

Between individually toggling to disable all plugins, custom CSS, and custom JS, maybe a better alternative to the above would be to have a troubleshooting mode toggle which just disables all plugins, custom css, and custom JS at once.

Settings > Interface > Basic Settings > Troubleshooting mode <toggle>

image

Gykes:
This PR adds a new toggle to `Settings -> Interface`(Image Below) which allows users to toggle Disable Customization in the new troubleshooting section. I don't have a reason for where I placed it on the page it was just kind of random. If someone has a better idea for where to place it then it's an easy change to make. I felt like adding in the new section was a good idea for future proofing but I can simplify it if needed.

This disables all plugins and custom css, JS, and localization.  To keep in line with other settings this will require a manual page reload by the user once enabled.

Tests: 
Enabled the Stash Notes plugin -> enable and refresh -> plugin gone from nav bar
Enabled custom css to change background -> enable and refresh -> background back to normal
Enabled custom JS to add a popup alert on reload -> enable and refresh -> popup no longer appears.
Enabled custom localization to change "Scenes" to "blah" -> enabled and refresh -> goes back to scenes

Fixes: #4508

UI Button:
Screenshot 2026-01-17 at 20 38 39

Confirm Modal: 
Screenshot 2026-01-17 at 20 38 46

Banner/Exit: 
Screenshot 2026-01-17 at 20 38 54

### DIFF
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -392,6 +392,9 @@ input ConfigInterfaceInput {
   customLocales: String
   customLocalesEnabled: Boolean
 
+  "When true, disables all customizations (plugins, CSS, JavaScript, locales) for troubleshooting"
+  disableCustomizations: Boolean
+
   "Interface language"
   language: String
 
@@ -464,6 +467,9 @@ type ConfigInterfaceResult {
   "Custom Locales"
   customLocales: String
   customLocalesEnabled: Boolean
+
+  "When true, disables all customizations (plugins, CSS, JavaScript, locales) for troubleshooting"
+  disableCustomizations: Boolean
 
   "Interface language"
   language: String
--- a/internal/api/resolver_mutation_configure.go
+++ b/internal/api/resolver_mutation_configure.go
@@ -513,6 +513,8 @@ func (r *mutationResolver) ConfigureInterface(ctx context.Context, input ConfigI
 
 	r.setConfigBool(config.CustomLocalesEnabled, input.CustomLocalesEnabled)
 
+	r.setConfigBool(config.DisableCustomizations, input.DisableCustomizations)
+
 	if input.DisableDropdownCreate != nil {
 		ddc := input.DisableDropdownCreate
 		r.setConfigBool(config.DisableDropdownCreatePerformer, ddc.Performer)
--- a/internal/api/resolver_query_configuration.go
+++ b/internal/api/resolver_query_configuration.go
@@ -156,6 +156,7 @@ func makeConfigInterfaceResult() *ConfigInterfaceResult {
 	javascriptEnabled := config.GetJavascriptEnabled()
 	customLocales := config.GetCustomLocales()
 	customLocalesEnabled := config.GetCustomLocalesEnabled()
+	disableCustomizations := config.GetDisableCustomizations()
 	language := config.GetLanguage()
 	handyKey := config.GetHandyKey()
 	scriptOffset := config.GetFunscriptOffset()
@@ -183,6 +184,7 @@ func makeConfigInterfaceResult() *ConfigInterfaceResult {
 		JavascriptEnabled:            &javascriptEnabled,
 		CustomLocales:                &customLocales,
 		CustomLocalesEnabled:         &customLocalesEnabled,
+		DisableCustomizations:        &disableCustomizations,
 		Language:                     &language,
 
 		ImageLightbox: &imageLightboxOptions,
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -421,7 +421,7 @@ func cssHandler(c *config.Config) func(w http.ResponseWriter, r *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
 		var paths []string
 
-		if c.GetCSSEnabled() {
+		if c.GetCSSEnabled() && !c.GetDisableCustomizations() {
 			// search for custom.css in current directory, then $HOME/.stash
 			fn := c.GetCSSPath()
 			exists, _ := fsutil.FileExists(fn)
@@ -439,7 +439,7 @@ func javascriptHandler(c *config.Config) func(w http.ResponseWriter, r *http.Req
 	return func(w http.ResponseWriter, r *http.Request) {
 		var paths []string
 
-		if c.GetJavascriptEnabled() {
+		if c.GetJavascriptEnabled() && !c.GetDisableCustomizations() {
 			// search for custom.js in current directory, then $HOME/.stash
 			fn := c.GetJavascriptPath()
 			exists, _ := fsutil.FileExists(fn)
@@ -457,7 +457,7 @@ func customLocalesHandler(c *config.Config) func(w http.ResponseWriter, r *http.
 	return func(w http.ResponseWriter, r *http.Request) {
 		buffer := bytes.Buffer{}
 
-		if c.GetCustomLocalesEnabled() {
+		if c.GetCustomLocalesEnabled() && !c.GetDisableCustomizations() {
 			// search for custom-locales.json in current directory, then $HOME/.stash
 			path := c.GetCustomLocalesPath()
 			exists, _ := fsutil.FileExists(path)
--- a/internal/manager/config/config.go
+++ b/internal/manager/config/config.go
@@ -194,6 +194,7 @@ const (
 	CSSEnabled                          = "cssenabled"
 	JavascriptEnabled                   = "javascriptenabled"
 	CustomLocalesEnabled                = "customlocalesenabled"
+	DisableCustomizations               = "disable_customizations"
 
 	ShowScrubber        = "show_scrubber"
 	showScrubberDefault = true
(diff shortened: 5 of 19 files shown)


sleetx:
A few thoughts:

- Is this necessary? or will it end up causing more confusion during troubleshooting? We could simply add a paragraph to the In-App Manual or the Stash Docs for troubleshooting tips like this.

- Assuming we move forward, the "Interface" settings category is probably not the best spot to place it, since it involves both scripting and plugin deactivation. Maybe under Tools, Logs, or Security? 

- I'd suggest branding it as "Disable Customizations" rather than "Troubleshooting". And a button rather than a toggle switch. After clicking the button, it could force the page to automatically reload. And then the alternative button appears as "Re-Enable Customizations".

- Are you storing exactly which options affected? It looks like you added a boolean. Stash instances will have various things turned on or off already. For example, users may already have inactive plugins that shouldn't be re-activated. Or they may have CSS enabled but not JS, etc.

Gykes:
> A few thoughts:
> 
>     * Is this necessary? or will it end up causing more confusion during troubleshooting? We could simply add a paragraph to the In-App Manual or the Stash Docs for troubleshooting tips like this.
> 
>     * Assuming we move forward, the "Interface" settings category is probably not the best spot to place it, since it involves both scripting and plugin deactivation. Maybe under Tools, Logs, or Security?
> 
>     * I'd suggest branding it as "Disable Customizations" rather than "Troubleshooting". And a button rather than a toggle switch. After clicking the button, it could force the page to automatically reload. And then the alternative button appears as "Re-Enable Customizations".
> 
>     * Are you storing exactly which options affected? It looks like you added a boolean. Stash instances will have various things turned on or off already. For example, users may already have inactive plugins that shouldn't be re-activated. Or they may have CSS enabled but not JS, etc.

- No, it was brought up and requested in discord as quite a few users have issues and the first step told to them is to disbale plugins as it seems like a plugin issue rather than a Stash issue. This is a quick and easy way to help users troubleshoot their problems.

- I put it there just due to the fact of it is dealing with interface stuff. Disabling CSS, JS, etc. I'm not married to where it is currently it was just a location that I figured might work, can be moved simply. 

- I created it as "Troubleshooting" as the category as we could potentially add new things there in the future but the actual toggle is labled as "Disable Customizations" within that category.

- So the actual toggle itself is a toggle and a bool, yes. How I have it currently is it does not modify your current settings at all. It doesn't toggle individual plugins on or off it either enables or disables the entire block. Imagine it like a breaker box. My implementation is flipping the main breaker to each section. If you have a breaker of it will continue to stay off regardless of if I flip the main breaker.


Edit: I specifically kept it so that it did not reload to page to keep it inline with other custom implementations. When you enable CSS you need to refresh the page to have it take effect. If I have it refresh the page on toggle then if someone is modifying settings, and clicks it by accident, it would refresh and they would lose their other settings. Just a small bad UX experience but if the consensus is to make it reload then i'm fine with that.

echo6ix:
@Gykes Thanks!

Suggestion:

1. You labeled the toggle `Disable customizations`. For utmost clarity I suggest renaming it to `Disable plugins and customizations`. Since it affects both, there’s no reason to omit "plugins" when calling out "customizations". And many people probably won't read the fine print😏
2. Customizations are in the `Interface` section, and plugins in `Plugins` section, but this toggle affects both items. Placing it in either section would exclude the other. It might be the most intuitive as a core toggle in the `System` section, located as the last entry.

feederbox826:
I think the concept itself is a great idea, I don't have much of an opinion on where it should go or what the naming should be. We've had multiple issues come up where it was so difficult to replicate and ended up being solved by "oh I had custom CSS" or "oh I had x y z theme and plugin conflict".

I would additionally add some sort of overlay to act as a notice, since the user would probably not want to run in this mode for extended periods of time

something akin to the "This copy if windows is not activated" overlay and additionally clickable to quickly turn it off

```
Troubleshooting Mode
stash v0.29.3 aaaaaa
```

Debug mode in logs would also be great but persistence would be difficult

Gykes:
> @Gykes Thanks!
> 
> Suggestion:
> 
>     1. You labeled the toggle `Disable customizations`. For utmost clarity I suggest renaming it to `Disable plugins and customizations`. Since it affects both, there’s no reason to omit "plugins" when calling out "customizations". And many people probably won't read the fine print😏
> 
>     2. Customizations are in the `Interface` section, and plugins in `Plugins` section, but this toggle affects both items. Placing it in either section would exclude the other. It might be the most intuitive as a core toggle in the `System` section, located as the last entry.

1) I kind of consider plugins to be part of customization for the most part and didn't want to make the toggle too lengthy but noted none the less.

2) I think you just mean from a UI perspective that it would exclude the other as at code level neither are. I'm okay with moving it to system, shouldn't be a problem

echo6ix:
I'm sure @WithoutPants would have a good idea of where this should be located in the settings.

But at the same time, I wouldn't waste much effort looking for an optimal location. Because it looks like the entire Settings hierarchy needs reorganized eventually, with more conventional user-friendly node headings. I've been using it since the first version and sometimes I'm still searching where settings are situated. 😅

WithoutPants:
My proposal is to have a button, probably in the left panel of the settings screen, to activate this mode. It will bring up a dialog explaining what it will do, and if the user confirms, the system will apply the mode and reload the UI. Having an overlay as feederbox mentioned is a great idea and also provides a quick and easy way to switch back to normal mode. Having it override the log level to at least debug would also be beneficial if we can easily do it in the backend.

Gykes:

Gykes:
@WithoutPants 

`My proposal is to have a button, probably in the left panel of the settings screen, to activate this mode`


Do you mean on the main page underneath Advanced Mode and have the same style toggle?

echo6ix:
> My proposal is to have a button, probably in the left panel of the settings screen, to activate this mode. It will bring up a dialog explaining what it will do, and if the user confirms, the system will apply the mode and reload the UI. Having an overlay as feederbox mentioned is a great idea and also provides a quick and easy way to switch back to normal mode. Having it override the log level to at least debug would also be beneficial if we can easily do it in the backend.

Good idea.

Before this gets finalized, it might be a good idea to revisit the optimal name for the toggle.

- `Troubleshoot Mode` is a clear-ish but long.
- `Debug Mode` is a concise, but a bit technical.
- `Safe Mode` is both concise and clear.  I believe this term is commonly used to describe a minimal fallback environment when booting into devices or operating systems for troubleshooting - exactly the intent here - perhaps best for a general audience. The only disadvantage is you just implemented a `SFW Content Mode` elsewhere - could these cause confusion with another? Maybe any confusion is not an issue with the confirmation dialog you mentioned.

sleetx:
I would not use "Safe Mode" as that could be confused with "SFW mode".

I like the idea of the logs being more easily accessible. Perhaps showing the most recent lines at the bottom of each screen, rather than needing to browse to Settings > Logs.

I also like @feederbox826 idea of a permanent overlay indicating this mode is turned on (and an easily accessible link to turn it off). Something like this: 

image

feederbox826:
maybe also add a default css blur to images (studio/ performer/ cover/ gallery), saves a lot of steps for censoring

both can be added via css, just add a root variable that activates it

Gykes:
> maybe also add a default css blur to images (studio/ performer/ cover/ gallery), saves a lot of steps for censoring
> 
> both can be added via css, just add a root variable that activates it

10 minutes, eh?

discourse-stashapp:

RyanAtNight:
I really like this PR idea and where this discussion has been going. I love the idea of putting the toggle in the left-hand pane under or above "Advanced Mode", but prompting the user with a dialog to confirm before refreshing the page.

Now for what I'd like to add/suggest on top of this:

> Before this gets finalized, it might be a good idea to revisit the optimal name for the toggle.
>
> Troubleshoot Mode is a clear-ish but long.
Debug Mode is a concise, but a bit technical.
Safe Mode is both concise and clear. I believe this term is commonly used to describe a minimal fallback environment when booting into devices or operating systems for troubleshooting - exactly the intent here - perhaps best for a general audience. The only disadvantage is you just implemented a SFW Content Mode elsewhere - could these cause confusion with another? Maybe any confusion is not an issue with the confirmation dialog you mentioned.

If we did do something like a "Troubleshoot Mode", I'd add a "?" icon next to it so that the user can hover their mouse pointer over it to get the full details in a tooltip.
image

And if I had a vote, I'd put it towards "Troubleshoot Mode" because a non-technical user might just outright dismiss and pass over a button for "debug mode" as it sounds like a programmer term. "Safe mode" would have worked, it it wasn't for the NSFW nature of this project.

"Troubleshoot mode" is the next concise option by process of elimination. And pair that name with a mouse-over tooltip, and it might prevent a lot of tech support requests.

Gykes:
Pushed some changes and update the UI images to reflect

DogmaDragon (changes requested):
Needs to be documented in the manual.

WithoutPants:
Changed the presentation of Troubleshooting Mode to have a red border around the viewport, and moved the alert text to the bottom right with a more terse exit message:

image

Also removed the troubleshooting mode button in the settings page when the mode is active.

WithoutPants:
I'll merge this as is for now. Any further refinements can be done in a separate PR. Edit: I'll add manual documentation first

WithoutPants (approved):

DogmaDragon (approved):

PR #6344, opened 2025-11-29
Feature: Scene Duplicate Filter

### LINKED ISSUE #3441: [Feature] Add "Duplicated (Stash ID)" filter
With the "Duplicated (phash)" filter, we can merge scenes that have a duplicate phash, but this doesn't account for all duplicates of a scene. Having a duplicated stash ID filter would give us another option.

A "Duplicated (Stash ID)" filter for scenes that would only show scenes that share a Stash ID with one or more other scenes. The working SQL for that is:
```sql
SELECT ssi.stash_id, sc.title
FROM scenes sc
  JOIN scene_stash_ids ssi
    ON sc.id = ssi.scene_id
  GROUP BY ssi.stash_id
  HAVING COUNT(*) > 1;
```

Unfortunately I don't know GraphQL (and don't really have the time to learn it) otherwise I would implement this myself. In theory it should be pretty simple to add.

An alternative I've thought about, that may be better in the long run, is to just have a "Duplicated" filter that brings up a modal similar to the tag filter, with options for each scene detail. This would render both this request and the current "Duplicated (phash)" filters obsolete. With this idea, you could select multiple fields and all would have to match.

Gykes:
Currently, we have a filter for duplicate PHash. I revamped it a little bit to allow other duplicate checks. I kept PHash check but added in StashID and TItle. THe stashID is a current open feature request and the title is from me as I like to dedupe via titles.

Fixes: https://github.com/stashapp/stash/issues/3441

EDIT: I added in URl(s) as well per Smiths request it's just not in the img
UI:
Screenshot 2025-11-28 at 22 48 38
Screenshot 2025-11-28 at 22 48 49


smith113-p:
Would it be hard to also add URL to this?

feederbox826:
> Would it be hard to also add URL to this?

Dicussed this in VC, didn't see much use or practical application for URLs that wouldn't already be covered by previous filters

Gykes:
@smith113-p I managed to get URls in there for you. This PR just got more complicated as I had to merge https://github.com/stashapp/stash/pull/6352 into this one.

Gykes:

Gykes:
I believe I will need to rework this some. With the URL changes being reverted from .30 I need to remove it from here as well.

SikrytSquare:
I have been waiting for this for too long. Hope it gets merged soon. Huge thanks @Gykes

discourse-stashapp:

discourse-stashapp:

https://discourse.stashapp.cc/t/finding-duplicates-by-title/5085/2

WithoutPants:
I think the graphql interface should be cleaned up a little. Here's my proposal:
- rename existing `PHashDuplicationCriterionInput` to `DuplicationCriterionInput`
- add optional boolean `phash` field. If present, scenes will be filtered based on if the phash is duplicated or not
- add optional boolean `url` field. Same logic as above.
- add optional boolean `stash_id` field. Same logic as above.
- deprecate existing `duplicated` field, to be replaced with `phash` field - it should still maintain the legacy behaviour.
- `distance` field should be documented to indicate that it is related to `phash`, but should also note that it is not implemented

This means that to filter on duplicate urls, you would have:
```
scene_filter: {
  duplicated: {
    url: true
  }
}
```

The UI code will need to be adapted to handle this style of criterion.

graphql/schema/types/filters.graphql
```diff
@@ -709,8 +715,8 @@ input FileFilterType {
   "Filter by modification time"
   mod_time: TimestampCriterionInput
 
-  "Filter files that have an exact match available"
-  duplicated: PHashDuplicationCriterionInput
+  "Filter files by duplication criteria (only phash applies to files)"
+  duplicated: DuplicationCriterionInput
```
WithoutPants: File should have its own duplicated criterion input - `FileDuplicateCriterionInput` which would only accept `phash` and `distance`.

graphql/schema/types/filters.graphql
```diff
@@ -75,10 +75,19 @@ input OrientationCriterionInput {
   value: [OrientationEnum!]!
 }
 
-input PHashDuplicationCriterionInput {
+input DuplicationCriterionInput {
+  "Deprecated: Use phash field instead. Kept for backwards compatibility."
```
WithoutPants: This needs to document what the behaviour of `duplicated` is now.
This should use `@deprecated(reason: ...)`

pkg/sqlite/scene_filter.go
```diff
@@ -277,26 +277,78 @@ func (qb *sceneFilterHandler) fileCountCriterionHandler(fileCount *models.IntCri
...
+		if duplicatedFilter.Duplicated != nil &&
+			duplicatedFilter.Phash == nil &&
+			duplicatedFilter.URL == nil &&
+			duplicatedFilter.StashID == nil &&
+			duplicatedFilter.Title == nil {
+			qb.addSceneFilesTable(f)
+			//nolint:staticcheck
+			qb.applyPhashDuplication(f, *duplicatedFilter.Duplicated)
+			return
+		}
```
WithoutPants: Why is it necessary to have a specific behaviour for the `duplicated` field that only works when no other fields are present? I'd have thought it should be identical behaviour to when `phash == true`.

WithoutPants (changes requested):
The criterion editor view needs to be implemented for the new criterion type:

Image

ui/v2.5/src/components/List/Filters/DuplicateFilter.tsx
```diff
@@ -0,0 +1,270 @@
...
+    <Form.Check
+      ref={ref}
+      id={id}
+      type="checkbox"
+      checked={value === true}
+      onChange={onClick}
+      label={label}
+    />
+  );
+};
```
WithoutPants: We have `IndeterminateCheckbox` (in `src/components/Shared`) already. It should be extended instead of creating a new duplicated component.

WithoutPants (changes requested):

WithoutPants (approved):

PR #6347, opened 2025-11-29
Feature: Stash ID Count Filter

### LINKED ISSUE #3890: Filter scenes with more than 1 stashid
I would like to see if a field can be added to the scene filter that would add sort by stashID count, since we now have a couple different endpoints, so it would be easier to see what scenes have more than one. I know you can filter by endpoint when the stashid is not null, but this isn't ideal when you're trying to scrape data from both TPDB and Stashdb, and match both up to the scene on the local instance.

Gykes:
This PR adds a new filter of "Stash Id Count" in the edit filter modal. This allows users to user greater than or less than to check for stashID counts against scenes. 

I decided against adding this to the sidebar as we dont current have a decent UI display similar to the slider. I feel like the slider wouldn't be good for this but I can do it that way if people want it.

Fixes: https://github.com/stashapp/stash/issues/3890

UI:
Screenshot 2025-11-29 at 00 41 07


Gykes:

discourse-stashapp:

RyanAtNight:
I love this PR so much that I'm rocking it in my own local fork right now. It solves a need I had where I wanted to filter for scenes that have not been scraped by any Stash-box endpoint yet. Before this, I had to set up a saved filter for every stash-box endpoint I use, and then hop between those.

I don't know how strict we are in this repo about writing tests, but I had Claude Code do a code review, and it recommended some integration tests. Then, I had it write some tests for edge cases to give a warning or error if something breaks this in the future.

I have the tests right here, if you're interested, in this commit: https://github.com/stashapp/stash/commit/a45ef97c7b27506dd5799137029225b27eccdcc0

Let me know if you would like me to submit a PR to your branch (Gykes:add-greather-or-less-than-to-stashid-filter)

smith113-p:
> filter for scenes that have not been scraped by any Stash-box endpoint yet.

You can do that with "stash id" "is null" and leaving the endpoint field empty

RyanAtNight:
> You can do that with "stash id" "is null" and leaving the endpoint field empty

Ya, you're right. It must not have been that, that I was trying and failing to do simply before because I do remember trying that and it working. I have all these multi-step processes in my head that required multiple "Saved filters" but I never wrote down the process.

I think the 2 saved filters that hop back and forth between are "Needs StashDB.org scrape" and "Needs theporndb.net scrape" which I would have preferred to combine those filtered lists. What I'm finding hard to remember is why, right now. It's been at least a month since I was thinking about that, and I'm kind of at HDD capacity right now. So, my "stash" is starting to settle in for now. I may update here again if I think of the reason why I thought this was useful.

RyanAtNight:
I remember what it was that this pull request would help me with, specifically now.

I was trying to filter my scenes to only show just the scenes that had both a StashDB and a PornDB ID. If so, I would select them all and mark them as organized.

Yes, I know that I can automatically mark scenes as organized the moment they get identified in the Settings -> Tasks page, but I disabled that feature because it didn't work how I liked. How it works now: it tries to identify each scene with the first "Source". If it is identified, even if there are more sources, it quits at that point and marks the scene as organized. Now, when I run the identify task again, it just passes up those scenes marked as organized. What I want to happen: I want to identify as many scenes as possible with both StashDB and PornDB, not just the first one to identify it.

Back to how this pull request relates to this: I can use this as a stand-in for marking all scenes that have 2 or more Stash ID endpoints as organized manually.

WithoutPants (approved):

Issue #5238, opened 2024-09-09
URLs should be sorted alphabetically


cj12312021:
With the addition of URLs in objects such as performers, movies, I would have expected those URLs to be sorted upon update similar to aliases. However that does not appear to be the case.

Gykes:
Currently, it appears that they are sorted by insert order. How do we expect these to be sorted?

DogmaDragon:
My preferense would be alphabetical order excluding `http(s)://www.`. That would naturally group based on the domain.

WithoutPants:
I think really URLs should probably be an ordered list with perhaps a description field, but that's a larger change. At least if they are sorted by insert order there is some semblance of control over how they are displayed.

Gykes:
Agree with WP on this one. Leaving them as insert order allows the user to make it any order they want, including alphabetical.

DogmaDragon:
That seems like niche use case. Do you really expect people to re-order 10-30 URLs per performer manually just to have custom order? While alphabetical order loses theoretical control, in practice it would be a lot more useful to a lot more people. 

Basic dupe hunting would be made a lot easier where domain change, URL schema changes or extra URL paramaters (including trailing `/`) often lead to duplicate links.

Gykes:
Points are valid, If WP is fine with it then I can do it.

WithoutPants:
I don't have a strong opinion, go for it.

WithoutPants:
I've reverted #6352 due to it breaking the workflow for users that sort their URLs manually. Reopening to re-solve this issue.

discourse-stashapp:
This issue has been mentioned on **Stash**. There might be relevant details there:

https://discourse.stashapp.cc/t/plugin-requests/4755/3

PR #6352, opened 2025-11-29
Feature Request: Sort All Urls Alphabetically

Gykes:
This makes it so that all URLs in all major objects will be sorted alphabetically rather than creation date. For currently existing objects you need to force a refresh of your URLs to get it to sort properly. Add a URL, remove a URL, or retag. All newly created objects will alphabetize by default.

Fixes: https://github.com/stashapp/stash/issues/5238

### DIFF
--- a/ui/v2.5/src/utils/url.ts
+++ b/ui/v2.5/src/utils/url.ts
@@ -1,14 +1,8 @@
 /**
  * Extracts the sortable portion of a URL by removing the protocol and www. prefix
  */
-function urlSortKey(url: string): string {
-  let key = url;
-  // Remove http:// or https://
-  key = key.replace(/^https?:\/\//, "");
-  // Remove www. prefix
-  key = key.replace(/^www\./, "");
-  return key.toLowerCase();
-}
+const urlSortKey = (url: string): string =>
+  url.toLowerCase().replace(/^(https?:\/\/)?(www\.)?/, "");
 
 /**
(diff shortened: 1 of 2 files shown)


WithoutPants:
This seems unnecessarily complicated considering we can just sort the URLs in the UI.

Gykes:
Hehe, I was going to do it that way but I figured you would want it done before it even reaches the front end. 

Will modify for frontend sorting only.

feederbox826:
Do you want me to do the FE sorting? should be less than 10LOC if there's not much component reuse

ui/v2.5/src/utils/url.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Extracts the sortable portion of a URL by removing the protocol and www. prefix
+ */
+function urlSortKey(url: string): string {
+  let key = url;
+  // Remove http:// or https://
+  key = key.replace(/^https?:\/\//, "");
+  // Remove www. prefix
+  key = key.replace(/^www\./, "");
+  return key.toLowerCase();
+}
```
feederbox826: https://jsperf.app/kimenu

I thought you loved regex?

```js
const urlSortKey = (url: string): string =>
	url.toLowerCase().replace(/^(https?:\/\/)?(www\.)?/, '')
```

Gykes: Yea, yea, yea.... updated :P

WithoutPants (approved):

PR #6354, opened 2025-12-01
Add option for instant transitions in lightbox

hckrman101:
Adds checkbox to disable animations in image preview lightbox.


WithoutPants (approved):
Having to do a graphql changes for lightbox options is pretty rough. We should move this into the UI config at some point.

Looks good.

PR #6359, opened 2025-12-02
Date precision

### LINKED ISSUE #1576: [Feature] Support shorter dates like YYYY-MM and YYYY
The day of month or even the month is not known on many files. Currently only YYYY-MM-DD date format is supported. Could we also support YYYY-MM and YYYY?

When I try to enter a YYYY-MM or YYYY as the date, I get no error. It seems like it's going to accept the value, but it doesn't get saved.

I'd like see support for YYYY-MM and YYYY, e.g. 2021 or 2021-07.

If for some reason the month and day of month can't be blank, maybe allow for something like this: 2021-00-00

WithoutPants:
Resolves #1576 

Adds new `_precision` columns for each date field in the database.

Sets the precision based on the input string and outputs based on the precision field. 

Adjusted the UI to account for the new date strings.

### DIFF
--- a/pkg/models/date.go
+++ b/pkg/models/date.go
@@ -22,7 +22,7 @@ type Date struct {
 	Precision DatePrecision
 }
 
-var dateFormatPrecision []string = []string{
+var dateFormatPrecision = []string{
 	"2006-01-02",
 	"2006-01",
 	"2006",


Jakan-Kink:
I don't see anything in here about changes to the GraphQL, are there going to be any changes there to support this?

discourse-stashapp:

https://discourse.stashapp.cc/t/renameronupdate/1754/8

PR #6320, opened 2025-11-25
Add sticky selection toolbar to filtered lists

WithoutPants:
Shows a sticky selection toolbar at the top of the page when one or more items are selected. This change is currently limited to non-scene list pages - changes will be ported to scene list is a separate PR. Shows the current amount of items selected, and provides access to select all/none buttons, operations and display mode controls.

Screenshots:

image

PR #6364, opened 2025-12-03
Make the entire image top bar sticky

ghost:
5 days ago #6320 was merged, which implements a sticky behavior for the image selection toolbar. However, the regular image top bar itself is _**not**_ sticky. This behavior is in contrast to the top bar of the Scene tab, which _**is**_ sticky.

This PR makes the entire Image top bar sticky (and adjusts some styles to make it clean), which matches the behavior of the top bar in Scene search.

scenes top-bar (vanilla):
image

image top-bar (vanilla):
image

image top-bar sticky (this PR):
image

Another side-effect of this change was that it caused the Scenes top-bar to move to the center, but I think it looks a lot better:

scenes top-bar (vanilla):
image

scenes top-bar (this PR):
image
image

I ran `make validate` and CSS linting passed.

### DIFF
--- a/ui/v2.5/src/components/List/styles.scss
+++ b/ui/v2.5/src/components/List/styles.scss
@@ -919,13 +919,16 @@ input[type="range"].zoom-slider {
 .filtered-list-toolbar {
   align-items: center;
   background-color: $body-bg;
+  border-radius: 0.5rem;
   gap: 0.5rem;
   justify-content: center;
 
   // offset the main padding
+  margin-left: auto;
+  margin-right: auto;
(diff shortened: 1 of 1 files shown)


ghost:
re: this PR causing the scenes top bar to move to the center

I'm not sure what y'all have in mind for the new Scene filter and side panel, so let me know if I should make this CSS change a more targeted fix for the image bar container specifically. Thanks!

ghost:
Likely redundant with the new changes coming in recently. Will re-open if I see a need

PR #6368, opened 2025-12-03
add autostart button to videoplayer

cj12312021:
This pull request adds the auto-start button, which was previously behind the interface settings, directly to the video player, similar to most modern video players.

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/styles.scss
+++ b/ui/v2.5/src/components/ScenePlayer/styles.scss
@@ -410,10 +410,6 @@ $sceneTabWidth: 450px;
         width: 2.5em;
       }
 
-      .vjs-autostart-button {
-        display: none;
-      }
-
       .vjs-progress-control {
         height: 2em;
         width: 100%;


DogmaDragon:
Please include a screenshot to pull requests making changes to the UI.

cj12312021:
Still working on this PR. I planned on providing screenshots once I've polished the styling

cj12312021:
I've gotten the CSS to a point I'm happy with. Here are pictures.

Acfive state:
Screenshot 2025-12-03 012029


Disabled state:
Screenshot 2025-12-03 011951

I would like to eventually create sections within the settings button where buttons such as the playback options can be tucked into. But that's a separate conversation. I'll address the remaining linter issues tomorrow.

DogmaDragon:
I like it. I have 2 questions.
1. Does the button have alt text? Most ofter buttons do (with the exception of the wheel) and I think it would add some clarity without any downsides. 
2. Will it properly work with **Settings** > **Interface** > **Scene Player** > **Auto-start video** and **Auto-start video when playing selected** toggles?

cj12312021:
This PR should be ready for a review now.

@DogmaDragon

> Does the button have alt text? Most ofter buttons do (with the exception of the wheel) and I think it would add some clarity without any downsides.

Yes, this new button also display descriptive text on hover. Trying to capture a screenshot of this while keep the text on screen proved difficult but here is the code:
```
    if (this.autostartEnabled) {
      this.addClass("vjs-icon-play-circle");
      this.controlText(this.localize("Auto-start enabled (click to disable)"));
    } else {
      this.addClass("vjs-icon-cancel");
      this.controlText(this.localize("Auto-start disabled (click to enable)"));
    }
```
The `Auto-start enabled (click to disable)` text will display on hover when the button is currently enabled. The `Auto-start disabled (click to enable)` text will display on hover when the button is disabled.

> Will it properly work with Settings > Interface > Scene Player > Auto-start video and Auto-start video when playing selected toggles?

Yes, the code will sync all toggle actions between the video player button as well as settings interface button.

ui/v2.5/src/components/ScenePlayer/autostart-button.ts
```diff
@@ -0,0 +1,126 @@
...
+    return `vjs-autostart-button ${super.buildCSSClass()}`;
+  }
+
+  private updateIcon() {
+    this.removeClass("vjs-icon-play-circle");
+    this.removeClass("vjs-icon-cancel");
+
+    if (this.autostartEnabled) {
+      this.addClass("vjs-icon-play-circle");
+      this.controlText(this.localize("Auto-start enabled (click to disable)"));
```
WithoutPants, line 31: Could we accept localised strings in the constructor options instead? That way we could put the strings into the localisation files.

WithoutPants (changes requested):
The styling needs to be adjusted for viewports between 577px and 700px. The cog icon and full screen buttons end up chopped off:

Image

ui/v2.5/src/components/ScenePlayer/autostart-button.ts
cj12312021: This was interesting. I saw we did localization this way for some of the other VideoJS buttons. I kept the approach for consistency, but I could add a localised string here if you don't mind the inconsistency.

WithoutPants: Fair point. Leave it as is for consistency. We should address localisation in another PR.

cj12312021:
> The styling needs to be adjusted for viewports between 577px and 700px. The cog icon and full-screen buttons end up chopped off

What are your thoughts on hiding the auto-play button for now on mobile until button consolidation has been done? The button overload is more apparent on mobile

WithoutPants:
> What are your thoughts on hiding the auto-play button for now on mobile until button consolidation has been done? The button overload is more apparent on mobile

Sounds good

ui/v2.5/src/components/ScenePlayer/styles.scss
WithoutPants, line 398: The buttons are still currently overflowing offscreen between `576px` and ~`676px` width. I reckon it would make sense to change this clause to `(max-width: 768px)` and move the hide autostart button to a separate `(max-width: 576px)` clause.

This is with that fix at `600px` width:
Image

cj12312021: Sorry, I'm just seeing your message. I'm currently unable to test the change, but I've applied the update as you've mentioned.

PR #6370, opened 2025-12-03
Query for image query result metadata separately

### LINKED ISSUE #6319: Make "Megapixels" and "Filesize" in the GraphQL query in Images page optional or remove altogether
This is not a bug report per se and should be labeled as "Improvement" but I can't edit the label.

As my Images count have been growing (+4M files) I've noticed that the Images page take a long time to load (~9s), it's not the images inside what makes it slow but the GraphQL Query. Checking the network tab I can see this:

Image

After playing with it in the GraphQL Playground I've seen a x10 improvement if I remove the megapixels and filesize from the query, if removed it only takes ~1s to load (Can be even faster removing the count part but that at least is interesting info)

```
query FindImages($filter: FindFilterType, $image_filter: ImageFilterType, $image_ids: [Int!]) {
    findImages(filter: $filter, image_filter: $image_filter, image_ids: $image_ids) {
        count
        megapixels
        filesize
        images {
          ...SlimImageData
          __typename
    }
    __typename
  }
}
```

I think the total filesize and mexapixel count of all the files is unnecessary and makes the Images page almost unusable on large DB's if it takes 9s each time I click next page. It should be an option to opt-in out or just removed from the query.

### Steps to reproduce

1. Have a large number of images
2. The images page will take a long time to load.

### Expected behaviour

Should not take a long time to load

### Stash version

v0.29.3

WithoutPants:
Makes the query for total megapixels and filesize separate from the main image query so that the former doesn't cause delays in returning the images themselves.

Shows `(...)` in the metadata info section while the metadata info query is running:

image

Haven't yet noticed a huge difference on my production system of >300k images. Going to try a larger db when I migrate it.

Resolves #6319

### DIFF
--- a/ui/v2.5/src/components/List/ListProvider.tsx
+++ b/ui/v2.5/src/components/List/ListProvider.tsx
@@ -1,10 +1,5 @@
 import React, { useMemo } from "react";
-import {
-  IListSelect,
-  useCachedQueryResult,
-  useListSelect,
-  useMetadataFilter,
-} from "./util";
+import { IListSelect, useCachedQueryResult, useListSelect } from "./util";
 import { isFunction } from "lodash-es";
 import { IHasID } from "src/utils/data";
 import { useFilter } from "./FilterProvider";
@@ -140,10 +135,12 @@ export const QueryResultContext = <
   }, [filter, filterHook]);
 
   // metadata filter is the effective filter with the sort, page size and page number removed
-  const metadataFilter = useMetadataFilter(effectiveFilter);
+  const metadataFilter = useMemo(
+    () => effectiveFilter.metadataInfo(),
+    [effectiveFilter]
+  );
 
   const result = useResult(effectiveFilter);
-
   const metadataInfo = useMetadataInfo?.(metadataFilter);
 
   // use cached query result for pagination
--- a/ui/v2.5/src/components/List/util.ts
+++ b/ui/v2.5/src/components/List/util.ts
@@ -483,19 +483,6 @@ export function useCachedQueryResult<T extends QueryResult>(
   return cachedResult;
 }
 
-// used to generate a metadata info filter that only updates when necessary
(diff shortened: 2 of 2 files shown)


z9gccs4zkw-ship-it:
Much Thanks! will try on my machine once pushed. for reference my system has 4.7TB of images (7,7M)

WithoutPants:
I've done a bit of testing on a system with 32M images. 

When getting everything at once (legacy behaviour), the full query for a single image sorted by path took 29s for the aggregate query (count, filesize, megapixels), followed by 42s to get the image ids, for a total wait of 71s.

When splitting it into two separate parallel queries, the query for the filesize and megapixels took 24s, then the query for the content took 5s for the count, and 42s to get the image ids. This means a total wait of 47s for the items to return, and 24s for the other info to populate. We could possibly shave some further time off that if we took out the count from the item query.

Performance for larger systems is still nowhere near acceptable, but this should improve things a little. Further work needs to be done at the schema level to determine what indexes are worth adding and what we can do at the sorting level to improve performance.

Related #6378

sleetx:
Is there a justification for calculating "total megapixels" at all? I find that metric to be completely irrelevant. Megapixels is useful when reviewing the properties of an individual image, not a whole collection.

Maybe "total megapixels" is best placed on the Stats page. But I don't think it's worth calculating every time you open the Images tab, especially given how much it taxes processing time.

WithoutPants:
> Is there a justification for calculating "total megapixels" at all? I find that metric to be completely irrelevant. Megapixels is useful when reviewing the properties of an individual image, not a whole collection.

Related feature request and PR: https://github.com/stashapp/stash/issues/1775

In my testing, adding megapixels added around 25% to the overhead when calculating total file size (which I think does belong). I am fairly ambivalent as to whether to keep it around - I think the utility is marginal at best.

Issue #3304, opened 2022-12-23
Error during scanning with a specific setup


HereIam80:
**Describe the bug**
There were multiple reports where during a scan the following error message appeared and no files were added:
`Error: exec: "ffprobe.exe": cannot run executable found relative to current directory`

Every time it was a new installation on windows and the users were just getting started.
Also ffprobe and ffmpeg were in the same folder as the stash-win.exe.
Stash was not located in it's default patch but on another drive, which was the same drive as the library (but a different folder).

The issue got resolved by either moving stash to it's default location or using the 0.17.0 version with the exact same setup so I believe this bug got introduced after that.

**To Reproduce**
I tried to recreate this issue to no avail.

**Stash Version: (from Settings -> About):**
0.18

**Additional context**
Here are the links to the three times I saw it reported and what got tried to resolve it:
https://discord.com/channels/559159668438728723/559159931505213440/1051590694940192849
https://discord.com/channels/559159668438728723/559159910550732809/1054405774274740234
https://discord.com/channels/559159668438728723/559159931505213440/1055782078077227088

iljarogoff:
For me this only happened when having a new install with .18
I reinstalled with .17 and then scanning worked. I then upgraded existing installation with latest dev build and now scanning is working. It seems it only happens with a fresh install with .18

Flashy78:
> 

Make sure you're installed in the default directory in Windows, so c:\Users\MYUSER\.stash\.
Or you could try what the person directly above you did to solve the issue.

datadragon61:
> > 
> 
> Make sure you're installed in the default directory in Windows, so c:\Users\MYUSER.stash. Or you could try what the person directly above you did to solve the issue.

I have mentioned that I am using default paths and settings.
I don't know how he reinstalled with .17, I can only see the option to download the latest version which is .18 and even if installing .17 solves it it is still a bug.

I have opened the bug report to tell the developer about a bug or that's an intended behavior that users have to download it, try to setup, get this bug and then waste time to go look for version .17? 😆

Flashy78:
@datadragon61 ok, sorry for trying to answer your question "What to do?".
If you want to draw attention to the original bug report submitted by HereIam80, add a thumbs up to it so it gets filtered up as an issue that multiple people are facing. Many many people have no problem running Stash on Windows, so it seems to be a unique bug that only a few people are encountering.
If you have any additional information beyond what the original but report described, please do contribute it to help with debugging.

bnkai:
[...]

PR #6373, opened 2025-12-03
Don't convert config file location to absolute during setup

### LINKED ISSUE #3029: [Bug Report] Paths should be relative to config file, NOT current working directory of the executable
Currently, relative paths (for `generated/`, `plugins/`, and `stash-go.sqlite`) are relative to the current working dir (cwd) of the executable. *This serves no useful purpose*. It also has the potential to cause unexpected behavior and even data loss.

I currently think that the simplest acceptable solution is to make relative paths relative to the config file.

I plan to work on a patch for this which is why I'm skipping a more detailed description and reproducer. But I still wanted to open an issue in case somebody else is thinking about this issue and wants to share an opinion or suggestion.

### LINKED ISSUE #6287: Make Scrapers and Plugins path relative / add to installer

Based on https://github.com/stashapp/stash/discussions/6285#discussioncomment-14993799

Please make the paths for Scrapers and Plugins relative to the app location by default, just like Cache and Generated Path. Right now they hardlink to the location of your installation without the user knowing.

They are not part of the initial installer. Maybe add them there.

No invisible dependencies.

Change location in the settings after installation.

### Additional context

_No response_

WithoutPants:
Converting the config file path to absolute during setup was originally done for #3304. The ffmpeg code has been redone since and this is no longer necessary. It was resulting in the scraper and plugin paths being made absolute during setup, despite all the others being relative to the provided config path.

Fixes #6287
Closes https://github.com/stashapp/stash/issues/3029

### DIFF
--- a/internal/manager/manager.go
+++ b/internal/manager/manager.go
@@ -219,8 +219,11 @@ func (s *Manager) Setup(ctx context.Context, input SetupInput) error {
 		// paths since they must not be relative. The config file property is
 		// resolved to an absolute path when stash is run normally, so convert
 		// relative paths to absolute paths during setup.
-		configFile, _ := filepath.Abs(input.ConfigLocation)
-
+		// #6287 - this should no longer be necessary since the ffmpeg code
+		// converts to absolute paths. Converting the config location to
+		// absolute means that scraper and plugin paths default to absolute
+		// which we don't want.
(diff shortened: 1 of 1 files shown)


feederbox826:
closes https://github.com/stashapp/stash/issues/6287

PR #6376, opened 2025-12-04
Respect user preference for type-to-create in image/scene multi-select form

ghost:
When selecting multiple images, you were previously unable to create new galleries from the edit images form itself. This PR adds the functionality, and corresponding disable behavior.

| Dropdown | Configuration |
|:---:|:---:|
| Create Gallery from Dropdown | Settings Option |

To be transparent, I am new to much of the tech stack in this repository, so I implemented this PR with some help from **Claude Opus 4.5**. If anything seems off, I'm happy to investigate and learn why. I would like to help enhance image-related features in the future, as it's my primary use case.

I get some linter errors when I validate, but I tested `develop` `HEAD` and get the same thing, so I'm ignoring.
image

### DIFF
--- a/ui/v2.5/src/components/Shared/MultiSet.tsx
+++ b/ui/v2.5/src/components/Shared/MultiSet.tsx
@@ -90,7 +90,7 @@ const Select: React.FC<IMultiSetProps> = (props) => {
         />
       );
     default:
-      return(
+      return (
         <FilterSelect
           type={type}
           isDisabled={disabled}
@@ -100,7 +100,7 @@ const Select: React.FC<IMultiSetProps> = (props) => {
           ids={props.ids ?? []}
           menuPortalTarget={props.menuPortalTarget}
         />
-      )
+      );
   }
 };
 


ghost:
I slightly misunderstood the codegen and was adding a redundant resolver, but I've cleaned it up now.

ghost:
I'm not sure how I didn't catch this before, but it's pretty no-brainer to enable this for performers and tags at the same time. I'll update this PR with that soon

ghost:
I noticed the other fields in scene/image multi-select don't respect the type-to-create config setting, and traced it back to `Select.tsx` explicitly setting the `createable` prop to false. I went ahead and removed this default prop, so it uses whatever the user has in their config for the behavior for all of these settings.

I don't think this interferes with anything else in the codebase, but let me know.

image

ghost:
This should be ready for review now.

1. Adds boilerplate to allow "Galleries" dropdown to be creatable
2. Removes forcing dropdowns in `Select.tsx` to **_not_** be creatable

feederbox826:
> I get some linter errors when I validate, but I tested develop HEAD and get the same thing, so I'm ignoring.

`pnpm install`

ui/v2.5/src/components/Shared/Select.tsx
```diff
@@ -377,17 +377,17 @@ export const TagSelect: React.FC<
...
     case "groups":
-      return <GroupSelect {...props} creatable={false} />;
+      return <GroupSelect {...props} />;
     case "galleries":
       return <GallerySelect {...props} />;
     default:
-      return <TagSelect {...props} creatable={false} />;
+      return <TagSelect {...props} />;
   }
 };
```
WithoutPants, line 392: `FilterSelect` is primarily used for the select controls in the filter dialogs - for which we don't really want to have new objects being creatable in - I'd lift this code out and put it into `MultiSet` instead of changing `FilterSelect`

ghost: Will do, I'll take a look at this sometime soon when I free up.

Should I add back `creatable={false}` and include it for galleries as well? Galleries do not currently have the prop (because they weren't creatable until this PR)

WithoutPants: > Will do, I'll take a look at this sometime soon when I free up.
> 
> Should I add back `creatable={false}` and include it for galleries as well? Galleries do not currently have the prop (because they weren't creatable until this PR)

Yes please.

ghost:
I decided to retain the base case, but just added a switch with the existing selection components to return versions where the props would populate creatable from the configuration. Thoughts?

PR #6377, opened 2025-12-04
Correct sidebar styling on details pages

### LINKED ISSUE #6375: Sidebar overlap and scrollbar issues on secondary list pages
### Describe the bug

Sidebar in the secondary list pages is taking visible space when extended or collapsed and creates a scrollbar and some visible anomalies (screenshots 1 and 2).
Sidebar also overlaps wih the sticky header (screenshot 3).

### Steps to reproduce

1. Open any secondary list page that has 0 scenes (Groups, Performers, Studios, Tags).
2. See the scrollbar when there shouldn't be one.

### Expected behaviour

No scrollbar until the sidebar is extended. 

Image

Image

Image

### Stash version

v0.29.3-93-g877491e6 (877491e6)

### Device details

Floorp 12.7.0@145.0.1 / Windows 10

### Relevant log output

```shell

```

WithoutPants:
Attempt at resolving #6375 

Appears to fix some of the styling issues on sub-pages with sidebar. It is still not stickying correctly on `xs` viewports. Input welcomed on how to fix this.

I'll take a further look tomorrow.

Fixes #6375

### DIFF
--- a/ui/v2.5/src/components/Shared/styles.scss
+++ b/ui/v2.5/src/components/Shared/styles.scss
@@ -929,7 +929,6 @@ $sticky-header-height: calc(50px + 3.3rem);
     align-self: flex-start;
 
     // take a further 15px padding to match the detail body
-    height: calc(100vh - $sticky-header-height - 15px);
     margin-top: -15px;
     max-height: calc(100vh - $sticky-header-height - 15px);
     overflow-y: auto;
@@ -941,6 +940,10 @@ $sticky-header-height: calc(50px + 3.3rem);
     .sidebar-toolbar {
       padding-top: 15px;
     }
+  }
+
+  .sidebar-pane:not(.hide-sidebar) .sidebar {
+    height: calc(100vh - $sticky-header-height - 15px);
   }
 
   .sidebar-pane.hide-sidebar .sidebar {


cj12312021:
I just took a look. I assume the issue you were referring to is the one I saw, which disrupts the sidebar's top position when scrolling to the very bottom of a sub-page. If so, modifying the following rule:
```
@media (max-width: 767.98px) {
    .detail-body .sidebar-pane:not(.hide-sidebar) .sidebar {
        margin-right: -250px;
    }
}
```
to:
```
@media (max-width: 767.98px) {
    .detail-body .sidebar-pane:not(.hide-sidebar) .sidebar {
        margin: 0 -250px 0 0;
    }
}
```
should fix it.

Removing the `margin-bottom: 48.75px;` property from `@media (max-width: 575.98px)  .sidebar-pane .sidebar` should also fix this, although that may be needed somewhere else.

WithoutPants:
Thanks for looking into hit. Here's some screenshots of what I mean:

This is what it looks like above `xs` viewport (correct):

image

As soon as the viewport drops below this breakpoint and the navbar hits the bottom, you get this:

image

Note the search input going off-screen and the border not hitting the bottom of the screen. This only happens when there isn't enough content in the content pane to scroll. If there are more than two scenes it appears correctly, but under that it is broken.

Changing `margin-right` into `margin: 0 0 -$sidebar-width 0;` makes it so that the sidebar no longer pops over the content, which we don't want in that viewport configuration. It also seems to interact weirdly with content above it:

image

However, overriding the `margin-bottom` looks to have fixed it. Thanks!

cj12312021:
I'm glad the latter solution worked, but note that the first solution effectively just overrode the `margin-bottom` rule that the first solution proposed removing. That first solution from my testing produced the same result. I'm unsure why it impacted the scene list real estate in your image. Perhaps some additional CSS was introduced that I didn't see when I pulled the branch earlier.
Screenshot 2025-12-04 154229

WithoutPants:
> I'm glad the latter solution worked, but note that the first solution effectively just overrode the `margin-bottom` rule that the first solution proposed removing. That first solution from my testing produced the same result. I'm unsure why it impacted the scene list real estate in your image.

I think it was my dumb mistake. In my comment above, you can see that I ordered the `margin` clause wrong, so that would explain it.

cj12312021:
Ah, that explains it.

Issue #2272, opened 2022-01-31
[Bug Report] Get Request loop bug?


T-prog3:
**Describe the bug**
I don't really know if this is a bug or an intended behavior.
I have this one performer created that have a gallery with around 8083 images in it.  And i noticed that if i leave the first page of the gallery by clicking next quick before everything is loaded i have to wait on the previous being done with all it's GET request before the current is being processed.

I stumbled upon this by accidentally when clicking on `perPage=1000` and then on the last button. This made me have to wait on the previous 1000 thumbnail GET Request before my current 83 images. 

Shouldn't the server stop processing when the user leave the page?

kermieisinthehouse:
So, there's currently no way to tell the server "the user left, stop processing". It just does everything it's told to do, in order. It shouldn't generate network traffic if you've navigated away though, that might be something to investigate

feederbox826:
Navigating away "cancels" the XHR request but we can't un-send the request, nor can we follow up or inform that the request has been cancelled. we just aren't around to receive the response. Unfortunately not technologically possible in the browser

PR #6378, opened 2025-12-05
SQL performance improvements

WithoutPants:
I fell down a bit of a rabbit hole while exploring for performance improvements for #6370.

I made the following improvements:
- changed `queryStruct` to use `tx.Get` instead of queryFunc
- fixed a latent bug in the table joining code that would add the join's `args` to the args list even if the join was not added
- omitted joining tables in the get aggregate query that were necessary only for ordering. Minor increase in performance when getting object counts, total file sizes, etc.
- made all calls to the sqlite interface pass the context

The last item was the most important one. This fixes what I believe is a pretty nasty latent bug, where queries would not be canceled when the applicable http request was canceled. 

An an example, on a database with 32m images - where each image request takes around a minute - I flicked through pages, changed the sort order and direction, and the page sizes all before any results came back. I had to force kill the process because it was cooking my server :grimacing:

A nice side-effect of this is that you can now kill the stash instance while a query is running if you cancel the request via navigating to another page in the UI (assuming the request came from the UI). Previously it would hang until the queries were complete - unless you `kill -9` it.

### DIFF
--- a/pkg/sqlite/repository.go
+++ b/pkg/sqlite/repository.go
@@ -96,7 +96,7 @@ func (r *repository) runIdsQuery(ctx context.Context, query string, args []inter
 }
 
 func (r *repository) queryFunc(ctx context.Context, query string, args []interface{}, single bool, f func(rows *sqlx.Rows) error) error {
-	rows, err := dbWrapper.Queryx(ctx, query, args...)
+	rows, err := dbWrapper.QueryxContext(ctx, query, args...)
 
 	if err != nil && !errors.Is(err, sql.ErrNoRows) {
 		return err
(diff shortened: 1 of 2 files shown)


smith113-p:
> The last item was the most important one. This fixes what I believe is a pretty nasty latent bug, where queries would not be canceled when the applicable http request was canceled.

I wonder if this is what #2272 was about

WithoutPants:
> I wonder if this is what #2272 was about

I think almost certainly.

discourse-stashapp:

https://discourse.stashapp.cc/t/stash-v0-30-release-development-thread/4484/1

PR #6396, opened 2025-12-11
Revert "Feature Request: Sort All Urls Alphabetically"

WithoutPants:
Reverts stashapp/stash#6352

There are users that sort URLs in a particular order. Ordering everything automatically causes a regression for those users.


DogmaDragon:
Well, there goes my most anticipated feature of v0.30. 😭

PR #6398, opened 2025-12-11
Optimize Tag List Page Performance

cj12312021:
Summary:
- Reduce tag list page query time by ~11x (tested on my production database) by eliminating expensive recursive count queries that weren't displayed on the list page
- Created optimized TagListData GraphQL fragment that excludes *_count_all fields
- Tag list page now loads in ~18ms instead of ~205ms (for 20 tags)

Problem:
The tag list page was slow to load because for every tag displayed, the TagData fragment requested 14 count fields:

- 7 direct counts (scene_count, image_count, etc.)
- 7 recursive counts (scene_count_all, image_count_all, etc. with depth: -1)

The recursive *_count_all fields use expensive recursive common table expressions to traverse the tag hierarchy and count all descendants. With 9M+ rows in my scenes_tags, this was extremely slow.
The *_count_all fields were fetched but never displayed on the list page - they're only used on the tag detail page.

Solution:
Created a new TagListData fragment and FindTagsForList query that only fetches the direct counts needed for the list page display.

### DIFF
--- a/ui/v2.5/graphql/data/tag.graphql
+++ b/ui/v2.5/graphql/data/tag.graphql
@@ -51,3 +51,31 @@ fragment SelectTagData on Tag {
     sort_name
   }
 }
+
+# Optimized fragment for tag list page - excludes expensive recursive *_count_all fields
+fragment TagListData on Tag {
+  id
+  name
+  sort_name
+  description
+  aliases
+  ignore_auto_tag
+  favorite
+  image_path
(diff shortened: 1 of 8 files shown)


cj12312021:
Did a bit more testing with this change on my DB. The result is so much better. I've moved the PR out of Draft.

WithoutPants (approved):

PR #6401, opened 2025-12-11
Implement stash_ids_endpoint for the SceneFilterType

### LINKED ISSUE #6400: Filter Scenes by Stash Ids

There is the GraphQL endpoint `findScenes`, which allows to query scenes by a number of settings.
```graphql
  "A function which queries Scene objects"
  findScenes(
    scene_filter: SceneFilterType
    scene_ids: [Int!] @deprecated(reason: "use ids")
    ids: [ID!]
    filter: FindFilterType
  ): FindScenesResultType!
```

Using `scene_filter.stash_id_endpoint`, it is possible to query a scene by its Stash box id, like such:
```graphql
query FindSceneById(
    $stashId: String
) {
    findScenes(
        scene_filter: {
            stash_id_endpoint: {
                stash_id: $stashId
                modifier: EQUALS
            }
        }
    ) {
        count
        scenes {
            ...SceneFragment
        }
    }
}
```

Unfortunately, the `StashIDCriterionInput` only allows to query for a single scene using a single id:
```graphql
input StashIDCriterionInput {
  """
  If present, this value is treated as a predicate.
  That is, it will filter based on stash_ids with the matching endpoint
  """
  endpoint: String
  stash_id: String
  modifier: CriterionModifier!
}
```

However, I would like to map a list of Stash box ids to Stash app scenes.

Therefore, I am proposing to add a `stash_ids_endpoint` to the `SceneFilterType`, as follows:
```graphql
input StashIDsCriterionInput {
  """
  If present, this value is treated as a predicate.
  That is, it will filter based on stash_ids with the matching endpoint
  """
  endpoint: String
  stash_ids: [String]
  modifier: CriterionModifier!
}
```

The resolver will attempt to resolve each id contained in `stash_ids` using the given `modifier` and `endpoint`.

What are your thoughts? Is this the way to implement this, or do you have a better idea?

The endpoint would allow to query a list of scenes by their stash ids with a single call, instead of having to make multiple calls to get all scenes by their Stash id.

This seems to be already possible by chaining many `SceneFilterType` via their `OR` field. Unfortunately, this creates deep nesting and is prohibitively expensive for the server, which will throw an exception when the nesting gets too deep.

### Additional context

_No response_

sashapp:
As mentioned in Issue #6400, I propose this change to allow to query an entire list of stash ids against the scenes in the database.

I have applied it to scenes, performers, and studios.

Resolves #6400

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -84,10 +84,20 @@ input PHashDuplicationCriterionInput {
 input StashIDCriterionInput {
   """
   If present, this value is treated as a predicate.
-  That is, it will filter based on stash_ids with the matching endpoint
+  That is, it will filter based on stash_id with the matching endpoint
   """
   endpoint: String
   stash_id: String
+  modifier: CriterionModifier!
+}
+
+input StashIDsCriterionInput {
+  """
+  If present, this value is treated as a predicate.
+  That is, it will filter based on stash_ids with the matching endpoint
+  """
+  endpoint: String
+  stash_ids: [String]
   modifier: CriterionModifier!
 }
 
@@ -156,6 +166,8 @@ input PerformerFilterType {
   o_counter: IntCriterionInput
   "Filter by StashID"
   stash_id_endpoint: StashIDCriterionInput
+  "Filter by StashIDs"
+  stash_ids_endpoint: StashIDsCriterionInput
   # rating expressed as 1-100
   rating100: IntCriterionInput
   "Filter by url"
@@ -292,6 +304,8 @@ input SceneFilterType {
   performer_count: IntCriterionInput
   "Filter by StashID"
   stash_id_endpoint: StashIDCriterionInput
+  "Filter by StashIDs"
+  stash_ids_endpoint: StashIDsCriterionInput
   "Filter by url"
   url: StringCriterionInput
   "Filter by interactive"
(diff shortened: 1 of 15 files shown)


WithoutPants (changes requested):
Please modify `sqlite.stashIDCriterionHandler` to use `stashIDsCriterionHandler` rather than (mostly) duplicating the logic.

I think the existing `stash_id_endpoint` fields should be marked as deprecated in the graphql interface.

sashapp:
Just to make sure that I understand you correctly. Do you mean I should modify `sqlite.stashIDCriterionHandler` to use `StashIDsCriterionInput` (you wrote handler in the second type name), such that no logic needs to be duplicated?

I could then just use the new handler, mark the old GraphQL endpoint as deprecated and redirect it to the new handler, wrapping its single id in a list?

WithoutPants:
> Just to make sure that I understand you correctly. Do you mean I should modify `sqlite.stashIDCriterionHandler` to use `StashIDsCriterionInput` (you wrote handler in the second type name), such that no logic needs to be duplicated?
> 
> I could then just use the new handler, mark the old GraphQL endpoint as deprecated and redirect it to the new handler, wrapping its single id in a list?

I think so. Change `stashIDCriterionHandler.handle` to convert the value of `c` into `StashIDsCriterionInput`, initialise a `stashIDsCriterionHandler` instance and call its `handle` method.

sashapp:
I have made the changes. Can you take another look?

WithoutPants (approved):

PR #6407, opened 2025-12-14
feat(dlna): add activity tracking for DLNA playback

### LINKED ISSUE #5710: [Feature] Add/edit scene play duration manually
Currently you can add play count/O count manually in history tab, but it's not possible to do the same with play duration, it's only possible to reset it. If you watch scene using DLNA or external video player like MPV, play duration is not counted.

It would be nice to manually edit play duration and save it or have a button like "+" to add full duration of a scene to total scene play duration. Another solution would be, if you manually add play count also add duration of that scene to play duration.

cj12312021:
## Summary
Adds time-based activity tracking for scenes played via DLNA, enabling play count, play duration, and resume time tracking similar to the web frontend.

## Changes
- New `ActivityTracker` in `internal/dlna/activity.go`
- Uses existing `trackActivity` UI setting (no new config needed)
- Time-based tracking approach (elapsed session time / video duration)
- 5-minute session timeout to handle aggressive client buffering

## Limitations
This is a heuristic-based approach with known limitations:
- Cannot detect actual playback position (only elapsed time)
- Cannot detect seeking or pause state
- Works best for linear playback

## Testing
- Unit tests added in `activity_test.go`
- Manually tested with Pigasus VR Media Player

Could close https://github.com/stashapp/stash/issues/5710,

### DIFF
--- a/internal/dlna/activity.go
+++ b/internal/dlna/activity.go
@@ -13,9 +13,9 @@ import (
 const (
 	// DefaultSessionTimeout is the time after which a session is considered complete
 	// if no new requests are received.
-	// This is set high (60 minutes) because DLNA clients buffer aggressively and may not
+	// This is set high (5 minutes) because DLNA clients buffer aggressively and may not
 	// send any HTTP requests for extended periods while the user is still watching.
-	DefaultSessionTimeout = 60 * time.Minute
+	DefaultSessionTimeout = 5 * time.Minute
 
 	// monitorInterval is how often we check for expired sessions.
 	monitorInterval = 10 * time.Second
@@ -176,8 +176,6 @@ func (t *ActivityTracker) Stop() {
 	t.mutex.Unlock()
 
 	for _, session := range sessions {
-		// update the last activity time to the current time
-		session.LastActivity = time.Now()
 		t.processCompletedSession(session)
 	}
 }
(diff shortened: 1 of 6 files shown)


cj12312021:
Also fixes the previously closed issue https://github.com/stashapp/stash/issues/3681. The solution in the PR does, in fact, infer playback activity by tracking the HTTP request. Without implementing a more involved AVTransport Service, this is likely the most optimal solution. The biggest limitation here, as mentioned earlier, is that the resume index won’t be very accurate if the user jumps between time indexes, but most DLNA services don’t make seeking forward or backward very convenient to begin with. I believe a similar implementation is used for Plex.

feederbox826:
My rationale for closing 3681 was that to do actual accurate tracking, you would have to reverse-engineer the HTTP range headers into the video block chunks and convert that into timestamps

The PR **does** do DLNA tracking but also makes a lot of assumptions as to watch activity and behaviour of the person - watching from the beginning to the end of a file linearly and continuously

cj12312021:
No shade was intended by my earlier message. I discovered the closed issue yesterday after taking another pass to look for the issue requesting this feature, which I was sure existed.

Based on my research, fine-grained support, like what we have on the front end, is possible. It would introduce a lot of complexity, which likely would not be worth the time investment. The level of support provided here provides the most bang for the buck and seems to be the approach favored by similar apps. I think most who turn activity tracking on would be happy with some level of DLNA support rather than no support.

Speaking personally now, my viewing habits when using DLNA do tend to be more linear.

PR #6425, opened 2025-12-18
Fix handy not working correctly

WithoutPants:
Related to upstream bug: https://github.com/defucilis/thehandy/issues/3

Fixes error message when trying to play a video with Handy:
```
Error: Need to setup the Handy with a script before calling HSSP Play!
```

Works around this by calling `getStatus()` to set the status correctly.


discourse-stashapp:

https://discourse.stashapp.cc/t/release-v0-30-1-stashapp-stash/4831/1

PR #6437, opened 2025-12-19
FR: Add Interfaces to Destroy File Database Entries

### LINKED ISSUE #6355: Add interfaces to destroy file database entries

There is currently no graphql-supported way to destroy file database entries. Historically, this has been because it was internally managed by the scan and clean subsystems. There are cases however where it is necessary to be able to destroy file entries. For example, when deleting a scene, we can set a flag to delete the file or not. When the file is not selected to be deleted, the file entry remains in the database, and a scene is created from the file on the next scan. Deleting the file removes the entry in the database, but it also deletes the file on the file system.

Proposal is to add the following interfaces to the graphql schema:
```
mutation {
  "destroy file entries in the database without deleting files from the filesystem. Files must not be assigned to objects."
  destroyFiles(ids: [ID!]!): Boolean
}
```

The `XDestroyInput` inputs should also be updated to add the following fields:
```
"If true, destroy the associated file entry from the database if the file is not assigned to any other objects."
destroy_file: Boolean
```

It is expected that these interfaces will only be used in the graphql playground for data correction purposes. I don't anticipate a need to include them in the UI, unless it is via an advanced user interface.

The necessity for this is mostly for data correction for circumstances where file entries are corrupted on some way. An example of this is #6326 where corrupt file entries will be created for files that have been renamed with a different case on filesystems with case-insensitivity. 

Manually deleting the file entries in the database via the `execSQL` interface or via an sqlite tool.

### Additional context

_No response_

Gykes:
This PR allows users to delete database duplicate entries via the playground. I was able to add a dummy secondary data to some images that were located in my database and then delete them accordingly. 

For WP: I changed up the verbage of the variables a little bit as I felt like "DestroyFiles" could be misinterpreted by users. I kept the logic the same but modified it to "DestroyFileEntries"

This example deletes an entry with the ID of 10:

```  
mutation {                                                                                                         
    deleteFileEntries(ids: ["10"])                                                                         
  }  
```

This is an example of removing a primary entry. It should error out and let the user know that it is a primary entry:

```
{
  "errors": [
    {
      "message": "cannot destroy primary file entry /Users/gykes/Desktop/blah.mkv",
      "path": [
        "deleteFileEntries"
      ]
    }
  ],
  "data": null
}
```

Also added `destroy_file_entry` option to SceneDestroyInput, ScenesDestroyInput, ImageDestroyInput, ImagesDestroyInput, and GalleryDestroyInput. This allows deleting the file DB entry (without removing the actual file) when destroying content.   

fixes: https://github.com/stashapp/stash/issues/6355

### DIFF
--- a/pkg/image/delete.go
+++ b/pkg/image/delete.go
@@ -37,8 +37,9 @@ func (d *FileDeleter) MarkGeneratedFiles(image *models.Image) error {
 }
 
 // Destroy destroys an image, optionally marking the file and generated files for deletion.
-func (s *Service) Destroy(ctx context.Context, i *models.Image, fileDeleter *FileDeleter, deleteGenerated, deleteFile bool) error {
-	return s.destroyImage(ctx, i, fileDeleter, deleteGenerated, deleteFile)
+func (s *Service) Destroy(ctx context.Context, i *models.Image, fileDeleter *FileDeleter, deleteGenerated, deleteFile bool, destroyFileEntry ...bool) error {
+	destroyEntry := len(destroyFileEntry) > 0 && destroyFileEntry[0]
+	return s.destroyImage(ctx, i, fileDeleter, deleteGenerated, deleteFile, destroyEntry)
 }
 
 // DestroyZipImages destroys all images in zip, optionally marking the files and generated files for deletion.
@@ -146,9 +147,15 @@ func (s *Service) DestroyFolderImages(ctx context.Context, folderID models.Folde
 }
 
 // Destroy destroys an image, optionally marking the file and generated files for deletion.
-func (s *Service) destroyImage(ctx context.Context, i *models.Image, fileDeleter *FileDeleter, deleteGenerated, deleteFile bool) error {
+func (s *Service) destroyImage(ctx context.Context, i *models.Image, fileDeleter *FileDeleter, deleteGenerated, deleteFile bool, destroyFileEntry ...bool) error {
+	destroyEntry := len(destroyFileEntry) > 0 && destroyFileEntry[0]
+
 	if deleteFile {
 		if err := s.deleteFiles(ctx, i, fileDeleter); err != nil {
+			return err
+		}
+	} else if destroyEntry {
+		if err := s.destroyFileEntries(ctx, i); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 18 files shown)


WithoutPants (changes requested):
> For WP: I changed up the verbage of the variables a little bit as I felt like "DestroyFiles" could be misinterpreted by users. I kept the logic the same but modified it to "DestroyFileEntries"

You've named it `deleteFileEntries` which is inconsistent with other destroy operations. I personally think `destroyFiles` is fine, assuming that it is properly documented, but `destroyFileEntries` is still better than the current naming.

Gykes:
> > For WP: I changed up the verbage of the variables a little bit as I felt like "DestroyFiles" could be misinterpreted by users. I kept the logic the same but modified it to "DestroyFileEntries"
> 
> You've named it `deleteFileEntries` which is inconsistent with other destroy operations. I personally think `destroyFiles` is fine, assuming that it is properly documented, but `destroyFileEntries` is still better than the current naming.

Yep, sorry went through like 4-5 naming various before pushing lol. I modified it up to use the `destroyFiles` naming.

pkg/image/delete.go
```diff
@@ -37,8 +37,9 @@ func (d *FileDeleter) MarkGeneratedFiles(image *models.Image) error {
 }
 
 // Destroy destroys an image, optionally marking the file and generated files for deletion.
-func (s *Service) Destroy(ctx context.Context, i *models.Image, fileDeleter *FileDeleter, deleteGenerated, deleteFile bool) error {
-	return s.destroyImage(ctx, i, fileDeleter, deleteGenerated, deleteFile)
+func (s *Service) Destroy(ctx context.Context, i *models.Image, fileDeleter *FileDeleter, deleteGenerated, deleteFile bool, destroyFileEntry ...bool) error {
```
WithoutPants: Why is it using `...bool`?

WithoutPants:
Why are the destroy methods using `...bool` parameters for the `destroyFileEntry` field?

Gykes:
It was used as a shortcut to make the parameter optional without updating existing callers. I can refactor it to use a regular bool parameter and update all call sites if you would prefer.

WithoutPants:
> It was used as a shortcut to make the parameter optional without updating existing callers. I can refactor it to use a regular bool parameter and update all call sites if you would prefer.

Yeah it doesn't feel very idiomatic to me this way. The alternative is to have an options struct, but either way its a change to the calling code, and a regular bool parameter is simplest.

WithoutPants (approved):

PR #6438, opened 2025-12-19
FR: Save & New Button on Objects

### LINKED ISSUE #5921: Add "Save & New" button in when creating new objects
Please add a "Save & New" button to Scenes, Studios, Performers...
I was surprised to see that Stash does not have this already, since it's a quite common feature ; )

Basically, if for example you add a performer, then the adding dialog has next to the "Save" button also a "Save & New" button.

Gykes:
This modifies the "Save" button into a split button and adds a "Save & New" option.

Updated: Performers, Studios, Tags, Scenes, Galleries, Groups

UI for Studios: 

Screenshot 2025-12-20 at 18-20-36 Studios Stash

Fixes: https://github.com/stashapp/stash/issues/5921

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryEditPanel.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryEditPanel.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useMemo, useState } from "react";
 import { FormattedMessage, useIntl } from "react-intl";
 import { Prompt } from "react-router-dom";
-import { Button, Form, Col, Row } from "react-bootstrap";
+import { Button, Dropdown, Form, Col, Row, SplitButton } from "react-bootstrap";
 import Mousetrap from "mousetrap";
 import * as GQL from "src/core/generated-graphql";
 import * as yup from "yup";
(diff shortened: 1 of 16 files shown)


echo6ix:
This seems like the perfect use-case for making `Save` a [split button](https://firefly.civica.com/assets/html-design-system/splitbutton.html), `Save & New` the alternative action.

> - Use a split button when there is a primary action that users are likely to perform most often, but there are also alternative actions they might occasionally choose.
> - To avoid cluttering the interface with multiple buttons, especially when the additional options are contextually related to the primary action.
> - When the secondary actions are contextually relevant to the primary action, providing logical grouping and enhancing user understanding of the available options.
> - When space is limited but you need to provide access to multiple related actions, a split button helps consolidate these actions in a compact format.

Example https://mui.com/material-ui/react-button-group/#split-button

@WithoutPants

Gykes:
> This seems like the perfect use-case for making `Save` a [split button](https://firefly.civica.com/assets/html-design-system/splitbutton.html), `Save & New` the alternative action.
> 
> > * Use a split button when there is a primary action that users are likely to perform most often, but there are also alternative actions they might occasionally choose.
> > * To avoid cluttering the interface with multiple buttons, especially when the additional options are contextually related to the primary action.
> > * When the secondary actions are contextually relevant to the primary action, providing logical grouping and enhancing user understanding of the available options.
> > * When space is limited but you need to provide access to multiple related actions, a split button helps consolidate these actions in a compact format.
> 
> Example https://mui.com/material-ui/react-button-group/#split-button
> 
> @WithoutPants

I like this. Never used it but seems like the better option. The next question will be do we want Save or Save & New to be default? I prefer Save

echo6ix:
@Gykes Definitely, `Save` would have to be the primary (default) button.

Gykes:
@echo6ix I updated the image above to reflect your recommendations. I definitely prefer your method better.

Thanks for the recommendation

ui/v2.5/src/components/Galleries/GalleryDetails/GalleryEditPanel.tsx
```diff
@@ -36,13 +36,15 @@ interface IProps {
   gallery: Partial<GQL.GalleryDataFragment>;
   isVisible: boolean;
   onSubmit: (input: GQL.GalleryCreateInput) => Promise<void>;
+  onSaveAndNew?: (input: GQL.GalleryCreateInput) => Promise<void>;
```
WithoutPants: Prefer to add `andNew` flag to `onSubmit`. That way we can condense the duplicate functions in `GalleryCreate.tsx`.

WithoutPants (changes requested):
This change is a bit more verbose than what is necessary. I've added a note to condense the gallery implementation, but this should be rolled out to all of the types.

ui/v2.5/src/components/Galleries/GalleryDetails/GalleryEditPanel.tsx
Gykes: Updated all objects

WithoutPants (approved):

PR #6444, opened 2025-12-21
FR: Add Delete Button For Scene Covers

### LINKED ISSUE #4789: Delete cover image
After an identify task, I go through and manually double check the scenes and delete the metadata for incorrectly identified ones. However I can't find a way to delete the cover image using the UI, only an option to replace it by adding image by URL or by file.

On the "edit" page/tab of a video. please include a button to delete/reset cover image to default/null. Similar to the minus button next to existing URLs, or the trash button next to stash IDs.

Gykes:
This PR adds a new clear image button next to cover image to allow users to delete the cover image of a given scene. 

It currently functions in the same way that adding a scene cover works just in reverse.

Click the new remove cover button
See notification about cover removed
save scene
profit

UI: 
Screenshot 2025-12-21 at 10 42 51

Fixes: https://github.com/stashapp/stash/issues/4789

### DIFF
--- a/internal/api/resolver_mutation_scene.go
+++ b/internal/api/resolver_mutation_scene.go
@@ -297,7 +297,8 @@ func (r *mutationResolver) sceneUpdate(ctx context.Context, input models.SceneUp
 	}
 
 	var coverImageData []byte
-	if input.CoverImage != nil {
+	coverImageProvided := input.CoverImage != nil
+	if coverImageProvided {
 		var err error
 		coverImageData, err = utils.ProcessImageInput(ctx, *input.CoverImage)
 		if err != nil {
@@ -310,21 +311,21 @@ func (r *mutationResolver) sceneUpdate(ctx context.Context, input models.SceneUp
 		return nil, err
 	}
 
-	if err := r.sceneUpdateCoverImage(ctx, scene, coverImageData); err != nil {
-		return nil, err
+	if coverImageProvided {
+		if err := r.sceneUpdateCoverImage(ctx, scene, coverImageData); err != nil {
+			return nil, err
+		}
 	}
 
 	return scene, nil
 }
 
 func (r *mutationResolver) sceneUpdateCoverImage(ctx context.Context, s *models.Scene, coverImageData []byte) error {
-	if len(coverImageData) > 0 {
-		qb := r.repository.Scene
+	qb := r.repository.Scene
 
-		// update cover table
-		if err := qb.UpdateCover(ctx, s.ID, coverImageData); err != nil {
-			return err
-		}
+	// update cover table - empty data will clear the cover
+	if err := qb.UpdateCover(ctx, s.ID, coverImageData); err != nil {
+		return err
 	}
 
 	return nil
--- a/internal/api/resolver_mutation_scene.go
+++ b/internal/api/resolver_mutation_scene.go
@@ -297,8 +297,8 @@ func (r *mutationResolver) sceneUpdate(ctx context.Context, input models.SceneUp
 	}
 
 	var coverImageData []byte
-	coverImageProvided := input.CoverImage != nil
-	if coverImageProvided {
+	coverImageIncluded := translator.hasField("cover_image")
+	if input.CoverImage != nil {
 		var err error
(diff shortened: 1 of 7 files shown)


DogmaDragon:
I'm bit hijacking this PR, but I would prefer to uniform the terminology used for the same task across different objects. As it stands right now:

Scenes
- `faTrashAlt`

Groups
- `Clear front image`
- `Clear back image`

Performers
- `Clear Image`

Studios
- `Clear Image`

Tags
- `Clear Image`

Gykes:
So you would rather it say "Clear Image" over the trash can?

If so I can take care of that tomorrow, easy fix

DogmaDragon:
Honestly not sure what is the best approach here, just wanted to bring up the inconsistency.

sleetx:
Just to be clear, is this simply unlinking the cover from the scene? or deleting/cleaning the generated cover from the database?

If it's the former, "clear image" makes sense. If it's the latter, probably the trash can or "delete cover".

Gykes:
> Just to be clear, is this simply unlinking the cover from the scene? or deleting/cleaning the generated cover from the database?
> 
> If it's the former, "clear image" makes sense. If it's the latter, probably the trash can or "delete cover".

The latter

echo6ix:
Similar to my recent split button feedback. All the actionable items related to the cover image can be consolidated into a single cover image button, and it matches users expectations from popular tools/sites like Facebook, Notion, etc.

The `Set image` button already does this, so the primary button label just requires a name change to account for the delete button.

```
[ 📷 Cover Image ] ▼
----------------------
Set image from file…
Set image from URL…
Remove image

```

Gykes:
> Similar to my recent split button feedback. All the actionable items related to the cover image can be consolidated into a single cover image button, and it matches users expectations from popular tools/sites like Facebook, Notion, etc.
> 
> The `Set image` button already does this, so the primary button label just requires a name change to account for the delete button.
> 
> ```
> [ 📷 Cover Image ] ▼
> ----------------------
> Set image from file…
> Set image from URL…
> Remove image
> ```

I think a destructive button shouldn't be in the same dropdown as a nondestructive dropdown. There's no reason per say just I feel like it's better to keep them somewhat segregated since they perform fundamentally different tasks.

I liked adding it to the Save & New as they were essentially the same. This feature is very different from what currently exists and doesn't align with the rest of the project when using the split button.

neru2132:
> I think a destructive button shouldn't be in the same dropdown as a nondestructive dropdown.

`Set image` is a destructive action if there is an existing image. For what it's worth, I have clicked on `Set image...` in the hopes that I could remove the image in the past, but I would also be okay with an extra button. What I don't like is seeing a bunch of administrative buttons when all I want to see is information about the scene, but that's more related to shortcomings of the `Details` tab.

Gykes:
> > I think a destructive button shouldn't be in the same dropdown as a nondestructive dropdown.
> 
> 
> 
> `Set image` is a destructive action if there is an existing image. For what it's worth, I have clicked on `Set image...` in the hopes that I could remove the image in the past, but I would also be okay with an extra button. What I don't like is seeing a bunch of administrative buttons when all I want to see is information about the scene, but that's more related to shortcomings of the `Details` tab.

I think it has a secondary purpose of a destructive action. Its main purpose is to set and remove/destroy if something currently exists. This feature is purely designed to destroy.

WithoutPants (changes requested):
Please look at the way that the performer image is cleared and use a similar implementation. It uses `changesetTranslator.hasField` to determine if the image is set, and uses `nil` to specify clearing the image, rather than `""`.

Clearing the image should show the default scene image rather than showing a toast notification. This will require modifying `api/sceneRoutes.Screenshot` to handle the `default` flag.

Gykes:
> Please look at the way that the performer image is cleared and use a similar implementation. It uses `changesetTranslator.hasField` to determine if the image is set, and uses `nil` to specify clearing the image, rather than `""`.
> 
> Clearing the image should show the default scene image rather than showing a toast notification. This will require modifying `api/sceneRoutes.Screenshot` to handle the `default` flag.

Yep, that way is a lot smarter. Didn't even think to check that.

WithoutPants (approved):

PR #6449, opened 2025-12-22
FR: Change Career Length to Career Start and Career End

### LINKED ISSUE #3348: [Feature] Change Career Length to 2 fields
Change Career Length from free text to 2 separate fields.  Carrer Start and End. Career Length could then be calculated.

None.  Career Length is currently a free text field

**Additional context**
Changing Career Length to 2 fields will allow for a calculation actual career length and possibly age started.

Gykes:
This is my first attempt at a pretty large front and backend change PR. Please be gentle.

Adds `career_start` and `career_end` integer fields to the Performer model, aligning with StashDB's  format. The existing `career_length` string field is now deprecated and computed from the new fields for backwards compatibility.

Closes: #3348

  Changes

  - Added `migration 76_performer_career_dates.up.sql` to add `career_start` and `career_end` integer columns to the performers table
  - Updated `appSchemaVersion` to 76

GraphQL Schema

  - Added `career_start` Int and `career_end` Int fields to:
    - `Performer type`
    - `PerformerCreateInput`
    - `PerformerUpdateInput`
    - `BulkPerformerUpdateInput`
    - `ScrapedPerformer` type
    - `ScrapedPerformerInput`
  - Deprecated `career_length` field with reason "Use career_start and career_end"

Backend

  - Added `CareerStart` and `CareerEnd` fields to:
    - `Performer model` struct
    - `PerformerPartial` struct
    -  `performerRow` SQLite struct
    -  `ScrapedPerformer` struct
  - Added `CareerLength` resolver that computes the display string from `career_start` and `career_end`, with fallback to stored `career_length` for backwards compatibility
  - Updated mutation resolvers to handle new fields in create/update operations
  - Updated `ToPerformer` method to convert scraped string values to integers

Frontend

  - Removed `career_length` input field from `PerformerEditPanel`
  - Added `career_start` and `career_end` number input fields to `PerformerEditPanel`
  - Updated `PerformerScrapeDialog` to show `career_start` and `career_end` instead of `career_length`
  - Updated Tagger `PerformerModal` to display and save `career_start` and `career_end`
  - Updated scraper GraphQL fragments to request new fields
  - Added localization strings for `career_start` and `career_end`

  StashBox Integration

  - Updated `performerFragmentToScrapedPerformer` to populate `CareerStart` and `CareerEnd` from StashDB's `CareerStartYear` and `CareerEndYear`
  - Updated `SubmitPerformerDraft` to use `CareerStart` and `CareerEnd` directly when submitting to StashBox, with fallback to parsing `CareerLength` for backwards compatibility

  Backwards Compatibility

  - Existing performers with only `career_length` populated will continue to display correctly via the resolver fallback
  - The `career_length` field remains in the GraphQL API (deprecated) and is computed from the new fields when available
  - StashBox scraping now populates the new fields directly


discourse-stashapp:

WithoutPants:
> * Added `migration 76_performer_career_dates.up.sql` to add `career_start` and `career_end` integer columns to the performers table

The migration should migrate existing `career_length` values into the new fields. This will likely require a post-migration to interpret the original fields. The post-migration should remove the existing `career_length` field. If the existing value cannot be parsed into start and end, then it should be logged and moved into a `career_length` custom attribute.

The mutation resolvers need to interpret the deprecated `career_length` fields and set the appropriate new fields.

The import/export functionality needs to be updated with the new fields _and_ handle the legacy field.

Gykes:
Honestly, I didn't even know you could do a post migration like that. I was curious how we were going to handle moving the data to the new field without having to have users retag all their stuff. Will more than likely start working on stuff tomorrow. Thanks for the info 👍

Gykes:
@WithoutPants 

I made a lot of additions/changes. Not going to lie after a certain point my eyes started to glaze over. I was able to test by using a backup DB and all the existing fields converted as expected.

pkg/models/model_scraped_item.go
```diff
@@ -659,3 +697,49 @@ type ScrapedGalleryInput struct {
 	// deprecated
 	URL *string `json:"url"`
 }
+
+// parseCareerLengthString parses a career_length string like "YYYY - YYYY" into start/end year ints.
+// This is a local copy to avoid circular imports with pkg/performer.
+func parseCareerLengthString(s string) (start *int, end *int, err error) {
```
WithoutPants: I think it might be better off consolidating the duplicated functions into `util` with a more generic name `ParseYearRangeString`. I think the `performer` package version looks like the better one to use.

pkg/models/model_scraped_item.go
```diff
@@ -659,3 +697,49 @@ type ScrapedGalleryInput struct {
...
+
+	lower := strings.ToLower(s)
+	lower = strings.ReplaceAll(lower, "present", "")
+
+	var parts []string
+	switch {
+	case strings.Contains(lower, " - "):
+		parts = strings.SplitN(lower, " - ", 2)
+	case strings.Contains(lower, "-"):
+		parts = strings.SplitN(lower, "-", 2)
```
WithoutPants: I think enforcing a space to either side of the `-` character is unnecessarily restrictive.

pkg/performer/import.go
```diff
@@ -282,5 +281,19 @@ func performerJSONToPerformer(performerJSON jsonschema.Performer) models.Perform
 		}
 	}
 
+	// prefer explicit career_start/career_end, fall back to parsing legacy career_length
+	if performerJSON.CareerStart != nil || performerJSON.CareerEnd != nil {
```
WithoutPants, line 288: The import unit test should be updated to cover both scenarios.

pkg/sqlite/migrations/78_postmigrate.go
```diff
@@ -0,0 +1,143 @@
...
+	)
+	return err
+}
+
+func (m *schema78Migrator) preserveAsCustomField(tx *sqlx.Tx, id int, value string) error {
+	// check if a career_length custom field already exists
+	var existing sql.NullString
+	err := tx.Get(&existing, "SELECT value FROM performer_custom_fields WHERE performer_id = ? AND field = 'career_length'", id)
+	if err == nil {
+		// already exists, skip
```
WithoutPants: We should log something here.

pkg/sqlite/performer_filter.go
```diff
@@ -129,7 +129,8 @@ func (qb *performerFilterHandler) criterionHandler() criterionHandler {
 			}
 		}),
 
-		stringCriterionHandler(filter.CareerLength, tableName+".career_length"),
```
WithoutPants, line 132: We should either handle the deprecated `career_length` filter criterion, or document that it is not used - here and in the graphql interface.

ui/v2.5/graphql/data/performer-slim.graphql
```diff
@@ -17,6 +17,8 @@ fragment SlimPerformerData on Performer {
   penis_length
   circumcised
   career_length
```
WithoutPants: This should no longer be necessary and can be removed.

ui/v2.5/graphql/data/performer.graphql
```diff
@@ -14,6 +14,8 @@ fragment PerformerData on Performer {
   penis_length
   circumcised
   career_length
```
WithoutPants: Remove

ui/v2.5/graphql/data/scrapers.graphql
```diff
@@ -39,6 +39,8 @@ fragment ScrapedPerformerData on ScrapedPerformer {
   penis_length
   circumcised
   career_length
```
WithoutPants: This should be able to be removed.

ui/v2.5/graphql/data/scrapers.graphql
```diff
@@ -69,6 +71,8 @@ fragment ScrapedScenePerformerData on ScrapedPerformer {
...
```
WithoutPants: Remove

ui/v2.5/src/components/Performers/PerformerDetails/PerformerDetailsPanel.tsx
```diff
@@ -173,8 +173,13 @@ export const PerformerDetailsPanel: React.FC<IPerformerDetails> =
           fullWidth={fullWidth}
         />
         <DetailItem
```
WithoutPants, line 176: Is there any benefit to separating out the career lengths in the details? I'd think it'd be more concise to use the existing format.

ui/v2.5/src/components/Performers/PerformerListTable.tsx
```diff
@@ -187,8 +187,12 @@ export const PerformerListTable: React.FC<IPerformerListTableProps> = (
     <>{FormatWeight(performer.weight)}</>
   );
 
-  const CareerLengthCell = (performer: GQL.PerformerDataFragment) => (
```
WithoutPants, line 190: Again, I think this is better displayed as career length range rather than splitting into separate fields.

WithoutPants (changes requested):
Nice work on a pretty involved change set. This is a static code review only.

ui/v2.5/src/components/Performers/PerformerDetails/PerformerDetailsPanel.tsx
Gykes: Not really. It's split out in the merge and scraper modals due to how it is received from stashbox and scrapers but will condense it down to one.

internal/api/resolver_mutation_performer.go
```diff
@@ -56,9 +57,13 @@ func (r *mutationResolver) PerformerCreate(ctx context.Context, input models.Per
 	newPerformer.CareerEnd = input.CareerEnd
 	// if career_start/career_end not provided, parse deprecated career_length
 	if newPerformer.CareerStart == nil && newPerformer.CareerEnd == nil && input.CareerLength != nil {
-		start, end, _ := performer.ParseCareerLength(*input.CareerLength)
-		newPerformer.CareerStart = start
-		newPerformer.CareerEnd = end
+		start, end, err := performer.ParseCareerLength(*input.CareerLength)
+		if err != nil {
+			logger.Warnf("could not parse deprecated career_length %q: %v", *input.CareerLength, err)
```
WithoutPants: There's no reason not to error out on a parse failure.

internal/api/resolver_mutation_performer.go
```diff
@@ -273,12 +278,16 @@ func performerPartialFromInput(input models.PerformerUpdateInput, translator cha
...
 	} else if translator.hasField("career_length") && input.CareerLength != nil {
-		start, end, _ := performer.ParseCareerLength(*input.CareerLength)
-		if start != nil {
-			updatedPerformer.CareerStart = models.NewOptionalInt(*start)
-		}
-		if end != nil {
-			updatedPerformer.CareerEnd = models.NewOptionalInt(*end)
+		start, end, err := performer.ParseCareerLength(*input.CareerLength)
+		if err != nil {
+			logger.Warnf("could not parse deprecated career_length %q: %v", *input.CareerLength, err)
```
WithoutPants: Return error

internal/api/resolver_mutation_performer.go
```diff
@@ -441,12 +450,16 @@ func (r *mutationResolver) BulkPerformerUpdate(ctx context.Context, input BulkPe
...
```
WithoutPants: Return error

pkg/performer/career.go
```diff
@@ -1,74 +1,11 @@
...
-	"fmt"
-	"strconv"
-	"strings"
+	"github.com/stashapp/stash/pkg/utils"
 )
 
 // ParseCareerLength parses a career_length string into start and end year integers.
-// Supported formats: "YYYY", "YYYY - YYYY", "YYYY-YYYY", "YYYY -", "- YYYY", "YYYY-present".
-// Returns nil for start/end if not present in the string.
+// Deprecated: Use utils.ParseYearRangeString directly.
```
WithoutPants: This is a new method - it should be removed instead of adding a new deprecated method to the code base.

pkg/performer/import.go
```diff
@@ -282,5 +281,19 @@ func performerJSONToPerformer(performerJSON jsonschema.Performer) models.Perform
...
+		newPerformer.CareerStart = performerJSON.CareerStart
+		newPerformer.CareerEnd = performerJSON.CareerEnd
+	} else if performerJSON.CareerLength != "" {
+		start, end, err := ParseCareerLength(performerJSON.CareerLength)
+		if err != nil {
+			logger.Warnf("error parsing career_length %q: %v", performerJSON.CareerLength, err)
```
WithoutPants: For an import operation, I'd expect an invalid career length to return an error and fail the import.

pkg/utils/date.go
```diff
@@ -25,3 +27,70 @@ func ParseDateStringAsTime(dateString string) (time.Time, error) {
...
+	lower := strings.ToLower(s)
+	lower = strings.ReplaceAll(lower, "present", "")
+
+	// split on " - ", "-", or " -"
+	var parts []string
+	switch {
+	case strings.Contains(lower, " - "):
+		parts = strings.SplitN(lower, " - ", 2)
+	case strings.Contains(lower, "-"):
+		parts = strings.SplitN(lower, "-", 2)
```
WithoutPants: I think this could be simplified, just split on `-` character if it contains it. You trim the space from the results further down anyway.

ui/v2.5/src/components/Performers/PerformerDetails/PerformerDetailsPanel.tsx
```diff
@@ -173,13 +173,14 @@ export const PerformerDetailsPanel: React.FC<IPerformerDetails> =
...
-        />
-        <DetailItem
-          id="career_end"
-          value={performer?.career_end?.toString()}
+          id="career_length"
+          value={
+            performer?.career_start || performer?.career_end
+              ? `${performer.career_start ?? ""} - ${
+                  performer.career_end ?? ""
+                }`
```
WithoutPants: Put this into a utility formatting function (ie `formatYearRange(start, end)`), since we're using it in a few places.

ui/v2.5/src/components/Performers/PerformerDetails/PerformerScrapeDialog.tsx
```diff
@@ -501,6 +501,9 @@ export const PerformerScrapeDialog: React.FC<IPerformerScrapeDialogProps> = (
           result={fakeTits}
           onChange={(value) => setFakeTits(value)}
         />
+        {/* career_start and career_end are kept as separate fields here
+            (unlike the combined display in the details panel) so users can
+            select the best value for each independently when scraping. */}
```
WithoutPants: This and the comment in the scrape dialog are probably not necessary. This behaviour is consistent with the edit panel/dialogs, so it's not significant that it's done here. The combined display only makes sense for the read-only fields.

WithoutPants (changes requested):

WithoutPants:
I made the following final changes:
- added basic legacy career length filter handler
- removed the dependency to `require` which wasn't previously used. `assert` has the same functionality
- added utility function to format year range and used it in the performer resolver and scraped performer post-processing
- changed the career start/end scraped performer fields to be int instead of string, and had the scrape post-processor fill in any missing career field based on the populated fields

By necessity I had to then add scraped number components in the UI.

WithoutPants (approved):

Gykes:
Got tired of correcting me, eh? Lol 

Ty for looking at it and getting it in.

echo6ix:
Thank you both for your work on changing the career length field to career start and end.

Is the range separator, ` - ` (space dash space), hardcoded or can it be over-ridden with custom locale? Say someone wanted to use an en dash with no spaces, `–`, to reflect how most style guides write date ranges.

Just bringing this up now while it's a fresh PR because you know it will come up eventually xD

WithoutPants:
> Is the range separator, `-` (space dash space), hardcoded or can it be over-ridden with custom locale? Say someone wanted to use an en dash with no spaces, `–`, to reflect how most style guides write date ranges.

It is hardcoded, but the spaces are optional and the result is trimmed.

PR #6450, opened 2025-12-22
[SFW] Show Performer Age in Performer Card when SFW turned on

bob12224:
## Description

This simple PR removes the explicit hiding of performer age from SFW.

I understand why `penis length` and `fake tits` are hidden in SFW, but `age` IS considered SFW (it is an editable field on performer pages). Thus I think it should be shown in SFW performer cards.

## User Story

* IWT: see the age of my family member in old footage.
    * PROBLEM: Must switch from SFW to NSFW to do so

### DIFF
--- a/ui/v2.5/src/sfw-mode.scss
+++ b/ui/v2.5/src/sfw-mode.scss
@@ -83,9 +83,4 @@
       display: none;
     }
   }
-
-  // hide performer age on performer cards
-  .performer-card__age {
-    display: none;
-  }
 }


bob12224:
To anyone who wants this before the PR is merged:
image

1. Turn on `Custom CSS` (Settings -> Interface)
2. `Edit` the Custom CSS
3. Paste this inside
    ````css
    .sfw-content-mode {
    .performer-card__age {
        display: block !important;
      }
    }
    ````
4. You *might* have to CTRL+SHIFT+R (refresh and ignore cache) afterwards

PR #6481, opened 2026-01-06
Align card images to center

ghost:
Currently, the magnifying glass for hovering over images is always centered over the center of the card itself, not the center of the fitted image. I tried to play around with a way to force this not to be the case using CSS only, but ran into a lot of difficulty.

I'm proposing we align these card images vertically, or do some manual layout logic to ensure that the magnifying glass always falls over the center of the image itself.

| BEFORE | AFTER |
|--------|-------|
|  Screenshot 2026-01-05 192836  | Screenshot 2026-01-05 192855 |

### DIFF
--- a/ui/v2.5/src/components/Images/styles.scss
+++ b/ui/v2.5/src/components/Images/styles.scss
@@ -86,6 +86,7 @@
   }
 
   &-preview {
+    align-items: center;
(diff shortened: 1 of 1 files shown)


WithoutPants (approved):
Looks good to me.

PR #6482, opened 2026-01-06
Plugin API - card grid components

Valkyr-JS:
Makes the following components patchable:

* `GalleryCardGrid`
* `GroupCardGrid`
* `ImageGridCard`
* `PerformerCardGrid`
* `SceneCardsGrid`
* `SceneMarkerCardsGrid`
* `StudioCardGrid`
* `TagCardGrid`

As can be seen from this list, the component naming differs slightly between each component, as do the filenames. I considered renaming them for consistency as part of this PR but thought that might be an unwanted change. I'd be happy to add the change to this PR or a separate one if given the go ahead.


WithoutPants:
Go ahead with the renaming for consistency. Like you've done above, a commit for each operation is ideal.

PR #6483, opened 2026-01-07
fix(dlna): improve activity tracking accuracy and efficiency

cj12312021:
## Summary
  - **Remove play duration tracking**: I think it is best due to how inaccurate the estimate would be for non-linear playback. Inaccurate estimates could skew the analytics that users may be running on that data.
  - **Combine database transactions**: Resume time and view count updates now happen in a single transaction instead of two separate ones. This provides atomicity (if one fails, both roll back) and better performance (one DB round-trip instead of two).

### DIFF
--- a/internal/dlna/activity.go
+++ b/internal/dlna/activity.go
@@ -82,16 +82,6 @@ func (s *streamSession) percentWatched() float64 {
 	return timeBasedPercent
 }
 
-// estimatedPlayDuration returns the estimated play duration in seconds.
-// Uses elapsed time from session start to last activity, capped by video duration.
-func (s *streamSession) estimatedPlayDuration() float64 {
-	elapsed := s.LastActivity.Sub(s.StartTime).Seconds()
-	if s.VideoDuration > 0 && elapsed > s.VideoDuration {
-		return s.VideoDuration
-	}
-	return elapsed
-}
-
 // estimatedResumeTime calculates the estimated resume time based on elapsed time.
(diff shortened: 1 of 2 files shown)


feederbox826:
Is there any point to the play duration check? Since DLNA buffers so aggressively would it be possible to not trigger it?

cj12312021:
> Is there any point to the play duration check? Since DLNA buffers so aggressively would it be possible to not trigger it?

The goal of this PR is to remove the play duration check/tracking from the DLNA service.

feederbox826:
Sorry, I meant `elapsed`

https://github.com/stashapp/stash/pull/6483/changes#diff-6cd6601a8d885f37fd5d16762252c4f038a5ee7feea29b7e64ec784f65805404R265

and the corresponding check
https://github.com/stashapp/stash/pull/6483/changes#diff-6cd6601a8d885f37fd5d16762252c4f038a5ee7feea29b7e64ec784f65805404R271

cj12312021:
The goal of that code was to prevent writes for sessions that lasted less than 5 seconds. The 5-second time should be more reliable to calculate, but I see the argument for removing it as well. I can push that change later today.

cj12312021:
@WithoutPants, whenever you get to this, feel free to leave your thoughts. I've honestly been going back and forward about this PR.

WithoutPants (approved):

WithoutPants:
I agree with the reasoning in the summary.

PR #6485, opened 2026-01-08
feat: Add .stashignore support for gitignore-style scan exclusions

### LINKED ISSUE #1139: [Feature] add ignore file to ignore a folder during scanning
I have some pictures and extras related to videos that I do not wish to be added to stash, but would like to keep inside the same folder structures as the videos.

Android, Emby and a lot of other programs check for a specific empty file before reading a folder. Emby for example checks if a .ignore file is present and will exclude folders with this file during a scan. I would like to request this functionality to be added to Stash

nandastone:
Adds support for `.stashignore` files that allow users to exclude files and directories from scanning using gitignore-style patterns. Closes #1139.

Some LLM usage as Go is not my preferred language.

## Changes

  - New `StashIgnoreFilter` in `pkg/file/stashignore.go` that checks for `.stashignore` files during scans
  - Integrated into `scanFilter` alongside existing exclusion logic. The scanner doesn't appear to expose plugin hooks for file filtering.
  - Supports nested `.stashignore` files (patterns cascade from library root to subdirectories)
  - Uses [go-gitignore](https://github.com/sabhiram/go-gitignore) library for pattern matching
  - Added user documentation in `Tasks.md`

## Usage

Place a `.stashignore` file in any directory within your library:

```
# Ignore temp files

*.tmp
temp/

# But keep this one

!important.tmp
```

## Test Plan

- 16 unit tests covering pattern syntax (wildcards, negation, nested files, etc.)
- 4 integration tests with real scanner and database
- All existing tests pass


feederbox826:
Why not extend or interface with it from the existing ignores menu?

nandastone:
> Why not extend or interface with it from the existing ignores menu?

Different mental models:

- Global excludes = "never scan X anywhere" (admin policy).  
- `.stashignore` = "in this folder, skip these" (local override). Both are valid, complementary use cases.

The `.stashignore` file lives with your content. You see what's ignored by looking at the folder. You can share it, back it up, version control it, etc.

And finally, glob syntax (as opposed to regex) is a familiar convention: `.gitignore`, `.dockerignore`, `.npmignore`, etc. The different syntax is why the PR doesn't "append" the discovered file rules to the existing regex rules system.

We also wouldn't want to try and "sync" the `.stashignore` rules to the regex rules, a two way sync would be brittle and complex.

pkg/file/stashignore.go
```diff
@@ -0,0 +1,203 @@
...
+// Accept returns true if the path should be included in the scan.
+// It checks for .stashignore files in the directory hierarchy and
+// applies gitignore-style pattern matching.
+// The libraryRoot parameter bounds the search for .stashignore files -
+// only directories within the library root are checked.
+func (f *StashIgnoreFilter) Accept(ctx context.Context, path string, info fs.FileInfo, libraryRoot string) bool {
+	// Always accept .stashignore files themselves so they can be read.
+	if filepath.Base(path) == stashIgnoreFilename {
+		return true
+	}
```
WithoutPants: Is this true? I don't think we actually need to accept the files themselves.

pkg/file/stashignore.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Try to load .stashignore from this directory.
+	stashIgnorePath := filepath.Join(dir, stashIgnoreFilename)
+	patterns, err := f.loadIgnoreFile(stashIgnorePath)
+	if err != nil || patterns == nil {
+		// Cache negative result (file doesn't exist or has no patterns).
+		f.cache.Store(dir, &ignoreEntry{patterns: nil, dir: dir})
+		return nil
+	}
```
WithoutPants, line 210: Shouldn't we log something if the file fails to be loaded?

pkg/file/stashignore.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+
+	return !ignored
+}
+
+// collectIgnoreEntries gathers all ignore entries from library root to the given directory.
+// It walks up the directory tree from dir to libraryRoot and returns entries in order
+// from root to most specific.
+func (f *StashIgnoreFilter) collectIgnoreEntries(dir string, libraryRoot string) []*ignoreEntry {
```
WithoutPants, line 99: Having to do this for every scanned file is likely to cause some non-trivial overhead. We'd perhaps get some benefit from caching the ignoreEntry list for a given directory path. It needn't be exhaustive - a LRU cache would probably do and would save calculating everything, it could be used to shortcut the calculation for sub-directories as well.

WithoutPants (changes requested):
Static code review only

pkg/file/stashignore.go
nandastone: Great point, I've implemented both optimisations:

1. LRU cache for collected entries keyed by dir + libraryRoot. This means files in the same directory will avoid re-walking the tree.
2. If parent's entries are cached, we just check if the child dir has a `.stashignore` and extend (if it has one), avoiding walking the tree.

All existing tests pass, but I didn't add tests specifically for the optimisations. Is that ok? if not, how might I approach those tests?

pkg/file/stashignore.go
nandastone: Now logging unexpected file system errors e.g. couldn't open file. "Expected" errors like "file doesn't exist" are ignored.

pkg/file/stashignore.go
nandastone: Good catch, we load directly from the FS, not via our scanner, so this exception isn't needed. Removed.

pkg/file/stashignore.go
WithoutPants: That's fine, no additional tests are needed.

WithoutPants (changes requested):
This needs to be implemented in the clean task as well so that entries can be removed post-adding an ignore file.

WithoutPants (approved):

nandastone:
Thanks @WithoutPants!

PR #6494, opened 2026-01-14
Add basename and parent_folders fields to Folder graphql interface

WithoutPants:
Preparation work for a folder filter. 

- Adds a `basename` field to the `Folder` type. This field is derived from the `path` field - more on this below.
- Adds `parent_folders` field to the `Folder` type, this provides a list of all parent folders for the applicable folder, with the first being the immediate parent and the last being the highest level folder. This means that you can get the full hierarchy of folders (and their ids) with a single query.
- Adds a schema migration to add a `basename` column. This is necessary for searching folders, since if you search for `foo` in the path, you will not only get the `/foo` directory but all of its subdirectories, since `foo` is part of the path. The `basename` column is automatically calculated from the `path` field during creates and updates.
- Adds a `basename` field to the `FolderFilterType` input, to allow filtering on `basename`.


pkg/sqlite/folder_filter_test.go
```diff
@@ -33,6 +33,17 @@ func TestFolderQuery(t *testing.T) {
 			includeIdxs: []int{folderIdxWithSubFolder, folderIdxWithParentFolder},
 			excludeIdxs: []int{folderIdxInZip},
 		},
+		{
+			name: "basename",
+			filter: &models.FolderFilterType{
+				Path: &models.StringCriterionInput{
```
Gykes: Should this be `Basename:`

WithoutPants:
Following up on #6608, the post-migration now corrects folder entries where there is no parent folder.

I've also made a change to the `zipfs` logic to treat errors encountered during the call to `filepath.Rel` as a file not found error. The error only occurs if the file entry's path cannot be made relative to the path of the parent zip file, so we should essentially treat it as not found. This was causing an issue during clean where a corrupt folder entry with a mismatched path to its zip file would not be cleaned since it emitted an error.

PR #6497, opened 2026-01-17
Image phash generation

WithoutPants:
Adds the ability to generate, view and filter image perceptual hashes. Uses the existing `goimagehash` library for generating perceptual hashes, and stores them in the `phash` fingerprint type.

- Added `Phash` field to the `File Info` tab in the image detail panel. Provides a link to show images with the same phash.
- Added `Phash` filter option, functioning the same as the scene filter criterion. 
- Added options to generate image phash to scan and generate tasks.
- Added ability to run the generate task on a list of image ids.
- Added generate menu option to image detail page
- Added generate menu option to the image list page


pkg/sqlite/criterion_handlers.go
```diff
@@ -1126,3 +1074,48 @@ func (h *relatedFilterHandler) handle(ctx context.Context, f *filterBuilder) {
...
+		distance = *phashDistance.Distance
+	}
+
+	if distance == 0 {
+		// use the default handler
+		intCriterionHandler(&models.IntCriterionInput{
+			Value:    int(value),
+			Modifier: phashDistance.Modifier,
+		}, "fingerprints_phash.fingerprint", nil)(ctx, f)
+	}
```
Gykes: Should there be a return after this block? When distance is 0 it calls `intcriterionHandler` then falls through to the switch default again, right?

DogmaDragon (changes requested):
Needs to be documented in the manual.

pkg/sqlite/criterion_handlers.go
WithoutPants: Thanks, addressed

DogmaDragon (approved):

Gykes:
> How do you view and delete duplicates?
> 
> Tools only has a Scene duplicate checker.

You can't currently. There is a PR open to add this functionality

evolite:
Hmm, i run phash generation but only 84 of my thousands of images get a phash. Trying to run the generate task again does not yield results.

2026-05-15 13:51:31Info    Generating 0 covers 0 sprites 0 previews 5 markers 0 phashes 0 image phashes

But, 
image

PR #6501, opened 2026-01-17
Add latest scene sort for performers and studios.

WeedLordVegeta420:
This adds a "Latest Scene" sort option for Studios and Performers (a similar option is already present in stash-box for Performers). This sorting is based on the max `date` value on associated Scenes - null values are effectively sorted as the earliest/lowest values.


sleetx:
Am I missing something? "Date" already sorts content by date, including on Performer and Studio pages.

WeedLordVegeta420:
@sleetx This adds an option when you're sorting the list of Performers or Studios (basically sorting the list of performers/studios based on when their most recent scene released). I think you're referring to the existing option of sorting the list of Scenes on an individual Studio/Performer page.

WithoutPants (approved):

Issue #12, opened 2019-02-11
[Feature] Account system


StashAppDev:
https://github.com/stashapp/StashServer/issues/43

StashAppDev:
User access rights https://github.com/stashapp/StashServer/issues/44

Issue #2337, opened 2022-02-21
[Feature] Multiple Users with Permissions


richardfoltin:
**Is your feature request related to a problem? Please describe.**
I would like to have my videos at a safe place. So my idea is that I would create 2 accounts for myself; an admin and a user. User account credentials are saved in the browser thus it is easy to access the videos. Admin account is only used if organinizing is needed.

**Describe the solution you'd like**
I would like to have the possiblity to have multiple accouns and to configure the permissions that an account can have. Especially in case of important features like Deleting videos.

QxxxGit:
I think multiple users with a permissions system could be pretty cool for Stash.

You could possibly see a wide array of uses if it's within the scope of the devs vision for Stash. Amateur/professionals could host their own install, allowing subscribers under a user role to see specific scenes/images/galleries. You would probably see some tube sites start using Stash to host content. Even on a local network, user permissions would be useful. My initial thoughts:

* Multiple users would necessitate an additional table for storing user data (name, pass, etc.). This table would evolve for feature requests (email validation and what have you).
* Permissions would range from viewing/creating/editing records (scenes, performers, etc.), managing users, and so on.
* Permissions could be set for individual media, collections/galleries, and for user roles. 

Of course something of this depth would be much further down the road; but I think it's worth getting ideas written out in advance.

kermieisinthehouse:
> You would probably see some tube sites start using Stash to host content. 

I think that, partially, using sqlite, and primarily, reading directly from block storage means that this will never scale past 25 users. It would be a huge rewrite to make stash "scalable"

RRocks01:
I came here to see if this was possible, I just wanted to create permissions for my wife, and ideally let her have her own "like" count, and more importantly keep mine private and library settings.  Would be nice to restrict content too.

WithoutPants:
This is something I'd like to add at some point. As others have pointed out, it will require a substantial amount of effort.

sleetx:
A multi-user environment would be great. We'll need a place in the settings where accounts can be managed. Plus a minimum of two assignable roles: admin and user. Admin is what we have now. Here are some requirements for a limited "user" role that is mostly read-only:

Access restrictions:
- No access to the "Settings" page (this includes browsing to the /settings URL manually)
- No access to the "Edit" tab on any page. (The "rating" option is here on some of them, and we probably want to let users rate things on their own. So ratings should be moved to another tab.)
- No access to the Tagger view or "Auto Tag" buttons
[...]

PR #6503, opened 2026-01-17
[WIP/PoC] Multi-user support (backend only)

WithoutPants:
This is a WIP implementation of multi-user support (#12). 

Currently stores user information in the config file. I expect to change this to store user information in the database before this can be considered ready to merge.

Adds fairly coarse-grained roles: `READ`, `MODIFY` and `ADMIN`. 

Requires that the first user created have the `ADMIN` role. Backwards compatible with the existing username/password config fields - which assume an `ADMIN` role.

Tested briefly by creating some users and logging into a `READ` user. Attempts to modify objects currently gives an `unauthorized` error, which should be improved to give a more verbose error.

Expected next steps:
- determine final role types
- add a schema migration and create a user store implementation
- ~~add user-specific api-keys~~

Potential future work:
- UI support
- Plugin permissions
- Adapt UI for various role types
- User-specific object ratings and favourites
- User-specific preferences
- Content permissions


Gykes:
Commenting to keep track so I can do frontend work

sleetx:
Feature request / discussion: #2337

WithoutPants:
I think that we'll need to include migrating existing ratings, favourites, saved filters and o/watch histories as part of the user migration. Doing this in a later migration after users have already been created will complicate the process significantly. Since this is a significant change in and of itself, I'll do this in a separate branch to merge into this one.

PR #6514, opened 2026-01-21
feat: auto-remove duplicate aliases

### LINKED ISSUE #5891: [Feature] Auto-remove duplicate aliases
Sometimes I find myself trying to save a performer where, for one reason or another, there is a duplicate alias. Stash prevents me from doing so with a helpful error message, which is easy to address, but is mildly annoying.

It would be helpful if, instead of an error message, stash simply removed the duplicate alias and continued with the save since that's all the user is going to do anyway. I don't believe this would have any negative side effects, because:
- No data is lost, as only a duplicate is removed
- The solution to the error message is to remove the duplicate, and this is just automating the process

Stash could present a dialog box to the user that allows them to choose whether or not to clean up the alias list, but this wouldn't really be an improvement over the current implementation.

With my proposed change, the system could also display a message to the user along the lines of "Removed duplicate alias [xyz]" if necessary.

**Additional context**
Here are some common situations where this occurs for me:
- Scraping data from theNude, which always adds the performer's name as an alias
- Changing the main performer name to one of their aliases and forgetting to also delete the alias
- Adding an alias to a performer who already has a long list of them, and not noticing that the alias was already present

xantror:
This PR implements automatic removal of duplicate aliases on performers, tags, and studios instead of showing a validation error.

Closes #5891

### DIFF
--- a/graphql/schema/types/performer.graphql
+++ b/graphql/schema/types/performer.graphql
@@ -80,6 +80,7 @@ input PerformerCreateInput {
   career_length: String
   tattoos: String
   piercings: String
+  "Duplicate aliases and those equal to name will be ignored (case-insensitive)"
   alias_list: [String!]
   twitter: String @deprecated(reason: "Use urls")
   instagram: String @deprecated(reason: "Use urls")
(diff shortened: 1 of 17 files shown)


WithoutPants (approved):
I don't love the automatic behaviour though I understand its necessity. There's also a discrepancy between the update and bulk update behaviours, but I'm happy just documenting the difference in the graphql interface.

PR #6519, opened 2026-01-23
Add Option to always show 'Open In External Player' for DeoVR and other VR Players

philpw99:
This will add a boolean option "Show 'Open In External Player' button" under the "Interface->Scene Player" setting.

With this option on, when user click on "Open In External Player", Stash will start to serve the prime original file directly to the browser, or to the program with webview.

Purpose: To allow VR Players like DeoVR and HereSphere to directly open the file via http serving. This way there is no need to generate huge .json files and potentially crash the VR Player. Users just need to open the Stash webpage in the VR Player, choose the scene and click on the "Open In External Player" button.

Note: It cannot handle the multi-resolution or multi-file scenario. It will only serve the prime file for that scene.


internal/api/routes_scene.go
```diff
@@ -99,6 +102,44 @@ func (rs sceneRoutes) StreamDirect(w http.ResponseWriter, r *http.Request) {
 	ss.StreamSceneDirect(scene, w, r)
 }
 
+func (rs sceneRoutes) StreamOrgDirect(w http.ResponseWriter, r *http.Request) {
```
WithoutPants: What is the purpose of this end point and how does it differ from the `/stream` end point?

package-lock.json
```diff
@@ -0,0 +1,22 @@
+{
```
WithoutPants: Why are there package files in the root directory?

WithoutPants (changes requested):
This PR seems to be a bit of a mess.

I would probably do away with having the option at all, but ensure that the link opens in a new tab/window if it isn't an android or apple device.

internal/api/routes_scene.go
philpw99: DeoVR and other VR player require a file name with extension. Just "/steam" as the end point won't work.
At least it should be like "/steam.mp4", but the current implementation of "/steam.mp4" is very slow, because it involves FFmpeg re-encoding.

Plus, when serving the media file to VR Players, it will help the player a lot by keeping the file name, like "my_video_3dh_180.mp4". VR players will recognize this media file as a 180 3D side-by-side video by using the original file name. This is a common feature for almost all VR players.

Moreover, a simple "/steam.mp4" will leave VR Players no idea what kind of video it actually is. And since every video in Stash shares the same name like "/steam.mp4", after you watching the 1st video, the second video will start playing at the 1st video's last position.

This end point will allow Stash to serve the original file at "/steam/org" route, if the scene has a file named "my_video_3dh_180.mp4", it can be served at "/steam/org/my_video_3dh_180.mp4". This way it serves the original file while keeping the original file name for the VR players.

Once the link is clicked, the VR player will open the media file and hide the browser. So there is no need to open a new browser tab. VR player's browser is quite rudimentary. So creating a new browser tab for the file will only complicate things and cover the browser with a blank page. Plus, if someone wants to play the original video inside a browser, he already has the media player in the same page.

I wish I didn't have to create this special end point to implement the simple file serving, but there is no other way I can see that can do the trick.

PS: The new end point also serves other file types in the same folder. As the example above, if there is a "my_video_3dh_180.srt" subtitle file in the same folder as the prime file, using "/steam/org/my_video_3dh_180.srt" will serve the subtitle file as well. This will allow VR players to open additional files like *.ass , *.funscript, *.hsp ... etc.

package-lock.json
philpw99: I will get rid of it.

internal/api/routes_scene.go
WithoutPants: This suggests to me a need for a direct link to the file, but I don't think it should be grouped with the `/stream` routes. I would probably define it as `/file/{filename}`. 

The route handler _must_ verify that the filename (without extension) matches one of the files associated with the scene. Otherwise we leak other files in the same directory. Access to funscript and caption files should be based on using the basename of the file. I don't want to allow any string to access any file within the directory.

Serving the video file itself should be done using the `SceneServer` so that it uses the `ReadLockManager`.

I would also suggest that we replace the existing `Stream` link in the `File Info` tab with this link. That way a separate option won't be necessary and the `Open in External Player` can maintain its existing behaviour.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -840,10 +840,11 @@
           },
           "disable_mobile_media_auto_rotate": "Disable auto-rotate of fullscreen media on mobile",
           "enable_chromecast": "Enable Chromecast",
-          "show_ab_loop_controls": "Show AB loop controls",
-          "show_scrubber": "Show scrubber",
-          "show_range_markers": "Show range markers",
-          "track_activity": "Enable scene play history",
```
DogmaDragon, line 859: Please reverse the casing changes.

internal/api/routes_scene.go
philpw99: > I would also suggest that we replace the existing Stream link in the File Info tab with this link. That way a separate option won't be necessary and the Open in External Player can maintain its existing behavior.

Can't agree more. A stream link should have its file name included. However, I don't really want to change the stream link in "File Info", because every time I open a scene, "File Info" is not visible. I have to click on "File Info" then click on the stream link to stream a scene. Too much trouble. I prefer something that's clickable right away. Only "Open In External Player" fits the bill.

Turns out it's not necessary to use a direct link to the file. It just need an URL with the ".mp4" or ".mkv" in the end for DeoVR to work. I don't really need "funscript" and "srt" support either. They won't work anyway.

Right now I am proposing using both `/stream/{filename}` and `/stream` at the same time. Both using the same stream routine. This way we can keep scripts using `/stream` working while streaming with the original file name is also possible. A lot of players can actually take the `/stream` link and work right away. Only DeoVR requires a filename.

There are too many browser agents out there. There is no way to reliably detect which OS the browser is running under. Like in Quest 3 it doesn't successfully detect the Meta's OS as Android. So forcing it to show is really necessary.

I didn't change the behavior of `Open in External Player` at all. It is more like giving it a fall back mechanism: If Stash can detect it under Android, it will do the "android.intent.action...". If it can detect it under iOS, it will do the iOS way. For other cases, it will simply direct stream the file with {filename} at the end of the URL.

ui/v2.5/src/locales/en-GB.json
philpw99: Sorry about this. I'm still a newbie in Golang.

Issue #4023, opened 2023-08-12
[Feature] Custom options for scrubber generation


stg-annon:
### **Is your feature request related to a problem? Please describe.**
A user may want to use a custom interval for the scrubber sprite they want to use, similar to how they are currently able to customize the preview generation video.

This may be even more useful with auto tagging tools that utilize the scrubber sprite with image recognition like [StashTag](https://github.com/cc1234475/stashtag)

potential options might look like:
- resolution 
- exclude start
- exclude end
- image interval

### **Describe the solution you'd like**
options within stash to customize how the scene sprite image is generated

### **Describe alternatives you've considered**
plugin or script that might replace the sprite image with a custom one but this may present stash with an unexpected format and cause issues where if stash is aware of the possibility of custom formats this would not be an issue

WeedLordVegeta420:
As someone with a lot of longer 2+ hour videos, having control over the interval would be nice. My assumption though is that if you have this customization, you would still have to always generate the default sprite since I believe it is used for the phash generation.

WeedLordVegeta420:
@DogmaDragon Ah. I do see now looking at the code that it generates its own sheet for generating the phash.

IAmKontrast:
Feature would be also useful for #4022

WanHaes:
Still no way to customize this? Maybe with a plugin?

PR #6522, opened 2026-01-24
Add options for sprite generation

ghost:
This PR adds some new configuration options for sprite generation:
- Minimum sprites
- Resolution of sprites
- time between two sprites

I've added the corresponding options to the system settings panel.

Sprites are now generated in a single row, to make the generation more manageable, especially for prime numbers.
I couldn't find any drawbacks to this approach since the phash generation is independent from sprite generation.

I've also added some locales using a LLM and as far as I can tell the translation is accurate. 

The added documentation contains some more detailed info about the new settings.

This is related to #4023


DogmaDragon:
Translations are exclusively handled via [Codeberg Translate](https://translate.codeberg.org/engage/stash/).

DogmaDragon:
Scene Tagger has a button to show raw sprite which inline generation probably breaks. 
image

ghost:
> Translations are exclusively handled via [Codeberg Translate](https://translate.codeberg.org/engage/stash/).

If I understand this correctly, I still have to add the values to the en-GB.json manually, right?
If so I will just remove the rest from this PR.



> Scene Tagger has a button to show raw sprite which inline generation probably breaks.

Very good catch, I will fix this.

DogmaDragon:
> If I understand this correctly, I still have to add the values to the en-GB.json manually, right?

Yes.

ghost:
I removed the unwanted locales and provided a fix to the mentioned tagger view problem.
The sprites are now generated in a dynamically sized grid.
The added functionality that comes with this PR remains untouched, while you can easily view the raw sprite in the tagger again.

graphql/schema/types/config.graphql
```diff
@@ -184,6 +184,13 @@ input ConfigGeneralInput {
   scraperPackageSources: [PackageSourceInput!]
   "Source of plugin packages"
   pluginPackageSources: [PackageSourceInput!]
+
+  "Time between two different scrubber sprites"
```
WithoutPants: Unit needs to be documented here and in the result.

WithoutPants (changes requested):
I'd prefer the default settings to be identical to the legacy behaviour if possible.

My proposal would be as follows:
- allow configuration of minimum _and_ maximum numbers of sprites, with the default for both set to the existing behaviour of 81 sprites - 0 is a valid value for both fields and means unset
- sprite interval value of 0 means that interval will be calculated based on maximum sprite number, using 81 as a fallback value if it is unset
- sprite generation should use sprite interval if set, calculating the number of sprites using the video duration. The interval will be adjusted down if the number of sprites falls below the minimum sprite value, and up if it is above the maximum sprite value.

ghost:
I don't get why this magic number '81' is so important. To me it seems like some random number that was chosen in the past.
Am I missing something here? Everything that was generated before this change would still work, because the `vtt` file is still unchanged. Why would one need exactly 81 newly generated sprites. I can only tell from my perspective, but for me this was very unpractical, especially for longer scenes. I, personally, don't care about the number of sprites, but usability.

WithoutPants:
> I don't get why this magic number '81' is so important. To me it seems like some random number that was chosen in the past. Am I missing something here? Everything that was generated before this change would still work, because the `vtt` file is still unchanged. Why would one need exactly 81 newly generated sprites. I can only tell from my perspective, but for me this was very unpractical, especially for longer scenes. I, personally, don't care about the number of sprites, but usability.

It's important in that I don't want newly generated sprite files to be suddenly changed for existing users, compared to the existing ones. 81 sprites is the legacy value.

ghost:
What do you think of a boolean setting "custom sprite generation" at the top of the new sprite generation settings?
This would be disabled by default and the generation would not change until enabled by the user.

WithoutPants:
> What do you think of a boolean setting "custom sprite generation" at the top of the new sprite generation settings? This would be disabled by default and the generation would not change until enabled by the user.

That will work.

ghost:
I implemented the new setting. Please have a look.

WithoutPants:
Superceded by #6588

PR #6523, opened 2026-01-25
Remove max-width from StudioTagger styles

Larsluph:
Currently the tagger view for studios is not taking the entire space available on the screen. From blame, it looks like it never got changed since the initial implementation of the view. Not sure why it was added in the first place.

<table>
<thead>
<th>Before</th>
<th>After</th>
</thead>
<tbody>
<tr>
<td>Before</td>
<td>After</td>
</tr>
</table>

### DIFF
--- a/ui/v2.5/src/components/Tagger/styles.scss
+++ b/ui/v2.5/src/components/Tagger/styles.scss
@@ -286,7 +286,6 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: center;
-  max-width: 1600px;
 
   &-header {
     color: white;


WithoutPants (changes requested):
This doesn't resolve the issue as-is. The `tagger-container` class also has the same `max-width` value. Removing that clause causes the other tagger views to render incorrectly, so a more specific clause will be necessary.

Larsluph:
You're right, I forgot about some custom CSS I added to enable full width on wide screens.
I just tried using the new "troubleshooting mode" and it would require more changes to properly allow widescreen rendering. I'll just stick to using custom CSS for this.

PR #6524, opened 2026-01-25
Captions on SlimSceneData

Valkyr-JS:
Captions data added to `SlimSceneData` fragment.
Added to allow card UI plugins to check for captions data and add a relevant icon.

### DIFF
--- a/ui/v2.5/graphql/data/scene-slim.graphql
+++ b/ui/v2.5/graphql/data/scene-slim.graphql
@@ -11,6 +11,10 @@ fragment SlimSceneData on Scene {
   organized
   interactive
   interactive_speed
+  captions {
+    language_code
+    caption_type
+  }
   resume_time
   play_duration
   play_count


WithoutPants:
I'm a bit concerned about the potential performance impact of this change. The `Captions` resolver doesn't currently have dataloader support, and I think that would be a minimum requirement for this change.

Valkyr-JS:
No problem. I'm not too familiar with GQL resolvers but I'll have a look see if I can figure it out as part of this PR. If performance is still poor after that, then no worries.

Gykes:
@Valkyr-JS Did you want to come back to this? If not I can close it out.

Valkyr-JS:
> @Valkyr-JS Did you want to come back to this? If not I can close it out.

Sorry for the late reply - yeah I'll come back to it later, but it's not something I'm gonna get round to particularly soon I don't think. I'm happy to close for now.

Issue #4136, opened 2023-09-18
[Bug Report] Can't cast any video to Chromecast


hecklawert:
**Describe the bug**
Every time I try to cast any video to Chromecast, the player hangs and doesn't play the video. Chromecast doesn't show any error; it only displays a cast icon. I've tried this with both HTTP and HTTPS, and I get the same error.

**To Reproduce**
1. Go to any video
2. Click on ChromeCast Button
4. Player hangs and doesn't play the video

**Expected behavior**
Video should play at TV with Chromecast

**Stash Version: (from Settings -> About):**
v0.22.1

**Desktop:**
 - OS: Linux Fedora 34 and Linux Fedora 38
 - Browser: Chrome
 - Version: 117.0.5938.62 Stable

**Smartphone:**
 - Device: One Plus Nord
 - OS: Android 10
 - Browser: Chrome and Brave

Eviepayne:
I also encounter this.
Not sure why it happens.
I have https, stash works normally.
the chromecast worked once and then never again.

Eviepayne:
@hecklawert I resolved my issue.
Do you have your stash on reverse proxy? I found that the issue was DNS.
chromecast hardcodes 8.8.8.8 as it's DNS provider. I added a outbound NAT on my firewall to redirect 8.8.8.8 to my local dns and it resolved my issue (note that my local dns had my stash domain)

tl;dr it could be you need to either do what I did or make your domain public DNS.

Qlupic:
I have the same problem and I tried to change DNS to Google both in the router and windows. Didn't help. I am using Windows 10. YouTube casting works as normal.

Eviepayne:
> I have the same problem and I tried to change DNS to Google both in the router and windows. Didn't help. I am using Windows 10. YouTube casting works as normal.

The issue isnt what dns you use. Its that google doesnt know your hostname assuming you have a domain on it.

you have 2 options.

(Not recommended)

1. publish the dns on the internet and make stash public facing.

(Recommended)

2. In your router create a NAT rule that redirects requests from google to your own dns resolver.

For ex. in my opnsense firewall I redirect all DNS queries to my Opnsense (localhost)
The DNS there is a recursor so I get internet published dns as well.

in your case you may want to only redirect 8.8.8.8 unless your dns server is a recursor as well.

Kylindrias:
[...]

PR #6529, opened 2026-01-27
feat: use signed urls for airplay videos

### LINKED ISSUE #2529: Airplay Non-Functional When Credentials is Enabled on Server
**Describe the bug**
Airplay is not functional when credentials is enabled on server.

**To Reproduce**
1. New Server
2. Airplay Initially Works
3. Enable Credentials
4. No More Airplay

### LINKED ISSUE #5015: [Bug Report] Chromecast direct streams with auth cannot be casted
**Describe the bug**
Externally initiated Chromecast casts fail when played from direct stream AND auth is enabled. This is not reproducible when casted from google's builtin cast sender, media controls or native `<video>` casting

**To Reproduce**
1. Connect Chromecast to cactoolv2 https://casttool.appspot.com/cactool/
2. Grab direct stream URL from stash with authentication (/scene/#/stream?apikey=ey)
3. Observe failiure as LOAD succeeds, but mediainfo and mediastatus are empty and chromecast is empty.
4. Load stream URL with transcode (/scene/#/stream.mp4?apikey=)
5. Observe stream.mp4 succeeding
6. Disable authentication (/scene/#/stream)
7. Load stream without authentication string
8. Observe success

**Expected behavior**
Chromecast streams should work with authentication enabled and with direct streams

**Screenshots**
N/A

v0.26.1-19-gb3d35dfa
v0.26.1

 - OS: W10 Pro
 - Browser Chrome
 - Version 126.0.6478.114

N/A

**Additional context**
I believe this might have something to do with how the link is interpreted or parsed.

An authenticated filebrowser instance (with JWT in queryStrings as well) also had issues casting via authenticated URL, unauthenticated filebrowser had no issues.
Caddy static serve had no issues either.
Home-Assistant also uses JWTs but seems to be [fine](https://www.github.com/home-assistant/core/issues/46280)

modal-error:
When using authentication for stash access, cast urls for airplay will not have access to cookies - meaning that Airplay will fail to pass authorization for video stream access.

Updated code to sign urls for videos and streams.

Signed urls by default expire after 24 hours.   Changing TTL is possible via configuration file.

* testing notes: **remote airplay from localhost won’t work,** since appletv’s perspective of localhost is different, try casting from IP address (192.168.0.x:9999) or other local DNS name.

Resolves #2529 
Closes https://github.com/stashapp/stash/issues/5015


modal-error:
Possibly related to https://github.com/stashapp/stash/issues/2529#event-20308350922

feederbox826:
def related

internal/api/authentication.go
```diff
@@ -40,6 +59,16 @@ func authenticateHandler() func(http.Handler) http.Handler {
 				return
 			}
 
+			// Check for signed media requests
+			if isSignedMediaRequest(r) {
+				// Allow signed requests
+				ctx := r.Context()
+				ctx = session.SetCurrentUserID(ctx, c.GetUsername())
```
WithoutPants: The username should be encoded in the signed URL, and it should be verified against the configured username.

internal/api/resolver_query_scene.go
```diff
@@ -39,7 +41,22 @@ func (r *queryResolver) SceneStreams(ctx context.Context, id *string) ([]*manage
 
 	baseURL, _ := ctx.Value(BaseURLCtxKey).(string)
 	builder := urlbuilders.NewSceneURLBuilder(baseURL, scene)
-	apiKey := config.GetAPIKey()
+	expires := time.Now().Add(24 * time.Hour)
```
WithoutPants: This is not using the configured expiry time.

pkg/signedurl/signedurl.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+)
+
+const (
+	ExpiresParam = "expires"
+	SigParam     = "signature"
+)
+
+// SignURL signs a URL with an expiration time using HMAC-SHA256
+func SignURL(rawURL string, secret []byte, expires time.Time) (string, error) {
```
WithoutPants: These functions should have unit tests

internal/api/resolver_model_scene.go
```diff
@@ -107,15 +108,29 @@ func (r *sceneResolver) Paths(ctx context.Context, obj *models.Scene) (*ScenePat
 	baseURL, _ := ctx.Value(BaseURLCtxKey).(string)
 	config := manager.GetInstance().Config
 	builder := urlbuilders.NewSceneURLBuilder(baseURL, obj)
+
+	// Use configurable expiry for signed URLs (only for AirPlay-compatible formats)
```
WithoutPants: We should only build signed URLs if authentication is active.

internal/api/resolver_model_scene.go
```diff
@@ -296,7 +311,22 @@ func (r *sceneResolver) SceneStreams(ctx context.Context, obj *models.Scene) ([]
 	builder := urlbuilders.NewSceneURLBuilder(baseURL, obj)
 	apiKey := config.GetAPIKey()
 
-	return manager.GetSceneStreamPaths(obj, builder.GetStreamURL(apiKey), config.GetMaxStreamingTranscodeSize())
+	endpoints, err := manager.GetSceneStreamPaths(obj, builder.GetStreamURL(apiKey), config.GetMaxStreamingTranscodeSize())
+	if err != nil {
+		return nil, err
+	}
+
+	// Sign each endpoint URL
```
WithoutPants: This should only apply if authentication is active.

WithoutPants (changes requested):

internal/api/authentication.go
modal-error: I wasn't certain exposing usernames in the url was desirable? or are you suggesting that it become additional salt?

WithoutPants: Ideally, the username would be embedded in the signed string so that it could be extracted at the other end. Would it be worth creating a signed JWT token instead that could be decoded? Similar to the way that api key is generated. That way the expiry wouldn't need to be included in the URL either.

modal-error: So in practice expiration and signature are some of the most common approaches to signed urls, my intent here was to follow patterns established in aws + gcloud for signed urls (often used for S3, etc).    

The point of a signed url is that it can't be guessed, is verifiable (can't be tampered with to access another resource) and is opaque  - it shouldn't expose user information or identity. 

https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html

WithoutPants: The example given in the [Google cloud docs](https://docs.cloud.google.com/storage/docs/access-control/signed-urls) passes the credentials used to create the signed URL. Given that we want to eventually add multiple users, including the username in the parameters will be required to associate the request with the original user.

WithoutPants:
I spent a bit of time trying to make this work, but I'm left with a few problems.

I'm not seeing how this will address #2529. The existing code includes the API key, which while not ideal does not really work that much differently to how a signed URL would. 

The implementation of the HLS and DASH manifests would not work since it's just passing through the manifest URL signature, but the segment URLs have a different URL path, so the verification would fail. The solution would be to sign each of the segment URLs, which is fairly straightforward in HLS, but non-trivial for DASH manifests, since it uses segment templates.

feederbox826:
> I'm not seeing how this will address https://github.com/stashapp/stash/issues/2529. 

can't speak for airplay specifically but from my research in https://github.com/stashapp/stash/issues/4136, the direct /stream URL is passed to chromecast. the link works in browsers since cookies are persistent and passed through, but cookies are not passed through to chromecast, only the URL is, which entails the need for a signed URL.

WithoutPants:
> can't speak for airplay specifically but from my research in #4136, the direct /stream URL is passed to chromecast. the link works in browsers since cookies are persistent and passed through, but cookies are not passed through to chromecast, only the URL is, which entails the need for a signed URL.

If API key is generated, it should be sending that through to the Chromecast.

feederbox826:
I can have a look once I fix up my chromecast, no access to airplay but it should follow the same principle

modal-error:
Cookies did seem to be the stumbling block orginally, I haven't abandoned the work, and intend on fixing things up, including shims for a future dereferecing of some kind of identity to satisfy the credential requirement.

Regarding APIKey - Honestly I missed that this exists.   It seems like the kind of credential you might not want to broadcast to a random device if it opens up every resource?  

You are correct about the manifests - which I think were at least partially implemented, for at least some formats - I know when testing some manifest based content I had to fix them, but suspect I only fixed one.

I liked the idea of not broadcasting an APIKey which essentially would open up access to potentially all content, and I liked the idea of not broadcasting usernames for credentials as well.

If there is not interest in this work I can pause as well.

modal-error:
Perhaps there is a simplication here - I think the original implementation here (ignoring some defects) is the idea of having a unique value for a given asset.   This is logically the same thing as the API Key, but is limited in scope from an access control perspective.   The current implementation was based on a shared key, to generate a resource specific key.   In my case I was using the internal credentials to generate it - it could just as easily use the APIKey instead.  

Both of these approaches don't assume a multi-user environment (credential mapping) to validate - which is something that I could only mock an API to support for this feature.   

Happy to pivot.

WithoutPants:
I've pushed the branch of as far as I got with my efforts. Branch here: https://github.com/WithoutPants/stash/tree/signed-urls

I agree that not broadcasting the api key is a decent goal, especially with multi-user support on the horizon and the increasing need to harden security.

I think a potential solution to the segmented streaming issue would be to pass through the existing signature when generating the manifest URLs, then have a special case when validating `/stream.mpd/*` and `/stream.m3u8/*` to validate against the original manifest URL instead.

My changes already incorporate the user name into the signed URL, so it should be good for multi-user support in future. It shouldn't need too much more work to get it working.

feederbox826:
> If API key is generated, it should be sending that through to the Chromecast.

Unable to replicate, see https://github.com/stashapp/stash/issues/5015 for repro and test tools

modal-error:
@WithoutPants @feederbox826 is this a dead end at this point?

WithoutPants (approved):
Looks good and tested ok. I did some minor tidying and lowered the default expiry to 4 hours. I also added documentation for the new config field. I don't really think encoding the user name is that helpful, but I don't care enough to change it. Thanks!

PR #6530, opened 2026-01-27
Create aliases when merging performers

sezzim:
Follow-up to #5910. Honestly, I'm not sure why the previous way didn't work, but the new way matches how it's handled for tag merges.


WithoutPants:
This is wrong, and I'm not quite sure what it's addressing. The merge dialog gives you the option to choose the alias list to use, and this change prevents the user from being able to choose. A quick test of the existing behaviour seems to verify that it works.

sezzim:
I tried to reproduce the behavior that prompted me to make this change and I cannot, so, I will close this.

PR #6533, opened 2026-01-28
New: Add Scene Production Date field

### LINKED ISSUE #1078: Add scene metadata field for original date (of production)
# Edit
(2026-07-21): Since writing this, I've learned that my understanding of `production date` was not completely correct. Because the post hinges on that definition, I'm not sure my original conclusions below are as valid as written. But I'll leave the original portion up for context.

### Definition of production date

It is my layman's understanding that under [18 U.S. Code § 2257](https://anonymiz.com/?https%3A%2F%2Fwww.law.cornell.edu%2Fuscode%2Ftext%2F18%2F2257), the legal definition of *produce* is broad. It encompasses not just filming, but explicitly includes "duplicating, reproducing, or reissuing" content.

There are two distinct categories of *produce*:

1. Primary producer, "any person who actually films, videotapes, photographs, or creates a digitally- or computer-manipulated image, a digital image, or a picture of, or who digitizes an image of, a visual depiction of [...]" aka _original date of production_.
2. Secondary producer, "any person who produces, assembles, manufactures, publishes, duplicates, reproduces, or reissues a book, magazine, periodical, film, videotape, or digitally- or computer-manipulated image, picture, or other matter intended for commercial distribution that contains a visual depiction of [...]"

### Why does this matter?

Because both are legally *productions*, and publishers cite the the term production broadly without specifying primary or secondary in their production dates - it is not necessarily synonymous for a filming date (primary production or original date of production).

We see this in the wild when looking at production dates from third-party retailers, or certain web sites, such as Karups. Karups makes a distinction between production date, filming date, and release date.

Users want to cite publisher production dates for the purpose of documenting original releases or calculating accurate performer scene ages, but given the broad usage of the term, this will often lead to inaccuracies.

### A better metadata field solution 
To maintain canonical accuracy and avoid interpretation errors I propose the following:

| Metadata field | Description|
|---|---|
| Filming date | The date the visual depiction was recorded on camera. (This would be the primary field used when calculating performer scene ages) |
| Production date | The exact 1:1 reflection of whatever the publisher claims as the production date. (Not to interpreted or adjusted since it could be primary or secondary production dates which would both be accurate)

# Original post

Many scenes, including website releases and movie scene compilations, are re-releases not accurately reflecting the original date of production.  This will reflect an inaccurate chronological representation of the scene, of the performers age during the scene, etc. 

Similar to id3v2 and MusicBrainz tag metadata, add metadata fields for original release dates for scenes:
| Metadata field | Description |
| --- | --- |
| originaldate | Original date of production (YYYY-MM-DD) of the earliest release |
| originaldate_url | Source URL for the original date |

**Additional context**
~~i would consider these _advanced metadata_ fields that would not be commonly used. Hence, to minimize confusion or accidental usage, their input forms should not be displayed in the scene metadata editor by default. i propose displaying them when activated via the scene editor's hamburger menu > show advanced tags (which would just change the css display value from hidden to block, or something along these lines)~~

Emilo2:
Add Scene Production Date field to scenes.
Works in same way as regular Date field.

Works with scrapers and stash-box instances
Draft submission is not yet working correctly on stash-box side. (Missing the `production_date` field on the `findDraft` query)

Closes https://github.com/stashapp/stash/issues/1078


feederbox826:
unsure if TODO or already done (models naming is confusing) but taking performer age from production date instead of date if set IIRC was implemented on stash-box

graphql/schema/types/filters.graphql
```diff
@@ -328,6 +328,8 @@ input SceneFilterType {
   last_played_at: TimestampCriterionInput
   "Filter by date"
   date: DateCriterionInput
+  "Filter by production date"
+  production_date: DateCriterionInput
```
Gykes, line 357: This adds `production_date` to the GraphQL filter type, but I don't see a corresponding change in `pkg/sqlite/scene_filter.go` to actually handle this filter. Without that, the filter option will appear in the UI but won't produce any results.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
```diff
@@ -825,6 +832,7 @@ export const SceneEditPanel: React.FC<IProps> = ({
             )}
 
             {renderDateField("date")}
+            {renderDateField("production_date")}
```
Gykes, line 859: The production date can be edited here, but it doesn't appear to be displayed in the scene detail view. `SceneDetailPanel.tsx` displays other scene metadata and `production_date` should probably be added to this as well.

internal/api/resolver_mutation_scene.go
```diff
@@ -57,6 +57,10 @@ func (r *mutationResolver) SceneCreate(ctx context.Context, input models.SceneCr
 	if err != nil {
 		return nil, fmt.Errorf("converting date: %w", err)
 	}
+	newScene.ProductionDate, err = translator.datePtr(input.ProductionDate)
+	if err != nil {
+		return nil, fmt.Errorf("converting date: %w", err)
```
Gykes: You have `converting date: %w` here but `converting production date: %w"` at line 207. Should probably update this one to match

Gykes (changes requested):

Emilo2:
> unsure if TODO or already done (models naming is confusing) but taking performer age from production date instead of date if set IIRC was implemented on stash-box

Didn't even think of this. It now uses the production date to calculate performer age if available.
Thanks.

graphql/schema/types/filters.graphql
Emilo2: Added the handler.
Also added IsMissing handler that was also missing.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
Emilo2: Added the production date to scene details. It shows below the `Updated At` if available.

internal/api/resolver_mutation_scene.go
Emilo2: Fixed the error message

pkg/sqlite/scene_filter.go
```diff
@@ -176,6 +176,7 @@ func (qb *sceneFilterHandler) criterionHandler() criterionHandler {
 		qb.performerAgeCriterionHandler(sceneFilter.PerformerAge),
 		qb.phashDuplicatedCriterionHandler(sceneFilter.Duplicated, qb.addSceneFilesTable),
 		&dateCriterionHandler{sceneFilter.Date, "scenes.date", nil},
+		&dateCriterionHandler{sceneFilter.Date, "scenes.production_date", nil},
```
Gykes: I believe this should be `sceneFilter.ProductionDate`. Probably just a copy paste mistake

If you use `.Date` it will just filter the same data as the regular date filter.

Emilo2: Thanks, this is now fixed also

Gykes (approved):
@WithoutPants Looks good to me. OP made plenty of fixes.

WithoutPants:
This needs a bit more design consideration. My assumption as to the meaning of the date field has always been the production date (since that's what we use to determine performer age). With the addition of a production field what is supposed to be the intent of the date field now?

I'm not entirely convinced that a single release date field is the correct way forward either. As noted in the original issue, scenes can be released multiple times. In the shorter term I'd be inclined to capture such information in scene custom fields (once implemented). Apart from a metadata completion point of view, I'm not sure what core functionality a release date field brings as a core field.

Gykes:
Date is the release date. 

Production date is the date it was recorded/created

You are correct that this is mostly just another metadata slot. I believe production date would be a more accurate representation of performer age at scene time rather than release date. Could have been produced 6m-1yr earlier. 

If StashDB already supports this then I see no reason to not support it in StashApp. If it is unsupported in StashDB then I see no reason to not support it and then get it added to StashDB.

Just my 2c

WithoutPants:
> Date is the release date.
> 
> Production date is the date it was recorded/created

This is based on your personal use and is arbitrary. We haven't dictated the meaning of the existing date field prior to now, so it's meaning is going to differ based on personal use and intent. Notably, release date is only relevant for studio-produced content, and not for amateur or personal content. I'm not keen on changing the meaning of the existing date field for users without a decent reason.

Gykes:
> We haven't dictated the meaning of the existing date field prior to now, so it's meaning is going to differ based on personal use and intent. 

True to a certain extent. With a roughly 30 second search on StashDB docs I found [this](https://guidelines.stashdb.org/docs/scenes/edit/scene-date/preferred-scene-dates/#preferred-scene-dates) 

It specifically states not to use production date for the scene date field. Granted it's not specifying what a Production date is but the verbage seems to lean that production date is different from a regular date. 

> Notably, release date is only relevant for studio-produced content, and not for amateur or personal content.

Disagree here, tons of amateur content is recorded/created in the past and then released on a schedule


Either way I wont be using it so I will let others who have some skin in the game speak to it.

cj12312021:
I had the same thought as WP when I first saw this PR. Providing this as a scene custom field seems like a better fit.

Emilo2:
Here's the stash-box side reasoning for the production date field https://github.com/stashapp/stash-box/issues/346
It has been functioning there already for quite some time and this just implements the provided field to stash-app side.

I have scenes that have been released first time this month with the production date from six years ago 2020. So now the performer age is correctly shown on the cards.

smith113-p:
> Providing this as a scene custom field seems like a better fit.

Scenes do not support custom fields currently, although I'd love it if they did.

AdultSun:
> This is based on your personal use and is arbitrary. We haven't dictated the meaning of the existing date field prior to now, so it's meaning is going to differ based on personal use and intent. Notably, release date is only relevant for studio-produced content, and not for amateur or personal content. I'm not keen on changing the meaning of the existing date field for users without a decent reason.

I disagree that its arbitrary. There's already a pretty strong argument that we *have* established "date" as "release date", even if we haven't stated that explicity.

First of all, Gykes's definition already matches StashDB's definition and usage. I realize StashDB is not the same as Stash, and not everybody using Stash uses StashDB, but it still sets an expectation within the userbase.

Second, I don't think there's a single scraper out there that would pull a production date into the "date" field instead of the release date. Truly, in all the discussions leading up to the addition of prod. dates in Stash-Box, and all the prod. date edits I've seen in StashDB afterwards, I have never seen an example of a scrapeable source for production date fields. It's always listed in the video as a production card or watermark, printed on a DVD back cover, mentioned on social media or a blog post, written out as a "comment" in IAFD, implied by embedded metadata, implied by an earlier release of the same content, etc. And that lack of a centralized source for prod. dates was a big motivator for adding them to StashDB in the first place.

If anything, I keep forgetting that the field in Stash and Stash-Box is still just named "Date" because everything within Stash's ecosystem and the wider community have always assumed it means "Release Date". And the only people using that field to track production dates (myself included but I've got everything flagged with a custom tag so it's fine) had to make a deliberate choice to overwrite any release dates they've scraped, or have populated their database without ever using scrapers in the first place. Either way, I wouldn't expect changing the label to "Release Date" and adding a "Production Date" alongside it to affect many people or to be entirely unexpected.

cj12312021:
I missed the detail that production date was already added as a field to StashDB. Given Stash's tight integration with StashDB, there may be an argument for supporting it in the scrape/submit workflow.

While I would also find value in the addition to Stash, I found it hard to argue that a new date field would not be confusing.

If I'm a new Stash user tagging content in my library (which may or may not contain adult content), I now have two date fields on a scene and no obvious guidance on which to populate. For users who aren't plugged into the StashDB ecosystem or its conventions, the distinction between Date and Production Date isn't self-evident.

There's also an entity consistency question. If the distinction between release date and production date is meaningful enough to warrant a core field on scenes, the same argument applies to groups, galleries, and images. Adding it to scenes alone either stays as a permanent inconsistency or implicitly commits to adding the field across all content types.

The subset of users who have access to production date data is likely relatively small, and the primary functional benefit (more accurate performer age) is narrow. The custom fields work planned for scenes seems like the better route forward, satisfying the date variation (original studio, repackager, etc.) use case without the trade-offs of a new core field.

Gykes:
> I missed the detail that production date was already added as a field to StashDB. Given Stash's tight integration with StashDB, there may be an argument for supporting it in the scrape/submit workflow.

I also believe that we should stay concurrent with StashDB. As the gap gets larger and larger then it will become harder to catch up. Someone already did the work for us so why not. StashApp and StashDB may be separate entities but they live under the same roof. I believe that if one is doing it then we should mirror/accommodate it.


> If I'm a new Stash user tagging content in my library (which may or may not contain adult content), I now have two date fields on a scene and no obvious guidance on which to populate. For users who aren't plugged into the StashDB ecosystem or its conventions, the distinction between Date and Production Date isn't self-evident.

I think you underestimate the average users IQ. If I ask you the date of Avengers Endgame you would more than likely give me April 22, 2019 with some minor variation depending on where you live. You aren't going to go around digging for the first day that the movie started production and then give me that.... it makes no sense. We would just update the documentation to represent that production date is not "date" (release date)

>There's also an entity consistency question. If the distinction between release date and production date is meaningful enough to warrant a core field on scenes, the same argument applies to groups, galleries, and images. Adding it to scenes alone either stays as a permanent inconsistency or implicitly commits to adding the field across all content types.

Agreed, I feel like this field should be applied to pretty much all objects where it is applicable.  We currently have tons of features that exist in scenes that do not in other objects. This isn't a hard rule of we wont have them there just more develop time has went into scenes than anything else due to the nature of StashApp.


> The subset of users who have access to production date data is likely relatively small, and the primary functional benefit (more accurate performer age) is narrow. The custom fields work planned for scenes seems like the better route forward, satisfying the date variation (original studio, repackager, etc.) use case without the trade-offs of a new core field.

Agreed but there are tons of users who have niche data. Are you saying that we should ignore that data? The point of Stash is to be a media manager/collector. I see no reason to not support another data field. It's not something that adds huge technical debt or requires a lot of work for future development.

cj12312021:
> I also believe that we should stay concurrent with StashDB...

I don't disagree with staying aligned with StashDB, but alignment doesn't have to mean mirroring every field as a core schema addition. Supporting production date in the scrape/submit workflow is different from adding a permanent column to the database. StashDB integration can cover the addition of custom fields.

> I think you underestimate the average users IQ...

This isn't about intelligence. There's an assumption here that most users have awareness of the studios behind the content they watch. In reality, most consumers exist completely outside of that world and don't pay attention to production details. For those users, two date fields with no obvious distinction is a UX problem, not a knowledge problem. That aside, Stash isn't only a tool for studio content. For users tagging personal or amateur content, the distinction between Date and Production Date is less intuitive.

> Are you saying that we should ignore that data...

Not at all. I'm saying custom fields are a better vehicle for it than a one-off core field. A custom fields system would support production date, original release date, repackage date, or whatever else users need, without each one requiring a migration, filter handler, and cross-entity consistency discussion. The data gets captured either way. The question is whether a purpose-built flexible system or a series of individual schema additions is the right path.

Gykes:
So, I feel like this has gotten to the point of further discussion of the fruition of this feature should probs move to Discourse or something lol.

There seems to be a desire for this from various users so I would take that into account. As I said before I have no skin in the game so I will leave my opinions as they are. WP can take a look and decide if it's worth adding or just using custom formats.

sleetx:
Regardless of video topic, every scene is produced at some point in time. Every scene is released at some point in time. They are usually, but not always, different. Thus, it would make sense to have two separate fields.

>We haven't dictated the meaning of the existing date field prior to now, so it's meaning is going to differ based on personal use and intent.

This is true, but it's almost universally used as release date. I struggle to find an example where it's used for something else. And as @AdultSun mentioned, I have never seen a scraper source that uses production date there.

> I'm not keen on changing the meaning of the existing date field for users without a decent reason.

Then can you clarify the meaning of the existing date field? There are a couple of places in the StashDocs that already refer to a scene's "release date", rather than just "date".

To @cj12312021's point, if the "Date" field title causes confusion, it might be worth re-labeling it as "Release Date" on the front end if we're adding "Production Date".

>The subset of users who have access to production date data is likely relatively small, and the primary functional benefit (more accurate performer age) is narrow.

Although I support adding the field, this is the counter-argument I could see some sense in. StashDB is adding production date metadata, but StashDB doesn't represent every user. If we're not seeing sufficient demand on the core app side, maybe it's not worth the trouble.

> Notably, release date is only relevant for studio-produced content, and not for amateur or personal content.

Just because a scene is released on a particular day, doesn't mean it was produced the same day. Regardless if it's by a professional or amateur. There is an argument for two separate fields on all types of scene content (see my intro sentence).

DogmaDragon:
> My assumption as to the meaning of the date field has always been the production date (since that's what we use to determine performer age).

Me and I think most other people were *de facto* reading **Date** field as release date despite it being technically undefined. Existing community scrapers, stash-box integration and common **date** term usage to refer to **release date** on other databases were influencing how the field was perceived.
 
> Apart from a metadata completion point of view, I'm not sure what core functionality a release date field brings as a core field.

Effectively release date is the only date related metadata point that is universally known. Be it mainstream-studio production, independent creator distribution, amateur recording, live-streamed content, pre-digital DVD releases or personal recording will technically have a known release date. It might be strectching a meaning a bit for personal recording, but I don't think it's outlandish to consider release date the same as production date for that type of content.

Production date on the other hand is often unknown. The only content where someone can be sure of the production date is personal recordings which I would dare to say is not what majority uses Stash for.
Even cases that do offer prodcution date like DVD releases, they are often partial and Stash didn't support partial dates until recently, which was another subconcious indication that **Date** field was *de facto* release date. 
Live-streamed content is often re-released as a VOD, so even when production date is seemingly obvious it's not always the case. 

I share the concern that having 2 date fields is not ideal, but I disagree that production date should be the default meaning. 

I would suggest making release date clearly denoted default date value and migrating production date to custom fields (which are now implemented).

smith113-p:
I agree that release date is the more obvious meaning of the field -- using it for the release date means that scenes will be sorted the same way as they are on the studio website, which makes it easier to see if you are missing any scenes.

Emilo2:
Custom field seems like a good solution, but it doesn't seem to support scrapers and stash-box drafts out of the box. So it would be trickier to implement. 

In a perfect world the Date object should have a type enum (Release, Production, etc.) and then just make it a list. But this would require tons of refactoring.

AdultSun:
Another concern I'd have with the custom field solution is I don't know how that would affect filtering and sorting based on production date.

I can see in the dev build that custom scene values can be filtered, but I'm not sure it expects or supports values that represents a date. I see "greater than / less than" and "between / not between" as options but I don't know how well those would work for YYYY-MM-DD values. I'm sure there's a way to incorporate that kind of support, but I don't know how long or complicated that would be to develop, or how intuitive it would be as a user.

Sorting seems trickier to me. Right now, the sort dropdown doesn't have any option for custom fields. And again, I don't know how well a generic option for custom values would work for dates anyway. And even then, I don't see how a generic option could understand the relationship between production dates and release dates compared to a hardcoded production date field.

Really the main usecase I have in mind for production dates within Stash is the ability to sort scenes into a chronological order based on our best understanding of when they were filmed. This would make it easier to mass tag performer attributes that change over time, recognize which scenes represent a performer's "first" of something, identify re-releases and redistributions, etc. But the only way that works is if we can easily and intuitively sort scenes according to production date first, with release date used as a fallback. As we've discussed, even estimated production dates are not available for most scenes, so any sorting option for production dates that doesn't consider release dates would have very limited usefulness.

DexXxter103:
For me personally, the most important thing is being able to follow a performer's development over time. Being able to tell that one scene is from her early days while another showcases her full experience after 10+ years — I find that really interesting.

Another reason is re-releases and quality upgrades: during 2020–2023, when relatively little new content was being produced, many studios went back and re-released older scenes in higher resolutions. What bothers me is when my Stash instance presents, say, a Natasha Nice looking like she did in 2016 as supposedly new content here in 2026, when the performer herself has changed quite a bit since then.

As a relatively new Stash user (I only discovered Stash for myself in late 2024), I started doing this early on: after scraping, I manually reset the date back to the original release date. This of course creates conflicts with StashDB and similar sources, where the policy seems to be — at least in some cases — that each release is its own separate entry, meaning a re-released scene ends up in the database twice. It also holds me back from submitting corrections to entries, because I'm worried my way of curating data might violate some guideline, and I'd rather avoid running into trouble. And yes, I'm probably one of those users contributing to scene entries having inconsistent timestamps — for example, I already had the original 2014 scene in 1080p in my database and then in 2024 updated it to the 2160p re-release, merging both scenes into one on my end in the process by keeping the old 1080p Stash ID reference.

And that ties back to what I said at the beginning: I enjoy following a performer's development. It just doesn't fit when a performer who has, for example, gotten heavily tattooed in the meantime because her personality has changed is suddenly shown to me in a scene from before that change, presented as if it were current.

> I agree that release date is the more obvious meaning of the field -- using it for the release date means that scenes will be sorted the same way as they are on the studio website, which makes it easier to see if you are missing any scenes.

I understand this approach, and that's often how I complete my collection too. It gets problematic, however, with studios like NaughtyAmerica and performers who have been in the business since 2010 or earlier: I often find myself thinking "Oh, she is active again? How could I miss this scene from 202x?" — only to then realize it's a re-release of a scene from 201x I've owned for years and already upgraded to.

Gykes:
I don't believe the main dev wants to introduce a new date field. It opens a can of worms that currently can be handled by custom fields. With that in mind I am closing this for now and if, in the future, the opinion changes we can reopen

DogmaDragon:
I think the can of worms needs to be solved either way because right now there is a conflict in how maintainers vs general public perceive field `date` meaning.

Gykes:
I think that is something we can put in the documentation. When I see a date on a movie poster that is the release date, not production date. I could be wrong but I think most people have the same thought process as that.

smith113-p:
Studios list their releases in order of release date. I think most people would prefer having the same order in their stash so they can easily compare what's missing; the production date does not have this property, if it's even known. Whereas, as mentioned, the production date (if available) does help with performer ages.

sleetx:
> I think that is something we can put in the documentation. 

The documentation already refers to "release date" in several places. [Example](https://docs.stashapp.cc/in-app-manual/tasks/jsonspec/?h=date#scenejson)

> When I see a date on a movie poster that is the release date, not production date. I could be wrong but I think most people have the same thought process as that.

We've got nearly unanimous agreement in this thread, and good replies to WP's discussion points since he last responded. It is premature to close this.

Gykes:
> The documentation already refers to "release date" in several places. [Example](https://docs.stashapp.cc/in-app-manual/tasks/jsonspec/?h=date#scenejson)

Problem solved then.

>  It is premature to close this.

I disagree. I have been in a few voice chats with WP and he has always leaned towards just keeping a single date and using custom fields for the extra ones. He also voiced that StashApp does not need to be a 1:1 clone of StashDB. Between his comments, the voice chats, and general vibes I have gotten from him on the matter i'm keeping it closed for various reasons. 

1) To clean up the PR list
2) To give some sort of actual hard decision. The decision can be overturned if WP feels otherwise and the PR reopened. There is no harm in closing

DexXxter103:
Could we please reopen this PR? Several developers had already expressed support for it, so I don't see a reason to close it.

If the PR was closed simply because you, @Gykes, don't want to keep long-running PRs open, I'd ask you to reconsider that decision.

Otherwise, I'll just open a new PR, because - as I mentioned in https://github.com/stashapp/stash/pull/6533#issuecomment-4237423848 - I consider this to be a very important feature.

For example, I'm always happy when studios like Naughty America re-release older scenes that I previously only owned in HD in a new 4K version. What I do find frustrating, however, is when the Stash community then ends up treating a scene that was originally produced in 2014 (where the performer may have long since retired from the industry) as if it were a scene from 2026.

To make matters worse, Naughty America sometimes even gives these re-releases completely new titles, which understandably provides arguments for those who believe they should be entered as entirely new scene records rather than being associated with the original 2014 scene.

For people like me, who place more value on knowing **when a scene was actually produced**—for example, to understand that *"the performer didn't have tattoos yet,"* *"this was before her breast augmentation,"* or simply to follow a performer's development over time—this creates a real problem.

If we merge the 4K re-release into the original 2014 scene, we also add its hash. If our viewpoint happens to be the majority, we're effectively forcing everyone else — including those who believe it should be a separate scene — to follow our interpretation for people identifying scenes by fragments. Right now, the opposite is happening. Yet I think we all ultimately share the same goal: maintaining high-quality data in StashDB.

**tl;dr**

I believe StashDB needs a dedicated **Production Date** field, and I don't think this PR should be closed without either a replacement PR or a clear statement from the developers that they have decided against this feature. The latter would at least provide a clear direction for users like me, allowing us to explore alternative solutions — including maintaining a fork of Stash (which I really would like to avoid) — because this is not something that can be cleanly implemented using custom fields.

smith113-p:
@DexXxter103 stashdb already has a production date field. this PR is about stash.

Gykes:
> If the PR was closed simply because you, @Gykes, don't want to keep long-running PRs open, I'd ask you to reconsider that decision. 

Get your AI to reread my comment and you will see that there are 2 reasons to close it. One for PR cleanliness, and one to give OP a hard decision. If you read directly above that you will fine the reasonings behind a hard decision. 

>Otherwise, I'll just open a new PR, because - as I mentioned in https://github.com/stashapp/stash/pull/6533#issuecomment-4237423848 - I consider this to be a very important feature.

You may do so and I will  close it then link it to this PR. There will be no benefit for you to do so. Just because a PR is closed does not mean it is gone forever. The code and branch still exist. If we want to implement this all I have to do is click one button and it all comes back to life.

I appreciate that it is important to you but, as of now, the dev team is leaning towards keeping it out. That may change and no decision is permeant. 

> I believe StashDB needs a dedicated Production Date field, and I don't think this PR should be closed without either a replacement PR or a clear statement from the developers that they have decided against this feature. The latter would at least provide a clear direction for users like me, allowing us to explore alternative solutions — including maintaining a fork of Stash (which I really would like to avoid) — because this is not something that can be cleanly implemented using custom fields.

Lots of people believe this and it has been said tons of times. This wall of AI text is just making an already clogged up PR comment section more clogged.

DexXxter103:
Please be careful with your wording. I am personally annoyed by the large number of AI-generated posts as well, but that should not lead us to immediately assume that every post is AI-generated.

Please excuse me if my English sounds somewhat overly elaborate. English is not my native language, and unfortunately I tend to express myself in a more detailed way rather than getting straight to the point, because I try to avoid potential misunderstandings upfront.

@smith113-p: Yes, I also noticed a few weeks ago that the field had been added to StashDB, and I have been eagerly waiting for the possibility to populate it through my local Stash instance. I already have a list of numerous JJ and NA scenes that I would like to update.

However, as I understand it, this also requires an implementation in Stash itself. And that was exactly the purpose of this PR, which now appears to have been closed. Am I wrong?

@Gykes Please, don't get me wrong and I don't want anyone to annoy. But in practice, closed issues and PRs are generally considered "done" or "rejected". At the very least, nobody will actively work on them anymore. That is my understanding of the current situation with this feature request: nobody is working on it anymore, and therefore it is unclear whether the ability to set the Production Date in Stash will ever become available.

And that is what concerns me. :(

Regarding the statement:

> I don't believe the main dev wants to introduce a new date field. It opens a can of worms that currently can be handled by custom fields. With that in mind I am closing this for now and if, in the future, the opinion changes we can reopen

I have to disagree: No, this cannot be properly implemented using custom fields.

The date is used in many different places throughout Stash. Trying to hook into all of these locations would not be practical. Additionally, the value would already need to be considered during sorting. However, this query is executed directly by Stash itself and not by Apollo, where most plugins hook into.

Furthermore, does this statement not contradict the decision made by the StashDB developers, who recently introduced this field there?

I'm sorry, but I genuinely cannot understand the decision to close this issue. Anyone having the same request in the future will search for it, find nothing because closed issues are filtered out by default, and new duplicate requests will be created.

Could we at least create a higher-level tracking issue for this topic and keep that open?

Gykes:
> Please be careful with your wording. I am personally annoyed by the large number of AI-generated posts as well, but that should not lead us to immediately assume that every post is AI-generated.

> Please excuse me if my English sounds somewhat overly elaborate. English is not my native language, and unfortunately I tend to express myself in a more detailed way rather than getting straight to the point, because I try to avoid potential misunderstandings upfront.

Has nothing to do with being elaborate. Has everything to do with grammar. People don't talk like how the comment was which is the assumption of AI. 

> However, as I understand it, this also requires an implementation in Stash itself. And that was exactly the purpose of this PR, which now appears to have been closed. Am I wrong?

You are not wrong just you listed StashDB in your above comment, not StashApp. Was more than likely a simple mistake.

>  Please, don't get me wrong and I don't want anyone to annoy. But in practice, closed issues and PRs are generally considered "done" or "rejected".

Yes, as of now it is rejected. I think I was pretty clear with that. As of this exact moment it is rejected. It can be reopened in the future if StashApp dev team opinions change. I have made a few PRs that got rejected then a few months later opened them up as StashApp team opinions changed. 

> and therefore it is unclear whether the ability to set the Production Date in Stash will ever become available.

Again, as of this moment, no. In the future, maybe. Keeping this PR open does NOTHING expect set an expectation that it is going to be added when there is no such expectation. 

> StashDB developers

StashDB dev != StashApp devs. I said it before that the main dev noted that StashApp and StashDB DO NOT need to be 1:1.

> I'm sorry, but I genuinely cannot understand the decision to close this issue. Anyone having the same request in the future will search for it, find nothing because closed issues are filtered out by default, and new duplicate requests will be created.

That's okay, mainly this is a PR and not an issue. If we had a PR open for every issue that might someday possible hypothetically maybe at some point make it into core open then our PR list would be a billion long. 

> Could we at least create a higher-level tracking issue for this topic and keep that open?

If there is not an issue open for it currently then you are welcome to make one. Issues are supposed to stay open for long standing issues NOT PRS.

AdultSun:
@DexXxter103 

> However, as I understand it, this also requires an implementation in Stash itself. And that was exactly the purpose of this PR, which now appears to have been closed. Am I wrong?

To be clear, you can still add or change a scene's production date in StashDB. You just can't populate the production date field automatically when submitting a draft from Stash. It has to be done manually inside of the edit form before sending it to the queue.

---

We've drifted quite a bit in this comment thread away from discussing the PR itself, so I'll jump over to issue #1078 to continue the larger discussion of how productions dates should be handled in Stash.

echo6ix:
> We've drifted quite a bit in this comment thread away from discussing the PR itself, so I'll jump over to issue #1078 to continue the larger discussion of how productions dates should be handled in Stash.

@AdultSun Thanks for posting this. It was written in 2021 and since then I've gained a better understanding of the term _production date_ as it's defined legally in 2257 and as it's used by publishers. FWIW original post has been edited to better reflect that understanding.

WithoutPants:
After discussion with the team, I'm reopening this. I still need some time to consider the impacts of this change, but this is largely how I'd implement it.

pkg/scene/import.go
```diff
@@ -110,6 +110,12 @@ func (i *Importer) sceneJSONToScene(sceneJSON jsonschema.Scene) models.Scene {
 			newScene.Date = &d
 		}
 	}
+	if sceneJSON.ProductionDate != "" {
+		d, err := models.ParseDate(sceneJSON.ProductionDate)
+		if err == nil {
```
Gykes: We should probably log a warning here rather than dropping silently. I know the date above doesn't log but that should be fixed at some point.

graphql/schema/types/scene.graphql
```diff
@@ -245,6 +249,7 @@ type SceneParserResult {
   director: String
   url: String
   date: String
+  production_date: String
```
Gykes: `SceneParserResult.production_date` is added here and in `filename_parser.go`, but nothing populates it. `sceneHolder` in `filename_parser.go` only ever sets `h.result.Date`, and `SceneFilenameParser` doesn't consume it. This will always return null. I would wire this in.

ui/v2.5/src/components/Tagger/scenes/StashSearchResult.tsx
```diff
@@ -643,6 +627,21 @@ const StashSearchResult: React.FC<IStashSearchResultProps> = ({
...
+      return (
+        <h5>
+          <OptionalField
+            exclude={excludedFields[fields.production_date]}
+            setExclude={(v) => setExcludedField(fields.production_date, v)}
+          >
+            {scene.production_date}
+          </OptionalField>
+        </h5>
+      );
```
Gykes, line 666: Wouldn't this just show two unlabeled dates stacked on the tagger?

Gykes (changes requested):
So, we got the go ahead from the main dev to push this further. I left some more comments and will put some more thoughts here. Also needs a rebase and a migration update. I didn't actually test this yet, all just code thought.

Scrapers: `mapped_results` dont seem to read a production date key so xpath and json scrapers cant produce the fields. I think that would effect most community scrapers. 

Identify: `getScenePartial` handles data but never `ProductionDate`. Identify will fetch it from stashbox and then throw it away. The identify dialog also needs the production date too.

Merge and bulk edit: The merge dialogues don't have a production_date row. I think the bulk scene update would be effected by this as well.

Cleanup: Rebase.

Gykes:
@Emilo2 If you would like to pick up the changes that I put above let me know. If not I can take over the work for you to get it over the finish line. 

Thanks!!

Emilo2:
Would be great if you could jump in. I'm currently quite busy with other projects.

Gykes:
**Tested:**
- Scene detail/edit panels: tested
- Scene modals (merge, bulk edit): tested
- Merging logic: tested
- Filtering: tested
- Sorting: tested
- Performer age (production date, falls back to date): tested
- Tagger scraping from stash-box: tested
- Import/export: unit tested
- Anonymiser : no change needed, doesn't touch dates
- Identify: tested

**Not tested live:**
- xpath/JSON scrapers: no scraper defines the field yet

**Notes:**
- Removed `production_date` from the `stash` scraper type older Stash servers reject the unknown field and the whole scrape fails. Doesn't affect stash-box.
- Dropped `SceneParserResult.production_date` was never populated.
- Only StashDB verified as serving the field; TPDB unchecked/JAVStash/FansDB Untested.

PR #6552, opened 2026-02-06
Merge tag values dialog

### LINKED ISSUE #6545: Tag merge does not allow for setting fields
### Describe the bug

Currently, merging tags deletes the source tags without copying any data other than the parent/child tags. I believe when the merge functionality was implemented, there was no or few other fields on tags.

The scene and performer merge dialogs give the option to set fields based on the source object. The tag merge dialog should be similarly updated.

The graphql schema will need to be updated to accept an optional `TagUpdateInput` to facilitate this.

### Steps to reproduce

Select the merge item from the dropdown menu on the tags page, after selecting tags to merge. The merge dialog will be shown allowing for selection of source and destination. Clicking the `Merge` button will immediately merge the tags, without giving an option to select data to save. 

### Expected behaviour

After clicking on the `Merge` button in the merge dialog, the UI should display another dialog allowing for selection of field values like performer/scene merge.

_No response_

### Stash version

v0.30.0

### Device details

_No response_

### Relevant log output

```shell

```

WithoutPants:
Resolves #6545 

Adds a subsequent step to the merge tags dialog which allows choosing the values to apply. Changed the `TagsMerge` interface to accept values. Merging of the tag hierarchy is no longer performed automatically in the backend, as it accepts the parent and sub tags instead.

image


ui/v2.5/src/components/Tags/TagMergeDialog.tsx
```diff
@@ -41,7 +445,18 @@ export const TagMergeModal: React.FC<ITagMergeModalProps> = ({
...
+  async function loadTags() {
+    const tagIDs = src.map((s) => s.id);
+    tagIDs.push(dest!.id);
+    const query = await queryFindTagsByID(tagIDs);
+    const { tags: loadedTags } = query.data.findTags;
+
+    setLoadedDest(loadedTags.find((s) => s.id === dest!.id));
+    setLoadedSources(loadedTags.filter((s) => s.id !== dest!.id));
+    setSecondStep(true);
+  }
```
Gykes, line 462: This function has no error handling. If `queryFindTagsByID` fails, the error is unhandled and the UI will hang with no feedback, right? . Would wrapping in try/catch with a toast    notification, similar to how `onMerge()` works, be a good idea?

Chances of this happening are almost nill but this was the only thing I could really find.

Gykes (changes requested):

PR #6556, opened 2026-02-07
Bugfix: AVIF Image PHash Support

### LINKED ISSUE #6535: Bug: Image Phashing Doesn't Support AVIF
### Describe the bug

WIth the addition of https://github.com/stashapp/stash/pull/6497 it looks like it has issues phashing AVIF files. This is more than likely due to the upstream issues that we encountered to even partially support importing and generating some content for them.

### Steps to reproduce

Try and generate a Phash for a AVIF Image

### Expected behaviour

Phash the AVIF Image

Dogma has volunteered his logs.

### Stash version

Develop

### Device details

_No response_

### Relevant log output

```shell

```

Gykes:
With the addition of https://github.com/stashapp/stash/pull/6497 it allows users to generate PHash for images. AVIF was spitting out errors due to FFMPEG not playing well with stdin and AVIF. This corrects the issue by converting to BMP first so that the image can be hashed. 

fixes: https://github.com/stashapp/stash/issues/6535

### DIFF
--- a/pkg/hash/imagephash/phash.go
+++ b/pkg/hash/imagephash/phash.go
@@ -36,6 +36,10 @@ func Generate(encoder *ffmpeg.FFMpeg, imageFile *models.ImageFile) (*uint64, err
 func loadImage(encoder *ffmpeg.FFMpeg, imageFile *models.ImageFile) (image.Image, error) {
 	ext := strings.ToLower(filepath.Ext(imageFile.Path))
 	if ext == ".avif" {
+		// AVIF in zip files is not supported - ffmpeg cannot read files inside zips
+		if imageFile.Base().ZipFileID != nil {
+			return nil, fmt.Errorf("AVIF images in zip files are not supported for phash generation")
+		}
 		return loadImageFFmpeg(encoder, imageFile.Path)
 	}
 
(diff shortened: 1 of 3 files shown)


Gykes:
Will be testing with HEIC tomorrow. If not I will make some additions to support.

pkg/hash/imagephash/phash.go
```diff
@@ -27,7 +32,13 @@ func Generate(imageFile *models.ImageFile) (*uint64, error) {
 }
 
 // loadImage loads an image from disk and decodes it.
-func loadImage(imageFile *models.ImageFile) (image.Image, error) {
+// For AVIF files, ffmpeg is used to convert to BMP first since Go has no built-in AVIF decoder.
+func loadImage(encoder *ffmpeg.FFMpeg, imageFile *models.ImageFile) (image.Image, error) {
+	ext := strings.ToLower(filepath.Ext(imageFile.Path))
```
WithoutPants, line 37: I believe this will break for images in zip files.

WithoutPants (changes requested):

pkg/hash/imagephash/phash.go
Gykes: We check for AVIF in zips at the initial scan and reject them if so. I can add in extra logic incase zips make it through but they shouldn't get this far.

WithoutPants: Yeah an extra sanity check can't hurt.

Gykes: Done

PR #6563, opened 2026-02-09
feat: add SHA1 as optional hashing algorithm for video files

hyper440:
## Summary

Add SHA1 checksum calculation as an optional hashing method alongside oshash and MD5. This provides an additional fingerprint option for video file identification.

## Changes

- Add `pkg/hash/sha1` package with SHA1 hash calculation functions
- Add `FingerprintTypeSHA1` constant to models
- Add `CalculateSHA1` config option with UI toggle in Settings
- Integrate SHA1 calculation in fingerprint calculator
- Update GraphQL schema and resolvers
- Update documentation to include SHA1 in hashing algorithms section

## Usage

SHA1 is disabled by default and can be enabled in:
**Settings → System → Hashing** (requires Advanced Mode toggle)

When enabled, SHA1 checksums are calculated during scan alongside oshash for video files.

## Testing

- Built and tested on Windows
- Verified SHA1 calculation works correctly during scan
- Confirmed setting persists and can be toggled on/off

### DIFF
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -91,6 +91,8 @@ input ConfigGeneralInput {
   ffprobePath: String
   "Whether to calculate MD5 checksums for scene video files"
   calculateMD5: Boolean
+  "Whether to calculate SHA1 checksums for scene video files"
+  calculateSHA1: Boolean
   "Hash algorithm to use for generated file naming"
   videoFileNamingAlgorithm: HashAlgorithm
   "Number of parallel tasks to start during scan/generate"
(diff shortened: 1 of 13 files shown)


Gykes:
What's the use case? SHA1 is slower than oshash but less secure/weaker than SHA-256. For file identification, oshash + MD5 already cover the fast and thorough hashing cases. What specific scenario does SHA1 enable that those two don't? Stash-box fingerprint matching uses oshash, MD5, and phash. It doesn't recognize SHA1, so the hash would be calculated and stored but never used by anything. Is there a third-party tool or service you're integrating with that needs SHA1 specifically?

Not trying to shoot you down just trying to understand the need

hyper440:
> Is there a third-party tool or service you're integrating with that needs SHA1 specifically?

Yes, i want to integrate stash with a 3rd party service that uses sha1

Gykes:
Then wouldn't this be better as a plugin or some sort of external support? I'm not the main dev so dont take my word as law but seems odd to include another hasher to support a 3rd party usage.

hyper440:
i just thought it would not hurt to add sha1 as an option. I'm currently using plugin for my needs.

stg-annon:
you can add 3rd party hashes to a file via the API without having to integrate the hash into the main codebase see [SHALookup](https://github.com/FansDB/metadata-scrapers/tree/main/scrapers/Coomer/SHALookup) / [FileSetFingerprints](https://github.com/stg-annon/stashapi/blob/main/src/stashapi/stashapp.py#L411C9-L422)  as an example

PR #6566, opened 2026-02-10
Performer image rewrite

### LINKED ISSUE #4257: [Feature] Transparent default/blank performer images
While I was working on a theme, I noticed that the blank performer images are themed around Stash's default theme. These could probably be changed to have transparent backgrounds, with the silhouette being a brightened-but-still-transparent overlay.

I checked on other default images like tags, and the only other forced background I noticed was the front cover for movies having a black background.

feederbox826:
Convert all placeholder performer images to SVGs (where possible) or PNGs if not possible
closes https://github.com/stashapp/stash/issues/4257

The original images from https://github.com/stashapp/stash/pull/109#issuecomment-524375798 were not all used under the proper licence. As a result, the following images were removed. Both ShutterStock and Adobe Stock prohibit the distribution of the source file

- NoName01 / NoName16 at [ShutterStock](https://www.shutterstock.com/image-vector/sexy-silhouettes-67804813)
- NoName15 is the flipped version of NoName01 
- NoName03 at [Adobe Stock](https://stock.adobe.com/images/silhouette-of-a-young-woman/1829808958)
- NoName08 at [ShutterStock](https://www.shutterstock.com/image-vector/silhouette-drawing-cowgirl-hat-whip-pistol-114518521)
- NoName18 at [ShutterStock](https://www.shutterstock.com/image-illustration/one-woman-silhouette-pink-hair-without-34462426)
- NoName40 at [ShutterStock](https://www.shutterstock.com/image-vector/set-naked-angel-devil-casual-women-49668595)

The following files could not (to the best of my ability) have their canonical sources located and were excluded
- NoName04
- NoName10
- NoName20


WithoutPants (approved):
Thanks for the effort here. I think I'll miss the more subtle colouring of the old silhouettes, but it's not a showstopper or anything, and I can likely adjust the colouring with a css snippet.

feederbox826:
should the PNGs be vectorized? I didnt think there was any need since it would just make the edges harsher and it's not like it'll be scaled up enough to see the difference

feederbox826:
> I think I'll miss the more subtle colouring of the old silhouettes, but it's not a showstopper or anything, and I can likely adjust the colouring with a css snippet.

Yeah I wanted to match the new sfw performer, having a background on the CSS kinda defeats the point of it being a vector as well. For the missing ones I would just recommend people to throw them back in the custom performer images, even though there were SVGs that were "obtainable" it was in stark contrast to the male images

WithoutPants:
> should the PNGs be vectorized? I didnt think there was any need since it would just make the edges harsher and it's not like it'll be scaled up enough to see the difference

It would be preferable to have them all of the same file type, but not imperative I guess.

feederbox826:
OK. I probably won't vectorize it then, it might actually increase the file size and i've seen a decrease in quality because of it when I've tried :/

PR #6570, opened 2026-02-12
Enhance backup and download warning message

Servbot91:
Added a warning to the backup and download description regarding database storage type verification. Since I was the sorry sap who ran into the issue, I thought it'd be a good idea to warn the next guy.

### DIFF
--- a/ui/v2.5/src/locales/en-GB.json
+++ b/ui/v2.5/src/locales/en-GB.json
@@ -496,7 +496,7 @@
       "auto_tag_based_on_filenames": "Auto tag content based on file paths.",
       "auto_tagging": "Auto tagging",
       "backing_up_database": "Backing up database",
-      "backup_and_download": "Performs a backup of the database and downloads the resulting file. WARNING: You must verify your Binary Data Storage Type within System -> Database or you risk an incomplete backup. The Database storage type must be selected and a blob migration must be performed for a full Database backup." ,
+      "backup_and_download": "Performs a backup of the database and downloads the resulting file. WARNING: Blobs must be backed up independently if binary data is stored outside the database file. See System > Database for more details." ,
       "backup_database": "Performs a backup of the database to the backups directory, with the filename format {filename_format}.",
       "cleanup_desc": "Check for missing files and remove them from the database. This is a destructive action.",
       "clean_generated": {
--- a/ui/v2.5/src/locales/en-GB.json
+++ b/ui/v2.5/src/locales/en-GB.json
@@ -496,7 +496,7 @@
       "auto_tag_based_on_filenames": "Auto tag content based on file paths.",
       "auto_tagging": "Auto tagging",
       "backing_up_database": "Backing up database",
-      "backup_and_download": "Performs a backup of the database and downloads the resulting file.",
+      "backup_and_download": "Performs a backup of the database and downloads the resulting file. WARNING: You must verify your Binary Data Storage Type within System -> Database or you risk an incomplete backup. The Database storage type must be selected and a blob migration must be performed for a full Database backup." ,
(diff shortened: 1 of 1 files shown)


feederbox826:
I think the warning message is actually quite unclear, and only adding it to one type just adds to the confusion. IMO much better targeted at https://github.com/stashapp/Stash-Docs/blob/main/docs/guides/backup-and-restore-database.md

Servbot91:
> I think the warning message is actually quite unclear, and only adding it to one type just adds to the confusion. IMO much better targeted at https://github.com/stashapp/Stash-Docs/blob/main/docs/guides/backup-and-restore-database.md

Id argue the opposite. Referring to a doc outside of the UI doesnt solve the issue and keeps the same problem of the user thinking they have a full db backup when downloading the file. The point is to bring attention to the storage types. In its current form a user must know or read documentation surrounding this to avoid making an incomplete backup. This removes that assumption. 

This warning in the UI at the very least makes the user ask the question, what is system storage types? what is this setting they are referring to? If they were to read the warning and visit the mentioned data type setting, it definitely makes alot more sense which is why I only mentioned one setting. Follow the work flow of the warning onn your own and it should make more sense to you. Put your mind in the end user. Take backup, see warning, investigate setting. But a warning 100% absolutely should be in the UI as I went through this myself.

It is extremely bold and assumptive to think that your average user is going to go mindfully find and read the doc.  Most users,  in any space, shoot first and ask questions later. This will definitely pump the brakes a bit.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -496,7 +496,7 @@
       "auto_tag_based_on_filenames": "Auto tag content based on file paths.",
       "auto_tagging": "Auto tagging",
       "backing_up_database": "Backing up database",
-      "backup_and_download": "Performs a backup of the database and downloads the resulting file.",
+      "backup_and_download": "Performs a backup of the database and downloads the resulting file. WARNING: You must verify your Binary Data Storage Type within System -> Database or you risk an incomplete backup. The Database storage type must be selected and a blob migration must be performed for a full Database backup." ,
```
DogmaDragon: I don't like this wording. Filesystem is the default binary data storage type for all fresh installations and blobs not being there by design doesn't make it incomplete. 

My proposed language would be something like this:

> WARNING: Blobs must be backed up independently if binary data is stored outside the database file. See System > Database for more details.

Servbot91: I think you are assuming too much with a devs eye. At a higher level, me as a end user goes to back up database, even if this is the selection by default. Me, an end user, not knowing any better is completely right to assume a database backup is a full database backup. There needs to be some kind of trigger to give a end user a heads up, at a glance, within the ui, and without having to look something up. Otherwise the end user is going to assume a download backup, is a full backup and that is objectively incorrect.

Even worse feederbox mentions the [documentation](https://github.com/stashapp/Stash-Docs/blob/main/docs/guides/backup-and-restore-database.md) which makes no mention what so ever of filesystem or database storage types. Its just setting so many people up to fail or they do fail, and realize their backup is no good when its already too late.

I do like your wording alot more and I also think its worth updating the [doc](https://github.com/stashapp/Stash-Docs/blob/main/docs/guides/backup-and-restore-database.md)

DogmaDragon: Thanks. I updated the corresponding guide on https://docs.stashapp.cc/guides/backup-and-restore-database/.

WithoutPants:
I think this is better addressed by modifying the behavior of backup instead. 

We should add options to include the filesystem blobs (if applicable) so that it is a valid backup solution. Could also optionally include the generated stuff too but it would significantly increase the resulting size. 

Edit: I can take a look this week.

Servbot91:
Superseded and addressed in #6586.

PR #6573, opened 2026-02-13
Add sidebar to group list

WithoutPants:
As per other sidebar work.

Notable change in this PR is the shifting of the `New` button back to the navbar. The group sub-group view was using the same icon as the create new operation. The navbar remains a not ideal location for the new button, but having it in the filter toolbar causes this inconsistency and it's location in the filter toolbar isn't well positioned in terms of the grouping of like operations.

### DIFF
--- a/ui/v2.5/src/components/Groups/GroupList.tsx
+++ b/ui/v2.5/src/components/Groups/GroupList.tsx
@@ -47,11 +47,6 @@ import { SidebarStudiosFilter } from "../List/Filters/StudiosFilter";
 import { SidebarTagsFilter } from "../List/Filters/TagsFilter";
 import { SidebarRatingFilter } from "../List/Filters/RatingFilter";
 import { Button } from "react-bootstrap";
-import { IListContextState, ListStateContext } from "../List/ListProvider";
-import {
-  FilterStateContext,
-  IFilterContextState,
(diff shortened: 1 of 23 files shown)


ui/v2.5/src/components/Groups/GroupList.tsx
```diff
@@ -11,233 +11,510 @@ import {
...
-          renderContent={renderContent}
-          renderEditDialog={renderEditDialog}
-          renderDeleteDialog={renderDeleteDialog}
-          renderToolbar={renderToolbar}
-        />
-      </GroupListContext>
+      <ListStateContext.Provider value={listContextState}>
+        <FilterStateContext.Provider value={filterStateContext}>
+          <div
+            className={cx("item-list-container gallery-list", {
```
Gykes: Should probably be updated to `group-list`

Gykes (changes requested):
Code review only, no testing:

Imma need you to stop being so good with coding. I only found 1 **_very_** small issue.

ui/v2.5/src/components/Groups/GroupList.tsx
WithoutPants: Good catch, thanks

PR #6584, opened 2026-02-16
Scene custom field backend support

WithoutPants:
Adds custom field support to scenes, in the backend only. Manually tested and unit tests added.


pkg/scene/import.go
```diff
@@ -449,6 +453,14 @@ func (i *Importer) PostImport(ctx context.Context, id int) error {
 		return err
 	}
 
+	if len(i.customFields) > 0 {
+		if err := i.ReaderWriter.SetCustomFields(ctx, id, models.CustomFieldsInput{
+			Full: i.customFields,
+		}); err != nil {
+			return fmt.Errorf("error setting tag custom fields: %v", err)
```
Gykes: should say scene instead of tag

Gykes (commented):
Static review

discourse-stashapp:

https://discourse.stashapp.cc/t/custom-scene-fields-or-private-notes-for-scenes/1691/4

PR #6586, opened 2026-02-17
Include blobs in backup 

WithoutPants:
Adds an optional `includeBlobs` flag to `BackupDatabaseInput`. When true, the output of the backup task will be a zip file consisting of the sqlite database file, and the blob files in their respective directories.

Re-formatted the backup task section:

image

Clicking on the `Backup...` button opens a dialog:

image

Note that blobs are included by default.

This is the dialog when download is selected and blobs are de-selected:

image

If the database uses database blobs, then the blobs option is checked and disabled:

image

I've also added a section to the `Tasks` page in the manual to give a little more detail.

Supercedes #6570 (ping @Servbot91)

### DIFF
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -325,6 +325,8 @@ input ImportObjectsInput {
 
 input BackupDatabaseInput {
   download: Boolean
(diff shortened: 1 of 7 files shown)


Servbot91:
Phenomenal work and surgically communicated. I have closed my PR and Branch.

PR #6587, opened 2026-02-17
[Feature] Reveal file in system file manager from file info panel

1509x:
# Scope

Adds a folder icon button next to the path field in the Scene, Image, and Gallery file info panels. Clicking it opens the file's enclosing directory in the system file manager (Finder on macOS, Explorer on Windows, xdg-open on Linux).

The path field previously rendered a `file://` URL link which browsers block for security reasons, making it non-functional. It is now plain text with the reveal button alongside it.

The button is only shown when accessing Stash from localhost, since the feature requires a local GUI file manager. It is hidden when accessed remotely.

The mutations are guarded so they only run if the request came from a loopback address.

Also fixes the existing `revealInFileManager` platform implementations which were constructing `exec.Command` but never calling `.Run()`, making them silent no-ops:

- **darwin**: add `.Run()` to `open -R`
- **windows**: add `.Run()` and fix flag to `/select,` with path as a separate argument
- **linux**: implement using `xdg-open` on the parent directory (was an empty stub)
- **desktop.go**: thread `os.FileInfo` through to platform implementations to avoid a redundant `os.Stat` call on Linux

## Closes/Fixes Issues

N/A

## Other testing QA Notes

- Open a scene/image/gallery detail page from localhost and click the folder icon next to the path field — the enclosing directory should open in the system file manager
- The button does not appear when accessing Stash from a non-localhost address
- Linux requires `xdg-open` to be available

### DIFF
--- a/internal/api/resolver_mutation_file.go
+++ b/internal/api/resolver_mutation_file.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/stashapp/stash/internal/desktop"
 	"github.com/stashapp/stash/internal/manager"
 	"github.com/stashapp/stash/pkg/file"
 	"github.com/stashapp/stash/pkg/fsutil"
@@ -324,5 +325,55 @@ func (r *mutationResolver) FileSetFingerprints(ctx context.Context, input FileSe
 		return false, err
 	}
 
+	return true, nil
+}
+
+func (r *mutationResolver) RevealFileInFileManager(ctx context.Context, id string) (bool, error) {
+	fileIDInt, err := strconv.Atoi(id)
+	if err != nil {
+		return false, fmt.Errorf("converting id: %w", err)
+	}
+
+	var filePath string
+	if err := r.withReadTxn(ctx, func(ctx context.Context) error {
+		files, err := r.repository.File.Find(ctx, models.FileID(fileIDInt))
+		if err != nil {
+			return fmt.Errorf("finding file: %w", err)
+		}
+		if len(files) == 0 {
+			return fmt.Errorf("file with id %d not found", fileIDInt)
+		}
+		filePath = files[0].Base().Path
+		return nil
+	}); err != nil {
+		return false, err
+	}
+
+	desktop.RevealInFileManager(filePath)
+	return true, nil
+}
+
+func (r *mutationResolver) RevealFolderInFileManager(ctx context.Context, id string) (bool, error) {
+	folderIDInt, err := strconv.Atoi(id)
+	if err != nil {
+		return false, fmt.Errorf("converting id: %w", err)
+	}
+
+	var folderPath string
+	if err := r.withReadTxn(ctx, func(ctx context.Context) error {
+		folder, err := r.repository.Folder.Find(ctx, models.FolderID(folderIDInt))
+		if err != nil {
+			return fmt.Errorf("finding folder: %w", err)
+		}
+		if folder == nil {
+			return fmt.Errorf("folder with id %d not found", folderIDInt)
+		}
+		folderPath = folder.Path
+		return nil
+	}); err != nil {
+		return false, err
+	}
+
+	desktop.RevealInFileManager(folderPath)
 	return true, nil
 }
--- a/internal/api/resolver_mutation_file.go
+++ b/internal/api/resolver_mutation_file.go
@@ -9,7 +9,9 @@ import (
 	"github.com/stashapp/stash/internal/manager"
 	"github.com/stashapp/stash/pkg/file"
 	"github.com/stashapp/stash/pkg/fsutil"
+	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/models"
+	"github.com/stashapp/stash/pkg/session"
 	"github.com/stashapp/stash/pkg/sliceutil/stringslice"
 )
 
@@ -329,6 +331,12 @@ func (r *mutationResolver) FileSetFingerprints(ctx context.Context, input FileSe
 }
 
 func (r *mutationResolver) RevealFileInFileManager(ctx context.Context, id string) (bool, error) {
+	// disallow if request did not come from localhost
+	if !session.IsLocalRequest(ctx) {
+		logger.Warnf("Attempt to reveal file in file manager from non-local request")
+		return false, fmt.Errorf("access denied")
+	}
+
 	fileIDInt, err := strconv.Atoi(id)
 	if err != nil {
 		return false, fmt.Errorf("converting id: %w", err)
@@ -354,6 +362,12 @@ func (r *mutationResolver) RevealFileInFileManager(ctx context.Context, id strin
 }
 
 func (r *mutationResolver) RevealFolderInFileManager(ctx context.Context, id string) (bool, error) {
+	// disallow if request did not come from localhost
+	if !session.IsLocalRequest(ctx) {
+		logger.Warnf("Attempt to reveal folder in file manager from non-local request")
+		return false, fmt.Errorf("access denied")
+	}
+
 	folderIDInt, err := strconv.Atoi(id)
 	if err != nil {
 		return false, fmt.Errorf("converting id: %w", err)
--- a/internal/api/resolver_mutation_file.go
+++ b/internal/api/resolver_mutation_file.go
@@ -357,7 +357,10 @@ func (r *mutationResolver) RevealFileInFileManager(ctx context.Context, id strin
 		return false, err
 	}
 
-	desktop.RevealInFileManager(filePath)
+	if err := desktop.RevealInFileManager(filePath); err != nil {
+		return false, err
+	}
+
 	return true, nil
 }
 
@@ -388,6 +391,9 @@ func (r *mutationResolver) RevealFolderInFileManager(ctx context.Context, id str
 		return false, err
 	}
 
-	desktop.RevealInFileManager(folderPath)
+	if err := desktop.RevealInFileManager(folderPath); err != nil {
+		return false, err
+	}
+
 	return true, nil
 }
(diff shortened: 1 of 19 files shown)


internal/api/resolver_mutation_file.go
```diff
@@ -326,3 +327,53 @@ func (r *mutationResolver) FileSetFingerprints(ctx context.Context, input FileSe
...
+			return fmt.Errorf("file with id %d not found", fileIDInt)
+		}
+		filePath = files[0].Base().Path
+		return nil
+	}); err != nil {
+		return false, err
+	}
+
+	desktop.RevealInFileManager(filePath)
+	return true, nil
```
Gykes, line 364: The mutation always returns (true, nil) regardless of whether the reveal actually did anything. When `IsDesktop()` is false (Docker, service mode, terminal), `RevealInFileManager` silently no-ops. Same when `os.Stat` fails or the platform `exec.Command` errors, all logged server-side only.

Should probably return an error or a bool so that the UI can handle the button on the fly depending on OS

internal/api/resolver_mutation_file.go
```diff
@@ -326,3 +327,53 @@ func (r *mutationResolver) FileSetFingerprints(ctx context.Context, input FileSe
...
+			return fmt.Errorf("folder with id %d not found", folderIDInt)
+		}
+		folderPath = folder.Path
+		return nil
+	}); err != nil {
+		return false, err
+	}
+
+	desktop.RevealInFileManager(folderPath)
+	return true, nil
```
Gykes, line 398: same as line 353

ui/v2.5/src/components/Images/ImageDetails/ImageFileInfoPanel.tsx
```diff
@@ -47,12 +52,18 @@ const FileInfoPanel: React.FC<IFileInfoPanelProps> = (
...
-          url={`file://${props.file.path}`}
-          value={`file://${props.file.path}`}
-          truncate
-        />
+        <TextField id="path">
+          <span className="d-flex align-items-center">
+            <TruncatedText text={props.file.path} />
+            <Button
+              className="minimal ml-1"
+              title="Reveal in file manager"
```
Gykes: This should be added to i18n instead of being hardcoded

ui/v2.5/src/components/Galleries/GalleryDetails/GalleryFileInfoPanel.tsx
```diff
@@ -38,12 +44,24 @@ const FileInfoPanel: React.FC<IFileInfoPanelProps> = (
...
-          url={`file://${path}`}
-          value={`file://${path}`}
-          truncate
-        />
+        <TextField id={id}>
+          <span className="d-flex align-items-center">
+            <TruncatedText text={path} />
+            <Button
+              className="minimal ml-1"
+              title="Reveal in file manager"
```
Gykes: This should be added to i18n instead of being hardcoded

ui/v2.5/src/components/Scenes/SceneDetails/SceneFileInfoPanel.tsx
```diff
@@ -70,12 +75,18 @@ const FileInfoPanel: React.FC<IFileInfoPanelProps> = (
           truncate
           internal
         />
-        <URLField
-          id="path"
-          url={`file://${props.file.path}`}
-          value={`file://${props.file.path}`}
-          truncate
-        />
+        <TextField id="path">
```
Gykes, line 74: Since this is the exact same on `ImageFileInfo` and `GalleryFileInfo` it might be a good idea to move these to a shared component to reduce code duplication. Not 100% needed imo but main dev might want it done.

ui/v2.5/src/components/Scenes/SceneDetails/SceneFileInfoPanel.tsx
```diff
@@ -70,12 +75,18 @@ const FileInfoPanel: React.FC<IFileInfoPanelProps> = (
...
+          <span className="d-flex align-items-center">
+            <TruncatedText text={props.file.path} />
+            <Button
+              className="minimal ml-1"
+              title="Reveal in file manager"
```
WithoutPants: Needs internationalisation

internal/desktop/desktop.go
```diff
@@ -156,14 +156,14 @@ func getIconPath() string {
...
 func RevealInFileManager(path string) {
-	exists, err := fsutil.FileExists(path)
-	if err != nil {
-		logger.Errorf("Error checking file: %s", err)
+	if !IsDesktop() {
 		return
 	}
-	if exists && IsDesktop() {
-		revealInFileManager(path)
+	if _, err := os.Stat(path); err != nil {
```
Gykes: `os.Stat` is called here to check the path exists, and then called again inside `revealInFileManager` on Linux to check if it's a file or directory. Could reuse the result from here instead of checking twice.

Not a large issue though

Gykes (changes requested):
Not the main dev, just making notes. I like the idea of this :)

internal/desktop/desktop.go
WithoutPants: Agree that we may as well. Change `revealInFileManager` to also accept `info`.

internal/api/resolver_mutation_file.go
```diff
@@ -326,3 +327,53 @@ func (r *mutationResolver) FileSetFingerprints(ctx context.Context, input FileSe
 
 	return true, nil
 }
+
+func (r *mutationResolver) RevealFileInFileManager(ctx context.Context, id string) (bool, error) {
```
WithoutPants, line 333: I'd like to see these mutations guarded so that they only run if the request came from a loopback address. I've coded up a middleware to set a flag if the request comes from a loopback address, which I've put in the `session` package. I added a call to `r = session.SetLocalRequest(r)` in `authenticateHandler` in `api/authentication.go`. The code for the handler is below:

```
package session

import (
	"context"
	"net"
	"net/http"

	"github.com/stashapp/stash/pkg/logger"
)

// SetLocalRequest checks if the request is from localhost and sets the context value accordingly.
// It returns the modified request with the updated context, or the original request if it did
// not come from localhost or if there was an error parsing the remote address.
func SetLocalRequest(r *http.Request) *http.Request {
	// determine if request is from localhost
	host, _, err := net.SplitHostPort(r.RemoteAddr)
	if err != nil {
		logger.Errorf("Error parsing remote address: %v", err)
		return r
	}

	ip := net.ParseIP(host)
	if ip == nil {
		logger.Errorf("Error parsing IP address: %s", host)
		return r
	}

	if ip.IsLoopback() {
		ctx := context.WithValue(r.Context(), contextLocalRequest, true)
		r = r.WithContext(ctx)
	}

	return r
}

// IsLocalRequest returns true if the request is from localhost, as determined by the context value set by SetLocalRequest.
// If the context value is not set, it returns false.
func IsLocalRequest(ctx context.Context) bool {
	val := ctx.Value(contextLocalRequest)
	if val == nil {
		return false
	}
	return val.(bool)
}
```

The mutations can then call `session.IsLocalRequest(ctx)` to determine if the request is local, and if not, the mutation should fail.

internal/desktop/desktop.go
```diff
@@ -156,14 +156,14 @@ func getIconPath() string {
 }
 
 func RevealInFileManager(path string) {
-	exists, err := fsutil.FileExists(path)
-	if err != nil {
-		logger.Errorf("Error checking file: %s", err)
+	if !IsDesktop() {
```
WithoutPants: This is unnecessary if we guard against non-local requests.

internal/desktop/desktop.go
```diff
@@ -156,14 +156,14 @@ func getIconPath() string {
...
 		return
 	}
-	if exists && IsDesktop() {
-		revealInFileManager(path)
+	if _, err := os.Stat(path); err != nil {
+		logger.Errorf("Error checking path: %s", err)
+		return
 	}
+	revealInFileManager(path)
```
WithoutPants: I think the path should be converted to absolute before being passed to `revealInFileManager`. It didn't work on Windows without doing so.

internal/desktop/desktop_platform_windows.go
```diff
@@ -84,5 +84,7 @@ func sendNotification(notificationTitle string, notificationText string) {
 }
 
 func revealInFileManager(path string) {
-	exec.Command(`explorer`, `\select`, path)
+	if err := exec.Command(`explorer`, `/select,`+path).Run(); err != nil {
```
WithoutPants: This needs to be changed to:
```
exec.Command(`explorer`, `/select,`, path)
``` 
(ie move `path` into separate argument instead of concatenating) for it to function correctly. Also as far as I can determine, running this _always_ returns an error, so we should document this but don't bother checking for error or logging anything.

ui/v2.5/src/components/Scenes/SceneDetails/SceneFileInfoPanel.tsx
WithoutPants: I agree, but I think we should consolidate the `Button` component here into a `RevealInFilesystemButton` component. I think that it should also include logic to be hidden if `getPlatformURL` returns a host that is _not_ `localhost`, so that it only shows if it can actually function correctly.

WithoutPants (changes requested):
Good ideas here. This needs a bit of refinement. I can push some changes to your branch if you prefer, as I've made some changes locally.

I've only tested this on Windows - ideally we'd want successful tests from the other platforms.

1509x:
@WithoutPants Yes please push your changes, happy to build on top of them for the remaining items (i18n, shared component, double-stat cleanup).

WithoutPants:
I've pushed my changes and resolved the applicable review notes.

WithoutPants (approved):

discourse-stashapp:

https://discourse.stashapp.cc/t/filepath-copy/1598/3

discourse-stashapp:

https://discourse.stashapp.cc/t/stash-open/1607/3

PR #6593, opened 2026-02-19
Add group scene count filter

WithoutPants:
Adds group scene count filter criterion


discourse-stashapp:

https://discourse.stashapp.cc/t/add-the-ability-to-filter-groups-by-scene-count/5751/3

PR #6601, opened 2026-02-24
FR: Custom Fields Frontend

Gykes:
Adds frontend support for custom fields on scenes, Images, Galleries, Studios, Groups, and tags. 

For transparency, I extracted the helper, did scenes, then had AI copy it all over to the rest of the objects. Cut my copy and pasting time down significantly which was cool.


ui/v2.5/graphql/data/scene-slim.graphql
DogmaDragon, line 1: Forgive my ignorance, but wouldn't this potentially increase the Scenes page loading time? And is it intentionally limited to only scenes object?

Gykes: Good catch, I was doing testing with it and forgot to remove it from the scene-slim. 

It has been removed.

WithoutPants (approved):
Thanks for this. I did some polishing up - made the styling consistent across object types and fixed minor issues. This is now looking good.

discourse-stashapp:

https://discourse.stashapp.cc/t/custom-scene-fields-or-private-notes-for-scenes/1691/4

PR #6604, opened 2026-02-25
Show unsupported filter criteria in filter tags

WithoutPants:
Alternative to #6602.

Shows a warning coloured filter tag, with warning icon and text "<type> (unsupported) ...". 

Cannot be edited, can only be removed. Won't be saved to saved filters.

Screenshot with `career_length` criterion:
image

Happy to adjust presentation based on feedback.


WithoutPants:
Added warning popover to front page content since that's where this issue was reported first. Recommendation rows with unsupported criteria will show a warning icon with the popover indicating what is not supported:

image

PR #6609, opened 2026-02-26
Update minimatch and ajv dependencies

WithoutPants:
Supercedes #6595
Supercedes #6605 
Dependabot doesn't seem to be updating the lock file correctly.

### DIFF
--- a/ui/v2.5/pnpm-lock.yaml
+++ b/ui/v2.5/pnpm-lock.yaml
@@ -2212,11 +2212,11 @@ packages:
     peerDependencies:
       ajv: ^6.9.1
 
-  ajv@6.12.6:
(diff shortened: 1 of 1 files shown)


feederbox826:
ajv is a major version change and it targets validation [and won't merge cleanly in CommunityScrapers](https://github.com/stashapp/CommunityScrapers/pull/2667), do you mind if I have a stab at it? updating just lockfile will just create a broken state

WithoutPants:
I tested the UI briefly and didn't notice any issues after upgrading. Note that it upgrades different ajv versions, `ajv@6.12.6 -> 6.14.0` and `ajv@8.17.1 -> 8.18.0` so as far as I can determine it's not doing a major version upgrade.

feederbox826:
ah I thought it was 6 -> 8, the multiple versions under different dependencies threw me off since scrapers is 7 -> 8

PR #6620, opened 2026-02-28
FR: Allow Tag Tagger to Autocreate Parent (Category) Tag

Gykes:
Currently, the studio tagger has the ability to auto create and connect the parent studio for a newly tagged studio. I am adding this into the tags tagger as well. Users will be able to select the toggle, exact same as Studio, and create the parent tag (category) for their newly tagged tag. Currently I do not believe that Categories have StashIDs similar to how tags do it so it is a name search only as of now unless someone corrects me.

UI: 
Screenshot 2026-02-27 at 17 15 30

Auto create link:
Screenshot 2026-02-27 at 17 16 27

### DIFF
--- a/graphql/schema/types/scraper.graphql
+++ b/graphql/schema/types/scraper.graphql
@@ -73,6 +73,7 @@ type ScrapedTag {
   name: String!
   description: String
   alias_list: [String!]
+  parent: ScrapedTag
   "Remote site ID, if applicable"
   remote_site_id: String
 }
--- a/graphql/stash-box/query.graphql
+++ b/graphql/stash-box/query.graphql
@@ -31,6 +31,11 @@ fragment TagFragment on Tag {
   id
   description
   aliases
+  category {
+    id
+    name
+    description
+  }
 }
 
 fragment MeasurementsFragment on Measurements {
--- a/internal/manager/manager_tasks.go
+++ b/internal/manager/manager_tasks.go
@@ -431,7 +431,7 @@ type StashBoxBatchTagInput struct {
 	ExcludeFields []string `json:"exclude_fields"`
 	// Refresh items already tagged by StashBox if true. Only tag items with no StashBox tagging if false
 	Refresh bool `json:"refresh"`
-	// If batch adding studios, should their parent studios also be created?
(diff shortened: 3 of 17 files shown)


feederbox826:
Categories have UUIDs https://stashdb.org/categories/939c769b-7c41-4a04-983e-5a30a8c949ea

but they do not operate in the same namespace as tags, adding them as stashids result in bad redirects

Gykes:
> Categories have UUIDs https://stashdb.org/categories/939c769b-7c41-4a04-983e-5a30a8c949ea
> 
> 
> 
> but they do not operate in the same namespace as tags, adding them as stashids result in bad redirects

Yep, that was the first thing I tried. I could expand it to pull the category uuid and utilize it but will be a tomorrow problem if we do it that way.

feederbox826:
it won't work since the stashid type is linked to tags

it goes endpoint/type/id

so it will only link to /tag/id

WithoutPants:
This looks a bit unfinished. There's no option to include/exclude parent tag in the exluded tags field, and using the tags tagger to update a tag doesn't include the parent category tag at all. Was the new behaviour only included for the batch operations?

Gykes:
> This looks a bit unfinished. There's no option to include/exclude parent tag in the exluded tags field, and using the tags tagger to update a tag doesn't include the parent category tag at all. Was the new behaviour only included for the batch operations?

Sorry m8, this is 100% not ready. I must have fat fingered the request button before closing out. I got caught up the the PR stuff and haven't got back to this yet.

discourse-stashapp:

https://discourse.stashapp.cc/t/stash-v0-31-release-development-thread/4881/2

Gykes:
Finished this up, seems to work as expected. 

Created various tags, used exclusion list, created a parent that already exists (failed) so modified to check for parent, all worked.

One thing I dont really like is how im currently searching for if a parent tag already exists. Perhaprs someone can help me find a better way. 

Also I think, in a different PR, these tagger modals can be updated to extract some stuff into shared components. There is a bit of duplication across them all

WithoutPants (approved):

PR #6621, opened 2026-02-28
Add Selective generate

### LINKED ISSUE #3287: [Feature] Selective Generate
I'd like a feature similar to Selective Scan and Selective Clean but for the Generate task.
Scanning does not generate supporting files for videos that are already present in the database. 
When only one folder has changed significantly, generating the supporting files for the entire database from scratch can be very time-consuming.

I'd therefore suggest an option to preform the selected generation tasks only in a specific folder and not the entire collection.
Thank you!

puc9:
As the title says, I've added the `Selective generate...` button and it functions the same as the other `Selective` buttons.

image

This closes #3287 but not its sub issue. I've built this for myself and only now, when creating the PR, I noticed the feature request.

The proposed workaround in issue #3287 is not ideal, IMHO, when you have considerably more than 1000 items that you wish to run generate on, so I think that the feature is still useful as implemented.

For example, you scan for new files and only generate scene and image phashes to quickly identify dups and only then you run the slower generation on the same folders.

### DIFF
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -26,7 +26,7 @@ input GenerateMetadataInput {
   imageIDs: [ID!]
   "gallery ids to generate for"
   galleryIDs: [ID!]
-
+  "paths to run generate on, in addition to the other ID lists"
   paths: [String!]
 
   "overwrite existing media"
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -29,7 +29,6 @@ input GenerateMetadataInput {
 
   paths: [String!]
 
-
   "overwrite existing media"
   overwrite: Boolean
 }
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -27,6 +27,9 @@ input GenerateMetadataInput {
   "gallery ids to generate for"
   galleryIDs: [ID!]
 
+  paths: [String!]
+
+
   "overwrite existing media"
   overwrite: Boolean
 }
--- a/internal/manager/task_generate.go
+++ b/internal/manager/task_generate.go
@@ -65,7 +65,6 @@ const generateQueueSize = 200000
 type GenerateJob struct {
 	repository models.Repository
 	input      GenerateMetadataInput
-	paths      []string
 
 	overwrite      bool
 	fileNamingAlgo models.HashAlgorithm
@@ -103,14 +102,6 @@ func (j *GenerateJob) Execute(ctx context.Context, progress *job.Progress) error
 
 	logger.Infof("Generate started with %d parallel tasks", parallelTasks)
 
-	if len(j.input.Paths) > 0 {
-		sp := getScanPaths(j.input.Paths)
-		j.paths = make([]string, len(sp))
-		for i, p := range sp {
-			j.paths[i] = p.Path
-		}
-	}
-
 	queue := make(chan Task, generateQueueSize)
 	go func() {
 		defer close(queue)
@@ -144,8 +135,13 @@ func (j *GenerateJob) Execute(ctx context.Context, progress *job.Progress) error
 		r := j.repository
 		if err := r.WithReadTxn(ctx, func(ctx context.Context) error {
 			qb := r.Scene
-			if len(j.input.SceneIDs) == 0 && len(j.input.MarkerIDs) == 0 && len(j.input.ImageIDs) == 0 && len(j.input.GalleryIDs) == 0 {
-				j.queueTasks(ctx, g, queue)
+			if len(j.input.SceneIDs) == 0 &&
+				len(j.input.MarkerIDs) == 0 &&
+				len(j.input.ImageIDs) == 0 &&
+				len(j.input.GalleryIDs) == 0 &&
+				len(j.input.Paths) == 0 {
+
+				j.queueTasks(ctx, g, nil, queue)
 			} else {
 				if len(j.input.SceneIDs) > 0 {
 					scenes, err = qb.FindMany(ctx, sceneIDs)
@@ -196,7 +192,8 @@ func (j *GenerateJob) Execute(ctx context.Context, progress *job.Progress) error
 				}
 
 				if len(j.input.Paths) > 0 {
-					j.queueTasks(ctx, g, queue)
+					paths := filterStashPaths(j.input.Paths)
+					j.queueTasks(ctx, g, paths, queue)
 				}
 			}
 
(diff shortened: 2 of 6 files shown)


graphql/schema/types/metadata.graphql
```diff
@@ -27,6 +27,8 @@ input GenerateMetadataInput {
   "gallery ids to generate for"
   galleryIDs: [ID!]
 
+  paths: [String!]
```
Gykes, line 30: Small, but I would add a comment here like the other fields on this page.

internal/manager/task_generate.go
```diff
@@ -43,6 +43,8 @@ type GenerateMetadataInput struct {
 	GalleryIDs []string `json:"galleryIDs"`
 	// overwrite existing media
 	Overwrite bool `json:"overwrite"`
+	// paths
```
Gykes: I would also change this comment up. `// paths` is somewhat unhelpful. Maybe `// paths to restrict generation to` or something like that.

internal/manager/task_generate.go
```diff
@@ -100,6 +103,12 @@ func (j *GenerateJob) Execute(ctx context.Context, progress *job.Progress) error
 
 	logger.Infof("Generate started with %d parallel tasks", parallelTasks)
 
+	sp := getScanPaths(j.input.Paths)
+	j.paths = make([]string, len(sp))
+	for i, p := range sp {
+		j.paths[i] = p.Path
+	}
```
Gykes: This always populates `j.paths`, even when `j.input.Paths` is empty, because `getScanPaths([])` returns all configured stash paths. That means `queueScenesTasks` and `queueImagesTasks` will always apply a path filter via `FilterFromPaths(j.paths)`, whereas previously they passed nil (no filter). This changes the normal "Generate All" behavior for scenes/images outside current stash paths would be silently skipped.

This should probably be modified so that it's guarded and only applied when paths are explicitly provided.

internal/manager/task_generate.go
```diff
@@ -183,6 +192,10 @@ func (j *GenerateJob) Execute(ctx context.Context, progress *job.Progress) error
 						}
 					}
 				}
+
+				if len(j.input.Paths) > 0 {
+					j.queueTasks(ctx, g, queue)
+				}
```
Gykes, line 197: Isn't this redundant? If both specific IDs and paths are provided, this would queue items by ID AND by path, potentially double-processing some items. This block shouldn't be needed, the path filtering already happens through `queueTasks` in the if branch above.

Gykes (changes requested):
Not the main dev, just noting some things. 

Edit: Static review only, no actual testing

internal/manager/task_generate.go
puc9: You can pass nothing to Generate, only paths, or you can also pass a combination of ids and paths. That's the purpose of the last conditional, I don't think is redundant, it allows for all possible combinations of parameters with minimum amount of changes.

graphql/schema/types/metadata.graphql
puc9: I followed how it was defined for the other mutations, as in without a comment. I'll see if I can find one that clarifies.

internal/manager/task_generate.go
Gykes: Fair enough, mostly I was concerned about double processing. The chances are minimal but guarding against it would probably be a good idea

graphql/schema/types/metadata.graphql
Gykes: Not a massive deal. In my opinion all of them should have comments to make it easier for future devs to understand what they are used for.

internal/manager/task_generate.go
puc9: 100% agree with your analysis, I'm just am not sure if it is an issue or not. I modeled my change for the autotag one, as that was the closest in functionality. I don't think that passing nil as before should be a goal here, but to make the code as similar with other parts as possible.

Additionally, if you fallow the calls, if a nil is passed there, eventually the makeQuery would actually create an empty filter: https://github.com/stashapp/stash/blob/c874bd560e24692d11585aff3b1397b09e8d1a4c/pkg/sqlite/scene.go#L975C2-L977C3

I'm obviously not opposed to change it, I just don't think it's necessary, and the maintainer usually adapts PRs if needed, so for now, I'll leave it as is.

Gykes: I understand the `makeQuery` point, both nil and an empty `SceneFilterType{}` are treated the same and produce no path filtering. But the issue isn't about nil vs empty filter. The issue is that `getScanPaths([])` returns all configured stash paths (not an empty slice). So when `j.input.Paths` is empty (normal "Generate"), `FilterFromPaths` receives actual paths and produces a real Path `= "/stash/path/%" OR ...` filter, not an empty one.

This differs from the autotag pattern. Autotag passes `input.Paths` directly to `FilterFromPaths` without going through `getScanPaths`, so when no paths are given, the filter is truly empty. The practical impact is narrow, it would only matter if a user removed a stash path but still has scenes from it in the database. Before this change they'd be included in "Generate All", after they'd be silently skipped. It's another edge case, but it is a behavioral change from the current code.

I'll let the main dev decide how he wants to handle it. Ultimately it's not a massive deal. He's on his weekend and might get to this next week

Gykes (approved):
Thanks for the work you put in. Looks good overall outside of what I noted in the comments. It's an edge case and I don't think it will break anything or cause issues. 

WP: I like the feature, it's been requested a decent amount, you can check comments to see how you would like to handle the potential edge case (Unless i'm way off base with how im reading the code)

internal/manager/task_generate.go
puc9: Hmm, I think I misunderstood you the first time, I can see now which edge case you are referring to. I'll have another read through the code.

internal/manager/task_generate.go
```diff
@@ -63,6 +65,7 @@ const generateQueueSize = 200000
 type GenerateJob struct {
 	repository models.Repository
 	input      GenerateMetadataInput
+	paths      []string
```
WithoutPants: This should not be necessary. Paths can be handled with local variables in `Execute`.

internal/manager/task_generate.go
WithoutPants: We should add a check for `len(j.input.Paths) == 0` here.

internal/manager/task_generate.go
WithoutPants: This should be changed to perform the processing of `j.Input.Paths` here rather than further up. 

I think `getScanPaths` doesn't quite fit what we want to do with these paths (filter out paths that are outside configured libraries), so I think we should just create a new function that does the filtering instead.

Processed paths should be sent to the `queueTasks` function.

internal/manager/task_generate.go
WithoutPants: Add paths string slice as input here and add to `queueScenesTasks` and `queueImagesTasks`.

WithoutPants: Add paths string slice as input here.

WithoutPants: Add paths string slice as input here.

internal/manager/task_generate.go
WithoutPants: Agree with Gykes that this should be changed. See other review comments.

WithoutPants (changes requested):

internal/manager/task_generate.go
puc9: Oh boy, how my reading comprehension is failing me... 👎 you are obviously right, I missed the part about the `getScanPaths`. I'll add a fix.

puc9: @WithoutPants I missed your comments from earlier, I'm reading through them right now.

internal/manager/task_generate.go
puc9: @Gykes after reading a bit more trough the code I agree that if a scene is passed in the input both as ID and from within one of the paths it will be processed twice, but I do not think that is easily fixable without building an index of which scenes/images have already been added to the queue. That looks like a larger change for limited benefit IMHO.
Additionally, that could not happen through regular UI, so I think it would be a rare occurrence. I've addressed all other comments, I hope.

WithoutPants (approved):

PR #6622, opened 2026-02-28
Fix stale browser-cached thumbnails after file content changes during scan.

XTVB:
### Problem

When a file's content changes (e.g. renaming files within a gallery zip), the scan handler updates fingerprints but never bumps the entity's `updated_at` timestamp. Since thumbnail URLs use `updated_at` as a cache buster and are served with
`immutable`/1-year cache headers, browsers indefinitely serve the old cached thumbnail.

### Solution

In the `associateExisting` method for scenes, images, and galleries, `UpdatePartial` was previously only called when a **new file association** was created (`!found`). This change extends it to also run when `updateExisting` is true (i.e.
`oldFile != nil` — the file existed before but its content changed). This bumps `updated_at`, invalidating the browser cache.

The fix restructures the existing conditional blocks so that `UpdatePartial` and the post-update plugin hook share the same `!found || updateExisting` guard, rather than having `UpdatePartial` nested only inside the `!found` branch.

### Testing

Added unit tests for `associateExisting` across all three packages (`pkg/scene`, `pkg/image`, `pkg/gallery`) covering:

- **Content change path** — `UpdatePartial` is called when the file is already associated but content has changed
- **No-op path** — `UpdatePartial` is *not* called when the file is already associated and unchanged
- **New file path** — `UpdatePartial` is called when a previously unknown file is added

Tests use the existing `mocks.Database` infrastructure. Added a `WithTxnCtx` helper to `mocks.Database` to provide a transaction context for code that registers plugin hooks.

### DIFF
--- a/pkg/gallery/scan_test.go
+++ b/pkg/gallery/scan_test.go
@@ -2,67 +2,14 @@ package gallery
 
 import (
 	"context"
-	"errors"
 	"testing"
 
 	"github.com/stashapp/stash/pkg/models"
+	"github.com/stashapp/stash/pkg/models/mocks"
 	"github.com/stashapp/stash/pkg/plugin"
-	"github.com/stashapp/stash/pkg/txn"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
-
-var errRollback = errors.New("rollback")
-
-type mockGalleryScanCU struct {
-	mock.Mock
-}
-
-func (m *mockGalleryScanCU) FindByFileID(ctx context.Context, fileID models.FileID) ([]*models.Gallery, error) {
-	args := m.Called(ctx, fileID)
-	return args.Get(0).([]*models.Gallery), args.Error(1)
-}
-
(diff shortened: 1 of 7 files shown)


pkg/gallery/scan_test.go
```diff
@@ -0,0 +1,161 @@
...
+	return ctx, nil
+}
+func (m *noopTxnManager) Commit(ctx context.Context) error   { return nil }
+func (m *noopTxnManager) Rollback(ctx context.Context) error { return nil }
+func (m *noopTxnManager) IsLocked(err error) bool            { return false }
+
+func withTxnCtx(fn func(ctx context.Context)) {
+	_ = txn.WithTxn(context.Background(), &noopTxnManager{}, func(ctx context.Context) error {
+		fn(ctx)
+		return errRollback
```
Gykes: The only thing that I can think of is that this is duplicated and we already have something existing that does essentially the same thing. You could utilize `mocks.Database` in `pkg/models/mocks/database.go` to reduce duplication. This would need to be updated on all the tests.

XTVB: Moved withTxnCtx to `pkg/models/mocks/database.go` and switched to using the existing mocks

WithoutPants (approved):

PR #6626, opened 2026-03-01
perf: use useMemo to cache performer sorting in Scene Tagger

Joshualover:
This PR fixes performance issue in Scene Tagger when creating performers. The issue was that performer sorting was executed on every render, causing UI to hang. Fixed by wrapping sorting in useMemo.


Gykes:
I would consider this a react "best practice" but, without proper testing, I feel like there would be more of an upstream issue that would cause this. 

Can you tell me the conditions in which you feel the hang? All the time, when x amount of performers are displayed on tagger, etc. Slice and sort is pretty cheap so I have a hard time believing that would cause the whole UI to hang

Edit: Also, do you mean when you click out of the performer create modal to save the performer? I see a roughly 6-7s hang when I hit create. I think the issue would be in the mutations and not in the sort. That's mostly just an educated guess and would need proper testing.

WithoutPants:
Closing as I doubt this is the cause of the hang, no response from OP, and the PR hasn't been formatted correctly.

PR #6629, opened 2026-03-01
Fix edit modal not opening inside gallery view

### LINKED ISSUE #6624: Edit modal fails to open inside gallery
### Describe the bug

The edit modal window fails to open when one or more images inside a gallery are selected.

Verified within troubleshooting mode - no plugin conflicts.

### Steps to reproduce

1. Go to a gallery, i.e. http://localhost:9999/galleries/1
2. Select one or more images
3. Click the `Edit` toolbar button or the keyboard shortcut `ee`

### Expected behaviour

Edit modal should open after step 3, but does not.

_No response_

### Stash version

v0.30.1-98-gc874bd56

### Device details

Chrome 145.0.7632.117 (Official Build) (64-bit) - Windows 11

### Relevant log output

```shell

```

dev-null-life:
## Summary
- Fix the edit/delete modal not appearing when images are selected inside a gallery
- The `{modal}` element was only rendered in the sidebar layout branch, but gallery images use the non-sidebar code path (`View.GalleryImages`), which returned content without rendering the modal to the DOM
- Also wrapped `onEdit`/`onDelete` in `useCallback` and added the missing dependency array to the Mousetrap `useEffect`, matching the pattern used in `SceneList` and other list components

Closes #6624

## Checklist
- [x] Reviewed existing PRs and issues to avoid duplicates
- [x] Code is readable and adequately commented
- [x] Evaluated impact on long-term maintainability
- [x] Minimal change — no unnecessary dependencies or scope creep

## Test plan
- [x] Go to a gallery containing images
- [x] Select one or more images
- [x] Click the Edit toolbar button — modal should now open
- [x] Press `e` keyboard shortcut — modal should now open
- [x] Press `d d` keyboard shortcut — delete dialog should open
- [x] Verify edit/delete modals still work on the standalone Images list page

### DIFF
--- a/ui/v2.5/src/components/Images/ImageList.tsx
+++ b/ui/v2.5/src/components/Images/ImageList.tsx
@@ -587,25 +587,6 @@ export const FilteredImageList = PatchComponent(
       setShowSidebar,
     });
 
-    useEffect(() => {
-      Mousetrap.bind("e", () => {
-        if (hasSelection) {
-          onEdit?.();
-        }
-      });
-
-      Mousetrap.bind("d d", () => {
-        if (hasSelection) {
-          onDelete?.();
-        }
-      });
-
-      return () => {
-        Mousetrap.unbind("e");
-        Mousetrap.unbind("d d");
-      };
-    });
-
     const onCloseEditDelete = useCloseEditDelete({
       closeModal,
       onSelectNone,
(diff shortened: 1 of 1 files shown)


ui/v2.5/src/components/Images/ImageList.tsx
```diff
@@ -787,7 +787,12 @@ export const FilteredImageList = PatchComponent(
     );
 
     if (!withSidebar) {
-      return content;
+      return (
+        <>
+          {modal}
+          {content}
+        </>
+      );
```
Gykes: Rather than rendering `{modal}` in both branches, you may want to render it once above the conditional. This matches how other list components handle it and avoids the duplication that likely caused this bug when the sidebar changes were introduced.

Gykes (changes requested):
Tested locally and worked as expected. The approach is consistent with how other list components in the codebase handle modal rendering and keyboard shortcuts. I left one small change that could be of benifit. Chances are this double modal is why it got forgotten in the sidebar change

ui/v2.5/src/components/Images/ImageList.tsx
dev-null-life: Good call, moved it above the conditional. Replaced the early return with a ternary so it only needs to be rendered once.

WithoutPants (approved):

PR #6632, opened 2026-03-01
Fix UNIQUE constraint failure in folder hierarchy correction

### LINKED ISSUE #6427: After 30.0 upgrade, can no longer run scans because folder.path unique constraint fails
### Describe the bug

After upgrading to 30.0 (not sure from which previous version, I didn't write it down beforehand), attempting to scan my library fails immediately with the following error:

error processing "e:\\content": correcting sub folder hierarchy for "e:\\content": updating folder path f:\content\sorted\yes -> e:\content\yes: updating folders: executing `UPDATE `folders` SET `created_at`=?,`id`=?,`mod_time`=?,`parent_folder_id`=?,`path`=?,`updated_at`=?,`zip_file_id`=? WHERE (`folders`.`id` = ?)` [[2024-12-23T03:36:03+12:00 614 2025-06-24T08:03:21+11:00 581 e:\content\yes 2024-12-23T03:36:03+12:00 <nil> 614]]: UNIQUE constraint failed: folders.path

It says `updating folder path f:\content\sorted\yes -> e:\content\yes`. That's not what should happen. I think this error arises from two things:

1. Drive E: no longer exists, it's F: now. f:\content\sorted\yes should become f:\content\sorted\yes.
2. Folders called "yes" exist at two points in the folder hierarchy: there's `\content\yes` and `\content\sorted\yes`.

It appears that Stash is realizing that #1 has happened but isn't taking #2 into account when updating the path.

Renaming the folders doesn't help.

### Steps to reproduce

N/A

### Expected behaviour

N/A

_No response_

### Stash version

30.0

### Device details

_No response_

### Relevant log output

```shell

```

dev-null-life:
## Summary
- Fixes #6427: `UNIQUE constraint failed: folders.path` when scanning after a folder move (e.g., drive letter change)
- Root cause: `correctSubFolderHierarchy` blindly trusted `parent_folder_id` relationships, so mis-parented children could produce colliding paths
- Adds three guard layers to `correctSubFolderHierarchy`:
  1. **Validate `filepath.Dir(child)` matches old parent path** — skips mis-parented children (root cause)
  2. **Skip children whose path is already correct** — avoids unnecessary DB writes
  3. **Check `FindByPath` before updating** — skips if target path is already taken by another folder

## Test plan
- [x] Added unit tests for `correctSubFolderHierarchy` covering:
  - Basic path update
  - Already-correct path skip
  - Mis-parented child skip
  - Pre-existing target path skip
  - Recursive sub-folder updates
  - Issue #6427 scenario with duplicate basenames at different hierarchy levels
  - Issue #6427 scenario with incorrect `parent_folder_id`
- [ ] Verify scan completes without error after drive letter change with duplicate folder basenames
- [ ] Verify scan completes without error when target folder paths already exist in DB

### DIFF
--- a/pkg/file/move.go
+++ b/pkg/file/move.go
@@ -189,29 +189,55 @@ func (m *Mover) rollback() {
 }
 
 // correctSubFolderHierarchy sets the path of all contained folders to be relative to the given folder.
+// oldParentPath is the folder's path before it was renamed. Children whose filepath.Dir
+// does not match oldParentPath are mis-parented and skipped.
 // It does not move the folder hierarchy in the filesystem.
-func correctSubFolderHierarchy(ctx context.Context, rw models.FolderReaderWriter, folder *models.Folder) error {
+func correctSubFolderHierarchy(ctx context.Context, rw models.FolderReaderWriter, folder *models.Folder, oldParentPath string) error {
 	folders, err := rw.FindByParentFolderID(ctx, folder.ID)
 	if err != nil {
 		return fmt.Errorf("finding contained folders in folder %s: %w", folder.Path, err)
 	}
 
-	folderPath := folder.Path
-
 	for _, f := range folders {
-		oldPath := f.Path
-		folderBasename := filepath.Base(f.Path)
-		correctPath := filepath.Join(folderPath, folderBasename)
+		oldChildPath := f.Path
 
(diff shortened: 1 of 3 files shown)


WithoutPants:
Rejecting this as I don't think it's a good fix for the underlying issue. I've raised #6634 instead.

PR #6633, opened 2026-03-01
Skip scanning zip contents when fingerprint is unchanged

### LINKED ISSUE #6512: scan gallery enclosure instead of children

When scanning galleries in zips, only scan and re-check the md5/oshash of the enclosing zip file (which should change if any of the contents are modified) instead of iterating within the zip

saves a lot of time when a user has a lot of galleries, being able to skip every single file in an enclosed gallery 

No

### Additional context

_No response_

dev-null-life:
## Summary
- When a zip-based gallery's modification time changes but its content fingerprint (oshash/md5) remains the same, skip walking and rescanning every file inside the zip
- Adds a `FingerprintChanged` field to `ScanFileResult` to track whether the file's content hash actually changed during rescan
- Uses the existing `Fingerprints.ContentsChanged()` method to compare old vs new fingerprints before deciding to walk zip contents

## Test plan
- [ ] Scan a library containing zip-based galleries
- [ ] `touch` a zip file (changes modtime but not content) and rescan — contents should be skipped
- [ ] Modify a zip file's contents and rescan — contents should be walked and updated
- [ ] Add a new zip file and scan — contents should be walked as before
- [ ] Force rescan with the rescan option — verify behavior is consistent

Closes #6512

### DIFF
--- a/internal/manager/task_scan.go
+++ b/internal/manager/task_scan.go
@@ -312,7 +312,7 @@ func (j *ScanJob) handleFile(ctx context.Context, f file.ScannedFile, progress *
 	// handle rename should have already handled the contents of the zip file
 	// so shouldn't need to scan it again
 
-	if (r.New || r.FingerprintChanged) && j.scanner.IsZipFile(f.Info.Name()) {
+	if (r.New || r.FingerprintChanged || (r.Updated && j.scanner.Rescan)) && j.scanner.IsZipFile(f.Info.Name()) {
 		ff := r.File
 		f.BaseFile = ff.Base()
 
@@ -324,6 +324,8 @@ func (j *ScanJob) handleFile(ctx context.Context, f file.ScannedFile, progress *
 		if err := j.scanZipFile(zipCtx, f, progress); err != nil {
 			logger.Errorf("Error scanning zip file %q: %v", f.Path, err)
 		}
+	} else if r.Updated && j.scanner.IsZipFile(f.Info.Name()) {
+		logger.Debugf("Skipping zip file scan for %q: fingerprint unchanged", f.Path)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


Gykes:
Can we slow down on these? You are pumping these out very fast with Claude Code with 0 discussion from Dev team on how we want to handle this.

Gykes:
You dont need to close it. Just give time to review and correct before pumping out a bunch more

internal/manager/task_scan.go
```diff
@@ -312,7 +312,7 @@ func (j *ScanJob) handleFile(ctx context.Context, f file.ScannedFile, progress *
 	// handle rename should have already handled the contents of the zip file
 	// so shouldn't need to scan it again
 
-	if (r.New || r.Updated) && j.scanner.IsZipFile(f.Info.Name()) {
+	if (r.New || r.FingerprintChanged) && j.scanner.IsZipFile(f.Info.Name()) {
```
Gykes: This may cause a regression on a force rescan, no? The rescan option exists in the scan options but with this change they'd be skipped if the fingerprint is unchanged.

```if (r.New || r.FingerprintChanged || (r.Updated && j.scanner.Rescan)) && j.scanner.IsZipFile(f.Info.Name())```

Perhaps something like that would get the desired outcome and not cause the regression? My recommendation is untested so please keep me honest and verify.

Gykes: You may also want to add a log in here somewhere so that users know that a zip was skipped and why.

Gykes (changes requested):
Static code review only. No actual testing was completed.

internal/manager/task_scan.go
dev-null-life: You're right, fixed. Added `(r.Updated && j.scanner.Rescan)` to the condition so force rescans still walk zip contents.

dev-null-life: Done, logs at debug level when a zip is skipped.

WithoutPants (approved):

stg-annon:
Would this also apply to Clean tasks as well? not sure what parts are shared if any

PR #6635, opened 2026-03-02
Fix new panic in IsFsPathCaseSensitive: Use filepath operations to check for file system case sensitivity

puc9:
If the path leading to the last component has any multi-byte chars then `IsFsPathCaseSensitive` panics.

Replacing `string` functions with `filepath` functions to address mismatch of indexes.

### DIFF
--- a/pkg/fsutil/fs.go
+++ b/pkg/fsutil/fs.go
@@ -29,7 +29,7 @@ func IsFsPathCaseSensitive(path string) (bool, error) {
 
 	flippedPath := filepath.Join(filepath.Dir(path), fBase)
 
-	fiCase, err := os.Stat(string(flippedPath))
+	fiCase, err := os.Stat(flippedPath)
 	if err != nil { // cannot stat the case flipped path
 		return true, nil // fs of path should be case sensitive
 	}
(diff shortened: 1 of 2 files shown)


pkg/fsutil/fs.go
```diff
@@ -27,18 +26,10 @@ func IsFsPathCaseSensitive(path string) (bool, error) {
...
 
-	flipped := []rune(path)
-	for _, c := range fBase { // replace base of path with the flipped one ( we need to flip the base or last dir part )
-		flipped[i] = c
-		i++
-	}
+	flippedPath := filepath.Join(filepath.Dir(path), fBase)
 
-	fiCase, err := os.Stat(string(flipped))
+	fiCase, err := os.Stat(string(flippedPath))
```
Gykes: Nitpicking: 

Since you are doing `filepath.Join` above it makes the string conversion redundant as it should already be a string.

Gykes: You may also want to edit the test case for this. The test in `pkg/fsutil/fs_test.go` would need an additional case to cover multi-byte characters in the parent path.

Gykes (commented):
Static review only, no actual testing was done. Not the main dev, just noticing things. 


Overall was good, just 1 small nitpick and a test expansion to cover the new potential case.

Gykes (approved):
Thanks for the work, OP 

@WithoutPants LGTM. Pulled it down to verify the new test passes and it does.

PR #6636, opened 2026-03-02
Add folder criteria to scenes, images and galleries and sidebars

WithoutPants:
Adds `folder` criterion to scenes and images filters, and `parent_folder` criterion to galleries filters. Adds these criteria to the sidebar for these objects:

image

Sidebar filter shows top-level folders allowing the user to expand into sub-folders. Specific folder or folder and sub-folders filtering can be selected. Editing criteria in the edit filter dialog allows for excluding folders, and selecting specific depth of sub-folders:

image


ui/v2.5/src/components/List/Filters/SelectableFilter.tsx
```diff
@@ -450,6 +450,45 @@ export const ObjectsFilter = <
   );
 };
 
+interface IHierarchicalObjectsFilter<T extends IHierarchicalLabeledIdCriterion>
+  extends IObjectsFilter<T> {}
```
Gykes: I know Typescript merges duplicate interface declarations. I'm not sure if this was intended or just a copy and paste mistake.  Looks like it is also declared the same way at line 492. 

I doubt this will break anything just noting it just to be safe.

ui/v2.5/src/components/List/Filters/FolderFilter.tsx
```diff
@@ -0,0 +1,554 @@
...
+          {canExclude && (
+            <Button
+              onClick={(e) => {
+                e.stopPropagation();
+                onSelect(folder, true);
+              }}
+              onKeyDown={(e) => e.stopPropagation()}
+              className="minimal exclude-button"
+            >
+              <span className="exclude-button-text">exclude</span>
```
Gykes: Internationalize

ui/v2.5/src/components/List/Filters/FolderFilter.tsx
```diff
@@ -0,0 +1,554 @@
...
+  criterion,
+  setCriterion,
+}) => {
+  const intl = useIntl();
+  const [query, setQuery] = useState("");
+  const [displayQuery, onQueryChange] = useDebouncedState(query, setQuery, 250);
+
+  const messages = defineMessages({
+    sub_folder_depth: {
+      id: "sub_folder_depth",
```
Gykes, line 329: I don't see this defined in en-GB.json

Gykes (changes requested):
Pulled it down and did some poking around. Seemed to work as expected. 

One thing that may be a feature/benefit is being able to type in/copy filepaths into the search bar and click enter on it. Im able to type in my path, for example, `/User/gykes/Desktop/AVIF` and it pulls it up in the filter list of paths but I have to then click on the AVIF folder. Would be nice to just be able to hit enter and have it add it to the list of paths.

WithoutPants:
Thanks for the review. I added the enter-to-select functionality to the folder select and to the other selection filters - performers, tags, studios.

PR #6637, opened 2026-03-03
New: Add From Clipboard to Set Image

### LINKED ISSUE #3416: [Bug Report] Short Form Subject (50 Chars or less)
**Describe the bug**
Adding a Studio Code to an existing scene in stash generates an image of that code and replaces existing cover image

**To Reproduce**
1. Go to existing scene
2. Scroll down to cover image to see existing
3. Enter Studio Code
4. Scroll down to cover image to see new image
5. See Error

**Expected behavior**
To be able to add a studio code WITHOUT it replacing existing cover image

**Screenshots**
StudioCode
StudioCodeCoverImage

**Stash Version: v0.18.0-49-gcb808c7b**

 - OS: Windows 10 Pro 22H2
 - Browser: Chrome
 - Version: 109.0.5414.120

### LINKED ISSUE #6625: Set image from clipboard (esp. performers)

Currently, I can set an image from a file or from a URL.

But especially for performers, I sometimes want to set the image from a video screencap, so it would be convenient to be able to set the image from the clipboard contents.

This may sometimes also be convenient for scene cover images when I could right click -> copy image from a website and set the image that way.

Easier way to set performer and scene images.

Not really, currently requires saving to a file

### Additional context

_No response_

Gykes:
Currently you can use `Crtl + V` or `Cmd + V` to paste an image into various objects edit modal for the images. The UI doesn't make it very obvious to users that this is possible. As a Stash user for several years I had no idea this was possible till that request for it was brought up and I found the code. It is documented in the manual but I feel like people don't go looking around the manual for fun and makes it not super obvious that this feature exists. 

Added a new "From Clipboard" option to the `Set image...` options list.  I did testing on Chromium based broswer and firefox/waterfox and Safari. 

Fixes: https://github.com/stashapp/stash/issues/6625

Chromium based usage: 

- The first time you use this you will get a permissions pop up allowing your browser to access your clip board. You can approve it and then it wont ask you again. If you deny permissions then the next time you try to use this feature you will get a warning starting that permissions are wrong and to go into your browser to modify. If the paste completes succesfully you will get a green message letting the user know that it has been applied. If you have nothing/ not an image in your clipboard a red error will appear letting the user know that there was no image in their clipboard

Firefox/Waterfox and Safari: 

- How they handle it is different and you need specific permission every time this is used (see Firefox image below). It functions the same just one more click due to how that browser handles this setup. Sadly there is no way around this and I don't think trying to avoid browser based security features is a good idea. 

UI: 
Screenshot 2026-03-02 at 21 55 11

Firefox/waterfox UI: 
Screenshot 2026-03-02 at 21 40 31

Documentation: 
https://caniuse.com/clipboard
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations

^^ Thanks Feederbox

Resolves #3416

### DIFF
--- a/ui/v2.5/src/utils/image.tsx
+++ b/ui/v2.5/src/utils/image.tsx
@@ -10,7 +10,9 @@ const blobToDataURL = (blob: Blob): Promise<string> =>
 
 const readImage = (file: File, onLoadEnd: (imageData: string) => void) => {
   // only proceed if no error encountered
-  blobToDataURL(file).then(onLoadEnd).catch(() => {});
+  blobToDataURL(file)
+    .then(onLoadEnd)
+    .catch(() => {});
 };
 
 const onImageChange = (
--- a/ui/v2.5/src/utils/image.tsx
+++ b/ui/v2.5/src/utils/image.tsx
@@ -13,46 +13,12 @@ const readImage = (file: File, onLoadEnd: (imageData: string) => void) => {
   blobToDataURL(file).then(onLoadEnd).catch(() => {});
 };
 
-const pasteImage = (
-  event: ClipboardEvent,
-  onLoadEnd: (imageData: string) => void
-) => {
-  const files = event?.clipboardData?.files;
-  if (!files?.length) return;
-
-  const file = files[0];
-  readImage(file, onLoadEnd);
-};
-
 const onImageChange = (
   event: React.FormEvent<HTMLInputElement>,
   onLoadEnd: (imageData: string) => void
 ) => {
   const file = event?.currentTarget?.files?.[0];
   if (file) readImage(file, onLoadEnd);
-};
-
-const usePasteImage = (
-  onLoadEnd: (imageData: string) => void,
-  isActive: boolean = true
-) => {
-  const encodeImage = useCallback(
-    (data: string) => {
-      onLoadEnd(data);
-    },
-    [onLoadEnd]
-  );
-
-  useEffect(() => {
-    const paste = (event: ClipboardEvent) => pasteImage(event, encodeImage);
-    if (isActive) {
-      document.addEventListener("paste", paste);
-    }
-
-    return () => document.removeEventListener("paste", paste);
-  }, [isActive, encodeImage]);
-
-  return false;
 };
 
 const imageToDataURL = async (url: string) => {
@@ -71,6 +37,32 @@ const readClipboardImage = async (): Promise<string | null> => {
     }
   }
   return null;
+};
+
+const usePasteImage = (
+  onLoadEnd: (imageData: string) => void,
+  isActive: boolean = true
+) => {
+  const encodeImage = useCallback(
+    (data: string) => {
+      onLoadEnd(data);
+    },
+    [onLoadEnd]
+  );
+
+  useEffect(() => {
+    const paste = async () => {
+      const data = await readClipboardImage();
+      if (data) encodeImage(data);
+    };
+    if (isActive) {
+      document.addEventListener("paste", paste);
+    }
+
+    return () => document.removeEventListener("paste", paste);
+  }, [isActive, encodeImage]);
+
+  return false;
 };
 
 const ImageUtils = {
--- a/ui/v2.5/src/utils/image.tsx
+++ b/ui/v2.5/src/utils/image.tsx
@@ -29,7 +29,24 @@ const imageToDataURL = async (url: string) => {
   return blobToDataURL(blob);
 };
 
+// uses event.clipboardData which works in all contexts including insecure HTTP
+const pasteImage = (
+  event: ClipboardEvent,
+  onLoadEnd: (imageData: string) => void
+) => {
+  const files = event?.clipboardData?.files;
+  if (!files?.length) return;
+
+  const file = Array.from(files).find((f) => f.type.startsWith("image/"));
+  if (file) readImage(file, onLoadEnd);
+};
+
+// uses Clipboard API which requires secure context (HTTPS or localhost)
 const readClipboardImage = async (): Promise<string | null> => {
+  if (!window.isSecureContext) {
+    return null;
+  }
+
   const items = await navigator.clipboard.read();
   for (const item of items) {
     const imageType = item.types.find((t) => t.startsWith("image/"));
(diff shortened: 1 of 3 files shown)


Gykes:
WP, I removed the old paste logic in lieu of mine. It also happens to fix this bug: https://github.com/stashapp/stash/issues/3416. My new code specifically checks for an image in the clipboard and I believe the old one was just taking the first item instead.

feederbox826 (approved):
The previous image pasting code https://github.com/Gykes/stash/blob/5dbf1797e999022b0de4ae8f5cf9798a6904d2c9/ui/v2.5/src/utils/image.tsx#L21

would take files[0], which would trigger #3416. TIL a single object can also have multiple types, repeat for the length of the clipboard

ui/v2.5/src/utils/image.tsx
```diff
@@ -30,6 +23,24 @@ const onImageChange = (
   if (file) readImage(file, onLoadEnd);
 };
 
+const imageToDataURL = async (url: string) => {
+  const response = await fetch(url);
+  const blob = await response.blob();
+  return blobToDataURL(blob);
+};
```
feederbox826, line 30: could be converted into an async arrow function to avoid await chains

```ts
const imageToDataURL = async (url: string) =>
  fetch(url)
    .then(response => response.blob())
    .then(blob => blobToDataURL(blob))
```

smith113-p:
Maybe change the text to something like "From clipboard (Ctrl+V)" as an indication that there's also a keyboard shortcut?

Gykes:
> Maybe change the text to something like "From clipboard (Ctrl+V)" as an indication that there's also a keyboard shortcut?

Ehh, seems kinda mouthy. Also it would need to be OS agnostic (ctrl vs cmd)

I could maybe add a tooltip to it that says that you can do it

WithoutPants:
This has broken the existing keyboard shortcut for me. Nothing happens when entering Ctrl-V on an eligible page. I also get the following error when clicking on the `From clipboard` button:

```
TypeError: Cannot read properties of undefined (reading 'read')
    at Object.readClipboardImage (image.tsx:33:43)
    at onPasteClipboard (ImageInput.tsx:66:39)
    at HTMLUnknownElement.callCallback2 (chunk-VW6ZD3TK.js?v=7eef757c:3899:22)
    at Object.invokeGuardedCallbackDev (chunk-VW6ZD3TK.js?v=7eef757c:3924:24)
    at invokeGuardedCallback (chunk-VW6ZD3TK.js?v=7eef757c:3958:39)
    at invokeGuardedCallbackAndCatchFirstError (chunk-VW6ZD3TK.js?v=7eef757c:3961:33)
    at executeDispatch (chunk-VW6ZD3TK.js?v=7eef757c:6741:11)
    at processDispatchQueueItemsInOrder (chunk-VW6ZD3TK.js?v=7eef757c:6761:15)
    at processDispatchQueue (chunk-VW6ZD3TK.js?v=7eef757c:6770:13)
    at dispatchEventsForPlugins (chunk-VW6ZD3TK.js?v=7eef757c:6778:11)
```

Gykes:
> This has broken the existing keyboard shortcut for me. Nothing happens when entering Ctrl-V on an eligible page. I also get the following error when clicking on the `From clipboard` button:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'read')
>     at Object.readClipboardImage (image.tsx:33:43)
>     at onPasteClipboard (ImageInput.tsx:66:39)
>     at HTMLUnknownElement.callCallback2 (chunk-VW6ZD3TK.js?v=7eef757c:3899:22)
>     at Object.invokeGuardedCallbackDev (chunk-VW6ZD3TK.js?v=7eef757c:3924:24)
>     at invokeGuardedCallback (chunk-VW6ZD3TK.js?v=7eef757c:3958:39)
>     at invokeGuardedCallbackAndCatchFirstError (chunk-VW6ZD3TK.js?v=7eef757c:3961:33)
>     at executeDispatch (chunk-VW6ZD3TK.js?v=7eef757c:6741:11)
>     at processDispatchQueueItemsInOrder (chunk-VW6ZD3TK.js?v=7eef757c:6761:15)
>     at processDispatchQueue (chunk-VW6ZD3TK.js?v=7eef757c:6770:13)
>     at dispatchEventsForPlugins (chunk-VW6ZD3TK.js?v=7eef757c:6778:11)
> ```

What's your setup? I tried on the 3 browsers listed above and had no issues.

Maista6969:
> What's your setup? I tried on the 3 browsers listed above and had no issues.

Are you running on HTTPS behind a reverse proxy? The [documentation for navigator.clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard) says it only works in secure contexts

also I smell bad lol

Gykes:
> > What's your setup? I tried on the 3 browsers listed above and had no issues.
> 
> Are you running on HTTPS behind a reverse proxy? The [documentation for navigator.clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard) says it only works in secure contexts

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard

Ahh, I did all my testing on localhost. Didn't even think to try unsecure connections

PS: Accidentally edited your comment lol. Not used to the edit button being there.

WithoutPants (approved):
Confirmed paste keyboard shortcut is working again. Verified that Set Image from Clipboard item is not shown on unsecure connection. Confirmed that button is shown on localhost connection and works correctly.

WithoutPants:
It looks like the keyboard shortcut has regressed the fix to #3416.

WithoutPants:
My suggested fix for #3416, which I think we can include here, is that `pasteImage` function should ensure the current `activeElement` is not an `HTMLInputElement`. This would stop the image pasting code from activating when pasting into an input field.

```
const pasteImage = (
  event: ClipboardEvent,
  onLoadEnd: (imageData: string) => void
) => {
  const files = event?.clipboardData?.files;
  if (!files?.length) return;

  if (document.activeElement instanceof HTMLInputElement) {
    // don't interfere with pasting text into inputs
    return;
  }

  const file = Array.from(files).find((f) => f.type.startsWith("image/"));
  if (file) readImage(file, onLoadEnd);
};
```

Gykes:
Good call, updated

PR #6638, opened 2026-03-03
Lint + build update and retooling

feederbox826:
Metrics:
- Reduced compiler image size by ~1GB
- Builds now take [6:32](https://github.com/feederbox826/stash/actions/runs/22656939313) down from  [~9:xx](https://github.com/stashapp/stash/actions/runs/22654149830)
- Lint now takes [0:46](https://github.com/stashapp/stash/actions/runs/22656937700) instead of [3:32](https://github.com/stashapp/stash/actions/runs/22654149827)

A non-exhaustive list of changes and improvments

## actions
- updated all image versions to latest version (and node 24 runtime if available)
- ~~removed references to "releases/**" branch~~

## lint
- hugely simplifed lint workflow (cr @Gykes)

## build
- used inline caching for node and golang
- build artifacts in parallel (cr @Gykes)
- updated binfmt (cross-platform image) from deprecated image

## compiler
- Add compiler build workflow through GHA instead of needing manual push
- switched to inline docker ADD for osxcross and OSX SDK (hugely better caching, smaller layer size, inline hash validation)
- downgraded to freebsd 12.4 to match targeted version
- removed deprecated nodesource, switched to pulling out of docker image for latest direct, static builds
- handle compilation in multi-image steps for hugely improved caching and parallel buidls
- remove ssh-keyscan 
- split off dependencies from earlier stages
- use official freebsd archive instead of third-party mirror


feederbox826:
Currently is on my custom image to demonstrate compiler improvments and to prove it works, will be changed once validated

discourse-stashapp:

https://discourse.stashapp.cc/t/stash-v0-31-release-development-thread/4881/3

.github/workflows/build.yml
```diff
@@ -2,10 +2,9 @@ name: Build
 
 on:
   push:
-    branches: 
+    branches:
       - develop
       - master
-      - 'releases/**'
```
WithoutPants, line 8: Why was this removed? This is used to build the hotfix release branches (prior to publishing).

.github/workflows/build-compiler.yml
```diff
@@ -0,0 +1,26 @@
+name: Compiler Build
```
WithoutPants, line 1: I'm a bit stumped as to what this workflow does?

.github/workflows/build.yml
```diff
@@ -114,31 +227,31 @@ jobs:
         echo "RELEASE_DATE=$(date +'%Y-%m-%d %H:%M:%S %Z')" >> $GITHUB_ENV
 
     - name: Upload Windows binary
-      # only upload binaries for pull requests
+      # only upload binaries for pull requests not targeting develop or master
```
WithoutPants: This doesn't seem correct, but I can't determine what the original purpose of this condition is. PRs that target these branches _do_ upload their artifacts.

WithoutPants (changes requested):
Please bump the version number in `docker/compiler/Makefile` to `13`.

.github/workflows/build.yml
feederbox826: oh I wasn't aware

.github/workflows/build-compiler.yml
feederbox826: It builds a new version of the compiler image

.github/workflows/build.yml
feederbox826: merged in from @Gykes

.github/workflows/build-compiler.yml
feederbox826: Pulling from dockerhub also takes longer, the PR'd image pulls from github and should be faster since it's on ajacent infrastructure

feederbox826: It's also possible to leave it on docker if you prefer the manual build-push approach and use the gcr.io proxy to avoid ratelimits

.github/workflows/build-compiler.yml
```diff
@@ -0,0 +1,26 @@
...
+
+on:
+  workflow_dispatch:
+
+env:
+  COMPILER_IMAGE: ghcr.io/feederbox826/stash-compiler
```
feederbox826:
```suggestion
  COMPILER_IMAGE: stashapp/compiler:13
```

leaving as blocking review

WithoutPants: Is this correct if we move to ghcr?

.github/workflows/build.yml
```diff
@@ -15,50 +15,161 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  COMPILER_IMAGE: stashapp/compiler:12
+  COMPILER_IMAGE: ghcr.io/feederbox826/stash-compiler
```
WithoutPants: This will need to be updated back to `stashapp/compiler` presumably?

WithoutPants (approved):
I'm happy to move ahead with this. I will do some local testing to ensure the binaries run correctly for platforms I have access to.

.github/workflows/build.yml
feederbox826: `ghcr.io/stashapp/compiler` if we move ahead with the switch to GHCR. left it as `compiler:13` in case we didn't

feederbox826: since its just for CI, didn't mirror to dockerhub so it will remain frozen at 12

docker/compiler/Makefile
```diff
@@ -1,16 +1,19 @@
+host=ghcr.io
 user=stashapp
 repo=compiler
 version=12
```
WithoutPants:
```suggestion
version=13
```

docker/compiler/Makefile
```diff
@@ -1,16 +1,19 @@
...
 
+VERSION_IMAGE = ${host}/${user}/${repo}:${version}
+LATEST_IMAGE = ${host}/${user}/${repo}:latest
+
+latest:
+	docker build -t ${LATEST_IMAGE} .
+
```
WithoutPants:
```suggestion
```

Duplicate target

WithoutPants (changes requested):

PR #6642, opened 2026-03-04
Sort performers and studios by scenes file size

### LINKED ISSUE #5530: [Feature] Short Form Title (50 chars or less.)
I'd like to be able to sort Performers and Studios by file size on disks. In particular for scenes, but It can be extended to gallery and images to.

Right now you can sort by scene count/gallery count/image count, this feature would add, the following options "Scenes size on disk", "Galleries size on disk" and "Images size on disk" and would display the total size instead of the scene/gallery/image count in the card/row of the corresponding view

See the following examples : 
Performer cards|200
Performer list|200
Studio cards|200

At least for the scenes (I don't really use gallery/image right now), I've made some graphql (with sql query) that will display performers/studios order by total file size but having it directly in the stash webui would be better.

**Additional context**

dev-null-life:
## Summary
- Adds `scenes_size` as a sort option for Performers and Studios, allowing users to sort by total file size of associated scenes on disk
- Follows the existing `scenes_duration` sort pattern (correlated subquery summing through `scenes_files` join)
- Adds frontend sort dropdown entries and i18n label

Closes #5530

## Changes
- **`pkg/sqlite/performer.go`** — SQL subquery (`selectPerformerScenesSizeSQL`), allowlist entry, switch case in `getPerformerSort`
- **`pkg/sqlite/studio.go`** — `sortByScenesSize()` method, allowlist entry, switch case in `getStudioSort`
- **`ui/v2.5/src/models/list-filter/performers.ts`** — `"scenes_size"` sort option
- **`ui/v2.5/src/models/list-filter/studios.ts`** — `"scenes_size"` sort option
- **`ui/v2.5/src/locales/en-GB.json`** — `"Scene Size"` i18n label

## Test plan
- [ ] Navigate to Performers list → Sort dropdown shows "Scene Size"
- [ ] Sort by "Scene Size" descending → performers with the most scene data appear first
- [ ] Navigate to Studios list → Sort dropdown shows "Scene Size"
- [ ] Sort by "Scene Size" descending → studios with the most scene data appear first
- [ ] Verify no errors in browser console or server logs

### DIFF
--- a/pkg/sqlite/performer.go
+++ b/pkg/sqlite/performer.go
@@ -778,6 +778,28 @@ func (qb *PerformerStore) sortByScenesDuration(direction string) string {
 	return " ORDER BY (" + selectPerformerScenesDurationSQL + ") " + direction
 }
 
+// used for sorting by total scene file size
+var selectPerformerScenesSizeSQL = utils.StrFormat(
+	"SELECT COALESCE(SUM({files}.size), 0) FROM {performers_scenes} s "+
+		"LEFT JOIN {scenes} ON {scenes}.id = s.{scene_id} "+
+		"LEFT JOIN {scenes_files} ON {scenes_files}.{scene_id} = {scenes}.id "+
+		"LEFT JOIN {files} ON {files}.id = {scenes_files}.file_id "+
+		"WHERE s.{performer_id} = {performers}.id",
+	map[string]interface{}{
+		"performer_id":      performerIDColumn,
+		"performers":        performerTable,
+		"performers_scenes": performersScenesTable,
+		"scenes":            sceneTable,
+		"scene_id":          sceneIDColumn,
+		"scenes_files":      scenesFilesTable,
+		"files":             fileTable,
+	},
+)
+
+func (qb *PerformerStore) sortByScenesSize(direction string) string {
+	return " ORDER BY (" + selectPerformerScenesSizeSQL + ") " + direction
+}
+
 var performerSortOptions = sortOptions{
 	"birthdate",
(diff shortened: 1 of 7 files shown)


Gykes:
Haven't reciew but just a 2 second look wouldn't `file_size` be better?

dev-null-life:
My thought is that `file_size` would be too ambiguous in the context of performers or studios. `scenes_size` is consistent with the other scene sorts.

Gykes (commented):
Overall it looks good. I can only really think of 2 things: 

1) Have you thought about extending this to tags/galleries?

2) Im still urked by `scenes_size`. I know it follows the rest of the setup just when I think of it my brain goes to resolution. I think `scenes_files_size` would be better and less ambiguous but I don't think it should stop this PR from merging.

dev-null-life:
Good points, thanks for the review!

1. Extended `scenes_size` sort to tags in the latest commit — follows the same pattern as `scenes_duration` for tags.

2. On the naming: I agree `scenes_size` could be read as resolution, but I think keeping it consistent with `scenes_duration` and `scenes_count` is the right call. `scenes_files_size` is more explicit but breaks that symmetry. Happy to change it if there's consensus.

Galleries are a different case since they have their own files rather than going through scenes, so I'd leave that as a separate PR if/when someone wants to tackle it.

Gykes:
I meant to say groups 🤦 

I'm not looking at the code rn to see if it's portable so if it's not then no worries. WP can make the final decision on the naming and get it in when he feels it's up to snuff

PR #6643, opened 2026-03-04
feat: add Cup Size sort option for performers

### LINKED ISSUE #4831: [Feature] Breast/Cup Size Sort on Performers Page
### Is your feature request related to a problem? Please describe:

There is a `Penis Size` sort but not a `Breast Size` sort option on the `/Performers` page to be able to sort the women as well as the men by their physical attributes.

### Describe the solution you'd like:

I would like to see a `Breast Size` or `Cup Size` sort added to allow sorting by increasing or decreasing cup size.

### Describe alternatives you've considered:

I am not particularly well-versed in JS or CSS to but I'd imagine that this would be possible by grabbing the characters between the number and first hyphen in a female performer's `Measurements` field as a sort of "top level" sort to sort by cup size "category" first. Then the numbers preceding the letter(s) could be included for further sorting.

dev-null-life:
## Summary
- Adds a "Cup Size" sort option to the Performers page, extracting cup size from the `measurements` string field (e.g. `34DD-24-34` → `DD` → normalized to `E`)
- Normalizes equivalent cup sizes: DD→E, DDD→F so they sort together
- Handles NULL, empty, and malformed measurements gracefully (sort to end)

Closes #4831

## Implementation
- **Computed SQL sort** — no schema migration needed, follows existing pattern for custom sorts (`sortByScenesDuration`, `sortByPlayCount`, etc.)
- SQL extracts cup letters via `LTRIM(SUBSTR(...), '0123456789')`, then normalizes with `REPLACE`
- Backend: new `sortByCupSize()` method + whitelist entry + switch case in `getPerformerSort()`
- Frontend: sort option added to performer list + `en-GB.json` translation

## Test plan
- [x] Go code compiles without errors
- [x] `TestPerformerQuerySortCupSize` test added (ASC + DESC, verifies no query errors)
- [ ] Manual: Performers page → Sort dropdown → "Cup Size" appears
- [ ] Manual: Sort ASC orders A < B < C < D < E (DD) < F (DDD) < G < ...
- [ ] Manual: Performers with no measurements sort to end

### DIFF
--- a/pkg/sqlite/performer.go
+++ b/pkg/sqlite/performer.go
@@ -778,11 +778,23 @@ func (qb *PerformerStore) sortByScenesDuration(direction string) string {
 	return " ORDER BY (" + selectPerformerScenesDurationSQL + ") " + direction
 }
 
+func (qb *PerformerStore) sortByCupSize(direction string) string {
+	direction = getSortDirection(direction)
+	return " ORDER BY CASE" +
+		" WHEN performers.measurements IS NULL OR performers.measurements = '' THEN NULL" +
+		" WHEN INSTR(performers.measurements, '-') = 0 THEN NULL" +
+		" ELSE REPLACE(REPLACE(" +
+		"LTRIM(SUBSTR(performers.measurements, 1, INSTR(performers.measurements, '-') - 1), '0123456789')," +
+		" 'DDD', 'F'), 'DD', 'E')" +
+		" END " + direction
+}
+
 var performerSortOptions = sortOptions{
 	"birthdate",
 	"career_start",
 	"career_end",
 	"created_at",
+	"cup_size",
 	"galleries_count",
 	"height",
 	"id",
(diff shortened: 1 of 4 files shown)


Gykes:
So, this feature has been requested quite a few times. I think the best way to handle this is to actually split the field into individual subfields. I thinking having one field makes it harder to sort/filter anything. Recently we broke out career length for performer https://github.com/stashapp/stash/pull/6449 and I think something like that change would be a better approach. 

I wont review it currently but if WP gives his okay I will give it a gander.

dev-null-life:
Yeah, I saw that as well. It looked a little too ambitious, so I chose the minimal route instead.

Gykes:
So, the main dev is on board with converting this PR out to break the single field into separate individual fields. You can reference https://github.com/stashapp/stash/pull/6449 to see kind of how to do it.

dev-null-life:
I'll take a look!

PR #6647, opened 2026-03-05
Refactor bulk edit dialogs

### LINKED ISSUE #1408: [Feature] Enhance bulk editing of Performers and Scenes to edit every allowed field

### LINKED ISSUE #6419: Want to clear studio ID in batch edit dialog

I would like it if the batch edit dialog let me clear the scene id. The reason is that tpdb studio IDs are often unhelpful, and if a studio doesn't really have IDs then the scraper doesn't overwrite the tpdb id.

More correct metadata, especially useful if I want to submit scenes to stash db.

Clear the id individually for each scene 

### Additional context

_No response_

WithoutPants:
Resolves #6419
Resolves #1408 

- added date and synopsis fields to bulk update group input
- change titles of bulk edit dialogs to `Edit x {entityType}`, adding the count of the entities being updated
- added bulk date input field and replaces existing date fields in the edit performers dialog
- restyled edit performer dialog to use inline fields, to use (slightly) less vertical real estate:
image

- refactored edit groups dialog, adding the following fields:
  - date
  - synopsis
image

- refactored edit galleries dialog, adding the following fields:
  - code
  - date
  - photographer
  - scenes
  - details
image

- refactored edit images dialog, adding the following fields:
  - code
  - date
  - photographer
  - details
image

- refactored edit scenes dialog, adding the following fields:
  - code
  - date
  - director
  - details
image


DogmaDragon:
I was recently asked if there was a way to bulk null ratings. I was under the impression it was possible via bulk dialog, but testing shown that it is not the case. Could we use this refactor as an opportunity to add support for it?

Gykes:
Being able to bulk remove StashID, on all modals that support them, has been requested as well.

smith113-p:
Why is director the only text field that has (what I assume is) a clear icon?

More generally, for text fields, how do you indicate "clear" vs "don't touch"?

ui/v2.5/src/components/Scenes/EditScenesDialog.tsx
```diff
@@ -109,136 +129,44 @@ export const EditScenesDialog: React.FC<IListOperationProps> = (
...
-  }, [organized, checkboxRef]);
-
-  function renderMultiSelect(
-    type: "performers" | "tags" | "groups",
-    ids: string[] | undefined
+  function renderTextField(
+    name: string,
+    value: string | undefined | null,
+    setter: (newValue: string | undefined) => void,
+    area: boolean = false
```
Gykes: This is copied a lot over various dialogue boxes. Should probably break this out at some point.

ui/v2.5/src/components/Scenes/EditScenesDialog.tsx
```diff
@@ -109,136 +129,44 @@ export const EditScenesDialog: React.FC<IListOperationProps> = (
...
-    } else if (organized === undefined) {
-      setOrganized(false);
-    } else {
-      setOrganized(true);
+    if (area) {
+      return (
+        <Form.Group
+          controlId={name}
+          data-field={name}
+          as={area ? undefined : Row}
```
Gykes: Is this dead logic? Since it's already inside the if branch doesn't that make area always true?

ui/v2.5/src/components/Shared/DateInput.tsx
```diff
@@ -70,28 +74,33 @@ const _DateInput: React.FC<IProps> = (props: IProps) => {
...
   });
 
+  const placeholderText = props.placeholder
+    ? `${props.placeholder} (${formatHint})`
+    : formatHint;
+
   return (
     <div>
-      <InputGroup hasValidation>
+      <InputGroup hasValidation className="bulk-update-date-input">
```
Gykes: You added this to the shared component. Wouldn't that mean that every `DateInput` in the app would be effected by this? Could potentially cause styling regressions.

Gykes (changes requested):

ui/v2.5/src/components/Shared/DateInput.tsx
WithoutPants: Good catch, fixed.

WithoutPants:
Made the following changes:
- added utility component `BulkUpdateFormGroup` to replace logic of `renderTextField`
- updated styling of text inputs to be consistent with the rest of the UI
- moved the unset button to be an appended button
- empty values now explicitly show an `<empty>` placeholder to differentiate between an empty value or multiple existing values
- added clear button and validation to date fields

image
image

WithoutPants:
> Why is director the only text field that has (what I assume is) a clear icon?
> 
> More generally, for text fields, how do you indicate "clear" vs "don't touch"?

I screwed up the screenshots a little. When editing a single object, the unset buttons do not appear. When modifying multiple objects, the edit dialog will show a value if it is the same for all objects, or an `<existing value>` placeholder otherwise. The user can modify the value to set that value for all objects, or reset to existing values by clicking on the unset button on the right of the text field.

> I was recently asked if there was a way to bulk null ratings. I was under the impression it was possible via bulk dialog, but testing shown that it is not the case. Could we use this refactor as an opportunity to add support for it?

It's not ideal, but the current workaround is to set a rating on all the objects, then edit again and clear the rating. The logic currently only clears the rating if all the objects have the same existing rating.

I don't have a ready solution to improve this, so I'd prefer to defer this for the time being.

Issue #6455, opened 2025-12-24
[FR] SQL Query Improvments for Larger DB


Gykes:
### Describe the feature you'd like

Currently, there are some edge case issue for larger databases. 

The main conversation for it starts on Discord [here](https://discord.com/channels/559159668438728723/559159931505213440/1453045631370395659)

My proposed fixes/helps are on Discord [here](https://discord.com/channels/559159668438728723/559159931505213440/1453197001490567410)

Current Behavior:

  When filtering scenes by folder path (e.g., /data/namer/), Stash builds a query like:
```
  SELECT DISTINCT scenes.id
  FROM scenes
  LEFT JOIN scenes_files ON scenes_files.scene_id = scenes.id
  LEFT JOIN files ON scenes_files.file_id = files.id
  LEFT JOIN folders ON files.parent_folder_id = folders.id
  ...
  WHERE (folders.path || '/' || files.basename LIKE '%/data/namer/%')
  AND (TRIM(scene_stash_ids.stash_id) = '' OR scene_stash_ids.stash_id IS NULL)
```

  This approach has several performance issues:

  1. `LIKE '%pattern%'` with leading wildcard prevents SQLite from using the index on `folders.path`, forcing a full table scan

  2. String concatenation `(folders.path || '/' || files.basename)`  must be computed for every row

  3. `TRIM()` function call prevents index usage and adds per-row overhead

  4. Join order: Starting from scenes and `LEFT JOIN` everything processes all 689k scenes before filtering

  Proposed Optimizations:

  1. Use `GLOB` instead of `LIKE` for prefix matching

   - Instead of: `folders.path LIKE '%/data/namer/%'` use: `folders.path GLOB '/data/namer/*'`

   - `GLOB` enables SQLite to use the `B-tree index` on `folders.path` for prefix matching. It's also case-sensitive, which is correct for filesystem paths and reduces comparison overhead.

  2. Reorder joins to filter early

 -  Start from folders (the most selective table after filtering) and `INNER JOIN` to `files` -> `scenes_files` -> `scenes`, rather than starting from scenes and `LEFT JOIN` everything. This lets SQLite narrow down to matching scenes immediately instead of processing all scenes first.

  3. Add INDEXED BY hints

  - Explicitly hint the query planner to use optimal indexes:
    - `INDEXED BY index_folders_on_path_unique` (TESTING REQUIRED)
    - `INDEXED BY index_scenes_files_file_id` (TESTING REQUIRED

[...]

PR #6648, opened 2026-03-05
Optimise table joins

WithoutPants:
Replaces left joins with inner joins where it makes sense to do so (ie where we are checking for the presence of related objects). Left joins are used only where we need to check for null related values.

Also removed the use of `trim` from the stash ids criterion where clauses.

While the unit tests pass, I will likely defer this until the following release to allow time for potential breakages to appear.

No actual performance comparison testing has been performed yet.

Related to #6455

### DIFF
--- a/pkg/sqlite/tag_filter.go
+++ b/pkg/sqlite/tag_filter.go
@@ -184,7 +184,7 @@ func (qb *tagFilterHandler) aliasCriterionHandler(alias *models.StringCriterionI
 		primaryFK:    tagIDColumn,
 		joinTable:    tagAliasesTable,
 		stringColumn: tagAliasColumn,
-		addJoinTable: func(f *filterBuilder) {
+		addJoinTable: func(f *filterBuilder, joinType joinType) {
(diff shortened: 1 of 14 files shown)


pkg/sqlite/tag_filter.go
```diff
@@ -184,8 +184,8 @@ func (qb *tagFilterHandler) aliasCriterionHandler(alias *models.StringCriterionI
 		primaryFK:    tagIDColumn,
 		joinTable:    tagAliasesTable,
 		stringColumn: tagAliasColumn,
-		addJoinTable: func(f *filterBuilder) {
-			tagRepository.aliases.join(f, "", "tags.id")
+		addJoinTable: func(f *filterBuilder, joinType joinType) {
+			tagRepository.aliases.leftJoin(f, "", "tags.id")
```
Gykes: `joinType` is accepted but ignored. Looks like it's always using `leftJoin`.

Gykes (changes requested):

PR #6650, opened 2026-03-06
add image scraping to built-in Auto-Tag scraper

spaceyuck:
The bulk Auto-Tag operations already supported images, but the built-in Auto-Tag scraper did not support them, therefore the option was missing in the image details scrape menu. It was already available for galleries and scenes, images just seem to have just slipped through. Seems to work fine, not sure if there was a reason why this wasn't implemented already.

Hopefully, this can be a first step in making Images a first class citizen in terms of scraping UI. But one step at a time.

Screenshot_20260320_001151
Screenshot_20260320_002326

### DIFF
--- a/pkg/scraper/autotag.go
+++ b/pkg/scraper/autotag.go
@@ -90,6 +90,11 @@ func autotagMatchTags(ctx context.Context, path string, tagReader models.TagAuto
 }
 
 func (s autotagScraper) viaImage(ctx context.Context, _client *http.Client, image *models.Image) (*models.ScrapedImage, error) {
+	path := image.Path
+	if path == "" {
+		return nil, nil
+	}
+
 	var ret *models.ScrapedImage
 
 	// only trim extension if image is file-based
@@ -97,11 +102,6 @@ func (s autotagScraper) viaImage(ctx context.Context, _client *http.Client, imag
 
 	// populate performers, studio and tags based on image path
 	if err := txn.WithReadTxn(ctx, s.txnManager, func(ctx context.Context) error {
-		path := image.Path
-		if path == "" {
-			return nil
-		}
-
 		performers, err := autotagMatchPerformers(ctx, path, s.performerReader, trimExt)
 		if err != nil {
 			return fmt.Errorf("autotag scraper viaImage: %w", err)
--- a/pkg/scraper/autotag.go
+++ b/pkg/scraper/autotag.go
@@ -89,11 +89,13 @@ func autotagMatchTags(ctx context.Context, path string, tagReader models.TagAuto
 	return ret, nil
 }
 
-func (s autotagScraper) viaImage(ctx context.Context, client *http.Client, image *models.Image) (*models.ScrapedImage, error) {
+func (s autotagScraper) viaImage(ctx context.Context, _client *http.Client, image *models.Image) (*models.ScrapedImage, error) {
 	var ret *models.ScrapedImage
-	const trimExt = false
 
-	// populate performers, studio and tags based on scene path
+	// only trim extension if image is file-based
+	trimExt := image.PrimaryFileID != nil
+
+	// populate performers, studio and tags based on image path
 	if err := txn.WithReadTxn(ctx, s.txnManager, func(ctx context.Context) error {
 		path := image.Path
 		if path == "" {
--- a/pkg/scraper/autotag.go
+++ b/pkg/scraper/autotag.go
@@ -89,6 +89,47 @@ func autotagMatchTags(ctx context.Context, path string, tagReader models.TagAuto
 	return ret, nil
 }
 
+func (s autotagScraper) viaImage(ctx context.Context, client *http.Client, image *models.Image) (*models.ScrapedImage, error) {
+	var ret *models.ScrapedImage
+	const trimExt = false
+
+	// populate performers, studio and tags based on scene path
+	if err := txn.WithReadTxn(ctx, s.txnManager, func(ctx context.Context) error {
+		path := image.Path
+		if path == "" {
+			return nil
+		}
+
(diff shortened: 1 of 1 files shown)


pkg/scraper/autotag.go
```diff
@@ -89,6 +89,47 @@ func autotagMatchTags(ctx context.Context, path string, tagReader models.TagAuto
 	return ret, nil
 }
 
+func (s autotagScraper) viaImage(ctx context.Context, client *http.Client, image *models.Image) (*models.ScrapedImage, error) {
```
Gykes: Shouldn't `client` be `_client` here?

pkg/scraper/autotag.go
```diff
@@ -89,6 +89,47 @@ func autotagMatchTags(ctx context.Context, path string, tagReader models.TagAuto
...
+	var ret *models.ScrapedImage
+	const trimExt = false
```
Gykes: Should this be true? 

`viaGallery` trims based on `gallery.PrimaryFileID != nil` to avoid the extension interfering with matches.

pkg/scraper/autotag.go
```diff
@@ -89,6 +89,47 @@ func autotagMatchTags(ctx context.Context, path string, tagReader models.TagAuto
...
+
+	// populate performers, studio and tags based on scene path
```
Gykes: `scene path` -> `image path`

Gykes (changes requested):
This looks reasonable as a gap-fill since scenes and galleries already support it. I believe that the practical value is narrower for images. Image filenames tend to be generic and images in galleries would typically inherit metadata at the gallery level rather than being individually auto-tagged. The main benefit is standalone images in well-organized folder structures.

pkg/scraper/autotag.go
spaceyuck: I've adopted the gallery solution (I'm unsure if an image witrhout a file is possible though, even if the model comment says so).

As a side note: the scene Auto-Tag scraper passes a constant false here too.

spaceyuck:
> The main benefit is standalone images in well-organized folder structures.

Definitely, it's a some metadata is better than no metadata kind of deal.

I've made some changes.

Gykes (changes requested):
After pulling it down I don't see the option in Settings -> Tasks -> Auto Tag to enable and disable images.

When I told it to auto tag I did see this in the logs:  `Auto-tagging images...` so it appears to be trying to work just there is no way to enable/disable. I know your PR title says that it gets added to the image details but if you run autotag app wide is still attempts to autotag images.

spaceyuck:
> After pulling it down I don't see the option in Settings -> Tasks -> Auto Tag to enable and disable images.

I'm not sure what you mean, but there are only toggles to select what to tag with (tag, performers, studios), not what to tag (image, scene, gallery)

> When I told it to auto tag I did see this in the logs: `Auto-tagging images...` so it appears to be trying to work just there is no way to enable/disable. I know your PR title says that it gets added to the image details but if you run autotag app wide is still attempts to autotag images.

The Bulk Auto-Tag mechanism from Settings Tasks and from Tags / Performers / Studios already supported images before this, and is run out of I think [task_autotag.go](https://github.com/stashapp/stash/blob/c832e1a8a29250cd2720bac3b3b042b0b62aee49/internal/manager/task_autotag.go#L606). 

This change should not have any effect on that whole logic. This change is to add the equivalent of the "imageByFragment" scrape operation to the builtin Auto-Tag scraper, so that it would show up in the image details scrape menu, and a (hopefully) future image Tagger list view.

I've reworded the PR title and description a bit to make it more clear what this actually does.

Gykes:
You're absolutely, right. I think I was extremely tired when I wrote that as I cant even remember what I was trying to get across lol. 

Will re-review in the next few days

pkg/scraper/autotag.go
```diff
@@ -89,6 +89,49 @@ func autotagMatchTags(ctx context.Context, path string, tagReader models.TagAuto
...
+
+	// only trim extension if image is file-based
+	trimExt := image.PrimaryFileID != nil
+
+	// populate performers, studio and tags based on image path
+	if err := txn.WithReadTxn(ctx, s.txnManager, func(ctx context.Context) error {
+		path := image.Path
+		if path == "" {
+			return nil
+		}
```
Gykes: The empty path check should probs be outside the transaction. No point opening a read transaction just to find out the path is empty.  You can reference viaGallery above to see what I mean.

Gykes (changes requested):

pkg/scraper/autotag.go
```diff
@@ -89,6 +89,49 @@ func autotagMatchTags(ctx context.Context, path string, tagReader models.TagAuto
 	return ret, nil
 }
 
+func (s autotagScraper) viaImage(ctx context.Context, _client *http.Client, image *models.Image) (*models.ScrapedImage, error) {
```
Gykes, line 92: So, a lot of this shares code with `viaScene` and `viaGallery` might be a good idea to break the shared code out to a private helper to reduce complexity.

Gykes (changes requested):
Documentation will also need to be updated to reflect this change.

pkg/scraper/autotag.go
spaceyuck: Well, I can try, but as a warning, I'm very new to and terrible at writing Go code from scratch. If this were Java or Python, I'd crank out crazy abstractions at high velocity no problem, but this might take a while and be quite bad.

Gykes: No problem. It's not a blocker for the PR or anything but not sure if the main dev will want you to do it or not.

WithoutPants (approved):
Looks good. Didn't see anywhere that documentation needed updating.

PR #6651, opened 2026-03-06
Fix: Maximum update depth exceeded Error on Gallery

### LINKED ISSUE #6652: Galleries with more than a single page of images freezes app
### Describe the bug

If the number of images in a gallery requires multiple pages, 2 or more, no images are displayed the App will freeze when attempting to edit move around the gallery. If the gallery has a single page worth of images, the gallery behaves normally.

### Steps to reproduce

1. Go to Galleries
2. Select a gallery that has more than 1 page of images
3. Try to click around the app or edit the gallery and the app freezes

### Expected behaviour

Gallery displays images

_No response_

### Stash version

v0.30.1-115-g74a8f2e5

### Device details

Chrome 146 on Windows 11

### Relevant log output

```shell

```

Gykes:
Issue raised in discord [here](https://discord.com/channels/559159668438728723/559159910550732809/1479254683804631102)

`GalleryPage` subscribes to the lightbox context via `useGalleryLightbox`. When `ImageList` updates lightbox state via `useLightbox`, the context change re-renders GalleryPage, which cascades to `GalleryImagesPanel` -> `FilteredImageList` -> `ImageList`. Unstable function references (`filterHook`, `onChangePage`) caused dependencies to appear changed on each pass, re-triggering the lightbox effect and creating an infinite loop.

The same inline `onChangePage` pattern exists in the top pagination of other list components. Those don't trigger infinite loops since their parent pages don't consume lightbox context, but could be updated to use `setPage` for consistency in a follow-up.

Fixes: #6652

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryAddPanel.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryAddPanel.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useCallback } from "react";
 import * as GQL from "src/core/generated-graphql";
 import { GalleriesCriterion } from "src/models/list-filter/criteria/galleries";
 import { ListFilterModel } from "src/models/list-filter/filter";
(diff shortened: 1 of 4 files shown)


feederbox826 (approved):
Passes the sniff test, not familar enough with react state to validate, but familiar enough with JS+DOM to understand why

Gykes:
So, I believe the underlying reason this loop only affects gallery pages is that `useGalleryLightbox` calls `useLightboxContext()`, subscribing `GalleryPage` to all lightbox state changes even though it only needs `setLightboxState` to write. Splitting the lightbox context into separate state/dispatch contexts would prevent this type of bug entirely by ensuring write-only consumers don't re-render on state changes. I think this would be better as a follow up PR to "refactor" how this current flow works.

SandiyosDev:
good PR, but context refactor deserves its own dedicated PR to avoid scope creep here.
hoping to have this PR and #6647 merged for 0.31.0 to be tagged and released sooner

discourse-stashapp:

https://discourse.stashapp.cc/t/galleries-not-working-properly-after-last-update/6049/2

PR #6654, opened 2026-03-07
fix: support string-based fingerprints in hashes filter

hyper440:
## Summary
Fixes the \`hashes\` filter in \`findFiles\` and related queries to properly support string-based fingerprints (md5, oshash, and custom types).
## Problem
The \`hashesCriterionHandler\` treats all fingerprint types as phash, converting values via \`StringToPhash\` (hex to int64). This breaks lookups for string-based fingerprints like md5 and oshash because:
1. \`StringToPhash\` fails to parse non-phash values and returns 0
2. The query uses integer comparison (\`WHERE fingerprint = 0\`), which never matches
## Solution
Added a type check to use the appropriate comparison method:
- **phash**: Uses existing integer comparison with optional hamming distance support
- **Other types** (md5, oshash, custom): Uses direct string equality
## Testing
Verified all fingerprint types work correctly:
- md5 lookup
- oshash lookup  
- phash lookup (unchanged behavior)


pkg/sqlite/file_filter.go
```diff
@@ -238,22 +228,28 @@ func (qb *fileFilterHandler) hashesCriterionHandler(hashes []*models.Fingerprint
...
 				distance = *hash.Distance
 			}
 
-			if distance > 0 {
-				// needed to avoid a type mismatch
-				f.addWhere(fmt.Sprintf("typeof(%s.fingerprint) = 'integer'", t))
-				f.addWhere(fmt.Sprintf("phash_distance(%s.fingerprint, ?) < ?", t), value, distance)
+			// Only phash supports distance matching and is stored as integer
+			if hash.Type == models.FingerprintTypePhash {
+				value, _ := utils.StringToPhash(hash.Value)
```
Gykes: `StringToPhash` error is silently discarded. Perhaps we should add some kind of guard here so that it errors out just in case.

pkg/sqlite/file_filter.go
```diff
@@ -238,22 +228,28 @@ func (qb *fileFilterHandler) hashesCriterionHandler(hashes []*models.Fingerprint
...
+				}
 			} else {
-				// use the default handler
-				intCriterionHandler(&models.IntCriterionInput{
-					Value:    int(value),
-					Modifier: models.CriterionModifierEquals,
-				}, t+".fingerprint", nil)(ctx, f)
+				// All other fingerprint types (md5, oshash, sha1, etc.) are stored as strings
+				// Use exact match for string-based fingerprints
+				f.addWhere(fmt.Sprintf("%s.fingerprint = ?", t), hash.Value)
```
Gykes, line 266: This new string matching path could use a test case in the existing fingerprint test suite to verify md5 and oshash lookups work through `hashesCriterionHandler`

Gykes (changes requested):
Mostly nit picky stuff

Gykes (approved):
The only thing I can think is that Phash wasn't added to the test case. The main phash logic wasn't really changed so shouldn't cause issues. Could request that the OP add in Phash next to the other 2 since they are here already.

PR #6655, opened 2026-03-08
Use StashIDPill in the performer modal dialog

smith113-p:
Currently, this dialog just shows a text "Stash-Box Source". This change instead re-uses the StashIDPill, with the main advantage that you can immediately tell which stash box is being used.

The studio modal has a similar issue, but it is harder to make this change because it does not have the link broken down into an endpoint and stash ID part.


smith113-p:
Old UI:
Screenshot 2026-03-07 202210

New UI:
Screenshot 2026-03-07 202932

To clarify, this is the dialog you get when you "refresh" a performer.

Gykes (approved):
I like this. Solid minor UI improvement. Thanks for this.

PR #6656, opened 2026-03-08
Show the stash box for each stash ID in the scene merge dialog

smith113-p:
Currently, this dialog only shows the ID but not the stash box it corresponds to. This is not very useful because the ID does not mean anything to a user.

This changes the dialog to use the standard Stash ID pill.

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneMergeDialog.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneMergeDialog.tsx
@@ -51,7 +51,7 @@ const StashIDsField: React.FC<IStashIDsField> = ({ values }) => {
   if (!values.length) return null;
 
   return (
-    <ul className="pl-0">
+    <ul className="pl-0 mw-100">
       {values.map((v) => (
         <li key={v.stash_id} className="row no-gutters">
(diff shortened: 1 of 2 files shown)


smith113-p:
Old UI:
Screenshot 2026-03-07 211437

New UI:
Screenshot 2026-03-07 211500

I don't know React well enough to use StashIDPill here, so I went with this smaller  change instead.

Gykes:
I'm not a huge fan of how this one looks tbh. I think using the StashIDPill would be better here. How you did it now duplicates some logic and doesn't look as good. I also think it means you could potentially move it to other modals as well for consistency.

If you would like assistance with it I am more than happy to do it for you or I can give you some pointers.

smith113-p:
I updated the pull request, it now looks like this:
Screenshot 2026-03-09 130700

Gykes:
Since you updated the styling for the pill on the css it will effect it across the whole app. I would check to make sure no other areas got messed with. I'm not home to do the check myself quite yet but, if you wanna be helpful, it's a good check to do.

smith113-p:
Yes, I checked a couple of places, I did not try to find all places this is used. Theoretically it should only make a difference if there is not enough space for the full stash ID.

PR #6663, opened 2026-03-09
Show scene resolution and duration in tagger

### LINKED ISSUE #4482: [Feature] Scene tagger, add duration of video
The problem is mostly for scenes from analvids/pissvids, but i'm sure it goes for others too.
Many scenes have wet and dry versions with the same studio code and almost the same name, the video duration if of course different.
On StashDB many of these scenes are tagged to be the wrong version so hashes can be for mixed lengths, on some scenes the wrong version has more hashes than the correct version too.

Example:
https://stashdb.org/scenes/9f135c8f-b2eb-49f8-a39e-61bc4e811427#fingerprints

The other version of the scene with 100% good matches:
https://stashdb.org/scenes/a8194fe5-3cea-442f-b298-e87a868e86ff#fingerprints

Another example with 100% of the matches wrong:
https://stashdb.org/scenes/547856f1-aea0-4e55-86d5-a4af3c7c84d3#fingerprints

They should instead be added to this video:
https://stashdb.org/scenes/df42eeb4-dd34-4555-a1eb-3e1b9ce7e60f#fingerprints

A simple solution to help on this would be to add duration for both the local file and any scraped matches to the scene tagger.

smith113-p:
A scene's duration and resolution is often useful to ensure you have found the right scene. This PR adds the same resolution/duration overlay from the grid view to the tagger view.

Resolves #4482

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneCard.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneCard.tsx
@@ -364,13 +364,21 @@ export const SceneSpecsOverlay: React.FC<ISceneSpecsOverlay> = ({ scene }) => {
       <span className="overlay-filesize extra-scene-info">
         <FileSize size={file.size} />
       </span>
-      <span className="overlay-resolution">
-        {" "}
-        {TextUtils.resolution(file.width, file.height)}
-      </span>
-      <span className="overlay-duration">
-        {TextUtils.secondsToTimestamp(file.duration)}
-      </span>
+      {file.width && file.height ? (
+        <span className="overlay-resolution">
+          {" "}
+          {TextUtils.resolution(file.width, file.height)}
+        </span>
+      ) : (
+        ""
+      )}
+      {(file.duration ?? 0) >= 1 ? (
+        <span className="overlay-duration">
+          {TextUtils.secondsToTimestamp(file.duration)}
+        </span>
+      ) : (
+        ""
+      )}
     </div>
   );
 };
--- a/ui/v2.5/src/components/Scenes/SceneCard.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneCard.tsx
@@ -352,6 +352,29 @@ const SceneCardOverlays = PatchComponent(
   }
 );
 
+interface ISceneSpecsOverlay {
+  scene: GQL.SlimSceneDataFragment;
+}
+
+export const SceneSpecsOverlay: React.FC<ISceneSpecsOverlay> = ({ scene }) => {
+  if (!scene.files.length) return null;
+  let file = scene.files[0];
+  return (
+    <div className="scene-specs-overlay">
+      <span className="overlay-filesize extra-scene-info">
+        <FileSize size={file.size} />
+      </span>
+      <span className="overlay-resolution">
+        {" "}
+        {TextUtils.resolution(file.width, file.height)}
+      </span>
+      <span className="overlay-duration">
+        {TextUtils.secondsToTimestamp(file.duration)}
+      </span>
+    </div>
+  );
+};
+
 const SceneCardImage = PatchComponent(
   "SceneCard.Image",
   (props: ISceneCardProps) => {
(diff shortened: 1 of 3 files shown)


smith113-p:
Looks like this:

Screenshot 2026-03-09 142931

ui/v2.5/src/components/Scenes/SceneCard.tsx
```diff
@@ -352,6 +352,29 @@ const SceneCardOverlays = PatchComponent(
...
+  scene: GQL.SlimSceneDataFragment;
+}
+
+export const SceneSpecsOverlay: React.FC<ISceneSpecsOverlay> = ({ scene }) => {
+  if (!scene.files.length) return null;
+  let file = scene.files[0];
+  return (
+    <div className="scene-specs-overlay">
+      <span className="overlay-filesize extra-scene-info">
+        <FileSize size={file.size} />
```
Gykes, line 365: The old code had conditional statements to display these. Was there a reason they were removed?

Gykes (changes requested):
Static review only

ui/v2.5/src/components/Scenes/SceneCard.tsx
smith113-p: Yes, I replaced them with a `if (!scene.files.length) return null` check at the beginning of the function and otherwise relied on the type system to ensure that everything else can't be null/undefined

Gykes: So the first one is okay but the old code had conditionals to only show if non zero essentially. A value can be 0 (Which is not null) and the old code checked for that.

Gykes: Actually, I take that back the first one isn't okay. Currently all you are doing is a null check on length and if not null proceed with everything. I think it would be better to make them individual checks similar to the old code and display accordingly.

smith113-p: Hm, why is the first one not okay? It previously used `file?.size !== undefined`. The type of SlimSceneDataFragment.files is as below, so the type system does not let files contain a null/undefined entry. So, file?.size can never be undefined.

You're probably right about the resolution/duration, certainly from a type system perspective, although I don't know if any of those can actually be 0. I will restore those checks though.

```ts
  files: Array<{
    __typename?: "VideoFile";
    id: string;
    path: string;
    size: number;
    mod_time: string;
    duration: number;
    video_codec: string;
    audio_codec: string;
    width: number;
    height: number;
    frame_rate: number;
    bit_rate: number;
    fingerprints: Array<{
      __typename?: "Fingerprint";
      type: string;
      value: string;
    }>;
  }>;
```

Gykes: The size one is against undefined. The rest are against an integer. If the duration is < 1 then it will not display. I'm not talking about verifying null checking i'm talking about verifying that duration and width + height are greater than 1.

I think it's there just so it doesn't display a 0. It either display the correct reso/duration or it displays nothing. A 0 in either in those fields could mean bad data/corruption in some way.

smith113-p: You wrote "Actually, I take that back the first one isn't okay" -- was that not referring to the file size check?

I added back the other checks now.

Gykes: > You wrote "Actually, I take that back the first one isn't okay" -- was that not referring to the file size check?

Yes but it was about how you used it, not the code itself. I just didn't like doing a size check and assuming all other data would be correct

How it's used now is a better approach imo. I personally have transcoded files that have a duration and reso that returns 0. This is just transcoding bugs/errors and can happen.

I'm making an assumption on the reasoning behind it but I see no reason to not check for it to display.

smith113-p: I can't quite tell if you're happy with the code now or want me to change anything else?

Gykes: It's fine now lol

Gykes (approved):
LGTM. 

Did some UI testing and seemed to be okay on Waterfox at 1080P and 1440p. 

Edit: The only thing I can think is to try and drop the new fields in the image some. Currently they sit roughly a quarter of the way up the image, might be better to have them in the bottom corner

discourse-stashapp:

https://discourse.stashapp.cc/t/overlay-info/1602/2

discourse-stashapp:

https://discourse.stashapp.cc/t/duration-in-tagger-view/4894/5

PR #6671, opened 2026-03-12
[Feature] Add `Frames` to VideoFile. UI and Graphql

### LINKED ISSUE #6230: Add Frames to all video files

When `ffprobe` is run, it retrieves the number of frames a video has, I have an extension that needs the accurate frame count.
NOTE: using the duration * frame rate does not always give accurate frame count.

1. `internal/manager/generator_sprite.go` -> `NewSpriteGenerator` would not need to calculate frame dynamically, it would be stored on initial scan
2. Future extensions that deal with video editing would benefit from not having to do frame calculations on their own

I could not find any way to add custom data to a `scene`, having custome data for every scene would allow my extension to add frames for each file, but it is a less elegant solution

### Additional context

Framecount is already done in the file `pkg/ffmpeg/ffprobe.go` line 308. This feature would only store it in the database and make it available in the graphql

bob12224:
---
name: Feature Addition
about: Add a feature to this project!
title: "[Feature] Add `Frames` to VideoFile. UI and Graphql"
labels: enhancement
assignees: 'WithoutPants, bnkai, Leopere'

---

# Scope

Grabbing the Frame Count from the FFProbe (which is already done on all Video Files) and saving it into DB.
Then making the value available via GraphQl and in the Scene page, under `File Info`.

## Closes/Fixes Issues

* Fixes #6230 

## Other testing QA Notes

I have a full local dev server running. I have tested the migrations and manually adding files + scan. Frames is correctly calculated and saved into the DB. It is available in graphql, and I also added it to the `File Info` Screen (for Scene).

NOTE: I chose to not add this as a filter or sorting option. My needs are to read the value, so if anyone wants to filter/sort on Frames, I can add it.


docs/DEVELOPMENT.md
```diff
@@ -34,6 +34,8 @@ NOTE: The `make` command in Windows will be `mingw32-make` with MinGW. For examp
 #### Ubuntu
 
 1. Install dependencies: `sudo apt-get install golang git gcc nodejs ffmpeg -y`
+    * To install `pnpm`: `npm install -g pnpm`
```
bob12224, line 37: I can remove this. Added it because the `Pre-requisites` kinda hides this for a new dev

pkg/ffmpeg/ffprobe.go
```diff
@@ -221,6 +221,7 @@ func (f *FFProbe) NewVideoFile(videoPath string) (*VideoFile, error) {
 		"-show_format",
 		"-show_streams",
 		"-show_error",
+		"-count_frames",
```
bob12224, line 224: There might be bigger changes required by this change. Now that we get the FrameCount by default, might be able to delete `GetReadFrameCount` function.

I leave larger refactoring to the maintainer to suggest

feederbox826:
Your PR contains commits by a different user. Please verify that the commits are properly attributed to the authors you want

Gykes:
Deferred till next release. Will review once this release closes out.

bob12224:
> Your PR contains commits by a different user. Please verify that the commits are properly attributed to the authors you want

I have rebased and renamed the author of those commits. Thanks for the catch

pkg/models/jsonschema/scene.go
```diff
@@ -31,6 +31,7 @@ type SceneFile struct {
 	Width      int           `json:"width"`
 	Height     int           `json:"height"`
 	Framerate  string        `json:"framerate"`
+	Frames     int           `json:"frame"`
```
feederbox826: inconsistent name?

pkg/sqlite/setup_test.go
```diff
@@ -934,6 +925,7 @@ func makeFile(i int) models.File {
 			VideoCodec: getFileStringValue(i, "videoCodec"),
 			AudioCodec: getFileStringValue(i, "audioCodec"),
 			FrameRate:  getFileDuration(i) * 2,
+			Frames:     int64(getFileDuration(i)) * 3,
```
feederbox826: shouldn't this be duration * frameRate?

pkg/sqlite/migrations/86_postmigrate.go
```diff
@@ -0,0 +1,124 @@
...
+				}
+
+				gotSome = true
+				lastID = fileID
+				count++
+
+				path := filepath.Join(dir, basename)
+
+				frames, err := m.ffprobe.GetReadFrameCount(path)
+				if err != nil || frames <= 0 {
```
feederbox826, line 96: is this the equality sign you want?

pkg/sqlite/migrations/85_video_file_frames.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE video_files ADD COLUMN frames INTEGER NOT NULL DEFAULT 0;
```
feederbox826: why not DEFAULT NULL and check for NULL in your postmigration?

feederbox826 (changes requested):
Not sure how useful this is to a broader audience, the ffprobe on post-migration is quite worrying. Frames can be esimated with duration * framerate and unless your entire library is restricted in duration, filters and sorting give you ????

pkg/sqlite/setup_test.go
bob12224: updated

pkg/models/jsonschema/scene.go
bob12224: updated

pkg/sqlite/migrations/86_postmigrate.go
bob12224: `IF (err != nil) || (frames <= 0)` -> do not update row
IF (there was an error) OR (frames is not a valid number)  -> do not update row

if `frames` is less than or equal to 0 that is considered an invalid number.
I could change it to `<`, but 0 frame video file seems invalid...

pkg/sqlite/migrations/85_video_file_frames.up.sql
bob12224: updated to use NULL here and in the postmigrate

bob12224:
> Not sure how useful this is to a broader audience, the ffprobe on post-migration is quite worrying. Frames can be estimated with duration * framerate and unless your entire library is restricted in duration, filters and sorting give you ????

@feederbox826 While frames can be estimated, it is an estimate that can be incorrect (hence the PR). Frames might not be useful to the average user, but it it is critical for anyone using video editors that require frame accuracy (plugin developers). I am using OTIO and it requires accurate frames.

An alternative was described in the original ticket, but it seemed much more ambitious

pkg/sqlite/migrations/86_postmigrate.go
```diff
@@ -0,0 +1,124 @@
...
+	const (
+		limit    = 1000
+		logEvery = 10000
+	)
+
+	result := struct {
+		Count int `db:"count"`
+	}{0}
+
+	if err := m.db.Get(&result, "SELECT COUNT(*) AS count FROM `video_files` WHERE `frames` IS NULL"); err != nil {
```
bob12224, line 45: @feederbox826 I could delete this `86_postmigrate.go` and if people wanted to back fill their frames data, they could re-generate it from the tasks menu?

feederbox826:
> An alternative was described in the original ticket, but it seemed much more ambitious

since then, custom filed support has landed for scenes specifically https://github.com/stashapp/stash/pull/6584 https://github.com/stashapp/stash/pull/6601

would this not address your usecase?

bob12224:
> > An alternative was described in the original ticket, but it seemed much more ambitious
> 
> since then, custom filed support has landed for scenes specifically #6584 #6601
> 
> would this not address your usecase?

custom fields would address my issue.

I hesitate to close this PR though, if anyone else wants to frames per video, they would need to download a plugin and frames would not be available in the file info tab

Gykes:
With only a general code check and not too much research my initial thoughts are this:

So it looks to me like, since this runs on every file, it will significantly increase scan times on every file. If i'm not mistaken `-count_frames` forces ffprobe to scan every single frame no matter what. Would probably be better to make this optional so that users don't all get wrecked with vastly increased scan times because a handful of people want this.

Same as above for the post migration, it will scan every file to generate this data being REALLY slow and people will just be loading on startup as the migration runs. Depending on how large a users library is this could take a while. I think this should be optional if users want to enable it via generate.

Fb's point of custom fields is valid imo. Not too many users will get a usecase from this and would be pretty heavy. Since the userbase is pretty small would a plugin not be valid here?

feederbox826:
> I hesitate to close this PR though, if anyone else wants to frames per video, they would need to download a plugin and frames would not be available in the file info tab

with stuff like HashTheStash that adds additional hashes and with the new fileinfo react hook, it's totally possible to just add frames in as a custom field.

I really don't see a very broad scope and usecase for your average user especially at the cost of the post-migration, moreso if users have offline or disconnected media

bob12224:
Closing the PR now that custom_fields are available for scenes

PR #6673, opened 2026-03-13
feat: add image details to search filter

notsafeforgit:
This pull request adds the image details field to the image search filter, improving discoverability of images based on their description.


WithoutPants (approved):

DogmaDragon:
Needs to be documented in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/Browsing.md.

PR #6675, opened 2026-03-13
feat: Implement Image Duplicate Checker

notsafeforgit:
This pull request introduces a new tool to identify duplicate images based on their perceptual hash (phash). The implementation is based directly on the one used for video pHashes and leverages a shared utility to ensure consistency and performance across the codebase.

It includes:
- Unified backend implementation for phash distance comparison and grouping.
- GraphQL schema updates and API resolvers.
- Enhanced frontend UI with pagination, group sorting, and detailed image previews.
- Unit tests for the image search and duplicate detection logic.

Example screenshot:
download

### DIFF
--- a/internal/api/resolver_query_find_image.go
+++ b/internal/api/resolver_query_find_image.go
@@ -135,6 +135,13 @@ func (r *queryResolver) AllImages(ctx context.Context) (ret []*models.Image, err
 	return ret, nil
 }
 
-func (r *queryResolver) FindDuplicateImages(ctx context.Context, distance int) ([][]*models.Image, error) {
-	return r.repository.Image.FindDuplicates(ctx, distance)
+func (r *queryResolver) FindDuplicateImages(ctx context.Context, distance int) (ret [][]*models.Image, err error) {
+	if err := r.withReadTxn(ctx, func(ctx context.Context) error {
+		ret, err = r.repository.Image.FindDuplicates(ctx, distance)
+		return err
+	}); err != nil {
+		return nil, err
+	}
+
+	return ret, nil
 }
--- a/internal/api/resolver_query_find_image.go
+++ b/internal/api/resolver_query_find_image.go
@@ -134,3 +134,7 @@ func (r *queryResolver) AllImages(ctx context.Context) (ret []*models.Image, err
 
 	return ret, nil
 }
+
+func (r *queryResolver) FindDuplicateImages(ctx context.Context, distance int) ([][]*models.Image, error) {
+	return r.repository.Image.FindDuplicates(ctx, distance)
+}
--- a/pkg/sqlite/image.go
+++ b/pkg/sqlite/image.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
-	"slices"
 	"strconv"
 	"strings"
 
@@ -412,6 +411,11 @@ func (qb *ImageStore) Find(ctx context.Context, id int) (*models.Image, error) {
 func (qb *ImageStore) FindMany(ctx context.Context, ids []int) ([]*models.Image, error) {
 	images := make([]*models.Image, len(ids))
 
+	idToIndex := make(map[int]int, len(ids))
+	for i, id := range ids {
+		idToIndex[id] = i
+	}
+
 	if err := batchExec(ids, defaultBatchSize, func(batch []int) error {
 		q := qb.selectDataset().Prepared(true).Where(qb.table().Col(idColumn).In(batch))
 		unsorted, err := qb.getMany(ctx, q)
@@ -420,8 +424,9 @@ func (qb *ImageStore) FindMany(ctx context.Context, ids []int) ([]*models.Image,
 		}
 
 		for _, s := range unsorted {
-			i := slices.Index(ids, s.ID)
-			images[i] = s
+			if i, ok := idToIndex[s.ID]; ok {
+				images[i] = s
+			}
 		}
 
 		return nil
@@ -1099,14 +1104,20 @@ func (qb *ImageStore) GetURLs(ctx context.Context, imageID int) ([]string, error
 
 var findExactImageDuplicateQuery = `
 SELECT GROUP_CONCAT(DISTINCT image_id) as ids
-FROM images_files
-JOIN files_fingerprints ON images_files.file_id = files_fingerprints.file_id
-WHERE files_fingerprints.type = 'phash' 
-  AND files_fingerprints.fingerprint != zeroblob(8)
-  AND files_fingerprints.fingerprint != ''
-GROUP BY fingerprint
+FROM (
+	SELECT images_files.image_id
+		 , files.size as file_size
+		 , files_fingerprints.fingerprint as phash
+	FROM images_files
+	JOIN files ON images_files.file_id = files.id
+	JOIN files_fingerprints ON images_files.file_id = files_fingerprints.file_id
+	WHERE files_fingerprints.type = 'phash' 
+	  AND files_fingerprints.fingerprint != zeroblob(8)
+	  AND files_fingerprints.fingerprint != ''
+)
+GROUP BY phash
 HAVING COUNT(DISTINCT image_id) > 1
-LIMIT 1000;
+ORDER BY SUM(file_size) DESC;
 `
 
 func (qb *ImageStore) FindDuplicates(ctx context.Context, distance int) ([][]*models.Image, error) {
--- a/pkg/sqlite/image.go
+++ b/pkg/sqlite/image.go
@@ -838,7 +838,7 @@ func (qb *ImageStore) makeQuery(ctx context.Context, imageFilter *models.ImageFi
 		)
 
 		filepathColumn := "folders.path || '" + string(filepath.Separator) + "' || files.basename"
-		searchColumns := []string{"images.title", "images.details", filepathColumn, "files_fingerprints.fingerprint"}
+		searchColumns := []string{"images.title", filepathColumn, "files_fingerprints.fingerprint"}
 		query.parseQueryString(searchColumns, *q)
 	}
 
@@ -1104,28 +1104,29 @@ func (qb *ImageStore) FindDuplicates(ctx context.Context, distance int) ([][]*mo
         WHERE files_fingerprints.type = 'phash'`
 
 	var hashes []*utils.Phash
-	err := imageRepository.queryStruct(ctx, query, nil, &hashes)
-	if err != nil {
-		return nil, err
-	}
+	if err := imageRepository.queryFunc(ctx, query, nil, false, func(rows *sqlx.Rows) error {
+		phash := utils.Phash{
+			Bucket:   -1,
+			Duration: -1,
+		}
+		if err := rows.StructScan(&phash); err != nil {
+			return err
+		}
 
-	for _, h := range hashes {
-		h.Bucket = -1
+		hashes = append(hashes, &phash)
+		return nil
+	}); err != nil {
+		return nil, err
 	}
 
 	dupeIds := utils.FindDuplicates(hashes, distance, -1)
 
 	var result [][]*models.Image
 	for _, comp := range dupeIds {
-		var group []*models.Image
-		for _, id := range comp {
-			img, err := qb.Find(ctx, id)
-			if err == nil && img != nil {
-				group = append(group, img)
+		if images, err := qb.FindMany(ctx, comp); err == nil {
+			if len(images) > 1 {
+				result = append(result, images)
 			}
-		}
-		if len(group) > 1 {
-			result = append(result, group)
 		}
 	}
 
--- a/pkg/sqlite/image.go
+++ b/pkg/sqlite/image.go
@@ -1096,10 +1096,6 @@ func (qb *ImageStore) GetURLs(ctx context.Context, imageID int) ([]string, error
 }
 
 func (qb *ImageStore) FindDuplicates(ctx context.Context, distance int) ([][]*models.Image, error) {
-	return qb.findPhashMatches(ctx, distance)
-}
-
-func (qb *ImageStore) findPhashMatches(ctx context.Context, distance int) ([][]*models.Image, error) {
 	query := `
         SELECT images.id, files_fingerprints.fingerprint as phash
         FROM images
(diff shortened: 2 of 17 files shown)


Gykes:
I'm getting build errors when trying to test. 

```
gykes@Mac-Air stash % make server-start
cd .local && go run -v -tags " sqlite_stat4 sqlite_math_functions"  -ldflags " -X 'github.com/stashapp/stash/internal/build.buildstamp=2026-03-13 16:05:16' -X 'github.com/stashapp/stash/internal/build.githash=af75c8c1b' -X 'github.com/stashapp/stash/internal/build.version=v0.30.1-122-gaf75c8c1b' -X 'github.com/stashapp/stash/internal/build.officialBuild=false'" ../cmd/stash
github.com/stashapp/stash/pkg/sqlite
# github.com/stashapp/stash/pkg/sqlite
../pkg/sqlite/image.go:10:2: "strconv" imported and not used
../pkg/sqlite/image.go:1106:16: undefined: utils
../pkg/sqlite/image.go:1116:13: undefined: utils
make: *** [server-start] Error 1
```

internal/api/resolver_query_find_image.go
```diff
@@ -134,3 +134,7 @@ func (r *queryResolver) AllImages(ctx context.Context) (ret []*models.Image, err
 
 	return ret, nil
 }
+
+func (r *queryResolver) FindDuplicateImages(ctx context.Context, distance int) ([][]*models.Image, error) {
```
Gykes: This needs to be wrapped in `r.withReadTxn()` like every other query resolver in this file. Without it, there's no database transaction in the context and the query will fail. See `FindDuplicateScenes` in `resolver_query_find_scene.go` for the correct pattern.

pkg/sqlite/image.go
```diff
@@ -1093,3 +1094,41 @@ func (qb *ImageStore) UpdateTags(ctx context.Context, imageID int, tagIDs []int)
...
+func (qb *ImageStore) FindDuplicates(ctx context.Context, distance int) ([][]*models.Image, error) {
+	query := `
+        SELECT images.id, files_fingerprints.fingerprint as phash
+        FROM images
+        JOIN images_files ON images.id = images_files.image_id
+        JOIN files_fingerprints ON images_files.file_id = files_fingerprints.file_id
+        WHERE files_fingerprints.type = 'phash'`
+
+	var hashes []*utils.Phash
+	err := imageRepository.queryStruct(ctx, query, nil, &hashes)
```
Gykes: `queryStruct` wraps `dbWrapper.Get()`, which is designed for single-row results. This will fail at runtime when the query returns multiple rows. The scene duplicate finder uses `queryFunc` with rows. Probably a good idea to implement this.

pkg/sqlite/image.go
```diff
@@ -1093,3 +1094,41 @@ func (qb *ImageStore) UpdateTags(ctx context.Context, imageID int, tagIDs []int)
...
+		h.Bucket = -1
+	}
+
+	dupeIds := utils.FindDuplicates(hashes, distance, -1)
+
+	var result [][]*models.Image
+	for _, comp := range dupeIds {
+		var group []*models.Image
+		for _, id := range comp {
+			img, err := qb.Find(ctx, id)
```
Gykes: This is an N+1 query, calling `qb.Find` individually for each duplicate image. The scene version uses `qb.FindMany(ctx, sceneIds)` to batch-load the group.

pkg/sqlite/image.go
```diff
@@ -837,7 +838,7 @@ func (qb *ImageStore) makeQuery(ctx context.Context, imageFilter *models.ImageFi
 		)
 
 		filepathColumn := "folders.path || '" + string(filepath.Separator) + "' || files.basename"
-		searchColumns := []string{"images.title", filepathColumn, "files_fingerprints.fingerprint"}
+		searchColumns := []string{"images.title", "images.details", filepathColumn, "files_fingerprints.fingerprint"}
```
Gykes: This appears to be from your other PR. I would extract this out and rebase off of develop.

Gykes (changes requested):
Just an initial static check. Once the build issues are fixed I can do another, and more detailed, review

ui/v2.5/src/locales/en-US.json
```diff
@@ -9,7 +9,9 @@
         "tools": {
             "scene_filename_parser": {
                 "ignore_organized": "Ignore organized scenes"
-            }
+            },
+            "image_tools": "Image Tools",
+            "image_duplicate_checker": "Image Duplicate Checker"
```
DogmaDragon: This change serves no purpose. Values are identical to `en-GB.json` which is default language. Also, any changes to different locales should be managed [through Codeberg Translate](https://translate.codeberg.org/projects/stash/stash/).

Gykes:
UI review: 

The initial page in settings -> Tools looks okay. The only thing here is that the casing of all the settings are bad but I know that's more of @DogmaDragon decision on how they want it as I've seen few PRs from them correcting issues like this. 
Screenshot 2026-03-14 at 11 14 32

Inside the image checker is where there are issues. It's white and looks nothing like the scene one. Is there a reason it looks so different from the scenes variation? There are also a lot of missing features. Search Accuracy, Search options, things of that nature.
Screenshot 2026-03-14 at 11 14 39

The final issue is when I click search I get a 400 error. I do have several images in my dev environment but non have hashes. If the images do not have hashes it should be handled appropriately and not 400 erroring  
Screenshot 2026-03-14 at 11 18 22

If you are using AI to build this, which I am not against, please make sure you are testing these features before PRing and requesting reviews from people. Both of the major issues I have pointed out could have been solved by you by just starting the app, seeing it fail, then testing the feature a little. 

Thank you for the work, and I am 100% for this feature, but I believe it needs more polish and work before I could give it an approval.

sleetx:
> Inside the image checker is where there are issues. It's white and looks nothing like the scene one. Is there a reason it looks so different from the scenes variation? There are also a lot of missing features. Search Accuracy, Search options, things of that nature. 

To be fair, the UI of the scene checker is not exactly great either, so I'm open to seeing changes if they make sense. But to your point, the color scheme is off and there are a lot of missing features that the scene checker currently has.

Gykes:
> To be fair, the UI of the scene checker is not exactly great either, so I'm open to seeing changes if they make sense. But to your point, the color scheme is off and there are a lot of missing features that the scene checker currently has.

Agreed, i'm definitely open to trying new things if there was an intent. This looks more like an over site rather than a deliberate attempt to refactor the UI to be a better UX. If OP wants to modify UX for this then I am happy to have it done and potentially port it over to scenes

notsafeforgit:
All UI issues mentioned should be fixed now, I have attached a screenshot. Polished it up quite a bit and hopefully it's closer to meeting the standards for stash now!

pkg/sqlite/image.go
```diff
@@ -837,7 +845,7 @@ func (qb *ImageStore) makeQuery(ctx context.Context, imageFilter *models.ImageFi
 		)
 
 		filepathColumn := "folders.path || '" + string(filepath.Separator) + "' || files.basename"
-		searchColumns := []string{"images.title", "images.details", filepathColumn, "files_fingerprints.fingerprint"}
+		searchColumns := []string{"images.title", filepathColumn, "files_fingerprints.fingerprint"}
```
Gykes, line 848: This removes details from the search. This seems like a regression

pkg/sqlite/image_test.go
```diff
@@ -1596,20 +1596,6 @@ func TestImageQueryQ(t *testing.T) {
...
-		q := getImageStringValue(imageIdx, detailsField)
-
-		sqb := db.Image
-
-		imageQueryQ(ctx, t, sqb, q, imageIdx)
-
-		return nil
-	})
-}
-
```
Gykes, line 1612: After you add details back on `image.go` you can add this back.

pkg/sqlite/image.go
```diff
@@ -1093,3 +1101,94 @@ func (qb *ImageStore) UpdateTags(ctx context.Context, imageID int, tagIDs []int)
...
+		return nil, err
+	}
+
+	var result [][]*models.Image
+	offset := 0
+	for _, comp := range dupeIds {
+		group := allImages[offset : offset+len(comp)]
+		result = append(result, group)
+		offset += len(comp)
+	}
```
Gykes, line 1191: So, if I am reading this properly there could potentially be a smell here. If a user kicks off a check on one tab and modifies/deletes an image in another before the scan completes then it would pop as nil. I'm not 100% sure if the downstream can handle this. We might want to filter nil entries when building the groups. I think this would only really be a concern for people with super large DBs where the scan take a minute.

I'm not 100% on this one so feel free top educate me if i'm misunderstanding. This could also be an issue with scenes, and if so, we can differ the fix to another PR for both.

ui/v2.5/src/components/ImageDuplicateChecker/ImageDuplicateChecker.tsx
```diff
@@ -0,0 +1,712 @@
...
+      return (
+        <option key={size} value={size}>
+          {size}
+        </option>
+      );
+    });
+  }, [allGroups.length]);
+
+  if (loading) return <LoadingIndicator />;
+  if (!data) return <ErrorMessage error="Error searching for duplicates." />;
```
Gykes, line 179: Internationilize

ui/v2.5/src/components/ImageDuplicateChecker/ImageDuplicateChecker.tsx
```diff
@@ -0,0 +1,712 @@
...
+                      </tr>
+                    </React.Fragment>
+                  );
+                })
+              )}
+            </tbody>
+          </Table>
+
+          {allGroups.length === 0 && !loading && (
+            <h4 className="text-center mt-4">No duplicates found.</h4>
```
Gykes, line 695: Internationilize

ui/v2.5/src/components/ImageDuplicateChecker/ImageDuplicateChecker.tsx
```diff
@@ -0,0 +1,712 @@
...
+          )}
+
+          {loading && (
+            <div className="text-center mt-4">
+              <Icon icon={faBox} spin className="fa-3x" />
+              <h4 className="mt-2">Loading...</h4>
```
Gykes, line 701: Internationilize

ui/v2.5/src/components/ImageDuplicateChecker/ImageDuplicateChecker.tsx
```diff
@@ -0,0 +1,712 @@
...
+            </div>
+          )}
```
Gykes, line 703: I think this whole block is just unreachable, no? The component already returns the loading indicator when true.

ui/v2.5/src/locales/en-GB.json
```diff
@@ -1133,12 +1136,12 @@
...
-    "select_all_but_largest_file": "Select every file in each duplicated group, except the largest file",
-    "select_all_but_largest_resolution": "Select every file in each duplicated group, except the file with highest resolution",
+    "select_all_but_largest_file": "Keep the largest file (select all but the largest)",
+    "select_all_but_largest_resolution": "Keep the highest resolution file (select all but the highest resolution)",
     "select_none": "Select None",
-    "select_oldest": "Select the oldest file in the duplicate group",
+    "select_oldest": "Keep the oldest file (select all but the oldest)",
     "select_options": "Select Options…",
-    "select_youngest": "Select the youngest file in the duplicate group",
+    "select_youngest": "Keep the youngest file (select all but the youngest)",
```
Gykes, line 1144: Was there a reason to modify these? I'm not too fussed but @DogmaDragon can give his opinion on how they want it done.

pkg/sqlite/image_filter.go
```diff
@@ -185,6 +185,10 @@ func (qb *imageFilterHandler) missingCriterionHandler(isMissing *string) criteri
 			case "tags":
 				imageRepository.tags.join(f, "tags_join", "images.id")
 				f.addWhere("tags_join.image_id IS NULL")
+			case "phash":
+				f.addInnerJoin("images_files", "", "images_files.image_id = images.id")
+				f.addLeftJoin(fingerprintTable, "fingerprints_phash", "images_files.file_id = fingerprints_phash.file_id AND fingerprints_phash.type = 'phash'")
+				f.addWhere("fingerprints_phash.fingerprint IS NULL")
```
Gykes, line 191: You added the filter but I don't see the front end side for it so wont be usable on the UI.

pkg/sqlite/image.go
```diff
@@ -1093,3 +1101,94 @@ func (qb *ImageStore) UpdateTags(ctx context.Context, imageID int, tagIDs []int)
 func (qb *ImageStore) GetURLs(ctx context.Context, imageID int) ([]string, error) {
 	return imagesURLsTableMgr.get(ctx, imageID)
 }
+
+var findExactImageDuplicateQuery = `
```
Gykes, line 1105: So, I had some thoughts on this: 

I know GROUP_CONCAT can truncate. What happens if a user has a lot of dupes would we lose the IDs? We might want to add some test cases for this. 

Wouldn't it also be faster to just return the IDs via the row directly rather than going through `string.split` + `strconv.Atoi`?

pkg/sqlite/image.go
```diff
@@ -1093,3 +1101,94 @@ func (qb *ImageStore) UpdateTags(ctx context.Context, imageID int, tagIDs []int)
...
+	WHERE files_fingerprints.type = 'phash' 
+	  AND files_fingerprints.fingerprint != zeroblob(8)
+	  AND files_fingerprints.fingerprint != ''
+)
+GROUP BY phash
+HAVING COUNT(DISTINCT image_id) > 1
+ORDER BY SUM(file_size) DESC;
+`
+
+func (qb *ImageStore) FindDuplicates(ctx context.Context, distance int) ([][]*models.Image, error) {
```
Gykes, line 1123: Your PR description said that it added tests but I dont see any. 

We would need test cases for the following:

Exact match path, Hammering-distance, and probably a regression test for the slice-offset group reconstruction when/if `FindMany returns nil.

Gykes (changes requested):
This is just more static review.

ui/v2.5/src/locales/en-GB.json
notsafeforgit: I thought that the old copy was rather confusing as "select the oldest" would actually check all the boxes of the "youngest" files and thus *keep* the oldest. So I tried to make it very clear what it would actually do.

spaceyuck:
Why was this abandoned? I was hoping at least for the server / JQL side of it for plugins.

notsafeforgit:
@spaceyuck because I'm just folding it in to my fork in which I've rebuilt the entire UI into a v3 from the ground up in react v19 with tanstack router, shadcn + baseui components, and many more nice UI features and enhancements. I'm not trying to release it publicly but it was taking too long to get the things I wanted into stash and by keeping a fork I can just rebase on top of the upstream occasionally, I can move much faster. Hopefully things in this PR can still be leveraged if helpful. My fork is here, https://github.com/notsafeforgit/stash with my hardware accelerated fork of stash-s6 here: https://github.com/notsafeforgit/stash-s6 which has image builds you can use (only linux amd64 for now). I have left the v2.5 UI fully backwards compatible and untouched, and you can enable V3 ui by running it with the env var `STASH_ENABLE_V3_UI=true`. Warning, I need to do some additional db migrations to support my additional features that aren't merged into upstream so make a database backup if you really wanted to use this.

PR #6680, opened 2026-03-15
Treat bad network path error as non-existing folder during folder rename check

WithoutPants:
Fixes issue on Windows where a bad network path error would prevent the folder rename check from completing.


Gykes:
I know something similar to this is handled in `scan.go` Would registering all `Lstat` errors be a better approach than just the ones that are baked into this fix?

WithoutPants:
> I know something similar to this is handled in `scan.go` Would registering all `Lstat` errors be a better approach than just the ones that are baked into this fix?

Yeah I think you're right. I wanted to isolate file not exist errors from other errors during rename checking, but I think just checking for any error should be sufficient.

PR #6682, opened 2026-03-15
Convert career length fields to dates

### LINKED ISSUE #6667: Career Start / End should be dates not integers.

Allow career start and career end fields to use the same imprecise/date types Stash supports for birth/death (e.g., full dates, partial dates like YYYY-MM or YYYY-MM-DD, and imprecise ranges) instead of being limited to integers (years).

- Preserves more accurate data for cam/OnlyFans/models whose careers have known month/day.
- Avoids users storing human-readable date strings in the integer fields or in notes.
- Enables consistent handling/validation, searching, filtering, and display of career ranges (e.g., "2017-01-16 to 2020-03-16 (3y 2m)").
- Improves interoperability with other StashDB-backed tools that already handle imprecise dates for birth/death.

No—current development builds expose career start/end as integers (years). Workarounds require storing precise dates in free-text fields or notes, which prevents proper querying, sorting, and filtering.

### Additional context

Recommend reusing Stash's existing imprecise-date schema and UI components used for birth/death to avoid duplication. If database schema constraints prevent changing integer fields, consider adding optional career_start_exact and career_end_exact imprecise-date fields while keeping the integer year fields for backward compatibility.

WithoutPants:
Resolves #6667 

Includes schema migration to convert career start/end to dates with year precision.

Verified that scraping works as expected with existing performer scrapers.


feederbox826:
Found? an obscure edge case for scraping - Will try to confirm and narrow it down for you

feederbox826:
If a scraper only returns one of CareerStart/ CareerEnd, the scraper terminates with "nil pointer dereference". This can be mitigated by passing in an invalid CareerLength that is not parsed and is not taken into account for data input, EVEN if it covers fields not covered

here's the scraper
```yml
name: break-6682

performerByURL:
  - action: scrapeXPath
    url:
      - https://stashapp.cc
      - https://example.org
    scraper: performerScraper

xPathScrapers:
  performerScraper:
    performer:
      Name:
        fixed: Performer
      CareerLength:
        fixed: 1900-2020
      CareerStart:
        fixed: 2000
      CareerEnd: 
        fixed: 2010
```

(stashapp.cc since the cert for example.org throws errors)

| Length | Start | End | error | Start | End |
|---|---|---|---|---|---|
| ❌ | ✅ | ✅ |  | 2000 | 2010 |
| ✅ | ✅ | ✅ |  | 2000 | 2010 |
| ❌ | ❌ | ✅ | nil pointer |  |  |
| ✅ | ❌ | ✅ |  |  | 2010 |
| ✅ | ✅ | ❌ |  | 2000 |  |
| ❌ | ✅ | ❌ | nil pointer |  |  |
| ✅ | ❌ | ❌ |  | 1900 | 2020 |
| ❌ | ✅ | ✅ |  | 2000 | 2010 |

✅ means the property is un-commented
❌ means the property is commented out

A lot of scrapers only present one useful date [IAFD always goes until current year](https://github.com/stashapp/CommunityScrapers/issues/2584), so this is a valid usecase

pkg/scraper/postprocessing.go
```diff
@@ -126,17 +126,35 @@ func (c *postScraper) postScrapePerformer(ctx context.Context, p models.ScrapedP
...
 
 	// populate career start/end from career length and vice versa
-	if !isEmptyStr(p.CareerLength) && isEmptyInt(p.CareerStart) && isEmptyInt(p.CareerEnd) {
-		p.CareerStart, p.CareerEnd, err = utils.ParseYearRangeString(*p.CareerLength)
+	if !isEmptyStr(p.CareerLength) && isEmptyStr(p.CareerStart) && isEmptyStr(p.CareerEnd) {
+		start, end, err := models.ParseYearRangeString(*p.CareerLength)
 		if err != nil {
 			logger.Warnf("Could not parse career length %s: %v", *p.CareerLength, err)
+		} else {
+			startStr := start.String()
```
Gykes, line 145: `ParseYearRangeString` can return start or end as nil. Both are dereferenced unconditionally causing a panic. I think we need to add a gaurd before calling a `.string()`. I think this is whats causing FBs errors.

pkg/scraper/postprocessing.go
```diff
@@ -126,17 +126,35 @@ func (c *postScraper) postScrapePerformer(ctx context.Context, p models.ScrapedP
...
+			endStr := end.String()
+			p.CareerStart = &startStr
+			p.CareerEnd = &endStr
+		}
+	} else if isEmptyStr(p.CareerLength) && (!isEmptyStr(p.CareerStart) || !isEmptyStr(p.CareerEnd)) {
+		ok := true
+		start, err := models.ParseDate(*p.CareerStart)
```
Gykes: Think we need another gaurd here for dereferenced.

internal/api/resolver_mutation_performer.go
```diff
@@ -100,6 +89,25 @@ func (r *mutationResolver) PerformerCreate(ctx context.Context, input models.Per
...
+	if newPerformer.CareerStart == nil && newPerformer.CareerEnd == nil && input.CareerLength != nil {
+		start, end, err := models.ParseYearRangeString(*input.CareerLength)
+		if err != nil {
+			return nil, fmt.Errorf("could not parse career_length %q: %w", *input.CareerLength, err)
+		}
+		newPerformer.CareerStart = start
+		newPerformer.CareerEnd = end
+	}
+
+	newPerformer.CareerStart, err = translator.datePtr(input.CareerStart)
```
Gykes: So, does this not unconditionally overwrite the `career_length` fallback set on 98+99. 

If someone only sends `career_length` then `datePtr` would return double nill which would wipe out the fallback.

ui/v2.5/src/components/Performers/EditPerformersDialog.tsx
```diff
@@ -428,21 +428,23 @@ export const EditPerformersDialog: React.FC<IListOperationProps> = (
...
-              value={updateInput.career_start?.toString()}
-              valueChanged={(v) =>
-                setUpdateField({ career_start: v ? parseInt(v) : undefined })
+            <BulkUpdateDateInput
+              value={updateInput.career_start}
+              valueChanged={(newValue) =>
+                setUpdateField({ career_start: newValue })
               }
               unsetDisabled={unsetDisabled}
+              error={birthdateError}
```
Gykes: Need correct states with the correct hooks for this and line 447

ui/v2.5/src/components/Performers/EditPerformersDialog.tsx
```diff
@@ -428,21 +428,23 @@ export const EditPerformersDialog: React.FC<IListOperationProps> = (
...
-              value={updateInput.career_end?.toString()}
-              valueChanged={(v) =>
-                setUpdateField({ career_end: v ? parseInt(v) : undefined })
+            <BulkUpdateDateInput
+              value={updateInput.career_end}
+              valueChanged={(newValue) =>
+                setUpdateField({ career_end: newValue })
               }
               unsetDisabled={unsetDisabled}
+              error={birthdateError}
```
Gykes: From above.

Gykes (changes requested):
Static review only.

internal/api/resolver_mutation_performer.go
WithoutPants: This was ordered incorrectly

pkg/sqlite/scene_test.go
```diff
@@ -2821,6 +2821,33 @@ func verifyIntPtr(t *testing.T, value *int, criterion models.IntCriterionInput)
...
 }
 
+func verifyDatePtr(t *testing.T, value *models.Date, criterion models.DateCriterionInput) {
+	t.Helper()
+	assert := assert.New(t)
+	if criterion.Modifier == models.CriterionModifierIsNull {
+		assert.Nil(value, "expect is null values to be null")
+	}
+	if criterion.Modifier == models.CriterionModifierNotNull {
+		assert.NotNil(value, "expect is null values to be null")
```
Gykes:
```suggestion
		assert.NotNil(value, "expect not null values to be not null")
```

Gykes (changes requested):

Gykes (approved):
Static testing only, didn't run it through a DB but logic tracks.

DogmaDragon:
Needs to be documented in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/ScraperDevelopment.md.

PR #6683, opened 2026-03-16
Patch scene specs

feederbox826:
Fixes """regression""" from https://github.com/stashapp/stash/pull/6663 + memoization which broke https://discourse.stashapp.cc/t/overlay-info/1602 (to more dismay than I imagined)


feederbox826:
Rebasing, my fork was out of date

PR #6684, opened 2026-03-16
add patching for scene-specs-overlay

feederbox826:
Rebased twice for good luck (#6683)

fixed """regression""" from https://github.com/stashapp/stash/pull/6663 + memoization which broke https://discourse.stashapp.cc/t/overlay-info/1602 to more dismay than I expected

cc @smith113-p

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneCard.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneCard.tsx
@@ -359,7 +359,6 @@ interface ISceneSpecsOverlay {
 export const SceneSpecsOverlay: React.FC<ISceneSpecsOverlay> = PatchComponent(
   "SceneCard.SceneSpecs",
   ({ scene }) => {
-    // try with more effort
     const file = scene.files?.[0];
     if (!file) return null;
     return (
--- a/ui/v2.5/src/components/Scenes/SceneCard.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneCard.tsx
@@ -356,32 +356,35 @@ interface ISceneSpecsOverlay {
   scene: GQL.SlimSceneDataFragment;
 }
 
-export const SceneSpecsOverlay: React.FC<ISceneSpecsOverlay> = ({ scene }) => {
-  if (!scene.files.length) return null;
-  let file = scene.files[0];
-  return (
-    <div className="scene-specs-overlay">
-      <span className="overlay-filesize extra-scene-info">
-        <FileSize size={file.size} />
-      </span>
-      {file.width && file.height ? (
-        <span className="overlay-resolution">
-          {" "}
-          {TextUtils.resolution(file.width, file.height)}
+export const SceneSpecsOverlay: React.FC<ISceneSpecsOverlay> = PatchComponent(
+  "SceneCard.SceneSpecs",
+  ({ scene }) => {
+    // try with more effort
+    const file = scene.files?.[0];
+    if (!file) return null;
+    return (
+      <div className="scene-specs-overlay">
+        <span className="overlay-filesize extra-scene-info">
+          <FileSize size={file.size} />
         </span>
-      ) : (
-        ""
-      )}
-      {(file.duration ?? 0) >= 1 ? (
-        <span className="overlay-duration">
-          {TextUtils.secondsToTimestamp(file.duration)}
-        </span>
-      ) : (
-        ""
-      )}
-    </div>
-  );
-};
+        {file.width && file.height ? (
+          <span className="overlay-resolution">
+            {TextUtils.resolution(file.width, file.height)}
+          </span>
+        ) : (
+          ""
+        )}
+        {file.duration > 0 ? (
+          <span className="overlay-duration">
+            {TextUtils.secondsToTimestamp(file.duration)}
+          </span>
+        ) : (
+          ""
+        )}
+      </div>
+    );
+  }
+);
 
 const SceneCardImage = PatchComponent(
   "SceneCard.Image",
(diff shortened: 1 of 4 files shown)


ui/v2.5/src/components/Scenes/SceneCard.tsx
```diff
@@ -356,32 +356,35 @@ interface ISceneSpecsOverlay {
...
-  if (!scene.files.length) return null;
-  let file = scene.files[0];
-  return (
-    <div className="scene-specs-overlay">
-      <span className="overlay-filesize extra-scene-info">
-        <FileSize size={file.size} />
-      </span>
-      {file.width && file.height ? (
-        <span className="overlay-resolution">
-          {" "}
```
feederbox826, line 369: The additional space was there for https://github.com/stashapp/stash/pull/6684/changes#diff-9956a6abfeb93196964da0d85ed2061f707ba675e2dc73198ba6edd02719fc3aL203

Without a whitespace, the lines would be squashed together. This was fixed with a CSS style instead of adding in arbitrary whitespaces

ui/v2.5/src/components/Scenes/SceneCard.tsx
```diff
@@ -356,32 +356,35 @@ interface ISceneSpecsOverlay {
...
-      ) : (
-        ""
-      )}
-      {(file.duration ?? 0) >= 1 ? (
-        <span className="overlay-duration">
-          {TextUtils.secondsToTimestamp(file.duration)}
+export const SceneSpecsOverlay: React.FC<ISceneSpecsOverlay> = PatchComponent(
+  "SceneCard.SceneSpecs",
+  ({ scene }) => {
+    // try with more effort
```
WithoutPants: Unnecessary comment

WithoutPants (changes requested):

ui/v2.5/src/components/Scenes/SceneCard.tsx
```diff
@@ -356,32 +356,35 @@ interface ISceneSpecsOverlay {
...
-  );
-};
+        {file.width && file.height ? (
+          <span className="overlay-resolution">
+            {TextUtils.resolution(file.width, file.height)}
+          </span>
+        ) : (
+          ""
+        )}
+        {file.duration > 0 ? (
```
feederbox826, line 376: No need for Nullish Coalescing (??) since we already validate that file.duration is not undefined, the condition should never fire (which was not true previously)

single comparison for > 0 is faster than >= 1 [benchmark](https://jsperf.app/zoseva). Saves microseconds since we save one comparison and most of our durations are expected to be greater than 1, which skews test results towards greater vs greaterEqual

ui/v2.5/src/components/Scenes/SceneCard.tsx
```diff
@@ -356,32 +356,35 @@ interface ISceneSpecsOverlay {
...
-      )}
-      {(file.duration ?? 0) >= 1 ? (
-        <span className="overlay-duration">
-          {TextUtils.secondsToTimestamp(file.duration)}
+export const SceneSpecsOverlay: React.FC<ISceneSpecsOverlay> = PatchComponent(
+  "SceneCard.SceneSpecs",
+  ({ scene }) => {
+    // try with more effort
+    const file = scene.files?.[0];
+    if (!file) return null;
```
feederbox826, line 363: Insetad of checking for length, just safely pull the first entry and fail if it's invalid

ui/v2.5/src/components/Scenes/styles.scss
```diff
@@ -190,6 +190,10 @@ textarea.scene-description {
   right: 0.7rem;
 }
 
+.scene-specs-overlay > span:not(:last-child) {
```
feederbox826, line 193: `:not(:last-child)` ensures we don't add it to overlay-duration at the end, while also not duplicating code by adding it to overlay-resolution AND overlay-filesize

ui/v2.5/src/docs/en/Manual/UIPluginApi.md
```diff
@@ -311,6 +311,7 @@ Returns `void`.
 - `SceneCard.Image`
 - `SceneCard.Overlays`
 - `SceneCard.Popovers`
+- `SceneCard.SceneSpecs`
```
feederbox826, line 314: Added a new name since Overlays only contains studio overlays

ui/v2.5/src/pluginApi.d.ts
```diff
@@ -746,6 +746,7 @@ declare namespace PluginApi {
     "SceneCard.Image": React.FC<any>;
     "SceneCard.Overlays": React.FC<any>;
     "SceneCard.Popovers": React.FC<any>;
+    "SceneCard.SceneSpecs": React.FC<any>;
```
feederbox826, line 749: Unsure if it should be properly typed?

feederbox826 (commented):
Cursor + Claude Opus was used to generate the initial commit (pre-rebase)

Code post-rebase was ~~lovingly~~ done by yours truly, even if it's mostly copy + paste

UI tested for general stability + extremely basic patcher to validate functionality

```js
const { React, patch } = window.PluginApi

const addOverlayInfo = (props, _, SceneCardSpecs) => {
  console.log(props)
  console.log(SceneCardSpecs)
  return SceneCardSpecs
}

patch.after("SceneCard.SceneSpecs", addOverlayInfo)
```

ui/v2.5/src/pluginApi.d.ts
Gykes: It "should" but that would open a can of worms that I don't want to dive into

Gykes (approved):
lgtm, didn't do any real testing outside of basic UI stuff.

PR #6688, opened 2026-03-17
Add stash ids to performer merge dialog

### LINKED ISSUE #6653: Merging performers does not merge stash IDs
### Describe the bug

When you merge two performers with stash IDs from different stash boxes, I would expect that the stash IDs get merged. Instead only one stash id survives.

### Steps to reproduce

1. Have two performers in stash
2. Have each performer be linked to a different stash box
3. Select both of them and click ... -> merge

### Expected behaviour

Resulting performer should have both stash ids

_No response_

### Stash version

v0.30.1-90-ged58d183

### Device details

_No response_

### Relevant log output

```shell

```

WithoutPants:
Resolves #6653 

Adds stash ids field to the performer merge dialog. Also makes the display of stash ids consistent between other merge dialogs. Changes the logic so that the field will be shown if either source or destination has stash ids (was hiding if they were the same or not present on source).

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerMergeDialog.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerMergeDialog.tsx
@@ -20,13 +20,14 @@ import { faExchangeAlt, faSignInAlt } from "@fortawesome/free-solid-svg-icons";
 import { ScrapeDialog } from "../Shared/ScrapeDialog/ScrapeDialog";
 import {
   ScrapedCustomFieldRows,
+  ScrapeDialogRow,
   ScrapedImageRow,
   ScrapedInputGroupRow,
   ScrapedStringListRow,
   ScrapedTextAreaRow,
 } from "../Shared/ScrapeDialog/ScrapeDialogRow";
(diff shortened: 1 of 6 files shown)


ui/v2.5/src/components/Performers/PerformerMergeDialog.tsx
```diff
@@ -327,6 +334,25 @@ const PerformerMergeDetails: React.FC<IPerformerMergeDetailsProps> = ({
...
+    );
+    setStashIDs(
+      new ScrapeResult(
+        dest.stash_ids,
+        all
+          .map((s) => s.stash_ids)
+          .flat()
+          .filter((s, index, a) => {
+            // remove entries with duplicate endpoints
+            return index === a.findIndex((ss) => ss.endpoint === s.endpoint);
```
smith113-p, line 351: I don't love this for performers... I do sometimes have multiple stash IDs from the same endpoint, especially if I submitted a merge and it hadn't been approved yet, though also with tpdb this can happen with their performer sites, or generally when I don't have enough evidence to submit a merge request.

PR #6694, opened 2026-03-18
feat: add per-alias autotag toggle + general api/app support for per-alias autotag

notsafeforgit:
Add DB support and graphQL API support to track alias matching for a performer being on or off on a per-alias basis. By default, new aliases (especially from scrapers) are set to be ignored for auto-tagging. The UI toggle has been updated to reflect an 'Auto tag' boolean toggle that allows users to opt-in specific aliases.

Attached is a screenshot of the new alias list UI with autotag toggles:

New performer edit page with alias toggles

### DIFF
--- a/pkg/performer/import.go
+++ b/pkg/performer/import.go
@@ -309,7 +309,7 @@ func performerJSONToPerformer(performerJSON jsonschema.Performer) (models.Perfor
 func toPerformerAliases(aliases []string) []models.PerformerAlias {
 	var res []models.PerformerAlias
 	for _, a := range aliases {
-		res = append(res, models.PerformerAlias{Alias: a, IgnoreAutoTag: false})
+		res = append(res, models.PerformerAlias{Alias: a, IgnoreAutoTag: true})
 	}
 	return res
 }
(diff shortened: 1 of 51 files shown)


ui/v2.5/src/utils/form.tsx
```diff
@@ -106,14 +107,34 @@ export function formikUtils<V extends FormikValues>(
...
+            errors.push(err);
+          }
+          errorIdx.push(i);
+        }
+      }
+      return [errors.join("\n"), errorIdx];
+    } else {
+      return [error, undefined];
     }
+  };
```
notsafeforgit: I’ve introduced `getFlattenedError` to centralize how we retrieve and format Formik error messages.

Previously, every field rendering function had to perform manual type assertions (e.g., `as ErrorMessage[]`) to handle Formik's complex error structure. This helper encapsulates that logic in one place, handling both simple string errors and array-based errors for list fields.

This significantly reduces code duplication and keeps the rendering functions clean and type-safe.

pkg/performer/import.go
```diff
@@ -305,3 +305,11 @@ func performerJSONToPerformer(performerJSON jsonschema.Performer) (models.Perfor
 
 	return newPerformer, nil
 }
+
+func toPerformerAliases(aliases []string) []models.PerformerAlias {
+	var res []models.PerformerAlias
+	for _, a := range aliases {
+		res = append(res, models.PerformerAlias{Alias: a, IgnoreAutoTag: true})
```
notsafeforgit, line 312: All current performers don't support autotagging aliases, so backups should be imported with autotag ignored for aliases set to true.

notsafeforgit:
Just wanted to comment that I'm now running a custom fork of stash with this change and have extensively tested this PR and it appears to work great. I have autotag enabled for many aliases for many performers and have seen no issues.

PR #6696, opened 2026-03-18
Fix black thumbnails for images with transparency (ffmpeg path)

### LINKED ISSUE #5850: [Bug Report] Thumbnails of line art images with transparency are unreadable (black thumbnails)
**Describe the bug**
Thumbnails of line art images with transparency are resulted in black thumbnails without transparency. This makes the thumbnails unreadable.

**To Reproduce**
1. Scan a line art image with transparency.
2. Click "Images" and find the line art image to see the thumbnail on the list.

**Expected behavior**
The thumbnail has transparency.

**Screenshots**
Image

v0.28.1

 - OS: Windows 11
 - Browser: Chrome
 - Browser Version: 135.0.7049.115

**Additional context**
None.

dev-null-life:
## Summary

- When ffmpeg converts a PNG/WebP image with an alpha channel to JPEG, transparent areas render as black since JPEG has no alpha support
- Added a `flattenAlphaToWhite` helper that composites the image against a white background using Go's `image/draw` before handing it to ffmpeg
- The vips path already produces the same result (vips `thumbnail` composites against white by default), so this brings the ffmpeg fallback into parity

## Alternative considered: WebP output

Switching the thumbnail format to WebP would preserve transparency rather than flatten it. However, that requires:
- Changing `GetThumbnailPath` from `.jpg` → `.webp` (affects cached files on disk)
- `http.ServeFile` infers `Content-Type` from the file extension, so existing `.jpg`-named files would be misserved if they contained WebP bytes
- Users would need to regenerate all existing thumbnails

Given that the vips path also flattens to white and the path/serving changes are non-trivial, this PR takes the simpler approach. A WebP thumbnail format could be explored as a separate follow-up.

## Test plan

- [ ] Scan a PNG or WebP image with a transparent background (e.g. line art)
- [ ] On a system without vips installed, confirm the generated thumbnail has a white background instead of black
- [ ] Confirm thumbnails for non-transparent images are unchanged

Fixes #5850

### DIFF
--- a/pkg/ffmpeg/options.go
+++ b/pkg/ffmpeg/options.go
@@ -80,6 +80,16 @@ func (a Args) VideoFilter(vf VideoFilter) Args {
 	return append(a, vf.Args()...)
 }
 
+// FilterComplex adds a -filter_complex argument and returns the result.
+func (a Args) FilterComplex(fc string) Args {
+	return append(a, "-filter_complex", fc)
+}
+
+// Map adds a -map argument and returns the result.
+func (a Args) Map(m string) Args {
+	return append(a, "-map", m)
+}
+
 // VSync adds the VsyncMethod and returns the result.
 func (a Args) VSync(m VSyncMethod) Args {
 	return append(a, m.Args()...)
--- a/pkg/ffmpeg/transcoder/image.go
+++ b/pkg/ffmpeg/transcoder/image.go
@@ -2,6 +2,7 @@ package transcoder
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/stashapp/stash/pkg/ffmpeg"
 )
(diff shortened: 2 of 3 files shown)


feederbox826:
Can be done natively in ffmpeg without intermediate png https://ffmpeg.org/ffmpeg.html#filter_005fcomplex_005foption

dev-null-life:
Good call — updated to use `filter_complex` natively:

```
color=white[bg];[0:v]format=rgba[fg];[bg][fg]scale2ref[bg2][fg2];[bg2][fg2]overlay=format=auto,scale=N:N:force_original_aspect_ratio=decrease[out]
```

`scale2ref` sizes the white source to match the input dimensions, so no fixed size is needed. Removed the Go-side image decode/composite/re-encode entirely.

PR #6699, opened 2026-03-18
Split performer measurements into individual fields

dev-null-life:
## Summary

Splits the single `measurements` string field on `Performer` into four typed columns, following the same pattern as the `career_length` → `career_start`/`career_end` split in #6449.

| Old | New |
|---|---|
| `measurements: String` | `band_size: Int` |
| | `cup_size: String` |
| | `waist_size: Int` |
| | `hip_size: Int` |

The `measurements` field is kept on all input/filter types as `@deprecated` and parsed as a fallback when the new fields are not provided, so existing integrations continue to work.

### Changes

- **Migration 86** — adds the four new columns, back-fills from existing `measurements` strings (format `34DD-24-34`), then drops the old column
- **Backend** — `Performer`, `PerformerPartial`, `PerformerCreateInput`, `PerformerUpdateInput`, `BulkPerformerUpdateInput`, `PerformerFilterType`, `ScrapedPerformer`, JSON import/export, stashbox integration, scraper
- **Filtering** — `band_size`/`waist_size`/`hip_size` use `IntCriterionInput`; `cup_size` uses `StringCriterionInput`
- **Sorting** — `cup_size` sorts by `LENGTH` first then alphabetically, so single-letter cups (A–G) precede double-letter ones (DD, EE), giving a natural size progression
- **Frontend** — edit panel, details panel, list table, scrape/merge dialogs, filter criteria, locale strings

## Test plan

- [ ] Existing performers with a `measurements` value have it correctly split after migration
- [ ] Performers with no `measurements` value migrate cleanly (all four new columns NULL)
- [ ] Creating/editing a performer with the new individual fields works
- [ ] Creating/editing via the deprecated `measurements` field still parses correctly
- [ ] Filter by `band_size`, `cup_size`, `waist_size`, `hip_size` returns expected results
- [ ] Sort by each new field works; `cup_size` sort follows natural size order
- [ ] `is_missing` works for each of the four new fields
- [ ] Scraper results populate the new fields correctly
- [ ] Performer merge dialog shows and applies measurement fields
- [ ] JSON export includes the four new fields; import back-fills from legacy `measurements` string


feederbox826:
is band size the correct term? it's technically the underbust. and "size" seems to be a very weird choice when we have penis_length. Would prefer some or literally any thought 

https://github.com/stashapp/stash/pull/6699/changes#diff-20371c3d894db65775b4a85cfbd0e6f015e747f0f708e31ecabc65ff78ba0752R889-R893

this is incorrect in so many ways

Gykes:
So, I appreciate the work you are trying to do. Lets go ahead and slow down though. You like to drop 3-4 Claude Code PR's within minutes of each other. We like to take things a little slower around here. Lets limit it to 1 PR and please either hop in Discord, Discourse, or start a discussion here on GH. Once you get the thumbs up from someone in the admin team then we can go ahead and start discussions on features. Spamming a bunch of low effort PRs is just more stuff for us to sludge through.  

Feel free to ping me at any point on any of those platforms... i'm normally around

PR #6700, opened 2026-03-18
Add option to ignore zip contents during clean

### LINKED ISSUE #5112: [Feature] Don't access all files in a gallery zip during a clean
When doing a clean, Stash assesses all the files inside a .zip gallery. This is troublesome for people who have many zip galleries, especially if they're on slow-access media such as cloud drives. (I have over 10k gallery zips on a cloud drive, it takes literally days to do a "clean" task)

Stash should see that the actual .zip file hasn't changed in size, modtime, or anything, and thus should continue directly, instead of opening every single file inside the .zip.

Only problem I see would be somewhat relevant to https://github.com/stashapp/stash/pull/4113 -- but I don't see it being Stash's problem if people edit zip files without editing modtime. And it probably wouldn't even be a problem regarding cleaning.

WithoutPants:
Resolves #5112 and [related request](https://discourse.stashapp.cc/t/option-to-not-check-contents-of-zip-gallery-files-when-cleaning/3677)

Adds `ignoreZipFileContents` option to `CleanMetadataInput`, defaulting to false. When true, omits files/folders within zip files from being assessed. This will be a significant speed up for users that don't remove files within zip files.

Adds flag to clean task in the UI:
image

### DIFF
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -131,6 +131,14 @@ type ScanMetadataOptions {
 input CleanMetadataInput {
   paths: [String!]
 
+  """
+  Don't check zip file contents when determining whether to clean a file.
+  This can significantly speed up the clean process, but will potentially miss removed files within zip files.
+  Where users do not modify zip files contents directly, this should be safe to use.
+  Defaults to false.
+  """
+  ignoreZipFileContents: Boolean
+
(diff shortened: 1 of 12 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/option-to-not-check-contents-of-zip-gallery-files-when-cleaning/3677/2

PR #6702, opened 2026-03-18
feat: align saved filter API shape, add saved_filter_id support and dynamic label mapping

notsafeforgit:
This PR refactors how saved filters are stored and how they're handled in the API to align them with standard input shapes. It also introduces the ability to query results directly using a `saved_filter_id`, which simplifies UI state management and improves performance.

Importantly, this change migrates labels from static storage within the filter JSON to a **dynamic resolver**, ensuring that labels (e.g., Tag names, Performer names) are always up-to-date even if the underlying objects are renamed.

### 1. API Shape Alignment & Migration
* **Serialized Shape:** The `object_filter` in the `saved_filters` table has been refactored to match the GraphQL input types (e.g., `SceneFilterType`). Previously, it stored a more complex structure that included hardcoded labels, which led to stale data.
* **Migration (v86):** Added a migration and a Go post-migration script to automatically convert existing saved filters. This process strips the static labels and converts the criteria into the strictly-typed format used by the API.

### 2. `saved_filter_id` Query Parameter
* Added a `saved_filter_id` parameter to all `find*` queries (Scenes, Images, Performers, Studios, etc.).
* **Functional Logic:** When provided, the API resolves the saved filter by ID, enforces that the mode matches the query type, and merges it with any provided `filter` (FindFilter) overrides like search string, paging, or sorting.

### 3. Dynamic Label Mapping Support
* **Live Resolvers:** Added a `label_mapping` field to the `SavedFilter` type that is resolved dynamically at request time.
* **Data Consistency:** Because labels are no longer stored in the database's JSON blob, the UI will always display the current name/title of a referenced object. If a Tag or Performer is renamed, all saved filters referencing that object will reflect the change immediately.
* **Reduced Redundancy:** The database now only stores the IDs of filtered objects, reducing the size of the `object_filter` blob and following better data normalization practices.

### 4. UI Integration
* Updated UI GraphQL fragments and models to support the new `label_mapping` and the simplified `object_filter` structure.
* Fixed TypeScript type mismatches resulting from the schema changes.

## Testing
* **Migration Testing:** Added `pkg/sqlite/migrations/86_postmigrate_test.go` to verify the conversion of various saved filter configurations.
* **Resolver Testing:** Added `internal/api/resolver_saved_filter_helper_test.go` to verify:
    * Successful resolution and mode-enforcement of saved filters.
    * Correct merging of `FindFilter` overrides.
    * Successful extraction and resolution of `label_mapping` for tags, performers, and other entities.
* **Verification:** Ran `make validate` (UI lint/check and backend lint/tests) and `make it` (integration tests).

### DIFF
--- a/ui/v2.5/src/models/list-filter/criteria/criterion.ts
+++ b/ui/v2.5/src/models/list-filter/criteria/criterion.ts
@@ -213,8 +213,23 @@ export abstract class ModifierCriterion<
 
   public setFromSavedCriterion(criterion: unknown) {
     const c = criterion as ISavedCriterion<V>;
+    const labelMapping =
+      ((criterion as Record<string, unknown>)._labelMapping as
+        | Record<string, string>
+        | undefined) || {};
     if (c.value !== undefined && c.value !== null) {
-      this.value = c.value;
+      if (
+        Array.isArray(c.value) &&
+        c.value.length > 0 &&
+        typeof c.value[0] === "string"
+      ) {
+        this.value = (c.value as unknown as string[]).map((id: string) => ({
+          id,
+          label: labelMapping[id] || id,
+        })) as unknown as V;
+      } else {
+        this.value = c.value;
+      }
     }
     this.modifier = c.modifier;
   }
@@ -417,14 +432,34 @@ export class IHierarchicalLabeledIdCriterion extends ModifierCriterion<IHierarch
   public setFromSavedCriterion(
     criterion: ISavedCriterion<IHierarchicalLabelValue>
   ) {
-    const { modifier, value } = criterion;
+    const c = criterion as unknown as Record<string, unknown>;
+    const modifier = c.modifier as CriterionModifier;
+    const value = c.value as IHierarchicalLabelValue | string[] | undefined;
+    const excludes = c.excludes as string[] | undefined;
+    const depth = c.depth as number | undefined;
+    const _labelMapping =
+      (c._labelMapping as Record<string, string> | undefined) || {};
 
     if (value !== undefined) {
-      this.value = {
-        items: value.items || [],
-        excluded: value.excluded || [],
-        depth: value.depth || 0,
-      };
+      if (Array.isArray(value)) {
+        this.value = {
+          items: value.map((id: string) => ({
+            id,
+            label: _labelMapping[id] || id,
+          })),
+          excluded: (excludes || []).map((id: string) => ({
+            id,
+            label: _labelMapping[id] || id,
+          })),
+          depth: depth || 0,
+        };
+      } else {
+        this.value = {
+          items: value.items || [],
+          excluded: value.excluded || [],
+          depth: value.depth || 0,
+        };
+      }
     }
 
     const modifierOptions =
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -12,6 +12,8 @@ type Query {
   "Queries for Files"
   findFiles(
     file_filter: FileFilterType
+    "Provide the ID of a saved filter instead of providing a file_filter. Cannot be used with file_filter "
+    saved_filter_id: ID
     filter: FindFilterType
     ids: [ID!]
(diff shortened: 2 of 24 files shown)


Gykes:
This is your 3rd PR. Lets go ahead and limit the amount that we have open till 1-2. Also, for larger refactors like this please reach out to someone on the dev team before blindly PRing. We like to take things slower around here and stacking all of these on top of each other makes our life harder. 

Please reach out on GH, Discord, or Discourse. You can ping me whenever, i'm around

ui/v2.5/src/models/list-filter/criteria/criterion.ts
```diff
@@ -213,8 +213,23 @@ export abstract class ModifierCriterion<
...
+        | Record<string, string>
+        | undefined) || {};
     if (c.value !== undefined && c.value !== null) {
-      this.value = c.value;
+      if (
+        Array.isArray(c.value) &&
+        c.value.length > 0 &&
+        typeof c.value[0] === "string"
+      ) {
+        this.value = (c.value as unknown as string[]).map((id: string) => ({
```
notsafeforgit, line 226: You'll notice some "ugly" double-casting. This was a necessary trade-off to support the new normalized ID-only storage format while maintaining the UI's rich model.

The `ModifierCriterion<V>` class is a generic base used by almost every filter in the UI. For many filters, `V` is a complex object (like `ILabeledId[]`), but the raw data coming from the saved filter's `object_filter` is now just a simple `string[]` (IDs only).

Because `V` is a generic type parameter, TypeScript won't allow a direct cast to `string[]`. We have to cast through `unknown` to "reset" the type checker so we can map the IDs back into the rich objects (`{id, label}`) the UI expects. The alternative would be a massive refactor of the entire filter type hierarchy to support union types (`V | string[]`) at every level. This would have touched dozens of files and significantly increased the risk of regressions.

The current implementation effectively "hydrates" the ID-only storage format back into the rich UI model using the `label_mapping` provided by the API, keeping the complexity isolated to this single base class.

notsafeforgit:
@Gykes  sorry about the 3 rapid fire PRs, this is all I have for now. I have just been bothered by these issues in stash for years now and finally got the time to really go heads down for the past week or so and iterate. Despite opening the PRs in quick succession, I spent a lot of time thinking about how to actually go about implementing these changes in a clean way, so they should hopefully not be a nightmare to review.

notsafeforgit:
Projects like https://github.com/1letzgo/stashy and https://github.com/secondfolder/stash-tv have long had to hack around the oddity of saved filters returning a different shape to clients vs what the find* apis actually accept, and most apps that use stash's api like these just have broken filters all the time. Hopefully this PR will make clients like them effortless to maintain when it comes to fetching the objects they want out of stash!

Gykes:
No problem, I don't want to make it seem like we don't want you making fixes and features, as we appreciate it. Currently we are getting near the end of the release and neither myself or the main dev has put any thought into how to handle this situation. Having a 30 minute convo with the admin team can go a long way of saving both of our times if we prefer a different implementation or just want to defer it completly.

WithoutPants:
I'm going to be blunt - this isn't a great submission. The description looks very LLM-generated and is not a good explanation. The code is undocumented in parts, and to me demonstrates a lack of understanding of the underlying system. In short, it looks like it's been vibe-coded.

The post-migration is - from what I can work out - converting the existing UI-coded to be in line with the graphql inputs, which makes sense, but there's nothing that I can see that is changing the way that we actually save the filters, so any new or modified saved filters will still have the old structure. 

I don't really like the mapping resolver solution done here. I don't have an immediate alternative solution, but I think that speaks to the idea that we need to actually properly design a solution to the original issue. 

The idea to accept `saved_filter_id` in the find queries is a good one, and conversely we _do_ need to massage the saved filter data to handle this, but I don't think this is a good solution to the issue and I don't think it's salvageable.

The _first_ thing I'd do to tackle this is to at least open an issue to discuss a proper approach to solve the underlying issue.

PR #6709, opened 2026-03-18
Feature: Add StashID guid consideration into performer search box

### LINKED ISSUE #5676: [Feature] Search performers by GUID in Scene selection in addition to name
For those with substantial performer collections and those who add from non-stash sources (groups, iafd, etc) it's common that sources like StashFace can identify a performer who has a single name that may have dozens if not hundreds of collissions.  StashFace will give you a GUID name, but commonly one must go find that performer in the performers section, find the identifying disambig and then select in the scene drop-down.  

It would be helpful to allow searching the GUID field in addition to names and aliases - perhaps only if a GUID is identified - but that may be fancy  Ideally when entering a guid it would suggest any performers that have that GUID (which really should be 0 or 1 total entries).

As listed above, it's a multi-step work around to look up the performer directly and then add them by disambiguation is time-consuming and inefficient.

**Additional context**

Image

Stash-KennyG:
Resolves #5676 

## Summary
Extend the scene (and shared) performer picker so it can search performers not only by name/alias but also by `stash_id` GUID.
## Details
- In `PerformerSelect.tsx`, the `loadPerformers` function now:
  - Detects GUID-like input using a simple 8‑4‑4‑4‑12 hex pattern.
  - When the input matches, performs a `findPerformers` query with a `stash_id_endpoint` criterion (endpoint blank, `stash_id` = input).
  - If any performers are found, returns those immediately.
  - If none are found, falls back to the existing name/alias search behavior.
- Normal name/alias search behavior is unchanged for non‑GUID input.
## Rationale
For users with large performer libraries and many name collisions, face matchers and external tools often provide a performer GUID. This change allows pasting that GUID directly into the performer picker on the scene page and selecting the correct performer without having to search the Performers list first.
## Testing
- `make validate-ui` (lint, typecheck, prettier) passes.
- Verified in the UI:
  - Typing a normal name still performs the existing name/alias search.
  - Pasting an existing performer `stash_id` GUID selects that performer via the picker.
  
Example: 
image

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerSelect.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerSelect.tsx
@@ -23,6 +23,7 @@ import {
   IFilterProps,
   IFilterValueProps,
   Option as SelectOption,
+  toOption,
 } from "../Shared/FilterSelect";
 import { useCompare } from "src/hooks/state";
 import { Link } from "react-router-dom";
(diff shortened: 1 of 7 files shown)


Stash-KennyG:
I was able to build in my branch, tested manually against a database of 22k performers - works great.
Resolves my own open feature request form 13 months ago!

PR #6716, opened 2026-03-19
Show studio name if studio image not set on detail pages

### LINKED ISSUE #5073: [Feature] Use text instead of studio image
It would be great if stash would display srudio names as text instead of empty placeholder image when no studio image is available or set. Not only does the placeholder take up a lot of space without providing any details, but it's unhelpful as one cannot see what studio is set unless editing the scene.

Either pure text like 'Studio: blastudio' or a dynamic placeholder image with the name of the studio would help greatly. Not only in the scene view, but also in the studio view are the empty placeholders useless. In the studio overview they also don't help when a lot of studios don't have logos (like for 3D animation creators). 

Screenshot_2024-07-14-01-23-32-69_40deb401b9ffe8e1df2f1cc5ba480b12

WithoutPants:
If no studio image is set, shows the studio icon with the studio name.

image

image

Happy to refine this as needed based on feedback. This is a low-risk PR so we could squeeze it into 0.31.

Resolves #5073

Added new interface option to always show studio text instead of logo:

image

Also clarified existing show studio as text option:

image

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryDetails/Gallery.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryDetails/Gallery.tsx
@@ -62,6 +62,7 @@ export const GalleryPage: React.FC<IProps> = ({ gallery, add }) => {
   const Toast = useToast();
   const intl = useIntl();
   const { configuration } = useConfigurationContext();
+  const { showStudioText } = configuration?.ui ?? {};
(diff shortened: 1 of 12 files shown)


DogmaDragon:
Could we add settings toggle to make it a permanent option, similar to how it works on scene card?

PR #6726, opened 2026-03-21
fix version string not being displayed

feederbox826:
tl;dr regression from https://github.com/stashapp/stash/pull/6638

fetching shallow history of git was erroring out git tag --describe, it needs to be added both in generated for UI (vite) to properly display it and in compile for graphql to display it


Gykes (approved):
Should get merged in sooner rather than later

Issue #2914, opened 2022-09-15
[Feature] Specify delay between 'identify' or 'scrape all' requests


acccelerated:
**Is your feature request related to a problem? Please describe.**
Some sites can be pretty aggressive about throttling/blocking users who are performing various quantities of automated requests. I have to babysit the process and issue them in small batches which is not a 'problem' but is suboptimal. I am also facing unnecessary risk of getting flagged as a problematic user or blocked as a result of not being able to control the rate at which these requests are sent. Additionally for 'friendly' servers being able to set this to a more generous interval would let me reduce the intensity of the 

**Describe the solution you'd like**
I would be able to specify a time period in ms between individual requests in batches when using the identify or scrape metadata feature in batches.

**Describe alternatives you've considered**
I'm not sure what form an alternative to my suggestion could take.

DogmaDragon:
That can be done at the scraper level.

feederbox826:
This could theoretically be done externally under "Identify". Have a plugin iterate through every identify call and on error, exponentially throttle for that endpoint. Pull identify settings from the instance and would also theoretically allow for multi-source identification

GalacticCheat:
Maista sucks cock
![Image](https://github.com/user-attachments/assets/ed12369e-cb07-434a-b49f-fbc470272035)
https://i.ibb.co/ds49xcXt/image-2-copie.jpg

Gykes:
Do we still want this in core? I know we recently added rate limiting to Stash-Box but if this can be done at the scraper level I would think it would be better to put it there.

feederbox826:
Not all of it can be done at the scraper level, but it would have to be per-scraper ratelimit, not a global one

PR #6731, opened 2026-03-21
Add exponential backoff for HTTP 429 rate limiting in scrapers

priv-r8s:
## Summary
- Add retry logic with exponential backoff when scrapers receive HTTP 429 (Too Many Requests)
- Backoff delay = Retry-After + exponential (2s, 4s, 8s, ...) so the server's requested wait is always respected
- If Retry-After exceeds 60s max, give up immediately instead of waiting excessively
- Parse Retry-After header (both seconds and HTTP-date formats) per RFC 9110
- 5-minute total timeout prevents rate-limit retries from running indefinitely
- Comprehensive unit tests for all backoff paths

## Test plan
- [x] Unit tests: exponential backoff without Retry-After header
- [x] Unit tests: Retry-After + exponential backoff (additive)
- [x] Unit tests: Retry-After via HTTP-date format
- [x] Unit tests: Retry-After exceeds max → immediate give-up
- [x] Unit tests: loadURL retry-and-succeed after 429s
- [x] Unit tests: loadURL retry exhaustion
- [x] Unit tests: context cancellation during retries
- [x] Live tested against r18.dev API — confirmed backoff works with real 429 responses (Retry-After: 10)

Replaces #6722 (closed due to fork rebuild).


Gykes (commented):
So, after a brief look over the biggest thing that's nagging me is the hard coded values. Each site is different and has more or less strict rules based on scraping. Hard coding these could increase wait times dramatically depending on the situation. I see 2 possible options and could use some feedback on best solutions. 

1) Change this up to allow individual scrapers to pass the values: 
 - `rate_limit_retries`. Allow scrapers to pass this individually per scraper. It would require all existing scrapers to be updated and pass these new values but would allow user flexibility depending on site. 

2) Set them as variables and add these settings to the scrapers settings in the app to allow users to modify them to what they see fit. We would need to find sane default for these but it could allow users to do their own testing with sites and find optimal settings.

Would be best to wait for some feedback from others before implementing any of these.

feederbox826:
Off the top of my head, not sure how helpful this would be. I haven't encountered one yml/json scraper that returns a 429 properly as a ratelimit backoff mechanism.

- redgifs (needs anon auth token so need to be py)
- nubiles (fake 429)
- TPDB (real 429, handled through stash-box concurrency limit)

Other than that I'm not sure how this would apply for bulk scrapes, would we need a global cooldown that freezes all http requests or is this better suited just as #2914

feederbox826:
closing in favor of #2914, global ratelimit would be better rather than an unlimited exponential ratelimit within a single scrape

PR #6733, opened 2026-03-22
Adds show favorites first for performers, studios, tags

### LINKED ISSUE #5785: [Feature] Display favourite tags firs
> For example, if I had tags: A, B, C, D, E, and F. Normally the tag filter would order them as such, however, if I had favourited the C and E tags, the order would now be C, E, A, B, D, F. Alternatively, the favourited tags could appear in the list twice; at the top and then again in their normal position, so using the previous example: C, E, A, B, C, D, E, F.

floyd352:
This feature adds the ability to favorite performers, studios, tags, which makes them always appear first.

CleanShot 2026-03-22 at 10 42 55

The [request](https://github.com/stashapp/stash/issues/5785#issue-2957985013) from @feederbox826 specifies tags but I went ahead and allowed for anything that can be favorited.  I figure we might as well allow it all.

These changes also apply to all the tabs pages like: Tags -> Performers tab, Tags -> Studio tab, Studios -> Performers tab.

Closes https://github.com/stashapp/stash/issues/5785

> [!NOTE]
> A number of the changes in here, like the Go code, are AI generated.  I've reviewed the code changes personally to the best of my ability.  I have manually tested it.

## Tags

CleanShot 2026-03-22 at 10 44 29

## Performers (favorited, number of gallery ASC)

CleanShot 2026-03-22 at 10 44 58

## Performers (favorited, random DESC)

CleanShot 2026-03-22 at 11 11 19

## Studios -> Performers tab (favorited, name DESC)

CleanShot 2026-03-22 at 11 11 59

### DIFF
--- a/pkg/sqlite/filter_internal_test.go
+++ b/pkg/sqlite/filter_internal_test.go
@@ -640,3 +640,134 @@ func TestStringCriterionHandlerNotNull(t *testing.T) {
 	assert.Equal(fmt.Sprintf("(%[1]s IS NOT NULL AND TRIM(%[1]s) != '')", column), f.whereClauses[0].sql)
 	assert.Len(f.whereClauses[0].args, 0)
 }
+
+func TestValidateSortFavoritesFirstPrefix(t *testing.T) {
+	opts := sortOptions{"name", "rating", "random", "scenes_count"}
+	assert := assert.New(t)
+
+	// valid: known sorts with the favorites_first_ prefix pass
+	assert.NoError(opts.validateSort("favorites_first_name"))
+	assert.NoError(opts.validateSort("favorites_first_rating"))
+	assert.NoError(opts.validateSort("favorites_first_scenes_count"))
+
+	// valid: random with prefix passes (random itself is valid)
+	assert.NoError(opts.validateSort("favorites_first_random"))
+	assert.NoError(opts.validateSort("favorites_first_random_42"))
+
+	// invalid: prefixed random with invalid seed
+	assert.Error(opts.validateSort("favorites_first_random_not-a-number"))
+
+	// invalid: unknown sort after prefix
+	assert.Error(opts.validateSort("favorites_first_unknown_sort"))
+
+	// invalid: nothing after prefix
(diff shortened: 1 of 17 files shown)


feederbox826:
Deferring for now - Actually a request I forwarded from the discord.

Bones look to be okay but theres a few quirks with ratings? that warrant a closer look

Gykes:
I haven't done research or looked at the code just my initial thoughts. 

Would it be better to just automatically have favorited content just go to the top? I'm not sure if I'm misremembering but I feel like quite a few apps just automatically do it. The settings page is already a cluster and putting more in there would just make it even worse.

Just looking for opinions

floyd352:
> theres a few quirks with ratings? that warrant a closer look

@feederbox826  Thanks for checking it out, can you share a bit more information to make it easier to reproduce

> Would it be better to just automatically have favorited content just go to the top?

@Gykes there's a [comment in the issue](https://github.com/stashapp/stash/issues/5785#issuecomment-3667468881) of "Should be optional", I originally agreed with it.
~It's a good question, we have ratings and we have favorites on the same items.  You can technically give 1 star to a favorited item which is a bit odd.  One difference is that you can't sort by favorites, as far as I can tell.  So even if you mark an item as favorite, if you can't sort, there's no easy way to find it in the library?
I like the idea of making this feature optional, but if you can't sort by favorite, I'm kind of lost what is the difference/point of favorite is compared to rating.  At the time of writing this, I think either we should be able to sort by favorite, or maybe it favorite does behave differently and bring to the top.~

EDIT: I'll wait for responses before suggesting something that might result in scope creep.

WithoutPants:
The original issue was referring to putting favourited items first in the _selector_ list - ie when selecting tags for a scene, (optionally) the favourite tags would be placed first, then the rest. I don't really like how this has been implemented. We should consider accepting multiple sort options which would better allow for doing something like this.

Gykes:
Closing as stale for now. OP is welcomed to open it back up for more clarification on lead devs thought process.

PR #6734, opened 2026-03-22
[ui] Fix Tag Modal cutting off

### LINKED ISSUE #6723: Tags tagger cuts off descriptions
### Describe the bug

Tags tagger is truncating descriptions to a single line which causes the cut off.

### Steps to reproduce

1. Go to Tags > Tagger.
2. Scrape tag that has multi-line description.

### Expected behaviour

Not truncating descriptionsto a single line at all or at the very least fixing the visual bug.

Image
Image

### Stash version

640d62c

### Device details

_No response_

### Relevant log output

```shell

```

feederbox826:
closes #6723

The actual issue is from the margins setting of `.no-gutters` interacting with the webkit display of `TruncatedText`. TruncatedText seems to have priority and cuts off the line, then the margin kicks in and wraps the overflow characters to the next line. This is only used in one other place, the description of studios when scraped, which doesn't exist on stash-box, so this hasn't come up

Instead, opted for parity with TagCardDetails to set lineCount to 3 https://github.com/stashapp/stash/blob/c832e1a8a29250cd2720bac3b3b042b0b62aee49/ui/v2.5/src/components/Tags/TagCard.tsx#L130-L140

UI validation on my machine
image

### DIFF
--- a/ui/v2.5/src/components/Tagger/tags/TagModal.tsx
+++ b/ui/v2.5/src/components/Tagger/tags/TagModal.tsx
@@ -103,7 +103,7 @@ const TagModal: React.FC<ITagModalProps> = ({
             <FormattedMessage id={id} />:
           </strong>
         </div>
-        <TruncatedText className="col-7" text={text} />
+        <TruncatedText className="col-7" text={text} lineCount={3} />
       </div>
     );
   }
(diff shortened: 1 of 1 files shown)


Gykes (approved):
Looks good.

Tested on some tags that had longer descriptions. Was unable to find one that had 4+ lines to test what it looks like but < 3 looks much better.

PR #6741, opened 2026-03-24
Add mutex to repositoryCache for thread safety

### LINKED ISSUE #6730: "fatal error: concurrent map writes" crash in repositoryCache.cacheList, no mutex on shared cache map
### Describe the bug

Hi! I wasn't able to find any AI policy for Stash, so please let me know if issues that have been written with help from AI are not desired. I dont write any go myself, so this was the only way for me to try to understand what is happening. The unfortunate bad thing about this is that I will suck at answering any follow-up questions 😅

The bug:
Stash crashes with `fatal error: concurrent map writes` at `pkg/pkg/cache.go:55` when concurrent `InstalledPackages` queries race to write to an unprotected map. The process just dies.
I ran a scan with 2+ tabs with Stash open in them. The `scanComplete` event fires on all 5 WebSocket connections at once, each tab re-queries `installedPackages`, and the goroutines collide. The crash dump below has 4 goroutines simultaneously inside `httpRepository.List` / `getCachedList` at time of crash.

### Steps to reproduce

Open Stash in 2+ browser tabs, run a metadata scan, and wait for it to complete and crash. 

### Expected behaviour

No crash.

**Crash output**
The full output:  [stash_go_crash.txt](https://github.com/user-attachments/files/26159782/stash_go_crash.txt)

Goroutines 517 and 290 are both running `ListInstalledRemotes` against the same `Manager` pointer (`0xc00020ddd0`), so they share the same `*repositoryCache`. Goroutines 251 and 248 are racing on a second `Manager` instance independently (`0xc0003dae10`), probably the other package type.

**Possible cause**
`repositoryCache` in `pkg/pkg/cache.go` has no mutex:

```
type repositoryCache struct {
    cache map[string]cacheEntry  // no synchronization
}

func (c *repositoryCache) cacheList(url string, lastModified time.Time, data []RemotePackage) {
    c.ensureCache()
    c.cache[url] = cacheEntry{...}  // line 55 — concurrent write → crash
}
```
`ensureCache()` has the same problem: two goroutines can both see `c.cache == nil` and both call `make()`.

The race path: `InstalledPackages` → `getInstalledPackagesWithUpgrades` → `ListInstalledRemotes` → `ListRemote` → `httpRepository.List`. Each call to `ListRemote` creates a fresh `httpRepository` via `remoteFromURL`, but all calls within the same Manager share the same `*repositoryCache` via `Manager.getCache()`. When N concurrent queries all find the cache stale, all N fetch the remote list independently and all call `cacheList` on return → simultaneous map writes, crash.

**Proposed fix**
Add `sync.RWMutex` to `repositoryCache` and handle the nil-map initialisation inside each method (removing the need for `ensureCache` entirely):

```
// pkg/pkg/cache.go
import "sync"

type repositoryCache struct {
    mu    sync.RWMutex
    cache map[string]cacheEntry
}

func (c *repositoryCache) cacheList(url string, lastModified time.Time, data []RemotePackage) {
    if c == nil {
        return
    }
    c.mu.Lock()
    defer c.mu.Unlock()
    if c.cache == nil {
        c.cache = make(map[string]cacheEntry)
    }
    c.cache[url] = cacheEntry{lastModified: lastModified, data: data}
}

func (c *repositoryCache) lastModified(url string) *time.Time {
    if c == nil {
        return nil
    }
    c.mu.RLock()
    defer c.mu.RUnlock()
    if c.cache == nil {
        return nil
    }
    e, found := c.cache[url]
    if !found {
        return nil
    }
    return &e.lastModified
}

func (c *repositoryCache) getPackageList(url string) []RemotePackage {
    c.mu.RLock()
    defer c.mu.RUnlock()
    if c.cache == nil {
        return nil
    }
    e, found := c.cache[url]
    if !found {
        return nil
    }
    return e.data
}
```
As a follow-on, `singleflight` at the `Manager.ListRemote` level would prevent the redundant HTTP fetches when many goroutines find the cache cold simultaneously, but the mutex above is the critical fix. (Note: `singleflight` on `httpRepository` itself wouldn't help since `remoteFromURL` creates a fresh instance per call.)

### Stash version

v0.30.1

### Device details

Vivaldi 7.9.3970.31 on Windows 11.

### Relevant log output

```shell

```

SandiyosDev:
added `sync.RWMutex` to `repositoryCache`... removed `ensureCache()` and handle nil-map init under the lock instead

should close #6730, see if @dustinruusu can confirm this fix

### DIFF
--- a/pkg/pkg/cache.go
+++ b/pkg/pkg/cache.go
@@ -1,6 +1,7 @@
 package pkg
 
 import (
+	"sync"
 	"time"
 )
 
@@ -10,14 +11,9 @@ type cacheEntry struct {
 }
 
 type repositoryCache struct {
+	mu sync.RWMutex
 	// cache maps the URL to the last modified time and the data
 	cache map[string]cacheEntry
-}
-
-func (c *repositoryCache) ensureCache() {
-	if c.cache == nil {
-		c.cache = make(map[string]cacheEntry)
-	}
 }
 
 func (c *repositoryCache) lastModified(url string) *time.Time {
@@ -25,7 +21,13 @@ func (c *repositoryCache) lastModified(url string) *time.Time {
 		return nil
 	}
 
-	c.ensureCache()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if c.cache == nil {
+		return nil
+	}
+
 	e, found := c.cache[url]
 
 	if !found {
@@ -36,7 +38,13 @@ func (c *repositoryCache) lastModified(url string) *time.Time {
 }
 
 func (c *repositoryCache) getPackageList(url string) []RemotePackage {
-	c.ensureCache()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if c.cache == nil {
+		return nil
+	}
+
 	e, found := c.cache[url]
 
 	if !found {
@@ -51,7 +59,13 @@ func (c *repositoryCache) cacheList(url string, lastModified time.Time, data []R
 		return
 	}
 
-	c.ensureCache()
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.cache == nil {
+		c.cache = make(map[string]cacheEntry)
+	}
+
 	c.cache[url] = cacheEntry{
 		lastModified: lastModified,
 		data:         data,
--- a/pkg/pkg/manager.go
+++ b/pkg/pkg/manager.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"path/filepath"
+	"sync"
 
 	"github.com/stashapp/stash/pkg/logger"
 	"github.com/stashapp/stash/pkg/models"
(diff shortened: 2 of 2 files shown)


Gykes (commented):
The method is called from multiple `goroutines` (via remoteFromURL), so two `goroutines` can both see `m.cache == nil`, each allocate a new `repositoryCache`, and end up with separate instances. At that point, the mutex inside each instance is protecting nothing shared and the goroutines are writing to different maps entirely.                                 
                             
Would a better fix be to initialize the cache when Manager is constructed so that `getCache()` is never racing on a nil check?

SandiyosDev:
updated, used sync.Once for getCache so it only inits once, didn't wanna touch Manager constructor in internal/manager
much appreciated

dustinruusu:
> added `sync.RWMutex` to `repositoryCache`... removed `ensureCache()` and handle nil-map init under the lock instead
> 
> should close #6730, see if @dustinruusu can confirm this fix

I don't have an easy way to do this right now, since I don't have Stash built from source and because I created a workaround so I currently don't run the code that caused this. 

Do the tests of Stash cover this? Or are you comfortable merging based on just code review?
If not, I need a bit of time to set up everything for properly testing this, so it'll be a few days.

WithoutPants (approved):

PR #6742, opened 2026-03-24
Fix 'not equals' custom field to include unset objects

### LINKED ISSUE #6611: Custom field filters do not inherently include "is not null"

A bit complicated, but all the custom field filters inherently include "is not null"
when searching "field" "is not", it will only return results that have "field" set.

It would be helpful for it to have a default state match of false, so any objects without the field set would also match "is not", since they have no value for that field

Coming from tags, custom fields are a much better way to store arbitrary data and labels that  would fit badly into the tags system, adopting a similar "is not" system would be able to replace arbitrary filters, fake URLs and other workarounds

revert to tags :(

### Additional context

would be possible with #3431 boolean "field" is not "values" OR "field" is null

SandiyosDev:
should close #6611

objects without custom field set were being excluded from "is not" results, switched NotEquals to use LEFT JOIN + OR IS NULL, its the same pattern excludes already uses

SQLite tests pass


Gykes (commented):
Pulled down the fix and tested. Worked as expected. 

I believe that Excludes and NotBetween will have this same issue. Could you update those ones too please?

SandiyosDev:
done

PR #6743, opened 2026-03-24
Fix lightbox not reading scale-up form config

### LINKED ISSUE #6504: "Scale up to fit" toggle does not work, has to be manually enabled each time
### Describe the bug

The "Scale up to fit" toggle under Image Lightbox setting does not work. Even with it turned on, "Scale up to fit" has to be manually selected in the lightbox each time.

### Steps to reproduce

1. Set "Fit to screen" and "scale up to fit."
2. Display an image smaller than the screen size 

### Expected behaviour

Setting "Scale up to fit" in Settings should make that the default for all new lightbox views.

_No response_

### Stash version

v0.30.1

### Device details

Mac 26.2 apple silicon

### Relevant log output

```shell

```

SandiyosDev:
close #6504
per @feederbox826

> Was only able to reproduce with the following steps
> IN settings -> interface -> image lightbox, set "scale up to fit" to true
> clear/ ensure localforage/ keyvaluepairs is clear (this will override the global settings)
> a. Clear cache/ localstorage works
> view settings box in individual lightbox
>This seemingly ONLY affects "scale to fit" I suspect because it's a binary value and it's evaluating undefined as falsey. > Regardless, localForage overrides stash config

figured out localForage was overriding the global config with undefined evaluating as falsey...
fixed for lightbox settings to now fall through localForage > config > default

Left contains current behavior, right contains this PR with desired behavior
2026-03-23-230418_2560x1440_scrot

thats all for tonight

### DIFF
--- a/ui/v2.5/src/hooks/Lightbox/Lightbox.tsx
+++ b/ui/v2.5/src/hooks/Lightbox/Lightbox.tsx
@@ -206,8 +206,25 @@ export const LightboxComponent: React.FC<IProps> = ({
     setLightboxSettings({ slideshowDelay: v });
   }
 
+  const scaleUp =
+    lightboxSettings?.scaleUp ??
+    config?.interface.imageLightbox.scaleUp ??
+    false;
+
+  const resetZoomOnNav =
+    lightboxSettings?.resetZoomOnNav ??
+    config?.interface.imageLightbox.resetZoomOnNav ??
+    false;
+
+  const scrollMode =
+    lightboxSettings?.scrollMode ??
+    config?.interface.imageLightbox.scrollMode ??
+    GQL.ImageLightboxScrollMode.Zoom;
+
   const displayMode =
-    lightboxSettings?.displayMode ?? GQL.ImageLightboxDisplayMode.FitXy;
+    lightboxSettings?.displayMode ??
+    config?.interface.imageLightbox.displayMode ??
+    GQL.ImageLightboxDisplayMode.FitXy;
   const oldDisplayMode = useRef(displayMode);
 
   function setDisplayMode(v: GQL.ImageLightboxDisplayMode) {
@@ -250,13 +267,13 @@ export const LightboxComponent: React.FC<IProps> = ({
     // reset zoom status
     // setResetZoom((r) => !r);
     // setZoomed(false);
-    if (lightboxSettings?.resetZoomOnNav) {
+    if (resetZoomOnNav) {
       setZoom(1);
     }
     setResetPosition((r) => !r);
 
     oldIndex.current = index;
-  }, [index, images.length, lightboxSettings?.resetZoomOnNav]);
+  }, [index, images.length, resetZoomOnNav]);
 
   const getNavOffset = useCallback(() => {
     if (images.length < 2) return;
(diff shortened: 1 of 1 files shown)


SandiyosDev:
I fixed the same pattern for the other lightbox settings while I was in there, but if that scope is too broad beyond this pr, i can trim it back

ui/v2.5/src/hooks/Lightbox/Lightbox.tsx
```diff
@@ -206,8 +206,25 @@ export const LightboxComponent: React.FC<IProps> = ({
...
   }
 
+  const scaleUp =
+    lightboxSettings?.scaleUp ??
+    config?.interface.imageLightbox.scaleUp ??
+    false;
+
+  const resetZoomOnNav =
+    lightboxSettings?.resetZoomOnNav ??
+    config?.interface.imageLightbox.resetZoomOnNav ??
```
feederbox826, line 216: ? what is the undefined fallback for? it's not chained all the way so not sure where it should stop

feederbox826:
not sure if hardcoding it will be the best choice or if we should memoize it/ whatever react does? or maybe make the lookups a function so that it will be reactive once it's changed on backend instead of on load

SandiyosDev:
> not sure if hardcoding it will be the best choice or if we should memoize it/ whatever react does? or maybe make the lookups a function so that it will be reactive once it's changed on backend instead of on load

The graphql schema has this as nullable (scaleUp: Boolean not Boolean!) so config can return null/undefined for them
Coding in fallbacks covers that

the chain is localForage > config > safe default, same way slideshowDelay already works a few lines above

If you have a preferred pattern, you can collab with me on this pr

feederbox826:
> The graphql schema has this as nullable (scaleUp: Boolean not Boolean!) so config can return null/undefined for them
Coding in fallbacks covers that

no it doesn't - your nullable is at the level of config?.interface and you'll hit index OOB

SandiyosDev:
@feederbox826 the existing code a few lines above already use the same pattern, useConfigurationContext throws if config is null, and the grqphql query always return the full imageLightbix object. if that chain was unsafe, the lightbox would alreayd be crashing on slideshowDelay

feederbox826:
> The graphql schema has this as nullable (scaleUp: Boolean not Boolean!) so config can return null/undefined for them

but your null check is at the level of config, it doesn't step that far in.

SandiyosDev:
hummm... the nullable is at the leaf level (scaleUp: Boolean), not at interface or imageLightbox

SandiyosDev:
@feederbox826 can you show me a case where it would crash? the existing slideshowDelay ilne uses the same chain and it hasn't crashed

feederbox826:
> @feederbox826 can you show me a case where it would crash? the existing slideshowDelay ilne uses the same chain and it hasn't crashed

Not about a case that crashes, your comment is conflicting with your code. This PR exists because of a nullable OOB not being accessed properly.

SandiyosDev:
I see...i was overcomplicating it
the fallback just covers when config fields come back null. would you rather i trim this to just scaleUp?

Gykes (approved):
LGTM. Follows the same pattern as `slideshowDelay`

PR #6746, opened 2026-03-24
Don't return heatmap path for non-interactive scenes

### LINKED ISSUE #6477: Deleting a Funscript does not remove the Heatmap on the Scene page.
### Describe the bug

Delete a Funscript and click Scan.
The heatmap is removed from the Scene's preview, as expected.
However within the Scene page, the old heatmap is still visible under the seek bar, which is not expected.

Troubleshooting:
Within the scene, clicking "Generate heatmap" with overwrite selected doesn't remove the old heatmap.
Within settings, clicking the Clean / Clean Generated Files / Optimise Database buttons doesn't remove the old heatmap either.

### Steps to reproduce

Outside of Stash, add a funscript file against a scene, labelled appropriately so it's detected by Stash.
Click scan to get the funscript linked. 
Generate heatmap and see the heatmap against the scene's preview, and within the scene's page.

So far so good!

Outside of Stash, delete the Funscript.
In Stash, click Scan for the Funscript to be forgotten.

The heatmap that was visible against the scene's preview disappears.
The heatmap shown within a scene page above the scrubber is still visible.

### Expected behaviour

When a funscript is deleted, expect the heatmap to disappear from the Scene page on Scan please.
When Clean Generated Files is run, expect the heatmap image to be deleted please.

Thank you!

_No response_

### Stash version

v0.30.1

### Device details

_No response_

### Relevant log output

```shell

```

SandiyosDev:
fixes #6477
Paths resolver was always returning the heatmap URL even after funscript being removed, now is only included when primary file is interactive

### DIFF
--- a/ui/v2.5/src/components/ScenePlayer/ScenePlayerScrubber.tsx
+++ b/ui/v2.5/src/components/ScenePlayer/ScenePlayerScrubber.tsx
@@ -349,9 +349,10 @@ export const ScenePlayerScrubber: React.FC<IScenePlayerScrubberProps> = ({
         <div
           className="scrubber-heatmap"
           style={{
-            backgroundImage: scene.paths.interactive_heatmap
-              ? `url(${scene.paths.interactive_heatmap})`
-              : undefined,
+            backgroundImage:
+              scene.interactive_speed && scene.paths.interactive_heatmap
+                ? `url(${scene.paths.interactive_heatmap})`
+                : undefined,
           }}
         />
         <div ref={indicatorEl} id="scrubber-position-indicator" />


SandiyosDev:
tested too

Gykes (commented):
So, this fix is rather expensive. Every scene will do a DB call to check for an interactive file. I think a better solution would be to fix clean so that it cleans up these files then the original code will return a 404 and be handled gracefully. This recommendation is a larger fix but I think it's better to solve it at the source.

SandiyosDev:
honestly not sure about the clean only route, would it make sense to handle this in the http handler instead (return 404 if the scene isn't interactive) or is there a better pattern here

SandiyosDev:
the scrubber checks if heatmap URL non-null and sets background image regardless of whether the file is there. so even after clean removes the file, resolver still returns URL and the scrubber still tries to load it

Gykes:
Good point, after some looking, `SceneCard.tsx:495` already handles this by checking `interactive_speed` before using the heatmap URL. It's not guarded against this is `ScenePlayerScrubber.tsx:352`. Adding a similar check there would fix the issue without adding a DB call to every paths resolution.

Gykes (approved):
Thanks for making the changes.

PR #6748, opened 2026-03-24
fix metadata export for nonASCII names

### LINKED ISSUE #5032: [Bug Report] Metadata export fails for non-alphanumeric and duplicate object names
**Describe the bug**
Metadata export does not work correctly for any Movie/Group, Performer, Studio and Tag where the name has no alphanumeric characters. The other exports fall back to their filename/hash and are not effected. 

On Windows, a ".json" file is created with the contents of the last exported object with no alphanumeric characters in the name. Sometimes the JSON produced is mangled:
`{
  "name": "测试",
  "created_at": "2024-07-01T14:31:02+03:00",
  "updated_at": "2024-07-01T14:35:50+03:00"
}_at": "2024-07-01T14:36:25+03:00"
}`

On Linux, nothing is exported.

Additionally, if you have multiple of these objects with same name, e.g. multiple Movies with the name "Test", a single json "Test.json" is created with data from the last one exported.

**To Reproduce**
1. Create two or more Movies/Groups, Performers, Studios or Tags where the name has no alphanumeric characters, e.g. Chinese, Korean or Japanese like テスト or 시험
2. In Tasks, run a Full Export in the Metadata section
3. In the metadata folder configured, there is either no json file created for them or a single ".json" on Windows

**Expected behavior**
Metadata from Full Export is complete and does not miss anything. 

v0.26.2-36-g70250c93

 - OS: Windows 10/11/Linux

**Additional context**
Happy to try fixing this if someone can suggest the best way for a Go novice - I think relaxing [SanitiseBasename ](https://github.com/stashapp/stash/blob/70250c93f16349cec6bef25dd7b8e803dc0e1616/pkg/fsutil/file.go#L153) to allow non-alphanumeric would be the simplest but would still allow any objects where the name isn't unique to overwrite each other. Using the internal ID as part of the file name would be the surest I can think of to fix that.

SandiyosDev:
closes #5032
SanitiseBasename regex strips everything not-alphanumeric giving empty filenames, took @bayured 's advice to allow unicode letters and numbers

added test cases too

### DIFF
--- a/pkg/fsutil/file.go
+++ b/pkg/fsutil/file.go
@@ -148,7 +148,7 @@ func Touch(path string) error {
 
 var (
 	replaceCharsRE = regexp.MustCompile(`[&=\\/:*"?_ ]`)
-	removeCharsRE  = regexp.MustCompile(`[^[:alnum:]-.]`)
+	removeCharsRE  = regexp.MustCompile(`[^\p{L}\p{N}\-.]`)
 	multiHyphenRE  = regexp.MustCompile(`\-+`)
 )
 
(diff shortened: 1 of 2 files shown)


SandiyosDev:
also referencing #2336 for the sake of record keeping

PR #6750, opened 2026-03-24
Added check for webp images when getting the src img for SceneWall.

bordet991:
Webp images are never used when browsing SceneWall, UI is always loading /previews aka mp4s. 
Seems to be the same all over the place, e.g hoovering a scene anywhere.

Added a check if webp images exists, in that case use it instead.
Guess you could also check "Preview type" in Interface settings but idk where that setting is located.. :)

I'm running this on my machine and it works fine anyway.

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneWallPanel.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneWallPanel.tsx
@@ -240,7 +240,9 @@ const SceneWall: React.FC<ISceneWallProps> = ({
       return {
         scene: s,
         src:
-          s.paths.preview && !erroredImgs.includes(s.paths.preview)
+          s.paths.webp && !erroredImgs.includes(s.paths.webp)
+            ? s.paths.webp
+            : s.paths.preview && !erroredImgs.includes(s.paths.preview)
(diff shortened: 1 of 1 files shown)


Gykes:
Settings -> Interface -> Scene / Marker Wall -> change to animated image

bordet991:
> Settings -> Interface -> Scene / Marker Wall -> change to animated image 

doesn't work, still loads mp4 via /preview ?

Gykes:
Do you have them generated?

EDIT: This is a support issue, not a PR issue. I would open an issue or ask on Discord/Discoure

bordet991:
Yes I have them generated.
As far as I can see the UI doesn't care and is hardcoded to always fetch the mp4 via /preview and fallback to using the screenshot. Webp images are not fetched.

So I changed it instead of creating a support ticket for it 🤷

Gykes:
So, after some further digging it looks like that page isn't respecting the settings. You found a bug but I think there is a better way to handle it. We should pass through this setting similar to how `WallItem.tsx:50-73` does it. `SceneMarkerWallPanel.tsx` will have the same issue more than likely.

PR #6757, opened 2026-03-26
Fix: Remove padFuzzyDate From Performer

### LINKED ISSUE #6752: Performer scraper does not keep date precision for birthday
### Describe the bug

I scraped a performer whose birthday in stashdb only had a year (1998). But when I scraped it, Stash converted it to 1998-01-01 even though Stash now supports partial dates.

### Steps to reproduce

1. Search for a performer from the performer tagger view:

Image

2. Click on the first Trixie E and get this dialog:

Image

### Expected behaviour

Birthday should show up as "1998" instead of "1998-01-01"

_No response_

### Stash version

v0.30.1-161-g2e48dbfc

### Device details

Chrome on Windows

### Relevant log output

```shell

```

Gykes:
Removes `padFuzzyDate` which was converting partial dates into full dates. `ParseDate()` Already supports partial dates anyway and this function was causing it to lose precision.

Fixes: https://github.com/stashapp/stash/issues/6752

UI: 
Screenshot 2026-03-26 at 16 18 55

### DIFF
--- a/pkg/stashbox/performer.go
+++ b/pkg/stashbox/performer.go
@@ -242,11 +242,11 @@ func performerFragmentToScrapedPerformer(p graphql.PerformerFragment) *models.Sc
 	}
 
 	if p.BirthDate != nil {
-		sp.Birthdate = padFuzzyDate(p.BirthDate)
+		sp.Birthdate = p.BirthDate
 	}
 
 	if p.DeathDate != nil {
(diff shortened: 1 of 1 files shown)


smith113-p:
In the bug @DogmaDragon said it's not limited to performer tagger, maybe more places need updating?

Gykes:
This is in the performer logic so it applies to any taggers.

PR #6759, opened 2026-03-27
Show stash-box name in studio/performer tagger

smith113-p:
Currently, the stash box selector component just shows the endpoint URL.

This PR makes it so that it shows the configured name as well.

I was debating between showing `Name: URL`, `Name` and `stash-box: Name`.
The latter matches what the scene tagger shows.

What I have implemented for now is this:
Screenshot 2026-03-27 140842


DogmaDragon:
Scene tagger uses `stash-box: Name` format. I would prefer to keep it consistent across all tagger interfaces.

smith113-p:
OK, now looks like this
image

PR #6761, opened 2026-03-29
Use StashIDPill for displaying the scraped stash ID

smith113-p:
This is more consistent with other places that stash IDs are shown, simplifies the code a bit, and lets you see at a glance which stash box is being used.


smith113-p:
Looks like this:
Screenshot 2026-03-29 125724

smith113-p:
For comparison, without this PR it looks like this:
image

Issue #5144, opened 2024-08-20
[Feature] Add option to generate/scrub waveforms along side videos


BonerFide:
**Is your feature request related to a problem? Please describe.**
Longer videos can be difficult to find periods of activity or dialogue in.

**Describe the solution you'd like**
Displaying a waveform similar to how funscripts/interactive videos are displayed allows active sections / dialogue sections of video to be identified quickly.

**Additional context**
FFMPEG can be used to generate waveform images
https://trac.ffmpeg.org/wiki/Waveform

discourse-stashapp:
This issue has been mentioned on **Stash**. There might be relevant details there:

https://discourse.stashapp.cc/t/audio-waveform-for-scenes/4662/2

PR #6770, opened 2026-03-30
Add per-file scene volume gain with browser local storage persistence

eb2292:
Adds per-file volume gain controls on the scene Edit tab, with values saved in the browser’s local storage and applied to the current primary file during playback. This keeps the feature fully client-side with no backend or DB changes, while preserving the existing player volume and mute behavior. This is helpful in the case of using multiple devices for playback (phone, PC, TV, etc.). 

image

_Not sure if you guys want a feature this large in core, but I’ve been using it for a while now and find it really handy for files that are too quiet or too loud. It’s especially useful for smoothing out inconsistent source volumes without changing the media files themselves. Feel free to reject._


DogmaDragon:
I believe we are moving away from storing settings in the browser. Custom fields or config file should be used for consistency.

Better placement for it would be under Filters tab instead of Edit tab as you are not editing metadata but effectively applying an effect to the player. 

I'm not a dev, so I'm not rejecting it but I think this would work better as a plugin. This is quite a niche issue and plugin system can easily implement what is needed.

Gykes:
Agreed this is probably better as a plugin.

feederbox826:
Cloning the audio context comes with a whole lot of complications - I have done it for a previous project for live waveform generation.

This looks clobbered together by an LLM, If this were to be implemented, you'd face a lot less difficulty hooking into the player and audio source directly as a video.js plugin. I think this has a lot of potential as a plugin, especialy once you export the AudioProcess you can do a lot of much cooler and practical things like [Waveform Generation](https://github.com/stashapp/stash/issues/5144) or a proper normalizer

that all being said, not a great place for it in core and especially not in it's current state (I see per-file settings but gain seems to be applied based on primary)

PR #6772, opened 2026-03-30
Update object fields in scraper documentation

DogmaDragon:
Document changes from https://github.com/stashapp/stash/pull/6449


Gykes (approved):

feederbox826 (approved):
looks like we removed URL as well. LGTM.

DogmaDragon:
> looks like we removed URL as well.

It's technically not removed, just deprecated, but I don't see a valid reason to list deprecated fields.

PR #6366, opened 2025-12-03
Custom favicon and title

WithoutPants:
Adds support for custom favicon by reading from `favicon.ico` from the config directory. Uses the custom favicon if it is present at startup.

Adds `Custom Title` setting to the interface settings. Replaces `Stash` with whatever is in the custom title. Defaults to `Stash` if empty.

Resolves #5072

Issue #5072, opened 2024-07-14
[Feature] Custom browser title and favicon


AnonTester:
It would be great if a custom browser title and favicon could be configured.

Use cases: 
- privacy to give it a different name and icon
- when using multiple instances (separate real from 3D/anime content) able to distinguish the instances in browser title and icon. Themes work for visual clue once tab is opened, but the titles/favicons remain the same.
-
Thanks.

feederbox826:
https://github.com/feederbox826/plugins/tree/main/plugins/rebrand custom browser title plugin but no favicon

PR #6775, opened 2026-03-30
populate custom title to navbar

feederbox826:
Adds on to #6366 

basic ui validation: 
image


WithoutPants:
This may not be desirable. I didn't consider it since the original issue #5072 didn't mention the navbar. I could see instances where a user wants to change the title for privacy reasons, but I don't think they'd expect the navbar title to change as well.

feederbox826:
I use it to more easily distinguish my instances as title bar is much smaller and it's at the end. In that case I'll resurrect my plugin :)

discourse-stashapp:

https://discourse.stashapp.cc/t/rebrand/1603/8

PR #6779, opened 2026-03-31
replace eslint, prettier with biome

feederbox826:
only added biome, formatter changes 300/400 files and eslint needs to be converted


feederbox826:
If it passes I'll just add the rules and revert code changes - keep it for another PR to not mix in anything else

slick-daddy:
I would prefer Oxlint since it has 100% Eslint test coverage and it is faster than Biome.

Link: https://oxc.rs/docs/guide/usage/linter.html

Migrate from Eslint: https://oxc.rs/docs/guide/usage/linter/migrate-from-eslint.html

feederbox826:
- matched stash-box https://github.com/stashapp/stash-box/pull/980
- no prettier

once stylelint/ scss support lands we can switch to one unified solution

WithoutPants:
Please rebase this against the latest `develop` version. `.stylelintcache` needs to be added to `.gitignore`.

WithoutPants:
Closing in favour of #6996.

PR #6783, opened 2026-03-31
replace thehandy with ive-connect

feederbox826:
## ***no integration testing done***

server time offset is done automatically by ive-connect in private functions so it was removed from context, no LLMs were used so there might be leftovers that eslint didn't catch. The API is much simpler but has a smaller surface, does csv conversion automatically.

Kind of clobbered together as I don't have much context for thehandy and ive-connect docs are sparse and a bit out of place


ui/v2.5/src/hooks/Interactive/interactive.ts
```diff
@@ -1,224 +1,106 @@
...
-      const fileName = `${Math.round(Math.random() * 100000000)}.csv`;
-      const csvFile = new File([csv], fileName);
-
-      funscriptUrl = await uploadCsv(csvFile).then((response) => response.url);
+    const result = await loadScript({
+      type: "funscript",
+      url: funscriptURL,
+    });
+    if (result.error) {
+      throw new Error(result.error);
```
xtc1337, line 72: This just loads the funscript data you will still need to call `device.prepareScript`

ui/v2.5/src/hooks/Interactive/interactive.ts
```diff
@@ -1,224 +1,106 @@
...
-    await this._handy.getStatus();
-  }
-
-  async sync() {
-    return this._handy.getServerTimeOffset();
   }
 
-  setServerTimeOffset(offset: number) {
-    this._handy.estimatedServerTimeOffset = offset;
+  sync() {
```
xtc1337, line 76: Could we still have this as  `async` to allow plugins to have more flexibility to the underlying `HapticDevice` needs.  I'm currently using this to support 3 different haptic interfaces (theHandy sdk we have now, `ive-connect` fw4 handy and `ive-connect` buttplug.io support). This will avoid having a breaking change.

ui/v2.5/src/hooks/Interactive/interactive.ts
```diff
@@ -1,224 +1,106 @@
...
     this.handyKey = config.connectionKey ?? this.handyKey;
-    this._handy.estimatedServerTimeOffset =
-      config.estimatedServerTimeOffset ?? this._handy.estimatedServerTimeOffset;
-    this.useStashHostedFunscript =
-      config.useStashHostedFunscript ?? this.useStashHostedFunscript;
   }
 
-  async play(position: number) {
-    if (!this._connected) {
+  async play(position: number, loop: boolean = false) {
```
xtc1337, line 86: if we are going to add this support I would also like to surface mimicking the `play` interface from `HapticDevice` which includes `playbackRate`
```ts
    /**
     * Play the loaded script at the specified time
     * @param timeMs Current time in milliseconds
     * @param playbackRate Playback rate (1.0 = normal speed)
     * @param loop Whether to loop the script
     */
    play(timeMs: number, playbackRate?: number, loop?: boolean): Promise<boolean>;
```

xtc1337:
Hi there, 
I wanted to offer some feedback on the direction here. Some background, I have a plugin([alpha](https://github.com/xtc1337/StashInteractiveTools/blob/alpha/stash-alpha.yml) version for testing) that currently uses both the default handy setup as well as `ive-connect`. This was possible via an addition I made a while back #5856.

If you have some time I would like to direct you to some reference work on some quirks with `ive-connect` in this type of flow. The currently api is pretty good and provide the calls. Anyways here are some examples 
- [patches](https://github.com/xtc1337/StashInteractiveTools/tree/alpha/src/utils/interactive/patches) - this was created in a way to move the management code outside of the main InteractiveClient/API. This allowed me to do other things a bit easier such as adding ivdb.io token support (there is a but in the lib that doesn't support it now) as well as handle the support for inmemory funscript alterations.
- [client.ts](https://github.com/xtc1337/StashInteractiveTools/blob/alpha/src/utils/interactive/client.ts) - this is the port to the `ive-connect` `HapticDevice` interface, it is basically a direct copy of the current `InteractiveAPI` code just routed `HapticDevice`. I will agree this abstraction is a bit much but it was needed to support in the current state.

The basic idea is as so :
- Swithch out the `InteractiveClientProvider` to take full control of what devices are supported
- Ported the current handy fw3 support to the `HapticDevice` interface to have the same api for all supported devices
- Provide a way for the user to choose which device via PluginSettings
- Handle any pre/after logic within the plugin itself before passing it to the `HapticDevice` (this is where we patch `InteractiveClientProvider`, I could have put that all within the provider itself but I wanted to have a bit more flexibility of where the code lives) 

One last note, buttplug and handy data flow are completely different. I know on discord there where some talks to support this, the only way to do that would be to provide a way to switch out `InteractiveAPI` instance to delegate the work to it (which can be provided by the work done in #5856).

Anyways, thanks for taking a look into this and adding it to core, my only real concern is the drop of `async` for sync, everything else wouldn't effect the extra support added by StashInteractiveTools (SIT)

feederbox826:
Thanks! I have absolutely no idea how any of this works so I wanted to get as close to drop-in as I could for integration testing. Once that's all clear then I'd be more comfortable making actual functional changes, otherwise we'd just be 4 steps back, 1 step forward. I'll try to keep that in mind for the next interation but i really just want to make sure this works first

ui/v2.5/src/hooks/Interactive/interactive.ts
feederbox826: but ive-connect has it sync? Can wrap it in a promise, sure

ui/v2.5/src/hooks/Interactive/interactive.ts
feederbox826: https://github.com/iveplay/ive-connect/blob/main/src/devices/handy/handy-device.ts#L287 ive-connect seems to just expose prepareScript which includes both?

ui/v2.5/src/hooks/Interactive/interactive.ts
feederbox826: Sure. IIRC this was a patch since ive-connect has no loop() function, it's tied into play

PR #6792, opened 2026-04-02
Fix: Update Postmigration 84 to Handle De-Duplicate of Folders.

### LINKED ISSUE #6788: DB migration error on Windows

### Describe the bug

I am seeing the following error message when attempting to upgrade to Stash 0.31.0:

> An error occurred migrating the database to the latest schema version. The backup database file was automatically renamed to restore the database.
error performing migration: error running migration for schema 84: running post migrations for schema version 84: migrating folders: error migrating folder 247 "\\data\\movies": UNIQUE constraint failed: folders.parent_folder_id, folders.basename

I am running the precompiled version of Stash on Windows Server 2019.  I am willing to send my entire database, but it's over 570mb in size so I can't attach it to this ticket.

### Steps to reproduce

1. Have an install of Stash 0.30.1 on Windows
2. Download Stash 0.31.0 on Windows
3. Attempt to upgrade.

### Expected behaviour

I... should be able to use Stash.

_No response_

### Stash version

0.31.0

### Device details

_No response_

### Relevant log output

```shell

```

Gykes:
!!!Drafted with AI!!! 
It generated so much extra junk and I pulled most of it out. Updated the comments as well to make it easier to read.

A couple of users have had issues with the recent migration. From what I can gather its mostly users who go from Linux -> Windows . What I think was happening is some old file paths were being left behind. Windows can handle `/` or `\` in it's pathing structure. Both forward and backslashes were getting normalized causing unique failures. This new addition to the migration checks for this, groups all the folders, checks for duplicates, tries to keep the folder structure that matches current OS convention, falls back to NEWEST ID if it cant be done, orphans the other one for the clean task to be able to clean, then loops till 0 conflicts exist. The principal of the newest ID being that this should be the most recent OS folder structure and would be preferred.

Didn't test this against an actual DB, logic seems sound.

Closes: https://github.com/stashapp/stash/issues/6788

### DIFF
--- a/pkg/sqlite/migrations/84_migrate.go
+++ b/pkg/sqlite/migrations/84_migrate.go
@@ -425,12 +425,32 @@ func (m *schema84Migrator) mergeFolder(tx *sqlx.Tx, keepID, dupID int) error {
 		return fmt.Errorf("re-parenting child folders: %w", err)
 	}
 
-	// Orphan the stale duplicate folder by clearing its parent so the UNIQUE
-	// constraint on (parent_folder_id, basename) won't be violated when
-	// migrateFolders sets basenames. Any stale file entries under it are left
-	// untouched — the clean task will handle them on the next scan.
-	if _, err := tx.Exec("UPDATE folders SET parent_folder_id = NULL WHERE id = ?", dupID); err != nil {
-		return fmt.Errorf("orphaning duplicate folder: %w", err)
+	// re-parent any files under the duplicate folder to the canonical folder.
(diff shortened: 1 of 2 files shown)


WithoutPants:
Updated after testing against user database exhibiting similar behaviour, as well as importing manually from the csv in the linked issue.

Split the post-migration execution to perform some steps prior to schema migration. I did this because in the other user database, there were duplicate folder entries with equal paths. This would cause the schema migration to fail due to the unique path value constraint. Changed the deduplication behaviour to also reparent files, and delete the duplicate folder entry if it is not related to a gallery.

WithoutPants (approved):

PR #6498, opened 2026-01-17
Refactor file scanning and handling logic

WithoutPants:
This is preparatory work for a future synchronous `scanFile` graphql interface. This PR moves the directory walking and queuing functionality out of the `file` package and into the calling scan task code. Includes removal of some vestigial functionality and some refactoring. No specific regressions are expected, but given the complexity of the scan code some regression testing will be necessary.

Closes #3800

PR #6796, opened 2026-04-04
fix memory leak

### LINKED ISSUE #6794: persistent memory leak post-scan

### Describe the bug

Memory usage seems to grow abnormally after scanning

### Steps to reproduce

1. Selective scan scenes
2. Watch memory usage explode

### Expected behaviour

Memory usage is GC'd after scanning

Reported by Eleniatari in discord and further investigated. It is confirmed not to be any plugins, it is not just the size of the library and it's not just images in zips blowing up the space. Something is not being GC'd

### Stash version

0.31.0, v0.30.1-177-g2da80743

bisect leads to #6498 

### Device details

Brave, Chrome, headless instances

### Relevant log output

| Canonical Name | Scene Size | Image Size | Pre-Scan | RSS After |
|---|---|---|---|---|
| main | 11.2T/4044 | 0 | 82M | 1G |
| phstash | 363G/940 | 70M/1673 | 60M | 125M |
| orgstash | 2.2T/1999 | 25G/9324 | 9M | 119M |
| vr | 1T/83 | 0 | 10M | 94M |
| scrape | 3.1G/5 | 9M/13 | 61M | 116M |
| scrape.feederbox.cc | 0 | 0 | 32M | 46M |
| scrape-ci | 0 | 0 | 26M | 38M |
| eleniatari| 28T/27K | 650G/165K | 25M | ~500M |

the mamory is not GC'd. it persists.

feederbox826:
fixes #6794. The logic according to LLM was that return can break the queue and cause it to remain allocated, and fileQueue and scanner both kill the channel and any hanging handlers. There were other fixes recommended but after trial and error this is the minimum necessary to prevent memory leak. The memory will spike (by ~14M) on first scan, more than 0.30.1, and does spike again on subsequent scans but does eventually GC down to the first spike instead of running away.

IO spike corresponds to 'scan" button being hit. First climb is from ~34M

image

### DIFF
--- a/internal/manager/task_scan.go
+++ b/internal/manager/task_scan.go
@@ -81,6 +81,7 @@ func (j *ScanJob) Execute(ctx context.Context, progress *job.Progress) error {
 	j.runJob(ctx, paths, nTasks, progress)
 
 	taskQueue.Close()
+	j.releaseScanResources()
 
 	if job.IsCancelled(ctx) {
 		logger.Info("Stopping due to user request")
@@ -129,6 +130,11 @@ func (j *ScanJob) runJob(ctx context.Context, paths []string, nTasks int, progre
 }
 
 const scanQueueSize = 200000
+
+func (j *ScanJob) releaseScanResources() {
+	j.fileQueue = nil
+	j.scanner = nil
+}
 
 func (j *ScanJob) queueFiles(ctx context.Context, paths []string, progress *job.Progress) error {
 	fs := &file.OsFS{}
@@ -283,8 +289,10 @@ func (j *ScanJob) processQueue(ctx context.Context, parallelTasks int, progress 
 
 		for f := range j.fileQueue {
 			logger.Tracef("Processing queued file %s", f.Path)
-			if err := ctx.Err(); err != nil {
-				return
+			if ctx.Err() != nil {
+				// Keep receiving until queueFiles closes the channel; otherwise
+				// the walker can block on send (full buffer) and never finish.
+				continue
 			}
 
 			wg.Add()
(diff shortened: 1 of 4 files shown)


Gykes (commented):
- The return -> continue changes are correct and I can agree with them.

- So, I haven't done any testing but I think the rest of this is more of a bandaid. I think the issue is in `manager.go:236`. From an overall look it seems that `*job` is keeping `exec` alive when it shouldn't be.  The job manangers graveyard is probably holding a reference keeping everything alive in memory rather than dumping it.

feederbox826:
> So, I haven't done any testing but I think the rest of this is more of a bandaid.

I agree but I haven't the faintest of where to look - if either one of fileQueue/ scanner is excluded, there is memory leak at a slower pace - consistent 5-10MB/scan instead

Gykes:
I believe the reason `scanner` and `fileQueue` aren't GC'd after `Execute()` returns is that the job manager's graveyard holds a reference to the `*Job`, which holds `exec` (the ScanJob), keeping everything alive. The graveyard never released executors, this was always the case but was benign until https://github.com/stashapp/stash/pull/6498 moved heavy resources (fileQueue, scanner) from local scope into `ScanJob` struct fields. That's what made the retention costly.

feederbox826:
> I believe the reason `scanner` and `fileQueue` aren't GC'd after `Execute()` returns is that the job manager's graveyard holds a reference to the `*Job`, which holds `exec` (the ScanJob), keeping everything alive. The graveyard never released executors, this was always the case but was benign until #6498 moved heavy resources (fileQueue, scanner) from local scope into `ScanJob` struct fields. That's what made the retention costly.

:-1:

image

feederbox826:
2fd7045e43d9c8a687ebe48dbab1faede124341a does the trick ✅

PR #6801, opened 2026-04-06
fix: run stash container as non-root user (closes #684)

### LINKED ISSUE #684: [Feature] non-privileged user in Docker build

TeapoyY:
## Summary

This PR fixes issue #684 by running the stash Docker container as a non-root user instead of root.

### docker/build/x86_64/Dockerfile
- Added gosu package for secure privilege dropping
- Added build args PUID and PGID (default: 9999) for configurable UID/GID
- Create a stash user with the configured UID/GID
- Set up /root/.stash directory owned by the stash user
- Use entrypoint.sh instead of running the binary directly

### docker/entrypoint.sh (new file)
- Entrypoint script that drops privileges to the stash user
- Supports PUID and PGID environment variables
- If running as root, creates/updates the stash user with the right UID/GID and switches to it using gosu
- If already running as the correct user, executes directly

### Default (UID 9999)
`ash
docker run -v /path/to/data:/root/.stash stashapp/stash
`

### Custom UID/GID (to match host user)
`ash
docker build --build-arg PUID=1000 --build-arg PGID=1000 -t stash .
docker run -e PUID=1000 -e PGID=1000 -v /path/to/data:/root/.stash stashapp/stash
`

## Security Impact

Running containers as root is a security risk. With this change:
- The stash container follows Docker security best practices
- The stash process runs with a non-privileged user
- UID/GID is configurable to support different volume mount scenarios

## References
- Fixes https://github.com/stashapp/stash/issues/684
- Bounty: OpenCollective bounty posted on issue

### DIFF
--- a/docker/build/x86_64/Dockerfile
+++ b/docker/build/x86_64/Dockerfile
@@ -30,15 +30,33 @@ COPY ./internal /stash/internal/
 # needed for generate-login-locale
 COPY ./ui /stash/ui/
 RUN make generate-backend generate-login-locale
-COPY --from=frontend /stash /stash/
+COPY --from=frontend /stash/stash/ /stash/
 ARG GITHASH
 ARG STASH_VERSION
 RUN make flags-release flags-pie stash
 
 # Final Runnable Image
+# Build arguments for custom UID/GID (default: 9999)
+ARG PUID=9999
+ARG PGID=9999
 FROM alpine:latest
-RUN apk add --no-cache ca-certificates vips-tools ffmpeg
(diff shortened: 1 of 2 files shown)


DogmaDragon:
Breaks image for existing users.

feederbox826:
1. you're creating the user twice instead of changing UID/GID
2. you're doing uid checks twice in entrypoint
3. still won't work since /root permissions are broken, unless you redefine HOME

Gykes:
Closing this as it seems to be llm nonsense. If OP wants to come back and have a discussion on it they are welcome to reopen to discuss.

PR #6802, opened 2026-04-06
Normalize manual headers

DogmaDragon:
- Normalized headers across all the manual pages
- Updated Browsing section to make it easier to read and understand
- Small formatting changes and typo fixes
- Added GitHub Sponsors link under financial contributions section

### DIFF
--- a/ui/v2.5/src/docs/en/Manual/ExternalPlugins.md
+++ b/ui/v2.5/src/docs/en/Manual/ExternalPlugins.md
@@ -1,4 +1,4 @@
-# External Plugin Tasks
+# External plugin tasks
 
 External plugin tasks are executed by running an external binary.
 
--- a/ui/v2.5/src/docs/en/Manual/AutoTagging.md
+++ b/ui/v2.5/src/docs/en/Manual/AutoTagging.md
@@ -1,4 +1,4 @@
-# Auto Tag
+# Auto tag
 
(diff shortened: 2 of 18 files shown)


feederbox826 (approved):

ui/v2.5/src/docs/en/Manual/ExternalPlugins.md
```diff
@@ -1,4 +1,4 @@
-# External Plugin Tasks
+# External plugin tasks
```
Gykes, line 1: Would it not be better to keep title casing for these?

Gykes (commented):
Wont make the same comment on every line but title casing?

ui/v2.5/src/docs/en/Manual/ExternalPlugins.md
DogmaDragon: If you mean the section headers specifically, I used the same sentence case as I don't like mixed capitalization rules. 

Technically we could use title case for level 1 headings and sentence case for the rest, but to me it looks unorganized.

Gykes: Okay, whatever way you wanna do it. Ultimately it's not a big deal to me.

Gykes (approved):

PR #6810, opened 2026-04-08
Fix WebSocket UTF-8 error for non-UTF-8 file paths in subscriptions

dev-null-life:
## Summary
- Sanitize log messages in `LoggingSubscribe` before sending over WebSocket
- Sanitize job description, subtasks, and error in `jobToJobModel` (used by `JobsSubscribe`, `JobQueue`, and `FindJob`)

## Why
WebSocket text frames must contain valid UTF-8. File paths with non-UTF-8 characters (common with certain media files or filesystem encodings) caused the browser to close the connection with: _"Could not decode a text frame as UTF-8."_

Invalid bytes are replaced with the Unicode replacement character (U+FFFD) so messages remain readable.

Note: the sanitization only applies to the API response layer — it operates on copies of the data structures used for the WebSocket response. The underlying stored data (database records, file paths on disk) is never touched.

## Test plan
- [ ] Scan a library containing files with non-UTF-8 characters in their paths
- [ ] Confirm the WebSocket connection stays open and log/job updates appear normally in the UI

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/internal/api/resolver_query_job.go
+++ b/internal/api/resolver_query_job.go
@@ -3,7 +3,6 @@ package api
 import (
 	"context"
 	"strconv"
-	"strings"
 
 	"github.com/stashapp/stash/internal/manager"
 	"github.com/stashapp/stash/pkg/job"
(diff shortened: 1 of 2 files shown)


WithoutPants:
Tested original issue with the following steps:
- download new image file into library directory
- rename file to include invalid utf character. In bash, this can be done like so: `mv <filename> $'renamed-\xff.jpg'`
- I had to comment out the throttling code in `job.updater.updateProgress` (pkg/job/manager.go`) so that it notified on every update, otherwise it would skip sending the message with the filename (only for job update messages; log subscription does not have this issue)
- run a scan on the path containing the file

The following message would be output to the browser console:
```
WebSocket connection to 'ws://<host>/graphql' failed: Could not decode a text frame as UTF-8.
```

The websocket connection is lost and reconnects afterwards.

Verified that the fix addresses the issue with both logging and job subscriptions, outputting the correct string with the invalid UTF-8 sequences replaced with �.

WithoutPants (approved):
I replaced the direct calls to `strings.ToValidUTF8` with a `sanitiseWebsocketString` function to better illustrate of this change.

PR #6828, opened 2026-04-14
[backend-only] cdp validation

feederbox826:
took more time then I would've liked but it's working. Unsure of how to proceed with UI as it's just a string input :/ or if it should be merged backend-only?


graphql/schema/schema.graphql
```diff
@@ -247,6 +247,9 @@ type Query {
   ): Directory!
   validateStashBoxCredentials(input: StashBoxInput!): StashBoxValidationResult!
 
+  # Validate remote CDP path
+  ValidateCDPPath: CDPValidationResult!
```
Gykes, line 251: Would it be good to accept an input parameter here so that users can test before saving? Currently you cant test and it just saves regardless of it it is correct or not. 

Should also do camel case here

pkg/scraper/url.go
```diff
@@ -329,3 +329,19 @@ func splitProxyAuth(proxyUrl string) (string, string, string) {
...
 }
+
+func TestRemoteCDP(globalConfig GlobalConfig) error {
+	cdpPath := globalConfig.GetScraperCDPPath()
+	if cdpPath == "" {
+		return fmt.Errorf("CDP path is empty")
+	}
+	if !isCDPPathHTTP(globalConfig) {
+		// unable to test non-http CDP
+		return fmt.Errorf("Unable to test non-http CDP paths")
```
Gykes, line 340: So, should this be an error? I'm not sure on this one but it seems to me that the path could be valid but just not tested. It would return `valid:false` then spit the error out to the user. It could be misleading because it's not technically an error just not tested.

Perhaps it should be info level or something unless it's actually invalid.

Gykes (changes requested):

pkg/scraper/url.go
feederbox826: using the current way to fetch it, it cant handle it. We would need to do an actual xpath scrape to validate PATHs.

graphql/schema/schema.graphql
feederbox826: debated doing that as well

PR #6838, opened 2026-04-16
Fix: Correct stash ID links in merge dialogs

### LINKED ISSUE #7064: Tag merge; stash ID uses scene url format instead of tags

### Describe the bug

When merging two tags. If one had a stashID and you follow the URL in the merge compare window. It uses the scene url format causing a 'scene not found' error. 

### Steps to reproduce

Merge two tags, one at least having a stash ID, before clicking accept follow the url. 

### Expected behaviour

Use the stash tag URL here. 

_No response_

### Stash version

v0.31.1

### Device details

_No response_

### Relevant log output

```shell

```

smith113-p:
The <StashIDsField/> element hardcoded a link type of "scenes", so the tag and performer merge dialogs had incorrect links.

Reported in Discord #bugs


Gykes (approved):
LGTM. Thanks for fixing it on the rest of the objects as well.

PR #6849, opened 2026-04-20
Add missing (empty) dependency array to `useEffect()`

megatwig:
`useEffect` without a dependency array is re-evaluated on [every single render](https://react.dev/reference/react/useEffect#specifying-reactive-dependencies:~:text=fix%20your%20logic.-,My%20Effect%20runs%20after%20every%20re%2Drender,First%2C%20check%20that%20you%20haven%E2%80%99t%20forgotten%20to%20specify%20the%20dependency%20array%3A,-useEffect).

Most of these are just binding (and unbinding) keyboard shortcuts, so should be pretty safe to update in bulk.

I did spot a couple that skip keybind listeners if `!isVisible`, so set as a dependency so they'll still get installed.

There's an [eslint check](https://react.dev/reference/eslint-plugin-react-hooks/lints/exhaustive-deps) which can flag this (and other deps issues) - but the version of `eslint-plugin-react-hooks` configured is ancient and doesn't seem to have it. I briefly tried upgrading and enabling, but `eslint-config-airbnb` complains it's too new, and it surfaces 57 new lint errors (222 with with full `react-hooks/recommended` enabled).

### DIFF
--- a/ui/v2.5/src/hooks/LocalForage.ts
+++ b/ui/v2.5/src/hooks/LocalForage.ts
@@ -83,7 +83,7 @@ export function useLocalForage<T extends {}>(
       };
       localForage.setItem(key, Cache[key]);
     }
-  });
+  }, []);
 
   const isLoading = loading || loading === undefined;
 
--- a/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryChaptersPanel.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryDetails/GalleryChaptersPanel.tsx
@@ -30,7 +30,7 @@ export const GalleryChapterPanel: React.FC<IGalleryChapterPanelProps> = ({
     return () => {
       Mousetrap.unbind("n");
     };
-  });
+  }, [isVisible]);
 
   function onOpenEditor(chapter?: GQL.GalleryChapterDataFragment) {
     setIsEditorOpen(true);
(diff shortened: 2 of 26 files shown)


ui/v2.5/src/hooks/LocalForage.ts
```diff
@@ -83,7 +83,7 @@ export function useLocalForage<T extends {}>(
       };
       localForage.setItem(key, Cache[key]);
     }
-  });
+  }, []);
```
Gykes, line 86: Would this not break config persistence? This would make it run only on mount then whenever `setData(...)` is called it wont rerun the effect. 

I think it should be `[key, data]`

Gykes (changes requested):
I started to review near the bottom but decided it would be better to just do a generic comment. 

My concern is that quite a few of these effects bind callbacks that close over state which changes after mount. Swapping them all the `[]` would freeze them as is and not update.

WithoutPants:
With the biome changes, the react dependencies are once again being validated correctly. Rebasing this onto the current develop branch shows 71 `lint/correctness/useExhaustiveDependencies` errors introduced with this change. This PR also reintroduced an dependency array in `useStopWheelScroll` which explicitly states why the array was removed. I'm not convinced that changing these things in bulk is a great idea. 

From memory, the keybind `useEffect` calls specifically don't use dependency arrays because the bind/unbind behaviour has a low performance impact, and having to manage dependencies for these adds too much cruft.

PR #6854, opened 2026-04-21
[Feature] Tagger performer match improvements

### LINKED ISSUE #6853: Feature Request: Performer Match Preview Card in Tagger
## Summary
When Scene Tagger proposes a performer match, show a hover preview card for the matched local performer so users can quickly verify identity before linking/saving.

## Problem
In Scene Tagger, proposed performer matches are currently text-only.  
As performer libraries grow, name collisions and near-matches become more common, increasing accidental cross-link risk.

Today, users often need to click into performer pages to confirm identity, which slows down tagging and increases the chance of incorrect links.

Image

## Proposed Behavior
For each **proposed matched performer** in Scene Tagger:

- Show a hover preview on the UI elements for matching simmilar to the performer hover behavior used in scene contexts.
- Preview card size: approximately **215x300px** (roughtly double the existing UI item on the scene card).
- Card content:
  - performer image
  - performer name
  - styled similarly to existing scene-card/performer card presentation
- The matched performer name should remain an anchor/link to open the performer page directly.

Image

## UX Notes
- Hovering the matched performer should show the preview quickly and consistently.
- If no performer image exists, show fallback text/thumbnail behavior consistent with existing performer UI patterns.
- Keep interaction lightweight; no extra clicks or buttons required.

This feature improves match confidence at the decision point, reducing accidental performer cross-linking when multiple people have similar names. For existing users, the outcome is faster and more accurate tagging because identity verification happens inline instead of requiring extra page navigation. In larger libraries, this should reduce correction/rework effort (unlinking and relinking performers) and improve overall data quality over time.

Users can partially work around this today by manually opening each matched performer link before saving, but that adds friction and slows down tagging workflows. It is also theoretically possible to script a custom solution via userscript/plugin, but that approach requires DOM parsing/scraping of match rows and is less accurate and less stable across UI changes. The proposed built-in behavior provides a consistent, reliable, first-class UX for all users.

### Additional context

_No response_

Stash-KennyG:
# Enhance Tagger performer match UX with hover previews and mismatch warnings #

Provides a performer hover-pop-over to the tagger UI so performers can be clearly seen and verified against potential match candidates before committing stashID changes in the local database.
  
Closes #6853 

Improves match confidence at the decision point, reducing accidental performer cross-linking when multiple people have similar names.

For existing users, this should improve speed and tagging accuracy because identity verification happens inline instead of requiring extra page navigation.

In larger libraries, this should reduce correction/rework effort (unlinking and relinking performers) and improve long-term data quality.

Additional operational benefits:

- lower cognitive load during tagging (less context switching);
- faster moderation/review workflows for users curating shared libraries;
- fewer follow-up data repairs caused by incorrect performer associations.

## Scope and Non-Goals

- Scope is intentionally limited to the **Tagger workflow** (Scene Tagger performer match context).
- This does **not** attempt to redesign global performer hover behavior across all pages.
- This is a UX assist for matching decisions, not an automatic identity resolution system.

## Platform and Interaction Considerations

- Current interaction model is hover-based, which is optimized for desktop pointer devices.
- On mobile/touch interfaces, hover interactions are naturally limited or unavailable.
- This is acceptable for initial delivery given the Tagger workflow is primarily desktop-driven, but should be disclosed for product discussion.

## Functionality Details
- Adds a hover preview for matched performer selection in Scene Tagger by attaching the preview to the picker control, when a match is suggested by name
image

- Adds a stashbox performer hover preview on the left performer label using scraped/remote data, while rendering through shared performer card UI for visual consistency.
image

- Adds mismatch context in the matched performer hover card for non-matching performer fields (including aliases/URLs as counts), to make potential cross-linkage easier to spot before save.
image

- Adds stash ID conflict warning inside the hover card when the local selected performer already has a stash ID for the same endpoint but with a different GUID.  Uses a compact red endpoint chip for stash ID conflict warning, with full conflicting GUID available via chip tooltip.
image

## Behavior Details
- **Picker hover target**: hover card opens from the selected picker control.
- **Left-side scraped hover**: hover on scraped performer text shows card built from scraped data (image/name/etc.) but styled with shared `PerformerCard`.
- **Mismatch rows**: shown only for fields where scraped has a value and differs from selected local value.
- **Aliases / URLs**: reported as counts only.
- **Stash ID conflict warning**:
  - shown only when `attached stash_id != incoming remote_site_id` AND `attached stash_endpoint == incoming stash_endpoint`
  - displayed inside card, above mismatch rows
  - compact red endpoint chip; GUID is tooltip text.

## Mismatch Fields Covers
- Birthdate
- Death Date
- Ethnicity
- Hair Color
- Eye Color
- Height
- Weight
- Penis Length
- Circumcised
- Measurements
- Fake Tits
- Tattoos
- Piercings
- Career Start
- Career End
- Aliases (count)
- URLs (count)

## Test Plan
- [✅] Hover selected performer picker in Scene Tagger matched result and confirm card opens below picker.
- [✅] Hover left scraped performer name and confirm scraped preview card appears and matches shared performer card styling.
- [✅] Confirm mismatch rows render only for differing values and are hidden for exact matches.
- [✅] Confirm aliases/URLs mismatch rows show only in Remote > Local.
- [✅] Confirm stash ID warning appears only when endpoint matches and GUID differs.
- [✅] Confirm warning chip is compact red endpoint chip and GUID is visible in tooltip.
- [✅] Confirm no stash ID warning appears when GUIDs match.
- [✅] Check regressions
  - Performer card on scene details unaffected
  - Performer popover on scene card unaffected
  - Tagger workflow unaffected on scenes and performer view
- [✅] Confirm lint/build checks pass.

## How to Demonstrate Feaure

Tagger will optomistically match when $stashDB.name == $local.name  and $local.count == 1.  We need to intentionally create a missmatch of the performer to create this opportunity to look at this functionality.  The easiest way to do this is:

* Take a scene that has an existing stash match in your library with a relatively unique performer name, i.e. "Angla White" not "Angela"
* Edit your existing local performer to remove their stashID.  
* Rescrape the scene
image

### Force a miss-match warning
* You can force a missmatch by intentionally skewing the performer data so that it will missmatch.
image

## Notes
- This feature is designed to improve confidence in match correctness before save, especially in larger datasets where name collisions are more common.

### LLM Details
[✔] LLM use is openly disclosed. Made with [Cursor](https://cursor.com/)
[✔] Code is reviewed by a human. 
[✔] Human testing and validation was performed. 
[✔] You take full responsibility for the code (including license compliance).


ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
```diff
@@ -12,6 +12,99 @@ import { getStashboxBase } from "src/utils/stashbox";
...
+  pushDeltaIfDifferent(rows, "Hair Color", remote.hair_color, local.hair_color);
+  pushDeltaIfDifferent(rows, "Eye Color", remote.eye_color, local.eye_color);
+  pushDeltaIfDifferent(rows, "Height", remote.height, local.height_cm);
+  pushDeltaIfDifferent(rows, "Weight", remote.weight, local.weight);
+  pushDeltaIfDifferent(
+    rows,
+    "Penis Length",
+    remote.penis_length,
+    local.penis_length
+  );
```
Gykes: All of these have existing internationalization. You can modify to take `labelId` and resolve via `intl.formatMessage`. I'm not going to comment on every single one but they all should be internationalized.

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
```diff
@@ -0,0 +1,88 @@
...
+    : [];
+
+  return {
+    id:
+      performer.stored_id ??
+      performer.remote_site_id ??
+      `scraped-${
+        performer.name?.replace(/\s+/g, "-").toLowerCase() ?? "performer"
+      }`,
+    name: performer.name ?? "Unknown performer",
```
Gykes: internationalize

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
```diff
@@ -73,8 +166,13 @@ const PerformerResult: React.FC<IPerformerResultProps> = ({
...
       stashID.stash_id === performer.remote_site_id
   );
-
   const [selectedPerformer, setSelectedPerformer] = useState<Performer>();
+  const { data: selectedPerformerData, loading: selectedPerformerLoading } =
+    GQL.useFindPerformerQuery({
+      variables: { id: selectedID ?? "" },
+      skip: !selectedID,
+    });
+  const selectedPerformerDetails = selectedPerformerData?.findPerformer;
```
Gykes: If i'm understanding here this will add an extra quert per row on top of the `selectedPerformer` that is already populated. For larger tagging batching, which I assume is what you are catering to, this would add a lot of unneeded fetches. We could reuse the data that already exists from `performerSelect` or lazy-fetch them on hover. 

Thoughts?

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
```diff
@@ -107,18 +205,21 @@ const PerformerResult: React.FC<IPerformerResultProps> = ({
     selectPerformer(undefined);
   };
 
-  if (stashLoading) return <div>Loading performer</div>;
+  if (stashLoading || selectedPerformerLoading)
+    return <div>Loading performer</div>;
```
Gykes: Wouldn't this cause the entire row, including the picker, to blank out to "Loading performer" every time a new match is picked?

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
```diff
@@ -12,6 +12,99 @@ import { getStashboxBase } from "src/utils/stashbox";
...
+
+interface IPerformerDeltaRow {
+  label: string;
+  value: string;
+}
+
+const normalizeValue = (value: unknown) =>
+  String(value ?? "")
+    .trim()
+    .toLowerCase();
```
Gykes: `toLoweCase()` shouldn't be used on fields like `height_cm`. This will cause `180.0` to not equal `180` and then fail the check. Should break numbers out and use `Number()`

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
```diff
@@ -0,0 +1,88 @@
...
+    ignore_auto_tag: false,
+    scene_count: null,
+    image_count: null,
+    gallery_count: null,
+    group_count: null,
+    performer_count: null,
+    o_counter: null,
+    rating100: null,
+    urls: performer.urls ?? [],
+  } as unknown as GQL.PerformerDataFragment;
```
Gykes: Not a fan of the as unknown as double cast here. It completely disables type checking, so if `PerformerCard` ever requires a new field this would break without warning. We could either loosen `PerformerCard` to accept a partial/minimal type, or build a lighter scraped-preview view instead of faking a full `PerformerDataFragment`?

Thoughts?

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
```diff
@@ -0,0 +1,88 @@
...
+  );
+};
+
+export const ScrapedPerformerPreview = ({
+  performer,
+  placement = "bottom",
+  children,
+}: IScrapedPerformerPreviewProps) => {
+  const { configuration: config } = useConfigurationContext();
+  const showPerformerCardOnHover = config?.ui.showTagCardOnHover ?? true;
```
Gykes: Is this the correct config?

Gykes (changes requested):
Static review only, no actual testing was done.

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
Stash-KennyG: Will take a look... good suggestion

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
Stash-KennyG: Valid critique here.  
Design consideration was to avoid introducing new types.  Using the full data fragment felt heavy, but I see your point.  I think I actually can just use the original ScrapedPerformer object. 
Working on that now.

Stash-KennyG: I refactored this so we no longer fake a PerformerDataFragment for scraped data.

Current approach:

Kept PerformerCard strict for local/full performer data.
Split preview rendering into two wrappers:
LocalPerformerCard (uses native PerformerCard with PerformerDataFragment and appends children)
RemotePerformerCard (generates a proxy Card  directly from ScrapedPerformer data)
The RemotePerformerCard uses the the remote renderer, preserving the same visual style as much as possible.
This removes the unsafe cast and avoids broadening PerformerCard’s contract.

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
Stash-KennyG: This is a continuation of the existing behavior where we are already doing this on the performer popover (mainline reference: line 54).

[`ui/v2.5/src/components/Performers/PerformerPopover.tsx`](https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/components/Performers/PerformerPopover.tsx#L54)
```tsx
const showPerformerCardOnHover = config?.ui.showTagCardOnHover ?? true;
```
This was added by https://github.com/stashapp/stash/pull/5739

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
Stash-KennyG: Por Que No Los Dos?
Practical pattern
Keep selectedPerformer as primary row state (no eager per-row full query).
In hover component (MatchedPerformerPreview / local preview wrapper), on first onOpen:
fetch findPerformer(id) if full data not already available.
Cache fetched result by performer ID (Apollo cache or local map) so repeated hovers don’t refetch.
Render:
immediate card from local/select data if available
enrich with warning/delta once lazy details arrive.

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
Stash-KennyG: Agreed, incoming.

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
Stash-KennyG: Agreed.

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
Gykes: That's potentially a bug I think. Currently if you disable show tag on hover it would disable for performer as well. We would either need to generalize that config. Something like "ShowPreviewOnHover" or "ShowPopoverOnHover" or create a completely separate option for performers and then add that to this.

To keep scope down you can keep this for now and we can add/fix that in another PR.

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
Gykes: Hybrid approach is fine. Cache should dedupe on ID then reuse cache after the first pull anyway.

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
Stash-KennyG: I just searched it - this is the only area I see that reference to ui.showTagCardOnHover that's not to the tag.
Happy to write it up, after I get this merged.

ui/v2.5/src/components/Tagger/scenes/PerformerResult.tsx
Stash-KennyG: I beleive the new implementation is much better.  I have removed that part as we are lazy loading only on hover/display so the result shouldn't need to pre-process and hold a loading here..

ui/v2.5/src/components/Tagger/scenes/MatchedPerformerPreview.tsx
```diff
@@ -0,0 +1,271 @@
...
+      onOpen={() => setIsOpened(true)}
+      content={
+        <div>
+          {performer ? (
+            <LocalPerformerCard performer={performer} />
+          ) : (
+            <div className="tag-popover-card tagger-matched-performer-popover p-3">
+              {selectedPerformerLoading ? (
+                <FormattedMessage
+                  id="ui.loading.generic"
```
Gykes: This exists as `loading.generic` in the root of `en-gb.json`.

ui/v2.5/src/components/Tagger/styles.scss
```diff
@@ -95,6 +95,49 @@
...
+    width: 200px;
+  }
+
+  .performer-card-image {
+    max-height: 300px;
+  }
+}
+
+.tagger-matched-performer-popover-extra {
+  border-top: 1px solid rgba(255, 255, 255, 0.12);
```
Gykes: I would use the theme color here and not hardcode it. If one doesn't exist for this exact color you could create it there and utilize it instead of hardcoding.

ui/v2.5/src/components/Performers/PerformerPopover.tsx
```diff
@@ -9,10 +9,12 @@ import { Placement } from "react-bootstrap/esm/Overlay";
 
 interface IPeromerPopoverCardProps {
   id: string;
+  cardClassName?: string;
```
Gykes: You added a few of these but I don't see where they are being utilized? I may be blind but if they are redundant then they should be removed.

ui/v2.5/src/components/Tagger/styles.scss
```diff
@@ -95,6 +95,49 @@
...
 
+.tagger-matched-performer-popover {
+  .card {
+    width: 200px;
+  }
+
+  .performer-card-image {
+    max-height: 300px;
+  }
+}
```
Gykes: This has the exact same css as `114-122`. We may be able to condense these down to save some line space.

ui/v2.5/src/components/Tagger/scenes/MatchedPerformerPreview.tsx
```diff
@@ -0,0 +1,271 @@
...
+  performerID,
+  scrapedPerformer,
+  endpoint,
+  placement = "bottom",
+  children,
+}: IMatchedPerformerPreviewProps) => {
+  const intl = useIntl();
+  const { configuration: config } = useConfigurationContext();
+  const showPerformerCardOnHover = config?.ui.showTagCardOnHover ?? true;
+  const [isOpened, setIsOpened] = useState(false);
```
Gykes: So, it seems like this would cause the query to remain open for the components lifetime. I don't think this is a massive issue as I believe Apollo cache would handle the repeat hovers. You could add an `onClose` if `HoverPopover` could support it.

ui/v2.5/src/components/Tagger/scenes/MatchedPerformerPreview.tsx
```diff
@@ -0,0 +1,271 @@
...
+const buildPerformerDeltaRows = (
+  remote: GQL.ScrapedPerformer,
+  local: GQL.PerformerDataFragment,
+  intl: IntlShape
+): IPerformerDeltaRow[] => {
+  const rows: IPerformerDeltaRow[] = [];
+
+  pushDeltaIfDifferent(
+    rows,
+    intl.formatMessage({ id: "birthdate", defaultMessage: "Birthdate" }),
```
Gykes: Do we really need the `defaultMessage`? The keys exist and it just takes up extra space. It's not a merge stopper just a nit pick.

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
```diff
@@ -0,0 +1,113 @@
...
+              <span className="performer-disambiguation">
+                {` (${performer.disambiguation})`}
+              </span>
+            )}
+          </div>
+          {age !== 0 && <div className="performer-card__age">{ageString}</div>}
+        </div>
+      </div>
+    </div>
+  );
```
Gykes: This manually rebuilds the PerformerCard DOM. If the markup changes then this wont change with it. We might be better of breaking it out to a shared component and utilizing that.

Gykes (dismissed):
Another static review. 

Looking better, once these are addressed/chatted about I will do app testing.

ui/v2.5/src/components/Tagger/scenes/MatchedPerformerPreview.tsx
Stash-KennyG: I'm not an expert on internationalization UI.  I did it looking for a safe fallback if the selected UI language doesn't have it...

I took a look at the implementation, I guess since we are merging up from en-gb we would be able to assume that if it exists there, it would always exist at some layer...

I could argue as a best practice it can help with context to prevent the need to swap back and forth to the source json and lookup.  May not be a big deal in the Birthdate sense, but on some edge cases it's less helpful.
I did a google search to support my thesis and it appears there are two distinct schools of thought.  I will not die on this hill, but I can factor to a draft of 
Only include messages when it's not the same as the id.

I.e.
`intl.formatMessage({ id: "birthdate"})`
vs
`intl.formatMessage({ id: "loading.generic", defaultMessage: "Loading..." })`

ui/v2.5/src/components/Performers/PerformerPopover.tsx
Stash-KennyG: This is an artifact from how I was intitially thinking about splitting the card.
I agree, not necessary.

ui/v2.5/src/components/Tagger/scenes/MatchedPerformerPreview.tsx
Stash-KennyG: Great idea.
Now fixed:

Added onClose={() => setIsOpened(false)} to HoverPopover in MatchedPerformerPreview.
Query skip: !performerID || !isOpened will now turn back on/off with hover open/close.

ui/v2.5/src/components/Tagger/scenes/ScrapedPerformerPreview.tsx
Stash-KennyG: I was trying to recycle the performer card, but it's just not perfect for this use case primarily with the different data requirement.  See your previous suggestion.
So based on this I'll move toward a specific component that serves both data methods.

Extending PerformerPopover to support both modes:
* existing id fetch mode
* new previewData mode (typed IPerformerPreviewData)
* Added optional loading, loadingText, cardExtras, onOpen, onClose props so Tagger can use lazy-loading +  extra mismatch/warning content without custom hover plumbing.
* Added shared renderer:
  - components/Performers/PerformerPreviewCard.tsx
* Migrated Tagger hover usage to PerformerPopover:
* ScrapedPerformerPreview now uses PerformerPopover with previewData
* MatchedPerformerPreview now uses PerformerPopover with previewData, loading state, and cardExtras

I like this implementation much more. Should remain consistent.
Moved CSS into the new target performer component css.

Stash-KennyG:
## What Changed Since PR Opened

- Refactored hover rendering so `PerformerPopover` is now the shared hover controller for this feature path.
- Added `previewData` support to `PerformerPopover` to safely render scraped performer previews without unsafe casting.
- Added `PerformerPreviewCard` in `components/Performers` as a shared typed preview renderer.
- Moved shared preview-popover styling ownership to `components/Performers/styles.scss` (`.performer-preview-popover`) and removed Tagger-owned shared styling.
- Preserved full `PerformerCard` behavior for ID-based popovers (favorite icon, counters/chips, overlays) by restoring `PerformerCard` in `PerformerPopover` local-ID mode.
- Kept Tagger-specific extras (stash ID conflict chip + mismatch rows) as appended card content inside matched popovers.

## Data Shape Strategy (No Unsafe Casts)

- `PerformerDataFragment` and `ScrapedPerformer` are intentionally handled through separate adapters.
- Scraped/remote preview uses typed `previewData` mapping.
- Local/matched preview uses `id` mode in `PerformerPopover` plus lazy details fetch for mismatch/warning logic.
- No `as unknown as` style casting is used to coerce incompatible performer types.

## Hover/Lazy-Load Behavior

- Matched performer detail query is hover-scoped:
  - Opens on hover (`onOpen`)
  - Stops querying on close (`onClose`)
- Prevents eager fetching for large scrape result pages while preserving content.

## UI/Styling Adjustments

- Consolidated duplicate preview card CSS.
- Updated border color to theme token (`$secondary`) for consistency with dark theme styling.

## CI/Validation Fixes Applied

- Fixed stylelint property ordering in `Tagger/styles.scss`.
- Applied Prettier formatting fixes for updated files.
- Confirmed latest PR checks pass after these adjustments.

ui/v2.5/src/components/Tagger/scenes/MatchedPerformerPreview.tsx
Gykes: It's not a hill I will die on either. The main dev can make that call. There's no negative to having it other than linespace.

Gykes:
So, after some quick testing of the feature it looks okay. There is a small UX thing that I noticed and not sure if it is intended. 

1) Search for a scene that has greater than 1 performer
2) Hover the first performer
3) The new modal is considered part of the hover and does not go away if you are hovering the name or the new modal
4) To hover the next performer you need to move outside of the box and it feels a little clunky.

From a UX standpoint I would image that, since there is no hyperlinked information in this modal, then it should go away when you remove your hover from the name of the performer on the tagger card. This is, of course, opinionated and I think no matter what way you do it you will get pushback. 

In my opinion, the hover should disappear when you move your mouse off of the performer name. With that in mind lets go ahead and go that route and if there is a large enough of a pushback then we can revert.

Stash-KennyG:
Current behavior is intentional. When you hover a performer and the modal appears, it remains active if the cursor moves into the modal. That allows click or ctrl/middle-click actions directly from the modal (open stash or stashdb).

The primary driver was the right side where that interaction did not exist before. The left side already supported it, but I kept behavior consistent across both sides rather than introducing different hover models.

Implementation differs slightly (inline block vs box picker), but both treat the modal as part of the hover region by design.

Net effect: dismissal is lateral movement, not vertical.

The suggestion is valid from a simplicity standpoint. Current design prioritizes interaction over quick dismissal. If that tradeoff proves wrong, it can be adjusted, but would lose the ability to interact with the card in case someone wanted to get more details.  Also, in highlighting performer mis-matches - interaction with middle click may be necssary...

Gykes:
If it is intentional then you can keep it as is. I imagine that people will want to be directed to the Stash Box link to verify performer at some point. Off the top of my head I don't have a solution for the issue I was having but I don't think it's a massive one. I think this is mostly ready for WP review anyway so will probably give it another quick look over and test then send it his way,

Gykes:
Coming back to this. I know some of the UI aspects were being questioned. Mostly I think it was the size of the new card. Was there any decision on this?

Stash-KennyG:
I have not seen anyone else comment on this.  Overall, this is the same size as the existing cards.  I'm all for size optimization, and would be happy to engage with any constuctive feedback but the current layout is intended to provide manual duplicate/match optimization.  There is a component of this where the size/detail kind of is the point.  The image generally matches the width of the underlying hover control and height flows accordingly.  If it scrolls vertically beyond that, I don't see this as a major negative issue.  Further this pulls from our existing card code-base so it cuts down on stylistic/manual maintenance of duplicate code.  Happy to take any further edits onboard.  Overall I'm very eager to get this into the next release.

ui/v2.5/src/components/Performers/PerformerPopover.tsx
```diff
@@ -8,15 +8,20 @@ import { useConfigurationContext } from "../../hooks/Config";
 import { Placement } from "react-bootstrap/esm/Overlay";
 
 interface IPeromerPopoverCardProps {
-  id: string;
+  id?: string;
+  cardContent?: React.ReactNode;
+  loading?: boolean;
+  loadingText?: string;
```
Gykes: `loading` and `loadingText` get added but I don't see them wired up.  Let me know if i'm missing it.

ui/v2.5/src/components/Performers/PerformerPopover.tsx
```diff
@@ -29,25 +34,77 @@ export const PerformerPopoverCard: React.FC<IPeromerPopoverCardProps> = ({
...
+  if (loading) {
+    return (
+      <>
+        <div className="tag-popover-card performer-preview-popover p-3">
+          {loadingText}
+        </div>
+        {cardExtras}
+      </>
+    );
+  }
```
Gykes: Is this branch reachable?

ui/v2.5/src/components/Tagger/scenes/TaggerPerformerPopover.tsx
```diff
@@ -0,0 +1,304 @@
...
+      cardExtras={
+        matchExtras || cardExtras ? (
+          <>
+            {matchExtras}
+            {cardExtras}
+          </>
+        ) : null
+      }
+      placement={placement}
+      enterDelay={1000}
```
Gykes: is 1s the standard? I think most other places in the codebase are 500ms. This is more of a design question than anything else.

Gykes (changes requested):
Couple more questions

I did more actual testing and it seemed to work as expected. I think after this it can be pushed up to WP for final review.

WithoutPants:
It's taken a little time to figure out how to make the mismatch information appear, so in future it'd be good to include some rough instructions on how to get specific behaviour to manifest.

For reference, I did the following:
- for a performer with populated fields and a stash id, add an alias for an unrelated performer
- lookup a scene using the tagger where the unrelated performer is in the scene
- the unrelated performer then appears as a match in the performer combobox

There's some usability issues I'd like to see addressed before merging this:
- the hover enter/leave delay has been changed and feels significantly different to the delays used elsewhere. In particularly the enter delay feels far too sluggish. If the intent is to avoid unwanted hovering when mousing over the dropdown list, then I'd suggest using a hoverable icon next to the performer name, but I'm not convinced that this will be necessary.
- the hover popover appears when hovering over the dropdown list, not just the current value. Makes it very difficult to choose a performer from the drop-down list.
- the presentation of the mismatched data needs a little further work. As it is currently, there's zero context of what `StashDB` tag actually means, nor for the mismatched data fields. 

My very rough suggestion would be to remove the mismatch data from the performer popover and just leave it as a quick reference. The mismatch information should be readily visible as soon as a potentially mismatched performer is selected. I don't have a specific design in mind, but some ideas might be:

- show a warning icon at the end of the line if there are mismatched fields - details would be shown by mousing over the icon:
image
- indicate issues with the matched performer below the performer row:
image

Ideally it should show the mismatched data behind the popover. For example, it might say `birthdate doesn't match` with the two values shown in a popover. You'd probably want to show only a short list of issues, with the rest behind a popover.

ui/v2.5/src/components/Performers/PerformerPopover.tsx
Stash-KennyG: This got refactored out from our earlier discussions from the way the query got reworked from your feedback a few months ago.

ui/v2.5/src/components/Performers/PerformerPopover.tsx
Stash-KennyG: No it is no longer - see previous comment - gone.

ui/v2.5/src/components/Tagger/scenes/TaggerPerformerPopover.tsx
Stash-KennyG: You are right.

PerformerSelect, TaggerPerformerPopover (except delays), and anything else using PerformerPopover without overrides → 500 / 100

Most other HoverPopover usages (scene cards, gallery cards, performer count button, etc.) → 200 / 200 

I'll set this at 500/100 but happy to take feedback here.

Stash-KennyG:
I should have some time tomorrow afternoon to look at making those updates - that makes a lot of sense.
I like the warning issues icon - may not be necessary to go beyond that.

It should also allow a cleanup of the PerformerPopover.extras that wouldn't be needed, but would create a PerformerWarningPopover tray, but should be pretty straight forward.  I'll take a look in the bright light of day.

Stash-KennyG:
Ok, @WithoutPants I think this should address your suggestions for improvements.

I have moved the collission warnings out of the popover and into a dedicated anchor.  
image
It opens on hover, or can be clicked for our touch friends.

I have also aligned the popover delay to be consistent with other controls in Stash and hover area on the picker to exclude the down button of the control to prevent it from interfearing with the picker.

WithoutPants:
This is looking better. Is there any reason we'd _want_ to turn off the warning of mismatched performers? It seems like an unnecessary option.

Stash-KennyG:
> This is looking better. Is there any reason we'd _want_ to turn off the warning of mismatched performers? It seems like an unnecessary option.

There was a discussion when working on the layout for the button as some felt it may have gotten in their way.  When we do a scrape it does add some amount of overhead.  If scraping a large volume of scenes with multi-performers per scene, maybe you just want to disable it...  It seemed straight forward enough to implement.  This is not a hill I'm looking to die on.  I can pull out the settings part if that's what we want to do, but I also don't think it materially complicates the mainteance area to have an option given what we already have as options for in the tagger settings.

Stash-KennyG:
* Bump

Any update here? I'd love to get this merged.

WithoutPants (approved):
Removed option as I don't think it adds much value. If it's necessary, we can add it in at a later date.

Issue #6852, opened 2026-04-21
Marker previews silently truncated at 20s; make `maxMarkerPreviewDuration` configurable


speckofthecosmos:
## Problem

When a scene marker is created with `end_seconds > 20`, Stash writes the full `end_seconds` value to the database and displays it in the marker form, but silently truncates the generated preview `.mp4` to 20 seconds. The cap is enforced at generation time by the `maxMarkerPreviewDuration = 20` constant in `pkg/scene/generate/marker_preview.go`. The UI provides no indication that the configured duration is being ignored.

**Repro:**
1. Create a marker with `Time = 10s`, `End Time = 40s` (30s intended duration)
2. Run Generate → Marker Previews
3. Observe: the preview file at `/generated/markers/{oshash}/10.mp4` is 20 seconds long, not 30

This is a least-surprise violation — user input is accepted, partially honored, and the disagreement is never surfaced.

## Impact

Users who treat markers as watchable clips — finding the good parts of a scene, setting end times, and consuming them as a preview wall — cannot get preview files that reflect their curated durations. A workaround community plugins that re-runs FFmpeg post-hoc would be fragile because the batch Generate task overwrites the extended output.

The cap itself exists for real reasons: markers can travel between instances via Stash's Export/Import pipeline and via community marker-sharing tooling, which means any install can receive markers carrying arbitrary `end_seconds` values from sources outside the user's direct control. A duration ceiling on generated previews protects against unintended disk blowup from a bad import. Keeping it configurable (rather than removing it entirely) preserves that safety for cautious users while unblocking curated-collection users who control their own data.

Context on why this is worth fixing now rather than as part of a larger clips feature: Discussion #6851 lays out a mobile-first direction for Stash starting with markers. This fix is the first concrete step in that direction, and it stands on its own merits regardless of how that broader conversation evolves. (An earlier discriminator-level proposal I made at #4594 is superseded in favor of this narrower approach.)

## Proposed fix

Replace the hardcoded `maxMarkerPreviewDuration = 20` constant with a configurable field on `ConfigGeneral`:

- Config key: `MaxMarkerPreviewDuration`
- **Default: `20`** — preserves current behavior exactly for all existing installs
- **Setting `0` disables the cap entirely** (unlimited preview duration) for users whose marker sources are fully trusted. Document this in the settings-panel field description: *"Maximum generated marker preview duration in seconds. Set to 0 for unlimited."*
[...]

Issue #4594, opened 2024-02-19
[Feature] Add ability to define Clips/Sub-Scenes from Scenes


UnwantedBunion:
Hi,

First, thanks to everyone who's worked on Stash, really appreciate your efforts.

I'm aware this is a near-duplicate of a couple of other feature requests, and there's a lot of overlap with several others. Please feel free to tag/move/etc. as appropriate, just thought I'd add my specific use cases.

**Is your feature request related to a problem? Please describe.**
Summary:
Remove unwanted parts/create "best bits" collection.

Detail:
Even for my favourite scenes, there are parts I'm not interested in. Either it's filler, or there are things going on I'd rather not watch. Having to fast-forward through a scene (or manually skip to the next marker) every time is a bit laborious. Being able to define the parts I do want, and automatically play only those, would be great.

Also, it'd be great to be able to put together a selection of "best" clips collection without having to manually process the file with ffmpeg etc.

**Describe the solution you'd like**
I'd propose adding the ability to define parts of Scenes as Clips/Sub-Scenes/whatever's the best name. This would involve being able to define a start and end time for the Clip. I'd imagine adding an end time to Markers would be the cleanest way to do thjs.

If there was a major barrier/opposition to extending Markers to include an end time, adding Clips to Scenes as a separate entity might be the way to go.

**Describe alternatives you've considered**
I've considered using Markers for this purpose, but they don't currently quite do 100% of what I'd like. There's no end time, so I can't define clips of arbitrary length. Sometimes the desired clip might be 5 seconds, sometimes it might be 20 mins.

**Additional context**
If they were to be implemented as a separate entity, I'd ideally like:
- The ability to add tags to a Clip (similar to Markers)
- A tab for Clips on the Scene page (similar to Markers)
- A separate Clips page to show all defined Clips (similar to Markers)

Regardless of implementation, to make this as useful as possible I'd like:
- The ability to, when playing a Scene, to only play the Clips for that scene rather than the whole lot. Have a toggleable setting that makes this the default behaviour. If this could include transitions (e.g. fade in/out) between Clips, that'd be perfect.
- The ability to download Clips as video files.

**Similar feature requests**

It's basically the same as these feature requests, with a bit more context:

[Feature] Optional end for markers to use them as clips
https://github.com/stashapp/stash/issues/3147

[Feature] End time for markers and a possibility to combine those subscenes to playlists
https://github.com/stashapp/stash/issues/2647

Other feature requests I could find that might use similar functionality:

[...]

PR #6855, opened 2026-04-22
Honor marker end times; make the ceiling configurable

speckofthecosmos:
Scene markers gained explicit end times in [PR #5311](https://github.com/stashapp/stash/pull/5311) (November 2024), but `pkg/scene/generate/marker_preview.go` was not updated to consume them. The file's `maxMarkerPreviewDuration = 20` constant predates end times by two years ([PR #2392](https://github.com/stashapp/stash/pull/2392), April 2022). For the 18 months since #5311 shipped, user-set end times beyond 20 seconds have been silently truncated at preview generation: the UI accepts the value, the database stores it, the marker form displays it, but the generated video is capped. The end-time feature has been shipping incomplete.

This PR closes the loop. Preview generation honors the marker's `endSeconds` by default. The 20s ceiling becomes opt-in via a new `MaxMarkerPreviewDuration` setting for users who want it as a safety against imports or data entry mistakes.

**The cost of delay is growing.** Every marker created since November 2024 with `endSeconds > 20` is a marker whose preview file doesn't match its stored metadata. That misalignment compounds daily — users shipping the fix later will have more stale preview files to regenerate than users shipping it now.

## Change

New `Max marker preview duration` setting under Settings → System → Preview Generation.

- **Default 0** — the marker's explicit end time is honored verbatim. This is a behavior change from the previous hardcoded 20s ceiling.
- **Positive N** — sets the ceiling in seconds. Users who import markers from untrusted sources or want defensive size limits can opt in.
- **Non-positive intervals** (`endSeconds <= seconds`) — preview generation is skipped with a warning log, rather than falling back to a default. If the user set zero or negative duration, either it's a data mistake (surface it, don't silently invent output) or it's an intentional point marker (respect the signal, don't generate a video they didn't ask for).
- Markers without an explicit `endSeconds` continue to use a fixed 20s default (deliberately out of scope for this PR — see below).

## Why the default is 0

A buried opt-in setting is the wrong interaction pattern for user-created data. Most users who set a marker end time expect it to be honored; the silent-truncation case penalized the majority for an edge case (imports with unusually long end times). With the flipped default, the ceiling-as-safety remains available via one setting for users who want it, and the common path matches user intent.

The 0-disables convention follows established precedent in `config.go` — `defaultLogFileMaxSize = 0 // megabytes, default disabled`.

## Incidental fix

The new `interval > 0` guard prevents zero/negative-interval markers from producing malformed ffmpeg calls — previously possible if `endSeconds <= seconds`, producing zero or negative duration arguments.

## Behavior change on upgrade

Existing marker preview files have had silently-truncated durations since #5311 shipped 18 months ago. This PR does not change those files retroactively — they remain until regenerated.

To refresh previews for markers whose end times have never been honored, run **Generate → Marker Previews** with **Overwrite Existing** enabled. Note: this regenerates ALL marker previews, not just the affected subset. For large libraries with few affected markers, this is imperfect — a targeted "refresh stale previews" task would be a natural follow-up enhancement and is out of scope here.

To keep the old 20s ceiling, set **Max marker preview duration** to 20 under Settings → System → Preview Generation.

## Closes/Fixes Issues

- Addresses [#6852](https://github.com/stashapp/stash/issues/6852) (closed for template non-conformance; content below)
- Context: [Discussion #6851](https://github.com/stashapp/stash/discussions/6851)
- Supersedes the discriminator-level proposal at [#4594](https://github.com/stashapp/stash/issues/4594)

## Testing QA Notes

- Default 0: marker with `endSeconds=40s, seconds=10s` renders full 30s interval ✓
- Cap set to 20: same marker renders at 20s (truncated at the ceiling) ✓
- Cap set to 60: same marker renders at 30s (within the ceiling) ✓
- Marker with nil `endSeconds`: uses 20s default regardless of cap setting ✓
- Marker with `endSeconds <= seconds`: generation skipped, warning log recorded ✓
- Clean codegen pass (`make generate`); backend and frontend both compile cleanly
- Manually verified end-to-end on a local build (config persistence, UI rendering, ffmpeg output duration matches expected)

## Out of scope

**Making the 20s default (for markers without explicit `endSeconds`) configurable.** Kept separate because the cap change fixes a demonstrable bug (silent truncation of user-provided end times), while making the nil-default configurable is a preference surface that invites distinct design discussion — what should it default to? Same as the cap? A 5–10s teaser? 0 meaning "skip generation for bookmark-style markers"?

Happy to bundle into this PR if reviewers think UI-coherence (changing Settings → System once rather than twice) outweighs narrow-scope. The incremental work is small:

```diff
// internal/manager/config/config.go
+ DefaultMarkerPreviewDuration        = "default_marker_preview_duration"
+ defaultMarkerPreviewDurationDefault = 20
+ func (i *Config) GetDefaultMarkerPreviewDuration() int {
+     return i.getInt(DefaultMarkerPreviewDuration)
+ }

// pkg/scene/generate/marker_preview.go
- const markerPreviewDefaultDuration = 20
+ func (g Generator) MarkerPreviewVideo(..., maxDuration int, defaultDuration int) error {
+     duration := float64(defaultDuration)
```

Plus mirror plumbing: `setConfigInt` in the resolver, field in GraphQL `ConfigGeneralInput`/`Result`, a second `<NumberSetting>` in `SettingsSystemPanel.tsx`, i18n strings for head/desc.

Otherwise, a follow-up PR would cite Discussion #6851 for framing and propose a sharper user story around the nil-default.

---

**Also worth a follow-up PR:** a "Refresh Outdated Marker Previews" maintenance task that detects markers whose existing preview duration doesn't match the expected duration from current config, and regenerates only those. Useful for this PR's migration story, useful anytime config changes affect preview output, and useful independently of this PR.


speckofthecosmos:
For anyone wanting to test this before merge: https://github.com/speckofthecosmos/stash-marker-cap-patch

Public multi-arch image at `ghcr.io/speckofthecosmos/stash-marker-cap-patch:develop` — rebuilds nightly against upstream develop. Compose-based deployment via `docker pull` + config swap.

Verified working on Synology NAS + verified end-to-end: 41 affected markers (with end_seconds beyond 20s) regenerated with correct durations.

ui/v2.5/src/components/Settings/SettingsSystemPanel.tsx
```diff
@@ -425,6 +425,14 @@ export const SettingsConfigurationPanel: React.FC = () => {
             return <></>;
           }}
         />
+
+        <NumberSetting
+          id="max-marker-preview-duration"
+          headingID="config.general.max_marker_preview_duration_head"
+          subHeadingID="config.general.max_marker_preview_duration_desc"
+          value={general.maxMarkerPreviewDuration ?? 20}
```
Gykes: Should the fallback here not be `?? 0` instead of `?? 20`

pkg/scene/generate/marker_preview.go
```diff
@@ -31,11 +35,24 @@ func (g Generator) MarkerPreviewVideo(ctx context.Context, input string, hash st
...
+
+	// Honor the marker's explicit interval when present and positive, capped
+	// by the configured safety ceiling. maxDuration <= 0 disables the ceiling.
+	// Non-positive intervals are treated as "no video wanted" and skipped —
+	// if the user intentionally set end = start they didn't want a preview,
+	// and if it's a data mistake we'd rather surface it than silently default.
+	if endSeconds != nil {
+		interval := *endSeconds - seconds
+		if interval <= 0 {
+			logger.Warnf("[generator] marker at %.2fs has non-positive interval (end=%.2f); skipping video preview generation", seconds, *endSeconds)
```
feederbox826: this should be guarded on frontend
start = end should default to 20s since that's the default behaviour

ui/v2.5/src/locales/en-GB.json
```diff
@@ -418,6 +418,8 @@
       "include_audio_desc": "Includes audio stream when generating previews.",
       "include_audio_head": "Include audio",
       "logging": "Logging",
+      "max_marker_preview_duration_desc": "Optional ceiling (in seconds) for marker preview videos with explicit end times. Default is 0 (no ceiling, the marker's end time is honored verbatim). Set a positive value to cap preview duration as a safety against imports or data entry mistakes. Markers without an end time use a fixed 20-second default, unaffected by this setting.",
```
feederbox826: this is way too verbose

feederbox826 (changes requested):

speckofthecosmos:
Pushed `9f8acb2` addressing both:

- **Fall-back on zero interval:** the renderer now treats `end <= start` the same as a missing end (warning + default duration) instead of returning early. The marker form already enforces `end >= start` and the resolver mirrors it (`validateSceneMarkerEndSeconds`) — both intentionally permit `end == start`, so the generator shouldn't draw a stricter line.
- **Locale copy trimmed** to match the surrounding settings' tone.

speckofthecosmos:
@feederbox826 — pinged for re-review. Both points addressed in `9f8acb2` above.

WeedLordVegeta420:
I'm hoping that this makes it into the next release. Seems like a relatively low impact way to expand Marker capabilities.

I do think one minor improvement would be to separate the default preview length (length when no end time is specified) from the max preview length (max length when end time is specified).

Gykes (changes requested):
Couple of thoughts: 

- Should probably update documentation to reflect this update.

- Should probably add some test cases are there are several branches that can be flowed through

- Defaulting to 0 would probably have existing users workflow changes, no? I'm not sure if this is an issue or not but just something to potentially warn/update existing users about.

WithoutPants:
> Making the 20s default (for markers without explicit endSeconds) configurable. Kept separate because the cap change fixes a demonstrable bug (silent truncation of user-provided end times), while making the nil-default configurable is a preference surface that invites distinct design discussion — what should it default to? Same as the cap? A 5–10s teaser? 0 meaning "skip generation for bookmark-style markers"?

> Happy to bundle into this PR if reviewers think UI-coherence (changing Settings → System once rather than twice) outweighs narrow-scope.

I think it makes more sense to include that in this PR. Making the default 20s maintains legacy behaviour, making it the most obvious choice imo. Setting the default duration <= 0 should be invalid, and generating marker previews where this value is <= 0 should be skipped with a warning log per the other preview generation behaviour.

speckofthecosmos:
Pushed `e0afb66`.

**@WithoutPants** — added the default-duration setting (default 20s; rejects `≤0` and skips with a warning, as you described). Max stays the ceiling (0 = none).

**@Gykes** — added branch tests (extracted the logic into a pure function, 13 cases), updated the Tasks manual + changelog for the behaviour change, and fixed the `?? 0` fallback.

**@feederbox826** — earlier points are in `9f8acb2`. Validation's at the resolver (added a `<0` guard on max too); the generator skip is per @WithoutPants' request above — happy to drop it if you'd prefer.

WithoutPants (approved):
- fixed lint error
- removed changelog entry - these are sorted out by the dev team, and that particular file was for the last release
- minor changes to the styling in the docs

ui/v2.5/src/locales/en-GB.json
DogmaDragon, line 1:
```
"markers_tooltip": "20 second videos which begin at the given timecode."
```
Hardcoded value needs to be updated.

WithoutPants: I'll address this in #7043

discourse-stashapp:

https://discourse.stashapp.cc/t/stash-tv/3627/136

PR #6856, opened 2026-04-22
New: Add File Paths to Image Scraper Input

NativeRavenclaw:
Tools like `gallery-dl --write-metadata` provide additional metadata to the download as separate file, i.e. `/stash/pic.png` gets an accompanying `/stash/pic.png.json`. 

Adding a `Path` field to the scraper args will allow scrapers to find such metadata files. This seems in line with scenes/galleries passing their Filename(s) to their scraper args.

### DIFF
--- a/pkg/models/image.go
+++ b/pkg/models/image.go
@@ -79,7 +79,6 @@ type ImageUpdateInput struct {
 	Urls             []string           `json:"urls"`
 	Date             *string            `json:"date"`
 	Details          *string            `json:"details"`
-	Path             *string            `json:"path"`
 	Photographer     *string            `json:"photographer"`
 	Rating100        *int               `json:"rating100"`
 	Organized        *bool              `json:"organized"`
--- a/pkg/models/image.go
+++ b/pkg/models/image.go
@@ -79,6 +79,7 @@ type ImageUpdateInput struct {
 	Urls             []string           `json:"urls"`
 	Date             *string            `json:"date"`
 	Details          *string            `json:"details"`
+	Path             *string            `json:"path"`
 	Photographer     *string            `json:"photographer"`
 	Rating100        *int               `json:"rating100"`
 	Organized        *bool              `json:"organized"`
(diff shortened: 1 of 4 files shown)


Gykes (commented):
So, i'm a little confused on what you are trying to accomplish here. If you are trying to do what I think you are trying to do them I believe that creating an `imageInput` similar to scene and gallery input with a `Files []fileInput` would be better. Again, without a full unerstanding of your goal then it's had to recommend any alternatives.

NativeRavenclaw:
I'm trying to get a custom python scraper to get the path/filename of an image from `scraper_args()`. Scenes have them - images don't.

```python
def main():
    op, args = scraper_args()
    log.info("op: %s\n" % op)
    log.info("args: %s\n" % args)

    result = None
    match op:
        case "image-by-fragment":
            input_data = args

            # path = input_data['files'][0]['path'] # <--- does not exist
            path = input_data['path']               # <--- exists with this PR
            result = image_by_fragment(path)

        case "scene-by-fragment":
            input_data = args

            path = input_data['files'][0]['path']  # <--- exists
            log.info("path: %s\n" % path)

            result = scene_by_fragment(path)
```

Using `imageInput` looks to be more in line with what the other entities get. I was a bit put off by the "input" naming, but will look into implementing it that way.

NativeRavenclaw:
I think that's better? Running it through a scraper through the UI, the `files` array now exists like for scene/gallery. I couldn't find any tests for this to adjust - tell me if I missed them.

Gykes (changes requested):
The code LGTM.

You are correct about no testing to update and I would like to change that. Could you please create a test for `imageToUpdateInput` with a `Files []fileInput` array. This can help protect against future regressions.

NativeRavenclaw:
`imageToUpdateInput` is now `imageInputFromImage` for symmetry with the other entities. I've added a test for that.

WithoutPants (approved):
Code looks good and tested ok.

Issue #3953, opened 2023-07-27
[Bug Report] No Title? then No scrape at all...


scruffynerf:
Initially discovered by trying to build a 'tag only' scraper, that just wouldn't work at all unless Title was returned with a non-empty value.  Then it works as expected.

**To Reproduce**
Confirmed by WP, using the enclosed yml file

Per discussion in Discord:

Me: using this yml in a fresh install of dev, no other scrapers installed (so it's not conflicting with ADE scraper, which it would right now)

WithoutPants:
I've reproduced. The parse error occurs because the Title field is present, and the code checks for the empty value of fixed - which is "", in which case it assumes there's a selector field - which is bad logic.

Removing the title field altogether ought to have the same effect, but like you say it's returning nothing. I'm debugging to find out why.

yep, it's a corner case. 

So: the xpath/json configs are split into scene fields (name, details etc) and relationships (performers, tags, etc). If no scene fields are populated then the code interprets that as no results, and the code that processes the relationships part never gets called.

So obviously a bug. For your use case, I can't think of a workaround - the only workaround there is is to set some scene field to something.

I was able to get the server to return results by adding a dummy Foo field in the yml, but the UI must have some similar logic, since it's still not displaying the scrape dialog. 

If you raise a bug I can try to get it into the current release.

**Expected behavior**
A scrape with any result should be returned in the UI with those results, even if other fields are empty.

**Stash Version: (from Settings -> About):**
v0.21.0-45-gb3fa3c32

[AdultEmpireSceneTags.yml.txt](https://github.com/stashapp/stash/files/12178223/AdultEmpireSceneTags.yml.txt)

PR #6857, opened 2026-04-22
Fix: Prevent Nil Pointer Dereference in scrapeScene

### LINKED ISSUE #6921: Panic: nil pointer in mappedScraper.processSceneRelationships when scene scrape returns zero results

### Describe the bug

Identify / scene scraping can panic with runtime error: invalid memory address or nil pointer dereference when a JSON (or mapped) scene scraper produces zero processed scene rows, but the scraper config still defines scene-level relationships (performers, tags, studio, movies, groups).

```
panic: runtime error: invalid memory address or nil pointer dereference
github.com/stashapp/stash/pkg/scraper.mappedScraper.processSceneRelationships(...)
	github.com/stashapp/stash/pkg/scraper/mapped.go:89
github.com/stashapp/stash/pkg/scraper.mappedScraper.scrapeScene(...)
	github.com/stashapp/stash/pkg/scraper/mapped.go:191
github.com/stashapp/stash/pkg/scraper.(*jsonFragmentScraper).scrapeSceneByScene(...)
	github.com/stashapp/stash/pkg/scraper/json.go:192
```

(Exact line numbers may vary slightly by release; the failure is a nil dereference inside processSceneRelationships when writing to ret.)

### Steps to reproduce

1. Use a mapped JSON scene scraper whose scene block includes relationship mappings (e.g. Performers, Tags, Studio).
2. Run a code path that calls mappedScraper.scrapeScene (e.g. Identify with scene by fragment / phash) such that sceneMap.process returns an empty results slice (API returns no matching video, wrong JSON shape, empty list, etc.).
3. Observe the job panic instead of returning “no scrape result”.

### Expected behaviour

When there are no mapped scene rows, Stash should either:
- skip relationship processing safely, or
- allocate an empty *models.ScrapedScene before applying relationships,
 
and the identify/scrape job should not panic.

## Actual behavior

`scrapeScene` leaves `ret` as `nil` when `len(results) == 0`, but still calls `processSceneRelationships(ctx, q, 0, ret)`, which assigns to fields on `ret` (e.g. `ret.Performers`), causing a **nil pointer dereference**.

Relevant logic (conceptually):

```go
var ret *models.ScrapedScene
if len(results) > 0 {
    ret = results[0].scrapedScene()
}
hasRelationships := s.processSceneRelationships(ctx, q, 0, ret) // ret may be nil
```

## Suggested fix

Before calling `processSceneRelationships`, ensure `ret` is non-nil when relationship maps may run, e.g.:

- `if ret == nil { ret = &models.ScrapedScene{} }` when `len(results) == 0` but `s.Scene` has any relationship config that `processSceneRelationships` will touch, **or**
- early-return from `processSceneRelationships` when `ret == nil` and no work is needed, **or**
- only call `processSceneRelationships` when `ret != nil` or after allocating a placeholder scene.

Align behavior with the comment in `scrapeScene` about returning when there are results **or** relationships—today the “relationships only” path cannot run safely without a non-nil `ret`.

### Stash version

v0.31.1

### Device details

Docker

### Relevant log output

```shell
26-05-12 14:25:31
Error   
goroutine 250603 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:26 +0x64
github.com/stashapp/stash/pkg/job.(*Manager).executeJob.func1()
	github.com/stashapp/stash/pkg/job/manager.go:200 +0x138
panic({0xffff8a94d960?, 0xffff8b822f00?})
	runtime/panic.go:792 +0x124
github.com/stashapp/stash/pkg/scraper.mappedScraper.processSceneRelationships({0x0, 0x4000ef2400, 0x0, 0x0, 0x0, 0x0, 0x0}, {0xffff8ab9aa90, 0x4000d2e390}, {0xffff8ab9c780, ...}, ...)
	github.com/stashapp/stash/pkg/scraper/mapped.go:89 +0x9c
github.com/stashapp/stash/pkg/scraper.mappedScraper.scrapeScene({0x0, 0x4000ef2400, 0x0, 0x0, 0x0, 0x0, 0x0}, {0xffff8ab9aa90, 0x4000d2e390}, {0xffff8ab9c780, ...})
	github.com/stashapp/stash/pkg/scraper/mapped.go:191 +0x134
github.com/stashapp/stash/pkg/scraper.(*jsonFragmentScraper).scrapeSceneByScene(0x4000e63a00, {0xffff8ab9aa90, 0x4000d2e390}, 0xffff8ab9b350?)
	github.com/stashapp/stash/pkg/scraper/json.go:192 +0x168
github.com/stashapp/stash/pkg/scraper.definedScraper.viaScene({{{0x4000d168a9, 0x8}, {0x4000d16880, 0x35}, {0x4000a46988, 0x8}, 0x0, 0x0, {0x0, 0x0, ...}, ...}, ...}, ...)
	github.com/stashapp/stash/pkg/scraper/defined_scraper.go:70 +0x4b0
github.com/stashapp/stash/pkg/scraper.Cache.ScrapeID({0x400020e780, 0x400145a600, {0xffff8ab9fc90, 0x40003d1680}, {{0xffff8ab9c4c0, 0x4000461440}, {0xffff3adbe650, 0x4000404580}, {0xffff3adbe690, 0x40004613c0}, ...}}, ...)
	github.com/stashapp/stash/pkg/scraper/cache.go:360 +0x6bc
github.com/stashapp/stash/internal/manager.scraperSource.ScrapeScenes({0x40000fc5a0?, {0x4000c872b0?, 0x40015bf338?}}, {0xffff8ab9aa90?, 0x4000d2e390?}, 0xffff8878fb74?)
	github.com/stashapp/stash/internal/manager/task_identify.go:313 +0x90
github.com/stashapp/stash/internal/identify.(*SceneIdentifier).scrapeScene(0x4001095688, {0xffff8ab9aa90, 0x4000d2e390}, 0x4000b0cea0)
	github.com/stashapp/stash/internal/identify/identify.go:106 +0xa0
github.com/stashapp/stash/internal/identify.(*SceneIdentifier).Identify(0x4001095688, {0xffff8ab9aa90, 0x4000d2e390}, 0x4000b0cea0)
	github.com/stashapp/stash/internal/identify/identify.go:61 +0x34
github.com/stashapp/stash/internal/manager.(*IdentifyJob).identifyScene.func1()
	github.com/stashapp/stash/internal/manager/task_identify.go:153 +0x1b8
github.com/stashapp/stash/pkg/job.(*Progress).ExecuteTask(0x40011b40f0, {0x4000dd56e0, 0x55}, 0x40015bf960)
	github.com/stashapp/stash/pkg/job/progress.go:175 +0xb0
github.com/stashapp/stash/internal/manager.(*IdentifyJob).identifyScene(0x40010f1f00, {0xffff8ab9aa90, 0x4000d2e390}, 0x4000b0cea0, {0x400212e2c0, 0x1, 0x1})
	github.com/stashapp/stash/internal/manager/task_identify.go:139 +0xf4
github.com/stashapp/stash/internal/manager.(*IdentifyJob).identifyAllScenes.func1(0x4000b0cea0)
	github.com/stashapp/stash/internal/manager/task_identify.go:128 +0x9c
github.com/stashapp/stash/pkg/scene.BatchProcess({0xffff8ab9aa90, 0x4000d2e390}, {0xffff3adf9080, 0x4000404580}, 0x40023f8800, 0x40023f8800?, 0x4001095bf0)
	github.com/stashapp/stash/pkg/scene/query.go:80 +0x1dc
github.com/stashapp/stash/internal/manager.(*IdentifyJob).identifyAllScenes(0x40010f1f00, {0xffff8ab9aa90, 0x4000d2e390}, {0x400212e2c0, 0x1, 0x1})
	github.com/stashapp/stash/internal/manager/task_identify.go:123 +0x254
github.com/stashapp/stash/internal/manager.(*IdentifyJob).Execute.func1({0xffff8ab9aa90?, 0x4000d2e390?})
	github.com/stashapp/stash/internal/manager/task_identify.go:58 +0xdc
github.com/stashapp/stash/pkg/txn.WithDatabase({0xffff8ab9aac8?, 0x40011b4050?}, {0xffff393531b8?, 0x4000461440?}, 0x40015bfe80)
	github.com/stashapp/stash/pkg/txn/transaction.go:135 +0x64
github.com/stashapp/stash/pkg/models.(*Repository).WithDB(...)
	github.com/stashapp/stash/pkg/models/repository.go:41
github.com/stashapp/stash/internal/manager.(*IdentifyJob).Execute(0x40010f1f00, {0xffff8ab9aac8, 0x40011b4050}, 0x40011b40f0)
	github.com/stashapp/stash/internal/manager/task_identify.go:56 +0x138
github.com/stashapp/stash/pkg/job.(*Manager).executeJob(0x40002b6150, {0xffff8ab9aac8, 0x40011b4050}, 0x4001222780, 0x40008785d0?)
	github.com/stashapp/stash/pkg/job/manager.go:209 +0x130
created by github.com/stashapp/stash/pkg/job.(*Manager).dispatch in goroutine 20
	github.com/stashapp/stash/pkg/job/manager.go:186 +0x1a8
2026-05-12 14:25:31
Error   
panic in job 8 - Identifying...: runtime error: invalid memory address or nil pointer dereference
```

dutchdevil-83:
## Description

`scrapeScene` declares `ret` as a nil `*models.ScrapedScene` and only initializes it when results are found. It then unconditionally calls `processSceneRelationships`, which dereferences `ret` at line 89:

```go
ret.Performers = s.processPerformers(ctx, scenePerformersMap, q)
```

This causes a panic when a scraper returns no direct scene fields but the function continues to process relationships.

## Root cause

```go
// before
var ret *models.ScrapedScene        // nil
if len(results) > 0 {
    ret = results[0].scrapedScene() // only set when results exist
}
hasRelationships := s.processSceneRelationships(ctx, q, 0, ret) // panics if ret == nil
```

## Fix

Initialize `ret` to an empty `&models.ScrapedScene{}` so `processSceneRelationships` always receives a valid pointer:

```go
// after
ret := &models.ScrapedScene{}
if len(results) > 0 {
    ret = results[0].scrapedScene()
}
hasRelationships := s.processSceneRelationships(ctx, q, 0, ret)
```

This also preserves the intent of #3953: when a scraper yields no non-relationship fields but does yield relationships, the function now correctly returns a scene populated with only those relationships instead of panicking.

## Panic trace

```
goroutine 124232 [running]:
github.com/stashapp/stash/pkg/scraper.mappedScraper.processSceneRelationships(...)
    github.com/stashapp/stash/pkg/scraper/mapped.go:89 +0xd9
github.com/stashapp/stash/pkg/scraper.mappedScraper.scrapeScene(...)
    github.com/stashapp/stash/pkg/scraper/mapped.go:191 +0x19b
...
panic in job 2 - Identifying...: runtime error: invalid memory address or nil pointer dereference
```

Closes https://github.com/stashapp/stash/issues/6921


Gykes (changes requested):
I would also recommend adding in some kind of regression test for the relationships-only scrape path.

Something that does the following:

  1. configures a scene scraper with empty mappedConfig results but a populated Performers/Tags block                                                          
  2. asserts the function does not panic                                                                                                          
  3. asserts the returned *ScrapedScene carries the relationships

dutchdevil-83:
Agreed — this path is worth covering. The production fix prevents the nil dereference, but a regression test should verify that a scene scraper with no direct scene-field results still returns a non-nil ScrapedScene populated with relationship data. I’d add a test in pkg/scraper/xpath_test.go that clears the scene mappedConfig, keeps the Performers/Tags mappings, asserts scrapeScene does not panic, and verifies the returned performers/tags.

dutchdevil-83:
Friendly ping: this fixes a nil pointer issue in scrapeScene. 
Let me know if anything needs changes 🙂

pkg/scraper/xpath_test.go
```diff
@@ -683,74 +683,65 @@ func verifyPerformers(t *testing.T, expectedNames []string, expectedURLs []strin
...
-	}
-
-	verifyPerformers(t, expectedPerformerNames, expectedPerformerURLs, scene.Performers)
-
-	const expectedStudioName = "Sis Loves Me"
-	const expectedStudioURL = "/channels/sis-loves-me"
-
-	verifyField(t, expectedStudioName, &scene.Studio.Name, "Studio.Name")
-	verifyField(t, expectedStudioURL, scene.Studio.URL, "Studio.URL")
+func TestApplySceneXPathConfigRelationshipsOnly(t *testing.T) {
```
Gykes, line 756: I think we should keep the original test and add a new one for your addition.

pkg/scraper/mapped.go
```diff
@@ -184,9 +184,14 @@ func (s mappedScraper) scrapeScene(ctx context.Context, q mappedQuery) (*models.
...
 	results := sceneMap.process(ctx, q, s.Common, urlsIsMulti)
 
+	// Ensure ret is non-nil before calling processSceneRelationships so it can
+	// safely populate relationship fields even when no direct results were found.
+	// This preserves the intent of #3953: returning a scene with only relationships.
 	var ret *models.ScrapedScene
 	if len(results) > 0 {
 		ret = results[0].scrapedScene()
+	} else {
+		ret = &models.ScrapedScene{}
```
Gykes: In the PR description this code block seems to be different. I prefer the version in the description. Was there a particular reason to swap it to this? Seems like the same thing just more verboose.

Gykes (changes requested):
Static review only, no testing done. 

Sorry, been AFK for a little bit now and just starting to catch back up.

I think `gofmt` needs to be ran. Seems like you use spaces rather than tabs and tests might fail, not 100%.

pkg/scraper/xpath_test.go
dutchdevil-83: Good call — I’ll keep the original scene test and add a separate regression test for the relationships-only path

pkg/scraper/mapped.go
dutchdevil-83: The only reason I changed it was to avoid the unnecessary allocation on the normal len(results) > 0 path. Behavior-wise it should be equivalent — both versions ensure processSceneRelationships gets a non-nil scene when there are relationship-only results. But I agree the version in the PR description reads better, so I can switch back.

dutchdevil-83:
@Gykes I pushed the requested updates:

restored the original scene XPath test and added a separate relationships-only regression test
switched scrapeScene back to the shorter initialization style from the PR description
ran gofmt
When you have a chance, could you take another look?

Gykes (approved):
LGTM

Testing Done: 
Ran new test to get expected pass.
Reverted the `mapped.go` fix but kept the test
Ran the test to expect a `invalid memory address or nil pointer dereference` error
Test failed as expected. 

Next step would be to create a dummy scraper and point it at a DB but I don't think that's needed tbh.

PR #6865, opened 2026-04-26
[Feature] Cover Image Usability Improvements

### LINKED ISSUE #6861: [Feature] Cover Image usability improvements + metadata
I wasn't sure if it would clutter the issue list if I broke these down, so I put all my thoughts on this topic in one issue. Sorry if that makes it more difficult to track the individual suggestions.

Generating custom cover images for scenes is awkward at the moment.

1) It seems to make a lot more sense to move the "Generate thumbnail from current" and "Generate default thumbnail" actions from the extraneous menu to the Cover Image section of the pane.

2) The Cover Image preview doesn't update when generating from the current frame or default thumbnail until you hit save.

3) One day I hope we can have frame forward and back controls, but for now the least the scene player could do is not resume playback when you seek paused video.

On to feature suggestions:

4) Does stash have a way to determine if a scene has a custom cover image or not? If a user is invested in making custom covers, they need a way to filter out scenes they have already done. I think a "custom_cover_image" variable in the "Is Missing" filter is the perfect place for it. This would apply for both frame-generated covers and covers from an image file.

5) When a user generates a cover image from a video frame, does stash remember the frame number in metadata? I couldn't find it. I think it could pay off later to start doing this now. One day if we have a realized scene database, users could contribute their custom cover frame numbers. Could these be heatmapped and auto-suggested when users fingerprint/request metadata for scenes? 
	
Concept images:

thumbnail generation custom_cover_image

Stash-KennyG:
Resolves #6861 

## Summary

- Improves Scene cover image workflows in the edit UI by consolidating actions into one grouped menu and keeping source metadata in sync.
- Creates a new `cover_image_source` tracking across UI and API, including fallback behavior for legacy/external callers.
- Adds File Info presentation for `cover_image_source` with readable, icon-based output for known source types.
- Generate flows refresh edit preview image without requiring full page reload.

## Why This Change

- Cover image actions were split between multiple controls and were harder to discover/use consistently.
- Interface had grown to set cover images through multiple paths; UI layout was inconsistent. 
- Source provenance is valuable for debugging and eventual scene handling through dupes/merges/etc.
- Existing external tools (userscripts/plugins) may update cover image without explicitly setting source metadata.

## User-Facing Improvements

- Scene edit panel now uses a single **Set image** menu near the preview window with grouped actions:
image
- FileInfo pane now includes cover image source data with icons consistent with the UI elements that generated them to help users connect behaviors.
**Null**
image
**Default**
image
**Clipboard**
image
**Userscript**
image
**URL**
image
**Stash**
image
**Timestamp**
image
**File**
image

## API / Data Contract
- `Scene.cover_image_source` is a nullable string field with semantic contract:
  - `NULL` means unknown / legacy / no cover.
  - Non-null values are provenance tokens describing the source of the current cover image.
- SQLite enforces allowed values with a CHECK constraint:
  - fixed tokens: `default`, `clipboard`, `userscript`
  - prefixed tokens: `url:*`, `stash:*`, `timestamp:*`, `file:*`
- GraphQL exposure:
  - field is available on `Scene`
  - writable via `SceneCreateInput` and `SceneUpdateInput`
- Server defaulting behavior for compatibility:
  - If a mutation provides `cover_image` but omits `cover_image_source`, source defaults to `userscript`.
  - Explicit `cover_image_source: null` is preserved as null (intentional clear).
  - Explicit non-null source is preserved as provided (after normalization/trim).
  - 
## Backwards Compatibility
- Legacy scenes with `NULL` source remain valid and do not display in UI  (no forced migration rewrite).
- Existing external clients that only set `cover_image` now get deterministic source (`userscript`) without requiring client changes.
- Existing clients that explicitly send `cover_image_source` continue to control source behavior.

## UI enhancement
- Added reusable `IconField` abstraction in `src/utils/field.tsx`.
- Extends the existing behavior to allow consistent rendering of a UI element with Stash token or UI icon + string/url.

### LLM Notes ###
I have done a first review of this and feel I've made some progress form feedback of others - thank you.
This was made with the help of [Cursor](cursor.ai)


Gykes (commented):
So I haven't actually reviewed the code but here are my 2 thoughts: 

1) The set image button should probably be changed to a split button similar to what was done [here](https://github.com/stashapp/stash/pull/6438)

2) The file info pane changes seem.... excessive. Is knowing that you pasted your image from clipboard really needed? I feel like most of those options are somewhat useless. This seems to just take up more screen space for not total gain. This is an opinion but that feature might need more discussion before we go ahead with it.

Stash-KennyG:
I like the button idea.
Happy to look at that.  Before I do, I will defer to the group whether we want to even keep,display the information.

I grabbed this off the roadmap as it was labeled 'ready' so I assuemd we were ready to move forward.

Overall, I can make the argument that being able to filter on 'custom/not-custom' has some value but I took the request from @feederbox826 on #6861 for feature 4 and 5 and just ran with them.  But, I'm also not going to war over this hill....

Once there is clear guidance from the group,  I'm happy to take whatever direction we want.

Gykes:
> I grabbed this off the roadmap as it was labeled 'ready' so I assuemd we were ready to move forward. 

It is ready but discussions still need to happen. WP/the community are always available to overturn me. I just personally feel like most of the options listed is very low value information and just increases core code requirements. 

> Overall, I can make the argument that being able to filter on 'custom/not-custom' has some value but I took the request from @feederbox826 on https://github.com/stashapp/stash/issues/6861 for feature 4 and 5 and just ran with them. But, I'm also not going to war over this hill....

I don't disagree just I'm not a huge fan of the current implementation.

WithoutPants:
The cover source stuff at the very least needs to be assessed separately in a different PR. As it is, you've got a fairly non-controversial UI change mixed in with something far more controversial that includes a schema change.

Stash-KennyG:
Based on this, it makes most sense for me to close this and cherry pick the non-controversial edits into a new PR.

PR #6867, opened 2026-04-26
[Feature] Add configurable default performer gender 

### LINKED ISSUE #3751: [Feature] Default gender of performer
As a gay man, gay porn actors always import as women which I just find mildly irritating.

A setting to default the gender thumbnail for performers 

Batch changing the gender flag on the performers page, but it is a pain compared to just having it done automagically.

**Additional context**

Stash-KennyG:
Fixes #3751 
## Summary
- add a new interface setting `defaultPerformerGender` (None + all gender enum values)
- apply the setting as a fallback when creating performers without explicit gender
- preselect default gender in new performer form, and apply it to non-stash scraper results when scraper gender is missing
- keep explicit user/scraper gender precedence and avoid mutating existing performers
## Backend
- extend config GraphQL schema for interface input/result
- add config key/getter and wire configure/query plumbing
- apply fallback in performer create path
- add tests for explicit-vs-default precedence helper behavior
## UI
- add setting in Interface > Performer List
- reuse shared gender list source
- fix locale key usage so dropdown shows localized labels (not raw keys)
- add English locale strings and changelog entry

image

### DIFF
--- a/internal/api/resolver_mutation_configure.go
+++ b/internal/api/resolver_mutation_configure.go
@@ -528,6 +528,7 @@ func (r *mutationResolver) ConfigureInterface(ctx context.Context, input ConfigI
 	r.setConfigBool(config.CustomLocalesEnabled, input.CustomLocalesEnabled)
 
 	r.setConfigBool(config.DisableCustomizations, input.DisableCustomizations)
+	r.setConfigString(config.DefaultPerformerGender, (*string)(input.DefaultPerformerGender))
 
 	if input.DisableDropdownCreate != nil {
 		ddc := input.DisableDropdownCreate
--- a/internal/api/resolver_mutation_configure.go
+++ b/internal/api/resolver_mutation_configure.go
@@ -64,8 +64,8 @@ func (r *mutationResolver) setConfigString(key string, value *string) {
 
 // applyDefaultPerformerGenderInput updates or clears DefaultPerformerGender.
 // Omit both fields to leave the stored value unchanged.
-func (r *mutationResolver) applyDefaultPerformerGenderInput(value *models.GenderEnum, clear *bool) error {
-	if clear != nil && *clear {
+func (r *mutationResolver) applyDefaultPerformerGenderInput(value *models.GenderEnum, shouldClear *bool) error {
+	if shouldClear != nil && *shouldClear {
 		if value != nil {
 			return fmt.Errorf("cannot set and clear default performer gender in the same request")
 		}
--- a/internal/api/resolver_mutation_configure.go
+++ b/internal/api/resolver_mutation_configure.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"strconv"
-	"strings"
 
 	"github.com/stashapp/stash/internal/manager"
 	"github.com/stashapp/stash/internal/manager/config"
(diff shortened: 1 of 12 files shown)


internal/api/resolver_mutation_configure.go
```diff
@@ -528,6 +528,7 @@ func (r *mutationResolver) ConfigureInterface(ctx context.Context, input ConfigI
 	r.setConfigBool(config.CustomLocalesEnabled, input.CustomLocalesEnabled)
 
 	r.setConfigBool(config.DisableCustomizations, input.DisableCustomizations)
+	r.setConfigString(config.DefaultPerformerGender, (*string)(input.DefaultPerformerGender))
```
Gykes: `setConfigStrong` is a no op when the input pointer is nul. When the UI sends`null`, for when the user sets the setting back to no default, the old setting is saved. 

Was able to confirm this was testing. Selecting "none" in the dropdown keeps the previous setting saved and I think this is the issue. 

Building a helper or having the UI send "" instead of null could be possible solutions.

internal/manager/config/config.go
```diff
@@ -1311,6 +1312,15 @@ func (i *Config) GetShowStudioAsText() bool {
...
 
+func (i *Config) GetDefaultPerformerGender() *models.GenderEnum {
+	g := models.GenderEnum(i.getString(DefaultPerformerGender))
+	if !g.IsValid() {
+		return nil
+	}
+
+	return &g
+}
+
```
Gykes, line 1329: Lets log a warning here on a failure rather than just silently ignoring it.

ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
```diff
@@ -424,16 +426,23 @@ export const PerformerEditPanel: React.FC<IPerformerDetails> = ({
...
       if (!result?.data?.scrapeSinglePerformer?.length) return;
+      const withDefaultGender = (
+        scrapedPerformerData: GQL.ScrapedPerformerDataFragment
+      ) =>
+        !scrapedPerformerData.gender && defaultPerformerGender
+          ? { ...scrapedPerformerData, gender: defaultPerformerGender }
+          : scrapedPerformerData;
+      const scrapedResult = withDefaultGender(
+        result.data.scrapeSinglePerformer[0]
+      );
```
Gykes: The default merge logic seems to be duplicated. Breaking it out to a helper might be better to reduce complexity.

Gykes (changes requested):
Just a quick overview. DogmaDragon will more than likely request that you update the manual to reflect this new setting so you may want to get ahead of it and add that too.

internal/api/resolver_mutation_configure.go
Stash-KennyG: Great note.  Refactor addressed in [17fe7b1](https://github.com/stashapp/stash/pull/6867/commits/17fe7b1092a7a9aec9ad15cb18dbca9849d4d7ae)

internal/manager/config/config.go
Stash-KennyG: [18eedfc](https://github.com/stashapp/stash/pull/6867/commits/18eedfc724acb1270358bbffaed976ec7018ecad)

ui/v2.5/src/components/Performers/PerformerDetails/PerformerEditPanel.tsx
Stash-KennyG: Broken out now.

WithoutPants:
I think at least as a first step, this should be a field in `UIConfig` which the UI uses to fill in the default performer gender. Having it in the backend means that it could conceivably be applied to the scrape and identify operations, which is a can of worms I'd like to avoid in the shorter term.

Stash-KennyG:
Closing to maybe refactor consistent with changes.

Issue #6088, opened 2025-09-15
[meta] O-Counter feature requests


feederbox826:
## Aggregation
- #5024

## New Counters
- #2836 
- #4680
- #4853

## UI / UX
- #5193
- #3855

Issue #2836, opened 2022-08-18
[Feature] Add O-Counter for Galleries and allow Sorting Galleries by O-Counter


Erovall:
**Is your feature request related to a problem? Please describe.**
Currently there is no way to add O-Counter for Galleries. O-Counter is added directly for Images, which does not achieve the same goal.

**Describe the solution you'd like**
 It will be very convenient to have an O-Counter for Galleries. Also add the ability to sort galleries by O-Counter.

feederbox826:
related: #5364 #4853

not closing as they all want very specific parts of gallery o-counters

PR #6878, opened 2026-04-30
Expose and filter gallery o_counter (image o_counter sum)

slick-daddy:
Adds support for summing up image `o_counter` values across a gallery and lets you filter galleries based on that total.
Related #2836 and #6088

## Description

- Added an `o_counter` field to the Gallery GraphQL type and updated the UI fragments so you can actually query `Gallery.o_counter`.
- Added `OCounter` to `GalleryFilterType` and wired up a filter handler (`galleryOCounterCriterionHandler`) that runs a SQL subquery to compute `SUM(images.o_counter)` per gallery, then applies integer criteria through the existing `getIntCriterionWhereClause`.
- Added `OCountByGalleryID` to `ImageStore`, which grabs `COALESCE(SUM(o_counter), 0)` for images tied to a gallery. Also updated the `ImageCounter` interface and hooked up the `galleryResolver.OCounter` resolver to call it.
- Updated the image reader/writer mocks to include `OCountByGalleryID`, and exposed `o_counter` as a filterable option in the UI list filters.

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -576,6 +576,8 @@ input GalleryFilterType {
   performer_age: IntCriterionInput
   "Filter by number of images in this gallery"
   image_count: IntCriterionInput
+  "Filter by o-counter"
+  o_counter: IntCriterionInput
   "Filter by url"
   url: StringCriterionInput
   "Filter by date"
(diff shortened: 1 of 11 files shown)


WithoutPants:
While there is room for a resolved sum o-counter, I'd not want it implemented as the o-counter for galleries. Galleries should really have their own o-counter like scenes, images and groups do.

Issue #5788, opened 2025-03-29
[meta] Sorting Feature Requests


feederbox826:
Since there's so much overlap in issues and sub-issues...

New Field:
- Breast/ Cup Size #4831 / #1661

Date
- Relative date #5010
- Performer age #2956
  - Age when filmed #2564 
- Folder modified date #2622
- "Liked" / Rating date #2492
- Recent activity: #3478

Average
- Studio/ Performer rating #3272

Aggregate:
- Studio-wide O-Counter #5024
- Image O-Count #2836 #5364
- Size of files on disk #5530
- Play Count/ Play Duration #4499
- Total duration #3470

Other fields:
- Studio name #5480
- Studio code #5480

Misc:
- Chained sorts: #1508
- UI/UX #1545
- Session persistence #1896

GalacticCheat:
Maista sucks cock
![Image](https://github.com/user-attachments/assets/ed12369e-cb07-434a-b49f-fbc470272035)
https://i.ibb.co/ds49xcXt/image-2-copie.jpg

efirlus:
Especially needs **[Multiple Sorts](https://github.com/stashapp/stash/issues/1508)**, so I made [stashsorter](https://github.com/efirlus/stashsorter). It's basically external script, so you can not control this within stash gui.

Issue #2956, opened 2022-09-28
[Feature] Add the ability to sort Galleries by Performer Age


Erovall:
Please the ability to sort Galleries by Performer Age.

PR #6879, opened 2026-04-30
Add gallery sorting by performer age

slick-daddy:
Provides the ability to sort galleries by performer age so users can find galleries based on the youngest or oldest performers associated with a gallery. Related #2956 and #5788 

### Description

- Add `performer_age` to `gallerySortOptions` and implement sorting logic in `setGallerySort` that uses a subquery to compute performer age with `MIN`/`MAX` aggregation and `COALESCE` fallbacks based on sort direction. 
- Add test cases for `performer_age` ascending and descending to `TestGalleryQuerySorting` in `gallery_test.go`.
- Expose the sort option in the frontend by adding the `performer_age` entry to `sortByOptions` and creating `PerformerAgeCriterionOption` in `ui/v2.5/src/models/list-filter/galleries.ts`.

### Testing

- Ran the targeted backend unit test `go test ./pkg/sqlite -run TestGalleryQuerySorting` which includes the new `performer_age` cases and it passed. 
- Ran the full backend test suite `go test ./...` and it completed successfully.

### DIFF
--- a/pkg/sqlite/gallery.go
+++ b/pkg/sqlite/gallery.go
@@ -860,10 +860,12 @@ func (qb *GalleryStore) setGallerySort(query *queryBuilder, findFilter *models.F
 	case "performer_count":
 		query.sortAndPagination += getCountSort(galleryTable, performersGalleriesTable, galleryIDColumn, direction)
 	case "performer_age":
-		// Looking at the youngest performer by default.
+		// Multi-performer semantics:
+		// - ASC sorts by the youngest performer in each gallery (MIN age)
+		// - DESC sorts by the oldest performer in each gallery (MAX age)
 		aggregation := "MIN"
 		if direction == "DESC" {
-			// When sorting by performer age DESC, consider oldest performer instead.
+			// DESC uses oldest performer age for each gallery.
 			aggregation = "MAX"
 		}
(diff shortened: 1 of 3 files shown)


pkg/sqlite/gallery.go
```diff
@@ -858,6 +859,34 @@ func (qb *GalleryStore) setGallerySort(query *queryBuilder, findFilter *models.F
...
+			aggregation = "MAX"
+		}
+		var fallback string
+		if direction == "ASC" {
+			// ASC puts NULL first by default, so coalesce to sqlite max int.
+			fallback = "9223372036854775807"
+		} else {
+			// DESC puts larger values first; coalesce NULL to sqlite min int to keep NULLs last.
+			fallback = "-9223372036854775808"
+		}
```
WithoutPants, line 878: This is different to the scene sorting code. Is there a good reason for it to be?

slick-daddy: Well, sort of. This number is one more higher from the highest 64 bit integer hence never conflicts with other numbers and from that we understand the case is invalid. It is a bit overkill, you can align it with scenes code if you want to keep feature parity, it makes sense too.

Issue #3573, opened 2023-03-21
[Feature] Filter out in Duplicate Checker


erebous:
It'd be nice if we can filter out, for example by path, in the Scene Duplicate Checker. In my case I have several deepfake videos marked as duplicates that I don't want to remove, adding too much noise to the list.

DogmaDragon:
Similar to https://github.com/stashapp/stash/issues/1656

PR #6884, opened 2026-05-03
feat(api): add scene_filter parameter to findDuplicateScenes

slick-daddy:
## Description

~~Closes~~ Related to #3573 (WithoutPants edit: does not include UI changes)

Adds `scene_filter` parameter support to the `findDuplicateScenes` query, allowing users to exclude specific scenes from duplicate checking results.

For example, a user can now filter out scenes by path, tag, or any other supported scene filter criteria. This means scenes they intentionally want to keep (e.g. deepfakes, alternate versions) will no longer appear as unwanted noise in the duplicate checker.

## Changes

- **Schema**: Added optional `scene_filter` argument to `findDuplicateScenes`
- **Backend**: Updated `SceneReaderWriter` interface and mock to accept a `SceneFilterType`
- **API**: Passes the scene filter from the resolver down to the duplicate checking repository
- **SQLite**: `FindDuplicates` query is now dynamically built based on the provided filter
- **Tests**: Updated SQLite-level duplicate checking tests to cover filtered queries

> **Note**: This PR implements the backend and GraphQL API only. 
> Frontend integration (adding filter UI to the duplicate checker) is not included.

### DIFF
--- a/pkg/sqlite/scene.go
+++ b/pkg/sqlite/scene.go
@@ -1430,11 +1430,58 @@ func (qb *SceneStore) GetStashIDs(ctx context.Context, sceneID int) ([]models.St
 	return sceneRepository.stashIDs.get(ctx, sceneID)
 }
 
-func (qb *SceneStore) FindDuplicates(ctx context.Context, distance int, durationDiff float64) ([][]*models.Scene, error) {
+func (qb *SceneStore) FindDuplicates(ctx context.Context, distance int, durationDiff float64, filter *models.SceneFilterType) ([][]*models.Scene, error) {
 	var dupeIds [][]int
+
+	query, err := qb.makeQuery(ctx, filter, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add necessary joins for duplicate checking
+	query.addJoins(
+		join{
+			table:    scenesFilesTable,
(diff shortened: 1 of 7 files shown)


pkg/sqlite/scene.go
```diff
@@ -41,41 +41,6 @@ const (
...
-		, abs(max(video_files.duration) OVER (PARTITION by files_fingerprints.fingerprint) - video_files.duration) as durationDiff
-	FROM scenes
-	INNER JOIN scenes_files ON (scenes.id = scenes_files.scene_id)
-	INNER JOIN files ON (scenes_files.file_id = files.id)
-	INNER JOIN files_fingerprints ON (scenes_files.file_id = files_fingerprints.file_id AND files_fingerprints.type = 'phash')
-	INNER JOIN video_files ON (files.id == video_files.file_id)
-)
-WHERE durationDiff <= ?1
-    OR ?1 < 0   --  Always TRUE if the parameter is negative.
-                --  That will disable the durationDiff checking.
```
Gykes, line 60: Seems like this comment was removed. Please add it back or explain the removal.

Gykes (changes requested):
The only test changes I see are passing a nil. Seems like the core feature is untested. Please add testing support for this.

PR #6885, opened 2026-05-04
[feat] Configure fields when batch tagging performers

Wasylq:
Added a "Configure fields" button that allows you to select which fields get refreshed when you do "batch update performers"

This field allows you to sync all data with exceptions, or merging your local edits with stashbox data

### Motivation

I have a well curated set of images per performer, images that I would like to keep, but still refresh other fields from the stashbox. I decided to add this little button that would allow you to choose which files get skipped during a batch refresh of performers. During development I figured, you might as well have possibility to merge stashbox data with your local for some fields (currently only URL and Aliases, couldn't find any other that might need that)

### Testing

The binary builds, during my tests the button behaves as expected, allowing you to sync only partial data from stashbox.


WithoutPants (approved):
Looks ok. Some comments that might steer future work:
- it's likely that users will expect the field selection to be saved between runs
- this creates a feature discrepancy between performers and all other batch updateable types, so might be worth adding to other types
- I don't love the necessity of including a legend underneath and it presents differently to the identify dialog, but I don't have any better ideas.

PR #6896, opened 2026-05-08
Add 'aliases' filter to Group (GraphQL, backend, tests, and UI)

### LINKED ISSUE #4383: [Feature] Ability to filter Movies by Aliases
Currently there's no option to filter Movies by their Aliases, even though filters for every other field of a Movie exists.

Add the option to filter Movies by Aliases. As the field takes in string values, the filter options like in Synopsis would be helpful, where you have options such as: is, is not, includes, excludes, etc.

slick-daddy:
Closes #4383

### Motivation
- Enable filtering groups by their aliases end-to-end so users can query and list groups by alias values. 
- Expose the new filter in the GraphQL schema and support it in the SQLite query builder. 
- Surface the option in the UI list filter so users can create alias-based criteria.

### Description
- Added `aliases: StringCriterionInput` to the GraphQL `GroupFilterType` in `filters.graphql`.
- Added `Aliases *StringCriterionInput` to the Go `models.GroupFilterType` and wired a `stringCriterionHandler` for `groups.aliases` in the SQLite group filter handler.
- Added unit test `TestGroupQueryAliases` in `pkg/sqlite/group_test.go` to validate alias filtering with multiple criterion modifiers.
- Exposed `aliases` in the UI list filter options by adding `createStringCriterionOption("aliases")` in `ui/v2.5/src/models/list-filter/groups.ts`.

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -423,6 +423,8 @@ input GroupFilterType {
   NOT: GroupFilterType
 
   name: StringCriterionInput
+  "Filter by aliases"
+  aliases: StringCriterionInput
   director: StringCriterionInput
   synopsis: StringCriterionInput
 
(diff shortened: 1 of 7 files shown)


Gykes:
I think this one should wait until https://github.com/stashapp/stash/pull/6951 lands.

PR #6900, opened 2026-05-08
Add sorting and filtering options to sub-groups UI

### LINKED ISSUE #6863: Add sorting and filtering options for sub groups

On the group details page, add the sorting and filtering options available for scenes and performers to subgroups as well.

Image

Currently sub-groups are sorted by the date they were last updated. Adding these sorting options would allow users to sort groups by commonly desirable methods - e.g. sorting a JAV series chronologically by using date.

None currently.

### Additional context

A similar issue is outlined in #6076, however it is requesting for an ability to add a custom ordering. I would prefer to be able to use the existing filters/sorting system in order to avoid having to specify an order for each group. However, I believe #6076 would still be useful for scenarios that the default filtering/sorting doesn't cover - e.g. having a custom ordering would be useful if the user wants to create a playlist of their favourite movies.

Stash-KennyG:
Closes [#6863](https://github.com/stashapp/stash/issues/6863): **Add sorting and filtering options for sub groups**

## Problem Statement

Sub-groups on the Group details page did not have parity with the standard list sorting experience. They were effectively constrained to the existing behavior, which made common ordering workflows (for example chronological or metadata-driven browsing) harder than elsewhere in the app.

## Design Intent

This change intentionally uses the existing Group sorting/filtering pipeline and removes  the one-off nature of the previous sub-group implementation.  The implementation was basically hiding the UI and using a field sort that was available elsewhere. This has some core benefits:

- Keep behavior simple and predictable by reusing existing list/filter architecture.
- Avoid special-case UI plumbing for sub-groups.
- Allso fixes an edge case where the content doesn't display properly on page-load when there are no scenes directly attached to the parent group.
- Simplifies shared SQL sort logic in one place so behavior is consistent across callers.

### UI Sample
image

## Changes

- Implemented sorting and filtering for sub-groups
- Added ability to sort by sub_group description value
- Updated SQL sort handling so natural sort fields are centralized, with null-safe handling for description (`COALESCE`).
- Removed hard-set sort from sub-groups allowing the same view sort mechanics as other pages - save, set-default etc
- Added `sub_group_description` to the standard Group sort options (instead of custom per-view options).
- Added localization for `sub_group_description` in `en-GB.json`.
- Added integration test coverage for subgroup description sorting, including:
  - ASC and DESC behavior
  - empty description handling
  - both relevant subgroup sort query paths
- Added a test for safety with our SQL function changes

### Eratta
I discovered one issue that I couldn't sort out.  The way we have built the page we are double off-setting the `FilterdListToolbar.tsx`/`SavedFilterList.tsx` window.  This moved it to a strange window location.  I have added a menuPortalTarget to `FilterListToolbar` element which will prevent this from happening.  I don't love it but I spent hours trying to trace the real cause and couldn't find a better way.  I found a few  `//hack` comments in the small areas I examined chasing this down, so one more  sub-optimal-fix here shouldn't be too bad.

Cursor helped me write this.  I reviewed it...


WithoutPants (approved):
I made some changes, the details of which can be viewing in the commit descriptions. It was quicker for me to make the changes rather than explaining what I wanted.

The short version is that I changed the manual ordering functionality so that it only applies when the manual sort by (`sub_group_order`) was selected. Otherwise, we'd end up with some odd behaviour when trying to reorder when sorting by a different field.

I also refactored the sort handling code to not put `description`-handling logic in the `getSort` function.

Code looks ok and testing worked as expected. Thanks for the submission.

PR #6905, opened 2026-05-09
Fix phash precision loss in JSON import/export

### LINKED ISSUE #6894: phash is imported and rounded as float64

### Describe the bug

When importing form json, phashes lose some precision due to (likely float64) interpretation

### Steps to reproduce

1. Scan a scene with a well-known phash
2. Export the database to .json backup
3. Import the JSON metadata

### Expected behaviour

phash is stored as string or hex representation to reduce precision loss

Attached is sqlite dump pre and post-import and included metadata from the export.

The phash (numberical) is as follows
```diff
+ 6391326009969271747
- 6391326009969272000
```
it's truncated at the 16th digit which according to [stack overflow](https://stackoverflow.com/a/69044661) is around the precision of float64. It's already exceeding 32bit int limit 

Originally reported by Gyvari, apparently a longstanding issue

### Stash version

v0.31.0, v0.31.1

### Device details

_No response_

### Relevant log output

```shell

```

slick-daddy:
## Description

This PR fixes phash precision loss when importing from JSON (fixes #6894).

## Problem

When importing from JSON, phashes lose precision because Go's JSON decoder unmarshals numbers into `float64` when decoding into `interface{}`. Float64 has only ~16 decimal digits of precision, but phashes are int64 values that can have up to 19 digits.

Example from issue #6894:
- Original: `6391326009969271747`
- After import: `6391326009969272000` (truncated)

## Solution

Convert phash values to hex strings during JSON export/import to avoid float64 precision truncation.

### Changes:

1. **Export** (`internal/manager/task_export.go`): 
   - Phash values are now stored as hex strings (e.g., `"58b3a02c9c9c27e3"`)
   - Uses existing `utils.PhashToString()` helper

2. **Import** (`pkg/file/import.go`):
   - Accepts both hex strings (new format) and numbers (old format)
   - Warns when numeric phashes are detected: `"Importing phash as float64 X - may have precision loss. Consider re-exporting from a newer version."`
   - Uses existing `utils.StringToPhash()` helper

## Backward Compatibility

- **New exports** use hex strings (no precision loss)
- **Old exports** with numeric phashes still import but emit a warning about potential precision loss

## Verification

After the fix, the export JSON will contain:
```json
{
  "type": "phash",
  "fingerprint": "58b3a02c9c9c27e3"
}
```

Instead of the problematic:
```json
{
  "type": "phash", 
  "fingerprint": 6391326009969272000
}
```


stashcollection14:
i test it, and this build fix works

pkg/file/import.go
```diff
@@ -104,9 +107,34 @@ func (i *Importer) baseFileJSONToBaseFile(ctx context.Context, baseJSON *jsonsch
...
+		if fp.Type == models.FingerprintTypePhash {
+			switch v := fp.Fingerprint.(type) {
+			case string:
+				// New format: hex string
+				phash, err := utils.StringToPhash(v)
+				if err != nil {
+					logger.Warnf("Error parsing phash hex string %q: %v", v, err)
+				} else {
+					fingerprintValue = phash
+				}
```
Gykes, line 150: If I am understanding this correctly then it retains the old, bad, value. I think it would be better to return an error or continue to skip it entirely. The main reson being is that I feel like no phash is better than a bad useless phash.

Gykes (changes requested):
Quick static review. Didn't actually test

WithoutPants (approved):
I changed this to use `json.RawMessage` for the fingerprint value. This allowed for manually unmarshalling into either a string or an int64 value instead of automatically unmarshalling numbers to floating point numbers. I added a check to disallow floating point numbers completely. 

With this change, legacy numeric values can be imported without precision loss and new exports will use the phash string value. Also added unit tests for the various scenarios.

Manually verified that legacy numeric values import correctly.

PR #6906, opened 2026-05-09
Add depth support to tag scene_count filter for hierarchical sidebar display

### LINKED ISSUE #6797: Empy parent tags cannot be selected in side menu

### Describe the bug

In the side menu for tags empty parent tags do not show up. They can be selected in the filter menu above the scenes and when selected they show up in the side menu as well.

For example I have a tag structure like
```
Europe
  France
  Germany
  United Kingdom
America
  Canada
  United States
```
Europe will not show up in the side menu unless at least one scene is tagged with it. It is available in the filter menu and then you can select include sub-tags to actually get a non-empty set of scenes.

### Steps to reproduce

1. Go to Scenes 
2. open side menu Tags
3. search for parent tag 
4. so such tag in list

Alternatively
1. Go to Scenes
2. open filter dialog
3. select Tags
4. search for and select parent tag
5. select include sub-tags
6. Get results

### Expected behaviour

Parent tags without scenes that carry it should be available in the side menu just as in the filter menu.

_No response_

### Stash version

v0.31.0

### Device details

_No response_

### Relevant log output

```shell

```

slick-daddy:
Fixes #6797

Added `scene_count_depth: Int` to` TagFilterType` in the GraphQL schema, backend model, and filter handler. The filter now supports a recursive CTE to count scenes across descendant tags. The sidebar query uses `scene_count_depth: -1` to include all descendants.

Files changed:

`graphql/schema/types/filters.graphql` — added `scene_count_depth` field
`pkg/models/tag.go` — added `SceneCountDepth` to struct
`pkg/sqlite/tag_filter.go` — added depth-aware scene count with recursive CTE
`ui/v2.5/src/components/List/Filters/TagsFilter.tsx` — set `scene_count_depth: -1` for sidebar queries
`pkg/sqlite/tag_test.go` — added tests for depth=nil, depth=-1, depth=1

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -635,6 +635,9 @@ input TagFilterType {
   "Filter by number of scenes with this tag"
   scene_count: IntCriterionInput
 
+  "Depth for scene count filter (0/nil = direct only, -1 = all descendants, N = N levels)"
+  scene_count_depth: Int
+
   "Filter by number of images with this tag"
   image_count: IntCriterionInput
 
--- a/pkg/models/tag.go
+++ b/pkg/models/tag.go
@@ -16,6 +16,8 @@ type TagFilterType struct {
 	IsMissing *string `json:"is_missing"`
 	// Filter by number of scenes with this tag
 	SceneCount *IntCriterionInput `json:"scene_count"`
+	// Depth for scene count filter (0/nil = direct only, -1 = all descendants, N = N levels)
+	SceneCountDepth *int `json:"scene_count_depth"`
 	// Filter by number of images with this tag
 	ImageCount *IntCriterionInput `json:"image_count"`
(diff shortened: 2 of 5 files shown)


WithoutPants:
I don't think a separate input field is a good solution, and this only addresses scenes - it does not address the issue for any of the other types that use tags.

My rough approach would be to create a new `HierarchicalCountInput` input, identical to `IntCriterionInput` but adding an optional `depth: Int` field. The existing `scene_count`, `image_count`, `gallery_count`, `performer_count`, `studio_count`, `movie_count`, `group_count`, and `marker_count` fields in the `TagFilterType` input would be changed to the new input type.

slick-daddy:
I agree with you. My initial thought was to implement this then extend the scope to other types you mentioned. Your approach is better from SWE perspective and more feasible for long term. I am closing this PR but at least we decided on the approach.

PR #6907, opened 2026-05-09
Move lightbox animation/scroll settings to LocalForage

### LINKED ISSUE #6864: Move disableAnimation lightbox option from GraphQL to UI config
Having to do a graphql changes for lightbox options is pretty rough. We should move this into the UI config at some point.

_Originally posted by @WithoutPants in https://github.com/stashapp/stash/pull/6354#pullrequestreview-3527977439_

slick-daddy:
Closes #6864

### Changes

- Removed `disableAnimation` and `scrollAttemptsBeforeChange` from `ConfigImageLightboxInput` and `ConfigImageLightboxResult` GraphQL types
- Removed `ImageLightboxDisableAnimation` and `ImageLightboxScrollAttemptsBeforeChange` constants and their reading logic from the backend config
- Removed `DisableAnimation` and `ScrollAttemptsBeforeChange` fields from the `ConfigImageLightboxResult` Go struct
- Removed the corresponding `setConfigBool`/`setConfigInt` calls from the `ConfigureInterface` mutation resolver
- Removed these fields from the frontend GraphQL query in `graphql/data/config.graphql`
- Updated `SettingsInterfacePanel.tsx` to read/write these settings via LocalForage instead of the GraphQL mutation
- Updated `Lightbox.tsx` to read these settings from LocalForage instead of server config
- Added `ILightboxConfig` local interface in `LocalForage.ts` to decouple LocalForage storage from GraphQL types

### Note

Users who had previously set non-default values for `scrollAttemptsBeforeChange` or `disableAnimation` will need to reconfigure these settings locally via the Settings UI after upgrading. A server-to-client migration was considered but is not feasible since these fields were removed from the GraphQL query.

### DIFF
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -361,8 +361,6 @@ input ConfigImageLightboxInput {
   scaleUp: Boolean
   resetZoomOnNav: Boolean
   scrollMode: ImageLightboxScrollMode
-  scrollAttemptsBeforeChange: Int
-  disableAnimation: Boolean
 }
 
 type ConfigImageLightboxResult {
@@ -371,8 +369,6 @@ type ConfigImageLightboxResult {
   scaleUp: Boolean
   resetZoomOnNav: Boolean
   scrollMode: ImageLightboxScrollMode
-  scrollAttemptsBeforeChange: Int!
-  disableAnimation: Boolean
 }
 
 input ConfigInterfaceInput {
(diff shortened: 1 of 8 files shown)


WithoutPants:
This is the wrong solution to the related issue. That issue refers to moving lightbox options in the UI config, _not_ the local forage. Furthermore, removing fields from the graphql schema causes compatibility issues with client code. They should be deprecated in the shorter term.

PR #6909, opened 2026-05-10
Add support for numeric ratings per-scene, per-tag

dontwannaname14:
Introduces the concept of tags that can support numeric ratings and allows a scene to have per-tag ratings.  Born from https://discourse.stashapp.cc/t/numeric-tag-ranking/7092/4


WithoutPants:
This isn't something I'd consider adding to the core stash system. I believe the use-case to be too niche to justify the maintenance overhead supporting it. I would consider adding custom fields to relationships like `scene -> tag` which would potentially facilitate something like this in a plugin.

dontwannaname14:
@WithoutPants any hints on how those custom fields might be implemented?  Im assuming it'd require a db schema change?   Does sound like it would be all an external-plugin would need to implement this

PR #6912, opened 2026-05-11
Withdrawn

callumlory:
Withdrawn.


feederbox826:
default nonroot is a breaking change. This can already be accomplished without a fake "nonroot" container. The container still has access to root privileges and this will break nonroot deployments like in k8s

PR #6916, opened 2026-05-11
Withdrawn

callumlory:
Withdrawn.

### DIFF
--- a/graphql/schema/types/metadata.graphql
+++ b/graphql/schema/types/metadata.graphql
@@ -17,6 +17,7 @@ input GenerateMetadataInput {
   imagePhashes: Boolean
   imageThumbnails: Boolean
   clipPreviews: Boolean
+  embeddedCaptions: Boolean
 
   "scene ids to generate for"
   sceneIDs: [ID!]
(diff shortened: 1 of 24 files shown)


smith113-p:
Hm, typically Stash does not modify the filesystem in this kind of automatic way...

MisterMcDuck:
FWIW I've wanted this feature for a while and went about testing it locally by merging the PR changes in. 

For anyone else who wanted to test, you'll need to increment the DB version for the change to apply:
```
diff --git a/pkg/sqlite/database.go b/pkg/sqlite/database.go
index 7c383dc4..026a18c0 100644
--- a/pkg/sqlite/database.go
+++ b/pkg/sqlite/database.go
@@ -34,7 +34,7 @@ const (
        cacheSizeEnv = "STASH_SQLITE_CACHE_SIZE"
 )
 
-var appSchemaVersion uint = 85
+var appSchemaVersion uint = 86
 
 //go:embed migrations/*.sql
 var migrationsBox embed.FS
```
Note: you may want to back your copy up, as this is a change to the DB that might leave you unable to continue forward if these changes aren't merged by the next release.

With that said, I've tried a few videos and note the captions are now showing when they did not previously, and can confirm > 1000 caption srts under the generated/captions folder. 

Thanks to @callumlory for implementing this change!

PR #6917, opened 2026-05-11
Persist list/grid display mode per page across navigation

### LINKED ISSUE #3773: [Feature] Persistent settings/toggles via cookies

slick-daddy:
Persist list/grid display mode per page across navigation

Store each page's `displayMode` in IndexedDB via `useInterfaceLocalForage` so
the user's list/grid preference is restored when navigating back to a page,
instead of always resetting to the default.

Applies to all entity list pages: Scenes, Images, Galleries, Performers,
Studios, Tags, Groups, and their detail sub-panels. URL `disp=` parameter
takes precedence over the persisted value.


slick-daddy:
I would be happy if I can get a general comment/review on this PR. It doesn't have to be detailed at the beginning. I want to know what people think about my approach. If we agree that the approach is correct, I think it is worth to get a proper review.

PR #6918, opened 2026-05-11
fix(sqlite): use LEFT JOIN for string-list criteria to fix OR filtering

### LINKED ISSUE #6914: `findTags` with OR not working as expected when tag has no aliases

### Describe the bug

In a plugin, I am using a call to `findTags` to retrieve a tag by name/alias. Since I don't know if what I'm searching for is a name or alias, I am searching for the name OR alias being equal to the provided value.

At some point this seems to have stopped working as expected. If the tag I am searching for has no aliases, it will never be returned by the query.

Looking at the trace logs, this seems to be due to the generated SQL query doing an `INNER JOIN` on the `tag_aliases` table. If this is a `LEFT JOIN`, the query returns the expected result.

Query in trace logs (edited for readability and to be more easily testable - full trace below):
```SQL
SELECT DISTINCT tags.id FROM tags 
INNER JOIN tag_aliases ON tag_aliases.tag_id = tags.id 
WHERE (tags.name LIKE '%Ahegao%') OR ((tag_aliases.alias LIKE '%Ahegao%')) 
```

### Steps to reproduce

1. Create a Tag with no aliases.
2. Use below query/filter to attempt to find this tag - replace `Ahegao` with your test tag name. No results are returned.
3. Add an alias to the tag (the value does not matter). Retry the prior query - the one expected result will be returned.

Query:
```GQL
query ($tag_filter: TagFilterType!, $filter: FindFilterType!) { findTags(filter: $filter, tag_filter: $tag_filter) { tags { id, name, sort_name, favorite, description, aliases, image_path, parents {id, name}, stash_ids {endpoint, stash_id, updated_at } } } }
```

Variables: 
```JSON
{
    "tag_filter": {
        "name": {
            "value": "Ahegao",
            "modifier": "EQUALS"
        },
        "OR": {
            "aliases": {
                "value": "Ahegao",
                "modifier": "EQUALS"
            }
        }
    },
    "filter": {
        "per_page": -1,
        "sort": "name"
    }
}
```

### Expected behaviour

A `findTags` query with multiple criteria joined by an "OR" and one of those criteria filtering on `aliases` should return all results that match a valid criteria regardless of if the expected result has any aliases. 

_No response_

### Stash version

v0.31.1-46-g01a75833

### Device details

_No response_

### Relevant log output

```shell
26-05-10 23:20:28
Trace   
SQL [516.9µs]: SELECT DISTINCT tags.id FROM tags INNER JOIN tag_aliases ON tag_aliases.tag_id = tags.id WHERE (tags.name LIKE ?) OR ((tag_aliases.alias LIKE ?)) ORDER BY COALESCE(tags.sort_name, tags.name) COLLATE NATURAL_CI ASC, COALESCE(tags.sort_name, tags.name, tags.id) COLLATE NATURAL_CI ASC , args: [Ahegao Ahegao]
2026-05-10 23:20:28
Trace   
SQL [524.1µs]: SELECT COUNT(*) as count FROM (SELECT DISTINCT tags.id FROM tags INNER JOIN tag_aliases ON tag_aliases.tag_id = tags.id WHERE (tags.name LIKE ?) OR ((tag_aliases.alias LIKE ?))) as temp, args: [Ahegao Ahegao]
```

slick-daddy:
Fixes #6914

## Problem

`findTags` (and equivalent queries for other entity types) silently dropped results when an OR filter combined a `name` condition with an `aliases` condition, and the matched entity had no aliases.

Root cause: `stringListCriterionHandlerBuilder.handler()` used `joinTypeInner` as the default join type for string-list tables (`tag_aliases`, `performer_urls`, etc.). In OR contexts, this caused the join to eliminate all rows without alias entries before the WHERE clause could evaluate the OR — making the `name` branch unreachable for those rows.

The bug was amplified by the join-merging pipeline in `filterBuilder.getAllJoins()` (`filter.go:425–436`): sub-filter joins are merged into the main query via `joins.addUnique()`, which upgrades any existing LEFT JOIN to INNER if an INNER join on the same table is encountered (`filter.go:157–159`). This meant even a correctly written outer query would inherit the INNER JOIN from the OR sub-filter.

Generated SQL (before):
```sql
SELECT DISTINCT tags.id FROM tags
INNER JOIN tag_aliases ON tag_aliases.tag_id = tags.id
WHERE (tags.name LIKE ?) OR (tag_aliases.alias LIKE ?)
```

Tags with no aliases produce no rows in `tag_aliases`, so they are eliminated before the WHERE clause runs — the `tags.name` branch never gets evaluated.

## Fix

Changed the default join type from `joinTypeInner` to `joinTypeLeft` in `stringListCriterionHandlerBuilder.handler()`.

Removed the now-redundant special case for `IsNull` and `NotMatchesRegex` modifiers — these existed solely to override the old `joinTypeInner` default. With `joinTypeLeft` as the default, both modifiers work correctly out of the box: the LEFT JOIN guarantees a row exists for every entity, so `(column IS NULL OR ...)` in the WHERE clause can reach entities without any list entries.

Generated SQL (after):
```sql
SELECT DISTINCT tags.id FROM tags
LEFT JOIN tag_aliases ON tag_aliases.tag_id = tags.id
WHERE (tags.name LIKE ?) OR (tag_aliases.alias LIKE ?)
```

## Safety

`LEFT JOIN + WHERE col = value` and `INNER JOIN + WHERE col = value` produce identical results in non-OR queries: `NULL = 'x'` evaluates to `NULL` (falsy), so rows without list entries are still filtered out. The behavioral difference only surfaces in OR contexts, which is exactly the bug being fixed.

## Scope

Affects all callers of `stringListCriterionHandlerBuilder` — tags, performers, studios, scenes, images, galleries, and files — all receive the same fix for their respective string-list fields (aliases, URLs, captions, fingerprints, etc.).

### DIFF
--- a/pkg/sqlite/criterion_handlers.go
+++ b/pkg/sqlite/criterion_handlers.go
@@ -629,10 +629,7 @@ func (m *stringListCriterionHandlerBuilder) handler(criterion *models.StringCrit
 				// 	Modifier: models.CriterionModifierNotNull,
 				// }, m.joinTable+"."+m.stringColumn)(ctx, f)
 			} else {
-				joinType := joinTypeInner
-				if criterion.Modifier == models.CriterionModifierIsNull || criterion.Modifier == models.CriterionModifierNotMatchesRegex {
-					joinType = joinTypeLeft
-				}
+				joinType := joinTypeLeft
 				m.addJoinTable(f, joinType)
 				stringCriterionHandler(criterion, m.joinTable+"."+m.stringColumn)(ctx, f)
 			}


WithoutPants:
This is a bandaid fix for a single scenario and doesn't fix the underlying issue introduced by #6648. Either all of the `left -> inner` join changes need to be reverted, or ideally a solution worked out to handle cases where the or sub-clause is used.

PR #6919, opened 2026-05-12
Fix gallery tab content not loading on performer page when opened in new tab

### LINKED ISSUE #6798: Content fails to render when opening performer page in new tab/window under certain conditions

### Describe the bug

When opening a performer's detail page in a new tab, the gallery cards are not displayed.

The last html structure that appears is `<div id="performer-tabs-tabpane-galleries" aria-labelledby="performer-tabs-tab-galleries" role="tabpanel" aria-hidden="false" class="fade tab-pane active show"></div>` aka the nav-tab container.

Conditions:
- The performer must have a gallery count > 0 & scene count = 0
- This does not effect performers with a scene count > 0

Refreshing the performer detail page corrects the issue.

### Steps to reproduce

1. Go to Performer grid view.
2. Filter performers to show only those with 0 scenes and more than 0 galleries.
3. Right-click one of the performers in the filtered list.
4. Select “Open in new tab” for their detail page.

### Expected behaviour

All content loads and toolbar loads after the nav-tabs.

Image

### Stash version

v0.31.0-2-g0ed2992a

### Device details

Chrome 146.0.7680.177 (Official Build) (64-bit)

### Relevant log output

```shell

```

slick-daddy:
Fixes #6798

`activeKey={tabKey ?? populatedDefaultTab}` ensures the correct tab is shown on the very first render, before the URL is updated.

Tested with one and two filters and both worked.

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerDetails/Performer.tsx
@@ -119,7 +119,7 @@ const PerformerTabs: React.FC<{
       id="performer-tabs"
       mountOnEnter
       unmountOnExit
-      activeKey={tabKey}
+      activeKey={tabKey ?? populatedDefaultTab}
       onSelect={setTabKey}
     >
       <Tab


WithoutPants:
The `tabKey` field is used in other places, and defaulting it here isn't really a good fix. Note that the same issue exists on the studio and tags pages, which wasn't address here either. 

Superceded by #6923

PR #6924, opened 2026-05-13
[Feature] Consolidate scene cover buttons

### LINKED ISSUE #1278: [Feature] Add "Generate Thumbnail From Content" function to "Set Image" button.
There are currently two places to set a scene thumbnail.

1. The "Generate" menu at the scene menu

Screen Shot 2021-04-08 at 11 49 23 PM

2. The "Set Image" button at the bottom

Screen Shot 2021-04-08 at 11 49 52 PM

Adding "From current" to the lower image will reduce confusion, since the intended use of the button is to have a list of all the different ways to set the thumbnail.

Add "From Current" to "set image" in "scene" view. Leave it in the context menu as well.  Additionally, standardize language between the buttons. Are we calling it an "image" or a "thumbnail"

Stash-KennyG:
Implementing part of #6861 
Continuing from #6865 

## Summary

- Improves Scene cover image workflows in the edit UI by consolidating actions into one grouped menu and keeping source metadata in sync.

## Why This Change

- Cover image actions were split between multiple controls and were harder to discover/use consistently.
- Interface had grown to set cover images through multiple paths; UI layout was inconsistent. 
- Issue 6861 was marked as ready in the backlog

## UI
Scene edit panel now uses a single Set image menu near the preview window with grouped actions
image

When the Generate screenshot methods are selected, we also now asynchronously update the preview window when the job finishes.

## Schema shape change
Per request of @WithoutPants the GraphQL sceneGenerateScreenshot now returns the job id.
```
Since we're properly returning the job ID, I think it's safe to change the return value of sceneGenerateScreenshot from String! to ID! in the graphql schema.
```

### LLM Notes
I have done a first review of this and feel I've made some progress form feedback of others - thank you.
This was made with the help of [Cursor](https://github.com/stashapp/stash/pull/cursor.ai)

Closes https://github.com/stashapp/stash/issues/1278


Gykes:
You may want to move this over to the split button idea that I brought up in the other PR. If you already have then you should update the UI picture to reflect.

Stash-KennyG:
In considering that I'm not sure I follow the intent of the split button for this implementation. There's no strong default behavior on the set image split that I can think of here. 

The reference in the other thread shows a save where the option is a "save and new" alternative. It's not jumping out to me what the default would be here. Thoughts?
I am more than happy to convert the menu on the split, but I am just struggling to come up with the appropriate default here.

Gykes:
I'm not married to any particular setting nor did I mean it had to exactly be like that. I just mean we kind of started a standard for the arrow and click up method of the save and new. Get creative with it, I don't have something exact in mind. If we are consolidating the settings I just feel like the arrow up is cleaner rather than ...

Stash-KennyG:
I'll give it a go. Perhaps default is current timestamp... let me see where the vibes take me.

Gykes:
You may just be better keeping it as it. WP was right on the other PR. Lets just keep it non controversial. I'm back at my new location and mostly setup so can start doing reviews and what not again probably here over the next few days.

ui/v2.5/src/components/Shared/ImageInput.tsx
```diff
@@ -22,6 +28,8 @@ interface IImageInput {
   onImageURL?: (url: string) => void;
   onReset?: () => void;
   acceptSVG?: boolean;
+  onGenerateDefault?: () => void;
+  onGenerateCurrent?: () => void;
```
Gykes: This is a shared component across 6 edit screens. This current verbiage may be slightly confusing (I know it was for me). It doesn't say whats being generated. If every entity needs to do it this way then it will scale rapidly. We could go with something a little more generic like `extraActions`?

Brain dump with no testing: 
`extraActions?: Array<{ icon: IconDefinition; labelId: string; onClick: () => Promise<void> }>; `

This way the scene panel passes its own actions in.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
```diff
@@ -64,6 +64,8 @@ interface IProps {
   initialCoverImage?: string;
   isNew?: boolean;
   isVisible: boolean;
+  onGenerateThumbFromCurrent?: () => Promise<void>;
+  onGenerateThumbDefault?: () => Promise<void>;
```
Gykes, line 73: These ones are `Promise<void>;`, which is correct, but on `imageInput.tsx` you used just regular void. If you don't go with my suggestion on that page then those should be updated to match this.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
```diff
@@ -877,15 +881,25 @@ export const SceneEditPanel: React.FC<IProps> = ({
             {renderDetailsField()}
             <Form.Group controlId="cover_image">
               <Form.Label>
-                <FormattedMessage id="cover_image" />
+                {intl.formatMessage({ id: "cover_image" })}
```
Gykes: Was there a reason to remove the declarative version of this?

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
```diff
@@ -877,15 +881,25 @@ export const SceneEditPanel: React.FC<IProps> = ({
...
               </Form.Label>
               {image}
-              <ImageInput
-                isEditing
-                onImageChange={onCoverImageChange}
-                onImageURL={onImageLoad}
-                onReset={scene.id ? onResetCover : undefined}
-              />
+              <div className="mt-2 d-flex align-items-center">
```
Gykes: We may be better using SCSS here instead of stacking bootstrap layers. It's mostly a personal thing as it can cause accidental regressions.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
```diff
@@ -877,15 +881,25 @@ export const SceneEditPanel: React.FC<IProps> = ({
...
+                {!isNew && (
```
Gykes: Wouldn't this remove the upload/URL/Clipboard for new scenes? I think some explanation would be needed and comment that or remove the gaurd.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
```diff
@@ -877,15 +881,25 @@ export const SceneEditPanel: React.FC<IProps> = ({
...
+                    onImageURL={onImageLoad}
+                    onGenerateDefault={onGenerateThumbDefault}
+                    onGenerateCurrent={onGenerateThumbFromCurrent}
+                  />
+                )}
+                {scene.id && (
+                  <Button variant="danger" onClick={() => onResetCover()}>
+                    {intl.formatMessage({ id: "actions.clear_image" })}
+                  </Button>
+                )}
```
Gykes: This duplicates the Clear button that ImageInput already renders via onReset (still used by performers, studios, etc.). Same label, same variant, same visual location just rendered twice in two different places depending on entity. Was there a reason not to just keep `onReset={scene.id ? onResetCover : undefined}` on the ImageInput above, matching the pattern used by every other caller?

Gykes (changes requested):
A few of these comments are opinions and not just blatant bugs that need fixing. Please feel free to push back if you have a valid reason.

ui/v2.5/src/components/Shared/ImageInput.tsx
Stash-KennyG: I like this pattern, I think it implements nicely and controls sprawl.
This should be addressed in next push.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
Stash-KennyG: Uncaught stylistic perference change.  Keeping this within the existing convention.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
Stash-KennyG: This has been moved per the refactor'd consolidated method referenced below.  Rather than re-implementing it all over again, we have a CustomFieldsInput now.

internal/api/resolver_mutation_scene.go
```diff
@@ -1319,11 +1319,12 @@ func (r *mutationResolver) SceneDeleteO(ctx context.Context, id string, t []*tim
...
 	if at != nil {
-		manager.GetInstance().GenerateScreenshot(ctx, id, *at)
+		jobID = manager.GetInstance().GenerateScreenshot(ctx, id, *at)
 	} else {
-		manager.GetInstance().GenerateDefaultScreenshot(ctx, id)
+		jobID = manager.GetInstance().GenerateDefaultScreenshot(ctx, id)
 	}
 
-	return "todo", nil
+	return strconv.Itoa(jobID), nil
```
WithoutPants, line 1329: Since we're properly returning the job ID, I think it's safe to change the return value of `sceneGenerateScreenshot` from `String!` to `ID!` in the graphql schema.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
WithoutPants: Indeed it does:

Image

The guard should be placed on the `extraActions` field instead.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
```diff
@@ -880,12 +920,17 @@ export const SceneEditPanel: React.FC<IProps> = ({
                 <FormattedMessage id="cover_image" />
               </Form.Label>
               {image}
-              <ImageInput
-                isEditing
-                onImageChange={onCoverImageChange}
-                onImageURL={onImageLoad}
-                onReset={scene.id ? onResetCover : undefined}
-              />
+              <div className="mt-2 details-edit">
```
WithoutPants: I don't think it's necessary to move the `ImageInput` into a div here.

WithoutPants (changes requested):
The usability of the new button is not great. I don't think a split button makes sense for this functionality since there's no obvious expected behaviour when clicking the main part of the button. Having an ellipsis in the button label also suggests to me that clicking the button will open a menu or dialog. To me it just works better as a dropdown menu. Removing it will also mean we can get rid of the `hasPlayerPlayedThisSession` stuff which just feels clunky.

I notice that the `From file...` option doesn't highlight when moused over. This also occurs in the existing menu, but we should fix it here. 

The item icons aren't well aligned and should be fixed:

Image

Stash-KennyG:
Argh, sloppy rollback/merge.  I was trying to avoid having to close and reopen a branch for this feature.
Let me try to cleanup the files that shouldn't have been touched in this PR...

Stash-KennyG:
Sorry @feederbox826, I don't know how to untag you now that I pulled out the sloppy merge changes.

I show only the expected files are changed.
The schema change is based on my interpretation of the request from @WithoutPants on changing the return shape for the generate job.
The description is now updated to cover the feature.
* Consolidated buttons
* Aligned formatting
* Updated the shape of the GraphQL endpoint to return the jobID.
* Dynamically updatating the UI when a generate job completes.

My initial review of the code looks good and I think we are ready for re-review.

ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
WithoutPants: Please revert this back to use `FormattedMessage`.

ui/v2.5/src/components/Shared/ImageInput.tsx
WithoutPants: This has been marked resolved but has not had the suggested changes made.

WithoutPants (changes requested):

ui/v2.5/src/components/Shared/ImageInput.tsx
Stash-KennyG: This was addressed but lost in a rollback.  I brought it back.  Next push should fix...

Stash-KennyG:
Sorry @WithoutPants, as above the miss-communication was a result in the rollback based on the rejected UX changes that ultimately threw out that other change.  I went back in and fixed that.  It's now been better gerneralized to address that.  Also the internationalization has been addressed as well.

WithoutPants (approved):

Issue #6564, opened 2026-02-10
Stash Object Sync Project


Gykes:
### Describe the feature you'd like

# Feature Parity Across Object Types

## Problem

Stash's object types (Scenes, Performers, Images, Galleries, Groups, Tags, Studios) have significant feature disparities. Scenes receive the most development attention, leaving other types with inconsistent sort options, filter criteria, and missing field support. This creates a fragmented user experience where workflows available for one type aren't available for others.

AI was used to generate all the graphs. I did fix some issues but there might be more after I corrected some.

### Common Features Across Types

| Feature | Scenes | Performers | Images | Galleries | Groups | Tags | Studios |
|---------|:------:|:----------:|:------:|:---------:|:------:|:----:|:-------:|
| Rating | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x: | :white_check_mark: |
| Favorite | :x: | :white_check_mark: | :x: | :x: | :x: | :white_check_mark: | :white_check_mark: |
| Organized | :white_check_mark: | :x: | :white_check_mark: | :white_check_mark: | :x: | :x: | :x: |
| O-Counter | :white_check_mark: | :white_check_mark:* | :white_check_mark: | :x: | :white_check_mark: | N/A | :white_check_mark:* |
| Play Count | :white_check_mark: | :white_check_mark:* | N/A | N/A | N/A | N/A | N/A |
| Last Played | :white_check_mark: | :white_check_mark:* | N/A | N/A | N/A | N/A | N/A |
| Stash ID | :white_check_mark: | :white_check_mark: | :x: | :x: | :x: | :white_check_mark: | :white_check_mark: |
| Custom Fields | :x: | :white_check_mark: | :x: | :x: | :x: | :white_check_mark: | :white_check_mark: |
| Tags | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | N/A | :white_check_mark: |
| Performers | :white_check_mark: | N/A | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x: | :x: |
| Studios | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x: | N/A |
| Date | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x: | :x: |
[...]

PR #6928, opened 2026-05-15
Add performers field to Tag type

slick-daddy:
## Summary

Adds a `performers: [Performer!]!` field to the `Tag` GraphQL type, allowing direct querying of performers associated with a specific tag.

The `performers_tags` junction table already existed in the database, so no migration is needed.

### Backend
- `graphql/schema/types/tag.graphql` — Added `performers: [Performer!]!` field to Tag type
- `pkg/models/model_tag.go` — Added `PerformerIDs RelatedIDs` field and `LoadPerformerIDs` method to Tag struct; added `PerformerIDs` to TagPartial
- `pkg/models/relationships.go` — Added `GetPerformerIDs` to TagRelationLoader interface
- `pkg/sqlite/tables.go` — Added `tagsPerformersTableMgr` join table manager
- `pkg/sqlite/tag.go` — Implemented `GetPerformerIDs` method querying `performers_tags`
- `internal/api/resolver_model_tag.go` — Added `Performers` resolver following the Scene/Image/Gallery pattern
- `pkg/models/mocks/TagReaderWriter.go` — Added `GetPerformerIDs` mock method

## Pattern

Follows the same IDs → loader → fill pattern used for `Scene`, `Image`, and `Gallery` resolvers.

Related #6564


WithoutPants:
I don't see a requirement for this. We can query performers for a given tag already using the `findPerformers` query which also provides proper paging etc. The referenced issue doesn't mention this as far as I can determine.

PR #6929, opened 2026-05-15
feat: Add depth support to tag count filters for hierarchical sidebar display

### LINKED ISSUE #6797: Empy parent tags cannot be selected in side menu

slick-daddy:
## Description
Second attempt to fix #6797

Parent tags with no directly-tagged content now appear in the sidebar because the `scene_count` (and `image_count`, `gallery_count`, etc.) filters use `depth: -1` to include descendant tags.
- Introduced `HierarchicalCountInput` (identical to `IntCriterionInput` + optional `depth`) in the GraphQL schema and Go model
- Changed all 8 `TagFilterType` count fields to use the new input type
- Tag count filter handlers use a recursive CTE when `depth` is set, walking `tags_relations` to aggregate across descendant tags
- Sidebar query passes `depth: -1` so parent tags show up even without direct content

### DIFF
--- a/graphql/schema/types/filters.graphql
+++ b/graphql/schema/types/filters.graphql
@@ -632,29 +632,29 @@ input TagFilterType {
   "Filter to only include tags missing this property"
   is_missing: String
 
-  "Filter by number of scenes with this tag"
-  scene_count: IntCriterionInput
+  "Filter by number of scenes with this tag. Depth controls whether descendant tags are included."
+  scene_count: HierarchicalCountInput
 
-  "Filter by number of images with this tag"
-  image_count: IntCriterionInput
+  "Filter by number of images with this tag. Depth controls whether descendant tags are included."
+  image_count: HierarchicalCountInput
 
-  "Filter by number of galleries with this tag"
-  gallery_count: IntCriterionInput
+  "Filter by number of galleries with this tag. Depth controls whether descendant tags are included."
+  gallery_count: HierarchicalCountInput
 
-  "Filter by number of performers with this tag"
-  performer_count: IntCriterionInput
+  "Filter by number of performers with this tag. Depth controls whether descendant tags are included."
+  performer_count: HierarchicalCountInput
 
-  "Filter by number of studios with this tag"
-  studio_count: IntCriterionInput
+  "Filter by number of studios with this tag. Depth controls whether descendant tags are included."
+  studio_count: HierarchicalCountInput
 
-  "Filter by number of movies with this tag"
-  movie_count: IntCriterionInput
+  "Filter by number of movies with this tag. Depth controls whether descendant tags are included."
+  movie_count: HierarchicalCountInput
 
-  "Filter by number of group with this tag"
-  group_count: IntCriterionInput
+  "Filter by number of group with this tag. Depth controls whether descendant tags are included."
+  group_count: HierarchicalCountInput
 
-  "Filter by number of markers with this tag"
-  marker_count: IntCriterionInput
+  "Filter by number of markers with this tag. Depth controls whether descendant tags are included."
+  marker_count: HierarchicalCountInput
 
   "Filter by parent tags"
   parents: HierarchicalMultiCriterionInput
@@ -912,6 +912,13 @@ input IntCriterionInput {
   value: Int!
   value2: Int
   modifier: CriterionModifier!
+}
+
+input HierarchicalCountInput {
+  value: Int!
+  value2: Int
+  modifier: CriterionModifier!
+  depth: Int
 }
 
 input FloatCriterionInput {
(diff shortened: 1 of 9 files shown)


graphql/schema/types/filters.graphql
```diff
@@ -632,29 +632,29 @@ input TagFilterType {
   "Filter to only include tags missing this property"
   is_missing: String
 
-  "Filter by number of scenes with this tag"
-  scene_count: IntCriterionInput
+  "Filter by number of scenes with this tag. Depth controls whether descendant tags are included."
```
WithoutPants: If this behaviour is common to all instances of `HierarchicalCountInput` it probably makes better sense to move the behaviour comment to the `depth` field. It should probably be reworded to indicate that it's the number of descendant levels to include.

pkg/sqlite/tag_filter.go
```diff
@@ -217,80 +218,100 @@ func (qb *tagFilterHandler) isMissingCriterionHandler(isMissing *string) criteri
...
-	return func(ctx context.Context, f *filterBuilder) {
-		if sceneCount != nil {
-			f.addLeftJoin("scenes_tags", "", "scenes_tags.tag_id = tags.id")
-			clause, args := getIntCriterionWhereClause("count(distinct scenes_tags.scene_id)", *sceneCount)
-
-			f.addHaving(clause, args...)
-		}
+func (qb *tagFilterHandler) addHierarchicalCountCTE(f *filterBuilder, cteAlias string, depth int) {
+	var depthCondition string
+	if depth >= 0 {
```
WithoutPants: Needs documentation as to what a `depth < 0` means.

WithoutPants (changes requested):
In future, please include what manual testing you've performed.

Static code review looks mostly ok with some minor changes. I'll give it a test later.

WithoutPants (changes requested):
I have redone the query count integration tests into table-based tests, and added a single test for scene count = 1 with depth = 1. This has revealed an error when using depth = 1:
```
--- FAIL: TestTagQueryCounts (0.00s)
    --- FAIL: TestTagQueryCounts/scene_count_equals_1_depth_1 (0.00s)
        /home/WithoutPants/dev/stash/dev/pkg/sqlite/tag_test.go:646: scene_count_equals_1_depth_1: Error querying tag: error executing count query with SQL: WITH  RECURSIVE scenes_tags_desc(root_id, descendant_id) AS (
            			SELECT id, id FROM tags
            			UNION ALL
            			SELECT td.root_id, tr.child_id
            			FROM tags_relations tr
            			INNER JOIN scenes_tags_desc td ON td.descendant_id = tr.parent_id WHERE depth < 1
            		) SELECT COUNT(*) as count FROM (SELECT DISTINCT tags.id FROM tags LEFT JOIN scenes_tags_desc ON scenes_tags_desc.root_id = tags.id LEFT JOIN scenes_tags ON scenes_tags.tag_id = scenes_tags_desc.descendant_id GROUP BY tags.id  HAVING (count(distinct scenes_tags.scene_id) = ?)) as temp, args: [1], error: error executing `WITH  RECURSIVE scenes_tags_desc(root_id, descendant_id) AS (
            			SELECT id, id FROM tags
            			UNION ALL
            			SELECT td.root_id, tr.child_id
            			FROM tags_relations tr
            			INNER JOIN scenes_tags_desc td ON td.descendant_id = tr.parent_id WHERE depth < 1
            		) SELECT COUNT(*) as count FROM (SELECT DISTINCT tags.id FROM tags LEFT JOIN scenes_tags_desc ON scenes_tags_desc.root_id = tags.id LEFT JOIN scenes_tags ON scenes_tags.tag_id = scenes_tags_desc.descendant_id GROUP BY tags.id  HAVING (count(distinct scenes_tags.scene_id) = ?)) as temp` [[1]]: no such column: depth
FAIL
FAIL	github.com/stashapp/stash/pkg/sqlite	0.313s
FAIL
```

Please add unit tests for the other count types and at least one scenario each for depth > 0, and depth < 0.

WithoutPants (approved):
Looks and tests ok.

PR #6930, opened 2026-05-15
fix: close HTTP response bodies and add nil guards for Primary() calls

slick-daddy:
### HTTP Response Body Leaks (pkg/)
`resp.Body.Close()` was deferred after the status code check, meaning non-2xx
responses returned early without closing the body. Over time this silently
exhausts the HTTP connection pool. Fixed by moving `defer resp.Body.Close()`
immediately after the `client.Do()` error check in all affected paths.

- `pkg/scraper/url.go`
- `pkg/scraper/image.go`
- `pkg/utils/image.go`
- `pkg/pkg/repository_http.go` (HEAD and GET paths)

### Nil Pointer Guards (internal/manager/)
`Files.Primary()` returns nil when a scene or image has no associated files.
Chaining `.Base().Path` or passing the result directly to an encoder without
a nil check causes a panic in the background goroutine, silently killing the
task. Added early `return` guards after each `Primary()` call.

- `task_transcode.go`
- `task_generate_interactive_heatmap_speed.go`
- `task_generate_image_thumbnail.go`
- `task_generate_clip_preview.go`

### File Handle Cleanup (internal/manager/)
`os.OpenFile` handles in `task_import.go` were closed manually at each return
path. If an intermediate step returned early, later handles could be skipped.
Replaced with `defer` calls.

- `internal/manager/task_import.go`

## Notes
Zero behavioral changes. All fixes are defensive — no new logic, no API changes.

### DIFF
--- a/internal/manager/task_import.go
+++ b/internal/manager/task_import.go
@@ -170,21 +170,17 @@ func (t *ImportTask) unzipFile() error {
 		if err != nil {
 			return err
 		}
+		defer o.Close()
 
 		i, err := f.Open()
 		if err != nil {
-			o.Close()
 			return err
 		}
+		defer i.Close()
 
 		if _, err := io.Copy(o, i); err != nil {
-			o.Close()
-			i.Close()
 			return err
 		}
-
-		o.Close()
-		i.Close()
 	}
 
 	return nil
--- a/internal/manager/task_import.go
+++ b/internal/manager/task_import.go
@@ -162,28 +162,33 @@ func (t *ImportTask) unzipFile() error {
 			continue
 		}
 
-		if err := os.MkdirAll(filepath.Dir(fn), os.ModePerm); err != nil {
-			return err
-		}
-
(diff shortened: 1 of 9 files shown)


internal/manager/task_import.go
```diff
@@ -170,21 +170,17 @@ func (t *ImportTask) unzipFile() error {
 		if err != nil {
 			return err
 		}
+		defer o.Close()
```
WithoutPants: I think this would make better sense if the for loop contents were extracted into a separate function. As it is, it'll leave all the file handles open until the `unzipFile` function finishes.

WithoutPants (changes requested):
This looks mostly ok, with only one thing I would like changed.

I would like to see at least one of the changed code paths in the scraper code (ie `scraper.loadURL`) manually tested to ensure that it correctly handles the status code >= 400 scenario. Everything else I'm satisfied should work ok.

internal/manager/task_import.go
slick-daddy: adressed

WithoutPants (approved):

PR #6931, opened 2026-05-15
fix: fix stale closures, ref leaks, and React warnings in UI components

slick-daddy:
### useEffect Dependency Arrays (7 files)
Effects were missing dependency arrays, causing them to re-run on every render.
Keyboard shortcuts were being rebound and IndexedDB was being written
unnecessarily. Fixed by adding correct dependency arrays. Where dependencies
were unstable references (inline functions, context values), the ref pattern
was used instead to keep the effect stable while always accessing fresh values.

- `hooks/LocalForage.ts` — `[key, data]`
- `hooks/keybinds.ts` — `[isVisible, ratingSystem]`
- `components/MainNavbar.tsx` — refs for `openManual`/`newPath`, `[goto, menuItems, history]`
- `components/Galleries/GalleryList.tsx` — refs for `onEdit`/`onDelete`, `[hasSelection]`
- `components/Studios/StudioList.tsx` — same
- `components/Performers/PerformerList.tsx` — same
- `components/Groups/GroupList.tsx` — same

### createRef → useRef (1 file)
`React.createRef()` creates a new ref object on every render, causing the
dependent `useEffect` to re-fire every render. `useRef` returns a stable
object, so the effect now runs once after mount.

- `hooks/Lightbox/LightboxImage.tsx`

### Fragment Key (1 file)
`<>...</>` shorthand inside `.map()` cannot accept a `key` prop. Replaced with
`<React.Fragment key={scene.id}>` to eliminate the React warning.

- `components/SceneDuplicateChecker/SceneDuplicateChecker.tsx`

### Console Cleanup (1 file)
Replaced `console.log` with `console.error` on a fetch error path.

- `src/App.tsx`

## Notes
Zero behavioral changes.

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryList.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryList.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect } from "react";
+import React, { useCallback, useEffect, useRef } from "react";
 import { FormattedMessage, useIntl } from "react-intl";
 import cloneDeep from "lodash-es/cloneDeep";
 import { useHistory } from "react-router-dom";
@@ -307,16 +307,21 @@ export const FilteredGalleryList = PatchComponent(
       setShowSidebar,
     });
 
+    const onEditRef = useRef(onEdit);
+    onEditRef.current = onEdit;
+    const onDeleteRef = useRef(onDelete);
+    onDeleteRef.current = onDelete;
+
     useEffect(() => {
       Mousetrap.bind("e", () => {
         if (hasSelection) {
-          onEdit?.();
+          onEditRef.current?.();
         }
       });
 
       Mousetrap.bind("d d", () => {
         if (hasSelection) {
-          onDelete?.();
+          onDeleteRef.current?.();
         }
       });
 
@@ -324,7 +329,7 @@ export const FilteredGalleryList = PatchComponent(
         Mousetrap.unbind("e");
         Mousetrap.unbind("d d");
       };
-    });
+    }, [hasSelection]);
 
     const onCloseEditDelete = useCloseEditDelete({
       closeModal,
--- a/ui/v2.5/src/components/Groups/GroupList.tsx
+++ b/ui/v2.5/src/components/Groups/GroupList.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect } from "react";
+import React, { useCallback, useEffect, useRef } from "react";
 import { FormattedMessage, useIntl } from "react-intl";
 import cloneDeep from "lodash-es/cloneDeep";
 import Mousetrap from "mousetrap";
@@ -271,16 +271,21 @@ export const FilteredGroupList = PatchComponent(
       setShowSidebar,
     });
 
+    const onEditRef = useRef(onEdit);
+    onEditRef.current = onEdit;
+    const onDeleteRef = useRef(onDelete);
+    onDeleteRef.current = onDelete;
+
     useEffect(() => {
       Mousetrap.bind("e", () => {
         if (hasSelection) {
-          onEdit?.();
+          onEditRef.current?.();
         }
       });
 
       Mousetrap.bind("d d", () => {
         if (hasSelection) {
-          onDelete?.();
+          onDeleteRef.current?.();
         }
       });
 
@@ -288,7 +293,7 @@ export const FilteredGroupList = PatchComponent(
         Mousetrap.unbind("e");
         Mousetrap.unbind("d d");
       };
-    });
+    }, [hasSelection]);
 
     const onCloseEditDelete = useCloseEditDelete({
       closeModal,
--- a/ui/v2.5/src/components/MainNavbar.tsx
+++ b/ui/v2.5/src/components/MainNavbar.tsx
@@ -254,6 +254,11 @@ export const MainNavbar: React.FC = () => {
     }
   }
 
+  const openManualRef = useRef(openManual);
+  openManualRef.current = openManual;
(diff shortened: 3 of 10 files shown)


ui/v2.5/src/components/Galleries/GalleryList.tsx
```diff
@@ -307,24 +307,29 @@ export const FilteredGalleryList = PatchComponent(
       setShowSidebar,
     });
 
+    const onEditRef = useRef(onEdit);
+    onEditRef.current = onEdit;
+    const onDeleteRef = useRef(onDelete);
+    onDeleteRef.current = onDelete;
```
WithoutPants, line 313: This is not an appropriate solution. The idiomatic solution is to convert the related function to a callback using `useCallback`.

ui/v2.5/src/components/Groups/GroupList.tsx
```diff
@@ -271,24 +271,29 @@ export const FilteredGroupList = PatchComponent(
       setShowSidebar,
     });
 
+    const onEditRef = useRef(onEdit);
+    onEditRef.current = onEdit;
+    const onDeleteRef = useRef(onDelete);
+    onDeleteRef.current = onDelete;
```
WithoutPants, line 277: Use `useCallback` instead.

ui/v2.5/src/components/SceneDuplicateChecker/SceneDuplicateChecker.tsx
```diff
@@ -850,7 +850,7 @@ export const SceneDuplicateChecker: React.FC = () => {
                   scene.files.length > 0 ? scene.files[0] : undefined;
 
                 return (
-                  <>
+                  <React.Fragment key={scene.id}>
```
WithoutPants, line 853: This isn't a good solution. It would be better to move the underlying code into a separate react component.

ui/v2.5/src/components/Studios/StudioList.tsx
```diff
@@ -264,24 +264,29 @@ export const FilteredStudioList = PatchComponent(
       setShowSidebar,
     });
 
+    const onEditRef = useRef(onEdit);
+    onEditRef.current = onEdit;
+    const onDeleteRef = useRef(onDelete);
+    onDeleteRef.current = onDelete;
```
WithoutPants, line 270: Use `useCallback`

ui/v2.5/src/components/Performers/PerformerList.tsx
```diff
@@ -430,24 +430,29 @@ export const FilteredPerformerList = PatchComponent(
       setShowSidebar,
     });
 
+    const onEditRef = useRef(onEdit);
+    onEditRef.current = onEdit;
+    const onDeleteRef = useRef(onDelete);
+    onDeleteRef.current = onDelete;
```
WithoutPants, line 436: Use `useCallback`

ui/v2.5/src/components/MainNavbar.tsx
```diff
@@ -254,29 +254,34 @@ export const MainNavbar: React.FC = () => {
     }
   }
 
+  const openManualRef = useRef(openManual);
+  openManualRef.current = openManual;
+  const newPathRef = useRef(newPath);
+  newPathRef.current = newPath;
+
```
WithoutPants, line 261: Use `useCallback`

ui/v2.5/src/hooks/Lightbox/LightboxImage.tsx
```diff
@@ -107,7 +107,7 @@ export const LightboxImage: React.FC<IProps> = ({
   const mouseDownEvent = useRef<MouseEvent>();
   const resetPositionRef = useRef(resetPosition);
 
-  const container = React.createRef<HTMLDivElement>();
+  const container = useRef<HTMLDivElement>(null);
```
WithoutPants, line 110: This breaks video playing

WithoutPants (changes requested):
I'm not convinced of the benefits of a lot of these changes, and given that there is a regression in the lightbox image code, it doesn't suggest to me that sufficient testing was performed. 

The keybind handling had dependencies omitted because the key bind/unbind operations are trivial and don't impact other functionality. Adding dependencies adds unnecessary complexity for very little benefit.

> IndexedDB was being written unnecessarily.

Do you have any proof of this? Did you test the before and after scenario?

slick-daddy:
Added refs for onEdit/onDelete with [hasSelection] deps because it was running on every render but now only runs when hasSelection actually changes if I am not mistaken. Same pattern applied to openManual/newPath.

 <React.Fragment key={scene.id}> in map() eliminates React warning about missing keys and has no downsides.

I agree with LightboxImage regression, it needs to be fixed.

WithoutPants:
> Added refs for onEdit/onDelete with [hasSelection] deps because it was running on every render but now only runs when hasSelection actually changes if I am not mistaken. Same pattern applied to openManual/newPath.

It's still an incorrect change. And you still didn't answer as to what meaningful benefit there is. Running on every render isn't necessarily an issue unless it's performing something taxing on every render, and you've given no indication or proof that it does. Nor have you indicated how you know that IndexedDB is being unnecessarily written to.

> 
> <React.Fragment key={scene.id}> in map() eliminates React warning about missing keys and has no downsides.

Sure does, but moving it into a separate component is a cleaner solution.

> I agree with LightboxImage regression, it needs to be fixed.

The apparent attitude of your overall response doesn't sit well with me. I don't pretend to have a full knowledge of react, but you haven't demonstrated a superior knowledge that warrants debating over what review feedback you can accept and reject.

PR #6932, opened 2026-05-15
feat: add scene marker count sort option for studios

### LINKED ISSUE #6662: Sort by Marker Count on Studios Tab

I would like there to be a way to sort studios by marker count (including sub studios).

This would provide another way for users to organize their Stash. For me personally, I search/scrape newly-added scenes using the StashDB stash box source but TPDB has better markers. I have the *TPDB Markers* plugin and would like to be able to bulk rescrape only the studios with the most logged markers on TPDB (e.g. NVG Network sites and ExCoGi sites).

I *could* just re-search all scenes based on a "missing TPDB Stash ID" filter but that would take too long. I've already re-searched by TPDB on the scenes in the aforementioned networks but I would like to be able to identify which other sites have the most markers already so that I can re-scrape newer scenes.

No.

### Additional context

_No response_

slick-daddy:
Related #5788, #6564 Closes #6662

Adds the ability to sort studios by scene marker count. It resolves from studios as requested and should be consistent with other sortings.

Tested and works.

In future, we need to decide how networks should behave when we implement same sorting options for scenes.


slick-daddy:
@WithoutPants Do you think we need to implement depth for this sorting? Tags have sub-tags but that is not the case for markers. If you think flat approach is good enough, I will leave this PR as is.

WithoutPants:
To be consistent with other sorting options, I think it should be counting only markers that are directly related to the studio via scene.

PR #6933, opened 2026-05-16
Add recursive sort by performer_count and o_counter for studios, o_counter depth parameter, display updates

slick-daddy:
Related #6564
## Summary

- **Studio o_counter(depth)**: `o_counter` now accepts a `depth: Int` parameter. Passing `depth: -1` (or any non-zero value) recursively counts o-history across all sub-studios using a `WITH RECURSIVE` CTE. Displayed on cards, list view, and detail page as `o_counter_all`.
- **Recursive sort**: Studios can be sorted by `performer_count` and `o_counter`, both resolved via recursive CTE to include sub-studios.
- **List view**: Added `o_counter` column to the studio list table.
- **Display consistency**: StudioCard, StudioListTable, and Studio details use `_all` variants (`scene_count_all`, `performer_count_all`, `o_counter_all`) for recursive counts across sub-studios.,,

## UI Changes

- Sort options: Studio list filter now has sort by "O-Count" and "Performer Count" options.
- List table: New "O-Count" column showing recursive o-counter with sub-studios included.
- Cards: Parent studio cards now display scene_count_all (scenes + sub-studios), performer_count_all (performers + sub-studios), and o_counter_all instead of flat counts.
- Detail page: The o-counter badge in the parent studio detail card now shows the recursive total (all sub-studios included).

Tested and works.

### DIFF
--- a/graphql/schema/types/studio.graphql
+++ b/graphql/schema/types/studio.graphql
@@ -26,8 +26,7 @@ type Studio {
   updated_at: Time!
   groups: [Group!]!
   movies: [Movie!]! @deprecated(reason: "use groups instead")
-  o_counter: Int
-
+  o_counter(depth: Int): Int! # Resolver
   custom_fields: Map!
 }
 
--- a/graphql/schema/types/studio.graphql
+++ b/graphql/schema/types/studio.graphql
@@ -11,12 +11,15 @@ type Studio {
   organized: Boolean!
 
   image_path: String # Resolver
+  
   scene_count(depth: Int): Int! # Resolver
   image_count(depth: Int): Int! # Resolver
   gallery_count(depth: Int): Int! # Resolver
   performer_count(depth: Int): Int! # Resolver
   group_count(depth: Int): Int! # Resolver
   movie_count(depth: Int): Int! @deprecated(reason: "use group_count instead") # Resolver
+  o_counter(depth: Int): Int! # Resolver
+
   stash_ids: [StashID!]!
   # rating expressed as 1-100
   rating100: Int
@@ -26,8 +29,6 @@ type Studio {
   updated_at: Time!
   groups: [Group!]!
   movies: [Movie!]! @deprecated(reason: "use groups instead")
-  # o_counter with depth: Int (default 0, flat count). Pass depth: -1 for recursive count across all sub-studios.
-  o_counter(depth: Int): Int! # Resolver
   custom_fields: Map!
 }
 
--- a/graphql/schema/types/studio.graphql
+++ b/graphql/schema/types/studio.graphql
@@ -26,6 +26,7 @@ type Studio {
   updated_at: Time!
   groups: [Group!]!
   movies: [Movie!]! @deprecated(reason: "use groups instead")
+  # o_counter with depth: Int (default 0, flat count). Pass depth: -1 for recursive count across all sub-studios.
   o_counter(depth: Int): Int! # Resolver
   custom_fields: Map!
 }
--- a/pkg/sqlite/image.go
+++ b/pkg/sqlite/image.go
@@ -736,31 +736,7 @@ func (qb *ImageStore) OCountByStudioID(ctx context.Context, studioID int, depth 
 	var ret int
 
 	if depth != 0 {
-		q := `
-		WITH RECURSIVE sub_studios AS (
-			SELECT id, 0 AS level FROM studios WHERE id = ?
-			UNION ALL
-			SELECT s.id, ss.level + 1 FROM studios s
-			INNER JOIN sub_studios ss ON s.parent_id = ss.id
-			WHERE ss.level < ? OR ? < 0
-		)
-		SELECT COALESCE(SUM(o_counter), 0) FROM images
-		WHERE images.studio_id IN (SELECT id FROM sub_studios)`
-
-		rows, err := dbWrapper.QueryxContext(ctx, q, studioID, depth, depth)
-		if err != nil {
-			return 0, fmt.Errorf("querying image o_count by studio: %w", err)
-		}
-		defer rows.Close()
-		for rows.Next() {
-			if err := rows.Scan(&ret); err != nil {
-				return 0, fmt.Errorf("scanning image o_count: %w", err)
-			}
-		}
-		if err := rows.Err(); err != nil {
-			return 0, fmt.Errorf("iterating image o_count rows: %w", err)
-		}
-		return ret, nil
+		return qb.oCountByStudioIDRecursive(ctx, studioID, depth)
 	}
 
 	table := qb.table()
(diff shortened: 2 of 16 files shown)


graphql/schema/types/studio.graphql
```diff
@@ -26,8 +26,7 @@ type Studio {
   updated_at: Time!
   groups: [Group!]!
   movies: [Movie!]! @deprecated(reason: "use groups instead")
-  o_counter: Int
-
+  o_counter(depth: Int): Int! # Resolver
```
WithoutPants: Needs documentation of what `depth` is and what the default value is.

pkg/sqlite/image.go
```diff
@@ -700,13 +700,41 @@ func (qb *ImageStore) OCountByPerformerID(ctx context.Context, performerID int)
 	return ret, nil
 }
 
-func (qb *ImageStore) OCountByStudioID(ctx context.Context, studioID int) (int, error) {
+func (qb *ImageStore) OCountByStudioID(ctx context.Context, studioID int, depth *int) (int, error) {
```
WithoutPants: Accept a value rather than a pointer. The zero-value has the same behaviour as `nil` here, so there's little point making it a pointer.

pkg/sqlite/image.go
```diff
@@ -700,13 +700,41 @@ func (qb *ImageStore) OCountByPerformerID(ctx context.Context, performerID int)
...
+	var ret int
+
+	if depth != nil && *depth != 0 {
```
WithoutPants: The `depth` value doesn't appear to even be used here?

pkg/sqlite/scene.go
```diff
@@ -826,7 +826,37 @@ func (qb *SceneStore) OCountByGroupID(ctx context.Context, groupID int) (int, er
 	return ret, nil
 }
 
-func (qb *SceneStore) OCountByStudioID(ctx context.Context, studioID int) (int, error) {
+func (qb *SceneStore) OCountByStudioID(ctx context.Context, studioID int, depth *int) (int, error) {
+	var ret int
+
+	if depth != nil && *depth != 0 {
```
WithoutPants: Depth value isn't being used.

pkg/sqlite/studio.go
```diff
@@ -658,6 +658,40 @@ func (qb *StudioStore) sortByLatestScene(direction string) string {
 	return " ORDER BY (" + selectStudioLatestSceneSQL + ") " + direction
 }
 
+func (qb *StudioStore) sortByPerformerCount(direction string) string {
```
WithoutPants, line 661: This behaviour is contradictory to all of the other sort options which don't include sub studios in their count sorting. I think a `performer_count` sort option should be consistent with the other count options. 

This leaves open the question of whether and how to implement sorting that includes counts of sub-studios, but that should probably be covered in a separate discussion and PR.

pkg/sqlite/studio.go
```diff
@@ -658,6 +658,40 @@ func (qb *StudioStore) sortByLatestScene(direction string) string {
...
+			INNER JOIN sub_studios ss ON ch.parent_id = ss.id
+		)
+		SELECT COUNT(DISTINCT %[2]s.performer_id)
+		FROM sub_studios ss
+		INNER JOIN %[4]s sc ON sc.%[3]s = ss.id
+		INNER JOIN %[2]s ON %[2]s.scene_id = sc.id
+	) %[5]s`, studioTable, performersScenesTable, studioIDColumn, sceneTable, getSortDirection(direction))
+}
+
+func (qb *StudioStore) sortByOCounter(direction string) string {
```
WithoutPants, line 685: As above, this is contradictory to other sort options.

ui/v2.5/src/components/Studios/StudioDetails/Studio.tsx
```diff
@@ -511,8 +511,8 @@ const StudioPage: React.FC<IProps> = ({ studio, tabKey }) => {
                   clickToRate
                   withoutContext
                 />
-                {!!studio.o_counter && (
-                  <OCounterButton value={studio.o_counter} />
+                {!!studio.o_counter_all && (
+                  <OCounterButton value={studio.o_counter_all} />
```
WithoutPants: I'm not convinced that this will be a desirable change for all users. It would probably be better to include a UI option of whether to include sub-studios in the o-counter counts.

ui/v2.5/src/components/Studios/StudioCard.tsx
```diff
@@ -98,13 +98,13 @@ export const StudioCard: React.FC<IProps> = PatchComponent(
     }
 
     function maybeRenderScenesPopoverButton() {
-      if (!studio.scene_count) return;
+      if (!studio.scene_count_all) return;
```
WithoutPants: As above, I don't know if this is necessarily desirable for all users, and would be better suited with an option.

WithoutPants (changes requested):
Please be a bit more thorough in your testing. The `depth` value not being used would have been identified had you actually tested your changes. Unit testing also would have caught it, and should be added for the new `*Store` functions.

slick-daddy:
I implemented O Count (All) and Performer Count (All) as a sorting option. Now it should be less controversial.

I realized some details while testings.

- Network cards don't show O count, performer count and scene count in their respective cards as a badge, but studios do. When sorted, this creates asymmetry. I think we shouldadd these badges to networks as well. It should resolve from sum of their respective entities. This would improve the general user experience. The only controversial part will be for the networks that has their own scenes (not binded to any of their sub-studios). I think this part should be discussed. It can be added as a config option to settings.

- If a network has only one sub-studio, that means their o count will be same. The algorithm first shows (sorts) the network and then the sub-studio. This is a nice touch.

pkg/sqlite/studio_test.go
```diff
@@ -1942,12 +1942,139 @@ func TestStudioQueryCustomFields(t *testing.T) {
...
 
-// TODO Create
-// TODO Update
-// TODO Destroy
-// TODO Find
-// TODO FindBySceneID
-// TODO Count
-// TODO All
-// TODO AllSlim
-// TODO Query
```
WithoutPants, line 1953: This should not have been removed.

pkg/sqlite/image.go
```diff
@@ -732,13 +732,42 @@ func (qb *ImageStore) OCountByPerformerID(ctx context.Context, performerID int)
 	return ret, nil
 }
 
-func (qb *ImageStore) OCountByStudioID(ctx context.Context, studioID int) (int, error) {
+func (qb *ImageStore) OCountByStudioID(ctx context.Context, studioID int, depth int) (int, error) {
+	var ret int
+
+	if depth != 0 {
+		q := `
```
WithoutPants: Move this into its own private function.

pkg/sqlite/scene.go
```diff
@@ -823,7 +823,38 @@ func (qb *SceneStore) OCountByGroupID(ctx context.Context, groupID int) (int, er
 	return ret, nil
 }
 
-func (qb *SceneStore) OCountByStudioID(ctx context.Context, studioID int) (int, error) {
+func (qb *SceneStore) OCountByStudioID(ctx context.Context, studioID int, depth int) (int, error) {
+	var ret int
+
+	if depth != 0 {
+		q := `
```
WithoutPants: Move into private function.

WithoutPants (changes requested):

graphql/schema/types/studio.graphql
WithoutPants: I chose to move this to the other `depth` accepting fields instead. They should probably be properly documented, but I'll defer that for now.

WithoutPants (approved):

PR #6934, opened 2026-05-16
Add click-to-sort column headers to all entity list table views

slick-daddy:
~~This PR shouldn't be merged before #6933 gets merged.~~ We are good now.

The only thing I didn't like about this PR is reverse mapping approach. Maybe we should edit onClick instead. I am open to feedbacks.

## Summary

Makes column headers clickable in list table views for Scenes, Galleries,
Performers, Studios, and Tags. The sort infrastructure is generic so all
entity tables benefit without per-entity duplication.
### Changes After Review
Server: favorite sort option for Performer, Studio, Tag sort options
Server: birthdate sort case for age column support
`StudioListTable`: Favourite column with FavoriteCell
`StudioListTable`: RelatedCell uses sub_studios pluralized label

### Generic sort support (`ListTable.tsx`)
- Added optional `sortable` flag to `IColumn` — columns without a backend
  sort field (image, composite, etc.) are marked `sortable: false`.
- Added `onSort`, `sortBy`, `sortDirection` props. Active sort column
  renders ▲ (ASC) or ▼ (DEC) indicator.
- Clicking a header fires `onSort(column.value)` — new column starts ASC,
  same column toggles direction.
### Per-entity wiring (`*List.tsx` + `*ListTable.tsx`)
Each entity follows the same pattern:
1. **`*ListTable.tsx`** — accepts `onSort`/`sortBy`/`sortDirection` props
   and threads them through to `ListTable`. Non-sortable columns marked
   `sortable: false`.
2. **`*List.tsx`** — `Filtered*List` defines a module-level
   `columnSortMap` translating column values to backend sort fields.
   `onSort` useCallback maps the column value via this map before calling
   `filter.setSortBy()`. The inner `*List` component derives
   `activeSortColumn` by reverse-mapping `filter.sortBy` back to the
   column value so `ListTable` matches correctly for arrow display.
### Column mappings (value → backend field)
| Entity | Mappings |
|---|---|
| Studios | `scene_count→scenes_count`, `image_count→images_count`, `gallery_count→galleries_count` |
| Scenes | `scene_code→code` |
| Galleries | `images→images_count` |
| Performers | `height_cm→height`, `penis_length_cm→penis_length`, `scene_count→scenes_count`, `gallery_count→galleries_count`, `image_count→images_count`, `age→birthdate` |
| Tags | `scene_count→scenes_count`, `gallery_count→galleries_count`, `image_count→images_count`, `group_count→groups_count`, `performer_count→performers_count`, `studio_count→studios_count` |
### Bug fixes
- **Active sort arrow not showing**: `filter.sortBy` contained the backend
  field name (e.g. `"scenes_count"`) but `ListTable` compared against
  `column.value` (e.g. `"scene_count"`). Fixed by reverse-mapping to
  column value before passing to the table.
- **TypeScript index error**: `filter.sortBy` could be `undefined`, which
  can't be used as an object index. Fixed with `?? ""` fallback.
### Studio child count label
- `StudioListTable` `RelatedCell` now shows "N Studios" (pluralized via
  `intl.formatMessage`) instead of a bare number.
## Files changed (11 files)
| File | Changes |
|---|---|
| `ui/.../List/ListTable.tsx` | Generic sort header support |
| `ui/.../Studios/StudioList.tsx` | Sort plumbing + column map |
| `ui/.../Studios/StudioListTable.tsx` | Sort props + non-sortable columns + "N Studios" label |
| `ui/.../Scenes/SceneList.tsx` | Sort plumbing + column map |
| `ui/.../Scenes/SceneListTable.tsx` | Sort props + non-sortable columns |
| `ui/.../Galleries/GalleryList.tsx` | Sort plumbing + column map |
| `ui/.../Galleries/GalleryListTable.tsx` | Sort props + non-sortable columns |
| `ui/.../Performers/PerformerList.tsx` | Sort plumbing + column map |
| `ui/.../Performers/PerformerListTable.tsx` | Sort props + non-sortable columns |
| `ui/.../Tags/TagList.tsx` | Sort plumbing + column map |
| `ui/.../Tags/TagListTable.tsx` | Sort props + non-sortable columns |


WithoutPants (changes requested):
Please include screenshots when making UI changes. Users shouldn't need to try it locally to find out what has changed.

For reference, here's how it looks on the scenes page, sorting by title ascending:

Image

I think it would look more consistent if we used the same icons as the sort selector and that the icon should be placed at the right edge:

Image

I would extract the `Icon` element from `ListFilter.tsx` and create an exported React component in the same file called `SortByIcon` which accepts a `SortDirectionEnum`. `SortBySelect` should replace its `Icon` with the new component. 

I was able to style it like above by putting contents of the column header `th` element into a `div` and styling it with `display: flex; align-items: center; justify-content: space-between`.

The other thing I noticed is that you've used `string` instead of `SortDirectionEnum` for the `sortBy` parameter. Please change these to use the enum.

While you are in there, please add `padding-bottom: 0` to the `.table-list .table thead th` css rule. The bottom padding is unnecessary and would be a good incidental fix.

ui/v2.5/src/components/Scenes/SceneList.tsx
```diff
@@ -112,7 +112,6 @@ function usePlaySelected(selectedIds: Set<string>) {
   const playScene = usePlayScene();
 
   const playSelected = useCallback(() => {
-    // populate queue and go to first scene
```
WithoutPants, line 115: Please revert these removals. There's no reason to remove them for this change.

ui/v2.5/src/components/Scenes/SceneList.tsx
```diff
@@ -179,15 +175,30 @@ function useAddKeybinds(filter: ListFilterModel, count: number) {
...
   "SceneList",
-  ({ scenes, filter, selectedIds, onSelectChange, fromGroupId }) => {
+  ({ scenes, filter, selectedIds, onSelectChange, fromGroupId, onSort }) => {
+    const reverseSortMap = useMemo(() => {
+      const rev: Record<string, string> = {};
+      Object.entries(sceneColumnSortMap).forEach(([k, v]) => {
+        rev[v] = k;
+      });
+      return rev;
+    }, []);
```
WithoutPants: This can be moved out of the function component. I would probably add a utility function for creating the reverse map (probably in `src/utils/data/ts`), and the reverse sort map can be declared below the sort map above.

Do this for the other list components as well.

ui/v2.5/src/components/List/ListTable.tsx
```diff
@@ -102,12 +111,33 @@ export const ListTable = <T extends { id: string }>(
...
 
   const columnHeaders = useMemo(() => {
-    return visibleColumns.map((column) => (
-      <th key={column.value} className={`${column.value}-head`}>
-        {column.label}
-      </th>
-    ));
-  }, [visibleColumns]);
+    return visibleColumns.map((column) => {
+      const isSortable = column.sortable !== false && onSort;
```
WithoutPants, line 115: Do sortable columns outnumber non-sortable columns? If not, then perhaps we should consider defaulting to `sortable = false`.

WithoutPants (changes requested):

WithoutPants:
There's a bunch of unrelated changes to the server code.

slick-daddy:
> There's a bunch of unrelated changes to the server code.

- Dropped unrelated 3 UI fix commits. My initial thought was it would be nice apply them on the way as well but I agree that they are out of scope, so let's drop them.

- Commit [4193f2a](https://github.com/stashapp/stash/pull/6934/commits/4193f2af0fc1175135a340edde3913d6908c714a) has also been dropped since it came from syncing fork by mistake.

- Rebased with cherry-picking and ready for review.

WithoutPants:
There's still changes to server files in this PR - specifically the `sqlite` package.

pkg/sqlite/performer.go
```diff
@@ -871,6 +872,17 @@ func (qb *PerformerStore) getPerformerSort(findFilter *models.FindFilterType) (s
 		sortQuery += qb.sortByLastOAt(direction)
 	case "latest_scene":
 		sortQuery += qb.sortByLatestScene(direction)
+	case "birthdate":
```
slick-daddy, line 875: Didn't drop favorite sort option and birthdate case because they are needed for Favourite/Age column clicks.

pkg/sqlite/performer.go
```diff
@@ -871,6 +872,17 @@ func (qb *PerformerStore) getPerformerSort(findFilter *models.FindFilterType) (s
...
+		// birthdate is inversely related to age (older = lower date).
+		// flip direction so DESC means "oldest first".
```
WithoutPants: This is wrong. We're sorting by birthdate and not by age. Reversing the sort direction for an age column in the UI should be performed by the UI, not by the server.

ui/v2.5/src/components/List/ListTable.tsx
WithoutPants: Answering my own question, it appears that there's more sortable than not. I don't love defaulting to false with a name like `sortable`, but I don't have an alternative.

ui/v2.5/src/components/Studios/StudioListTable.tsx
```diff
@@ -115,7 +148,11 @@ export const StudioListTable: React.FC<IStudioListTableProps> = ({
     const childLink =
       studio.child_studios && studio.child_studios.length > 0 ? (
         <Link to={NavUtils.makeChildStudiosUrl(studio)}>
-          {studio.child_studios.length}
+          {studio.child_studios.length}{" "}
+          {intl.formatMessage(
+            { id: "studios" },
```
WithoutPants: I think it would be better to introduce a new `sub-studios` localisation string to use here.

pkg/sqlite/performer.go
WithoutPants: You didn't include these changes in the original description.

WithoutPants (changes requested):
Please ensure that the description in your PRs matches the changes you've made. You made undocumented changes in this PR, specifically changes to the server sorting behaviour, changing the related studios column presentation, and the favourite column to the studio list table. Having to pore through the changes to discover undocumented changes is significantly time-consuming.

ui/v2.5/src/components/Studios/StudioDetails/Studio.tsx
```diff
@@ -511,9 +511,9 @@ const StudioPage: React.FC<IProps> = PatchComponent(
                     clickToRate
                     withoutContext
                   />
-                  {showAllCounts && !!studio.o_counter_all ? (
+                  {showAllCounts && studio.o_counter_all ? (
```
slick-daddy, line 514: This was an incidental fix from `pnpm run lint:js:fix`. It flagged unnecessary double-negation (!!) on o_counter fields which are already numbers, not booleans.

PR #6935, opened 2026-05-16
perf: memoize card components and fix IntersectionObserver churn

slick-daddy:
## Summary

- Wrap all SceneCard/GalleryCard sub-components in React.memo
  (PatchComponent pattern: _Inner → React.memo(_Inner))
- Move all onScrubberClick declarations to useCallback with proper deps
- Fix IntersectionObserver in ScenePreview: add [] dep array + disconnect() cleanup
- Strip scenes fragment in gallery-slim.graphql from SlimSceneData (89 fields) to id/title (2 fields)

## Impact
- Eliminates unnecessary re-renders of all cards on selection, hover and
  filter changes — noticeably smoother interaction on 40-card grids
- Stops 40 IntersectionObserver instances being created and destroyed on
  every render; removes observer churn during scroll and hover entirely
- Reduces gallery list query payload from ~89 fields per scene down to 2;
  significantly less network transfer for galleries with many scenes
  
  Note: Components using PatchComponent() are wrapped as:
  const _Inner = PatchComponent(...) 
  const Inner = React.memo(_Inner)
This order is required — wrapping PatchComponent itself breaks $$typeof.

Edit: It would be great to get some feedback from @ikmckenz on this PR.

### DIFF
--- a/ui/v2.5/graphql/data/gallery-slim.graphql
+++ b/ui/v2.5/graphql/data/gallery-slim.graphql
@@ -37,7 +37,11 @@ fragment SlimGalleryData on Gallery {
     image_path
   }
   scenes {
-    ...SlimSceneData
+    id
+    title
+    files {
+      path
+    }
   }
   paths {
     cover
--- a/ui/v2.5/src/components/Galleries/GalleryCard.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryCard.tsx
@@ -24,38 +24,36 @@ interface IGalleryPreviewProps {
   disabled?: boolean;
 }
 
-export const GalleryPreview: React.FC<IGalleryPreviewProps> = React.memo(({
-  gallery,
-  onScrubberClick,
-  disabled,
-}) => {
-  const [imgSrc, setImgSrc] = useState<string | undefined>(
-    gallery.paths.cover ?? undefined
-  );
+export const GalleryPreview: React.FC<IGalleryPreviewProps> = React.memo(
+  ({ gallery, onScrubberClick, disabled }) => {
+    const [imgSrc, setImgSrc] = useState<string | undefined>(
+      gallery.paths.cover ?? undefined
+    );
 
-  return (
-    <div className={cx("gallery-card-cover")}>
-      {!!imgSrc && (
-        <img
-          loading="lazy"
-          className="gallery-card-image"
-          alt={gallery.title ?? ""}
-          src={imgSrc}
-        />
-      )}
-      {gallery.image_count > 0 && (
-        <GalleryPreviewScrubber
-          previewPath={gallery.paths.preview}
-          defaultPath={gallery.paths.cover ?? ""}
-          imageCount={gallery.image_count}
-          onClick={onScrubberClick}
-          onPathChanged={setImgSrc}
-          disabled={disabled}
-        />
-      )}
-    </div>
-  );
-});
+    return (
+      <div className={cx("gallery-card-cover")}>
+        {!!imgSrc && (
+          <img
+            loading="lazy"
+            className="gallery-card-image"
+            alt={gallery.title ?? ""}
+            src={imgSrc}
+          />
+        )}
+        {gallery.image_count > 0 && (
+          <GalleryPreviewScrubber
+            previewPath={gallery.paths.preview}
+            defaultPath={gallery.paths.cover ?? ""}
+            imageCount={gallery.image_count}
+            onClick={onScrubberClick}
+            onPathChanged={setImgSrc}
+            disabled={disabled}
+          />
+        )}
+      </div>
+    );
+  }
+);
 
 interface IGalleryCardProps {
   gallery: GQL.SlimGalleryDataFragment;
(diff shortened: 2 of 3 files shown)


ui/v2.5/src/components/Scenes/SceneList.tsx
```diff
@@ -112,7 +112,6 @@ function usePlaySelected(selectedIds: Set<string>) {
   const playScene = usePlayScene();
 
   const playSelected = useCallback(() => {
-    // populate queue and go to first scene
```
Gykes, line 115: Why were these comments removed? I don't see a reason to remove them. There are a few on this page.

ui/v2.5/graphql/data/gallery-slim.graphql
```diff
@@ -37,7 +37,8 @@ fragment SlimGalleryData on Gallery {
     image_path
   }
   scenes {
-    ...SlimSceneData
+    id
+    title
```
Gykes, line 41: This trim only modifies the `.graphql` source. If im not mistaken `gallery.scenes` gets passed to `<SceneCard>` which needs paths/files/studio/etc. 

This would more than likely break various parts on other pages that rely on this. Can we get some testing on this please?

ui/v2.5/src/components/List/ListTable.tsx
```diff
@@ -102,12 +109,25 @@ export const ListTable = <T extends { id: string }>(
...
+    return visibleColumns.map((column) => {
+      const isSortable = column.sortable !== false && onSort;
+      const isActive = sortBy === column.value;
+
+      return (
+        <th
+          key={column.value}
+          className={`${column.value}-head${isSortable ? " sortable" : ""}`}
+          onClick={isSortable ? () => onSort(column.value) : undefined}
+          style={isSortable ? { cursor: "pointer" } : undefined}
```
Gykes: Would this be better in the SCSS class rather than inline? Sort direction should also probably compare against `GQL.SortDirectionEnum.Asc` rather than just a plain string.

Gykes (changes requested):
Static review, no actual testing done.

I think this PR covers 2 different things: 

1) Memoization, IntersectionObserver, and slim-fragment stuff.
2) A new sortable column header feature that is wired through `ListTable` and 5 components.

I think 2) should be stripped out into its own PR to reduce creep scope. I believe that forcing ASC would be jarring to users as each column should have a different default.

Gykes:
Just looking at the next PR it looks like that is the PR for your sortable stuff. Chances are you created a branch from a branch and that is what it is showing up in this PR.

slick-daddy:
> Just looking at the next PR it looks like that is the PR for your sortable stuff. Chances are you created a branch from a branch and that is what it is showing up in this PR.

I just realized this is correct, I will try to fix it.

slick-daddy:
Fixed and rebased.

>I believe that forcing ASC would be jarring to users as each column should have a different default.

List view feels very dry without sorting feature. There are many entities that can be sorted and we have to come up with a logic to be able to sort everything. Entities can be selected from dropdown menu, meaning users don't have lose their way while sorting. Also I think it adds flexibility to find specific things (e.g favourited performers). If we don't force a way, it will be impossible implement due to different data types (int, string, boolean).

ui/v2.5/graphql/data/gallery-slim.graphql
WithoutPants: It does indeed break the display of the scene title, which requires the `files` field.

I think this can be resolved by adding `files { path }`.

ui/v2.5/src/components/Galleries/GalleryCard.tsx
```diff
@@ -66,7 +64,7 @@ interface IGalleryCardProps {
   onSelectedChanged?: (selected: boolean, shiftKey: boolean) => void;
 }
 
-const GalleryCardPopovers = PatchComponent(
+const _GalleryCardPopovers = PatchComponent(
```
WithoutPants: I don't understand why you can't have `React.memo(PatchComponent(...))`?

ui/v2.5/src/components/Scenes/SceneCard.tsx
```diff
@@ -43,62 +43,66 @@ interface IScenePreviewProps {
...
-  const videoEl = useRef<HTMLVideoElement>(null);
-
-  useEffect(() => {
-    const observer = new IntersectionObserver((entries) => {
-      entries.forEach((entry) => {
-        if (entry.intersectionRatio > 0)
-          // Catch is necessary due to DOMException if user hovers before clicking on page
-          videoEl.current?.play()?.catch(() => {});
-        else videoEl.current?.pause();
+export const ScenePreview: React.FC<IScenePreviewProps> = React.memo(
```
WithoutPants: Adding `React.memo` made an absolute meal of this changeset. Can you separate each of your changes into its own commit so I make review each changeset more easily? This is a good practice to get into.

WithoutPants (changes requested):

WithoutPants (approved):

PR #6936, opened 2026-05-17
Add support for sorting images by performer's age

eugene-shovel:
Just like in the title, it closes the gap in this regard between scenes and images. Related to #2956

That's my first contribution so apologies in advance if I missed some crucial step. Everything's tested locally, of course.

### DIFF
--- a/pkg/sqlite/image.go
+++ b/pkg/sqlite/image.go
@@ -1032,12 +1032,15 @@ func (qb *ImageStore) setImageSortAndPagination(q *queryBuilder, findFilter *mod
 			addFolderJoin()
 			sortClause = " ORDER BY COALESCE(images.title, files.basename) COLLATE NATURAL_CI " + direction + ", folders.path COLLATE NATURAL_CI " + direction
 		case "performer_age":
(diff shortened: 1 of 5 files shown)


Gykes (changes requested):
Static review only, no actual testing was done. 

Overall looks fine. Please add test coverage to the `image_test.go` similar to `scene_test.go:4258`.

There is also minor code duplication but I think it is fine for now.

WithoutPants:
@eugene-shovel are you going to add unit tests per @Gykes comment?

eugene-shovel:
@WithoutPants yes, sorry for the delayed communication. I plan to finish it this or the next week and then hopefully move to more contributions.

Gykes (approved):
For the sake of simplicity I am going to keep the same logic for sorting by age as with scenes. If people prefer to fall back to actual performer age, and not calculated, then it's a very easy fix and can either be done in this PR before merge or a followup.I have updated the documents to reflect that images without a date will not be sorted properly due to how age is calculated. 

@DogmaDragon Please feel free to modify the documentation as you see fit.

PR #6941, opened 2026-05-19
fix: address clawpatch security and reliability findings

slick-daddy:
## Summary

Fixes several confirmed bugs and risk findings reported by clawpatch across the `phasher` CLI, `stash` CLI, and `Makefile`.

### `cmd/phasher/main.go`
- Return a clear error from `getPaths` when `ffmpeg` or `ffprobe` is not found in `PATH`, instead of silently passing empty strings downstream
- Guard against empty file extension in `printPhash` to prevent a runtime panic on files without an extension (e.g. `Makefile`, `README`)
- Exit with a non-zero code when any file fails during batch processing

### `cmd/stash/main.go`
- `initProfiling` now returns `(*os.File, error)` so the caller can close the file descriptor after `pprof.StopCPUProfile`; previously the fd was leaked for the entire process lifetime

### `internal/api/models.go`
- `convertBaseFile` now returns `(BaseFile, error)` instead of panicking on unknown file types, matching the pattern used by `convertVisualFile`
- `convertBaseFiles` updated accordingly

### `Makefile`
- Fixed `install` target: introduced `STASH_BINARY` to track the actual binary name separately from the `-o` flag, preventing broken installs on both Unix and Windows
- Added `generate-backend` as an explicit dependency of `build-release` to prevent race conditions under `make -j`

### `cmd/stash/main_test.go`
- Replaced empty `TestStub` with real tests for `exitError`, `recoverPanic`, and `initLogTemp`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,10 @@ ifdef OUTPUT
   PHASHER_OUTPUT := $(OUTPUT)
 endif
 ifdef STASH_OUTPUT
+  STASH_BINARY := $(STASH_OUTPUT)
   STASH_OUTPUT := -o $(STASH_OUTPUT)
+else
+  STASH_BINARY := stash
 endif
 ifdef PHASHER_OUTPUT
   PHASHER_OUTPUT := -o $(PHASHER_OUTPUT)
@@ -446,8 +449,8 @@ remove-compiler-container:
 install: build-release
 ifdef IS_WIN_SHELL
 	@if not exist "$(PREFIX)" mkdir $(PREFIX)
-	@copy "dist\\stash-win.exe" "$(PREFIX)\\stash-win.exe"
+	@copy "$(STASH_BINARY).exe" "$(PREFIX)\\stash.exe"
 else
 	@mkdir -p $(PREFIX)/bin
-	@install -m 755 $(STASH_OUTPUT) $(PREFIX)/bin/stash
+	@install -m 755 $(STASH_BINARY) $(PREFIX)/bin/stash
 endif
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ build: stash
 
 # builds dynamically-linked PIE release binaries
 .PHONY: build-release
-build-release: generate-backend flags-release flags-pie build
+build-release: flags-release flags-pie build
 
 # compile and bundle into Stash.app
 # for when on macOS itself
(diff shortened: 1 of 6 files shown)


cmd/stash/main.go
```diff
@@ -69,11 +69,13 @@ func main() {
...
 
 	if cpuProfilePath != "" {
-		if err := initProfiling(cpuProfilePath); err != nil {
+		f, err := initProfiling(cpuProfilePath)
+		if err != nil {
 			exitError(err)
 			return
 		}
 		defer pprof.StopCPUProfile()
+		defer f.Close()
```
WithoutPants, line 77: I think the order should be switched if I understand `defer` execution order correctly. We should stop profiling and then close the file handle, but `defer` statements are executed last in first out.

internal/api/models.go
```diff
@@ -28,28 +27,34 @@ func convertVisualFile(f models.File) (VisualFile, error) {
 	}
 }
 
-func convertBaseFile(f models.File) BaseFile {
+func convertBaseFile(f models.File) (BaseFile, error) {
```
WithoutPants: Is this change actually necessary? It's assumed that the input `f` is one of the types listed in the switch case, and any exception should be considered a programming error, which is more suitable for a `panic`.

Makefile
```diff
@@ -132,7 +135,7 @@ build: stash
 
 # builds dynamically-linked PIE release binaries
 .PHONY: build-release
-build-release: flags-release flags-pie build
+build-release: generate-backend flags-release flags-pie build
```
WithoutPants: I don't understand what this fixes. #6990 should have fixed the issues running `make -j x release`, and I don't see any current problems running `build-release`.

WithoutPants (changes requested):

internal/api/models.go
```diff
@@ -41,15 +39,18 @@ func convertBaseFile(f models.File) BaseFile {
...
 	}
 }
 
 func convertBaseFiles(files []models.File) []BaseFile {
-	return sliceutil.Map(files, convertBaseFile)
+	ret := make([]BaseFile, len(files))
+	for i, f := range files {
+		ret[i] = convertBaseFile(f)
+	}
+	return ret
```
WithoutPants: This does exactly what it already did but with more LOC. Why was this changed?

WithoutPants (changes requested):

internal/api/models.go
slick-daddy: Reverted.

WithoutPants (approved):

PR #6947, opened 2026-05-20
Attempt to start adding per-user agent playback

### LINKED ISSUE #4175: [Feature] Set Default live transcode or change priority
My WMV files seek and load faster when using DASH/HLS instead of the default mp4 live transcode

 A way to set the default transcode or reorder the list in the player.

 I checked the yaml config file as well and couldn't find anything.

valarauca:
Create a system to remember somebody's playback preferences, mostly by useragent.

It is a little crude but this should solve one of the primary pain-points of "_Why do I have to keep selecting HLS to play back videos on mobile_". 

Resolves https://github.com/stashapp/stash/issues/4175


valarauca:
Attempts to address https://github.com/stashapp/stash/issues/4175

ui/v2.5/src/components/ScenePlayer/source-selector.ts
```diff
@@ -123,6 +123,8 @@ class SourceSelectorPlugin extends videojs.getPlugin("plugin") {
 
       const loadSrc = this.sources[this.selectedIndex];
 
+      player.trigger("playbackSourceSelected", source);
```
Gykes: So the backed returns default: true but I don't see it being consumed anywhere?

WithoutPants:
Why would you persist this information in the database? 

This doesn't PR demonstrate good knowledge of the code base and as such this issue is probably not suited for a first-time contribution.

PR #6948, opened 2026-05-20
fix: strip surrounding quotes from Windows paths via shared helpers

### LINKED ISSUE #294: [Feature] [Windows] Make init flow compatible with quoted paths
The fields for entering folder paths for cache/generated/metadata should support quoted paths. The current situation is that providing quoted paths to use will generate an internal server error.

The reason for this is that, on Windows, the Shift+Context Menu for a folder allows you to **Copy As Path**, but this creates a quoted path.

image

slick-daddy:
Fixes #294

When users copy file paths from Windows and paste them into the library path input, the pasted text often includes surrounding double quotes (e.g. 'C:\path\generated'). These quotes cause the backend to fail with 'The filename, directory name, or volume label syntax is incorrect'.

Consolidate quote-stripping into `fsutil.StripPathQuotes` (Go) and `TextUtils.stripPathQuotes` (TS) instead of duplicating inline logic across resolvers and components.

Testing:
- Go to Library, click on "add directory"
- Paste a directory with quotes, e.g "C:\path\generated"
- It strips quotes and accepts


WithoutPants:
Please include what testing you've performed in the description.

slick-daddy:
Done.

ui/v2.5/src/components/Shared/FolderSelect/FolderSelect.tsx
```diff
@@ -44,15 +49,21 @@ const _FolderSelect: React.FC<IProps> = ({
   const debouncedSetDirectory = useDebounce(setPath, 250);
 
   function setInstant(value: string) {
+    const stripped = TextUtils.stripQuotes(value);
     const normalizedValue =
-      quotePath && value.includes(" ") ? TextUtils.addQuotes(value) : value;
+      quotePath && stripped.includes(" ")
+        ? TextUtils.addQuotes(stripped)
+        : stripped;
+    setDisplayPath(normalizedValue);
```
WithoutPants: This code change needs documentation.

WithoutPants (changes requested):

WithoutPants:
I'm not convinced that the UI changes are good. Let's just limit this to the server-side for the time being.

ui/v2.5/src/components/Shared/FolderSelect/FolderSelect.tsx
```diff
@@ -49,21 +44,15 @@ const _FolderSelect: React.FC<IProps> = ({
   const debouncedSetDirectory = useDebounce(setPath, 250);
 
   function setInstant(value: string) {
-    const stripped = TextUtils.stripQuotes(value);
     const normalizedValue =
-      quotePath && stripped.includes(" ")
-        ? TextUtils.addQuotes(stripped)
-        : stripped;
-    setDisplayPath(normalizedValue);
+      quotePath && value.includes(" ") ? TextUtils.addQuotes(value) : value;
```
slick-daddy, line 48: One minor cosmetic note: if you paste and save quoted path, the quotes will briefly appear in the paths list because the UI no longer strips them immediately. After refreshing the page, the list reflects the cleaned value from the server. The path is stored correctly either way.

WithoutPants: Yeah the stash configuration changes were good. I've restored them.

WithoutPants (approved):

PR #6951, opened 2026-05-21
Convert group aliases to list

### LINKED ISSUE #5593: [Feature] Multiple aliases for groups
Performers, studios and tags currently support multiple aliases, however groups do not. It would be great for them all to function in the same way.

babechan233:
## Description

Group aliases are currently stored as a single string on the `groups` table. This PR migrates them to a dedicated join table (`group_aliases`), matching the existing pattern used by performers, studios and tags.

**Changes:**
- New migration moves alias data to a `group_aliases` join table
- `Group.aliases` in the GraphQL schema changed to `[String!]`
- UI updated to use multi-value input; scrape dialog merges aliases
- Import supports old single-string format for backward compatibility

## Related Issue

Closes #5593 

## Testing

- [x] Added resolver and validate unit tests. All tests run successfully
- [x] Basic validation from UI of add/edit/delete
- [x] Searching utilizes aliases 
- [x] Data migration works seemlessly

## Screenshots

**Before:**
alias_before

**After:**
alias_after

## Checklist

## AI Usage Disclosure

- LLM assisted primarily for writing and fixing tests. Other changes largely done manually.
- All changes thoroughly reviewed by both LLM and myself.

## Additional Context

**Migration note:** I recommend that existing alias be added as-is without comma splitting, as group names can legitimately contain commas. Doing the comma-splitting ourselves adds burden and also messes up the data in case of genuine comma. Users who were storing comma-separated aliases will need to split them manually after upgrading though.


graphql/schema/types/movie.graphql
```diff
@@ -1,6 +1,7 @@
 type Movie {
   id: ID!
   name: String!
+  "Comma-delimited list of aliases"
```
babechan233: `Movie` has not been modified simply because they are to be deprecated. I will remove this comment.

internal/api/resolver_model_movie.go
```diff
@@ -228,3 +228,28 @@ func (r *groupResolver) CustomFields(ctx context.Context, obj *models.Group) (ma
...
+			return nil, err
+		}
+	}
+
+	return obj.Aliases.List(), nil
+}
+
+// movieResolver.Aliases overrides groupResolver.Aliases to return a single string
+// for backward compatibility with the deprecated Movie type.
+func (r *movieResolver) Aliases(ctx context.Context, obj *models.Group) (*string, error) {
```
babechan233, line 246: Only basic changes done to movie related codebase to ensure existing stuff doesn't break & code compiles correctly.

pkg/models/mocks/GroupReaderWriter.go
```diff
@@ -621,3 +621,26 @@ func (_m *GroupReaderWriter) UpdatePartial(ctx context.Context, id int, updatedG
 
 	return r0, r1
 }
+
+// GetAliases provides a mock function with given fields: ctx, relatedID
+func (_m *GroupReaderWriter) GetAliases(ctx context.Context, relatedID int) ([]string, error) {
```
babechan233: Copied from `StudioReaderWriter` / `PerformerReaderWriter`

pkg/sqlite/migrations/86_group_aliases.up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE `group_aliases` (
+  `group_id` integer,
+  `alias` varchar(255) NOT NULL,
+  foreign key(`group_id`) references `groups`(`id`) on delete CASCADE
+);
+
```
babechan233, line 7: multiple groups can have same aliases, this is not considered a problem, hence no `UNIQUE INDEX`.

pkg/sqlite/migrations/86_group_aliases.up.sql
Gykes, line 1: So, the biggest thing that I see with this is how do we want to handle people who split data? We just need to make a decision is all. 

If a user wants multiple aliases they COULD be doing like `abc, 123, XYZ`. Now, current implementation that would still be one alias but to the user it could mean multiple. Do we want to convert the data as a one to one or split data on on commas or something like that?

pkg/sqlite/migrations/86_group_aliases.up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE `group_aliases` (
+  `group_id` integer,
```
Gykes: This should probably be `NOT NULL`

pkg/sqlite/migrations/86_group_aliases.up.sql
```diff
@@ -0,0 +1,20 @@
...
+  `alias` varchar(255) NOT NULL,
+  foreign key(`group_id`) references `groups`(`id`) on delete CASCADE
```
Gykes, line 5: I think we need a `PRIMARY KEY` here as well.

pkg/sqlite/group.go
```diff
@@ -178,6 +178,16 @@ func (qb *GroupStore) Create(ctx context.Context, newObject *models.Group) error
 		return err
 	}
 
+	if newObject.Aliases.Loaded() {
+		if err := group.ValidateAliases(ctx, newObject.Aliases.List()); err != nil {
```
Gykes: I think validation would be better in the server layer along side the other group validators, no? I took a quick glance at performers and on `/pkg/performer/validate.go:53` it seems to be stored there.

Gykes (changes requested):
Didn't do a full review, just a quick glance while I have a little time to kill. 

Feel free to educate me on anything if I misunderstood your intent.

pkg/sqlite/migrations/86_group_aliases.up.sql
babechan233: Yes. I was thinking on the same. I think I should have mentioned this as a comment.

To be fair, for groups, I find the comma-separation to be brittle because group names are more generic than tags and performers and studios, I have actually run into scenarios where scraped movies (groups) have commas in their titles etc so they end up breaking.

That being said, how many folks are really using/relying on multiple aliases via comma separation? Not sure if SQL can do it but i guess we can do this in post migration if need be.

tl;dr , 2 options -
- split existing aliases on commas and populate
- populate existing data as a single alias (no splitting)

i favour the second one but i feel the element of least surprise for users would be the first one.

Gykes: My initial thoughts are similar to yours. 

Since this iteration of aliases were only designed for single use we should probably assume that users were doing that and any other way of doing it is unsupported. I think the second option is the "This is how it was designed" option and probably the best bet.

The first option adds complications as if there are `,` in the actual alias it would break the splitting.

babechan233: Yes. Should I add this discussion/detail as comments somewhere? Or in PR description? Do we have a way to communicate this information to users?

Gykes: You can update the PR description with the "This is what I think we should do". WP will ultimately have the final say but I doubt he will want to add the complication. It's not getting merged now or anything but users will be warned and DBs will be backed up automatically anyway when they migrate.

babechan233: updated! thanks!

babechan233:
s1

There is a problem though. We display aliases as comma-separated. Here, even though - B, C, AA - all 3 are different aliases, they are displayed as comma separated. This may be a visual problem when displaying group names containing commas.

s2

Clearly here `A,B` is a single alias but since the aliases list is  comma separated, it looks more like 2 different aliases.

pkg/sqlite/migrations/86_group_aliases.up.sql
```diff
@@ -0,0 +1,21 @@
...
+INSERT INTO `group_aliases` (
+    `group_id`,
+    `alias`
+) SELECT
+    `id`,
+    `aliases`
+FROM `groups`
+WHERE `aliases` IS NOT NULL AND TRIM(`aliases`) != '';
+
+ALTER TABLE `groups` DROP COLUMN `aliases`;
```
Gykes, line 21: Here you are using `!=''`. I'm pretty sure `SELECT` would take the raw value. If the user has leading or trailing whitespace would it not get regected by the `ValidateAliases`? 

Off the top of my head I cant remember if we removed whitespaces from Aliases. I remember doing it for various other things a while ago. Would need some verification.

ui/v2.5/src/components/Groups/GroupDetails/GroupEditPanel.tsx
```diff
@@ -71,7 +72,7 @@ export const GroupEditPanel: React.FC<IGroupEditPanel> = ({
 
   const schema = yup.object({
     name: yup.string().required(),
-    aliases: yup.string().ensure(),
+    aliases: yupRequiredStringArray(intl).defined(),
```
Gykes, line 75: I think this would cause duplicate aliases to get deleted by the backend with `UniqueExcludeFold` on save but would never warn the user. It should either be moved over to `YupUniqueStringList(intl)` or have it in the logs/document that duplicate aliases are not supported.

Gykes (changes requested):
Another quick review. Please update any documentation that would be required as well. 

EDIT: Per your UI question. I cant think of a good way around it. Users should probably just understand what their aliases are and know what it ascually means. Outside of wrapping aliases in quotes it just seems like a small UI thing people will need to get used to.

pkg/sqlite/migrations/86_group_aliases.up.sql
babechan233: For both studios and performers, the aliases are cleaned up, ie, leading & trailing whitespaces removed and duplicates removed.

I will need to check a bit if we really need it in the db query though.

ui/v2.5/src/components/Groups/GroupDetails/GroupEditPanel.tsx
babechan233: Duplicate aliases dont seem to offer any value, atleast to me. Having X twice is the same as having X once - primary usecases of aliases are display and search, both of which need not care about duplicates,

Further, studios and performers handle this the same way - by silently making the duplicates go away, without any user warning.

babechan233: #6514 decided that duplicates should be silently consumed without any warning to the user.

.mockery.yml
```diff
@@ -1,4 +1,27 @@
...
+      StudioReaderWriter:
+      TagReaderWriter:
+      SavedFilterReaderWriter:
+  github.com/stashapp/stash/internal/manager:
+    config:
+      dir: ./internal/manager/mocks
+      filename: "{{.InterfaceName}}.go"
+      outpkg: mocks
+    interfaces:
+      GroupService:
```
babechan233, line 27: More service interfaces exist but mocks for those can be generated as and when required.

go.mod
```diff
@@ -55,7 +55,7 @@ require (
 	github.com/vearutop/statigz v1.4.0
 	github.com/vektah/dataloaden v0.3.0
 	github.com/vektah/gqlparser/v2 v2.5.27
-	github.com/vektra/mockery/v2 v2.10.0
+	github.com/vektra/mockery/v2 v2.21.6
```
babechan233, line 58: Updated to use `packages` field in `.mockery.yml`

babechan233 (commented):
- Update docs (only single place)
- Updated mockery to allow service layer mocks generation
- Added resolver test for aliases

pkg/sqlite/migrations/86_group_aliases.up.sql
babechan233: Updated, `TRIM` is not really required. Only DB constraint we have is non-null. Others are small issues which will be fixed on next group update anyhow.

babechan233: `ValidateAliases` only runs during create/update. Not during read. During reads, UI will show the aliases without being trimmed. Not an issue in my opinion.

ui/v2.5/src/docs/en/Manual/Browsing.md
```diff
@@ -10,9 +10,9 @@ The text field allows you to search using keywords. Keyword searching matches on
 |------|-----------------|
 | Scene | Title, Details, Path, OSHash, Checksum, Marker titles |
 | Image | Title, Details, Path, Checksum |
-| Group | Title |
 | Marker | Title, Scene title |
 | Gallery | Title, Path, Checksum |
+| Group | Name, Aliases |
```
DogmaDragon: Why the order change? Before it aligned with the order of menu items, which was intuitive.

babechan233: I see, my bad. I ordered it so that all "Name" users would be together. Will change.

DogmaDragon (approved):
Documentation check passed.

internal/api/testutil/gql.go
```diff
@@ -0,0 +1,40 @@
+package testutil
```
babechan233, line 1: This file was completely generated by LLM. I understand what it is doing but I dont understand if/whether the solution it utilizes is idiomatic or even the "right" solution. Will request greater reviewer attention here.

graphql/schema/types/group.graphql
```diff
@@ -7,7 +7,7 @@ type GroupDescription {
 type Group {
   id: ID!
   name: String!
-  aliases: String
+  aliases: [String!]
```
WithoutPants, line 10: This change breaks compatibility with existing external systems. You will need to deprecate the existing `aliases` field and add a new `alias_list` field instead. The existing field will need a resolver to build the string value.

We did the same with performer aliases before we removed the deprecated field. See https://github.com/WithoutPants/stash/blob/f65e87773c4b9fb053f4c5f23bcb77c03adbbccd/internal/api/resolver_model_performer.go#L21

Will need to apply this to the other alias field changes.

internal/api/resolver_model_movie.go
WithoutPants: This can be repurposed to be on `groupResolver` when you update the graphql fields.

internal/api/testutil/gql.go
WithoutPants: I'd remove this and the unit test. We do need to add testing to the graphql layer, but I think it's probably better to do it with integration testing and not something within this scope.

pkg/sqlite/migrations/86_group_aliases.up.sql
WithoutPants: For migration 42 where we changed performer aliases from string to string list, we had a post-migration to separate comma-delimited aliases into alias entries. It should be applied here as well. 

I think adding `!= ''` to the insert where clause is a good sanity check. Perhaps not necessary, but it doesn't hurt.

pkg/sqlite/group.go
```diff
@@ -730,3 +748,7 @@ func (qb *GroupStore) sortByOCounter(direction string) string {
 	// need to sum the o_counter from scenes and images
 	return " ORDER BY (" + selectGroupOCountSQL + ") " + direction
 }
+
+func (qb *GroupStore) GetAliases(ctx context.Context, groupId int) ([]string, error) {
```
WithoutPants, line 752: Please add a unit test to exercise this function.

pkg/sqlite/group_filter.go
```diff
@@ -131,9 +131,12 @@ func (qb *groupFilterHandler) missingCriterionHandler(isMissing *string) criteri
 			case "tags":
 				groupRepository.tags.leftJoin(f, "tags_join", "groups.id")
 				f.addWhere("tags_join.group_id IS NULL")
+			case "aliases":
```
WithoutPants, line 134: Please add unit test cases to `TestGroupQuery` to cover this.

.mockery.yml
WithoutPants: These changes should largely not be necessary if we remove the unit test from the api package.

WithoutPants (changes requested):

slick-daddy:
Branch has merge conflicts.

PR #6952, opened 2026-05-21
Run Docker image as configurable non-root user

### LINKED ISSUE #684: [Feature] non-privileged user in Docker build

Kausha3:
## Summary
- add a production Docker entrypoint that chowns mounted app directories and drops to the configured PUID/PGID before running stash
- keep the existing /root/.stash config path for compatibility while making HOME writable for scraper/plugin cache use
- document PUID/PGID in the production Docker README and compose example

Fixes #684

## Verification
- Built the production Dockerfile with a dummy linux/amd64 stash binary from a dist-style context:
  `docker build --platform linux/amd64 -f docker/ci/x86_64/Dockerfile /private/tmp/stash-docker-test -t stash-nonroot-test`
- Ran default UID/GID check:
  `docker run --rm stash-nonroot-test` -> `uid=1000 gid=1000 home=/root config=/root/.stash/config.yml writable=yes`
- Ran custom UID/GID check:
  `docker run --rm -e PUID=1234 -e PGID=1235 stash-nonroot-test` -> `uid=1234 gid=1235 home=/root config=/root/.stash/config.yml writable=yes`

### DIFF
--- a/docker/ci/x86_64/Dockerfile
+++ b/docker/ci/x86_64/Dockerfile
@@ -13,7 +13,33 @@ FROM --platform=$TARGETPLATFORM alpine:latest AS app
 COPY --from=binary /stash /usr/bin/
 
 RUN apk add --no-cache ca-certificates python3 py3-requests py3-requests-toolbelt py3-lxml py3-pip ffmpeg tzdata vips vips-tools vips-heif \
-    && pip install --break-system-packages mechanicalsoup cloudscraper stashapp-tools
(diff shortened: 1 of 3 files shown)


ghost:
+1

feederbox826:
Default nonroot is a breaking change. This can already be accomplished without a fake "nonroot" container. The container still has access to root privileges and this will break nonroot deployments like in k8s

PR #6954, opened 2026-05-21
docker: run production image as non-root with configurable PUID/PGID

### LINKED ISSUE #684: [Feature] non-privileged user in Docker build

liangtovi-debug:
## Summary
- run the production container with a non-root user by default
- add `PUID`/`PGID` support in the image entrypoint so host file ownership is controllable
- move the default config location from `/root/.stash` to `/config` for non-root compatibility
- update production compose + docs to reflect the new defaults

Fixes #684.

## Validation
- `sh -n docker/build/x86_64/docker-entrypoint.sh`
- `docker compose -f docker/production/docker-compose.yml config`

## Notes
- explicit runtime `--user` overrides remain supported (entrypoint bypasses user creation when already non-root)

### DIFF
--- a/docker/build/x86_64/Dockerfile
+++ b/docker/build/x86_64/Dockerfile
@@ -37,8 +37,11 @@ RUN make flags-release flags-pie stash
 
 # Final Runnable Image
 FROM alpine:latest
-RUN apk add --no-cache ca-certificates vips-tools ffmpeg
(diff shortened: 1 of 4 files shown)


feederbox826:
default nonroot is a breaking change. This can already be accomplished without a fake "nonroot" container. The container still has access to root privileges and this will break nonroot deployments like in k8s

PR #6961, opened 2026-05-23
Fix: Warped Image Aspect Ratios in Lightbox on Safari

### LINKED ISSUE #5087: Warped image aspect ratios in lightbox on iOS/Safari
Several people have reported incorrect image aspect ratios when opening images in the lightbox on iOS/Safari devices. The images appear warped horizontally. 

When opening the image files outside the stash lightbox on those devices, the image is normal.

Relevant discord messages:
https://discord.com/channels/559159668438728723/1192901861053837583/1256620987923955743
https://discord.com/channels/559159668438728723/1192901861053837583/1256845471473995857
https://discord.com/channels/559159668438728723/1192901861053837583/1258307948132831284
https://discord.com/channels/559159668438728723/1192901861053837583/1263769456375627808

void-function865:
# Scope

Safari has a rendering bug where `transform: scale` applied to an `` element with very large intrinsic dimensions (e.g. 10246×13662) renders the image content distorted, even though the element's bounding box reports the correct portrait size. The lightbox computes `defaultZoom` to fit natural-size images into view via `transform: scale` on the `<picture>` wrapping the img, which triggers this bug on macOS and iOS Safari.

The fix moves the `transform` from the `<picture>` element onto a wrapping `<div>` so the scale never targets a `<picture>`/`` chain. The img keeps its natural dimensions inside the wrapper, and Safari renders it correctly. The change also drops the unused `<source srcSet>` (it pointed at the same URL as the fallback `` and provided no responsive behavior) and the dead `picture > div` CSS rule that had no matching element. `object-fit: contain` is removed from the img — it interacted badly with the wrapper transform in Safari and was a no-op anyway since the img has no CSS-constrained dimensions.

## Closes/Fixes Issues

Fixes #5087

## Other testing QA Notes

Verified in Safari (macOS), Chrome, and Firefox on a real gallery with a 10246×13662 portrait image — aspect ratio renders correctly in all three. The fix should also resolve the iOS Safari case originally reported in the issue, since it's the same underlying browser bug.

`make validate-ui-quick` passes (eslint, stylelint, prettier all clean).


Gykes (changes requested):
Static review.No actual testing was done.

Overall looks okay just a quick question or two. 

- Did you test on small/normal images as well to make sure there was no weird regression?

- Can you please add a comment above the div wrapper explaining the non obvious choice? Something like "Safari has a known bug this is a required workaround". Doesn't need to be exactly that but just something along those lines.

void-function865:
Thanks for the review @Gykes!

Both points addressed:

- **Comment added**: pushed an amend with an inline JSX comment above the wrapper div explaining the Safari workaround and linking to #5087.
- **Regression check on small/normal images**: tested in Safari, Chrome, and Firefox with regular-sized images in addition to the original 10246×13662 image — no issues in any browser.

Gykes (approved):
LGTM.

Tested on develop and was able to reproduce your bug
With fix confirmed that bug is no longer present. 
Attempted to use smaller images and various browsers and all worked as expected.

PR #6965, opened 2026-05-25
Create gclib-based docker images for python compatiblity

lostLegion15:
## Main Issue

This has been an issue I've been dealing with for sometime, but have kind of kicked it down the road until in the last month I decided to take a look at it.

All of this is specifically around docker images (non CUDA ones).

There are a number of community plugins that fail when you attempt to run them, they call `PythonDepManager`. Two things seem to consistently fail across multiple plugins:

1. `git` is not installed in the host container, so you have to manually go into the container and add it via `apk add git`, or add this to the start-up script
2. Once that is installed, when `PythonDepManager` attempts to install packages it fails due to python packages being incompatible with alpine's `musl`, as there isn't always pre-compiled versions of those packages (python's PyTorch is a good example of this, `torch`).

Screenshot_20260525_155137

To attempt to solve this problem this PR moves the base deployed docker image and backend go compilation to debian-based containers, which use `gclib` instead of `msul`.

After building the container with the below when I run the install it via the community plugins it runs without me having to jump into the container and run any manual commands:

Screenshot_20260525_154710

## Additional Drivers

A common request is drivers for Intel/AMD (since there's already a separate CUDA-container). This change also installs the AMD/Old Intel drivers (`mesa-va-drivers`) and installs the latest "non-free" Intel drivers as well for modern Intel GPU compatibility, `intel-media-va-driver-non-free`.

I totally understand if we want those specific installs removed from the base docker image and we can let the individual users auto-run install commands of whatever drivers they need after the container starts.

With this specific change if you map the appropriate `/dev/dri` and/or `/dev/kfd` mounts into the container, `ffmpeg` picks up the changes.

## Additional Notes

This PR just adds in a new debian-based option to avoid breaking any existing images that may have `apk`-based commands that run on container start, etc.

I'd like to also possibly publish the debain-based one, but for right now this acts like the CUDA one where it's an option but isn't published to the docker repository.

## Disclosures

I used local LLMs to assist in this change. I'm semi-unsure if the makefile changes were necessary but I couldn't get the `make docker-build` to work without adding in that additional makefile combination.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -371,6 +371,10 @@ endif
 .PHONY: ui
 ui: pre-ui generate ui-only generate-login-locale
 
+.PHONY: ui-build
+ui-build: ui-env pre-ui generate-ui
+	cd ui/v2.5 && npm run build
+
 .PHONY: ui-only
 ui-only: ui-env generate ui
 	cd ui/v2.5 && npm run build
@@ -423,6 +427,11 @@ validate: validate-ui validate-backend
 .PHONY: docker-build
 docker-build: build-info
 	docker build --build-arg GITHASH=$(GITHASH) --build-arg STASH_VERSION=$(STASH_VERSION) -t stash/build -f docker/build/x86_64/Dockerfile .
+
+# locally builds and tags a 'stash/build-debian' docker image (debian-based, improved non-free repo setup)
+.PHONY: docker-build-debian
+docker-build-debian: build-info
+	docker build --build-arg GITHASH=$(GITHASH) --build-arg STASH_VERSION=$(STASH_VERSION) -t stash/build-debian -f docker/build/x86_64/Dockerfile-debian .
 
 # locally builds and tags a 'stash/cuda-build' docker image
 .PHONY: docker-cuda-build
--- a/docker/build/x86_64/Dockerfile-debian
+++ b/docker/build/x86_64/Dockerfile-debian
@@ -1,0 +1,57 @@
+# This dockerfile should be built with `make docker-build-debian` from the stash root.
+# Debian-based variant with improved non-free repo setup for Intel GPU drivers.
+
+# Build Frontend
+FROM node:24-alpine AS frontend
+RUN apk add --no-cache make git
+## cache node_modules separately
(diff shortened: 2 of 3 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/haven-vlm-connector/5464/18

feederbox826:
This has been brought up before and ultimately, it's very scattered to make multiple images and variants. It would be much better rolled up into the CUDA image, as well as node should be switched to trixie as bookworm is now oldstable. We also target python 3.14 for Community* plugins. 

At the end of the day this seems to be a customized image for your own needs. I suggest you just make your own image with stashapp/stash as the base, it'll be much easier to customize or rip the binary out yourself.

lostLegion15:
@feederbox826

To be clear, beyond the GPU driver additions, this was not "customized for my own needs". It was to ensure that the stash container image supports all of the community plugins available from a clean install (which is currently not the case, as evidenced by the above screenshots).

Currently you can install a clean official docker image of stash and some of the community plugins will fail to install due to the gclib incompatibility with python binaries (shown in the screenshots). If we're not willing to address that incompatibility by updating the base image to support those plugins, then we should also remove all plugins that are incompatible with the official image as well?

Seems strange to say "here's the official image", you install it, and then go to the community plugins (that's already there from a clean install), install one, and it fails to actually work because the official image is incompatible with the community plugins?

I'm not saying all community plugins must be compatible with the alpine-based image. "3rd party" repos for plugins should not be required to work on the base docker image. I'm saying anything that comes pre-installed with stash's docker image should work, without further image modifications.

lostLegion15:
Also, in case it wasn't clear from the PR, the Python version was a red herring in trying to figure out what was going on. The issue was not that the python dependencies for those libraries didn't have 3.14 compatible versions, but they had no alpine-based, `musl` pre-compiled binaries for any python version. `gclib`-based distros have much higher levels of compatibility with python binaries, so this was an attempt to slowly shift the base image from `alpine` based docker images to `debian` to ensure full compatibility with "out of the box" community plugins that are available on clean installs.

PR #6967, opened 2026-05-25
New: Make StudioPage, TagPage, GroupPage Patchable

### LINKED ISSUE #6846: Wrap TagPage / GroupPage / StudioPage in PatchComponent for plugin symmetry

`PerformerPage` is wrapped in `PatchComponent("PerformerPage", ...)`, giving plugins a first-class extension point for the performer detail page via the patch system (`PluginApi.patch.before/instead/after`). The sibling detail-page components for other entity types are NOT wrapped:

  - `ui/v2.5/src/components/Tags/TagDetails/Tag.tsx` — plain `React.FC`, not wrapped.
  - `ui/v2.5/src/components/Groups/GroupDetails/Group.tsx` — plain `React.FC`, not wrapped.
  - `ui/v2.5/src/components/Studios/StudioDetails/Studio.tsx` — plain `React.FC`, not wrapped. (Note: the inner `StudioDetailsPanel` IS wrapped at `Studios/StudioDetails/StudioDetailsPanel.tsx:16`, so the pattern is already partially in place for studios.)

This asymmetry means plugins that want to extend tag / group / studio detail pages the same way they extend performer pages have no React-level hook. Our workarounds fall back to `MutationObserver` against DOM selectors, which is fragile across Stash UI iterations and fights React reconciliation during edit-mode mount/unmount.

Plugins gain access to the same extension mechanism they have for performer pages:
  - patch.instead to rebuild the detail page entirely (e.g. custom layouts, inline edit forms).
  - patch.before / patch.after to augment the existing render (e.g. inject panels, modify props).

Without this, plugins must use DOM-scraping approaches, which:
  - Break silently when internal class names change.
  - Can't cleanly intercept edit-mode state transitions.
  - Require per-plugin MutationObserver machinery that duplicates across plugins.

Impact
  - Core behavior: zero. The PatchComponent wrapper is a transparent Proxy when no patches are registered (ui/v2.5/src/patch.tsx).
  - Bundle size: negligible (one additional function call per page module).
  - Existing plugins: no change

Wrap each page component in `PatchComponent`, matching the `PerformerPage` pattern. Single-line change per file:

  ```tsx
  // Before (Tag.tsx, Group.tsx, Studio.tsx)
  const TagPage: React.FC<IProps> = (props) => {
    // existing implementation
  };

  // After
  const TagPage: React.FC<IProps> = PatchComponent("TagPage", (props) => {
    // existing implementation
  });
```
Same edit for GroupPage and StudioPage. No behavioral change to core — the Proxy wrapper is transparent when no plugins have registered patches for those names.

### Additional context

I've been working on a UI adjustment plugin, and the individual Performer Detail page was wrapped with PatchComponent so I was sort of able to rebuild it using a plugin.  I'd like to do the same across other detail pages to make it uniform, I think that would be a solid addition to the great work already done in Stash.

ordureconnoisseur:
## Summary

Wraps the three remaining detail-page components — `StudioPage`, `TagPage`, `GroupPage` — in `PatchComponent`, so plugins can use `PluginApi.patch.{before,instead,after}` on them. This matches the pattern already used for `PerformerPage` (#5703) and `ScenePage`.

Fixes #6846. Also addresses the corresponding ask in the umbrella #4510. cc @ezgoing269 @DogmaDragon

## Changes

- `Studio.tsx`, `Tag.tsx`, `Group.tsx`: wrap each page component in `PatchComponent("<Name>", ...)`
- `pluginApi.d.ts`: add `StudioPage`, `TagPage`, `GroupPage` to the components type list
- `UIPluginApi.md`: add the three new entries to the patchable components list

The large diff size is from re-indenting the function bodies one level deeper (same as PR #5703 did for `PerformerPage`); the semantic change per file is ~5 lines. `PatchComponent` is a transparent Proxy when no patches are registered, so core behavior is unchanged.

### DIFF
--- a/ui/v2.5/src/components/Groups/GroupDetails/Group.tsx
+++ b/ui/v2.5/src/components/Groups/GroupDetails/Group.tsx
@@ -44,6 +44,7 @@ import { GroupSubGroupsPanel } from "./GroupSubGroupsPanel";
 import { GroupPerformersPanel } from "./GroupPerformersPanel";
 import { Icon } from "src/components/Shared/Icon";
 import { goBackOrReplace } from "src/utils/history";
+import { PatchComponent } from "src/patch";
 
 const validTabs = ["default", "scenes", "performers", "subgroups"] as const;
 type TabKey = (typeof validTabs)[number];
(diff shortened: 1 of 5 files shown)


Gykes (approved):
LGTM. Pretty small PR. If WP doesn't get a look at it I will probs merge this in within the next few days.

PR #6971, opened 2026-05-27
New: Remove OSHASH From Submissions and Queries

Gykes:
Some chat has been going around about potentially wanting to remove OSHASH from StashDB as it's a somewhat useless hash since so many people transcode and just causes bad matching.

This PR removes OSHASH submissions from the submit fingerprints and removes the OSHASH query from scraping and tagging.

Testing done: 
I added debug lines and ran scrapes and then submissions
- Scrape results:  26-05-27 16:12:31Debug   stash-box query fingerprint: algorithm=PHASH hash=bf1ac88ccdb36252

- Submit results: 26-05-27 16:13:58Debug   stash-box submit fingerprint: sceneID=0917cc21-61fb-4f56-a17a-5955337c24ae algorithm=PHASH hash=bf1ac88ccdb36252

Currently the only Hashes being scraped/tagged and submitted is PHASH

Stashbox convo: https://github.com/stashapp/stash-box/issues/1115

### DIFF
--- a/pkg/stashbox/scene.go
+++ b/pkg/stashbox/scene.go
@@ -45,7 +45,7 @@ func (c Client) QueryScene(ctx context.Context, queryStr string) ([]*models.Scra
 }
 
 // FindStashBoxScenesByFingerprints queries stash-box for a scene using the
-// scene's MD5/OSHASH checksum, or PHash.
+// scene's MD5 checksum, or PHash.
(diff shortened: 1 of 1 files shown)


DogmaDragon:
I think this is a wrong approach. As was mentioned by @InfiniteStash on Discord, it would be better handled by stash-box instances deciding which fingerprints types they want to accept instead.

Gykes:
This was put up when "remove" was still the idea. The idea then expanded to the whole handshake situation. This will solve the former but not the latter, mostly just a design decision at this point.

PR #6981, opened 2026-06-01
Withdrawn

### LINKED ISSUE #3530: [Feature] Support multiple Scenes in a single file

AlonePenguin:
Withdrawn by author.

### DIFF
--- a/graphql/schema/types/scene.graphql
+++ b/graphql/schema/types/scene.graphql
@@ -60,6 +60,12 @@ type Scene {
   resume_time: Float
   "The total time a scene has spent playing"
   play_duration: Float
+  "Start time within the primary file, in seconds"
+  start_time: Float
+  "End time within the primary file, in seconds"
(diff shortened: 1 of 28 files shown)


WithoutPants:
This is not how I want to implement this feature, nor is this a suitable issue for a first-time contributor.

PR #6986, opened 2026-06-02
New: Fingerprint Submission Queue (Batching)

Gykes:
It has been a long standing issue that, when you have a lot of fingerprints, it runs longer than the timeout then they get lost/removed and not submitted. This iteration of this fix modifies the fingerprint submission to use the new StashBox submission mutation. Currently it is hard coded at 50 at a time. If someone could educate me on a better number, as 50 is arbitrary, then I would love to see the science behind it and we can raise/lower the limit as needed. I think in the future we could make this variable an option in the UI that allows users to increase/decrease as they see fit. 

Testing done (not too much):
- Built, messed around with UI
- Submitted fingerprints to Stashbox and recieved the appropriate toast message and logs.
- Mostly used built in tests to test the new submitting function. 

Sadly I don't have an ass load of fingerprints in my dev version that I can do this with and i'm not too sure how to do it. I could potentially copy over my 50k or so fingerprints from my main machine. I guess we can see if the code is any good and I can run that if needed.

### DIFF
--- a/graphql/schema/schema.graphql
+++ b/graphql/schema/schema.graphql
@@ -561,9 +561,7 @@ type Mutation {
   stopAllJobs: Boolean!
 
   "Submit fingerprints to stash-box instance. Returns the ID of the job that performs the submission."
-  submitStashBoxFingerprints(
-    input: StashBoxFingerprintSubmissionInput!
-  ): ID!
+  submitStashBoxFingerprints(input: StashBoxFingerprintSubmissionInput!): ID!
 
   "Submit scene as draft to stash-box instance"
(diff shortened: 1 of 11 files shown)


WithoutPants:
Is this not already superceded by https://github.com/stashapp/stash/pull/6782?

Gykes:
No, their version was done before the new stashbox batch submission were put into place. That one is also a lot more verboose and was a refactor of various other things. Mine is more pointed to a specific pain point without doing a whole refactor.

Gykes:
Converting to draft as I need to add batch size and timeout configuration.

PR #6988, opened 2026-06-02
Fix: Handling UTF-8 BOM in Funscript Parsing

CynicalAtropos:
## Summary

Fixes funscript parsing when a `.funscript` file is encoded as UTF-8 with BOM.

Go's JSON decoder rejects a leading BOM before the opening `{`, causing heatmap generation to fail with:

`invalid character 'ï' looking for beginning of value`

This trims a leading UTF-8 BOM at the funscript JSON unmarshalling boundary and applies the same behavior to both heatmap generation and funscript CSV conversion.

## Changes

- Added shared funscript unmarshalling helper that trims a UTF-8 BOM before `json.Unmarshal`
- Updated both funscript load paths to use the helper
- Added coverage for BOM-prefixed funscript files

## Testing

- `mingw32-make fmt`
- `go test ./internal/manager`
- `golangci-lint run`
- Tested on both working UTF8 and previously erroring UTF8BOM funscripts to confirm that heatmap generation still works as expected.

## Notes

As with my last PR, I used Codex to help investigate, review my code/check for go specific issues, and also write the test file.

### DIFF
--- a/internal/manager/generator_interactive_heatmap_speed.go
+++ b/internal/manager/generator_interactive_heatmap_speed.go
@@ -48,6 +48,12 @@ type Action struct {
 	Speed float64
 }
 
+var utf8BOM = []byte{0xEF, 0xBB, 0xBF}
+
+func unmarshalFunscriptData(data []byte, funscript *Script) error {
+	return json.Unmarshal(bytes.TrimPrefix(data, utf8BOM), funscript)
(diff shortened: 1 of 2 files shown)


Gykes (approved):
Static review looks good. I don't have a good way to test this in reality.

PR #6992, opened 2026-06-03
chore: bunch of security fixes

slick-daddy:
- SQL injection fix: Parameterize stash-box endpoint values in stashIDCriterionHandler and stashIDsCriterionHandler to prevent SQL injection via crafted endpoint strings.
- Regex hoisting: Move regexp.MustCompile calls to package level in xpath.go, url.go, and image.go to avoid recompilation on every call.

### DIFF
--- a/pkg/scraper/url.go
+++ b/pkg/scraper/url.go
@@ -27,6 +27,8 @@ import (
 
 const scrapeDefaultSleep = time.Second * 2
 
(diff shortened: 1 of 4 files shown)


babechan233:
[not a review] man sqli in these times is crazy :(

WithoutPants:
> Observer leak fix: Add cleanup and dependency array to IntersectionObserver in ScenePreview to prevent leaked observers on every render.

You already did this in #6935.

slick-daddy:
Dropped observer leak fix since it overlaps with another PR. The rest is exactly same and should work as intended.

WithoutPants (approved):

PR #6995, opened 2026-06-03
Add AI usage policy, update contributing section

DogmaDragon:
- Added reference to `ARCHITECTURE.md` in `README.md`.
- Updates badges to display related logo.
- Added badge for Codeberg Translate.
- Refactored `README.md` for better consistency.
- Replaced StashDB related link with Stash documentation.
- Added `AI_POLICY.md` document.
- Updated `CONTRIBUTING.md` document to better .
- Removed old and added a new pull request template.

PR #6998, opened 2026-06-04
Use file checksum for image URL cache-busting

void-function865:
## Description

Image URLs (`paths.image`, `paths.thumbnail`, `paths.preview`) append a cache-buster derived from the image's `updated_at`. That value bumps on *any* metadata edit — rating, o-counter, organized, tags — none of which change the image bytes. Because the URL changes, the browser re-downloads the identical image. It's most visible in the gallery lightbox: setting a rating makes the current image visibly reload/flicker (the carousel also uses the path as its React `key`, so the element remounts), but it affects image/thumbnail caching everywhere (grids, detail pages, wall view).

This changes the cache-buster to use the primary file's checksum instead, so the URL is stable across metadata edits and only changes when the file content changes — preserving the original anti-stale intent of the timestamp suffix. When an image has no primary file the checksum is empty, in which case it falls back to `updated_at`.

Scenes and galleries are intentionally left on `updated_at`: their thumbnails can change without the underlying content changing (e.g. reassigning a gallery cover), so a content-based key would be wrong for them. They use separate URL builders, so this change is scoped to images only.

## Related Issue

Discussed and agreed approach in #6975 (@WithoutPants  confirmed checksum-based cache-busting is fine for images, and flagged the scenes/galleries distinction that this PR respects).

## Testing

- Added unit tests for the image URL builder (`internal/api/urlbuilders/image_test.go`) covering: checksum present, empty-checksum fallback to `updated_at`, and the full URL shape for `GetImageURL`/`GetThumbnailURL`.
- Built the full app (UI + backend) and verified against a real library:
  - Image URL is now `/image/1/image?t=<md5 checksum>` instead of a timestamp.
  - Setting a rating bumps `updated_at` but the image/thumbnail URL stays byte-for-byte
    identical, so the browser keeps its cached copy and the image no longer reloads.
  - The image/thumbnail endpoints still serve correctly (the server ignores the `?t=`
    value; it is purely a client cache key).

Note: image thumbnail/preview *regeneration* never bumped `image.updated_at` (the generate tasks only write the file to disk; unlike the scene screenshot/heatmap tasks they don't call `UpdatePartial`), so a force-regenerated preview already served from the client cache before this change. This PR preserves that existing behavior rather
than introducing a regression.

## Checklist

- [x] I have read and understood the [Contributing](docs/CONTRIBUTING.md) document.
- [x] I have read and understood the [AI Usage Policy](docs/AI_POLICY.md) document.

## AI Usage Disclosure

I used Claude (Anthropic) as an assistant during this work: to help trace the call paths in the codebase, to draft this description, and to run the verification steps above. I reviewed the change, understand every line, and can explain and defend the design decisions in review. The manual testing described above was performed and confirmed against a real library.

## Additional Context

Possible follow-up (not in this PR): the lightbox carousel keys items by `image.paths.image`. With this fix the path is stable across edits so it no longer churns, but `key={image.id}` would be more semantically correct. Happy to do it here or as a separate change if preferred.

### DIFF
--- a/internal/api/urlbuilders/image.go
+++ b/internal/api/urlbuilders/image.go
@@ -24,17 +24,30 @@ func NewImageURLBuilder(baseURL string, image *models.Image) ImageURLBuilder {
 	}
 }
 
+// cacheBuster returns the value used to bust client-side caches. The image
+// content is immutable for a given file, so the checksum of the primary file
+// is preferred: this keeps the URL stable across metadata edits (rating,
+// o-counter, tags, etc.) so the browser can reuse its cached copy. When no
+// primary file is present the checksum is empty, in which case we fall back to
+// the updated timestamp.
+func (b ImageURLBuilder) cacheBuster() string {
+	if b.Checksum != "" {
+		return b.Checksum
+	}
+	return b.UpdatedAt
+}
+
 func (b ImageURLBuilder) GetImageURL() string {
-	return b.BaseURL + "/image/" + b.ImageID + "/image?t=" + b.UpdatedAt
+	return b.BaseURL + "/image/" + b.ImageID + "/image?t=" + b.cacheBuster()
 }
 
 func (b ImageURLBuilder) GetThumbnailURL() string {
-	return b.BaseURL + "/image/" + b.ImageID + "/thumbnail?t=" + b.UpdatedAt
+	return b.BaseURL + "/image/" + b.ImageID + "/thumbnail?t=" + b.cacheBuster()
 }
 
 func (b ImageURLBuilder) GetPreviewURL() string {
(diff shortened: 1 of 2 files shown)


DogmaDragon:
Please use the current pull request template.

void-function865:
> Please use the current pull request template.

@DogmaDragon I switched to the current template — I've updated the description to match it, including the AI usage disclosure.

One bit of context: the new template and AI policy were merged in #6995 only a few hours before I opened this. My local fork predates that commit, so the template I followed was the one that existed in my checkout — I wasn't ignoring the current one, I didn't have it yet. I'd also raised the approach in discussion #6975 first and got a maintainer's sign-off before writing any code, specifically to avoid wasting review time.

I mention it only because "Please use the current pull request template" + immediate close, hours after the rule changed, lands as more abrupt than I suspect was intended. The change itself is small, tested against a real library, and I'm glad to walk through any of it in review.

WithoutPants (approved):

PR #7035, opened 2026-06-13
Open gallery slideshow from the galleries page

### LINKED ISSUE #1579: [Feature] Play button on gallery page
Gallery slideshows can only be started by opening a gallery, switching to Wall display, opening a picture and then clicking Play.

A play button directly on the gallery page, regardless of display mode. Possibly also when opening a picture from Grid mode.

void-function865:
## Description

Starting a gallery slideshow currently takes several steps (open the gallery → switch to Wall view → open a picture → press Play). This PR lets you open a gallery's slideshow in the lightbox directly from the **galleries page**:

- On the **grid**, hovering a gallery cover shows a **magnifying-glass** button that opens that gallery's images in the lightbox, mirroring the existing image-card interaction.
- A new **"Auto-start gallery slideshow"** interface setting (Settings → Interface → Image Lightbox) controls whether the slideshow begins playing automatically when opened from either the grid or wall galleries-page entry points.
- The cover's hover-scrubber opens the lightbox at the scrubbed image, on both the grid and Wall view.
- Clicking the cover or title still navigates to the gallery page.

Key design choices:

- **Auto-start is scoped to the galleries-page entry points** — the grid card (magnifying glass or scrubber) and Wall-view covers/scrubbers. Opening the lightbox from *within* a gallery (clicking an individual image) or via a detail-page chapter marker is arguably a different user intent — you want to view *that* image, not start a slideshow — so it never auto-starts there. Auto-start is a per-open instruction that is cleared when the lightbox closes, so it can't leak into a lightbox subsequently opened from another entry point.
- **The setting is a backend interface setting**, implemented as the gallery analogue of the scene player's `autostartVideo` ("auto-start playback") preference: a typed `ConfigInterfaceResult` boolean (`autostartGallerySlideshow`), wired through the config getter and the configure/query resolvers, defaulting to `false`. It lives alongside `autostartVideo` and the lightbox `slideshowDelay` rather than in the front-end-only UI config.
- **The hover-scrubber opens the lightbox at the scrubbed image, consistently on the grid and in Wall view.** Previously the grid card's scrubber navigated to the single-image viewer page (`/galleries/{id}/images/{i}`) while the Wall-view scrubber opened the lightbox. It now opens the lightbox at the scrubbed index in both, so scrubbing-then-clicking behaves the same everywhere. (This intentionally replaces the grid card's prior navigate-to-viewer behavior in favor of consistency.)
- **Reuse the image-card UI for the trigger.** The magnifying-glass button reuses the image card's existing `.preview-button` markup/styling; the only gallery-specific CSS is a one-line `top: 0` (the gallery cover is a block container rather than a flex one).
- **A single list-level lightbox controller (new code).** `useGalleriesLightbox` drives the lightbox for the whole list rather than a hook per card.
  - Versus a per-card hook: one lazy query for the list instead of one per gallery; the gallery is chosen at `show()` time rather than baked into the hook; query variables are passed to the lazy-query *execute* function, which avoids the `useLazyQuery` "pass variables to execute" deprecation warning; and the lightbox opens only in response to an explicit `show()`, rather than from a "data is present → open" effect that races across instances on the shared lightbox state.
  - The existing `useGalleryLightbox` (plugin API; also used by the gallery detail page) now delegates to it, so the gallery paging/fetch logic lives in one place.
  - As a drive-by, routing every gallery entry point through this one controller also fixes **path-dependent** behavior of the lightbox's thumbnail-navigation strip (`showNavigation`). It defaults to *shown* in the shared lightbox state; only the image list and the gallery viewer ever explicitly set it to hidden, while the gallery detail page's chapter markers and the Wall-view covers never set it at all — so whether the strip appeared depended on which lightbox you had opened previously in the session (e.g. on a fresh page load a Wall-view or chapter open would show it, but once you'd opened the in-gallery image viewer, which sets it hidden, it would stay hidden). The controller sets `showNavigation: false`, so the strip is consistently hidden regardless of path.

## Related Issue

Closes #1579

## Testing

- Galleries grid: hovering a cover shows the magnifying glass; clicking opens the lightbox at image 1 without navigating away.
- Wall view: clicking a cover (or scrubbing to an index) opens the lightbox; the slideshow auto-starts when the setting is on and stays paused when off.
- Setting off → lightbox opens paused; setting on → slideshow auto-advances.
- Opening the lightbox from the images list / within a gallery does **not** auto-start, even after a galleries-page auto-start open (verified the per-open flag is cleared on close).
- Cover/title click still routes to the gallery page.
- Hover-scrubber click (grid and Wall view): scrubbing to a position and clicking opens the lightbox at that image — it stays on `/galleries` (no longer navigates to the single-image viewer) and opens a different image than the magnifying glass at index 0, confirming the scrubbed index is honored.
- Multi-page galleries (>40 images) page correctly in the lightbox — via the new grid entry point and the existing detail-page chapters / Wall cards that now share the same controller.
- Selecting mode hides the magnifying glass (drag/box-select unaffected).
- `make validate` passes (Go build/vet/gofmt, tsc, biome, stylelint).

## Screenshots

**Before:** hovering a gallery cover on the galleries only shows the select box; there is no way to open the gallery's slideshow from the grid.

before

**After:** hovering a gallery cover also shows a magnifying-glass button; clicking it opens the gallery's images in the lightbox (auto-starting the slideshow when the new setting is enabled).

after

## Checklist

## AI Usage Disclosure

I used Claude Code (Anthropic) to assist with this PR. It helped explore the existing image-card/lightbox implementations, propose and implement the approach (the config setting and resolvers, the gallery-card magnifying glass, and the list-level lightbox controller). I directed the design decisions, reviewed all changes, and manually verified the behavior against a running instance (including the auto-start scoping, scrubber/title behavior, multi-page paging, and selection mode).

## Additional Context

The new setting (`autostartGallerySlideshow`) requires regenerating the GraphQL bindings (`make generate`); no generated files are committed, consistent with the repo's setup. The setting defaults to `false`, so existing behavior is unchanged unless a user opts in.

### DIFF
--- a/ui/v2.5/src/components/Galleries/GalleryCard.tsx
+++ b/ui/v2.5/src/components/Galleries/GalleryCard.tsx
@@ -1,5 +1,5 @@
 import { Button, ButtonGroup, OverlayTrigger, Tooltip } from "react-bootstrap";
-import React, { useCallback, useMemo, useState } from "react";
+import React, { MouseEvent, useMemo, useState } from "react";
 import * as GQL from "src/core/generated-graphql";
 import { GridCard } from "../Shared/GridCard/GridCard";
 import { HoverPopover } from "../Shared/HoverPopover";
@@ -10,22 +10,27 @@ import { PerformerPopoverButton } from "../Shared/PerformerPopoverButton";
 import { PopoverCountButton } from "../Shared/PopoverCountButton";
 import NavUtils from "src/utils/navigation";
 import { RatingBanner } from "../Shared/RatingBanner";
-import { faBox, faPlayCircle, faTag } from "@fortawesome/free-solid-svg-icons";
+import {
+  faBox,
+  faPlayCircle,
+  faSearch,
+  faTag,
+} from "@fortawesome/free-solid-svg-icons";
 import { galleryTitle } from "src/core/galleries";
 import { StudioOverlay } from "../Shared/GridCard/StudioOverlay";
 import { GalleryPreviewScrubber } from "./GalleryPreviewScrubber";
 import cx from "classnames";
-import { useHistory } from "react-router-dom";
 import { PatchComponent } from "src/patch";
 
 interface IGalleryPreviewProps {
   gallery: GQL.SlimGalleryDataFragment;
   onScrubberClick?: (index: number) => void;
+  onPreview?: (ev: MouseEvent) => void;
   disabled?: boolean;
 }
 
 export const GalleryPreview: React.FC<IGalleryPreviewProps> = React.memo(
-  ({ gallery, onScrubberClick, disabled }) => {
+  ({ gallery, onScrubberClick, onPreview, disabled }) => {
     const [imgSrc, setImgSrc] = useState<string | undefined>(
       gallery.paths.cover ?? undefined
     );
@@ -50,6 +55,13 @@ export const GalleryPreview: React.FC<IGalleryPreviewProps> = React.memo(
             disabled={disabled}
           />
         )}
+        {onPreview && gallery.image_count > 0 ? (
+          <div className="preview-button">
+            <Button onClick={onPreview}>
+              <Icon icon={faSearch} />
+            </Button>
+          </div>
+        ) : undefined}
       </div>
     );
   }
@@ -62,6 +74,7 @@ interface IGalleryCardProps {
   selected?: boolean | undefined;
   zoomIndex?: number;
   onSelectedChanged?: (selected: boolean, shiftKey: boolean) => void;
+  onPreview?: (index: number, ev?: MouseEvent) => void;
 }
 
 const GalleryCardPopovers = React.memo(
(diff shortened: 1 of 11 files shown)


graphql/schema/types/config.graphql
```diff
@@ -401,6 +401,8 @@ input ConfigInterfaceInput {
   autostartVideoOnPlaySelected: Boolean
   "If true, next scene in playlist will be played at video end by default"
   continuePlaylistDefault: Boolean
+  "If true, the slideshow will autostart when opening a gallery's lightbox from the galleries page"
+  autostartGallerySlideshow: Boolean
```
WithoutPants: Adding this to the graphql schema is unnecessary for a UI-only option. Use the ui settings instead (see `src/core/config.ts`). This will eliminate the need for server-side changes.

WithoutPants (changes requested):
Concept sounds good and the proposal looks sound. I'll test after the config stuff is addressed.

graphql/schema/types/config.graphql
void-function865: I dropped the GraphQL schema field and the config key/getter/resolvers entirely. It's now a UI-only setting stored via the existing UI config (`src/core/config.ts` / `saveUI`), so there are no server-side changes. Ready for testing whenever you get a chance.

WithoutPants (approved):

PR #7040, opened 2026-06-14
Fix: json.Number Custom Field Filters

### LINKED ISSUE #7031: Numeric custom-field filters broken on scenes (work on performers)

### Describe the bug

Found via my plugin [Restash](https://github.com/Espionage9248/Restash), which writes an integer `restash_score` into both scene and performer `custom_fields` (`restash_score` = a 0-100 freshness score; stored as a JSON int, confirmed by reading `custom_fields` back). Filtering that field works on **performers** but is broken on **scenes**, despite identical stored values.

### Symptom (scenes)

For a numeric custom field, every modifier is wrong:
- `LESS_THAN` → matches **all** rows
- `GREATER_THAN` / `EQUALS` / `BETWEEN` → match **none**

…even when rows with a qualifying value demonstrably exist. Native scene fields (`rating100`) filter correctly.

### Root cause

The signature (`<` matches all, `>`/`=` match none) is what happens when the **operand is bound as TEXT** against a numeric stored value: in SQLite's storage-class ordering (`NULL < INTEGER/REAL < TEXT`), any number sorts below a text operand. The performer path binds it numerically and works; the scene path does not.

This is the performer bug from #5682 (fixed in #5685) recurring on the scene filter path. Binding numeric custom-field operands to the column's numeric type (or `CAST`) on the scene path should fix it.

### Steps to reproduce

Same field/value, two entity types:

```graphql
query($s: SceneFilterType, $p: PerformerFilterType) {
  s: findScenes(filter:{per_page:1}, scene_filter:$s) { count }
  p: findPerformers(filter:{per_page:1}, performer_filter:$p) { count }
}
```

| Criterion (`restash_score`, int) | Scenes | Performers |
|---|---|---|
| `EQUALS [N]` (a row holds exactly N) | **0** | ≥1 ✅ |
| `GREATER_THAN [90]` | **0** | correct subset, all verified >90 ✅ |
| `LESS_THAN [10]` | **all rows** | correct subset, all verified <10 ✅ |

Reproduces with both int (`restash_score`) and float (`restash_raw`) fields.

### Expected behaviour

Scene numeric custom-field criteria compare numerically - matching the performer path and native scene fields. `restash_score GREATER_THAN [90]` returns scenes scoring above 90; `LESS_THAN [10]` returns only those below 10; `EQUALS [N]` returns exact matches.

_No response_

### Stash version

v0.31.1

### Device details

Unraid 7.2.6

### Relevant log output

```shell

```

CynicalAtropos:
## Description

Fixes numeric custom-field filters when GraphQL `Any` values arrive as `json.Number`.

The previous fix (#5685) for performer filters normalized `json.Number` in the top-level `findPerformers` resolver, but similar filters for scenes, images, galleries, studios, tags, groups, etc still had the issue as well as nested filters such as `scene_filter.performers_filter.custom_fields` that bypass that resolver. This change moves the conversion into the shared SQLite custom-field value conversion path so direct and nested custom-field filters bind numeric operands as numbers instead of text.

Adds regression coverage for direct scene custom-field filtering and the nested performer-filter bypass path.

## Related Issue

Fixes #7031

## Testing

- `go test -tags integration ./pkg/sqlite -run "TestSceneQueryCustomFields/(json_number_greater_than|json_number_equals|json_number_between|nested_performer_json_number_greater_than|nested_performer_json_number_less_than)" -count=1`
- `go test -tags integration ./pkg/sqlite -run "TestSceneQueryCustomFields|Test(Performer|Gallery|Image|Group|Studio|Tag)QueryCustomFields" -count=1`
- `go test -tags "integration sqlite_stat4 sqlite_math_functions" ./pkg/sqlite -count=1`
- `git diff --check`
- `mingw32-make lint`
- Manually testing with custom fields and filters following the general steps in the reported issue to confirm that it resolves the filtering. 

## Screenshots

N/A - backend filtering fix.

## Checklist

## AI Usage Disclosure

Used Codex to investigate the filter paths, identify the nested-filter bypass, draft the shared SQLite conversion fix, and write targeted regression tests. I reviewed the changes, test results, and manually tested myself.

## Additional Context

The fix is intentionally centralized in the SQLite custom-field conversion path rather than adding resolver-level conversions for each entity. This covers top-level custom-field filters and nested related filters that do not pass through the top-level entity resolvers.

### DIFF
--- a/internal/api/json.go
+++ b/internal/api/json.go
@@ -3,8 +3,6 @@ package api
 import (
 	"encoding/json"
 	"strings"
-
-	"github.com/stashapp/stash/pkg/models"
 )
 
 // jsonNumberToNumber converts a JSON number to either a float64 or int64.
@@ -15,15 +13,6 @@ func jsonNumberToNumber(n json.Number) interface{} {
 	}
 	ret, _ := n.Int64()
 	return ret
-}
-
-// anyJSONNumberToNumber converts a JSON number using jsonNumberToNumber, otherwise it returns the existing value
-func anyJSONNumberToNumber(v any) any {
-	if n, ok := v.(json.Number); ok {
-		return jsonNumberToNumber(n)
-	}
-
-	return v
 }
 
 // ConvertMapJSONNumbers converts all JSON numbers in a map to either float64 or int64.
@@ -41,24 +30,6 @@ func convertMapJSONNumbers(m map[string]interface{}) (ret map[string]interface{}
 		} else {
 			ret[k] = v
 		}
-	}
-
-	return ret
-}
-
-func convertCustomFieldCriterionValues(c models.CustomFieldCriterionInput) models.CustomFieldCriterionInput {
-	nv := make([]any, len(c.Value))
-	for i, v := range c.Value {
-		nv[i] = anyJSONNumberToNumber(v)
-	}
-	c.Value = nv
-	return c
-}
-
-func convertCustomFieldCriterionInputJSONNumbers(c []models.CustomFieldCriterionInput) []models.CustomFieldCriterionInput {
-	ret := make([]models.CustomFieldCriterionInput, len(c))
-	for i, v := range c {
-		ret[i] = convertCustomFieldCriterionValues(v)
 	}
 
 	return ret
--- a/internal/api/resolver_query_find_performer.go
+++ b/internal/api/resolver_query_find_performer.go
@@ -31,11 +31,6 @@ func (r *queryResolver) FindPerformers(ctx context.Context, performerFilter *mod
 		}
 	}
 
-	// #5682 - convert JSON numbers to float64 or int64
-	if performerFilter != nil {
-		performerFilter.CustomFields = convertCustomFieldCriterionInputJSONNumbers(performerFilter.CustomFields)
-	}
-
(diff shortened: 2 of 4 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/restash/9008/14

Gykes (changes requested):
You didn't remove the old `convertCustomFieldCriterionInputJSONNumbers(...)` is there a particular reason? How I see it now is it's already a Go number whenever SQlite sees the value. If it's unused I would remove it or comment on it stating why it should exist.

CynicalAtropos:
> You didn't remove the old `convertCustomFieldCriterionInputJSONNumbers(...)` is there a particular reason? How I see it now is it's already a Go number whenever SQlite sees the value. If it's unused I would remove it or comment on it stating why it should exist.

No compelling reason other than limiting the scope of the PR, so I went ahead and pushed an update to remove the obsolete code and more cleanly bring performers in line with the rest.  Re-ran the tests and did a manual smoke test to make sure performers still filtered fine.

Gykes (approved):
LGTM. Tested across various objects and it worked as expected with no obvious regressions

discourse-stashapp:

https://discourse.stashapp.cc/t/restash/9008/17

PR #7055, opened 2026-06-17
Extract embedded subtitles as captions

q36zhd46w17o:
## Summary
- detect embedded subtitle streams from ffprobe while scanning video files
- extract each new supported embedded subtitle stream to a WebVTT sidecar without overwriting existing caption files
- register extracted or already-present VTT files through the existing scene caption path so the current player can display them

Fixes #3875.

## Bounty
- OpenCollective bounty: $50 contribution 745825, assigned in #3875.

## Testing
- `PATH=/tmp/codex-go1.25.0/go/bin:$PATH go test ./pkg/file/video ./pkg/scene ./pkg/models`
- `make touch-ui`
- `PATH=/tmp/codex-go1.25.0/go/bin:$PATH go test ./internal/manager`
- manual ffmpeg smoke test: generated a tiny MKV with an embedded SRT stream and extracted it with `-map 0:1 -c:s webvtt -f webvtt`, producing valid WebVTT output

Issue #6064, opened 2025-08-25
[Bug Report] QSV hwaccel on ffmpeg v8 is broken


feederbox826:
If affected, downgrade to v7.x, the issue seems to be upstream or they have not updated their documentation yet

**Describe the bug**
ffmpeg v8.0 hangs indefinitely when running QSV hwaccel tests

**To Reproduce**
1. enable hardware acceleration
2. On windows, download gyan.dev ffmpeg
3. start stash to start inithw check
4. Indefinite hang

**Screenshots**
```pwsh
./ffmpeg -hide_banner -v warning -init_hw_device qsv=hw -filter_hw_device hw -f lavfi -i color=c=red:s=1280x720 -t 0.1 -c:v h264_qsv -global_quality 20 -preset faster -vf hwupload=extra_hw_frames=64,format=qsv,scale_qsv=854:480 -f null -
[AVHWFramesContext @ 0000025386bcfbc0] Error synchronizing the operation
[hwupload @ 0000025386bdb700] Failed to upload frame: -1313558101.
[vf#0:0 @ 0000025386bcae80] Error while filtering: Unknown error occurred
Received > 3 system signals, hard exiting

./ffmpeg -hide_banner -v warning -init_hw_device qsv=hw -filter_hw_device hw -f lavfi -i color=c=red:s=1280x720 -t 0.1 -c:v vp9_qsv -global_quality 20 -preset faster -vf hwupload=extra_hw_frames=64,format=qsv,scale_qsv=854:480 -f null -
>>
[AVHWFramesContext @ 000001fa9a3ed7c0] Error synchronizing the operation
[hwupload @ 000001fa9a3f88c0] Failed to upload frame: -1313558101.
[vf#0:0 @ 000001fa9a3e6440] Error while filtering: Unknown error occurred
Received > 3 system signals, hard exiting
```

feederbox826:
Since someone mentioned it on the discourse: You can find old downloads from gyan.dev for the last version of ffmpeg near the bottom of the section here

https://www.gyan.dev/ffmpeg/builds/#release-builds

discourse-stashapp:
This issue has been mentioned on **Stash**. There might be relevant details there:

https://discourse.stashapp.cc/t/initial-installation-hangs-on-optimizing-database/4739/2

PR #7057, opened 2026-06-19
feat: extract embedded subtitles as VTT captions during scan

### LINKED ISSUE #3875: [Feature] Display embedded subtitles

idan57570idan-svg:
## Description

When scanning a video file that contains embedded subtitle streams (e.g. SRT, ASS/SSA, or WebVTT tracks embedded in MKV/MP4), the scanner now automatically extracts each supported stream to a WebVTT sidecar file and registers it via the existing `CaptionUpdater` pipeline. This makes embedded subtitles available to the web player without the user needing to manually extract them.

**Changes:**

- `pkg/ffmpeg/ffprobe.go` — added `SubtitleStreams []*FFProbeStream` field to `VideoFile` and `getSubtitleStreams()` method, populated during `parse()`.
- `pkg/file/video/embedded_caption.go` (**new file**) — `ExtractEmbeddedCaptions()` extracts each supported subtitle stream via `ffmpeg -c:s webvtt` and registers it through `CaptionUpdater`. Skips streams with unsupported codecs and languages already registered.
- `internal/manager/task_scan.go` — after new/fingerprint-changed video files are stored, calls `ExtractEmbeddedCaptions` using the manager's `FFProbe`/`FFMpeg` instances.

Supported codecs: `subrip`, `srt`, `ass`, `ssa`, `webvtt`, `mov_text`.

## Related Issue

Fixes #3875

## Testing

Tested manually with an MKV containing embedded SRT (English + French) and an ASS track:
1. Added the file to a stash library and ran a full scan.
2. Confirmed two `.vtt` sidecar files were created next to the video (`video.en.vtt`, `video.fr.vtt`).
3. Confirmed the captions appeared in the scene player caption selector.
4. Re-scanned the file — confirmed extraction was skipped (files already existed, captions already registered).
5. Verified files without subtitle streams are unaffected.

## Screenshots

N/A (backend-only change).

## Checklist

## AI Usage Disclosure

Claude Code (Anthropic) was used to help generate the Go implementation. I reviewed every line, understand the design, and manually tested the behaviour as described above. All architectural decisions (hooking into `handleFile`, using the existing `CaptionUpdater` interface, codec allowlist) were made by me.

## Additional Context

Previous attempt: #7055 (closed for not following the PR template — this PR follows the template).

### DIFF
--- a/internal/manager/task_scan.go
+++ b/internal/manager/task_scan.go
@@ -362,6 +362,24 @@ func (j *ScanJob) handleFile(ctx context.Context, f file.ScannedFile, progress *
 		}
 	}
 
+	// extract embedded subtitle streams for new or changed video files
+	if r.New || r.FingerprintChanged {
+		if videoFile, ok := r.File.(*models.VideoFile); ok {
+			mgr := GetInstance()
+			if mgr.FFProbe != nil && mgr.FFMpeg != nil {
+				if probed, err := mgr.FFProbe.NewVideoFile(videoFile.Path); err == nil && len(probed.SubtitleStreams) > 0 {
+					txnMgr := j.scanner.Repository.TxnManager
+					fileRepo := j.scanner.Repository.File
+					if txnErr := txn.WithTxn(ctx, txnMgr, func(ctx context.Context) error {
+						return video.ExtractEmbeddedCaptions(ctx, videoFile.Path, probed.SubtitleStreams, mgr.FFMpeg.Path(), videoFile.ID, fileRepo)
+					}); txnErr != nil {
+						logger.Errorf("Error extracting embedded captions for %s: %v", videoFile.Path, txnErr)
+					}
+				}
+			}
+		}
+	}
+
 	// clean captions - scene handler handles this as well, but
 	// unchanged files aren't processed by the scene handler
 	if r.IsUnchanged() {
--- a/pkg/ffmpeg/ffprobe.go
+++ b/pkg/ffmpeg/ffprobe.go
@@ -85,9 +85,10 @@ func ResolveFFProbe(path string, fallbackPath string) string {
 
 // VideoFile represents the ffprobe output for a video file.
 type VideoFile struct {
-	JSON        FFProbeJSON
-	AudioStream *FFProbeStream
-	VideoStream *FFProbeStream
+	JSON            FFProbeJSON
+	AudioStream     *FFProbeStream
+	VideoStream     *FFProbeStream
+	SubtitleStreams  []*FFProbeStream
 
 	Path      string
 	Title     string
@@ -343,6 +344,8 @@ func parse(filePath string, probeJSON *FFProbeJSON) (*VideoFile, error) {
 		}
 	}
 
+	result.SubtitleStreams = result.getSubtitleStreams()
+
 	return result, nil
 }
 
@@ -363,6 +366,17 @@ func isRotated(s *FFProbeStream) bool {
 	}
 
 	return false
+}
+
+func (v *VideoFile) getSubtitleStreams() []*FFProbeStream {
+	var result []*FFProbeStream
+	for i := range v.JSON.Streams {
+		s := &v.JSON.Streams[i]
+		if s.CodecType == "subtitle" && s.Disposition.AttachedPic == 0 {
+			result = append(result, s)
+		}
+	}
+	return result
 }
 
 func (v *VideoFile) getAudioStream() *FFProbeStream {
--- a/pkg/file/video/embedded_caption.go
+++ b/pkg/file/video/embedded_caption.go
@@ -1,0 +1,89 @@
+package video
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	stashExec "github.com/stashapp/stash/pkg/exec"
+	"github.com/stashapp/stash/pkg/ffmpeg"
+	"github.com/stashapp/stash/pkg/logger"
+	"github.com/stashapp/stash/pkg/models"
+)
+
+var supportedSubtitleCodecs = map[string]struct{}{
+	"subrip":   {},
+	"srt":      {},
+	"ass":      {},
+	"ssa":      {},
+	"webvtt":   {},
+	"mov_text": {},
+}
+
+// ExtractEmbeddedCaptions extracts embedded subtitle streams from a video file
+// to WebVTT sidecar files and registers them via the provided CaptionUpdater.
+// Streams with unsupported codecs or that already have a registered VTT caption
+// for the same language are silently skipped.
+func ExtractEmbeddedCaptions(ctx context.Context, videoPath string, streams []*ffmpeg.FFProbeStream, ffmpegPath string, fileID models.FileID, w CaptionUpdater) error {
+	captions, err := w.GetCaptions(ctx, fileID)
+	if err != nil {
+		return fmt.Errorf("getting captions for %s: %w", videoPath, err)
+	}
+
+	changed := false
+	for _, s := range streams {
+		if _, ok := supportedSubtitleCodecs[s.CodecName]; !ok {
+			logger.Debugf("[embedded captions] unsupported codec %q in %s, skipping stream %d", s.CodecName, videoPath, s.Index)
+			continue
+		}
+
+		lang := LangUnknown
+		if IsValidLanguage(s.Tags.Language) {
+			lang = s.Tags.Language
+		}
+
+		if IsLangInCaptions(lang, "vtt", captions) {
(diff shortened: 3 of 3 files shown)


glatinone (commented):
Thanks for picking this up — extracting embedded subs on scan and feeding them through the existing caption pipeline is the right shape for this. A few things I'd want addressed before merge, mostly around the ffprobe re-invocation and where the ffmpeg calls happen relative to the DB transaction:

**1. The re-probe in `task_scan.go` duplicates work `Decorator.Decorate` already did, and drops the context in the process.**

`pkg/file/video/scan.go:30` already calls `probe.NewVideoFile(base.Path)` for every new/changed file as part of the normal scan (that's where `Format`/`VideoCodec`/`Width`/etc. come from). This PR adds `SubtitleStreams` to `ffmpeg.VideoFile` right there, then discards it and calls `NewVideoFile` a *second* time per file in `task_scan.go:370` just to get it back.

Beyond the extra subprocess spawn per file (real cost across a large library), that second call is also made with `stashExec.Command(...)`, not `CommandContext`, so it has no way to respect cancellation or a deadline. This codebase already hit this exact class of bug once — see the `// #6064 - add timeout to context to prevent hangs` fix in `pkg/ffmpeg/codec_hardware.go`. A hung/corrupt file or slow network-mounted media file here would stall the scan job on this file indefinitely.

Suggest threading `SubtitleStreams` through from the existing `Decorate()` probe (e.g. onto `models.VideoFile`, or via the `ScanResult`) instead of probing again. That removes the redundant spawn and the missing-context gap in one move.

**2. `ExtractEmbeddedCaptions` runs ffmpeg subprocess calls inside `txn.WithTxn`.**

Every other `txn.WithTxn` usage in the codebase (`caption.go`, `identify.go`, `autotag/*.go`, etc.) wraps DB-only work. Here it wraps N sequential ffmpeg extractions of unbounded duration. Under a parallel scan (multiple files being handled concurrently), that's N transactions held open across external subprocess calls simultaneously — worth moving the ffmpeg extraction outside the txn and only wrapping the final `UpdateCaptions` DB write.

**3. Pre-existing file check doesn't validate the file.**

```go
if _, statErr := os.Stat(vttPath); statErr != nil {
```
This treats any file at `vttPath` as a successful prior extraction, including a zero-byte/truncated one left by an ffmpeg process that got killed mid-write (crash, OOM, or a future timeout once #1 is addressed). That caption would then get registered and never retried. Worth at least a minimum-size check, or writing to a temp path and renaming into place on success.

**4. Formatting** — `SubtitleStreams []*FFProbeStream` in the `VideoFile` struct looks like it might not be `gofmt`-clean relative to the other field alignment; worth double-checking `gofmt -l` picks up nothing here (I don't have a Go toolchain handy to verify this myself).

None of this is a knock on the overall approach — just want the transaction/timeout issues sorted before this lands, since they're the kind of thing that only shows up under load on a big library scan.

glatinone:
Small follow-up/correction to my review above:

On point 4 (formatting) — I want to correct myself: it's a cosmetic whitespace issue only (the extra space before `[]*FFProbeStream` relative to the other fields' alignment), not a memory/padding concern. I double-checked the full `VideoFile` struct and every field in it is already an 8-byte-aligned type (pointers, slices, strings, float64/int64, `time.Time`), so there's no padding to reclaim by reordering — reordering wouldn't change the struct's size here. Just run it through `gofmt -w` and it's a non-issue.

Also, to be precise on points 1-3: I traced the mechanisms (redundant un-cancellable probe, ffmpeg calls inside `txn.WithTxn`, unvalidated pre-existing file) directly in the code and against how the rest of the codebase does things, but I haven't reproduced or measured an actual hang/connection-pool-exhaustion under load — so treat those as risks worth closing off before merge, not confirmed production incidents.

glatinone:
Following up on my earlier review with concrete suggestions for the three points, since they interact with each other a bit:

**1. Move the ffmpeg extraction outside the transaction (point 2 from before)**

Right now `ExtractEmbeddedCaptions` does the read, the ffmpeg work, and the write all inside whatever transaction the caller wraps it in. Splitting into three phases keeps the transaction boundary tight around DB-only work:

```go
func ExtractEmbeddedCaptions(ctx context.Context, videoPath string, streams []*ffmpeg.FFProbeStream, ffmpegPath string, fileID models.FileID, w CaptionUpdater) error {
	// Phase 1: read existing captions (brief read tx, owned by the caller or a
	// short-lived txn.WithTxn here — no external process calls inside it).
	captions, err := w.GetCaptions(ctx, fileID)
	if err != nil {
		return fmt.Errorf("getting captions for %s: %w", videoPath, err)
	}

	// Phase 2: all ffmpeg subprocess work happens here, fully outside any
	// transaction. This is the part with unbounded, IO-dependent duration.
	// captions grows in-place (same as the current code) so that two streams
	// sharing a language within the same run still correctly dedupe against
	// each other, not just against what was already registered before this call.
	changed := false
	for _, s := range streams {
		if _, ok := supportedSubtitleCodecs[s.CodecName]; !ok {
			continue
		}
		lang := LangUnknown
		if IsValidLanguage(s.Tags.Language) {
			lang = s.Tags.Language
		}
		if IsLangInCaptions(lang, "vtt", captions) {
			continue
		}

		vttPath := GetCaptionPath(videoPath, lang, "vtt")
		if err := ensureVTTFile(ctx, ffmpegPath, videoPath, s.Index, vttPath); err != nil {
			logger.Warnf("[embedded captions] failed to extract stream %d from %s: %v", s.Index, videoPath, err)
			continue
		}
		captions = append(captions, &models.VideoCaption{
			LanguageCode: lang,
			Filename:     filepath.Base(vttPath),
			CaptionType:  "vtt",
		})
		changed = true
	}

	if !changed {
		return nil
	}

	// Phase 3: minimal write transaction — just the DB update.
	return w.UpdateCaptions(ctx, fileID, captions)
}
```

The caller (`task_scan.go`) then doesn't need to wrap this whole call in `txn.WithTxn` at all — `GetCaptions`/`UpdateCaptions` can each own their own short transaction internally (however `CaptionUpdater`'s real implementation already handles that), rather than the caller holding one open across the ffmpeg calls.

**2. Extract the zero-byte guard into its own function (point 3)**

Pulling the `os.Stat` check into a named helper makes the "trust existing file" logic testable on its own, and gives you a natural place to add the size/re-extract guard:

```go
func ensureVTTFile(ctx context.Context, ffmpegPath, videoPath string, streamIndex int, vttPath string) error {
	if info, statErr := os.Stat(vttPath); statErr == nil {
		if info.Size() > 0 {
			return nil // valid file already present
		}
		// Zero-byte leftover from a prior killed/interrupted extraction — remove
		// and re-extract rather than trusting it.
		if err := os.Remove(vttPath); err != nil {
			return fmt.Errorf("removing zero-byte caption %s: %w", vttPath, err)
		}
	}

	if err := extractSubtitleStream(ctx, ffmpegPath, videoPath, streamIndex, vttPath); err != nil {
		return err
	}

	// Confirm the extraction actually produced a non-empty file before the
	// caller registers it — ffmpeg can exit 0 in some edge cases without
	// writing usable output.
	info, err := os.Stat(vttPath)
	if err != nil {
		return fmt.Errorf("stat after extraction for %s: %w", vttPath, err)
	}
	if info.Size() == 0 {
		return fmt.Errorf("extraction produced a zero-byte file: %s", vttPath)
	}
	return nil
}
```

**3. task_scan.go — reuse the existing probe instead of re-running ffprobe**

This is the same point as before: `pkg/file/video/scan.go`'s `Decorator.Decorate` already runs `probe.NewVideoFile(base.Path)` for every new/changed file, and `SubtitleStreams` is already part of that result — it just gets dropped when only `Format`/`VideoCodec`/etc. are copied into `models.VideoFile`. Threading `SubtitleStreams` through onto `models.VideoFile` (or via whatever result type `ScanFile` returns) lets `task_scan.go` read `videoFile.SubtitleStreams` directly, removing both the second `ffprobe` subprocess spawn and the outer transaction wrapper around the whole block (since with phases 1–3 above, the DB work is already appropriately scoped).

None of this is meant as a required rewrite — just a concrete shape for closing out the three points, since "decouple from the transaction" and "validate the file" are easier to see with actual code in front of them.

Gykes:
I believe this would be better suited as a plugin. Stash already auto-associates .vtt sidecars during a scan. A plugin that extracts the embedded streams to `video.en.vtt` should get picked up automatically on the next scan. I think plumbing all of this into core is just a lot of extra maintenance. 

With that in mind i'm going to close this for now. Feel free to reach out if you would like to discuss it.

Issue #3727, opened 2023-05-10
[Bug Report] Japansese voiced consonant mark bug


suzuhiroruri:
**Describe the bug**
Japansese voiced consonant mark is not imported property.
example:
・"が" is Japanese one word, but stash app import as two words as "か" and "゛".
・"ぱ" is Japanese one word, but stash app import as two words as "は" and "゜".

**To Reproduce**
1. Import video file which has name Japansese voiced consonant mark.

**Expected behavior**

**Stash Version: (from Settings -> About):**
Voiced consonant mark is imported as one word.

**Desktop (please complete the following information):**
 - mac OS

Releck:
Tried to reproduce this with a file named `がぎぐげご.mp4` and couldn't. Each character is correctly imported with no separate characters for dakutens.

Tested on 0069c48

suzuhiroruri:
This file is separated characters for dakutens.
https://github.com/stashapp/stash/assets/1904259/990e1959-87ca-4921-b7b5-1d8e58a2281e

Releck:
Can you please copy the filename with `岬あずさ` here. I suspect that somewhere along the line dakutens are manifesting as unicode control characters and you actually have `岬あずさ` (different character on the す) in your filename. More information about your system locale would help. I'm not convinced this is a stash bug.

suzuhiroruri:
岬あずさ

suzuhiroruri:
Sorry, I made mistake.
https://github.com/stashapp/stash/issues/3727#issuecomment-1547115845
is imported filename.

These are filename:
```
OKS-128 岬あずさ 濡れてテカってピッタリ密着 神スク水 可愛い女子のスクール水着姿をじっとりと堪能！着替え盗撮から始まり貧乳から巨乳にパイパン、ハミ毛、ジョリワキ等のフェチ接写やローションソーププレイやスク水ぶっかけ等を完全着衣で楽しむAV
```

```
岬あずさ
```

suzuhiroruri:
Thanks, I understand that.
However I'm bothered, when I search dakuten word, becourse search dakuten words are not same as imported words. 
For example, file name has ```岬あずさ``` is not same word with search word ```岬あずさ```.

suzuhirokakaku:
This is mac OS problem.
[...]

PR #7070, opened 2026-06-25
fix: normalize macOS NFD filenames to NFC when scanning

### LINKED ISSUE #4425: [Bug Report] mac-os NFD/NFC file name problem
**Describe the bug**
Japansese voiced consonant mark is not imported property.
example:
・"が" is Japanese one word, but stash app import as two words as "か" and "゛".
・"ぱ" is Japanese one word, but stash app import as two words as "は" and "゜".

**To Reproduce**
1. Import video file which has name Japansese voiced consonant mark.

**Expected behavior**

Voiced consonant mark is imported as one word.

 - mac OS

suzuhiroruri:
## Description

On macOS, filenames are reported by the filesystem in NFD (decomposed) form — for example `が` is stored as `か` plus a combining mark. SQLite compares strings byte-wise, so a search for the composed (NFC) form that I actually type never matches the decomposed form that was stored during scanning.

This normalizes scanned paths to NFC before storing them, so stored data matches what users type. A custom collation was suggested in the issue, but collations don't apply to `LIKE` (which the `q` search uses), so normalizing at scan time is the approach taken here.

Details:
- Normalization is applied only on macOS, whose filesystems are normalization-insensitive (the on-disk file is still found via its NFC path). It is a no-op elsewhere, because on a byte-exact filesystem (e.g. Linux) storing NFC would break access to files whose on-disk names are NFD.
- Path containment checks (library membership / generated folder) are made normalization-insensitive.
- File/folder rename detection treats an entry differing only by normalization as the same entry, so rescanning an existing library updates paths in place rather than creating duplicates.
- Entries inside zip files are left byte-exact.

## Related Issue

Fixes #4425 (also the earlier #3727)

## Testing

- [x] Added unit tests for `NormalizePath` and the normalization-insensitive `IsPathInDir`
- [x] Built locally and ran a scan against a file whose on-disk name is NFD; confirmed the stored `basename` is NFC and that a composed-form search returns the file
- [x] Confirmed the on-disk file is still accessible via the normalized path

## Screenshots

N/A (no visual changes)

## Checklist

- [ ] I have made corresponding changes to the documentation (if applicable). — N/A

## AI Usage Disclosure

I used Claude Code (Anthropic) to assist with implementing the change, writing the unit tests, running the manual scan/search verification, and drafting this description. I have reviewed and understand all of the changes, performed the testing described above, and take full responsibility for the code.

## Additional Context

This is a small, focused bug fix for an issue I reported, not a large generated feature.

### DIFF
--- a/internal/manager/manager_tasks.go
+++ b/internal/manager/manager_tasks.go
@@ -145,7 +145,9 @@ func (s *Manager) Scan(ctx context.Context, input ScanMetadataInput) (int, error
 		ZipFileExtensions:     cfg.GetGalleryExtensions(),
 		// ScanFilters is set in ScanJob.Execute
 		// HandlerRequiredFilters is set in ScanJob.Execute
-		RootPaths: cfg.GetStashPaths().Paths(),
+		// #4425 - normalize root paths so they match the NFC-normalized paths
+		// stored during scanning (see fsutil.NormalizePath).
+		RootPaths: fsutil.NormalizePaths(cfg.GetStashPaths().Paths()),
 		Rescan:    input.Rescan,
 	}
 
--- a/internal/manager/task_scan.go
+++ b/internal/manager/task_scan.go
@@ -207,13 +207,24 @@ func (j *ScanJob) queueFileFunc(ctx context.Context, f models.FS, zipFile *file.
 			return err
 		}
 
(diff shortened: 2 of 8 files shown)


DogmaDragon:
Please follow the pull request template.

PR #7077, opened 2026-06-27
Fix: Nested Library Path Matching For Propagating Library Type Exclusions

### LINKED ISSUE #4956: [Bug Report] Nested library path quirks
**Describe the bug**
Library path exclusions propogate to subdirectories and don't act as intended

**To Reproduce**
1. Create library at `/media/stash/images` for images
2. Create library at `/media/stash` for videos
3. Add libraries to stash with only videos/images enabled for each directory
4. Content of subdirectories are excluded as they are in the parent directory
5. excluded content in the subdirectory is included as part of the parent directory

The video/images selection of parent directories is propogated to and overrides subdirectories, preventing subdirectory exclusion/ specific inclusion

**Expected behavior**
Subdirectories would be handled independently regardless of file structure

**Screenshots**
N/A

v0.26.1

N/A

N/A

**Additional context**
https://discord.com/channels/559159668438728723/559159910550732809/1249850025509392496

CynicalAtropos:
## Description

Fixes nested library path matching so overlapping library entries consistently use the most specific matching path.

Previously, library path resolution returned the first configured path that contained the scanned file. For nested libraries, this made media-type inclusion/exclusion order-dependent. A child library could inherit the parent library’s video/image settings if the parent appeared first in the config.

This updates `StashConfigs.GetStashFromDirPath` to choose the deepest matching library path, and routes `GetStashFromPath` through the same logic. This keeps scan, clean, and related path classification behavior consistent.

## Related Issue

Fixes #4956 

## Testing

- `golangci-lint run ./internal/manager/config`
- `git diff --check`
- `go test ./internal/... `
- Manually tested this with nested library paths with mixed images/videos in both the parent and child.  Confirmed scan, clean, and generate with the nest library and swapping the video/image exclusions between them as well as swapping the order.
  - Also did a quick smoke test of image clips to make sure they still followed the exclusions

## Screenshots

N/A

## Checklist

I don't believe any doc changes are necessary since this is just aligning it with the expected behavior.  One possible exception would be if a specific direction for .stashignore is chosen.

## AI Usage Disclosure

Used Codex to investigate, review my changes, and write the testing.

## Additional Context

This change also makes the `.stashignore` boundary behavior deterministic for nested library entries, but that behavior may need explicit review/confirmation.

Without this fix, the `.stashignore` boundary already depends on library order because `GetStashFromDirPath` returns the first matching library.

For a nested path such as `/media/images/foo.jpg`:

- Config order `/media`, then `/media/images`: old code picks `/media`, so parent `.stashignore` files cascade, but the parent video/image exclusions incorrectly win.
- Config order `/media/images`, then `/media`: old code picks `/media/images`, so the child exclusions win, but parent `/media/.stashignore` no longer participates.

So this patch does not introduce the possible `.stashignore` discrepancy from nothing. It resolves order-dependent media-type behavior by deterministically choosing the child library config. If parent `.stashignore` files are intended to cascade across nested library entries, then the selected library config probably should not also be used as the `.stashignore` root without a separate explicit decision.

### DIFF
--- a/internal/manager/config/stash_config.go
+++ b/internal/manager/config/stash_config.go
@@ -22,21 +22,26 @@ type StashConfig struct {
 type StashConfigs []*StashConfig
 
 func (s StashConfigs) GetStashFromPath(path string) *StashConfig {
-	for _, f := range s {
-		if fsutil.IsPathInDir(f.Path, filepath.Dir(path)) {
-			return f
-		}
-	}
-	return nil
+	return s.GetStashFromDirPath(filepath.Dir(path))
 }
 
 func (s StashConfigs) GetStashFromDirPath(dirPath string) *StashConfig {
+	var ret *StashConfig
+	longestPath := -1
+
 	for _, f := range s {
-		if fsutil.IsPathInDir(f.Path, dirPath) {
-			return f
+		if f == nil {
+			continue
+		}
+
+		path := filepath.Clean(f.Path)
+		if fsutil.IsPathInDir(path, dirPath) && len(path) > longestPath {
+			ret = f
+			longestPath = len(path)
 		}
 	}
-	return nil
+
+	return ret
 }
 
 func (s StashConfigs) Paths() []string {
--- a/internal/manager/config/stash_config.go
+++ b/internal/manager/config/stash_config.go
@@ -21,10 +21,12 @@ type StashConfig struct {
 
 type StashConfigs []*StashConfig
 
+// GetStashFromPath returns the most specific stash configuration containing path.
 func (s StashConfigs) GetStashFromPath(path string) *StashConfig {
 	return s.GetStashFromDirPath(filepath.Dir(path))
 }
 
+// GetStashFromDirPath returns the most specific stash configuration containing dirPath.
 func (s StashConfigs) GetStashFromDirPath(dirPath string) *StashConfig {
 	var ret *StashConfig
 	longestPath := -1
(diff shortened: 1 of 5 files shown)


internal/manager/config/stash_config.go
```diff
@@ -22,21 +22,26 @@ type StashConfig struct {
 type StashConfigs []*StashConfig
 
 func (s StashConfigs) GetStashFromPath(path string) *StashConfig {
```
WithoutPants, line 25: Let's document these functions to explicitly state that the most specific path will be chosen.

WithoutPants (changes requested):
> So this patch does not introduce the possible .stashignore discrepancy from nothing. It resolves order-dependent media-type behavior by deterministically choosing the child library config. If parent .stashignore files are intended to cascade across nested library entries, then the selected library config probably should not also be used as the .stashignore root without a separate explicit decision.

I think `.stashignore` files should be unaffected by the vagaries of the library configuration. The root path is only supposed to be used as a bounds to prevent searching unrelated parent directories. As such, I think in this case, the `stashIgnoreFilter.Accept` function should use the shortest path, rather than the longest.

CynicalAtropos:
Updated this to separate the two path-matching cases:

- `GetStashFromPath` / `GetStashFromDirPath` now explicitly document that they return the most specific library config.
- `.stashignore` now uses the topmost containing library root from `GetStashRootFromDirPath` as its traversal bound, so parent `.stashignore` files continue to cascade through nested library entries.
- Added tests for order-independent topmost root selection and for scan/clean honoring a parent `.stashignore` under a nested library.

Validation:
- `go test ./internal/manager/config`
- `go test ./internal/manager`
- `git diff --check`
- Manually tested .stashignore with nested libraries to confirm they cascade regardless of the library config.  Also just double checked no regression with the library type exclusions in nested libraries.

WithoutPants (approved):

PR #7078, opened 2026-06-27
Add text filter to sort by select

WithoutPants:
## Description

There is a growing number of sort by options, and the list is becoming very long. This PR adds a text filter to the sort by select dropdown. Incidentally moves the sort by selector into its own file. Text field is focused when sort by dropdown is opened on non-touch devices.

Currently, the filter does _not_ clear after selecting or closing the dropdown. I can change this here or in a follow up PR based on user testing.

## Testing

- opened sort by select dropdown and ensured that entering text into the field filters the available options.
- selected items, and closed dropdown without selecting to test filter text persistence between opening

## Screenshots

Before:
image
After:
image

## AI Usage Disclosure

## Additional Context
I have a follow up branch to pin options to the top of the list.

### DIFF
--- a/ui/v2.5/src/components/List/styles.scss
+++ b/ui/v2.5/src/components/List/styles.scss
@@ -1241,5 +1241,5 @@ ul.selectable-list {
       background-color: unset;
       border: unset;
     }
-  }  
+  }
 }
--- a/ui/v2.5/src/components/List/styles.scss
+++ b/ui/v2.5/src/components/List/styles.scss
@@ -1214,3 +1214,32 @@ ul.selectable-list {
     top: 0;
   }
 }
+
+.sort-by-select .dropdown-menu {
+  min-width: 250px;
+  padding-top: 0;
+
+  .sort-by-filter-container {
+    background-color: $secondary;
+    border-bottom: solid 1px $textfield-bg;
+    display: flex;
+    padding: 5px;
+    position: sticky;
+    top: 0;
+    z-index: 1;
+
+    .clearable-input-group {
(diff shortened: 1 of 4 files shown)


ui/v2.5/src/components/List/styles.scss
```diff
@@ -1214,3 +1214,32 @@ ul.selectable-list {
...
+    .clearable-text-field {
+      background-color: $textfield-bg;
+      width: 100%;
+    }
+
+    .clearable-text-field-clear {
+      background-color: unset;
+      border: unset;
+    }
+  }
```
CynicalAtropos: Minor note that there is trailing whitespace here

CynicalAtropos (commented):
Did a quick review and test and it seems to work well.  Really appreciate this feature and pinning will be another nice addition.  
My vote regarding whether or not to keep the search text on close is to keep it how it is.  It's easy to misclick or want to swap back and forth and you can always wipe it by navigating to another page or clicking the x if you want.

WithoutPants:
> Did a quick review and test and it seems to work well. Really appreciate this feature and pinning will be another nice addition. My vote regarding whether or not to keep the search text on close is to keep it how it is. It's easy to misclick or want to swap back and forth and you can always wipe it by navigating to another page or clicking the x if you want.

Appreciate the review, testing and feedback :)

ui/v2.5/src/components/List/styles.scss
WithoutPants: Thanks, fixed.

PR #7080, opened 2026-06-27
Add credentials step to initial setup wizard

WithoutPants:
## Description

Adds a step to the setup wizard to setup an initial username and password. This closes some janky behaviour for public-facing systems where (presumably) the security tripwire would be fired after setup since there's no way to set the credentials during the setup phase.

Credentials step occurs after the paths step and before the confirmation step. If populated, username and masked password are shown on the confirmation page. It also includes a warning indicating that login will be required after clicking finish if the credentials are set.

Includes incidental separation of the welcome page into a separate file and route entry, since the login redirect would otherwise lose the welcome page.

## Testing

Created a new system a few times, with and without credentials.

## Screenshots

image

image

### DIFF
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -13,6 +13,9 @@ input SetupInput {
   storeBlobsInDatabase: Boolean!
   "Empty to indicate default - only applicable if storeBlobsInDatabase is false"
   blobsLocation: String!
+  
+  initialUsername: String
+  initialPassword: String
 }
 
 enum StreamingResolutionEnum {
--- a/graphql/schema/types/config.graphql
+++ b/graphql/schema/types/config.graphql
@@ -13,7 +13,7 @@ input SetupInput {
   storeBlobsInDatabase: Boolean!
   "Empty to indicate default - only applicable if storeBlobsInDatabase is false"
   blobsLocation: String!
-  
+
   initialUsername: String
   initialPassword: String
 }
--- a/ui/v2.5/src/components/Setup/Setup.tsx
+++ b/ui/v2.5/src/components/Setup/Setup.tsx
@@ -9,11 +9,7 @@ import {
   InputGroup,
 } from "react-bootstrap";
 import * as GQL from "src/core/generated-graphql";
-import {
-  mutateSetup,
-  useConfigureUI,
-  useSystemStatus,
-} from "src/core/StashService";
+import { mutateSetup, useSystemStatus } from "src/core/StashService";
 import { useHistory } from "react-router-dom";
 import { useConfigurationContext } from "src/hooks/Config";
 import StashConfiguration from "../Settings/StashConfiguration";
@@ -24,9 +20,7 @@ import { FolderSelectDialog } from "../Shared/FolderSelect/FolderSelectDialog";
 import {
   faEllipsisH,
   faExclamationTriangle,
-  faQuestionCircle,
 } from "@fortawesome/free-solid-svg-icons";
-import { releaseNotes } from "src/docs/en/ReleaseNotes";
 import { ExternalLink } from "../Shared/ExternalLink";
 
 interface ISetupContextState {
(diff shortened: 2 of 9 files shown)


ui/v2.5/src/components/Setup/Setup.tsx
```diff
@@ -672,6 +666,110 @@ const StashExclusions: React.FC<{ stash: GQL.StashConfig }> = ({ stash }) => {
...
+        <Form.Group controlId="username">
+          <Form.Label>
+            <FormattedMessage id="login.username" />:
+          </Form.Label>
+          <Form.Control
+            className="text-input"
+            placeholder={intl.formatMessage({
+              id: "login.username",
+            })}
+            value={username}
```
CynicalAtropos, line 766: I think it should probably have `isInvalid={!usernameValid}` to get the bootstrap state on the input (red outline and red exclamation mark with a circle icon)

ui/v2.5/src/components/Setup/Setup.tsx
```diff
@@ -672,6 +666,110 @@ const StashExclusions: React.FC<{ stash: GQL.StashConfig }> = ({ stash }) => {
...
+            <FormattedMessage id="login.password" />:
+          </Form.Label>
+          <Form.Group id="password">
+            <Form.Control
+              className="text-input"
+              type="password"
+              placeholder={intl.formatMessage({
+                id: "login.password",
+              })}
+              value={password}
```
CynicalAtropos: Same here with a `isInvalid={!passwordValid}`

graphql/schema/types/config.graphql
```diff
@@ -13,6 +13,9 @@ input SetupInput {
   storeBlobsInDatabase: Boolean!
   "Empty to indicate default - only applicable if storeBlobsInDatabase is false"
   blobsLocation: String!
+
```
CynicalAtropos: Trailing whitespace

CynicalAtropos (commented):
Performed a quick review and did some testing with auth and no auth.  Also confirmed it works on my dev setup with both vite on :3000 (no auth) and with a built ui on :9999.
It seems like a good feature to have in general for setup.

Not sure if it is a urgent matter, but I do think it might be worth updating DEVELOPMENT.md, specifically the Local development quickstart section, with the extra step and a warning to not enable auth as it locks up the instance and it isn't immediately intuitive how to get out of the state for a new dev.

sleetx:
New users may be confused about internet accessibility. I would suggest clarifying the last sentence to mention that accessing Stash from outside would require a nonstandard setup. Maybe even link the relevant [StashDocs page](https://docs.stashapp.cc/networking/authentication-required-when-accessing-stash-from-the-internet/)

slick-daddy:
Another good idea would be adding an eye icon to right side of the password box. When clicked, it should show the password in plain text. This would prevent mistyping.

graphql/schema/types/config.graphql
WithoutPants: Biome isn't formatting the files in the `graphql` directory, presumably because it is outside the `ui/v2.5` directory. A bit frustrating.

WithoutPants:
Thanks for the feedback!

I've addressed the review feedback and made the following changes:
- added show/hide password button per @slick-daddy suggestion
- added note to the development notes to indicate credentials can't be used with UI in dev mode per @CynicalAtropos suggestion
- slightly reworded the credentials requirement text to hopefully make it clearer

ui/v2.5/src/locales/en-GB.json
```diff
@@ -1459,6 +1459,12 @@
   "seconds": "Seconds",
   "settings": "Settings",
   "setup": {
+    "credentials": {
+      "description": "Enter the username and password you want to use to login to stash system. These can be left blank if you do not want to enter a username/password to access the system. Credentials are required if you are accessing the system from the public internet.",
```
DogmaDragon:
```suggestion
      "description": "Enter the username and password you want to use to login to Stash system. These can be left blank if you do not want to enter a username/password to access the system. Credentials are required if you are accessing the system from the public internet.",
```

PR #7083, opened 2026-06-27
Fix lightbox jumping to a page's last image when crossing a page boundary

### LINKED ISSUE #7082: Lightbox lands on a page's last image when crossing a gallery page boundary

### Describe the bug

In a gallery whose images span more than one page, opening the lightbox and navigating across a page boundary can land on the **last** image of the new page instead of the **first**. With a 77-image gallery (two pages of 40), crossing forward onto page 2 should show its first image — indicator `1 / 37`, header `Page 2 / 2` — but instead it can jump to the last image of the page, `37 / 37`.

It's a timing/race condition, so it's most reliable when navigating quickly and once the adjacent page has already been loaded once (cached). It happens with both the keyboard arrow keys and the on-screen prev/next arrows (and clicking the left/right edges of the image) — they share the same navigation code.

### Steps to reproduce

1. Open a gallery with more than one page of images (default page size 40 — e.g. a gallery of 77 images = page 1: images 1–40, page 2: images 41–77).
2. Open the lightbox at the **last image of page 1** (image 40).
3. Press **Right arrow** → moves to image 41, the first image of page 2 (`1 / 37`, `Page 2 / 2`). ✅
4. Press **Left arrow** → moves back to image 40, the last image of page 1 (`40 / 40`, `Page 1 / 2`). ✅
5. Press **Right arrow** again — quickly.
6. **Observed:** the lightbox jumps to the **last** image of page 2 (image 77), indicator `37 / 37`, instead of the first image (`1 / 37`).

Doing steps 3–5 in quick succession makes it reproduce reliably. The same happens when clicking the on-screen ‹ › arrows or the image edges instead of the keyboard.

### Expected behaviour

Crossing forward to the next page should always land on that page's **first** image (image 41 — `1 / 37`, `Page 2 / 2`); crossing backward should land on the previous page's **last** image. Navigation should never jump to the far end of the page.

_No response_

### Stash version

develop @ 10c0994e

### Device details

- Manual reproduction: Safari on macOS 26.5.1 (Apple Silicon).
- Also exercised under Playwright 1.61.1 on the desktop Chromium, Firefox and WebKit engines. The cause is a React render-timing race, so it isn't specific to any one browser engine (WebKit is the same engine as the manual Safari run).

### Relevant log output

```shell

```

void-function865:
## Description

In a paginated gallery's lightbox, navigating across a page boundary could land on the **last** image of the new page instead of the first (e.g. on a 77-image gallery the indicator jumps to `37 / 37` / `Page 2 / 2`, the last image of page 2, rather than `1 / 37`, its first). It reproduces when navigating quickly across the boundary — forward, back, then forward again — and once the adjacent page is cached, with both the keyboard and the on-screen arrows.

Two compounding races in `Lightbox.tsx` cause it:

1. `handleLeft`/`handleRight` run on raw keydown/click events, which can arrive faster than React re-renders, so they read a **stale** `isSwitchingPage` (and `index`) from their closure — the page-switch guard didn't actually block the rapid presses/clicks.
2. `pageCallback(...)` was called **before** the handler set its guard and reset the index. When the adjacent page is already cached, it swaps `images` in *synchronously*, producing a render with the old index against the new, shorter page; the index-range effect then clamped that stale index to `images.length - 1` — the last image.

The fix reconciles the index off the **`page` prop** instead of the timing-sensitive switch flag. The parent updates `page` and `images` together, so a `page` change is a race-free signal to land on the new page's first image (forward) or last (backward). The out-of-range clamp now only runs for a genuine same-page shrink (e.g. a deleted image). A synchronous ref still mirrors `isSwitchingPage` solely to drop rapid inputs mid-switch, and it (plus the target index) is now set **before** `pageCallback` so a cached page can't outrun it.

## Related Issue

Fixes #7082 

## Testing

Manually, in a real browser, on a gallery of 77 images (two pages of 40):

- Open the lightbox at image 40 (last of page 1), then **forward → back → forward** across the boundary, quickly. Before: jumps to the last image of page 2, image 77 (`37 / 37`). After: lands on the first image, image 41 (`1 / 37`), every time.
- Repeated with the on-screen ‹ › arrows (and image-edge clicks) — same result; the path is shared with the keyboard.
- Checked single forward and backward crossings, multi-press runs, and rapid mashing at a range of speeds — always lands on the page's first/last image, never the far end.
- Verified normal navigation, opening at a specific image, and the slider→close→reopen flow are unchanged.
- Reproduced/verified on Chromium, Firefox and WebKit.

## Screenshots

not applicable

## Checklist

## AI Usage Disclosure

I used [Claude Code](https://claude.ai/code) as a coding assistant for this fix. It helped reproduce the race in a real browser, trace the render-by-render cause (the navigation handlers reading stale state on rapid input, and an already-cached page swapping its images in synchronously before the index reset — after which the index-range guard clamped the stale index onto the shorter page), and implement the page-keyed reconciliation. I directed the investigation and the approach, reviewed all changes, and manually verified the behavior on a running instance — crossing the page boundary forward/back/forward with both the keyboard and the on-screen arrows.

## Additional Context

Single commit, one file (`ui/v2.5/src/hooks/Lightbox/Lightbox.tsx`); removes the now-unused `oldImages` ref and the old switch-clear effect, folding their job into the page-keyed index reconciliation. No schema or API changes.

### DIFF
--- a/ui/v2.5/src/hooks/Lightbox/Lightbox.tsx
+++ b/ui/v2.5/src/hooks/Lightbox/Lightbox.tsx
@@ -144,7 +144,19 @@ export const LightboxComponent: React.FC<IProps> = ({
   const lastDKeyTime = useRef<number>(0);
   const [navOffset, setNavOffset] = useState<React.CSSProperties | undefined>();
 
-  const oldImages = useRef<ILightboxImage[]>([]);
+  // An in-flight page switch's intended landing, set synchronously by the nav
+  // handlers / chapter jump *before* pageCallback. The landing is an explicit
+  // target — "first"/"last"/a chapter index — NOT derived from the page-number
+  // direction (a wrap moves the page number opposite to nav direction, so
+  // deriving from direction breaks first/last wraparound). It is resolved to a
+  // concrete index by the settle effect once the new page has loaded.
+  const pendingTarget = useRef<"first" | "last" | number | null>(null);
+  // The page number at switch start. The settle effect treats the switch as
+  // complete when the page number changes (the parent swaps page + images in one
+  // update) — a reliable trigger, unlike comparing the images array identity,
+  // which also changes on same-page refetch/reopen and could fire the settle
+  // early, consuming the switch before the new page arrived.
+  const switchFromPage = useRef<number | undefined>(page);
 
   const [zoom, setZoom] = useState(1);
 
@@ -266,25 +278,37 @@ export const LightboxComponent: React.FC<IProps> = ({
     useState<string>((slideshowDelay / SECONDS_TO_MS).toString());
 
   useEffect(() => {
-    if (images !== oldImages.current && isSwitchingPage) {
-      // The new page's images have arrived. Resolve the handler's intended
-      // landing: a -1 sentinel (set by a backward cross or a backward wrap)
-      // means the new page's last image; any other value is the handler's own
-      // target — 0 for a forward cross/wrap, indexInPage for a chapter jump —
-      // and is kept as-is. Clear the ref synchronously so the index-range
-      // effect can resume guarding same-page shrinks.
-      if (index === -1) setIndex(images.length - 1);
-      // Pin the baseline to the page we just settled. Without this, on engines
-      // where the parent's new `images` and the handler's local setIndex(-1)
-      // land in *different* commits, this effect can fire on the new images
-      // before the -1 has applied, clear isSwitchingPage early, and then the -1
-      // renders (a stray "0 / N"). Re-pinning makes the guard below false until
-      // the next switch actually swaps the array.
-      oldImages.current = images;
+    const target = pendingTarget.current;
+    if (target !== null) {
+      // A page switch is in flight. It completes when the page NUMBER changes:
+      // the parent updates page + images together, so a changed page number
+      // means the new page's images are present. Resolve the handler's explicit
+      // landing then — "last"/"first"/a chapter index — and only then drop the
+      // switch flags. The target is never derived from the page-number
+      // direction, so first/last wraparound and chapter jumps land on the right
+      // image regardless of which way the page number moved.
+      if (page !== switchFromPage.current) {
+        setIndex(
+          target === "last"
+            ? images.length - 1
+            : target === "first"
+              ? 0
+              : target
+        );
+        pendingTarget.current = null;
+        switchFromPage.current = page;
+        isSwitchingPageRef.current = false;
+        setIsSwitchingPage(false);
+      }
+      return;
+    }
+    // No switch pending: clear the switching flag once the first page's images
+    // have loaded (the flag starts true on open so the loader shows until then).
+    if (isSwitchingPage && !isLoading && images.length > 0) {
       isSwitchingPageRef.current = false;
       setIsSwitchingPage(false);
     }
-  }, [isSwitchingPage, images, index]);
+  }, [isSwitchingPage, isLoading, images, page]);
 
   const disableInstantTransition = useDebounce(
     () => setInstantTransition(false),
@@ -394,7 +418,7 @@ export const LightboxComponent: React.FC<IProps> = ({
 
   const handleLeft = useCallback(
     (isUserAction = true) => {
-      if (isSwitchingPageRef.current || index === -1) return;
+      if (isSwitchingPageRef.current) return;
 
       if (disableAnimation) {
         setInstant();
@@ -405,16 +429,22 @@ export const LightboxComponent: React.FC<IProps> = ({
 
       if (index === 0) {
         // go to previous page, or loop back if no callback is set. Raise the
-        // guard and set the target index *before* pageCallback: an
+        // guard and record the landing intent *before* pageCallback: an
         // already-cached page can swap its images in synchronously, so the
         // guard must be up first or a rapid input (and the index-range effect)
-        // races the swap. -1 is the sentinel the settle effect resolves to the
-        // new page's last image.
+        // races the swap. "last" is resolved by the settle effect to the new
+        // page's last image once the page number changes — no index is pre-set,
+        // so there is no transient sentinel index to leak into render.
         if (pageCallback) {
           isSwitchingPageRef.current = true;
           setIsSwitchingPage(true);
-          setIndex(-1);
-          oldImages.current = images;
+          switchFromPage.current = page;
+          pendingTarget.current = "last";
+          // Park the index at a value that is valid on any non-empty page (0)
+          // while the new page loads. The settle effect moves it to the target
+          // once the page arrives; parking it keeps it in range so the
+          // index-range clamp can't fire on a shorter incoming page.
+          setIndex(0);
           pageCallback({ direction: -1 });
         } else setIndex(images.length - 1);
       } else setIndex((index ?? 0) - 1);
@@ -423,7 +453,7 @@ export const LightboxComponent: React.FC<IProps> = ({
         resetIntervalCallback.current();
       }
     },
-    [images, pageCallback, index, disableAnimation, setInstant]
+    [images, pageCallback, index, page, disableAnimation, setInstant]
   );
 
   const handleRight = useCallback(
@@ -438,14 +468,15 @@ export const LightboxComponent: React.FC<IProps> = ({
       setShowChapters(false);
 
       if (index === images.length - 1) {
-        // go to next page, or loop back if no callback is set. Guard and target
-        // index set before pageCallback (see handleLeft): land on the new
+        // go to next page, or loop back if no callback is set. Guard and landing
+        // intent recorded before pageCallback (see handleLeft): land on the new
         // page's first image, even when a cached page swaps in synchronously.
         if (pageCallback) {
           isSwitchingPageRef.current = true;
           setIsSwitchingPage(true);
-          setIndex(0);
-          oldImages.current = images;
+          switchFromPage.current = page;
+          pendingTarget.current = "first";
+          setIndex(0); // keep in range while the page loads (see handleLeft)
           pageCallback({ direction: 1 });
         } else setIndex(0);
       } else setIndex((index ?? 0) + 1);
@@ -454,7 +485,7 @@ export const LightboxComponent: React.FC<IProps> = ({
         resetIntervalCallback.current();
       }
     },
-    [images, pageCallback, index, disableAnimation, setInstant]
+    [images, pageCallback, index, page, disableAnimation, setInstant]
   );
 
   const firstScroll = useRef<number | null>(null);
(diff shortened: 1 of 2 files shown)


CynicalAtropos:
Did some testing and was not able to replicate the original issue when I clicked on the gallery details and then entered lightbox (I believe this is expected unless you were also seeing it there).  However, I was able to replicate from the galleries page clicking the magnifying glass, but I didn't see a race condition based on rapid input. No matter how slow, it always went from 1/2 40/40 to 2/2 38/38 if I used keyboard and went to 2/2 1/38 with clicking on the right arrow. 

However, I think the change to using the page change as the signal is likely not the right direction as it introduces several regressions. Specifically, when chapter navigation changes pages it resets the index incorrectly and wraparound navigation is broken (i.e. go backwards from 1/40 and it will land on 1/34 or going from 34/34 to the right lands on 40/40).

I did a quick test patch where I limited it to changing the ordering around pageCallback (which by itself fixed the issues I was seeing) and then using a ref guard to block potential rapid input race conditions.  This means the target index remains in the control of the action with the user's intent instead of just page changing.

It's possible I am missing something especially since I couldn't seem to replicate the rapid input issue so feel free to try out my test fix here: https://github.com/CynicalAtropos/stash/tree/fix/lightbox-simple-race-fix

void-function865:
Thanks for the careful review @CynicalAtropos - you're right on both counts, and I've reworked the fix (pushed on top of the reviewed commit so the history stays intact).

**On the regressions:** keying the landing index off the `page`-number direction was the wrong signal. On a *wrap* the page number moves opposite to the navigation: going back from the first image of page 1 wraps the page *forward* to the last page, so the reconciliation lands on that page's first image instead of the gallery's last; forward from the last image is the mirror. Chapter jumps had the same problem - `gotoPage` picks a specific in-page index and the page-keyed effect overrode it.

**Reworked along the lines you suggested:** the landing index is back under the handlers' control - the `-1` sentinel resolves to the new page's last image (backward), `0` to its first (forward), and `indexInPage` is preserved for chapter jumps. The two races are closed directly instead:

1. A synchronous `isSwitchingPageRef` gates `handleLeft`/`handleRight`, so rapid input during a switch is dropped even when events arrive faster than React re-renders.
2. The guard ref and the target index are set **before** `pageCallback`, so an already-cached page that swaps its images in synchronously can't outrun them.
3. The index-range clamp is gated on that same ref, so it can never clamp a stale index while a switch is in flight - it now runs only for a genuine same-page shrink (a deleted image). That's the one thing I added beyond the reorder: in tracing it, the reorder alone closes the window for the key/click paths because the parent delivers the new `images` a commit after the handler's own `setIndex`, but that's a commit-ordering coincidence rather than a guarantee; gating the clamp makes the clamp-to-last impossible by construction.

**On not being able to reproduce the rapid-input race:** I actually ran into this one the other way round - I hit it first in normal manual use, mashing forward → back → forward across the boundary fast in my own browser and watching it land on the last image of the new page. So it's real and it does happen by hand; it's just very timing-dependent. It needs two things at once: the adjacent page already cached (so its images swap in synchronously) **and** input faster than a React render, so a second press lands while the index is still stale. "No matter how slow" is exactly the condition that *avoids* it - slow input gives each press its own render and always resolves cleanly, which is probably why it didn't show for you. And when it is in range it's a dead heat - the handler's correct `setIndex` and the clamp's stale `setIndex(last)` race, and which one wins depends on cache state and input speed - so eyeballing the final indicator isn't reliable; you have to assert that the clamp branch never wins.

That's where the testing setup comes in. I've been trialing [Playwright](https://playwright.dev) end-to-end tests for the web UI (driving Chromium, Firefox and WebKit), and it's what let me pin both of these down deterministically:

- **Your wraparound regression** - backward wrap from the first image must land on the gallery's last image, forward wrap from the last on its first - reproduced against the page-keyed version and now passing.
- **The original #7082 race** - the forward → back → forward round-trip with the adjacent page pre-warmed (cached), driven at several sub-150 ms inter-press gaps, asserting it lands on the new page's first image and **never** the clamp-to-last. It reproduces on unpatched develop and passes with the rework, on all three engines.

For what it's worth, the library I'm testing against isn't lightweight either - the gallery is a single ~531 MB zip of 77 images (two pages of 40 + 37), with large images: ~3.3k–14.2k px per side, many in the 100–200 MP range, ~7 MB on average and up to ~20 MB each. I'd expect a heavy library to make the race *easier* to trigger: the metadata swap that drives the bug is React-render timing (independent of file size), but decoding and painting images that large puts long tasks on the main thread, which can delay React committing the index reset and stretch the exact window the clamp races in - and the lightbox sits in its switching state a little longer, so fast input is more likely to overlap it. The necessary conditions are the same either way (adjacent page cached + input faster than a render); image weight just widens the window.

If you'd like to look, the two specs are on a branch in my fork: [`lightbox-wraparound.spec.ts`](https://github.com/void-function865/stash/blob/void/e2e-trial/e2e/lightbox-wraparound.spec.ts) (your regression) and [`lightbox-boundary-race.spec.ts`](https://github.com/void-function865/stash/blob/void/e2e-trial/e2e/lightbox-boundary-race.spec.ts) (the #7082 round-trip).

One caveat: there's no committed fixture for this yet - the tests currently run against a specific local library - so it isn't something I'm proposing to add to the PR. But if the Playwright e2e trial keeps proving useful for exactly this kind of timing/interaction bug that the unit and type checks can't catch, I'd like to float a proposal to add a proper fixtured UI-test setup to the project separately.

Thanks again - the test branch pointed straight at the right fix.

CynicalAtropos:
Retested the updated fix and it passed everything I tried including the previously mentioned regressions.  Looks good to me.
 
Also, the playwright e2e tests seem like a great idea.  These type of ui conditions can be hard to catch regressions of while manually testing if they have tight requirements and you aren't specifically trying for them.  I took a look and may try running them later to try them out.

void-function865:
@CynicalAtropos — follow-up commit pushed (`5cfbefc1`) that fixes a regression I introduced in the rework, while keeping the handler-controlled landing you asked for.

### The bug in the reworked version

The rework keyed the "new page has arrived, resolve the landing" signal on the **images array identity** (`images !== oldImages.current`). That turns out to be unreliable: the pinned reference can be stale relative to the live `images` at the moment a switch begins. Two ways that happens:

- the initial open arrives with an empty list and then the real page, which consumes the initial `isSwitchingPage` flag on the empty array; and
- reopening the lightbox leaves `oldImages` pinned to a previous page's array.

When the reference is stale, the settle effect fires **prematurely on the old page**, keeps the old index, and clears `isSwitchingPage`. By the time the new page actually swaps in, the flag is already `false`, so the `-1` landing sentinel is never resolved — it leaks straight to render as a blank image, with the counter showing `(page-1)*pageSize + 0` (e.g. a backward wrap from the first image shows `40/77` instead of `77/77`, with no image).

It only reproduces after the lightbox has been opened/closed/reopened with a page cached, which is why the first round of testing missed it.

### The fix

Trigger the settle on the **page number changing** instead of array identity. The parent updates `page` and `images` together, so a changed page number is a reliable signal that the new page is present — and unlike the array reference it doesn't change on a same-page refetch/reopen.

Crucially, this is **not** a return to deriving the index from the page-number direction (the thing that broke wraparound originally). The landing is still **handler-controlled**: each nav handler / chapter jump records an explicit target (`"first"` / `"last"` / the chapter's `indexInPage`) before `pageCallback`, and the settle effect resolves that target once the page changes. So first/last wraparound and chapter jumps land correctly regardless of which way the page number moved. No `-1` sentinel is pre-set; the index is parked at `0` (valid on any non-empty page) during the switch so the index-range clamp can't fire on a shorter incoming page.

### Verification

Driven with the Playwright trial harness across Chromium/Firefox/WebKit:

- boundary race (forward → back → forward, page 2 cached) at several inter-press gaps → always the new page's first image, clamp-to-last never fires;
- backward wrap (incl. the page-2-cached reopen case that exposed this) → global last image, with the image shown;
- forward wrap → global first image;
- reopen and chapter navigation unaffected.

WithoutPants:
This change currently breaks the keyboard navigation. Pressing right jumped the page (not image) forward from 1 to 3, then to 5, then it stopped working. Pressing left put the page back to 3, then it wouldn't do anything.

void-function865:
@WithoutPants, I cannot reproduce your issue yet. Drove it with Playwright in Chromium/Firefox/WebKit — single presses across several boundaries, held-key bursts, a second press mid-switch, `findImages` delayed 900ms, CPU throttled 6×, and a 77-image/100-200MP gallery — and also tried by hand in all three browsers. Arrows step one image at a time, cross correctly, and the switch never wedges. Tested at `5cfbefc1` (this branch) and merged onto current develop `d4e02f75` — both clean.

To narrow it down:

1. Were you tapping the arrow keys one at a time, or holding them down (OS key-repeat)?
2. How big is the gallery — roughly how many images, what page size, and are the images large?
3. Which browser did you see it in?
4. When it stopped responding, did a page reload bring it back, or was it wedged until you closed the lightbox?

void-function865:
@WithoutPants — I still can't reproduce your sequence, and I've tried hard to. Rather than keep guessing at it silently, here's exactly what I tested, plus a latent issue I found and hardened along the way. I've pushed the changes in `1eee9759`; I can't claim they're your bug, but a couple of them are worth having regardless.

### What I tried (no repro)

Across Chromium, Firefox and WebKit, and by hand in a real browser:

- single arrow presses across many boundaries, and held keys / OS auto-repeat;
- a **very small page size** (1 and 2) so that *every* keypress crosses a page boundary — the regime where a page-skip would show up most easily;
- the heavy library from my earlier comment — the 77-image gallery with 100–200 MP images — at page size 1, so each cross carries a full large-image decode;
- `findImages` delayed, and CPU throttled 6×.

In every case the counter tracks the presses exactly: N presses advance N pages, one at a time, and navigation never wedges. The page never jumps by two on a single press, and the guard never sticks. So I can't confirm what you saw, and I don't want to pretend I've diagnosed it.

### What I changed anyway

Reviewing the settle path, I did find one genuinely fragile spot: the switch guard cleared as soon as the **page number** changed, on the assumption that the new page's images arrive in the same update. That holds for the in-tree callers today, but nothing enforces it — a caller that advanced the page a commit before its images would settle the switch against the old page and re-admit input mid-switch. I've tightened it to settle only when the page number **and** the images array have both changed, with a fallback for callers that drive the lightbox without a page prop. It's defensive — it doesn't change behaviour on any current path — but it removes an assumption that could bite exactly this kind of race.

The same commit also has a few concrete fixes that stand on their own, independent of your report:

- the `d d` delete shortcut no longer fires mid page-switch (the index is parked then, so it could target an image you weren't looking at);
- the delete dialog captures its target image when it opens instead of reading it back at confirm time, so a navigation while it's open can't retarget it;
- a stale chapter target is clamped to the incoming page;
- a missing `.catch` on the gallery page fetch.

### If you can still trigger it

A retest of `1eee9759` would tell us whether it's gone or untouched. And if it still reproduces, the four questions from before would let me chase it properly: were you tapping or holding the key; roughly how many images, what page size, and how large; which browser; and did a reload clear it or was it wedged until you closed the lightbox?

WithoutPants:
I was able to reproduce briefly in `b5bb99db` (but was unable to reproduce reliably) but not in `1eee975`. 

To reproduce:
- navigate to images page
- set page size to 5
- open lightbox by clicking magnifier icon on an image
- press right arrow button

Testing with the current version seems ok, so I think this is good to go. I think for future work we need to simplify the lightbox design. It has too many moving parts and points of failure - mostly due to the connection with the underlying query page and the image animations.

WithoutPants (approved):

PR #7085, opened 2026-06-27
Replace security tripwire mechanism

### LINKED ISSUE #6474: Allowlist IPs for security tripwire
### Describe the bug

Since switching my home network to fully IPv6, I've noticed I have to set credentials on my stash even for local access from some devices.

When I use the domain name ending with `.local` that is offered by my router, devices look this up using IPv6 and then make connections using their public IP address, not their `fe80::/10` link-local address.

This, of course, triggers the `security_tripwire_accessed_from_public_internet` because these are not local IP addresses.

I can work around this by setting a password or by setting the `dangerous_allow_public_without_auth` to `true`, but I'd love a middle ground where I can set the `dangerous_allow_public_without_auth` to a specific IPv6 subnet, corresponding to the subnet that my router allocates, as these are still devices that are accessing it locally, not over the internet.

### Steps to reproduce

1. Have an IPv6 home network
2. Remove any credentials
3. Access your stash locally over DNS using http://server.namelocal:9999/

### Expected behaviour

I would expect to be able to access it without hitting the tripwire, even if it requires me to manually specify the local subnet somehow.

_No response_

### Stash version

v0.30.1

### Device details

Chrome on Android

### Relevant log output

```shell

```

WithoutPants:
## Description

Ideally should follow merging of #7080. 

Replaces the existing tripwire mechanism, which would shut down the system if it had no credentials configured and it was accessed from a public IP address (unless `dangerous_allow_public_without_auth` flag was set).

Instead, if no credentials are configured, the server will reject any requests from public IP addresses with a 403 error.

Public IP addresses may be whitelisted using the `public_whitelist` config field, which accepts IP address ranges in CIDR format (eg: `192.168.1.0/24`).

Also added a `trusted_proxies` config field to allow for configuring proxies from non-local IP addresses. Requests with `X-Forwarded-For` header will be rejected if the requests comes from a non-local non-trusted IP address.

The `dangerous_allow_public_without_auth` flag and `security_tripwire_accessed_from_public_internet` flags have been removed.

## Related Issue

Resolves #6474 

## Testing

Manual testing by hard-coding `api.isLocalIP` to return false, then tried accessing the server with and without credentials configured. 

**Trusted proxy setting has not tested.**

## Additional Context

There is potentially scope for adding an IP whitelist setting that gets applied even when credentials are configured.

I may restore the `dangerous_allow_public_without_auth` flag if the whitelist option isn't a workable solution for users that know what they are doing.


DogmaDragon (approved):
Documentation check passed.

Note to myself: External tripwire FAQ document needs to be updated after release.

PR #7022, opened 2026-06-07
New: Add delete image shortcut and button to lightbox

void-function865:
## Description

Adds the ability to delete the current image directly from the lightbox viewer without having to close it and return to the grid.

Two interaction paths are supported:

- **`d d` keyboard shortcut** — double-tap `d` within one second to open the delete confirmation dialog. Consistent with the existing `d d` shortcut on the images page (grid and wall views).
- **`Enter` to confirm** — once the delete confirmation dialog is open, pressing `Enter` confirms the deletion. This completes a fully keyboard-driven culling flow: `d d` to open, `Enter` to confirm.
- **Trash icon button** — a delete button is shown next to the image filename in the lightbox footer, using the same `danger` styling as the delete button in the images page toolbar. This goes beyond the scope of the issue, but provides a touch-accessible path for mobile users where keyboard shortcuts are unavailable. A swipe-to-delete gesture would be an alternative mobile interaction (swipe-to-delete is a [frequently requested feature in comparable apps such as Immich](https://github.com/immich-app/immich/discussions/24059) as well), and I am happy to implement that instead if preferred.

Both paths open the existing `DeleteImagesDialog`, which gives the user the option to also delete the underlying file and/or generated supporting files, matching the behaviour of every other delete flow in the application.

After confirming deletion the lightbox automatically advances to the next image, or closes if the deleted image was the last one.

## Related Issue

Closes #4512

## Testing

Tested manually in Safari, Firefox, and Chrome.

1. Opened a gallery and entered the lightbox.
2. Pressed `d` once — nothing happened.
3. Pressed `d` a second time within one second — the delete confirmation dialog appeared showing the current image filename.
4. Pressed `Enter` — deletion was confirmed, image removed, lightbox advanced to the next image.
5. Triggered the dialog again and clicked **Cancel** — dialog closed, image was still present, lightbox remained open.
6. Repeated until only one image remained and confirmed deletion — the lightbox closed.
7. Clicked the trash icon in the footer next to the filename — the same confirmation dialog appeared.
8. Verified the **Delete file** and **Delete generated supporting files** checkboxes behave as expected and respect the defaults configured in Settings.

## Screenshots

**Before** — no delete action available in the lightbox:
before

**After** — trash icon next to filename, `d d` + `Enter` shortcut supported:
after

---

*Screenshot images: Photo by [Oskar Kadaksoo](https://unsplash.com/@oskark) on [Unsplash](https://unsplash.com/photos/V_FR5YYMfVE)*

## Checklist

[...]

PR #7088, opened 2026-06-27
Add Rating Keyboard Shortcuts for Images

### LINKED ISSUE #5616: [Feature] Implement ratings keyboard shortcut for images in lightbox mode
When viewing images in lightbox, the rating keyboard shortcuts, _r + n{0-5}_, have not been implemented.

Implement rating keyboard shortcuts for lightbox mode

FR created on behalf of https://github.com/stashapp/stash/discussions/5614#discussioncomment-11886486

void-function865:
## Description

This implements the rating keyboard shortcuts (`r` then a digit) for images in the lightbox, honoring whichever rating system is configured and matching the scene and image detail pages:

- **Stars** — `r 1`–`r 5` set the rating, `r 0` clears it.
- **Decimal** — `r {0-9}{0-9}` sets the rating (`r 0 0` for `10.0`), `` r ` `` clears it.

Key design choices:

- **Reuse `useRatingKeybinds` via a lightbox-scoped Mousetrap instance.** The lightbox calls `Mousetrap.pause()` while open to suppress global shortcuts behind the overlay, which is why the existing Mousetrap-based hook never fired there. Rather than re-implement the rating logic by hand, the lightbox now owns its own `new Mousetrap()` — the `mousetrap-pause` plugin tracks the paused flag per instance, so a separate instance keeps firing while the global singleton stays paused. The hook gains an optional Mousetrap-instance parameter that defaults to the global singleton, so the scene and image detail pages are unchanged, and the lightbox reuses the hook verbatim instead of duplicating its logic.
- **Move the lightbox's `d d` delete onto the same shared mechanism.** The delete shortcut now uses the standard `Mousetrap.bind("d d", …)` sequence used everywhere else in the app, on the lightbox's own instance, replacing the bespoke timestamp-based double-tap detector that only existed because the global Mousetrap was paused. I added that detector in #7022; folding its removal in here keeps both lightbox sequence shortcuts on the shared mechanism rather than bespoke code, and only the activation glue stays per-consumer (the global hook binds digit keys dynamically through Mousetrap; the lightbox feeds the same hook through its scoped instance).
- **Fix a pre-existing unbind-timer bug in `useRatingKeybinds` that the reuse surfaced.** Every `r` press scheduled its own independent timer to unbind the digit keys after one second but never cancelled the previous one, so a quick second sequence (for example `r 3` then `r 4`) whose digit landed after the first timer fired had its keys unbound mid-sequence and the keypress dropped — the rating did not always update. The timer is now cancelled and rescheduled per sequence, and the decimal entry buffer is reset at the start of each sequence. This fix also benefits the scene and image detail pages, which share the hook.

## Related Issue

Closes #5616.

## Testing

Verified against a local instance with both rating systems, plus an automated Playwright suite run in my fork (driving a real browser against a dev build of this branch):

- Lightbox, Stars: `r 3` sets three stars and persists across closing and reopening the image; `r 0` clears it.
- Lightbox, Decimal: `r 3 5` sets 3.5 and persists; `` r ` `` clears it.
- Rapid sequence regression: `r 3` immediately followed by `r 4` ends on four stars (this deterministically failed before the timer fix, with the `4` dropped, and passes after it).
- Existing lightbox shortcuts unchanged: arrow-key navigation, `Escape` to close, `d d` to open the delete dialog, and global shortcuts (such as `g s`) staying inert while the lightbox is open.
- Off the lightbox: global navigation shortcuts still route, and the rating shortcuts (including the rapid-sequence case) still work on the scene and image detail pages, confirming the shared-hook change is backward compatible.

## Screenshots

No new UI.

## Checklist

## AI Usage Disclosure

I used Claude Code (Anthropic) to assist with this PR. It helped explore the lightbox and `useRatingKeybinds` implementations, propose and implement the approach (the scoped Mousetrap instance, the hook parameter, the `d d` dedup, and the unbind-timer fix), and author the Playwright end-to-end tests used to verify it. I directed the design decisions, reviewed all changes, and validated the behavior manually and through the automated tests before submitting.

## Additional Context

This is a front-end-only change with no schema or generated-binding changes. The keyboard shortcuts are also documented in the keyboard shortcuts manual (`KeyboardShortcuts.md`).

### DIFF
--- a/ui/v2.5/src/hooks/keybinds.ts
+++ b/ui/v2.5/src/hooks/keybinds.ts
@@ -2,86 +2,56 @@ import Mousetrap from "mousetrap";
 import { useEffect, useRef } from "react";
 import { RatingSystemType } from "src/utils/rating";
 
+const starRatingShortcuts: { [char: string]: number } = {
+  "0": NaN,
+  "1": 20,
+  "2": 40,
+  "3": 60,
+  "4": 80,
+  "5": 100,
+};
+
+type RatingSequenceMode = "idle" | "star" | "decimal";
+
 export function useRatingKeybinds(
   isVisible: boolean,
   ratingSystem: RatingSystemType | undefined,
   setRating: (v: number) => void,
   mousetrap: Pick<Mousetrap.MousetrapInstance, "bind" | "unbind"> = Mousetrap
 ) {
+  // setRating/ratingSystem are recreated every render by every caller (they
+  // close over the currently displayed entity). Reading them through refs,
+  // updated unconditionally on each render, lets the bind effect below key
+  // only off isVisible/mousetrap while still always acting on the latest
+  // values -- rebinding "r" itself isn't needed to pick up a fresh setRating.
+  const setRatingRef = useRef(setRating);
+  setRatingRef.current = setRating;
+  const ratingSystemRef = useRef(ratingSystem);
+  ratingSystemRef.current = ratingSystem;
+
+  const mode = useRef<RatingSequenceMode>("idle");
   const firstChar = useRef<string | undefined>(undefined);
-  const ratingTimeout = useRef<ReturnType<typeof setTimeout>>();
-
-  // (Re)start the 1s window after each "r"-initiated sequence, cancelling any
-  // pending unbind. Without this, pressing "r" again before the window elapses
-  // leaves the earlier timeout scheduled, which then unbinds the digit keys
-  // mid-sequence and drops the next keypress (e.g. a quick "r 3" then "r 4").
-  function restartRatingTimeout(unbind: () => void) {
-    if (ratingTimeout.current) clearTimeout(ratingTimeout.current);
-    ratingTimeout.current = setTimeout(() => {
-      ratingTimeout.current = undefined;
-      unbind();
-    }, 1000);
-  }
-
-  const starRatingShortcuts: { [char: string]: number } = {
-    "0": NaN,
-    "1": 20,
-    "2": 40,
-    "3": 60,
-    "4": 80,
-    "5": 100,
-  };
-
-  function handleStarRatingKeybinds() {
-    for (const key in starRatingShortcuts) {
-      mousetrap.bind(key, () => setRating(starRatingShortcuts[key]));
-    }
-
-    restartRatingTimeout(() => {
-      for (const key in starRatingShortcuts) {
-        mousetrap.unbind(key);
-      }
-    });
-  }
-
-  function handleDecimalKeybinds() {
-    // start each sequence fresh so a new "r" doesn't combine with a digit left
-    // buffered from a previous, abandoned sequence
-    firstChar.current = undefined;
-
-    mousetrap.bind("`", () => {
-      setRating(NaN);
-    });
-
-    for (let i = 0; i <= 9; ++i) {
-      mousetrap.bind(i.toString(), () => {
-        if (firstChar.current !== undefined) {
-          let combined = parseInt(firstChar.current + i.toString(), 10);
-          if (combined === 0) {
-            combined = 100;
-          }
-
-          setRating(combined);
-          firstChar.current = undefined;
-        } else {
-          firstChar.current = i.toString();
-        }
-      });
-    }
-
-    restartRatingTimeout(() => {
-      firstChar.current = undefined;
-
-      mousetrap.unbind("`");
-      for (let i = 0; i <= 9; ++i) {
-        mousetrap.unbind(i.toString());
-      }
-    });
-  }
+  const sequenceTimeout = useRef<ReturnType<typeof setTimeout>>();
 
   useEffect(() => {
     if (!isVisible) return;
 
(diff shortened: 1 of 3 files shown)


DogmaDragon (approved):
Documentation check passed.

ui/v2.5/src/hooks/keybinds.ts
```diff
@@ -79,7 +97,7 @@ export function useRatingKeybinds(
     });
 
     return () => {
-      Mousetrap.unbind("r");
+      mousetrap.unbind("r");
     };
```
Gykes, line 83: The restart timer is never cleared when `isVisable` goes false. If the user presses r then closes the lightbox within 1s the timeout still fires. Not a massive deal as it probably wont be happening very often but still good to clean it up.

ui/v2.5/src/hooks/keybinds.ts
Gykes: This is preexisting: 

This effect doesnt seem to have a dep array. It seems to unbind/remount on every render. Just another small clean up while we are working on this. Could do : `[isVisible, ratingSystem, setRating, mousetrap]`

Gykes (changes requested):
Static review only, most of this is cleanup/pre existing issues

ui/v2.5/src/hooks/keybinds.ts
Gykes: Hmm, this seems to have used the original page rather than the new page. It is supposed to be `Keybinds.ts:82`

ui/v2.5/src/hooks/keybinds.ts
void-function865: Confirmed — worse than a wasted timer, actually: the lightbox doesn't unmount on close (`isVisible` just toggles an early `return null`), and its scoped Mousetrap instance is never paused (only the global singleton is, by design). So `r 3` followed by closing within 1s left the digit/backtick keys bound on that instance for up to a second after close — a stray `3` keypress anywhere afterward would silently re-fire `setRating`.

Fixed in bd4e5c5f7: the effect's cleanup now cancels the pending timeout and runs the queued unbind immediately, so nothing outlives the effect.

ui/v2.5/src/hooks/keybinds.ts
void-function865: Fixed in bd4e5c5f7, but not with the literal array you listed — running it through this repo's biome `useExhaustiveDependencies` rule flagged that as wrong: the effect doesn't call `setRating` directly, it calls `handleStarRatingKeybinds`/`handleDecimalKeybinds` (which close over it), so biome wants those two in the deps instead of `setRating`. Wrapped both handlers (and the timer helper) in `useCallback` and depend on those; `biome lint` and `tsc --noEmit` are both clean now.

One caveat: this doesn't fully stop the effect from re-running on every render for the six existing scene/image/etc. callers — their `setRating` is a plain function redefined every render there too, not `useCallback`-wrapped, so it still gets a new identity each time. That's a pre-existing pattern across all six of those files and out of scope for this PR; happy to file it separately if useful.

Verified against a locally-built matching backend with the Playwright e2e suite (lightbox rating stars/decimal, the rapid r-sequence boundary regression, d d, arrow/Escape, and the scene/image detail-page rating shortcuts) — all green.

Gykes (approved):
Tested and it worked as expected. No obvious regressions were found. 

From the actual code itself I think how this was done isn't 100% the best way. Now I'm not a TS or front end guy so I could be wrong. I think if we decouple the cancel timer and unbind now from the every render it would be better. I'm not 100% sure on that and I think this is "good enough".

Gykes:
@void-function865 Just now looking it seems like the merge conflict is from more upstream changes to the lightbox. Can you please take a look at those?

void-function865:
Pushed a rewrite addressing the design concern from the last review round.

**What changed in `useRatingKeybinds`:** instead of dynamically binding/unbinding the digit and backtick keys per sequence, `r`, the digits, and `` ` `` are now bound once for the hook's lifetime, and a ref-tracked sequence state (`idle`/`star`/`decimal`) plus a single timeout gate their behavior instead. This fully decouples the digit bindings' lifetime from the "r"-binding effect's render-driven churn: since `setRating` is never memoized by any caller (it closes over the currently displayed image), that effect previously reran on every render, and its cleanup (added to fix the earlier "r 3 then r 4" issue) flushed the pending unbind on *any* re-render, not just a second "r" press — so an unrelated re-render mid-sequence could still drop a digit before the 1s window elapsed. The rewrite removes that coupling entirely rather than special-casing it further.

Also rebased onto current `develop` and resolved the conflict with the page-boundary fix (#7083, merged since): kept `deleteTarget`'s capture-at-press-time and the mid-switch guard, but moved the `d d` binding onto the lightbox-scoped Mousetrap instance per this PR's original design, so both fixes' guarantees hold together.

Manually verified in a running instance, including the specific regression this addresses (an unrelated re-render — e.g. toggling the lightbox options panel, nothing to do with rating — landing mid-sequence no longer drops the pending digit).

PR #7104, opened 2026-07-05
Fix A/V desync after seeking when video is stream-copied

### LINKED ISSUE #7103: Audio desyncs after seeking a live-transcoded stream with copied video but re-encoded audio

### Describe the bug

When a live transcode stream-copies the video while re-encoding the audio, seeking causes an audio/video desync for the rest of playback. The audio runs ahead of the video by the distance between the seek target and the previous video keyframe (up to a full GOP). Files with sparse keyframes (e.g. 10s GOPs) desync by several seconds, which is very noticeable. Playing the same file from the start without seeking stays perfectly in sync.

### Steps to reproduce

First: prepare symptomatic video file. The two properties that matter below are the AC-3 audio (forces the live-transcode path with video copy) and the long GOPs (makes the desync large; short-GOP files have the same defect but the offset is small enough to go unnoticed).

```
curl -LO https://download.blender.org/demo/movies/BBB/bbb_sunflower_1080p_30fps_normal.mp4.zip
unzip bbb_sunflower_1080p_30fps_normal.mp4.zip
ffmpeg -ss 60 -t 150 -i bbb_sunflower_1080p_30fps_normal.mp4 \
  -map 0:v:0 -map 0:a:1 -vf scale=640:360 \
  -c:v libx264 -preset veryfast -crf 30 -pix_fmt yuv420p \
  -x264-params keyint=300:min-keyint=300:scenecut=0 \
  -c:a ac3 -ac 2 -b:a 128k -movflags +faststart bbb_seek_desync_repro.mp4
```

Then: add to stash and play the scene. Seek to e.g. 1:45 (mid-GOP). Audio is ~5 seconds ahead of video from that point on. Seeking to other positions gives a desync between 0 and 10s depending on the distance to the previous keyframe.

### Expected behaviour

Audio and video always stay in sync after seeking.

Quick workaround: adding -noaccurate_seek to the live transcode input args restores sync. As a config option it applies to all live transcode invocations, which is broader than needed. Note that with this change, the scrubber can land up to one GOP before the requested position. This is the same keyframe-aligned seeking many video players perform by default.

### Stash version

v0.31.1

### Device details

Replicated it using two instances, one on a Fedora server and locally on macOS (using Podman and Docker respectively). Client browser in both instances was Chromium

### Relevant log output

```shell

```

dpusceddu:
## Description
When a live transcode stream-copies the video while re-encoding the audio, ffmpeg's accurate seek trims the decoded audio exactly to the -ss position, but the copied video can only start at the preceding keyframe, leaving the audio ahead of the video by up to a GOP for the rest of playback.

This PR adds -noaccurate_seek to the live transcode args only when the selected video codec is VideoCodecCopy and a start time is set, so both streams start together at the preceding keyframe. Re-encoded streams keep ffmpeg's default accurate seek, and non-seek playback is untouched. The same arg builder serves the MP4, WebM and MKV endpoints, so all copy-capable paths are covered.

## Related Issue
Closes #7103.

## Testing
Built and tested inside the stashapp/stash:v0.31.1 Docker image against the repro file from #7103.

## Screenshots

Not a visual change.

## Checklist

## AI Usage Disclosure

I have used Claude Code with the Fable 5 model to assist me in finding the bug and reasoning about it.

## Additional Context
With this change, the scrubber can land up to one GOP before the requested position. This is the same keyframe-aligned seeking many video players perform by default.

### DIFF
--- a/pkg/ffmpeg/stream_transcode.go
+++ b/pkg/ffmpeg/stream_transcode.go
@@ -202,6 +202,13 @@ func (o TranscodeOptions) makeStreamArgs(sm *StreamManager) Args {
 	args = append(args, extraInputArgs...)
 
 	if o.StartTime != 0 {
+		if codec == VideoCodecCopy {
+			// Stream-copied video can only start at the keyframe preceding the
+			// seek point, while re-encoded audio would be trimmed exactly to it,
+			// leaving the output desynced by up to a GOP. Disable accurate seek
+			// so that both streams start at the keyframe.
+			args = args.NoAccurateSeek()
+		}
 		args = args.Seek(o.StartTime)
 	}
 
--- a/pkg/ffmpeg/stream_transcode.go
+++ b/pkg/ffmpeg/stream_transcode.go
@@ -203,10 +203,7 @@ func (o TranscodeOptions) makeStreamArgs(sm *StreamManager) Args {
 
 	if o.StartTime != 0 {
 		if codec == VideoCodecCopy {
-			// Stream-copied video can only start at the keyframe preceding the
-			// seek point, while re-encoded audio would be trimmed exactly to it,
-			// leaving the output desynced by up to a GOP. Disable accurate seek
-			// so that both streams start at the keyframe.
+			// #7103 - keep audio aligned with copied video, which can only start at a keyframe
 			args = args.NoAccurateSeek()
 		}
 		args = args.Seek(o.StartTime)
--- a/pkg/ffmpeg/options.go
+++ b/pkg/ffmpeg/options.go
@@ -37,6 +37,11 @@ func (a Args) Seek(seconds float64) Args {
 	return append(a, "-ss", fmt.Sprint(seconds))
 }
 
+// NoAccurateSeek adds the -noaccurate_seek flag and returns the result.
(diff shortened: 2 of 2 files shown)


pkg/ffmpeg/stream_transcode.go
```diff
@@ -202,6 +202,13 @@ func (o TranscodeOptions) makeStreamArgs(sm *StreamManager) Args {
 	args = append(args, extraInputArgs...)
 
 	if o.StartTime != 0 {
+		if codec == VideoCodecCopy {
+			// Stream-copied video can only start at the keyframe preceding the
+			// seek point, while re-encoded audio would be trimmed exactly to it,
+			// leaving the output desynced by up to a GOP. Disable accurate seek
+			// so that both streams start at the keyframe.
```
slick-daddy: A bit too verbose in my opinion.

dpusceddu: > A bit too verbose in my opinion.

New commit, I made it as brief as I could + referenced the issue number like I saw in other comments in this codebase

Gykes (approved):

PR #7115, opened 2026-07-18
fix(ci): increase test timeout to prevent flaky pkg/sqlite failures

slick-daddy:
## Description
CI intermittently fails with a hard test timeout on `pkg/sqlite`. I encountered this bug while I was working on my own branch. It happened because of slow runners. Go's default timeout for tests is 10 minutes and my run failed at 11th minute, even though it has no bugs or any other issue. I am adding timeout flag to solve this issue. It doesn't touch `make it` command.

Proof for failed run can be found [here](https://github.com/slick-daddy/stash/actions/runs/29625640460/job/88029616607?pr=27).

Especially see:
> *** Test killed with quit: ran too long (11m0s). (**Fails at 11th minute.**)
> FAIL	github.com/stashapp/stash/pkg/sqlite	660.039s
> ?   	github.com/stashapp/stash/pkg/sqlite/blob	[no test files]
> ?   	github.com/stashapp/stash/pkg/sqlite/migrations	[no test files]
> ?   	github.com/stashapp/stash/pkg/stashbox	[no test files]
> ?   	github.com/stashapp/stash/pkg/stashbox/graphql	[no test files]
> ok  	github.com/stashapp/stash/pkg/studio	0.006s
> ok  	github.com/stashapp/stash/pkg/tag	0.008s
> ?   	github.com/stashapp/stash/pkg/txn	[no test files]
> ok  	github.com/stashapp/stash/pkg/utils	0.010s
> ?   	github.com/stashapp/stash/ui	[no test files]
> FAIL
> make: *** [Makefile:334: it] Error 1
> Error: Process completed with exit code 2.

## Checklist

## AI Usage Disclosure
No AI used.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -113,7 +113,7 @@ jobs:
         name: generated
 
     - name: Test Backend
-      run: make it
+      run: go test -timeout 30m -tags "sqlite_stat4 sqlite_math_functions integration" ./...
 
   # Job 3: Cross-compile for all platforms
   # Each platform gets its own runner and Docker container (ghcr.io/stashapp/compiler:13).
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -113,7 +113,9 @@ jobs:
         name: generated
 
     - name: Test Backend
-      run: go test -timeout 30m -tags "sqlite_stat4 sqlite_math_functions integration" ./...
+      run: make it
+      env:
+        GOTESTFLAGS: -timeout 30m
 
   # Job 3: Cross-compile for all platforms
   # Each platform gets its own runner and Docker container (ghcr.io/stashapp/compiler:13).
--- a/Makefile
+++ b/Makefile
@@ -331,7 +331,7 @@ test:
 .PHONY: it
 it:
 	$(eval GO_BUILD_TAGS += integration)
-	go test -tags "$(GO_BUILD_TAGS)" ./...
+	go test $(GOTESTFLAGS) -tags "$(GO_BUILD_TAGS)" ./...
 
 # generates test mocks
 .PHONY: generate-test-mocks


Gykes:
I'm more inclined to figure out why the tests are taking that long vs 3x the timeout time. I don't think increasing the timeout time is the best bet if GH runners are just slow. I feel like just rerunning the test is better. Having such a large timeout time can hide actual errors down the road, no?

slick-daddy:
> I'm more inclined to figure out why the tests are taking that long vs 3x the timeout time.

I thought the same thing actually but error log doesn't provide anything useful, except running for too long.

>I feel like just rerunning the test is better.

I agree but I run the test 3 times and it is still failing. If the fails are unrelated to timeout (which is possible), we still need to understand why it happens. If related, timeout will solve it.

>Having such a large timeout time can hide actual errors down the road

We can change it to 20 or even 15 minutes.

slick-daddy:
I asked Github Copilot out of curiosity to understand the underlying issue. This is the report and suggested fix:

> Looking at the logs, the tests are being killed after 11 minutes: *** Test killed with quit: ran too long (11m0s).
Reason: Go's default test timeout is 10 minutes. The internal/autotag and pkg/sqlite test suites exceed this limit and are forcibly terminated.
Solution: Update the test command in .github/workflows/build.yml line 116 to increase the timeout.

Gykes:
Well, AI's default option is to either sleep or increase timeouts. It's why, as of now, you can't blindly trust it. Currently on develop I don't have an issue. As a example you and I got very different results when throwing the issue at AI. I tossed this question at Opus 4.8 and got the following result: 

```
I looked at the goroutine dump in that job log, and I don't think this is a slow-runner timeout. I believe it's a deadlock on your branch. A few things point that way:

  1. Both failing packages died at exactly the same time:
  FAIL github.com/stashapp/stash/internal/autotag  660.005s
  FAIL github.com/stashapp/stash/pkg/sqlite        660.039s
  660s is precisely the 11m watchdog kill (Go's 10m -timeout plus the 1m SIGQUIT grace). A genuinely slow suite lands on varying times; a hang lands on the timeout exactly. Note it also wasn't only pkg/sqlite — internal/autotag went down the same way.

  2. The dump names a specific stack, identical in both binaries, goroutine 1 [sync.RWMutex.RLock, 10 minutes]:
  config.(*Config).GetUIConfiguration           config.go:1378   ← blocked on RLock
  config.(*Config).MigrateImageLightboxConfig   config.go:2004   
  migrations.post84                             84_migrate.go:60
  sqlite.(*Migrator).runCustomMigration         
  sqlite.(*Database).Open
  autotag.runTests → TestMain
  MigrateImageLightboxConfig appears to hold the config write lock and then call GetUIConfiguration(), which takes RLock() on that same mutex. Go's sync.RWMutex isn't reentrant, so that's an unconditional self-deadlock. It fires inside db.Open() during TestMain, before
  any test body runs — which is why every package that opens a DB hangs identically. The fix is to call the unlocked internal accessor from inside the migration rather than the public locking getter.

  3. MigrateImageLightboxConfig doesn't exist on develop — post84 there only does the folder migration. So this is specific to your branch, not something CI hits on develop.

  I also ran the full suite against current develop locally (with the Makefile's tags: sqlite_stat4 sqlite_math_functions integration):
  real 9.47s        # entire ./... suite
  ok  pkg/sqlite         1.222s
  ok  internal/autotag   5.285s

  Given that, raising the timeout to 30m wouldn't have unblocked your branch — it would have turned an 11-minute failure into a 30-minute one on every CI run, and it makes real hangs meaningfully more expensive to detect for everyone else.

  That said, the GOTESTFLAGS passthrough itself seems like a reasonable thing to have on its own merits. If you'd like to keep that part, it'd probably want: the same treatment on the test: target for consistency, no -timeout default in the workflow (or a much smaller
  bump than 3×), and a rationale of "let CI tune test flags" rather than "fix flaky failures." Worth checking whether the maintainers would rather just use Go's built-in GOFLAGS env var, which needs no Makefile change at all.
```

Now, I did exactly 0 verification of this. I just know I have had 0 issues with running tests and never had a test hang so I get the feeling there is a smell and increasing the timeout does not seem like the correct answer. If you can give me something more than "Copilot said so" then we can talk about it.

slick-daddy:
After pushing [this](https://github.com/slick-daddy/stash/pull/27/changes/a3508dd8dbb447590259b8af2a11f0d177a901f8#diff-cff70f89f8904970f88b652a4a7f1d0cbd86d1661897200a179c4434ad788e32L2001-L2002) commit, CI runs as it should.

Turns out, there was a deadlock that causes CI to flake. All I did was removing the following two lines:
`i.Lock()`
`defer i.Unlock()`

I don't believe it was a race condition. The bug is worth investigating and makes this PR sort of trivial, hence closing without  merge.

Issue #1789, opened 2021-10-01
[Feature] Support for externally hosted content. eg: S3


stashTX:
I would like to see stash support S3 buckets for content and/or metadata storage.

Perhaps then there could be an option for the client to pull the content from the stash server which would proxy the data from the S3 compatible storage,  or there could be an option for the client to request the content directly from S3 if permissions allowed.

This could also be extended to allow the content to point to a hosted m3u file where the client could request it directly allowing for the leveraging off offsite storage and CDN's.

hamboneZA:
I've got a setup with S3 working well using a local mount with [rclone](https://rclone.org/commands/rclone_mount/).

sacredantelope:
In my Stash setup I've noticed the system completely freezes quite often when scrubbing or switching scenes. I think hosting on S3 and using presigned URLs could do a lot to, as @stashTX said, leveraging the offsite storage and CDN. It would also make it a lot easier to spread Stash over a networked setup and not requiring a ton of hard storage on the server.

Is there any interest from the project to potentially receive a PR for this? Is there any knowledge about the complexity of such a feature or known hurdles? @StashAppDev

DingDongSoLong4:
@sacredantelope Have you tried rclone, as mentioned above? It's a very powerful and supports a ton of storage providers, not just S3. Simply mount a remote to a local path and point stash to it, and everything should just work.

Given that rclone exists, if you want this feature in stash itself, you're probably going to have to do it yourself. But even then, since the changes likely won't be trivial, it will introduce some amount of maintenance burden, so it likely wouldn't be merged for that reason anyway.

All that is to say, is there a reason you're not using rclone? Is it deficient in some way? It really is the most sensible solution here, rather than a change to Stash itself.

stashTX:
rclone does exist and I do use it.

Native support of object storage systems though would be preferred rather than a 3rd party OS level driver/shim.
This would allow this feature to work cross platform.

Additionally it would potentially allow the client to stream the video directly off the object store rather than the stash server itself. One could then have a stash server as essentially the metadata provider and user interface to videos/photos stored in other locations. It is an architectural decoupling and perhaps not trivial but does allow a lot of flexibility.

DingDongSoLong4:
Rclone is quite cross-platform (certainly much more than Stash is), but I see your point.

[...]

PR #7116, opened 2026-07-18
Add S3-compatible blob storage backend

Sl0thC0der:
Adds a third `blobs_storage` type, `S3`, storing blobs (scene covers, performer/studio/tag images, ...) in S3-compatible object storage — AWS S3, MinIO, RustFS, Garage, Backblaze B2, Wasabi, etc. Requested in #1789.

## Usage

```yaml
blobs_storage: S3
s3:
  endpoint: minio:9000        # or s3.amazonaws.com
  access_key: ...
  secret_key: ...
  bucket: stash-blobs
  blobs_prefix: blobs/        # optional key prefix
  region: ""                  # optional
  use_ssl: false
```

The storage type and the S3 connection settings are also editable in the UI under Settings → System (the S3 fields appear when the S3 storage type is selected).

## Design

- Objects are keyed by `prefix + md5-checksum` — the same checksum already stored in the `blobs` table, so the database remains the reference index and identical blobs deduplicate naturally.
- Implemented with `aws-sdk-go-v2` using path-style addressing, so self-hosted S3-compatible services work out of the box.
- The client is created lazily; misconfiguration surfaces as a normal error on first use rather than at startup.
- Reads fall back between stores (S3 → filesystem and vice versa), so switching storage types never breaks existing blobs mid-migration.
- The existing **Migrate blobs** task now migrates between any pair of database/filesystem/S3, and with *delete old* removes the blob from all non-target stores.
- A missing object maps to `fs.ErrNotExist`, matching the filesystem store semantics.
- Config validation requires `s3.endpoint` and `s3.bucket` when the S3 storage type is active (both at startup and when changed via `configureGeneral`).

## Notes

- `DeleteObject` on a missing key is treated as success (idempotent deletes).
- Only the en-GB locale contains the new strings; other locales fall back as usual.

Tested against MinIO and RustFS: fresh writes during scan/generation, serving covers and performer images, and full migrations filesystem→S3 and S3→filesystem including *delete old*.


DogmaDragon:
Please follow the pull request template.

joyov:
Is this still being worked on? Would be great to have S3 support in Stash.

PR #7123, opened 2026-07-22
Fix: Gallery Parent Folder Exclusion SQL Error

### LINKED ISSUE #6993: SQL error when filtering out Parent Folder in Galleries

### Describe the bug

In the picture galleries section, excluding results by parent folder is broken.

This happens for me on and folder or subfolder.

A similar error also occurs when the `Include sub-folders` option is checked.

### Steps to reproduce

1. go to the Galleries page
2. click the Filter icon (funnel)
3. open the Parent Folder section
4. for any folder or subfolder listed, click the minus button to exclude it

### Expected behaviour

Galleries are displayed that do not match the selected folder

I have performed the Maintenance -> Clean function, to no effect. 

### Stash version

v0.31.1 (4de2351e)

### Device details

Any browser.

### Relevant log output

```shell
findGalleries: input: findGalleries error executing count query with SQL: SELECT COUNT(*) as count FROM (SELECT DISTINCT galleries.id FROM galleries LEFT JOIN galleries_files ON galleries_files.gallery_id = galleries.id LEFT JOIN files ON galleries_files.file_id = files.id LEFT JOIN folders AS gallery_folder ON galleries.folder_id = gallery_folder.id WHERE ((files.parent_folder_id NOT IN (SELECT column2 FROM (VALUES(76,76))) OR folders.parent_folder_id IS NULL) AND (gallery_folder.parent_folder_id NOT IN (SELECT column2 FROM (VALUES(76,76))) OR gallery_folder.parent_folder_id IS NULL))) as temp, args: [], error: error executing `SELECT COUNT(*) as count FROM (SELECT DISTINCT galleries.id FROM galleries LEFT JOIN galleries_files ON galleries_files.gallery_id = galleries.id LEFT JOIN files ON galleries_files.file_id = files.id LEFT JOIN folders AS gallery_folder ON galleries.folder_id = gallery_folder.id WHERE ((files.parent_folder_id NOT IN (SELECT column2 FROM (VALUES(76,76))) OR folders.parent_folder_id IS NULL) AND (gallery_folder.parent_folder_id NOT IN (SELECT column2 FROM (VALUES(76,76))) OR gallery_folder.parent_folder_id IS NULL))) as temp` [[]]: no such column: folders.parent_folder_id
```

CynicalAtropos:
## Description

Fixes the gallery parent-folder exclusion query using an incorrect table alias.

The ZIP-gallery condition filtered on `files.parent_folder_id`, but its `IS NULL` branch still referenced `folders.parent_folder_id`. The `folders` table is not present under that name in this query, causing SQLite to return `no such column: folders.parent_folder_id`.

The null check now consistently uses `files.parent_folder_id`.

Adds regression coverage for both exact-folder and recursive parent-folder exclusions.

## Related Issue

Fixes #6993 

## Testing

- `go test -tags="integration sqlite_stat4 sqlite_math_functions" ./pkg/sqlite -count=1`
- `git diff --check`
- `gofmt -w pkg/sqlite/gallery_filter.go pkg/sqlite/gallery_test.go`

Manual testing:

- Applied an exact parent-folder exclusion to galleries.
- Applied the same exclusion recursively with subfolders included.
- Confirmed ZIP-backed galleries in the excluded folders were removed.
- Confirmed galleries outside the excluded hierarchy remained visible.
- Confirmed the GraphQL request completed without a SQLite error.

## Screenshots

N/A

## Checklist

## AI Usage Disclosure

Used codex as assistant to investigate and write the regression tests.  Implemented and manually tested the fix myself.

## Additional Context

This change is intentionally limited to the SQL error reported in #6993.

During testing, two adjacent behaviors were identified but are not changed here:

- A multi-file gallery remains visible when at least one associated file is outside the excluded folder hierarchy.
- A configured library nested inside another configured library is treated as an independent folder root for recursive filtering.

They were the result of intentional design choices elsewhere, but I am not sure if this is an intended behavior for this filtering.

### DIFF
--- a/pkg/sqlite/gallery_filter.go
+++ b/pkg/sqlite/gallery_filter.go
@@ -332,7 +332,7 @@ func (qb *galleryFilterHandler) parentFolderCriterionHandler(folder *models.Hier
 				return
 			}
 
-			f.addWhere(fmt.Sprintf("files.parent_folder_id NOT IN (SELECT column2 FROM (%s)) OR folders.parent_folder_id IS NULL", valuesClause))
+			f.addWhere(fmt.Sprintf("files.parent_folder_id NOT IN (SELECT column2 FROM (%s)) OR files.parent_folder_id IS NULL", valuesClause))
 			f.addWhere(fmt.Sprintf("gallery_folder.parent_folder_id NOT IN (SELECT column2 FROM (%s)) OR gallery_folder.parent_folder_id IS NULL", valuesClause))
 		}
 	}
--- a/pkg/sqlite/gallery_test.go
+++ b/pkg/sqlite/gallery_test.go
@@ -1586,6 +1586,43 @@ func TestGalleryQueryPath(t *testing.T) {
 	}
 }
 
+func TestGalleryQueryParentFolderExcludes(t *testing.T) {
+	tests := []struct {
+		name      string
+		folderIdx int
+		depth     int
+	}{
+		{
+			name:      "exact folder",
+			folderIdx: folderIdxWithGalleryFiles,
+			depth:     0,
+		},
+		{
+			name:      "parent folder recursively",
+			folderIdx: folderIdxForObjectFiles,
+			depth:     -1,
+		},
+	}
+
+	for _, tt := range tests {
+		runWithRollbackTxn(t, tt.name, func(t *testing.T, ctx context.Context) {
+			galleries, count, err := db.Gallery.Query(ctx, &models.GalleryFilterType{
+				ParentFolder: &models.HierarchicalMultiCriterionInput{
+					Value: []string{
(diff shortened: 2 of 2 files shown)


slick-daddy:
Looks good to me. It can be merged safely.

> A configured library nested inside another configured library is treated as an independent folder root for recursive filtering.

Isn't this the intended behaviour? This allows users to add `files` and `files/images` paths seperately to allow flexibility. As long as scanning doesn't duplicate the results, I wouldn't call this a bug.

It should be documented regardless, though.

**When we should consider it as a bug?**
If recursive filtering incorrectly excludes the files/images library by treating it as a subdirectory of the `files` library.

Gykes (approved):
LGTM. Thanks for adding in the testing as well.

PR #7137, opened 2026-08-02
Fix: JSON Based Scrapers Fail With Chrome CDP

### LINKED ISSUE #7136: JSON based scrapers fail with Chrome CDP

### Describe the bug

While trying to troubleshoot why R18.dev scraper was getting 403s I modified the R18.dev [scraper](https://github.com/stashapp/CommunityScrapers/blob/43e272c50c129207f90ad2f45eb5f263d81251ab/scrapers/R18.dev.yml) locally to add a driver block to use CDP. This exposed a bug in the CDP implementation. With CDP the scape was returning "not valid json" even though r18.dev was returning a valid json structure. 

### Steps to reproduce

1. Add `driver:
   useCDP: true` to R18.dev.yml
2. Confirm Chrome CDP Path is configured under Settings → Metadata Providers, pointing at a reachable headless Chrome instance
3. Navigate to a scene named with a valid JAV code and search with R18.dev scraper (eg SIMM-812)
4. Receive red error dialog containing "scraper R18.dev: not valid json"

### Expected behaviour

Should return the scene data, since the underlying [URL](https://r18.dev/videos/vod/movies/detail/-/dvd_id=simm812/json) genuinely returns valid JSON when fetched directly

_No response_

### Stash version

v0.31.1

### Device details

Chrome 150.0.7871.187 (Official Build) (arm64) macOS Version 26.5.2

### Relevant log output

```shell
2026-08-01 21:45:29
Error   
scrapeSingleScene: input: scrapeSingleScene error while name scraping with scraper R18.dev: not valid json
2026-08-01 21:45:29
Info    
loadURL (https://r18.dev/videos/vod/movies/detail/-/dvd_id=SIMM-812/json)
2026-08-01 21:45:26
Debug   
Remote cdp instance found ws://localhost:9222/devtools/browser/15a205c5-ecb4-464e-8a40-b17bbd5f81f0
```

wraithfive:
## Description
When a `useCDP` scraper navigates directly to a URL that Chrome identifies as JSON content, Chrome's own built-in JSON viewer wraps the raw JSON in an HTML pretty-printer document before it ever reaches the page DOM. `urlFromCDP` always extracted content via `chromedp.OuterHTML`, so `scrapeJson`-type scrapers using `useCDP` would receive this HTML wrapper instead of the JSON itself and fail with "not valid json".

## Related Issue
Closes #7136

## Testing
- `go build`, `go vet`, and `go test ./pkg/scraper/...` all pass, including
  new unit tests for the MIME-type decision logic (`url_test.go`)
- Built a full patched Stash image from source and ran it as a fully isolated instance (separate database,
  separate config, pointed at the same headless Chrome CDP sidecar) to verify against the real R18.dev scraper:
- Before the fix: `not valid json`
- After the fix: correctly returns real scene data (title, date, etc.)
- Regression-checked an HTML-based `useCDP` scraper (JavLibrary) and a non-CDP scraper (JavBus) to confirm unaffected behavior
    
## To verify this fix
1. Configure a reachable CDP endpoint under Settings → Metadata Providers (Chrome CDP Path), pointing at any Chrome/chromedp instance with `--remote-debugging-port` exposed.
2. To test the exact scraper I used, add `driver: { useCDP: true }` to R18.dev.yml. Any scraper using `action: scrapeJson` with `driver: { useCDP: true }` should work. 
3. Trigger a scrape against a real JSON endpoint via the scene Edit tab  "Scrape with..." or in the GraphQL playground with 
```
query {
  scrapeSingleScene(
    source: { scraper_id: "R18.dev" }
    input: { query: "SIMM-812" }
  ) {
    title
    date
    details
  }
}
```

Before this fix: fails with `not valid json` regardless of the endpoint actually returning valid JSON. After: succeeds normally.

## Checklist

## AI Usage Disclosure

The root cause was diagnosed and this fix was drafted with Claude Sonnet 5. 

## Additional Context
1. if `Network.getResponseBody` fails (e.g., the response got evicted from Chrome's cache before the fetch), the fix falls back to the original OuterHTML behavior rather than erroring out.

### DIFF
--- a/pkg/scraper/url.go
+++ b/pkg/scraper/url.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/chromedp/cdproto/cdp"
@@ -259,15 +260,21 @@ func urlFromCDP(ctx context.Context, urlCDP string, driverOptions scraperDriverO
 	// responses in an HTML pretty-printer when navigated to directly, so
 	// OuterHTML on such a page returns HTML containing the JSON rather than
 	// the JSON itself.
-	var jsonRequestID network.RequestID
-	var isJSONDocument bool
+	// jsonDoc is written from chromedp's event-processing goroutine (in the
+	// ListenTarget callback below) and read from the action sequence passed
+	// to chromedp.Run, so access must be synchronized.
+	var jsonDoc struct {
+		sync.Mutex
+		requestID network.RequestID
+		isJSON    bool
+	}
 	chromedp.ListenTarget(ctx, func(ev interface{}) {
 		if ev, ok := ev.(*network.EventResponseReceived); ok {
-			if ev.Type == network.ResourceTypeDocument && !isJSONDocument {
-				if isJSONMimeType(ev.Response.MimeType) {
-					jsonRequestID = ev.RequestID
-					isJSONDocument = true
-				}
+			jsonDoc.Lock()
+			defer jsonDoc.Unlock()
+			if ev.Type == network.ResourceTypeDocument && !jsonDoc.isJSON && isJSONMimeType(ev.Response.MimeType) {
+				jsonDoc.requestID = ev.RequestID
+				jsonDoc.isJSON = true
 			}
 		}
 	})
@@ -314,12 +321,16 @@ func urlFromCDP(ctx context.Context, urlCDP string, driverOptions scraperDriverO
 		chromedp.Sleep(sleepDuration),
 		setCDPClicks(driverOptions),
 		chromedp.ActionFunc(func(ctx context.Context) error {
-			if isJSONDocument {
-				body, err := network.GetResponseBody(jsonRequestID).Do(ctx)
+			jsonDoc.Lock()
+			isJSON, requestID := jsonDoc.isJSON, jsonDoc.requestID
+			jsonDoc.Unlock()
+
+			if isJSON {
+				body, err := network.GetResponseBody(requestID).Do(ctx)
 				if err != nil {
 					// fall back to OuterHTML if the response body is no
 					// longer available (e.g. evicted from Chrome's cache)
-					logger.Debugf("[scraper] could not get raw response body for JSON document, falling back to OuterHTML: %v", err)
+					logger.Warnf("[scraper] could not get raw response body for JSON document, falling back to OuterHTML: %v", err)
 					return chromedp.OuterHTML("html", &res, chromedp.ByQuery).Do(ctx)
 				}
 				res = string(body)
--- a/pkg/scraper/url.go
+++ b/pkg/scraper/url.go
@@ -260,21 +260,11 @@ func urlFromCDP(ctx context.Context, urlCDP string, driverOptions scraperDriverO
 	// responses in an HTML pretty-printer when navigated to directly, so
 	// OuterHTML on such a page returns HTML containing the JSON rather than
 	// the JSON itself.
-	// jsonDoc is written from chromedp's event-processing goroutine (in the
-	// ListenTarget callback below) and read from the action sequence passed
-	// to chromedp.Run, so access must be synchronized.
-	var jsonDoc struct {
-		mutex     sync.Mutex
-		requestID network.RequestID
-		isJSON    bool
-	}
+	var jsonDoc jsonDocumentTracker
 	chromedp.ListenTarget(ctx, func(ev interface{}) {
 		if ev, ok := ev.(*network.EventResponseReceived); ok {
-			jsonDoc.mutex.Lock()
-			defer jsonDoc.mutex.Unlock()
-			if ev.Type == network.ResourceTypeDocument && !jsonDoc.isJSON && isJSONMimeType(ev.Response.MimeType) {
-				jsonDoc.requestID = ev.RequestID
-				jsonDoc.isJSON = true
+			if ev.Type == network.ResourceTypeDocument {
+				jsonDoc.markJSON(ev.RequestID, ev.Response.MimeType)
 			}
 		}
 	})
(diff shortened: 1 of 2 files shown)


pkg/scraper/url.go
```diff
@@ -253,6 +253,25 @@ func urlFromCDP(ctx context.Context, urlCDP string, driverOptions scraperDriverO
...
+	// we can pull the raw response body via CDP's Network domain instead of
+	// reading the rendered DOM. Chrome's built-in JSON viewer wraps raw JSON
+	// responses in an HTML pretty-printer when navigated to directly, so
+	// OuterHTML on such a page returns HTML containing the JSON rather than
+	// the JSON itself.
+	var jsonRequestID network.RequestID
+	var isJSONDocument bool
+	chromedp.ListenTarget(ctx, func(ev interface{}) {
+		if ev, ok := ev.(*network.EventResponseReceived); ok {
+			if ev.Type == network.ResourceTypeDocument && !isJSONDocument {
```
Gykes: These are written on chromedps event processing georoute but read from chromedp.Run georoute and they're not synced. I think this could create a race.

pkg/scraper/url.go
```diff
@@ -253,6 +253,25 @@ func urlFromCDP(ctx context.Context, urlCDP string, driverOptions scraperDriverO
...
+				if isJSONMimeType(ev.Response.MimeType) {
```
Gykes: These 2 conditions can be be condensed.

pkg/scraper/url.go
```diff
@@ -294,7 +313,20 @@ func urlFromCDP(ctx context.Context, urlCDP string, driverOptions scraperDriverO
...
 		chromedp.Sleep(sleepDuration),
 		setCDPClicks(driverOptions),
-		chromedp.OuterHTML("html", &res, chromedp.ByQuery),
+		chromedp.ActionFunc(func(ctx context.Context) error {
+			if isJSONDocument {
+				body, err := network.GetResponseBody(jsonRequestID).Do(ctx)
+				if err != nil {
+					// fall back to OuterHTML if the response body is no
+					// longer available (e.g. evicted from Chrome's cache)
+					logger.Debugf("[scraper] could not get raw response body for JSON document, falling back to OuterHTML: %v", err)
```
Gykes: Would making this a warning rather than a debug be better so people can see it in the logs easier?

Gykes (approved):
Quick static review.

Gykes:
Meant to request changes instead of approve, sorry

pkg/scraper/url.go
wraithfive: Good catch

wraithfive:
Pushed a fix for all three in 253f27e. Then thought to do what I probably should have done in the first place and checked the style of other mutex's in the project and extracted it to a named field.

After that I wanted to make sure a test would actually catch this kind of race if it ever came back. The tracked state was living in an anonymous struct scoped inside urlFromCDP itself so I pulled it out into its own small type first (jsonDocumentTracker) so it could be tested without needing chromedp at all. Then added a test that hammers its write and read methods concurrently — confirmed it fails without the mutex and passes with it.

A review by Fable pointed out that Network.responseReceived fires for iframe navigations too, not just the top-level one, so the tracker could in theory latch onto an iframe's JSON response and hijack extraction for an unrelated HTML scraper. Fixed that by only ever tracking the first Document response instead of the first JSON one, and added a test covering it.

pkg/scraper/url.go
```diff
@@ -361,6 +397,72 @@ func getRemoteCDPWSAddress(ctx context.Context, url string) (string, error) {
...
+// request instead), so the first Document response reliably corresponds to
+// the top-level navigation. A click-triggered navigation (via
+// setCDPClicks, which runs after the initial Navigate) is not treated as a
+// new "first" document either, so if a click leads to a JSON page, this
+// still reports the original navigation's result.
+//
+// It's written from chromedp's event-processing goroutine (via
+// markDocument, called from a ListenTarget callback) and read from the
+// action sequence passed to chromedp.Run (via mainDocument), so access is
+// synchronized with a mutex.
```
Gykes: I presume these are AI comments. Could you please condense these down to be a little more readable? This is a general statement for all of them comments that were added.

Gykes (changes requested):

pkg/scraper/url.go
wraithfive: Yeah sorry about that. My day job is really pushing us to use AI everywhere and this was a good chance to get used to using it for coding tasks I'd normally do manually that don't risk killing someone (Healthcare software). Noted for future to make sure I give it instructions to be concise. I kind of liked the verbosity while actually reviewing the code it had written but agree it's too much to leave in.

Gykes: No worries, I have the same issue at my day job :D

pkg/scraper/url.go
```diff
@@ -361,6 +387,50 @@ func getRemoteCDPWSAddress(ctx context.Context, url string) (string, error) {
...
+	recorded  bool
+}
+
+// markDocument records requestID as the main document if no document has
+// been recorded yet. Subsequent calls after the first are no-ops.
+func (t *jsonDocumentTracker) markDocument(requestID network.RequestID, mimeType string) {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+
+	if t.recorded {
```
Gykes, line 417: Does this need to be the first document or the last main-frame one? 

If the URL hits a JS/CF challenge or a redirect page before landing on the JSON the first document is an HTML. That would take us down the "not valid JSON" path.

Gykes (changes requested):
Sorry, hopefully this is the last review.

pkg/scraper/url.go
wraithfive: Good Point. I didn't think about pages that redirect or sit behind cloudflare browser challenges before landing on the actual JSON. chromedp will get through those on its own (that's what the sleepDuration setting is there for, giving it a second to settle). It won't handle a captcha challenge but I don't want to get into trying to auto solve those in this. So "first document" was the wrong thing to grab, since by the time it settles we're way past that first page.

I'm thinking instead we keep track of the actual top-level page via page.EventFrameNavigated and only pay attention to Document responses on that frame. The top level should be the one where Frame.ParentID == "". Instead of keeping the first one we see, keep overwriting it so we end up with whatever loaded last. That way an iframe's response still can't sneak in (wrong frame, so it's filtered out no matter when it shows up), but a redirect or challenge page doesn't get stuck as the "final" answer either. Whatever the browser actually ends up sitting on wins.

Does that logic sound right before I make the change?

PR #7143, opened 2026-08-04
Restore scroll position when navigating back to a list

### LINKED ISSUE #7142: Scroll position is lost when navigating back to a list

### Describe the bug

Going back to a list view lands at the top of the list instead of the position you left. Scroll down the Scenes list, open a scene, press the browser back button, and the list comes back scrolled to the top. Finding where you were means scrolling down again, which gets worse the further into a long list you were.

It happens on any list that is tall enough to scroll, and it happens whether or not you changed anything on the scene page.

The cause is that list content loads asynchronously. At the moment the history entry is popped the query hasn't resolved, so the document is one viewport tall and the browser clamps its restored position to 0. When the cards render a moment later, nothing scrolls back. Instrumented in Chromium:

```
popstate  scrollY=0  documentElement.scrollHeight=900   <- content not rendered yet
+2s       scrollY=0  documentElement.scrollHeight=1978  <- content rendered, no scroll back
```

React Router 5 does not do scroll restoration of its own, so nothing in the app compensates for this.

### Steps to reproduce

1. Go to Scenes, with enough scenes for the page to scroll.
2. Scroll down a few rows.
3. Click a scene to open it.
4. Press the browser back button.
5. The list is back, scrolled to the top.

### Expected behaviour

The list comes back at the position it was left at, the way it does on a server-rendered page.

Two related cases that behave the same way:

- If the list was on page 2 or later, coming back also loses the scroll position, and the filter re-parse on the way back triggers the list's own scroll-to-top.
- Reloading the scene page and then going back has the same result.

_No response_

### Stash version

Development build of `develop` at 2926d02c.

### Device details

Chromium 143 WSL2

### Relevant log output

```shell

```

tassilogravura:
## Description
React Router v5 has no scroll restoration, so the app relied on the browser's native restoration. That doesn't work for the list views: at the point the history entry is popped, the list content hasn't been fetched and rendered yet, so the document is one viewport tall and the position is clamped to the top. Nothing scrolls back once the content arrives, leaving the user at the top of the list.

Record the scroll position for each history entry and re-apply it on POP until the page is able to scroll there, holding it briefly while the list settles. Positions are kept in session storage so they survive a page reload.

Coming back to a list that was showing a page other than the first re-parses the filter from the query string, which looks like a page change to useScrollToTopOnPageChange and scrolled over the restored position, so that hook now defers to an in-progress restoration.

## Related Issue
Closes https://github.com/stashapp/stash/issues/7142

## Testing
manual testing in my own browser

## Screenshots
-

## Checklist

## AI Usage Disclosure

- Claude Opus 5

### DIFF
--- a/ui/v2.5/src/App.tsx
+++ b/ui/v2.5/src/App.tsx
@@ -29,6 +29,7 @@ import { MainNavbar } from "./components/MainNavbar";
 import { PageNotFound } from "./components/PageNotFound";
 import * as GQL from "./core/generated-graphql";
 import { makeTitleProps } from "./hooks/title";
+import { useScrollRestoration } from "./hooks/scrollRestoration";
 import { LoadingIndicator } from "./components/Shared/LoadingIndicator";
 
 import {
@@ -205,6 +206,8 @@ export const App: React.FC = () => {
   const location = useLocation();
   const history = useHistory();
   const setupMatch = useRouteMatch(["/setup", "/migrate", "/welcome"]);
+
+  useScrollRestoration();
 
   // dispatch event when location changes
   useEffect(() => {
--- a/ui/v2.5/src/components/List/PagedList.tsx
+++ b/ui/v2.5/src/components/List/PagedList.tsx
@@ -5,6 +5,7 @@ import { Pagination, PaginationIndex } from "./Pagination";
 import { LoadingIndicator } from "../Shared/LoadingIndicator";
 import { ErrorMessage } from "../Shared/ErrorMessage";
 import { FormattedMessage } from "react-intl";
+import { useContentLoading } from "src/hooks/scrollRestoration";
 
 export const LoadedContent: React.FC<
   PropsWithChildren<{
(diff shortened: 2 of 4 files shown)


hyper440:
One concern: `scrollToWhenReady` relies on fixed 3s/750ms timers, which could fail on slow mobile queries. Could restoration run once the list finishes loading, with a `ResizeObserver` if the layout still needs to settle?

tassilogravura:
> One concern: `scrollToWhenReady` relies on fixed 3s/750ms timers, which could fail on slow mobile queries. Could restoration run once the list finishes loading, with a `ResizeObserver` if the layout still needs to settle?

Good catch, the 3s is real. With an 8s query it gives up and you land at the
top, which is the exact bug this fixes.

Deadline is conditional now: lists report while loading (one call in PagedList,
which all of them go through) and the clock only runs when nothing is loading.
3s stays as a fallback, 30s hard cap.

Took the ResizeObserver suggestion too, so re-applying is driven by the document
growing instead of a rAF loop.

Kept the settle hold but cut it to 250ms, measured from when loading finishes.
It isn't waiting for layout, it holds the position against
useScrollToTopOnPageChange, which scrolls without changing geometry, so no
resize fires for it.

PR #7144, opened 2026-08-06
feat: support multiple scenes in a single file via scene-file range (#3530)

benkwok1983-cmd:
## Summary
Implements part of #3530 (Support multiple scenes in a single file): adds a nullable `start_time`/`end_time` range to the scene-file relationship (`scenes_files`), so multiple scenes can reference the same physical file with different time ranges without re-encoding.

## Design
- Both `start_time`/`end_time` NULL: full file (existing behaviour, backward compatible)
- Only `start_time`: play from that point to end of file
- Both set: bounded range
- Range is valid only for a scene's primary file
- `duration` remains derived from the source file (not stored separately)

## Changes
- **Migration 86**: `ALTER TABLE scenes_files ADD COLUMN start_time REAL NULL; ADD COLUMN end_time REAL NULL;`
- **pkg/models**: `Scene` gets transient `StartTime`/`EndTime` (`*float64`, json `start_time`/`end_time`)
- **pkg/sqlite**: `selectDataset` selects the range columns with the primary file join; `sceneQueryRow.resolve()` populates them; `relatedFilesTable.setRange()` writes the range; `SceneStore.Create` sets the range on the primary file when provided
- **GraphQL**: `Scene.start_time`/`Scene.end_time` exposed; `SceneCreateInput`/`SceneUpdateInput` accept `start_time`/`end_time`
- **API**: `SceneCreate` passes the range through to the new scene

## Scope note
This PR covers the data model + API layer (migration, persistence, GraphQL exposure). Playback range handling, UI fields, and the `sceneCreateFromFileRange` operation are follow-ups — happy to add them here if preferred.

## Testing
- `go generate` + backend integration tests run via CI (generate job runs gqlgen)
- Manual verification of the migration against a test DB pending maintainer direction

Refs: #3530

### DIFF
--- a/SCENE_RANGE_PLAN.md
+++ b/SCENE_RANGE_PLAN.md
@@ -1,0 +1,97 @@
+# stashapp #3530 — 单文件多场景（Scene Range）实现方案
+
+> 目标：$450 赏金（OpenCollective，支持 PayPal 提现）
+> 认领：2026-08-06 10:00 UTC 已在 issue #3530 评论认领
+> 仓库：stashapp/stash (Go 后端 + React 前端, 12.8k star)
+> Fork：benkwok1983-cmd/stash → 本地 E:\Hermes\bounties\stash
+
+## 需求
+一个视频文件可以包含多个场景（按时间范围切分），无需重新编码。
+维护者倾向"类似 Markers 的机制"（issue #260/#779 相关）。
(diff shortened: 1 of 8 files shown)


DogmaDragon:
https://github.com/stashapp/stash/blob/develop/docs/CONTRIBUTING.md

PR #7151, opened 2026-08-11
Fix: Make the Merge Modals Patchable + onClose Modal Fix

### LINKED ISSUE #7150: Expose the Scene and Performer merge modals for plugins

I am writing a plugin that supplements the existing duplicate tool (in the vein of [DupeFinder](https://github.com/zzzinsCode/stashapp-plugins/tree/main/plugins/DupeFinder) and the original [stash-plugin-duplicate-finder](https://github.com/WithoutPants/stash-plugin-duplicate-finder)) but I'd like to use the native merge modals instead of having to write my own clumsy copies

The existing duplicate finder is based on PHASH similarity but many users are experiencing duplicates in their collections for other reasons, such as mismatched scenes from stash-boxes. These are not "true duplicates" in that they don't always refer to the exact same scene, they're either redistributions, re-releases or straight up mismatched fingerprints. The problem remains though, and they can be hard to find so I'm writing this plugin to make it easier 🙂 

This can be be done in a painstakingly manual way by running SQL queries through the GraphQL playground

### Additional context

_No response_

Maista6969:
## Description

Follows the same style as earlier PRs that expose more components through the PluginApi extension point: the PR looks bigger than it is because of the whitespace changes so it's helpful for reviewers to turn that off (git diff -w in your terminal or [directly on GitHub](https://github.com/stashapp/stash/compare/develop...Maista6969:stash:plugin-api-merge-modals?diff=split&w=1))

After writing the issue I realized that TagMergeModal was also relevant now that tags can have StashIDs so I included it as well.

I cheekily snuck in a small bug fix commit that can be removed if needed: it's a small bug in that the Scene and Performer merge modals call their `onClose` callback twice on a successful merge, first with the scene ID (which is correct) and then with no arguments, which means undefined, so the receiver gets mixed messages. I traced through the calling code and native Stash could never trigger this (which is why it's survived for four years) but a plugin like mine could end up relying on the return value in a way that would run into this.

## Related Issue
This is a solution to my own issue #7150

## Testing
I did not add anything to the test suite because this does not introduce any new functionality. The bug fix is minor in that the Stash components that use the onClose

I am running this branch on my development instance of Stash and the modals are exposed and work as expected. They continue working in their original contexts too, of course.

## Checklist

## AI Usage Disclosure

### DIFF
--- a/ui/v2.5/src/components/Performers/PerformerMergeDialog.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerMergeDialog.tsx
@@ -799,7 +799,6 @@ export const PerformerMergeModal: React.FC<IPerformerMergeModalProps> =
           Toast.success(intl.formatMessage({ id: "toast.merged_performers" }));
           onClose(destPerformer[0].id);
         }
-        onClose();
       } catch (e) {
         Toast.error(e);
       } finally {
--- a/ui/v2.5/src/components/Performers/PerformerMergeDialog.tsx
+++ b/ui/v2.5/src/components/Performers/PerformerMergeDialog.tsx
@@ -46,6 +46,7 @@ import {
 import { PerformerSelect } from "./PerformerSelect";
 import { uniq } from "lodash-es";
 import { StashIDsField } from "../Shared/StashID";
+import { PatchComponent } from "src/patch";
 
 type MergeOptions = {
   values: GQL.PerformerUpdateInput;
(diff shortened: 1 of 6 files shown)


DogmaDragon (approved):
Documentation check passed.

Gykes (approved):
Overall looks good. I didn't actually test the patches but seems sound.

THe only other thing I can think is that by moving the `onClose` the modal no longer closes if the mutation resolves without throwing. I think that's an extreme edge case and the fix is correct overall.

PR #7159, opened 2026-08-14
Support explicit non-root Docker users

### LINKED ISSUE #684: [Feature] non-privileged user in Docker build

ImAurum11:
## Description

I went through the previous attempts and the rootless guide before working on this. I ended up keeping the current image behavior unchanged and relying on Docker's own `user: UID:GID` support instead of adding another PUID/PGID entrypoint.

The main thing in the way was Stash trying to look up the current user. Numeric container users don't always exist in `/etc/passwd`, so Stash now uses `HOME` first and falls back to the previous lookup when needed. The compose example shows the matching home and config settings without changing anything for existing installations.

CUDA actually needed a small exception because its optional NVIDIA patch writes to system directories. It still works exactly as before when running as root, while a non-root container skips the patch and starts Stash normally.

## Related Issue

Closes #684.

## Testing

I tested this with UID/GID `12345:23456` and no matching user entry, and success! Stash started in the official image, responded over HTTP, and wrote files with the expected ownership.

The Go tests, configuration tests, Compose validation, shell checks and full backend build also pass. Should be good to go.

## Screenshots

Not applicable.

## Checklist

## AI Usage Disclosure

I used Codex while researching the previous attempts, tidying/tying up the implementation and running some tests. I reviewed all of its changes and take full responsibility for them.

On a side note, sorry again for the earlier PR missing the template, sometimes I skip through reading unintentionally. Appreciate your patience.

### DIFF
--- a/docker/build/x86_64/Dockerfile-CUDA
+++ b/docker/build/x86_64/Dockerfile-CUDA
@@ -54,13 +54,13 @@ COPY --from=backend --chmod=555 /stash/stash /usr/bin/
 RUN mkdir -p /usr/local/bin /patched-lib
 ADD --chmod=555 https://raw.githubusercontent.com/keylase/nvidia-patch/master/patch.sh /usr/local/bin/patch.sh
 ADD --chmod=555 https://raw.githubusercontent.com/keylase/nvidia-patch/master/docker-entrypoint.sh /usr/local/bin/nvidia-patch-entrypoint.sh
-COPY --chmod=555 ./docker/build/x86_64/cuda-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+COPY --chmod=555 ./docker/build/x86_64/cuda-entrypoint.sh /usr/local/bin/cuda-entrypoint.sh
 
 ENV LANG=C.UTF-8
 ENV NVIDIA_VISIBLE_DEVICES=all
 ENV NVIDIA_DRIVER_CAPABILITIES=video,utility
 ENV STASH_CONFIG_FILE=/root/.stash/config.yml
 EXPOSE 9999
-ENTRYPOINT ["docker-entrypoint.sh", "stash"]
+ENTRYPOINT ["/usr/local/bin/cuda-entrypoint.sh", "stash"]
 
 # vim: ft=dockerfile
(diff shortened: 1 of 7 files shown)


docker/build/x86_64/cuda-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -eu
+
+if [ "$(id -u)" -eq 0 ]; then
+  exec /usr/local/bin/nvidia-patch-entrypoint.sh "$@"
+fi
+
+exec "$@"
```
aserv92, line 8: The current implementation will re-run stash (one more time) when the user ID is 0 (root) and stash exits with exit code 0 on the first run.

```suggestion
if [ "$(id -u)" -eq 0 ]; then
  exec /usr/local/bin/nvidia-patch-entrypoint.sh "$@"
else
  exec "$@"
fi
```
or
```suggestion
if [ "$(id -u)" -eq 0 ]; then
  exec /usr/local/bin/nvidia-patch-entrypoint.sh "$@"
  exit $?
fi

exec "$@"
```

docker/build/x86_64/Dockerfile-CUDA
```diff
@@ -53,7 +53,8 @@ COPY --from=backend --chmod=555 /stash/stash /usr/bin/
 # NVENC Patch
 RUN mkdir -p /usr/local/bin /patched-lib
 ADD --chmod=555 https://raw.githubusercontent.com/keylase/nvidia-patch/master/patch.sh /usr/local/bin/patch.sh
-ADD --chmod=555 https://raw.githubusercontent.com/keylase/nvidia-patch/master/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+ADD --chmod=555 https://raw.githubusercontent.com/keylase/nvidia-patch/master/docker-entrypoint.sh /usr/local/bin/nvidia-patch-entrypoint.sh
+COPY --chmod=555 ./docker/build/x86_64/cuda-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```
aserv92: I recommend using a better name for the entry point as both entry points are docker entry points.

aserv92 (commented):
I also have concerns about running stash as a user that does not actually exist from the containers perspective.

ImAurum11:
These are interesting suggestions ngl, appreciate you taking the time. Sorry I didn't see them earlier, GitHub didn't really notify me in my inbox or email haha.

I took the naming point and changed the wrapper to `cuda-entrypoint.sh`, which makes its role clearer next to `nvidia-patch-entrypoint.sh`. I also clarified in the docs that the numeric UID doesn't need an `/etc/passwd` entry as long as `HOME` points to a writable directory. That's the exact setup I tested with UID/GID `12345:23456`.

I took another look at the `exec` part, but I left it as-is because it can't run Stash twice. The first `exec` replaces the wrapper with the NVIDIA script, and that script ends with `exec "$@"`. The `if/else` version would behave the same, while the other suggestion has an unreachable `exit` after `exec`.

Thanks again.

PR #7166, opened 2026-08-15
Allow plugin settings to add connect-src CSP sources

### LINKED ISSUE #7165: Plugin settings should be able to add connect-src CSP sources

Add a mechanism for plugin developers to contribute additional connect-src entries to Stash’s page Content Security Policy through plugin settings, rather than only through the static ui.csp.connect-src value in the plugin’s yml.

Currently, a plugin’s CSP contributions are hard-coded in its shipped yml file. This forces plugins that talk to a user-configurable backend to either ship a broad wildcard (e.g. http://*:7860), which is poor security policy, or require users to hand-edit the plugin’s yml on every update. A cleaner approach is to let a plugin define a settings entry whose value is treated as a validated connect-src source when building the page CSP header. Concretely: settings whose key starts with csp_ and whose value is a valid concrete http/https URL get appended to the page’s connect-src directive for that plugin.

Validation would be strict: values must be concrete http/https URLs with a host, and anything else (wildcards like http://*:7860, non-URLs, disallowed schemes, empty strings, values containing CSP meta-characters) is ignored, so invalid settings can never corrupt the header.

- Users of plugins that proxy to a configurable backend (face-recognition servers, tag scrapers, etc.) can configure the backend endpoint in Stash’s plugin settings UI, and the correct `connect-src` entry is applied automatically on the next page load, no hand-editing of plugin files.
- Because plugin settings persist in Stash’s own config (not the plugin’s shipped files), they survive plugin updates. A plugin can ship a narrow default and let users add exactly the host they need.
- It replaces the anti-pattern of shipping broad wildcards, giving tighter CSPs without a per-plugin maintenance burden.

Yes, partially: a plugin can ship `ui.csp.connect-src` entries in its static `yml`. However, that value is fixed at install time and cannot be changed from the plugin settings UI, and any change is lost on plugin update. There is currently no runtime/configurable path.

### Additional context

I have a reference implementation on branch `csp-plugin-settings` (PR to follow) that adds a pure helper in `internal/api/server.go` reading `GetPluginConfiguration(pluginID)` and appending validated `csp_*` settings to the `connect-src` header, with strict URL validation (rejects wildcards, disallowed schemes, whitespace/`;` breakout attempts, degenerate hosts, and non-string values), plus unit tests.

cc1234475:
## Description
Adds a mechanism for plugins to contribute validated `connect-src` entries to Stash's page Content Security Policy via plugin settings, rather than only through the static `ui.csp.connect-src` in the plugin `yml`.

Any plugin setting whose key starts with `csp_` and whose value is a valid concrete `http`/`https` URL is appended to the page's `connect-src` directive (per plugin, only for enabled plugins). This lets plugins that talk to a user-configurable backend ship a narrow default and have users set their exact backend endpoint in the plugin settings UI, applied automatically on the next page load — no hand-editing of plugin files, and the value survives plugin updates because it's stored in Stash's config rather than the shipped files.

Currently the static `ui.csp.connect-src` forces plugins with user-configurable backends to either ship a broad wildcard (e.g. `http://*:7860`) — poor security policy — or require users to edit the plugin `yml`, which is overwritten on every update.

Implementation details:
- `internal/api/server.go`: new pure helper `cspConnectSrcFromSettings(settings)` returns validated connect-src URLs plus keys skipped as invalid, and `isValidConnectSrcURL(s)` validates strictly via `net/url`. In `setPageSecurityHeaders`, inside the existing per-plugin loop, `GetPluginConfiguration(plugin.ID)` is read and validated `csp_*` values are appended to `connect-src`. Invalid values are skipped with a debug log (key + plugin ID).
- `internal/api/server_test.go`: table-driven tests covering valid URLs, disallowed schemes, non-URLs, non-string values, wildcard hosts, CSP directive-breakout attempts (whitespace/`;`/quotes), degenerate hosts, userinfo, empty strings, and mixed valid/invalid.

Validation intentionally rejects: anything other than concrete `http`/`https` URLs with a host, wildcards (`http://*:7860`), non-string values, empty strings, values containing CSP meta-characters, degenerate hosts, and URLs with userinfo — so a misconfigured or hostile setting can never corrupt or disable the CSP header.

`pkg/plugin` is unchanged; only `connect-src` is affected (not `script-src`/`style-src`). No reload or restart is needed — the header is rebuilt per request.

## Related Issue
Closes #7165

## Testing
Unit tests: `go test ./internal/api/ -run TestCspConnectSrcFromSettings` — 14/14 pass. Full `go build ./...` succeeds.

Manual verification (local instance, plugin with `settings: csp_apiEndpoint`):
1. Built the branch and ran a local Stash instance with a test plugin defining a `csp_` setting.
2. Confirmed the served page `Content-Security-Policy` `connect-src` header initially excludes the setting's URL.
3. Set the setting to `https://api.example.com` via the `configurePlugin` GraphQL mutation — the URL appeared in the served `connect-src` header on the next request, without reload or restart.
4. Set the setting to a CSP-breakout value (`https://evil.com/; script-src 'none'`) — it was rejected, and the header remained valid and unchanged.
5. Set the setting to a wildcard (`http://*:7860`) — rejected, not added to the header.
6. Confirmed the value persists in Stash's `config.yml` under the plugin ID.

## Screenshots
No UI changes; this is a server-side CSP header change, so there are no screenshots.

## Checklist
- [x] I have read and understood the [[Contributing](https://github.com/stashapp/stash/blob/develop/docs/CONTRIBUTING.md)](https://github.com/stashapp/stash/blob/develop/docs/CONTRIBUTING.md) document.
- [x] I have read and understood the [[AI Usage Policy](https://github.com/stashapp/stash/blob/develop/docs/AI_POLICY.md)](https://github.com/stashapp/stash/blob/develop/docs/AI_POLICY.md) document.

## AI Usage Disclosure

I used an AI coding assistant (OpenCode with an Deepseek v4 Flash) to help generate and refine the implementation and the accompanying tests. I have reviewed and understood every line and design decision, performed the manual testing described above, and take full responsibility for the change and its AGPL licensing.

## Additional Context
This addresses a design gap where plugins with user-configurable backends cannot express per-user connect-src entries, forcing either broad wildcards or manual `yml` edits that are lost on plugin update. A follow-up would be for affected plugins (e.g. a face-recognition userscript) to adopt this mechanism and drop their wildcard entries.

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -574,7 +574,7 @@ func cspConnectSrcFromSettings(settings map[string]interface{}) (valid []string,
 }
 
 func isValidConnectSrcURL(s string) bool {
-	if strings.ContainsAny(s, " \t\r\n;\"'") {
+	if strings.ContainsAny(s, " ,\t\r\n;\"'") {
 		return false
 	}
 	u, err := url.Parse(s)
@@ -642,11 +642,11 @@ func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request, plugins []*p
 
 		connectSrcSlice = append(connectSrcSlice, ui.CSP.ConnectSrc...)
 
-		if settings := c.GetPluginConfiguration(plugin.ID); settings != nil {
+		if settings := c.GetPluginConfiguration(plugin.ID); settings != nil && ui.CSPSettings {
 			valid, skippedKeys := cspConnectSrcFromSettings(settings)
 			connectSrcSlice = append(connectSrcSlice, valid...)
 			for _, key := range skippedKeys {
-				logger.Debugf("skipping invalid csp_ setting %q for plugin %q", key, plugin.ID)
+				logger.Warnf("skipping invalid csp_ setting %q for plugin %q", key, plugin.ID)
 			}
 		}
 
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -574,11 +574,14 @@ func cspConnectSrcFromSettings(settings map[string]interface{}) (valid []string,
 }
 
 func isValidConnectSrcURL(s string) bool {
+	if strings.ContainsAny(s, " \t\r\n;\"'") {
+		return false
+	}
 	u, err := url.Parse(s)
 	if err != nil {
 		return false
 	}
-	return (u.Scheme == "http" || u.Scheme == "https") && u.Host != "" && !strings.Contains(u.Host, "*")
+	return (u.Scheme == "http" || u.Scheme == "https") && u.Host != "" && u.Hostname() != "" && u.User == nil && !strings.Contains(u.Host, "*")
 }
 
 func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request, plugins []*plugin.Plugin) {
@@ -639,7 +642,7 @@ func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request, plugins []*p
 
 		connectSrcSlice = append(connectSrcSlice, ui.CSP.ConnectSrc...)
 
-		if settings := config.GetInstance().GetPluginConfiguration(plugin.ID); settings != nil {
+		if settings := c.GetPluginConfiguration(plugin.ID); settings != nil {
 			valid, skippedKeys := cspConnectSrcFromSettings(settings)
 			connectSrcSlice = append(connectSrcSlice, valid...)
 			for _, key := range skippedKeys {
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/fs"
 	"net/http"
+	"net/url"
 	"os"
 	"path"
 	"path/filepath"
@@ -552,6 +553,34 @@ func isURL(s string) bool {
 	return strings.HasPrefix(s, "http://") || strings.HasPrefix(s, "https://")
 }
 
+const cspSettingPrefix = "csp_"
+
+// cspConnectSrcFromSettings returns validated http(s) connect-src URLs from
+// the plugin's settings, plus the keys that were skipped as invalid.
+// Settings keys beginning with "csp_" are treated as connect-src sources.
+func cspConnectSrcFromSettings(settings map[string]interface{}) (valid []string, skippedKeys []string) {
+	for k, v := range settings {
+		if !strings.HasPrefix(k, cspSettingPrefix) {
+			continue
+		}
+		s, ok := v.(string)
+		if !ok || !isValidConnectSrcURL(s) {
+			skippedKeys = append(skippedKeys, k)
+			continue
+		}
+		valid = append(valid, s)
+	}
+	return valid, skippedKeys
+}
+
+func isValidConnectSrcURL(s string) bool {
+	u, err := url.Parse(s)
+	if err != nil {
+		return false
+	}
+	return (u.Scheme == "http" || u.Scheme == "https") && u.Host != "" && !strings.Contains(u.Host, "*")
+}
+
 func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request, plugins []*plugin.Plugin) {
 	c := config.GetInstance()
 
@@ -609,6 +638,15 @@ func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request, plugins []*p
 		}
 
 		connectSrcSlice = append(connectSrcSlice, ui.CSP.ConnectSrc...)
+
+		if settings := config.GetInstance().GetPluginConfiguration(plugin.ID); settings != nil {
+			valid, skippedKeys := cspConnectSrcFromSettings(settings)
+			connectSrcSlice = append(connectSrcSlice, valid...)
+			for _, key := range skippedKeys {
+				logger.Debugf("skipping invalid csp_ setting %q for plugin %q", key, plugin.ID)
+			}
+		}
+
 		scriptSrcSlice = append(scriptSrcSlice, ui.CSP.ScriptSrc...)
 		styleSrcSlice = append(styleSrcSlice, ui.CSP.StyleSrc...)
 	}
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -14,8 +14,10 @@ import (
 	"path"
 	"path/filepath"
 	"runtime/debug"
+	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	gqlHandler "github.com/99designs/gqlgen/graphql/handler"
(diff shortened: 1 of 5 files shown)


internal/api/server.go
```diff
@@ -552,6 +553,37 @@ func isURL(s string) bool {
...
+			skippedKeys = append(skippedKeys, k)
+			continue
+		}
+		valid = append(valid, s)
+	}
+	return valid, skippedKeys
+}
+
+func isValidConnectSrcURL(s string) bool {
+	if strings.ContainsAny(s, " \t\r\n;\"'") {
```
Gykes: Are commas filtered here? Could that potentially break headers if someone was misconfigured?

internal/api/server.go
```diff
@@ -609,6 +641,15 @@ func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request, plugins []*p
 		}
 
 		connectSrcSlice = append(connectSrcSlice, ui.CSP.ConnectSrc...)
+
+		if settings := c.GetPluginConfiguration(plugin.ID); settings != nil {
+			valid, skippedKeys := cspConnectSrcFromSettings(settings)
```
Gykes: So, I think the prefix you went with will collide with the generic plugin-settings namespace. I think this would inject or degub on every page load and there's no opt in. Not sure the best way to handle this tbh.

internal/api/server.go
```diff
@@ -609,6 +641,15 @@ func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request, plugins []*p
...
+			connectSrcSlice = append(connectSrcSlice, valid...)
+			for _, key := range skippedKeys {
+				logger.Debugf("skipping invalid csp_ setting %q for plugin %q", key, plugin.ID)
```
Gykes: Would this be better as a warn?

Gykes (changes requested):
I will preface this by saying I don't have a massive amount of knowledge in this area. Feel free to push back and defend any decisions you made.

internal/api/server.go
cc1234475: Good catch. Added `,` to the blocked character list in `isValidConnectSrcURL`. Commas are valid in URLs but could cause confusion if someone uses them as CSP list separators. No harm in rejecting them defensively.

Resolved in the latest commit.

internal/api/server.go
cc1234475: Agreed, this was a real concern. Added an opt-in mechanism: plugins must now set `csp-settings: true` in their `ui` section to enable the `csp_` setting prefix. Without it, `csp_`-prefixed settings are ignored for CSP purposes.

This prevents accidental namespace collisions. The flag is plumbed through `UIConfig` (yaml: `csp-settings`) and `PluginUI` (json: `csp_settings`), and the `cspConnectSrcFromSettings` call is gated on `ui.CSPSettings`.

Also updated Plugins.md to document the opt-in requirement.

internal/api/server.go
cc1234475: Done. Changed `Debugf` to `Warnf`. A skipped CSP setting means a plugin author misconfigured something that silently breaks functionality — worth surfacing at warn level.

internal/api/server.go
```diff
@@ -609,6 +641,15 @@ func setPageSecurityHeaders(w http.ResponseWriter, r *http.Request, plugins []*p
 		}
 
 		connectSrcSlice = append(connectSrcSlice, ui.CSP.ConnectSrc...)
+
+		if settings := c.GetPluginConfiguration(plugin.ID); settings != nil && ui.CSPSettings {
```
Gykes: This seems to run unconditionally so I think `GetPluginConfiguration` is called for every enabled plugin on every page load, regardless of the opt in. I think the opt-in flag should have a gate of some kind.

Here's a quick rough example not an actual fix: 

```
if ui.CSPSettings {
       if settings := c.GetPluginConfiguration(plugin.ID); settings != nil {
          valid, skippedKeys := cspConnectSrcFromSettings(settings)
          ...
      }
  }
```

internal/api/server_test.go
```diff
@@ -0,0 +1,125 @@
...
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			valid, skippedKeys := cspConnectSrcFromSettings(tt.settings)
+			assert.ElementsMatch(t, tt.valid, valid)
+			assert.ElementsMatch(t, tt.skippedKeys, skippedKeys)
+		})
+	}
+}
+
+func TestSetPageSecurityHeaders_CSPSettingsOptIn(t *testing.T) {
```
Gykes: Unless im misunderstanding this test then it's not really testing anything. How it looks is you are setting true on the flag and then just calling that to validate that it's true. I think it would be better to call `setPageSecurityHeaders` and check if the `csp_` url shows up in the header.

Gykes (changes requested):
Another static review

internal/api/server.go
cc1234475: Good catch, you were right. The old `settings != nil && ui.CSPSettings` still called `GetPluginConfiguration` (and took its RLock) for every enabled plugin on every page request, and only then discarded the result. Gated it exactly as you wrote it in `70ded0a0`:

```go
// only read plugin settings if the plugin opted in to the csp_ prefix
if ui.CSPSettings {
    if settings := c.GetPluginConfiguration(plugin.ID); settings != nil {
        ...
    }
}
```

While in there I fixed two related things that your comment made obvious:

- The `Warnf` you asked for earlier fired on **every page request** for a misconfigured setting, which is log spam. It is now deduplicated per plugin ID + setting key, so it warns once and re-warns only if the user changes the setting to a different invalid value. The log records the key only, never the value.
- `settings` is a map, so the emitted `connect-src` ordering varied between requests. The validated URLs are now sorted, making the header stable.

internal/api/server_test.go
cc1234475: You understood it correctly, and you were right: that test asserted on the struct literal it had just built, not on any behaviour. Replaced in `70ded0a0` with one that calls `setPageSecurityHeaders` through `httptest` and asserts on the emitted `Content-Security-Policy` header:

```go
func connectSrc(t *testing.T, plugins []*plugin.Plugin, pluginConfig map[string]interface{}) string {
    c := config.InitializeEmpty()
    for _, p := range plugins {
        c.SetPluginConfiguration(p.ID, pluginConfig)
    }

    w := httptest.NewRecorder()
    setPageSecurityHeaders(w, httptest.NewRequest(http.MethodGet, "/", nil), plugins)

    return w.Header().Get("Content-Security-Policy")
}
```

Three cases, all against the real header:

- **opted in**: the valid `csp_` URL is present, the wildcard value is not, and the non-`csp_` setting value is not.
- **not opted in**: the valid `csp_` URL is absent.
- **disabled plugin**: the valid `csp_` URL is absent.

`go test ./...` passes (884 tests, 66 packages).

cc1234475:
Pushed `70ded0a0`, which addresses both outstanding review comments (the `GetPluginConfiguration` gate and the tautological opt-in test), plus two issues that fell out of them: the warning no longer repeats on every page request, and the emitted `connect-src` is now deterministically ordered.

One thing worth stating explicitly rather than leaving for a reviewer to find, since it is a deliberate design choice:

**`configurePlugin` does not validate keys against the plugin's declared settings.** It writes whatever map it is given. So with the `csp_` prefix, a `csp_*` key can be added at runtime that the plugin `yml` never declared. That is intentional and is most of the point of #7165: a user can point a plugin at their own backend without editing plugin files that an update would overwrite. The settings UI itself only renders declared settings, so this only applies to the GraphQL mutation path.

The trade-off: anything able to call `configurePlugin` can widen `connect-src` for the whole page, for any enabled plugin that has opted in. Validation constrains it to concrete `http`/`https` origins, so the header cannot be corrupted or a directive broken out of, but it is still an added destination the plugin author did not sanction.

The stricter alternative is an explicit allowlist in the `yml`, e.g. `ui.csp.connect-src-settings: [apiEndpoint]`, so only named settings feed the policy. It costs nothing for the settings-UI flow (the author writes one static line either way, the user still picks the value), and it closes the runtime-key path. It does lose the ability to add a source the plugin author never anticipated.

Happy to switch to the allowlist if you prefer the tighter contract, it is a small change. Flagging the choice rather than making it silently.

PR #7169, opened 2026-08-17
Fix: Fetch Stash-Box Studio Parents in the Scene Query

InfiniteStash:
## Description

Adds studio parent query to studio fragment. This avoids the large number of round-trips to fetch single studios with delays in between. Not sure why it was ever implemented like that.

## Testing

Tested locally and works fine. On the whole it should be a speedup with no other consequences.

## Checklist

## AI Usage Disclosure

Drafted with claude.

### DIFF
--- a/graphql/stash-box/query.graphql
+++ b/graphql/stash-box/query.graphql
@@ -20,6 +20,13 @@ fragment StudioFragment on Studio {
   parent {
     name
     id
+    aliases
+    urls {
+      ...URLFragment
+    }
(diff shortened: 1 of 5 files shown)


Gykes (approved):
LGTM. Was able to build and do some minor testing with no obvious regressions. 

The only thing I can think of is adding some smaller tests but shouldn't stop this going through

slick-daddy:
Safe to merge 👍🏻

PR #7171, opened 2026-08-18
Fix: Handle Long Log Lines From Scrapers and Plugins

### LINKED ISSUE #7170: Plugins and scrapers break on long log lines

### Describe the bug

The way Stash reads stderr from script-based plugins and scrapers fails for log lines that contain 65536 bytes or more.

[PluginLogger.ReadLogMessages](https://github.com/stashapp/stash/blob/4c3d3a90f35c1656b89c4f3f13fc3e03b8431800/pkg/logger/plugin.go#L195-L211) reads plugin/scraper stderr with a bufio.Scanner which has a default token limit of bufio.MaxScanTokenSize (64KiB), and once it exceeds that limit it fails permanently: Scan() returns false with bufio.ErrTooLong and can never recover

Because the return value is treated as "end of stream", the loop exits and immediately runs src.Close(), closing the read end of the child process's stderr pipe. From there:

1. The child's next write to stderr gets EPIPE. In Python that surfaces as BrokenPipeError, which propagates out of the scraper
2. The scraper therefore dies before it can write its JSON result to stdout
3. Stash's decode in runScraperScript sees an empty stdout and reports could not unmarshal json from script output: EOF

This crashes the scraper or plugin and the error message in the logs is misleading because it just says `could not unmarshal json from script output: EOF` (since it never checks the error from the bufio.Scanner)

### Steps to reproduce

1. Create oversized.py in your scrapers directory
```py
import sys

print("\x01i\x02", "before the long line", file=sys.stderr, flush=True)
print("\x01d\x02", "X" * 70000, file=sys.stderr, flush=True)
print("\x01i\x02", "after the long line", file=sys.stderr, flush=True)
print('{"title": "it worked"}')
```
2. Create oversized.yml next to oversized.py in your scrapers directory
```yml
name: Oversized log line
sceneByURL:
  - action: script
    url:
      - example.com
    script:
      - python
      - oversized.py
```
3. Reload scrapers and then scrape `https://example.com` on any scene

### Expected behaviour

A scraper or plugin outputting long log lines should not kill the run. The long line could (should?) be truncated and marked rather than dropped, and if the reader bails out it should be logged rather than passing silently

I have fixed this for all the scrapers in the CommunityScrapers repo in [2a2149d](https://github.com/stashapp/CommunityScrapers/commit/2a2149de4241f51b175362fa40de286b98a851d2), which is nice because it will cover all versions of Stash even when this is fixed

It does not, however, protect people writing their own scrapers or plugins that do not rely on `py_common`

### Stash version

v0.31.1

Maista6969:
## Description

This changes the logging infrastructure used by scrapers and plugins to handle log lines that exceed 64KB by replacing the  bufio.Scanner with a bufio.Reader that reads their stderr in chunks. I could have raised the buffer size on the bufio.Scanner but that would just be moving the breakpoint instead of fixing the core issue.

## Related Issue
Resolves #7170 

## Testing
I have added a small test that exercises the fix and there is no change in behavior that scrapers or plugins need to be aware of.

## Checklist

## AI Usage Disclosure

### DIFF
--- a/pkg/logger/plugin.go
+++ b/pkg/logger/plugin.go
@@ -187,25 +187,57 @@ func (log *PluginLogger) handleStderrLine(line string) {
 	}
 }
 
+// maxLogLineLength is the maximum length of a single log line read from a plugin.
+// Longer lines are truncated to this length.
+const maxLogLineLength = 64 * 1024
+
+const truncatedSuffix = " [truncated]"
+
 // ReadLogMessages reads plugin log messages from src, forwarding them to the PluginLoggers Logger.
 // ProgressLevel messages are parsed as float64 and forwarded to ProgressChan. If ProgressChan is full,
 // then the progress message is not forwarded.
+// Log lines longer than maxLogLineLength are truncated.
 // This method only returns when it reaches the end of src or encounters an error while reading src.
 // This method closes src before returning.
 func (log *PluginLogger) ReadLogMessages(src io.ReadCloser) {
-	// pipe plugin stderr to our logging
-	scanner := bufio.NewScanner(src)
-	for scanner.Scan() {
-		str := scanner.Text()
-		if str != "" {
-			log.handleStderrLine(str)
+	defer src.Close()
+
+	// pipe plugin stderr to our logging.
+	// bufio.Scanner is unsuitable here: it aborts permanently on lines longer than
+	// its buffer, which would stop us reading the remainder of the plugin's output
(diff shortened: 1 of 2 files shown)


Gykes (approved):
Overall seems fine. The only thing I can think of is the logic might break if there are like emojis and stuff in which makes the byte calculations weird. I don't see there ever being that issue with logs but wanted to document it.

PR #7189, opened 2026-08-28
Avoid setting WAL mode on read connections

smith113-p:
## Description
I have found that I often get an error message on the scenes page saying that there is a database lock. Codex has discovered that even read-only connections set WAL mode, which is a database-wide setting that is not read-only and can trigger this situation.

Hence, this PR only sets WAL mode on read-write connections.

## Testing
`go test ./pkg/sqlite`

## Screenshots
n/a

## Checklist

- [X] I have read and understood the [Contributing](https://github.com/stashapp/stash/blob/develop/docs/CONTRIBUTING.md) document.
- [X] I have read and understood the [AI Usage Policy](https://github.com/stashapp/stash/blob/develop/docs/AI_POLICY.md) document.

## AI Usage Disclosure

- [X] I have used AI tools to assist with this pull request, and I have disclosed the tools and how I used them below.

I have used ChatGPT (formerly known as Codex) to figure out what might be causing frequent database locks, and it has identified this specific code and it wrote the CL for me.

The AI policy says that comments must not be written by AI although the comment in this CL seems pretty clear as it is and I don't see the value in rewriting it. But let me know if this is an issue.

### DIFF
--- a/pkg/sqlite/database.go
+++ b/pkg/sqlite/database.go
@@ -236,12 +236,26 @@ func (db *Database) Close() error {
 
 func (db *Database) open(disableForeignKeys bool, writable bool) (*sqlx.DB, error) {
 	// https://github.com/mattn/go-sqlite3
-	url := "file:" + db.dbPath + "?_journal=WAL&_sync=NORMAL&_busy_timeout=50"
+	url := db.connectionURL(disableForeignKeys, writable)
+
+	conn, err := sqlx.Open(sqlite3Driver, url)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


Maista6969:
Ask your bot if WAL is per-connection

smith113-p:
It's per-database, and because it is I have gotten more skeptical that this PR will fix anything (which is why I kept it in draft state).

Basically I am not sure that the underlying PRAGMA journal_mode will trigger the write lock if the DB is already in WAL mode.

Gykes:
Are we good to close this out?

PR #7190, opened 2026-08-28
Increase busy timeout to 5s

### LINKED ISSUE #7188: Frequent "database is locked" errors on the scenes page

### Describe the bug

Often, when I load the scenes page, I get an error that the database is locked.

Possibly related: my database seems kind of slow for reasons I don't understand

### Steps to reproduce

1. Load http://localhost:9999/scenes

### Expected behaviour

Scenes page should load

I am filing this issue primarily so I can reference is in PRs I am about to make. I am not really expecting anyone to be able to usefully investigate this with the level of detail I currently have, although I am hoping to debug this some more locally.

### Stash version

v0.31.1-160-gc1fa78e0

### Device details

Seen with Chrome and Firefox

### Relevant log output

```shell
nothing interesting in the log
```

smith113-p:
## Description

PR #3593 set the timeout to 50ms, implying that the default was 0 (i.e. that in case of a lock a query would fail immediately).

As best as I can tell, the default timeout has been five seconds for a very long time, see e.g.:
https://github.com/mattn/go-sqlite3/commit/d754d2db456cb912005948bfa82b7c06f3733274

So this commit just restores the timeout to the default 5s, which should reduce the number of "database locked" errors when there is a slow(ish) write operation running.

## Related Issue

Closes #7188 

## Screenshots

n/a

## Checklist

## AI Usage Disclosure

I used Codex/ChatGPT to find this code and understand context but the PR was written manually.

### DIFF
--- a/pkg/sqlite/database.go
+++ b/pkg/sqlite/database.go
@@ -236,6 +236,7 @@ func (db *Database) Close() error {
 
 func (db *Database) open(disableForeignKeys bool, writable bool) (*sqlx.DB, error) {
 	// https://github.com/mattn/go-sqlite3
+	// 5s timeout is the default.
 	url := "file:" + db.dbPath + "?_journal=WAL&_sync=NORMAL&_busy_timeout=5000"
 	if !disableForeignKeys {
 		url += "&_fk=true"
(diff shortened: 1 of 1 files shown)


Gykes (changes requested):
If you could add a quick comment as to why we are moving it back to 5s. just like "// 5s is default. 'commit number' ".

smith113-p:
> If you could add a quick comment as to why we are moving it back to 5s. just like "// 5s is default. 'commit number' ".

Thanks, done.

Gykes (approved):
Thanks for the comment. Checked out the commit and you seem to be correct. Not a whole lot of testing I can do here to verify this. I don't believe anything else could be broken by bumping the timeout.

PR #7193, opened 2026-08-31
New: Show Remote Duration in Tagger

### LINKED ISSUE #7201: Show remote duration in scene tagger

This would especially useful if the remote has a duration but no fingerprints (this can frequently happen with TPDB). Even when there are fingerprints, it can be useful to easily see exactly how the local duration differs from the remote duration.

Note that the "Duration is off by at least NNNs" is based on fingerprints, not the remote's duration field.

Easier to verify that the matched remote scene is correct

Have to click through to the stash ID to verify the duration

### Additional context

#4482 requested this as a secondary request, but that issue was closed when tagger started showing the local scene's duration.

smith113-p:
## Description
In scene tagger, if the stash box provided a duration, shows it on top of the scene image similar to how the local scene duration is displayed.

This is especially useful if the remote has a duration but no fingerprints (this can frequently happen with TPDB). Even when there are fingerprints, it can be useful to easily see exactly how the local duration differs from the remote duration.

## Related Issue

Fixes #7201 (this was also a secondary request in #4482)

## Testing
Manually scraped scenes with and without duration

## Screenshots
Before:
image

After: 
Screenshot 2026-08-31 142252

## Checklist

## AI Usage Disclosure

I used Codex/ChatGPT to write some of this code, especially the CSS. I did read and understand every line of the PR, though.

### DIFF
--- a/ui/v2.5/src/components/Tagger/scenes/StashSearchResult.tsx
+++ b/ui/v2.5/src/components/Tagger/scenes/StashSearchResult.tsx
@@ -19,6 +19,7 @@ import { Icon } from "src/components/Shared/Icon";
 import { SuccessIcon } from "src/components/Shared/SuccessIcon";
 import { LoadingIndicator } from "src/components/Shared/LoadingIndicator";
 import { TagSelect } from "src/components/Shared/Select";
+import TextUtils from "src/utils/text";
 import { TruncatedText } from "src/components/Shared/TruncatedText";
 import { OperationButton } from "src/components/Shared/OperationButton";
 import * as FormUtils from "src/utils/form";
@@ -553,11 +554,22 @@ const StashSearchResult: React.FC<IStashSearchResultProps> = ({
             }
             setExclude={(v) => setExcludedField(fields.cover_image, v)}
           >
-            <img
-              src={scene.image}
-              alt=""
-              className="align-self-center scene-image"
-            />
+            <div className="scene-image-overlay-container">
+              <img
+                src={scene.image}
+                alt=""
+                className="align-self-center scene-image"
+              />
+              {scene.duration ? (
+                <div className="scene-specs-overlay">
+                  <span className="overlay-duration">
+                    {TextUtils.secondsToTimestamp(scene.duration)}
+                  </span>
+                </div>
+              ) : (
+                ""
+              )}
+            </div>
           </OptionalField>
         </div>
       );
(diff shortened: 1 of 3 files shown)


smith113-p:
FYI here is how it looks when the cover image is landscape:
image

Here is how it looks when no duration is available:
image

ui/v2.5/src/components/Tagger/scenes/StashSearchResult.tsx
```diff
@@ -553,11 +554,22 @@ const StashSearchResult: React.FC<IStashSearchResultProps> = ({
             }
             setExclude={(v) => setExcludedField(fields.cover_image, v)}
           >
-            <img
-              src={scene.image}
-              alt=""
-              className="align-self-center scene-image"
-            />
+            <div className="scene-image-overlay-container">
```
Gykes, line 557: This is nested inside the cover image. If there is no cover image will it return nothing? Is that expected?

ui/v2.5/src/components/Tagger/scenes/StashSearchResult.tsx
```diff
@@ -553,11 +554,22 @@ const StashSearchResult: React.FC<IStashSearchResultProps> = ({
...
+              <img
+                src={scene.image}
+                alt=""
+                className="align-self-center scene-image"
+              />
+              {scene.duration ? (
+                <div className="scene-specs-overlay">
+                  <span className="overlay-duration">
```
Gykes, line 565: The "" branch would render and empty text node rather than nothing. I think returning null or just using && is a better way to do this. 

Something like so: 

```
 {scene.duration && (
   <div className="scene-specs-overlay">
     <span className="overlay-duration">
       {TextUtils.secondsToTimestamp(scene.duration)}
     </span>
   </div>
 )}
```

Gykes (changes requested):
Just a quick static check. No actual testing was done.

ui/v2.5/src/components/Tagger/scenes/StashSearchResult.tsx
smith113-p: Done. I copied this from https://github.com/stashapp/stash/blob/ebe326a25e0f0653e83e3d613af34447e232cbfe/ui/v2.5/src/components/Scenes/SceneCard.tsx#L379...

ui/v2.5/src/components/Tagger/scenes/StashSearchResult.tsx
smith113-p: Yeah I would say this is expected. If we have no image we would need a different kind of UI. I'm not immediately sure what that UI would look like, and it is rare to not have an image (rare on TPDB, almost never happens on stashdb), so it is probably not worth the code complexity.

smith113-p (commented):
sorry for the delay, see responses below

Gykes (approved):
LGTM, did some quick testing and worked as expected

PR #7199, opened 2026-09-02
Install all package requirements when updating a package

### LINKED ISSUE #7198: Updating scrapers does not install new requirements

### Describe the bug

I have been updating a lot of scrapers in the Community repository lately and several users have reported this bug to me. If a package is updated upstream in a way that it gains a new dependency, updating that package does _not_ install the new dependency which makes it seem broken. Take for example the Evil Angel metadata scraper which was recently updated to have the Altwolia package as a dependency.

This happens because the requirements for packages are only resolved in the "Available scrapers" window, where the resulting resolution is also used to show which transitive dependencies will be installed alongside whatever package the user chose to install: choosing Evil Angel here would include Altwolia (a direct dependency of Evil Angel) as well as py_common, a dependency of Altwolia.

The current workaround for this is telling users to install Altwolia manually.

### Steps to reproduce

1. Have an older version of a scraper like Evil Angel, xEmpire, Wicked, TabooHeat that depends on AlgoliaAPI installed
2. Update that scraper to a newer version that has removed its AlgoliaAPI dependency and gained an Altwolia dependency
3. Try using this scraper: it will fail to resolve Altwolia

### Expected behaviour

The `installPackage` GraphQL call should update dependencies as well as installing new dependencies. It should not remove old dependencies because there's no way to know if those are still relevant for other packages.

### Stash version

v0.31.1

Maista6969:
## Description
Resolves and installs package requirements in the backend itself. This is necessary to ensure updating packages get their full set of required packages if they were added in an update. This also makes it safe to naively install packages through the GraphQL API for tools that integrate with Stash such as [stash-scrape-ci-standalone](<https://github.com/feederbox826/stash-scrape-ci-standalone>)

## Related Issue
This resolves #7198 

## Testing
The easiest way to test this is to install a package that has multiple dependencies, for example the Evil Angel scraper mentioned in the original issue and this will pull in both the Altwolia and py_common packages. Immediately uninstall those two packages, then go edit the manifest for Evil Angel to modify its version to make Stash think it's outdated.

At this point we're essentially at the state of users who have updated Evil Angel on v0.31.1, the scraper does not work because the dependencies are missing.

But now you can update the package either through the API or the UI and it should be updated as expected, but it should also reinstall those missing packages you uninstalled earlier. At this point the scraper works, and users will no stop reporting this bug! :)

## Checklist

### DIFF
--- a/pkg/pkg/manager.go
+++ b/pkg/pkg/manager.go
@@ -151,6 +151,19 @@ func (m *Manager) getStore(remoteURL string) *Store {
 }
 
 func (m *Manager) Install(ctx context.Context, spec models.PackageSpecInput) error {
+	return m.install(ctx, spec, make(map[string]struct{}))
+}
+
+// installs the package identified by spec, then recursively installs
+// or updates any of its requirements that are missing or outdated. installing
+// tracks package IDs currently being installed in this call tree to guard
+// against dependency cycles
+func (m *Manager) install(ctx context.Context, spec models.PackageSpecInput, installing map[string]struct{}) error {
+	if _, inProgress := installing[spec.ID]; inProgress {
+		return nil
+	}
+	installing[spec.ID] = struct{}{}
+
 	remote, err := m.remoteFromURL(spec.SourceURL)
 	if err != nil {
 		return fmt.Errorf("creating remote repository: %w", err)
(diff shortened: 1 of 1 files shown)


pkg/pkg/manager.go
```diff
@@ -196,6 +209,44 @@ func (m *Manager) Install(ctx context.Context, spec models.PackageSpecInput) err
...
+		if err == nil {
+			// already installed - only reinstall if the source has a newer version
+			remotePkg, err := m.packageByID(ctx, reqSpec)
+			if err != nil {
+				return fmt.Errorf("getting remote package %s, required by %s: %w", reqID, pkg.ID, err)
+			}
+
+			if remotePkg == nil || !local.Upgradable(remotePkg.PackageVersion) {
+				continue
+			}
```
Gykes: When a required package is present and not upgradable the loop continues without recursing. i think this means it's own requirements are never checked, right? I don't believe we are walking the entire dependency tree.

PR #7203, opened 2026-09-03
Add warning for scraper URL conflicts

### LINKED ISSUE #7200: Warn about overlapping URL patterns in scrapers

Most scrapers are URL scrapers, and the way they work is by matching whatever URL is in the field for any scene/performer/gallery/image to a pattern declared in the scraper.

The problem arises when two or more scrapers can scrape the same pattern, a situation that can occur when scrapers are consolidated into larger scrapers that can cover multiple domains because they use the same CMS or belong to the same network. A recent example would be the new MojoHost scraper introduced in [CommunityScrapers 3bdeead](https://github.com/stashapp/CommunityScrapers/commit/3bdeead917fe8650e565a5a5b13363eacc9e5c56) that moved patterns from over a dozen scrapers into one big scraper which will be easier for us maintainers to keep updated.

Users who have the old FFCSH scraper installed can install the new MojoHost scraper and suddenly their scraping of fantasyflipflop.com becomes non-deterministic because sometimes Stash will use the new MojoHost scraper and sometimes it will use the old FFCSH scraper. Scraper definitions are stored in maps, and the [iteration order of Golang maps](https://go.dev/blog/maps#iteration-order) is not guaranteed (and is indeed random), which means the scraper used can and will vary.

This is incredibly confusing for users so I think we should warn them when two or more of their scrapers share URL patterns. I think most common cause of this problem is the aforementioned updating/consolidation of scrapers so we should suggest to users that the most probable correct resolution for this problem is to uninstall the older package, but not to do it automatically. 

A plugin could in theory iterate over all scrapers and their patterns and run this check in the browser, but in my opinion this should be in the core of Stash because it's not their fault that I love tinkering with the CommunityScrapers repo.

### Additional context

_No response_

Maista6969:
## Description

Adds a warning in the `Scrapers` list in `Settings > Metadata Providers` when a scraper has a pattern that overlaps with another installed scraper. Clicking the warning icon will open a modal that explains what the issue is and suggests resolving it by uninstalling one of the scrapers, heuristically suggesting that the older one is probably the least useful one. Usually a more recently updated scraper will be more correct, and hopefully that's what the user would want.

## Related Issue

This resolves #7200

## Testing
Install two scrapers from the Community feed that overlap: right now I've left intentionally overlapping scrapers ATK and ATKGirlfriends in the repository, but I'll fix them when/if this issue is addressed. There are also false positive matches like Reality Kings (which defines the pattern `rk.com`) and Digital Playground (which defines the pattern `digitalplaygroundnetwork.com` where the substring `rk.com` also appears)

## Screenshots
Before
image
After
image

This is the warning modal that appears when you click the warning icon
image

## Checklist

### DIFF
--- a/internal/api/resolver_model_scraper_url_conflict.go
+++ b/internal/api/resolver_model_scraper_url_conflict.go
@@ -1,0 +1,56 @@
+package api
+
+import (
+	"context"
+	"path/filepath"
+	"strings"
+
+	"github.com/stashapp/stash/pkg/scraper"
+)
+
+func (r *scraperURLConflictResolver) ScraperPackage(ctx context.Context, obj *scraper.ScraperURLConflict) (*ScraperURLConflictPackage, error) {
+	return findInstalledScraperPackage(ctx, obj.ScraperID)
+}
+
+func (r *scraperURLConflictResolver) OtherScraperPackage(ctx context.Context, obj *scraper.ScraperURLConflict) (*ScraperURLConflictPackage, error) {
+	return findInstalledScraperPackage(ctx, obj.OtherScraperID)
+}
+
+// ASSUMPTION: the scraper name is the same as the package name
+// if this assumption is not true, we safely return nil and leave it up to the user
+func findInstalledScraperPackage(ctx context.Context, scraperID string) (*ScraperURLConflictPackage, error) {
+	pm, err := getPackageManager(PackageTypeScraper)
+	if err != nil {
+		return nil, err
+	}
+
+	installed, err := pm.ListInstalled(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	for spec, manifest := range installed {
+		if spec.ID != scraperID {
+			continue
+		}
+
+		var siblings []string
+		for _, f := range manifest.Files {
+			if filepath.Ext(f) != ".yml" {
+				continue
+			}
+
+			id := strings.TrimSuffix(filepath.Base(f), ".yml")
+			if id != scraperID {
+				siblings = append(siblings, id)
+			}
+		}
+
+		return &ScraperURLConflictPackage{
+			Package:           manifestToPackage(manifest),
+			SiblingScraperIds: siblings,
+		}, nil
+	}
+
+	return nil, nil
+}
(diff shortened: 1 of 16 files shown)


DogmaDragon:
Could you document in https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Manual/Scraping.md that scraper list can be used in this manner. I feel like unless people know of it, they wouldn't really be able to take advantage of it.

Maista6969:
Thanks for the reminder, I've added a few lines where I think they make sense :slightly_smiling_face:

DogmaDragon (approved):
Documentation check passed.

internal/api/resolver_model_scraper_url_conflict.go
```diff
@@ -0,0 +1,56 @@
...
+
+// ASSUMPTION: the scraper name is the same as the package name
+// if this assumption is not true, we safely return nil and leave it up to the user
+func findInstalledScraperPackage(ctx context.Context, scraperID string) (*ScraperURLConflictPackage, error) {
+	pm, err := getPackageManager(PackageTypeScraper)
+	if err != nil {
+		return nil, err
+	}
+
+	installed, err := pm.ListInstalled(ctx)
```
Gykes: This is a N+1 issue I believe. This seems to get called once per resolved, which is twice per conflice, and each call does a `os.ReadDir` over the scraper dir plus YAML parse of every installed package manifest.

ui/v2.5/src/components/Settings/ScraperURLConflicts.tsx
```diff
@@ -0,0 +1,281 @@
...
+  const mineIsOlder = packageIsOlder(
+    conflict.minePackage,
+    conflict.otherPackage
+  );
+  const otherIsOlder = packageIsOlder(
+    conflict.otherPackage,
+    conflict.minePackage
+  );
+
+  async function confirmUninstall() {
```
Gykes: I don't see any error handing for this function. `AlertModal` calls it as `onClick={() => onConfirm()}` so a rejected mutation is a unhandled rejection, `setUninstallTarget(undefined)` is never reached so I think the modal would just stick open, no? I would add a try catch and a `useToast`

PR #7213, opened 2026-09-07
Chore: Make SceneScrapeDialog Patchable

dude-stash:
Requested in https://github.com/stashapp/stash/issues/4510#issuecomment-5564326675.

`SceneScrapeDialog` is exported but never registered, so it never appears in `PluginApi.components` and a plugin cannot render it. Listing it in `loadableComponents` doesn't help either - `loadComponents` awaits the import and discards the module's exports, so that only benefits components which register themselves on import.

This wraps the export in `PatchComponent`, the same way the other patchable components do, and adds it to the patchable list in the UI plugin API manual.

### Why

I'm writing a UI plugin that adds a stash-box scene search with the filters the built-in search doesn't expose (`queryScenes` by title/code/url/date/performers/tags/studio, and `queryPerformers(performed_with:)` for pairings). Applying a result to the local scene wants exactly the field-by-field dialog that "Scrape with..." already opens on the Edit tab. Without access to it a plugin has to either reimplement that dialog, which then drifts from this one, or skip the field sync and only write the `stash_id`.

### Impact on existing callers

None intended. `SceneEditPanel` imports it exactly as before, and `PatchComponent` is a pass-through when nothing is patched.

### Testing

Done so far:

- `pnpm exec biome check` clean on the changed file.
- `make ui-only` builds, and the registration is present in the emitted `SceneScrapeDialog` chunk.
- Exercise the native "Scrape with..." flow on a scene's Edit tab and confirm it is unchanged. [Video](https://www.loom.com/share/5ffb717d78304465810eb2567a2440c5)
- Exercise a plugin rendering `PluginApi.components.SceneScrapeDialog` end to end. [Video](https://www.loom.com/share/c18298ff62594e92b02400d1608a5970)

### AI disclosure

This change was written with AI assistance (Claude Code). The scope is the diff in this PR - the two-line registration and the manual list entry - plus the separate plugin that motivated it, which is not part of this repo. I have reviewed the change, understand what each line does and why, and have tested it as described above.


Gykes (approved):
LGTM

PR #7215, opened 2026-09-07
Auto-save confirmed scene edit fields (opt-in)

### LINKED ISSUE #4353: [Feature] Auto-save metadata fields that already include a confirmation
Manually adding a tag to a scene already requires quite a few clicks. There is already a feature to auto-save when changing toggles from the Settings page. Using this auto-save feature for adding Tags to a scene would reduce the work required to manually add a tag to a scene.

Under the Details tab on the Scene viewer, there is a dropdown to select new tags to add. Typing a tag name is not enough to add it, you need to also click or press enter, thus confirming that the tag name is what you want.

Even this confirmation is not enough, however - accepting a tag simplify activates a Save button back at the top that must be clicked to actually save the new tag. 

This second step could be replaced with the same auto-save behavior from the settings page.

A separate top-level tagging flow that pops up suggestions and allows the user to add tags easily.

**Additional context**
image
image
image
image

dude-stash:
## Description

Adds an opt-in "Auto-save on selection" setting under Settings → Interface → Editing (off by default). When enabled, adding or removing a tag, performer, studio, gallery, group, or stash-box ID selection on the scene edit form immediately saves the form, skipping the extra manual "Save" click. It **does not** apply to free-text fields or while creating a new (unsaved) scene. Applying a scrape result can also fire auto-save when tag, studio or performers are part of what was applied. Otherwise it still requires pressing the Save button.

When the setting is on, Save keeps the edit form mounted and dims it in place so scroll position is preserved. With the setting off, Save still uses the original full-panel spinner.

No backend/GraphQL schema changes — the setting is a client-side UI config field.

## Related Issue

Closes #4353

## Testing

`make validate-ui` (biome lint, biome format, stylelint, tsc) passed on this branch.

Manual pass against a running dev instance (backend on `:9999`, this branch's UI on `:3000`).

**Setting off (default) — original save behavior**

- Settings → Interface → Editing shows "Auto-save on selection", off by default.
- Selecting a tag, performer, studio, or group does not save until Save is clicked.
- Save still uses the original full-panel spinner: the form unmounts, then remounts.
- Groups picker is unchanged: adding a group removes it from the add list immediately; removing it puts it back.

**Setting on**

- Toggled the setting on and confirmed it persisted.
- On an existing scene edit tab, selecting a tag, performer, studio, or group saved immediately without clicking Save. Removing a performer or group also auto-saved.
- During that save the form stays mounted, dims, and ignores clicks; scroll position is kept. After save, an "Updated scene" toast appears.
- Groups keep the same picker behavior as with the setting off: an added group is no longer offered in the add list; a removed group is available again. (The in-place save path is what required a small extra fix here so that list would still refresh.)
- Confirmed via a GraphQL query against the dev database that the new tag was actually persisted, then reverted the test scene and turned the setting back off.

Free-text fields and creating a new scene were not auto-saved, as intended.

## Screenshots

A [video](https://www.loom.com/share/e8a8fe1412fe4ff89ede0f3c9e381884) is worth a thousand screenshots

## Checklist

## AI Usage Disclosure

Implemented with Claude Code (Anthropic): the feature implementation (frontend changes, setting/locale strings), the manual testing pass described above.

### DIFF
--- a/ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
+++ b/ui/v2.5/src/components/Scenes/SceneDetails/SceneEditPanel.tsx
@@ -28,6 +28,7 @@ import { addUpdateStashID, getStashIDs } from "src/utils/stashIds";
 import { useFormik } from "formik";
 import { Prompt } from "react-router-dom";
 import { useConfigurationContext } from "src/hooks/Config";
+import { IUIConfig } from "src/core/config";
 import { IGroupEntry, SceneGroupTable } from "./SceneGroupTable";
 import {
   faSearch,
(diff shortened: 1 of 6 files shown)


dude-stash:
@DogmaDragon I hope this time I got it right! :)

PR #7218, opened 2026-09-09
Scene time ranges: multiple scenes per file (backend)

597226617:
Backend portion of #3530, following the PR-1 scope WithoutPants laid out in the issue (ranges capability first; UI, streaming and generation follow in later PRs).

## Approach

Nullable `start_time`/`end_time` on the scene-file relation rather than touching the source file at scan time. Existing rows are untouched — a scene with NULL range behaves exactly as before, so no data migration for current installs. Primary scenes are unaffected by the constraint questions raised in the issue: ranges are only validated per scene-file pair, and shared files stay legal.

## What's in here

- migration 87 (`scenes_files.start_time`/`end_time`, nullable)
- repository layer: range-aware create/update, scan updates the ranges instead of duplicating rows
- GraphQL: scene `file { start_time, end_time }` exposed, mutations accept ranges
- tests: unit tests for the range queries (30 assertions), 4 integration scenarios (scan, re-scan, per-file ranges, shared file), 12-case invalid-input matrix
- DESIGN.md — schema decisions and the three spots where I deviated from the obvious approach, with reasons

## What's deliberately not in here

UI, streaming detection, and generation, per the issue discussion. The schema supports adding those without another migration.

Built and tested against go1.25.5, `go vet` clean. Deviations from the original design sketch are called out in DESIGN.md rather than hidden — happy to walk through any of them.


Gykes:
Please utilize the PR template.

597226617:
Saw this got closed — no worries if the scope or timing wasn't right. If there's a version of this you'd actually want (e.g. smaller, discussion-first, or a different schema direction), happy to rework. The issue thread's design discussion pointed at nullable ranges, which is what I built, but I might have read the PR-1 boundary wrong.

Gykes:
I just want you to use the PR template as I stated in my initial comment.

PR #7220, opened 2026-09-10
Fix Chromecast Integration on Stash app

### LINKED ISSUE #4136: [Bug Report] Can't cast any video to Chromecast

dude-stash:
## Description

Chromecast doesn't work: you press Cast, the device connects, but the TV only shows the Cast logo and nothing plays, with no error.

Two reasons:

1. Stash used the Video.js Chromecast *tech*. A tech is a player backend, so it sends the TV the same stream the browser is playing (Direct stream), and Stash reports that as `video/mp4` even when the file is MKV. Chrome accepts that; a Chromecast cannot play MKV at all, so it just waits.
2. The URL is `http://localhost:9999`. The Chromecast downloads the video itself, and for the Chromecast, `localhost` means the Chromecast.

This is why the workarounds in the issue help some people and not others: casting from an IP fixes the address, DNS tricks fix the hostname, but neither fixes the MKV problem.

**What this does:**

Stash now talks to the Google Cast framework directly from the scene player, so it can choose what to send:

- Pick the stream from the real video and audio codecs, not from the MIME type. Direct stream only for H.264 + AAC in mp4, otherwise the original transcode, otherwise HLS.
- Replace `localhost` with a LAN address of this machine. The browser can't read network interfaces, so the server reports them in a new `systemStatus.localIPs` field.
- Keep the page on HTTPS or `http://localhost`, since the Cast SDK won't start anywhere else — only the media URL changes.
- If a scene can't be cast, show an error instead of hanging.

Signed URLs from #6529 still work, so authentication isn't affected.

**UI-wise, nothing changes** — same button, same place, same icon, same setting. `@silvermine/videojs-chromecast` is removed (Stash draws the button itself now), which also removes the only UI lockfile dependency that wasn't from npm — it was a GitHub tarball that fails behind a strict proxy, so `pnpm install` now works there too.

**Known issue:** Chrome draws its own Cast button in the video's top-left corner that casts the tab, not the video. That comes from the browser, not Stash — `disableRemotePlayback` didn't hide it, so I left it for a follow-up.

**Next:** TV-side player controls (play/pause/seek) and simultaneous browser+TV playback, then AirPlay (same localhost problem).

## Related Issue

Fixes #4136.

## Testing

- 29 new unit tests for stream selection and URL rewriting (added Vitest, since the UI had no JS test runner; it now runs as part of `pnpm run validate`, so CI picks it up with no workflow change)
- `make test`: 34 packages, no failures
- `make it`: pass
- `golangci-lint` v2.11.4 (same version as CI): 0 issues
- `pnpm run validate`: biome, stylelint, tsc, format, tests all pass
- UI build: pass

Manual testing with a real Chromecast device:

- MKV scene: plays on the TV (first seconds take a while due to the transcode)
- H.264/AAC mp4 scene: plays with no transcode
- Chromecast setting toggled off: no cast button shown, `cast_sender.js` not loaded
- Control bar looks the same as before

## Screenshots

No visual changes — the Cast button keeps the same icon, placement, and behavior from the user's perspective.

## Checklist

## AI Usage Disclosure

I used Claude Code to help implement this feature and write the documentation in this PR. I reviewed, tested, and take responsibility for all of the code and the manual Chromecast testing described above.

## Additional Context

None.


docs/CHROMECAST.md
DogmaDragon, line 1: This doesn't belond here. Sounds like LLM musings.

ui/v2.5/src/docs/en/Manual/Interface.md
DogmaDragon, line 1: Mid-sentence line break should be removed. Also please write it in human language, not LLM verbose nonsense.

DogmaDragon (changes requested):

Issue #7247, opened 2026-09-20
VR videos get super bright and washed out when played in VR mode


GoldstarZV:
### Have you enabled troubleshooting mode?

- [x] I confirm that the troubleshooting mode is enabled.

### Describe the bug

When playing VR videos, in regular video player the video looks fine, but when switching to any of the videojs-vr player mode, colors and brightness are washed out and too bright. 

I have tried troubleshooting mode, in a different browser, and this behaviour was corroborated by [someone on discord. ](https://discord.com/channels/559159668438728723/1550917242349420575) 

### Steps to reproduce

1. Define a VR tag in Settings > Interface > Scene Player
2. Add a VR video to your stash scenes 
3. Give the new video the VR tag defined in step 1
4. In the bottom right corner of the video player, press on the VR headset button to switch to 180 lr (or any other VR mode) 
<img width="194" height="142" alt="Image" src="https://github.com/user-attachments/assets/f22afe6b-6ebf-4c3a-9ef0-276597218b63" />

5. Observe the video brightness becomes too high and colors become washed out
 

### Expected behaviour

Colors should remain consistent

### Screenshots or additional context

**Normal video player:**
<img width="2000" height="1000" alt="Image" src="https://github.com/user-attachments/assets/448ead51-0b83-4ed1-99f7-b8726f1afdf4" />

**Played in VR player (180 lr but it happens in all modes)**
<img width="2000" height="1038" alt="Image" src="https://github.com/user-attachments/assets/4466575d-2303-4919-98a7-b3f4a309d07e" />

### Stash version

v0.31.1-167-gb872e360 Build hash: b872e360

### Device details

Windows 11
Chrome Version 153.0.8010.48 (Official Build) (64-bit)
But also tested on Firefox 

### Relevant log output

```shell

```

GoldstarZV:
I think I have a fix for this, just putting up the PR now

PR #7248, opened 2026-09-20
Fix: VR player too bright & washed out colors

GoldstarZV:
## Description

Fixes a bug with videojs-vr which causes VR videos playing in VR mode to be too bright and desaturated. 

videojs-vr does not set a colour space. three.js therefore treats the already sRGB as linear and encodes it to sRGB again on output, so VR playback looks overly brightened and desaturated.

The fix is to tag the plugin's textures as sRGB after it creates them, on both `"initialized"` and `"loadedmetadata"`. the latter covers the branch where `init()` rebuilds the texture during an XR session and returns before firing `"initialized"`.

## Related Issue

https://github.com/stashapp/stash/issues/7247

## Testing

1. Built from develop with the fix and ran it against a real library
2. Opened a VR scene (4096×2048, 180° side-by-side) and played it in 180 LR
3. Compared with non-VR mode, blacks and saturation now match perfectly. 

## Screenshots

Video played in regular video player:
image

Video played in videojs-vr **in `development`** as of `b872e36`
image

Video played in videojs-vr **with this fix:**
image

## Checklist

- [ ] I have made corresponding changes to the documentation (**not** applicable).

## AI Usage Disclosure

Claude diagnosed and wrote the fix for this issue. I verified the fix myself and did my best to ensure the fix is reliable and every line of change is necessary. I wrote every line of text written in the issue as well as in this PR.


Gykes (changes requested):
General Thoughts: 

This seems more like an upstream bug that should be fixed there. If/when this gets fixed there this code will become dead/redundant. I'm okay with putting it in for now just because I have no idea when it will be fixed upstream but we need to document that this needs to be removed at some point. Please put a comment in the code stating that this is an upstream issue, this is a short term bandaid, and needs to be removed once fixed. 


Review:

Overall it looks fine just a small request. 

I don't believe this will fix the poster properly. It doesn't fire and event when the poster loads, only when the video starts to play. What you could do is catch the moment the plugin sets `posterTexture` and replace the plain property with a getter/setter right after `this.vr = this.player.vr():`. `fixTextureColorSpace()` would then only need to handle `videoTexture` and poster can come out of the loop.

Issue #6276, opened 2025-11-14
Part of the UI unclickable on mobile when viewing portrait scenes


farp33:
### Describe the bug

When viewing a scene in portrait mode on mobile, the non-videoplayer portion of the UI is inaccessible.

Tested on Android 14 and 15, in Chrome and Firefox Mobile

### Steps to reproduce

1. Open a scene with portrait ratio
2. Try click any UI element but the video player, for example the "markers" tab or performer portrait
3. Open a scene with landscape ratio
4. Notice the UI behaving as expected 

### Expected behaviour

Expected UI to be accessible. Ik looks like the video player element overlays the rest of the UI, as there seems to be an extra scrollbars visible 

### Screenshots or additional context

Screenshot taken with Firefox devtools. Purple line shows flex box boundary of the `video-wrapper`

<img width="445" height="823" alt="Image" src="https://github.com/user-attachments/assets/95640621-41ff-45df-9c1c-9eeacbe642d3" />

### Stash version

v0.29.3

### Device details

Firefox Mobile and Chrome on Android 14 & 15

### Relevant log output

```shell

```

feederbox826:
Unable to replicate on Android 15, by scrolling all the other elements are fully accessible, when the phone is in landscape or portrait

farp33:
I've been able to reproduce it in the Firefox Devtools responsive design mode on desktop. It looks like the `height: 177.78vw;` property causes the `video-wrapper` to overlay `scene-tabs`. 
https://github.com/stashapp/stash/blob/2332401dbfbcdbb943dac09e89e8f9ed87856a02/ui/v2.5/src/components/ScenePlayer/styles.scss#L21-L23

Turning it off seems to fix this issue, but I'm not sure it would break anything else though

farp33:
I've added a screenshot to the issue description, I hope this clarifies it a bit further

farp33:
The problem was caused by some custom CSS, I forgot I had it enabled. Closing the issue

Issue #6526, opened 2026-01-26
[Bug][Mobile][Portrait] Player bottom controls are clipped (fullscreen missing) + menus overlap/are occluded (z-index/stacking issue)


booth7850:
### Describe the bug

On Android mobile browsers in portrait orientation, the Stash web video player UI renders incorrectly 100% reproducibly on my setup:

The bottom control bar is clipped/truncated (right side of the controls is cut off), causing the fullscreen button to be missing / not visible.

Player popup menus (e.g. quality/source/settings menus) are overlapping / occluded by the player controls, and/or rendered partially off-screen.

There appears to be a z-index / stacking context issue where the menu and the bottom controls layer incorrectly, sometimes also resulting in tap/clicks being blocked.

### Steps to reproduce

1. Open Stash Web UI on an Android phone browser.

2. Open any Scene and start playback.

3. Keep the phone in portrait mode.

4. Tap the video to show the controls.

5. Observe the bottom control bar: the right side is clipped and the fullscreen button is not visible.

6. Open the menu (quality/source/settings); observe it overlaps/is occluded and may not be fully tappable.

### Expected behaviour

Bottom controls should be fully visible in portrait mode, including the fullscreen button.

Popup menus should be fully within the viewport and not overlap/occlude controls.

Layering should be correct (menus on top, controls below) and all menu items should be tappable.

### Screenshots or additional context

![Image](https://github.com/user-attachments/assets/315fa23b-1d8e-4ed3-a955-579194841674)

![Image](https://github.com/user-attachments/assets/b32a466e-01ed-4b9c-a94a-39e76278bd78)

### Stash version

v30.0.2

### Relevant log output

```shell

```

booth7850:
gen this custom css by gpt，hide unneeded button in portrait and ajust some zindex:
```
@media (max-width: 900px) and (orientation: portrait) {
  .video-js .vjs-playback-rate {
    display: none !important;
  }

  .video-js .vjs-subs-caps-button {
    display: flex !important;
    visibility: visible !important;
    opacity: 1 !important;
  }

  .video-js .vjs-control-bar {
    flex-wrap: nowrap !important;
  }
}
@media (max-width: 900px) and (orientation: portrait) {
  .video-js .vjs-control-bar {
    z-index: 3000 !important;
  }

  .video-js .vjs-touch-overlay,
  .video-js .vjs-big-play-button,
  .video-js .vjs-overlay {
    z-index: 1000 !important;
  }
[...]

PR #7249, opened 2026-09-20
fix: cap portrait video wrapper height to prevent click-blocking overlay

quietsignal44:
## Summary

Portrait/vertical videos set `.video-wrapper` height to `177.78vw`, which on wide viewports extends far below the visible video area (e.g. **3413px on a 1920px-wide screen**). This creates an invisible overlay that blocks clicks on scene tabs, rating buttons, and other UI elements positioned below or beside the player.

### Root cause

In `ui/v2.5/src/components/ScenePlayer/styles.scss` (lines 21-23):
```scss
&.portrait .video-wrapper {
    height: 177.78vw;
}
```

The `177.78vw` value (100vw × 16/9) is intended to make portrait videos fill the viewport width at the correct aspect ratio, but there is no upper bound. The wrapper's box extends past the visible video (which is clipped by `overflow: hidden`) and intercepts pointer events on elements below.

### Fix

Add `max-height: calc(100vh - 4rem)` so the wrapper never exceeds the available viewport height:
```scss
&.portrait .video-wrapper {
    height: 177.78vw;
    max-height: calc(100vh - #{$menuHeight});
}
```

This preserves the existing behavior for viewports where `177.78vw` fits, while capping the height on wider screens where it would otherwise overflow.

### Related issues

- #6276 — "Part of the UI unclickable on mobile when viewing portrait scenes" (closed, but the underlying CSS issue remains in the codebase)
- #6526 — Player z-index/stacking issues (related stacking context problem)

https://claude.ai/code/session_01S4siNHnHXrvxFCU7uYaka3


quietsignal44:
@WithoutPants @Gykes — Would appreciate your eyes on this one. It's a one-line CSS fix for portrait/vertical videos creating an invisible overlay that blocks clicks on scene tabs and rating buttons below the player.

The root cause is the `height: 177.78vw` on `.VideoPlayer.portrait .video-wrapper` with no upper bound — on wider viewports the wrapper extends thousands of pixels past the visible video area. The fix adds `max-height: calc(100vh - 4rem)` to cap it.

Related to the closed #6276 (same symptoms, but the underlying CSS issue was never addressed).

